PR #741, opened 2015-08-13
Selenium longer per-test timeout

cy6erGn0m:
Increase time out to 5s because 1s for the whole test suite is too strong restriction + minor code cleanup


bashor:
ok

cy6erGn0m:
Merged manually

PR #745, opened 2015-08-26
kotlin builder: print 'label in local history' message to log with 'debug' level and once per build

chashnikov:
Currently the same line is printed to build.log two times for each module, i.e. >1000 times in case of IDEA project.


bashor:
ok, please merge manually

chashnikov:
Merged to 'master'.

PR #750, opened 2015-09-08
Kotlin compile daemon 2

ligee:
Next variant of compiler daemon (following the PR#743 - https://github.com/JetBrains/kotlin/pull/743)
Important changes to the first version:
- new instances management scheme - separate daemon for each compiler (determined by compiler classpath)
- reporting to compiler output, controlled by command line options, off by default
- many refactorings, security improvements, and cleanups

Known issues:
- Build cancellation support is not implemented yet


compiler/rmi/kotlinr/src/org/jetbrains/kotlin/rmi/kotlinr/KotlinCompilerClient.kt
```diff
@@ -18,246 +18,376 @@ package org.jetbrains.kotlin.rmi.kotlinr
...
+        return (localId.compilerVersion.isEmpty() || localId.compilerVersion == remoteId.compilerVersion) &&
+               (localId.compilerClasspath.all { remoteId.compilerClasspath.contains(it) }) &&
+               (localId.compilerDigest.isEmpty() || remoteId.compilerDigest.isEmpty() || localId.compilerDigest == remoteId.compilerDigest)
+    }
 
-                compilerId.updateDigest()
+
+    class FileBasedLock(compilerId: CompilerId, daemonOptions: DaemonOptions) {
+        private val lockFile: File =
+                File(daemonOptions.runFilesPath,
```
cy6erGn0m: do we need to keep them in home dir? can we use tempdir instead to eliminate possible issues caused by NFS-backed homedir, etc ?

compiler/rmi/kotlinr/src/org/jetbrains/kotlin/rmi/kotlinr/KotlinCompilerClient.kt
```diff
@@ -18,246 +18,376 @@ package org.jetbrains.kotlin.rmi.kotlinr
...
+        }
 
-            val daemon = connectToCompileService(compilerId, daemonLaunchingOptions, daemonOptions, System.out, autostart = !clientOptions.stop, checkId = !clientOptions.stop)
+        private val channel = if (locked) RandomAccessFile(lockFile, "rw").channel else null
+        private val lock = channel?.lock()
 
-            if (daemon == null) {
-                if (clientOptions.stop) println("No daemon found to shut down")
-                else throw Exception("Unable to connect to daemon")
+        synchronized private fun acquireLockFile(lockFile: File): Boolean {
```
cy6erGn0m: Why do we have `:Boolean` function that always returns `true` ?

compiler/rmi/kotlinr/src/org/jetbrains/kotlin/rmi/kotlinr/KotlinCompilerClient.kt
```diff
@@ -18,246 +18,376 @@ package org.jetbrains.kotlin.rmi.kotlinr
...
+                // Ignoring it - assuming it is another client process owning it
             }
+            var attempts = 0L
+            while (lockFile.exists() && attempts++ * COMPILE_DAEMON_STARTUP_LOCK_TIMEOUT_CHECK_MS < COMPILE_DAEMON_STARTUP_LOCK_TIMEOUT_MS) {
+                Thread.sleep(COMPILE_DAEMON_STARTUP_LOCK_TIMEOUT_CHECK_MS)
+            }
+            if (lockFile.exists())
+                throw IOException("Timeout waiting the release of the lock file '${lockFile.absolutePath}")
+
+            return lockFile.createNewFile()
```
cy6erGn0m: Can't we replace it with something like that:
```kotlin
do {
    if (lockFile.createNewFile()) return true
    Thread.sleep(...)
} while (!timeout)

throw timeout exception
```

It is better because the function has only two cases: return or exception. Now we have true/false/exception

compiler/rmi/rmi-interface/src/org/jetbrains/kotlin/rmi/LoopbackSocketFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.io.Serializable
+import java.net.InetAddress
+import java.net.ServerSocket
+import java.net.Socket
+import java.rmi.server.RMIClientSocketFactory
+import java.rmi.server.RMIServerSocketFactory
+
+
+// TODO switch to InetAddress.getLoopbackAddress on java 7+
+val loopbackAddrName by lazy { if (java.net.InetAddress.getLocalHost() is java.net.Inet6Address) "::1" else "127.0.0.1" }
```
cy6erGn0m: `java.net.InetAddress.getLocalHost()` may crash with IOException on such machines due to strange network configuration. In some cases it couldn't be fixed on some corporate machines. I faced it few times in different companies. It is possible at least on Linux and Solaris machines

jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.java
```diff
@@ -128,36 +128,71 @@ private static String execCompiler(
...
+            }
+
+            if (daemon != null) {
+                ByteArrayOutputStream compilerOut = new ByteArrayOutputStream();
+                ByteArrayOutputStream daemonOut = new ByteArrayOutputStream();
+
+                Integer res = KotlinCompilerClient.incrementalCompile(daemon, argsArray, incrementalCaches, compilerOut, daemonOut);
+
+                ProcessCompilerOutput(messageCollector, collector, compilerOut, res.toString());
+                BufferedReader reader = new BufferedReader(new StringReader(daemonOut.toString()));
```
cy6erGn0m: will we see non-latin characters? it could be significant for error messages with path to files, for example

compiler/rmi/rmi-server/src/org/jetbrains/kotlin/rmi/service/CompileServiceImpl.kt
```diff
@@ -39,54 +39,111 @@ import kotlin.concurrent.read
 import kotlin.concurrent.write
 
 
+fun nowSeconds() = System.nanoTime() / 1000000000L
+
 class CompileServiceImpl<Compiler: CLICompiler<*>>(
         val registry: Registry,
```
cy6erGn0m: How long this thing work? Does it shutdown itself?

compiler/rmi/kotlinr/src/org/jetbrains/kotlin/rmi/kotlinr/KotlinCompilerClient.kt
ligee: Taking into account that I need recognisable file names and use parts of the names for discovery, it seems that straightforward approach with temdir has several potential security flaws, to fix which I'd need a platform-specific code with native function calls (at least until an access to a robust way to create a secure temp subdirs, implemented in java 7).
Another problem is that e.g. Idea substitutes "java.io.tmpdir", so it is no longer points to the system one, which with current mechanism makes sharing daemons between various clients impossible.
Homedir solution probably is not ideal, but seems safe enough, and there is a fallback possibility to specify run directory for daemons.

compiler/rmi/rmi-interface/src/org/jetbrains/kotlin/rmi/LoopbackSocketFactory.kt
ligee: Ok, thanks, I'll wrap it into try/catch accordingly.

cy6erGn0m: You also can run `host` utility in catch block

compiler/rmi/kotlinr/src/org/jetbrains/kotlin/rmi/kotlinr/KotlinCompilerClient.kt
ligee: There were no false case in fact. But I rewrote it a bit differently, may be it is more clear and robust now

compiler/rmi/rmi-interface/src/org/jetbrains/kotlin/rmi/LoopbackSocketFactory.kt
ligee: This is a bit too much for this case, imho. I even don't need localhost address as such, just the default addressing scheme.

jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.java
ligee: At least I do not see how unicode support could be broken here, provided that String supports it, and all streams used here are not touching contents.

compiler/rmi/rmi-server/src/org/jetbrains/kotlin/rmi/service/CompileServiceImpl.kt
ligee: What is the root of the questions? What concerns you have here?
The daemon may stay alive arbitrary long time, but current default idle timeout is 2 hours, after that it shuts itself down.

idea/src/org/jetbrains/kotlin/idea/compiler/configuration/KotlinBuildProcessParametersProvider.kt
```diff
@@ -17,12 +17,22 @@
...
+public class KotlinBuildProcessParametersProvider(private val compilerWorkspaceSettings: KotlinCompilerWorkspaceSettings,
+                                                  private val kotlinPluginStartupComponent: PluginStartupComponent
+): BuildProcessParametersProvider() {
     override fun getVMArguments(): MutableList<String> {
-        return if (compilerWorkspaceSettings.incrementalCompilationEnabled)
-            arrayListOf()
-        else
-            arrayListOf("-Dkotlin.incremental.compilation=false")
+        val res = arrayListOf<String>()
+        if (compilerWorkspaceSettings.incrementalCompilationEnabled) {
```
bashor: Is it ok that behavior changed?

ligee: No, it is not ok. Thank you!

jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.java
cy6erGn0m: Ah I see: as far as `daemonOut.toString()` uses `Charset.defaultCharset()` it seems to should work

PR #753, opened 2015-09-22
	Fix for KT-8020.

abond:


ilya-g:
https://youtrack.jetbrains.com/issue/KT-8020 is going to be fixed in Kotlin/JS compiler in 1.1.4

PR #757, opened 2015-09-29
Local delegated properties

dotlin:
This PR adds support for local delegated properties on both JVM/JavaScript platforms save for top-level variables in scripts


abreslav:
Thanks for your contribution! We will review and integrate it as soon as we can, but I'm afraid, it may take some time

max-kammerer:
Big thanks! Merged manually.

PR #758, opened 2015-10-02
compile service 4

ligee:
fixing jps hanging on windows
fixing compilation failures due to too many connections ("error during JRMP connection establishment")
adding support for JS compilation on daemon
detailed performance reporting
performance improvements:
- connection caching in JPS
- reducing number of rpc callbacks from daemon


ligee:
merged into master

PR #764, opened 2015-10-10
Replace @HiddenDeclaration with @Deprecated(level = HIDDEN), support @Deprecated(level = ERROR)

abreslav:


compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java
```diff
@@ -288,8 +288,8 @@ public String render(@NotNull BadNamedArgumentsTarget target) {
 
         MAP.put(MANY_COMPANION_OBJECTS, "Only one companion object is allowed per class");
 
-        MAP.put(DEPRECATED_SYMBOL, "''{0}'' is deprecated.", DEPRECATION_RENDERER);
-        MAP.put(DEPRECATED_SYMBOL_WITH_MESSAGE, "''{0}'' is deprecated. {1}", DEPRECATION_RENDERER, STRING);
+        MAP.put(DEPRECATION, "''{0}'' is deprecated. {1}", DEPRECATION_RENDERER, STRING);
+        MAP.put(DEPRECATION_ERROR, "Calling ''{0}'' is an error. {1}", DEPRECATION_RENDERER, STRING);
```
udalov: I think this error message looks weird on classes in type positions...

abreslav: Good point. Will change to "Using ..."

Have you looked through other changes?

compiler/frontend/src/org/jetbrains/kotlin/resolve/deprecationUtil.kt
```diff
@@ -0,0 +1,103 @@
...
+
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.descriptors.*
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget
+import org.jetbrains.kotlin.descriptors.annotations.Annotations
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.resolve.annotations.argumentValue
+
+private val JAVA_DEPRECATED = FqName(Deprecated::class.java.name)
```
udalov: This was corrupted by the refactoring and is now in fact "kotlin.Deprecated". Let's use a hard-coded string instead.

core/descriptors/src/org/jetbrains/kotlin/resolve/annotationsForResolve.kt
```diff
@@ -16,14 +16,9 @@
...
 
-import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.descriptors.annotations.Annotated
 import org.jetbrains.kotlin.name.FqName
 
-private val HIDDEN_ANNOTATION_FQ_NAME = FqName("kotlin.HiddenDeclaration")
-
-public fun DeclarationDescriptor.isAnnotatedAsHidden(): Boolean = annotations.findAnnotation(HIDDEN_ANNOTATION_FQ_NAME) != null
-
 private val NO_INFER_ANNOTATION_FQ_NAME = FqName("kotlin.NoInfer")
```
udalov: I think this whole file should be moved to `frontend`...

compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java
udalov: Yes

core/descriptors/src/org/jetbrains/kotlin/resolve/annotationsForResolve.kt
abreslav: Can/t do. This is used in descriptors

abreslav:
Bootstrapped and pushed two new commits that finish the transition

PR #768, opened 2015-10-14
Drothmaler/stdlib string to char sequence

drothmaler:
Hi I did put some effort into changing many stdlib methods to support CharSequence instead of String.
Pleas see if you like it :-)
This would also fix [KT-8651](https://youtrack.jetbrains.com/issue/KT-8651) and parts of [KT-9031](https://youtrack.jetbrains.com/issue/KT-9031).


ilya-g:
merged manually

PR #769, opened 2015-10-16
Rr/compile service 6

ligee:
fixing problems related to parallel compilation and enabling it back
daemon lifetime control updates
removing redundant dependencies (cherrypicket from appropriate fix in M15


ligee:
application environment lifetime change breaks too many tests

PR #773, opened 2015-10-21
Add toBigInteger extenstion to Numbers

alien11689:


HaasJona:
:-1: This implementation results in overflows for large `double`s and `float`s.  Either restrict this function to integer types (would be preferred by me) or do something like this for doubles and floats:

`public fun Number.toBigInteger() = BigDecimal(doubleValue).toBigInteger()`

Also you need to be careful that `bigInteger.toBigInteger()` doesn't convert to a Long in between which would result in overflow again (ideally just should return `this`).

PR #775, opened 2015-10-27
Allow dependency versions to be omitted for Gradle plugin

DanielThomas:
This allows the plugin to provide the Kotlin version (if omitted) to `org.jetbrains.kotlin` dependencies:

```
dependencies {
  compile 'org.jetbrains.kotlin:kotlin-stdlib'
}
```

Avoids having to resort to extension properties to keep the version consistent in the build (and once the plugin is in the Gradle plugin portal, will simplify the example to a plugins block and a dependencies block).

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
@@ -1,40 +1,42 @@
 package org.jetbrains.kotlin.gradle.plugin
 
-import org.gradle.api.Plugin
-import org.gradle.api.Project
-import org.gradle.api.tasks.SourceSet
-import org.jetbrains.kotlin.gradle.internal.KotlinSourceSetImpl
-import org.gradle.api.internal.project.ProjectInternal
-import org.gradle.api.internal.HasConvention
-import org.jetbrains.kotlin.gradle.internal.KotlinSourceSet
-import java.io.File
-import org.gradle.api.Action
-import org.gradle.api.tasks.compile.AbstractCompile
 import com.android.build.gradle.BaseExtension
-import com.android.build.gradle.api.AndroidSourceSet
 import com.android.build.gradle.BasePlugin
+import com.android.build.gradle.api.AndroidSourceSet
 import com.android.build.gradle.internal.variant.BaseVariantData
 import com.android.build.gradle.internal.variant.BaseVariantOutputData
-import org.gradle.api.artifacts.dsl.DependencyHandler
-import org.gradle.api.artifacts.ConfigurationContainer
-import org.gradle.api.initialization.dsl.ScriptHandler
-import org.jetbrains.kotlin.gradle.plugin.android.AndroidGradleWrapper
-import javax.inject.Inject
-import org.gradle.api.file.SourceDirectorySet
-import kotlin.properties.Delegates
-import org.gradle.api.tasks.Delete
 import groovy.lang.Closure
+import org.gradle.api.Action
+import org.gradle.api.Plugin
+import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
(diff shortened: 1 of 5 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -248,6 +250,18 @@ abstract class AbstractKotlinPlugin @Inject constructor(val scriptHandler: Scrip
...
         })
     }
+
+    private fun configureResolutionStrategy(project: Project) {
+        val projectVersion = loadKotlinVersionFromResource(project.logger)
+        project.configurations.all(Action<Configuration> { configuration ->
+            configuration.resolutionStrategy.eachDependency { details ->
+                val requested = details.requested
+                if (requested.group.equals("org.jetbrains.kotlin") && requested.version.isEmpty()) {
+                    details.useTarget("${requested.group}:${requested.name}:$projectVersion")
```
cy6erGn0m: Why do you put `project.version` here? shouldn't you somehow lookup Kotlin plugin version here?

cy6erGn0m:
it shouldn't be so much changes in the imports section

libraries/tools/kotlin-gradle-plugin/src/test/resources/testProject/kotlinProjectOmitVersion/src/main/kotlin/helloWorld.kt
```diff
@@ -0,0 +1,22 @@
...
+class KotlinGreetingJoiner(val greeter : Greeter) {
+
+    val names = ArrayList<String?>()
+
+    fun addName(name : String?): Unit{
+        names.add(name)
+    }
+
+    fun getJoinedGreeting() : String? {
+        val joiner = Joiner.on(" and ").skipNulls();
```
cy6erGn0m: why do we need guava in such simple example? we can replace ugly `Joiner.on(" and ").skipNulls()` with `names.filterNotNull().joinToString(" and ")`. We also don't need `com.google.common.primitives.Ints` at all...

DanielThomas: I just copied the kotlinProject test so there's something meaningful in there.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
DanielThomas: I just followed the existing convention for the variable name - can change it to `kotlinVersion` if you prefer?

DanielThomas:
Re: imports - I just organized them w/ IntelliJ. Is that not desirable?

cy6erGn0m:
@DanielThomas it's a bit strange that so much changed, ok, let it be

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
cy6erGn0m: here it is not clear which project version it will be: kotlin's version or user's project version. So I think you better to rename it and ensure it actually contains valid version (you can change sample project's version to some particular value (for example, 1.0.0))

h0tk3y:
@DanielThomas Thanks for the pull request! We've [manually taken the changes](https://github.com/JetBrains/kotlin/commit/ae087e4fe582aeba59b66ca930e11642f071f6f6) you propose, therefore I'm closing the PR.

DanielThomas:
@h0tk3y That's awesome! Would love to do away with https://github.com/nebula-plugins/nebula-kotlin-plugin completely - I'll have to PR the other features (setting jvm target version and adding the right version of stdlib implicitly based on the Java target version).

PR #776, opened 2015-10-28
Extract DOM utilities from stdlib

cy6erGn0m:
- remove all DOM-related things out of stdlib
- put corresponding declarations to kotlin-js library to keep compatibility for now
- remove deprecated declarations, including outdated JS declarations
- fix related and unrelated tests
- remove corresponding tests

TopLevelFromStandardLibraryWithoutParam.kt - removed as discussed with Valentin K

**do not merge (will merge manually later)**

Merge should be in-sync with kotlinx.dom publication


cy6erGn0m:
Merged manually

PR #778, opened 2015-11-02
rename "Jet" from names of IDE classes; consistently use Kt prefix fo…

yole:
…r PSI elements and classes directly related to specific PSI elements


yole:
Merged manually.

PR #779, opened 2015-11-02
#KT-6274 Fixed

mcgee:
I think it is very convenient to add interface ``KtValVarKeywordProvider`` for the classes who have ``getValOrVarKeyword`` method in order to fix the bug. However I was forced to add unsafe cast to ``AutomaticVariableRenamer.kt``.

I also fixed ``DataFlowInMultiDeclInFor.kt`` test because it became failing due to my changes. This test case looks a little bit strange now because ``var`` keyword is not allowed in ``for`` loops.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/KtValVarKeywordProvider.java
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/KtValVarKeywordProvider.java
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.psi;
 import com.intellij.psi.PsiElement;
 import org.jetbrains.annotations.Nullable;
 
-public interface KtVariableDeclaration extends KtCallableDeclaration, KtWithExpressionInitializer, KtValVarKeywordProvider {
-    boolean isVar();
+public interface KtValVarKeywordProvider extends PsiElement {
+    @Nullable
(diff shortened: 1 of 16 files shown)


abreslav:
Thanks for your contribution, we'll review it soon!

compiler/frontend/src/org/jetbrains/kotlin/psi/KtValVarKeywordProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.psi;
+
+import com.intellij.psi.PsiElement;
+import org.jetbrains.annotations.Nullable;
+
+public interface KtValVarKeywordProvider extends PsiElement {
```
abreslav: "Provider" is not a great name for it, I think

mglukhikh: I would suggest "KtValVarOwner" instead

mglukhikh:
Ok, I will do a pair of minor changes myself and then push your commit into the main repo. Thanks again for your contribution!

mglukhikh:
Edited and pushed manually

PR #780, opened 2015-11-03
Rr/compile service 7

ligee:
Numerous reliability improvements around the daemon
Enabling parallel compilation without the daemon
Some refactorings


ligee:
Merged into master

PR #782, opened 2015-11-06
#KT-5115 Fixed

mcgee:
https://youtrack.jetbrains.com/issue/KT-5115


mcgee:
I've fixed the comments except for the one about case with the only setter. If you still don't agree with me after explanation, I'll remove it.

mcgee:
I've fixed all comments except for the case with the only setter. If you still don't agree with me that it could be useful case, I'll remove it.

mcgee:
Sorry, I amended the commit, so comments are not visible in this pull request anymore. They can be seen here: https://github.com/mcgee/kotlin/commit/60cd5ae131f134dae9ae831b17fb6fe2e4b74fee

yole:
Thanks for your contribution!

PR #784, opened 2015-11-16
#KT-9893 Fixed

mcgee:
It was hard and I hope that I've understood the problem correctly

### DIFF
--- a/compiler/testData/diagnostics/tests/substitutions/recurisveStarProjectionsNullability.kt
+++ b/compiler/testData/diagnostics/tests/substitutions/recurisveStarProjectionsNullability.kt
@@ -1,0 +1,10 @@
+public interface Rec<T : Rec<T>> {
+    public fun foo(): T?
+}
+
+fun acceptRec(<!UNUSED_PARAMETER!>rec<!>: Rec<*>) {
+}
+
+fun checkRec(rec: Rec<*>) {
+    acceptRec(<!TYPE_MISMATCH!>rec.foo()<!>)
+}
--- a/compiler/testData/diagnostics/tests/substitutions/recurisveStarProjectionsNullability.txt
+++ b/compiler/testData/diagnostics/tests/substitutions/recurisveStarProjectionsNullability.txt
@@ -1,0 +1,11 @@
+package
+
+public fun acceptRec(/*0*/ rec: Rec<*>): kotlin.Unit
+public fun checkRec(/*0*/ rec: Rec<*>): kotlin.Unit
+
+public interface Rec</*0*/ T : Rec<T>> {
+    public open override /*1*/ /*fake_override*/ fun equals(/*0*/ other: kotlin.Any?): kotlin.Boolean
+    public abstract fun foo(): T?
+    public open override /*1*/ /*fake_override*/ fun hashCode(): kotlin.Int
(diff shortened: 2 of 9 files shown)


dzharkov:
Hi!

Thank you for contribution, your commit seems very reasonable, but as far as I can see it opens some problems with recursive cases again. For example:
```
interface A<T : A<T?>?> {
    fun foo(): T?
}
fun testA(a: A<*>) {
    a.foo()
}
```
Compilation of such code fails with StackOverflowError if you're changes are applied.
I'm not sure but think that such problems could be solved by making type of star projection lazy in a sense that calculation of it's type constructor and arguments is deferred, and when they are needed, then next recursive substitution step should be done.

mcgee:
Thank you for the new test case. I'll think about your lazy type suggestion and try to implement it.

mcgee:
As far as I understand, star projection is already lazy and it causes original bug (KT-9893). It is "too lazy" and substitutes star projection "as is" in place of nullable ``T``

There are 2 conflicting cases in which ``TypeSubstitutor`` should behave differently:
1. Star projection tries to determine type of itself. It is needed to be sure that star projection's getType() will not be called inside, so ``TypeSubstitutor`` should substitute star projection "as is" 
2. Substitution of star projection into nullable function return type T?. In this case ``TypeSubstitutor`` definitely should check return type on nullability.

As a solution I've added parameter to ``TypeSubstitutor`` in order to control its behavior for star projections. I've amended the commit so you can check it out.

dzharkov:
Current solution seems to be better, but IMO it still looks fragile.
In particular I would not like adding new field to `TypeSubstitutor` class.

My point was that in current version of `StarProjectionImpl` it defers calculation of type as whole object, while such calculation may lead to infinite recursion, so we have to place `isStarProjection` check almost everywhere we retrieve type from projection.

But we could make type for `StarProjectionImpl` lazy in different aspect, instead of deferring itself type
calculation we defer calculation of it's contents.

You can see something similar here: https://github.com/JetBrains/kotlin/blob/32151c077e3718851f7a3ca9a55913099d95c2a0/compiler/frontend/src/org/jetbrains/kotlin/resolve/TypeResolver.kt#L85

By doing this we win in a sense that calling `StartProjectionImpl.getType` NEVER leads to StackOverflow and even getting it's contents won't be unsafe anymore (because nested occurrences of projection will be calculated as lazy entities in the same way).

After such changes a lot of `isStarProjection` usages become redundant, and what's important here,
we can remove this check:
```
if (replacement.isStarProjection() && (substituteStarProjectionAsIs || !type.isMarkedNullable())) {
   return replacement;
}
```
and therefore to solve #KT-9893

dzharkov:
@mcgee I'm sorry, but I think, that idea my from previous post can be dangerous too.
Please let me some time to think about our solution

mcgee:
@dzharkov I suppose this pull request can be closed because you fixed the issue in https://github.com/jetbrains/kotlin/commit/e2c02f825f0bc812e48fd692ca657c30e87b47a4

I see that you fixed the issue in more general way. Unfortunately, I cannot grasp your fix right now. Despite the fact I have contributed nothing, it was interesting experience. I hope one day I will be able to solve problems like that. Thanks for your patience :)

PR #787, opened 2015-11-18
Rr/compile service 8

ligee:
New daemon control
Cleanup after compilation (fixes e.g. jar locking issues)
Fixing log-based daemon tests on windows


ligee:
Merged into beta 3

bashor:
how about master?

ligee:
It was merged into master before branching beta 3. With the message I meant that it is in the mainline since beta 3.

bashor:
ok

PR #788, opened 2015-11-20
KT-10114 Maven: plugin doesn't recognize sourceDirs configured in the goal configuration

cy6erGn0m:
As far as Maven doesn't provide any standard way to configure multiple source locations we have our own configuration entries in kotlin-maven-plugin. However IDEA doesn't recognize them so the source roots are not autodetected. This is a custom `MavenImporter` implementation that does the trick. 

See build status: https://teamcity.jetbrains.com/viewType.html?buildTypeId=bt345&tab=buildTypeStatusDiv&branch_Kotlin=rr%2Fcy%2Fmaven-import

Perhaps to be merged manually after tests and discussion

There is also an issue discovered in IDEA. See https://youtrack.jetbrains.com/issue/IDEA-148280 for details. this maven importer workarounds the issue but not completely.


cy6erGn0m:
@yole all tests seems to be passed

yole:
Feel free to merge.

cy6erGn0m:
Merged manually

PR #789, opened 2015-11-26
Rr/cy/extract testing

cy6erGn0m:
Extract tests from stdlib

- added kotlin-test maven module
- added kotlin-test to kotlinc dist (required for IDEA tests) and as a module dependencies where required
- removed assert/fail usages from PROD code
- some tests were migrated to use another functions rather than `assertNNN` while the others now pick up kotlin-test.jar as a dependency 

remote builds: https://teamcity.jetbrains.com/viewType.html?buildTypeId=bt345&branch_Kotlin=rr%2Fcy%2Fextract-testing&tab=buildTypeStatusDiv


cy6erGn0m:
@udalov haha, yes, thank you. I was looking at kotlin-reflect when was introducing kotlin-test.jar. Will fix

cy6erGn0m:
merged manually

udalov:
^ BTW manifest name still isn't fixed

cy6erGn0m:
@udalov ops, fixed in master https://github.com/JetBrains/kotlin/commit/c5e43a132767af71331126cb98981d9318e6e3cd

PR #790, opened 2015-11-27
New attempt to fix tests when parallel compilation is enabled

ligee:


udalov:
I don't like this approach because the order of tests may now have a subtle effect on their outcomes, is there any other solution?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/Properties.kt
```diff
@@ -17,3 +17,11 @@
 package org.jetbrains.kotlin.cli.common
 
 public val KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY = "kotlin.environment.keepalive"
+
+
+fun String?.toBooleanLenient(): Boolean? = when (this?.toLowerCase()) {
```
MichaelNedzelsky: It seems it is possible to simplify conditional expressions by using the following function instead of `toBooleanLenient`
```
fun String.systemPropertyAsBooleanOrTrueOtherwise(negate: Boolean): Boolean {
    val propValue = System.getProperty(this) ?: return true

    return when (propValue.toLowerCase()) {
        in listOf("", "yes", "true", "on", "y") -> !negate
        in listOf("no", "false", "off", "n") -> negate
        else -> true
    }
}
```
See comments below.

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt
```diff
@@ -257,7 +258,7 @@ public class KotlinCoreEnvironment private constructor(
             val appEnv = getOrCreateApplicationEnvironmentForProduction(configuration, configFilePaths)
             // Disposing of the environment is unsafe in production then parallel builds are enabled, but turning it off universally
             // breaks a lot of tests, therefore it is disabled for production and enabled for tests
-            if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY) == null) {
+            if (!(System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY).toBooleanLenient() ?: false)) {
```
MichaelNedzelsky: This could be written as
```
if (KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY.systemPropertyAsBooleanOrTrueOtherwise(negate = true)) {
```

compiler/tests/org/jetbrains/kotlin/repl/AbstractReplInterpreterTest.kt
```diff
@@ -45,6 +47,13 @@ public abstract class AbstractReplInterpreterTest : UsefulTestCase() {
 
     private data class OneLine(val code: String, val expected: String)
 
+    override fun setUp() {
+        super.setUp()
+        // set to false if not present or set to true or unknown (empty considered true)
+        if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY)?.let { it.toBooleanLenient() ?: true } ?: true )
```
MichaelNedzelsky: This could be written as
```
if (KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY.systemPropertyAsBooleanOrTrueOtherwise(negate = false)) {
```

idea/idea-test-framework/src/org/jetbrains/kotlin/idea/test/KotlinLightCodeInsightFixtureTestCase.kt
```diff
@@ -63,6 +65,10 @@ public abstract class KotlinLightCodeInsightFixtureTestCase : LightCodeInsightFi
                 super.processError(message, t, details, logger)
             }
         })
+
+        // set to false if not present or set to true or unknown (empty considered true)
+        if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY)?.let { it.toBooleanLenient() ?: true } ?: true )
```
MichaelNedzelsky: This could be written as
```
if (KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY.systemPropertyAsBooleanOrTrueOtherwise(negate = false)) {
```

compiler/tests/org/jetbrains/kotlin/repl/AbstractReplInterpreterTest.kt
MichaelNedzelsky: Also, I suggest to store previous value of property and restore it in `tearDown`.

ligee:
There is is another solution - find out how to properly detect testing mode in our ApplicationEnvironment and dispose stuff in tests using that detection. But it seems require much more effort from my side, since I'm not familiar with that part, and much more testing as well, since it may affect many parts. So I'd prefer if this solution will be accepted for now.
Will it look better if I'll implement storing and restoring the value of the property on setup and shutdown respectively, as proposed by Michael Nedzelsky?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/Properties.kt
ligee: I do not like the suggestion - looks too cumbersome and hiding the logic. But may be it will look better if combined with storing and restoring of the property value

compiler/tests/org/jetbrains/kotlin/repl/AbstractReplInterpreterTest.kt
ligee: That's a good one, thanks!

udalov:
> find out how to properly detect testing mode in our ApplicationEnvironment

`ApplicationManager.getApplication().isUnitTestMode()`?

> Will it look good enough if I'll implement storing and restoring the value of the property on setup and shutdown respectively, as proposed by Michael Nedzelsky?

I'm OK with it, although it's still not perfect because `tearDown` is not called if a test fails in `setUp`

compiler/tests/org/jetbrains/kotlin/scripts/ScriptTest.java
```diff
@@ -41,12 +41,21 @@
...
 
 public class ScriptTest {
+
+    //@Before
+    //public void setUp() throws Exception {
+    //    // set to false if not present or set to true or unknown (empty considered true)
+    //    if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY)?.let { it.toBooleanLenient() ?: true } ?: true )
+    //    System.setProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY, "false")
+    //}
+
```
udalov: Why is this needed here?

ligee:
> >  find out how to properly detect testing mode in our ApplicationEnvironment

> ApplicationManager.getApplication().isUnitTestMode()?

This is how it is implemented now (MockApplication.java:85):
```java
 public boolean isUnitTestMode() {
   return true;
 }
```

PR #791, opened 2015-11-27
#KT-9938 Fixed

mcgee:


mcgee:
Thanks for your comments

Fixed and amended the commit

mcgee:
Sorry, false alarm, I forgot to fix resources and I need to fix some tests.

mcgee:
Ok, I think I'm ready now 

What have been done:
* Check if caret position is on initializer
* Check if property is not an extension one
* Check if property is not a local variable or top level declaration
* Registered in ``plugin.xml``
* Added resources to ``idea/resources/intentionDescriptions`` directory

mcgee:
@yole fixed, I've made an additional commit, it should be easier to review 

I can melt them into one if everything ok

mcgee:
@yole, ok, done

yole:
Merged. Thanks! Sorry that it took so long; to make this faster, consider joining #kontributors in the Kotlin Slack

mcgee:
@yole, yay! :tada:
I've joined #kontributors a while ago

PR #793, opened 2015-12-01
N+k attempt to fix parallel compilation

ligee:
enabling parallel compilation
keepalive property controls destruction of the environment
controlling keepalive property in compiler use sites, but in jps taking parallel compilation option into account, so it should be now not set in tests

### DIFF
--- a/jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.kt
+++ b/jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.kt
@@ -17,6 +17,7 @@
 package org.jetbrains.kotlin.compilerRunner
 
 import com.intellij.util.xmlb.XmlSerializerUtil
+import org.jetbrains.jps.api.GlobalOptions
 import org.jetbrains.kotlin.cli.common.ExitCode
 import org.jetbrains.kotlin.cli.common.KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY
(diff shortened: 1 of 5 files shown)


jps-plugin/src/org/jetbrains/kotlin/compilerRunner/KotlinCompilerRunner.kt
```diff
@@ -117,9 +118,8 @@ public object KotlinCompilerRunner {
                 val stream = ByteArrayOutputStream()
                 val out = PrintStream(stream)
 
-// Uncomment after resolving problems with parallel compilation and tests
-//                if (System.getProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY) == null)
-//                    System.setProperty(KOTLIN_COMPILER_ENVIRONMENT_KEEPALIVE_PROPERTY, "")
+                if (java.lang.Boolean.parseBoolean(System.getProperty(GlobalOptions.COMPILE_PARALLEL_OPTION, "false")))
```
udalov: use `String#toBoolean()`

ligee: I'm using exactly the code used in idea now to avoid different interpretation.

udalov: `String#toBoolean()` is just a Kotlin way of saying `Boolean.parseBoolean`, look at its implementation

ligee: ok

PR #794, opened 2015-12-06
Changed kotlin.test.fail method's signature to return Nothing.

RodrigoQuesadaDev:
Hi, I just updated kotlin.test.fail to return Nothing as that's a better implementation (e.g. it allows you to have a function like this one: *fun getTheFreakingValue(): String = value ?: fail("Value is null!")*).


RodrigoQuesadaDev:
I just created ticket [KT-10289](https://youtrack.jetbrains.com/issue/KT-10289) to track this small PR. :)

PR #796, opened 2015-12-12
[stdlib]Add foldIndexed and reduceIndexed groups of functions

borgesgabriel:
I asked about the lack of those functions in the stdlib channel. No one came up with a reason for the absence, so I put together a PR.

- `foldIndexed`, `foldRightIndexed`, `reduceIndexed` and `reduceRightIndexed` have been added, in line with `filterIndexed` etc.;
- Test cases added appropriately for the new functions.

### DIFF
--- a/libraries/stdlib/src/generated/_Arrays.kt
+++ b/libraries/stdlib/src/generated/_Arrays.kt
@@ -8010,6 +8010,105 @@ public inline fun <R> ShortArray.fold(initial: R, operation: (R, Short) -> R): R
 }
 
 /**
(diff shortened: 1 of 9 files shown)


borgesgabriel:
@ilya-g, would you like me to break this down into smaller commits -- one for each of the functions proposed along with its test cases?

ilya-g:
No, it's not required, the implementation is fine, but we need some time to discuss internally about inclusion of these functions in the stdlib.

ilya-g:
Thanks, Gabriel, I've merged this PR manually.

PR #798, opened 2015-12-17
KT-8275  Unterminated multi-line comment should be compilation error (+cosmetics)

geevee:


compiler/frontend/src/org/jetbrains/kotlin/parsing/KotlinParsing.java
```diff
@@ -112,9 +112,20 @@ void parseFile() {
...
 
+        checkUnclosedBlockComment();
         fileMarker.done(KT_FILE);
     }
 
+    private void checkUnclosedBlockComment() {
+        if (TokenSet.create(BLOCK_COMMENT, DOC_COMMENT).contains(myBuilder.rawLookup(-1))) {
+            int startOffset = myBuilder.rawTokenTypeStart(-1);
+            int endOffset = myBuilder.rawTokenTypeStart(0);
+            if (!myBuilder.getOriginalText().subSequence(startOffset, endOffset).toString().endsWith("*/")) {
```
abreslav: This may materialize a huge file aa a single string (in case of this file being commented entirely)

We could probably use other means, such as:

* changing the lexer to emit a special token type on unfinished comments
* using the CharSequence API to only materialize the last two characters as String, and not the entire text

compiler/frontend/src/org/jetbrains/kotlin/parsing/KotlinParsing.java
```diff
@@ -112,9 +112,20 @@ void parseFile() {
...
+                error("Unclosed comment");
```
abreslav: This error is placed before the comment, in Java it is at the end of the file

abreslav: Looks like this fixes it:

```
                PsiBuilder.Marker marker = myBuilder.mark();
                marker.error("Unclosed comment");
                marker.setCustomEdgeTokenBinders(WhitespacesBinders.GREEDY_RIGHT_BINDER, WhitespacesBinders.GREEDY_RIGHT_BINDER);
```

But I think the Java version uses some other technique

abreslav:
@geevee What are your plans regarding this?

geevee:
I would like to finish it next week.

On Tuesday, 19 January 2016, Andrey Breslav <notifications@github.com>
wrote:

> @geevee <https://github.com/geevee> What are your plans regarding this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JetBrains/kotlin/pull/798#issuecomment-172847311>.
>


-- 
-- 
Evgeny Gerashchenko

abreslav:
@geevee It's becoming urgent, RC is close, so let us know if you won't be working on it after all

abreslav:
@geevee Any news?

geevee:
Since I have access to main kotlin repo no more, I have created new pull request with my fork. See #813

compiler/frontend/src/org/jetbrains/kotlin/parsing/KotlinParsing.java
geevee: Thank you, I have rewritten to materialize last two characters as String.

compiler/frontend/src/org/jetbrains/kotlin/parsing/KotlinParsing.java
geevee: Thank you! I have used your idea. Java doesn't have check for unclosed comments in parser at all. It is performed in `HighlightUtil` instead.

PR #803, opened 2015-12-21
#KT-10313 Fixed

mcgee:
https://youtrack.jetbrains.com/issue/KT-10313


udalov:
I think you can just do something like this:
```
    value.put(parameterDescriptorType, codegen.v);
    StackValue.coerce(parameterDescriptorType, parameterType, codegen.v);
```

No checks for primitives are needed, `put` invokes `coerce` which handles that by itself.

Also please add a box (`BlackBoxCodegenTestGenerated`) test which would produce ClassCastException now and remove the bytecode text test. We're trying to add tests on the bytecode text only when absolutely necessary, because they are more fragile than box tests.

Thanks!

mcgee:
Checks for primitives were added because of SAM conversions tests were failing. Code without checks tries to cast box to ``kotlin.jvm.functions.Function0``
I was not sure that there were no more cases that my code could brake. So I added strict checks that covered implicit number conversions only.

I can go another way and check only for SAM. Something like this:
```
if (codegen.bindingContext.get(SAM_VALUE, argumentExpression) != null) {
    value.put(parameterType, codegen.v)
}
else {
    value.put(parameterDescriptorType, codegen.v)
    StackValue.coerce(parameterDescriptorType, parameterType, codegen.v)
}
```

Will it be correct?

udalov:
I guess it's OK if there's no easy way to solve it on another level.

max-kammerer:
Vararg and inline cases missed:
class Box<T>(val s: T)

inline fun <T> calc(z: T): Box<T> {
    return Box<T>(z)
}

fun <T> run(vararg z: T) {    
}

     calc<Long>(-1)
     run(-1, -1, -1)
}

max-kammerer:
And super constructor calls:

open class Base<T>(val s: T)
class Box(): Base<Long>(-1)

udalov:
After an offline discussion with @max-kammerer, it became clear that the problem is reproducible only for constants which are not instances of `KtConstantExpression`. For the latter, `ExpressionCodegen#visitConstantExpression` works fine by generating the value computed eariler by the front-end of the compiler. Non-constants are fine because you can't pass an `Int` where `Long` is expected and vice versa.

An expression such as `-1` is a unary expression and codegen ends up in `ExpressionCodegen#visitUnaryExpression` where it doesn't then try to lookup the constant value like it does in `visitConstantExpression`, although that value is computed and is stored in the corresponding slice of the binding context. The same happens for binary expressions in `visitBinaryExpression`.

The solution we've came up with is to manually check for evaluated compile time constants in those methods (just like in `visitConstantExpression`) and avoid the normal code generation which would box the primitive in a wrong wrapper class. Besides unary and binary expressions we should also consider labeled and parenthesized expressions and maybe something else I overlooked which `ConstantExpressionEvaluatorVisitor` is able to process but codegen is ignoring.

The major problem of that approach is that, if implemented as described, it would start to inline values of Kotlin `const` properties sometimes (values of `const` properties are currently *not* inlined and that is correct, we're not sure we should do that yet). For simple expressions like `x` (where `x` is declared as `const val x = 0`) codegen ends up in `ExpressionCodegen#visitSimpleNameExpression` where it eventually creates a `StackValue.Property` instance which is able to prevent inlining in case when it's a Kotlin `const` val, that is fine and will continue to work. But if you have an expression like `-x`, you can't unconditionally get its constant value and put it to the stack like `visitConstantExpression` does, because that would change semantics, we want invokevirtual + lneg still generated here. Note that it's relevant only for Kotlin properties, Java fields should be inlined.

One workaround to this would be of course to manually check if an expression which has an evaluated constant value contains a Kotlin property referenced somewhere in it, with a custom `KtVisitor`. This sounds pretty fragile to me but it's the best that we could think of. Maybe there are other solutions, if you find one, please share. Overall this issue turned out to be a rather difficult one from my point of view, feel free to withdraw if you feel the same way :)

mcgee:
Wow :confused: 
OK, please give me some time to grasp your comment and figure out what I can do.

mcgee:
Okay, so...

I extracted compile time constant in both ``ExpressionCodegen#visitBinaryExpression`` and ``ExpressionCodegen#visitPrefixExpression``. I saw that compiler generated ``LDC`` instruction and correct Long boxing instruction. That's great.

About inlining of Kotlin ``const``'s: it's worse :) 
The compiler inlines even non-const reference expressions. Consider example:
```
val a = false
val b = false
val c = false
if (!(a && b && c)) {
    "then"
} else {
    "else"
}
```
It generates this:
```
L0
  LINENUMBER 2 L0
  ICONST_0
  ISTORE 0
L1
  LINENUMBER 3 L1
  ICONST_0
  ISTORE 1
L2
  LINENUMBER 4 L2
  ICONST_0
  ISTORE 2
L3
  LINENUMBER 5 L3
L4
  LINENUMBER 6 L4
  LDC "then"
  POP
```
So it seems to be sensible to make custom visitor that looks for property references. Thus I am trying to figure out how to write it for now. 

However, may be there is a simpler way. Front-end already knows that ``foo<Long>(1 + 2)`` should pass typecheck and ``foo<Long>(1 + x)`` shouldn't. May it be possible to use the same mechanism in backend instead of custom visitor?

mcgee:
Anyway, even if I implement looking for any reference visitor the problem with too aggressive inlining still exists. Consider example:
```
if (!(1 < 2)) {
    val p = 1
}
```
Compiler produces following code:
```
L0
  LINENUMBER 2 L0
L1
  LINENUMBER 3 L1
L2
  LINENUMBER 5 L2
  RETURN
```
:neutral_face:

udalov:
@mcgee Sorry for the delay. It seems that the issues you mentioned have been fixed in the compiler, would you be willing to continue working on this issue? Thanks!

mcgee:
Yes, I would like to continue

mcgee:
@udalov As far as I understand the issue about inlining of
```
if (!(1 < 2)) {
    val p = 1
}
``` 
wasn't fixed. Test ``negateConstantCompare.kt`` still fails. I've committed my changes, you can check that I've done as you suggested.

max-kammerer:
@mcgee, Could you rebase your commits on current master?

Both mentioned cases was fixed in master branch:
```val a = false
val b = false
val c = false
if (!(a && b && c)) {
    "then"
} else {
    "else"
}```
 and 
```
if (!(1 < 2)) {
    val p = 1
}
```

**Comment by mcgee** [comment]
@max-kammerer, I merged ``upstream/master`` to my branch before. But it's okay, I've rebased now to be 100% sure. The second case (``negateConstantCompare.kt``) still fails, I'm sorry

File: compiler/backend/src/org/jetbrains/kotlin/codegen/CallGenerator.kt [hunk 1]
```diff
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.codegen
 import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.resolve.calls.model.ResolvedCall
+import org.jetbrains.kotlin.codegen.binding.CodegenBinding.*
```
max-kammerer: Unused import

compiler/testData/codegen/boxWithJava/casts/literalExpressionAsGenericArgument/literalExpressionAsGenericArgument.kt
```diff
@@ -0,0 +1,5 @@
+fun box(): String {
+    val sub = Box<Long>(-1)
+    println(sub.value == 1L)
```
max-kammerer: replace 'println' with if check

max-kammerer:
I just noticed that we already have some additional information about constant expressions: check CompileTimeConstant.Parameters class. Seems 'ReferenceExpressionChecker' could be replaced with 'pure' check

mcgee:
@max-kammerer, thanks for your comments. I've fixed them

However I still have the problem with 4 failing tests (including ``negateConstantCompare.kt``). Moreover, I think it is pretty logical that ``!(1 < 2)`` is inlined. To be honest, I don't see the reason why it should not be inlined. Could you explain me that?

max-kammerer:
@mcgee, yes you are right, 'negateConstantCompare.kt' test is about 'check' and 'goto' instructions optimization. To keep this test you can replace '1' and  '2' with const vals. What about another 3 test?

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3975,4 +3981,19 @@ private NonLocalReturnInfo(Type type, String name) {
             labelName = name;
         }
     }
+
+    private StackValue getCompileTimeConstant(@NotNull KtExpression expression) {
```
max-kammerer: Combine this method with original one by adding checkPure flag, and create overload with original signature to keep old invocations

mcgee:
@max-kammerer, another 3 is 
``bytecodeText/contains/byte.kt``:
```
val a: Byte = 1 + 1

// 1 I2B
```
``bytecodeText/contains/short.kt``:
```
val a: Short = 1 + 1

// 1 I2S
```
``bytecodeText/contains/nullableByteAndShort.kt``:
```
val a: Byte? = -1
val b: Short? = -1

// 1 I2B
// 1 I2S
```

Am I right that this 3 also could be resolved by const vals?

mcgee:
Ahh, sorry, that won't compile

mcgee:
@max-kammerer, I've fixed your comments, thanks for good advice!

max-kammerer:
@mcgee, thanks for your pull request. I merged it mannually. I also updated codegen to generate byte and short constants instead of 'ldc' instuctions (they replaced I2B and I2S instuctions in tests): https://github.com/JetBrains/kotlin/commit/384d2ea0d121c57a503d43e9edafe0722d9508af

mcgee:
@max-kammerer, thanks for your patience

PR #804, opened 2015-12-23
Remove which dependency from startup scripts

NeQuissimus:
I would like to get rid of the startup scripts' dependency on `which`.
Some lightweight Linux distributions do not ship with the command by default (in my case NixOS) and Kotlin seems to require it as a dependency only for its startup scripts.

Rather than relying on `dirname` and `which`, we should be able to just use built-in `bash` functionality.

I wrote a short script that verifies equivalent functionality:

```bash
echo $0
echo "$(dirname "$(which "$0")")" #Current version
echo "${0%/*}" #Proposed version
```

```
~> ./paths.sh
./paths.sh
.
.
~> /Users/tsteinbach/paths.sh
/Users/tsteinbach/paths.sh
/Users/tsteinbach
/Users/tsteinbach
~> ln -s /Users/tsteinbach/paths.sh /Users/tsteinbach/Desktop/paths.sh
~> /Users/tsteinbach/Desktop/paths.sh
/Users/tsteinbach/Desktop/paths.sh
/Users/tsteinbach/Desktop
/Users/tsteinbach/Desktop
```

The behaviour for direct access and via symbolic links is identical.

### DIFF
--- a/compiler/cli/bin/kotlin
+++ b/compiler/cli/bin/kotlin
@@ -16,4 +16,7 @@
 
 export KOTLIN_RUNNER=1
 
-"$(dirname "$(which "$0")")"/kotlinc "$@"
+DIR="${BASH_SOURCE[0]%/*}"
+: ${DIR:="."}
+
+"${DIR}"/kotlinc "$@"
--- a/compiler/cli/bin/kotlinc-js
+++ b/compiler/cli/bin/kotlinc-js
@@ -16,4 +16,7 @@
 
 export KOTLIN_COMPILER=org.jetbrains.kotlin.cli.js.K2JSCompiler
 
-"$(dirname "$(which "$0")")"/kotlinc "$@"
+DIR="${BASH_SOURCE[0]%/*}"
+: ${DIR:="."}
+
+"${DIR}"/kotlinc "$@"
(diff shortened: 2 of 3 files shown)


udalov:
This seems to break when the usage does not contain a slash, i.e. please try `. paths.sh` or `source paths.sh`. Do you maybe know the solution to that problem?

Also `which` is so basic I couldn't imagine it's missing on some systems. Even MinGW has it. I'm not sure how NixOS is even usable without `which`? :)

NeQuissimus:
NixOS is very much usable :) You should check it out ;)


And you are right, I was not anticipating somebody sourcing the file.
This is the correct way of doing it, I believe:


New paths.sh

```
echo $0
echo "$(dirname "$(which "$0")")" # Current version
echo "${0%/*}" # Won't work with sourcing

DIR="${BASH_SOURCE[0]%/*}"
: ${DIR:="."}
echo "${DIR}" # Equivalent to current version
```

Output

```
~> ./paths.sh
./paths.sh
.
.
.
~> source paths.sh
paths.sh
.
paths.sh
.
~> . paths.sh
paths.sh
.
paths.sh
.
~> ~/Desktop/paths.sh #symlink
/Users/tsteinbach/Desktop/paths.sh
/Users/tsteinbach/Desktop
/Users/tsteinbach/Desktop
/Users/tsteinbach/Desktop
```

udalov:
Sorry for the delay! I've merged your commit manually: https://github.com/JetBrains/kotlin/commit/7cad65a1e707c73f594e1ae8cf9c75b609f177c5

Thank you!

PR #807, opened 2016-01-04
Apply inline fun use() to AutoCloseable

lukaseder:
The Java 7 try-with-resources statement operates on all `AutoCloseable` resources, not just the `Closeable` (legacy) ones. E.g. JDBC resources are `AutoCloseable`, not `Closeable`


yole:
We can't accept this pull request because the Kotlin stdlib needs to be compilable against Java 6, which doesn't have AutoCloseable. We do plan to support stdlib multitargeting in future versions of Kotlin, which will allow us to address this problem propertly. See https://youtrack.jetbrains.com/issue/KT-5899 for the issue tracking this.

lukaseder:
Thanks for the feedback, I wasn't aware of this dependency.

PR #809, opened 2016-01-14
Rr/build lib 2

ligee:
- a new module - build(lib) - containing jps-independent parts of incremental compilation supporting classes and utilities
- jps-plugin switched to the build module
- gradle plugin switched to the build module with basic (so far) implementation of incremental compilation support


bashor:
Why do you use module.xml file for gradle?

ligee:
The build lib (build-common) parts are merged into master, after some rework by bashor, the rest (gradle plugin changes) will be extracted into separate branch

PR #810, opened 2016-01-19
handle IndexNotReadyException in KtFile#getClasses

vsch:
otherwise Plugin causes IDE exceptions when index updates are triggered.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/KtFile.java
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/KtFile.java
@@ -21,6 +21,7 @@ import com.intellij.lang.ASTNode;
 import com.intellij.lang.FileASTNode;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.fileTypes.FileType;
+import com.intellij.openapi.project.IndexNotReadyException;
 import com.intellij.psi.*;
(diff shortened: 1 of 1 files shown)


yole:
I don't think that't the right fix for the issue. Silently returning incorrect data is not the way index updates are supposed to be handled.

vsch:
@yole, I started to write almost identical wording on the PR comment but what had me stumped is that the call is coming in from IDE projectView ClassesTreeStructureProvider on the EDT thread. So delaying the return until the indexes are computed is not really an improvement, neither is re-throwing the exception or leaving it as is, because it generates exceptions with frequent regularity.

I would check if we are in the EDT thread and ignore the exception, since the UI will be updated when indexing is complete and if not then rethrow the exception so the caller can properly handle the index not ready exception. 

Is this the accepted practice for these types of issues in the IDE?

yole:
The correct fix for the project view exception is to mark IDEA ClassesTreeStructureProvider as non-DumbAware.

vsch:
Thanks, I'll do that.

vsch:
@yole, ClassesTreeStructureProvider is DumbAware but not checking if indexes are ready at that point in the code.

Is an acceptable solution to catch the IndexNotReadyException and throw a ProcessCanceledException so the tree updater can properly handle it?

PR #812, opened 2016-01-28
Fix dead links.

hotchemi:
Just fix some dead links.


yole:
These links are not dead. The file is interpreted by Dokka, and the links are the way they need to be for Dokka-generated documentation to work correctly.

PR #814, opened 2016-01-30
Fix KT-10821 Many small drops on a Sequence cause huge slowdown

sargunster:
https://youtrack.jetbrains.com/issue/KT-10821


ilya-g:
Thanks for PR, I've merged it manually

PR #815, opened 2016-01-31
Fixes KT-1494 Inspection to highlight missing documentation for public declarations

manas-chaudhari:
https://youtrack.jetbrains.com/issue/KT-1494

Currently, this checks for explicitly public declarations. Need suggestions on handling these scenarios:
- Should overriden public elements also be highlighted? What if the element in super class is protected and it has been made public in overriding class?
- When visibility is not specified, it is public implicitly. In this case should inspection report problems?

### DIFF
--- a/idea/idea-analysis/src/org/jetbrains/kotlin/idea/inspections/IntentionBasedInspection.kt
+++ b/idea/idea-analysis/src/org/jetbrains/kotlin/idea/inspections/IntentionBasedInspection.kt
@@ -122,15 +122,15 @@ abstract class IntentionBasedInspection<TElement : KtElement>(
             editor?.caretModel?.moveToOffset(startElement.textOffset)
             intention.applyTo(startElement as TElement, editor)
         }
-
-        private fun PsiElement.findExistingEditor(): Editor? {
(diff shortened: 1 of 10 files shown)


yole:
* Since docs are inherited in Kotlin, you should not report an overridden public method if the base class method has documentation.
* If an element was protected and is now public, it needs documentation.
* The "public" keyword is highlighted as redundant in Kotlin, so you definitely should report problems where the visibility is implicitly public.

manas-chaudhari:
Have taken care of above scenarios. Can proceed to review.

Another issue which bothers me:
- `main` gets highlighted for Missing Documentation. I am not sure if it is a good idea to make it an exception. If exceptions need to be made, are there other such keywords?

manas-chaudhari:
Thanks. Will send over an update asap.

manas-chaudhari:
Updated. Please check

manas-chaudhari:
Have fixed the duplicate method issue. Also added a test for quickfix.

yole:
Thanks for your contribution! I've squashed the commits in this PR, disabled the inspection by default and merged the PR.

PR #817, opened 2016-02-03
Add support for AOSP's nullability annotations.

JakeWharton:
These are different than 'android.support.annotation' in that they are used on the JVM for projects within AOSP.


dzharkov:
I think you forgot to commit generated test.
You can generate them either with 'Generate Tests' action in IDEA project or just run `org.jetbrains.kotlin.generators.tests.GenerateTestsKt`

JakeWharton:
Oops! You're right. Will do soon. Thanks.

JakeWharton:
Updated!

dzharkov:
I've merged your changes manually. Thank you!

PR #818, opened 2016-02-09
Adds a kapt setting to the gradle plugin for controlling the generated source output directory

sagara-:
This patch adds a new setting to kapt for controlling the output directory of generated sources; this doesn't affect the tmp directory used for storing stubs and other data, which is currently left as-is.

Currently, the output directory is hard-coded to be stored under the build directory. However, this presents an issue with IDEs such as IDEA which don't allow marking generated source directories as source while under an excluded directory (such as the build directory). The only current workaround is to manually unexclude build/ and exclude all its subdirs aside from generated/, and then mark generated/ as a src dir every time the project settings are refreshed.

When using an android project, due to android's modified gradle model this isn't an issue, as the build directory itself isn't marked as excluded, only some of its subdirectories. However, when creating a normal application via gradle, the generated sources under the build directory aren't indexed by IDEA properly, and setting idea.modules.generatedSourceDirs has no effect due to the aforementioned limitation. While the application will still compile correctly, the IDE will report missing class errors, as well as being unable to provide autocompletion for generated classes.

The provided patch provides kapt.generatedFilesBaseDir (name taken from the protobuf-gradle-plugin) with 3 overloads, one for String, one for File and one for taking a Closure<File> with a single Project argument, which can override the root generation directory for the current project. If not provided, defaults to the current directory (generated/source/kapt/).

This can be used to put the generated sources under a different directory, which can then be added to the IDE's src directories via gradle plugins, such as using the idea plugin's idea.modules.sourceDirs+generatedSourceDirs.

A caveat of this is that the gradle clean task will no longer delete the directory; it's up to the user to extend the clean task (eg: via clean.doFirst) to clean up the directory.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KaptExtension.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KaptExtension.kt
@@ -18,6 +18,7 @@ package org.jetbrains.kotlin.gradle.plugin
 
 import groovy.lang.Closure
 import org.gradle.api.Project
+import java.io.File
 
 public open class KaptExtension {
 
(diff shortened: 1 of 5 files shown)


yanex:
Project importing in Kotlin now correctly marks the kapt output directory as "generated source set" (not only for Android projects but also for plain "Java-style" Gradle projects). Please check if the project importing works properly for you.
So looks like remapping destination directories for AP is not needed anymore. @sagara-, what do you think?

yanex:
Looks like this is not actual anymore.
Please [file an issue](https://kotl.in/issue) to our issue tracker if you still have a problem with this.

PR #820, opened 2016-02-11
Include source folders from registered source-generating tasks.

JakeWharton:
This includes the output directories from tasks which generate Java code which were registered with `registerJavaGeneratingTask()`.

I'm not sure how to implement a fallback for this for newer Android plugins as was added recently for the other source folders lookup.

cc @NataliaUkhorskaya

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/android/AndroidGradleWrapper.groovy
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/android/AndroidGradleWrapper.groovy
@@ -105,13 +105,13 @@ class AndroidGradleWrapper {
   }
 
   static def List<File> getGeneratedSourceDirs(BaseVariantData variantData) {
+    def result = new ArrayList<File>()
+
     def getJavaSourcesMethod = variantData.getMetaClass().getMetaMethod("getJavaSources")
     if (getJavaSourcesMethod.returnType.metaClass == Object[].metaClass) {
-      return variantData.getJavaSources().findAll { it instanceof File }
+      result.addAll(variantData.getJavaSources().findAll { it instanceof File })
     }
     else {
-      def result = new ArrayList<File>()
-
       if (variantData.scope.getGenerateRClassTask() != null) {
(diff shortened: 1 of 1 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/android/AndroidGradleWrapper.groovy
```diff
@@ -132,8 +132,13 @@ class AndroidGradleWrapper {
               && variantData.scope.getRenderscriptCompileTask() != null) {
         result.add(variantData.scope.getRenderscriptSourceOutputDir());
       }
+    }
 
-      return result
+    def getExtraSourcesMethod = variantData.getMetaClass().getMetaMethod("getExtraGeneratedSourceFolders")
+    if (getExtraSourcesMethod.returnType.metaClass == List.metaClass) {
+      result.addAll(variantData.getExtraGeneratedSourceFolders())
```
NataliaUkhorskaya: variantData.getExtraGeneratedSourceFolders() is Nullable, so you should check for null, before calling addAll

JakeWharton: Returning null on something that returns a collection. Those developers should have their programming license revoked.

Good catch, will update tomorrow.

sindrenm: Haha, agreed, @JakeWharton. :stuck_out_tongue:

JakeWharton: Updated!

JakeWharton:
Friendly (non-weekend) ping!

NataliaUkhorskaya:
Thanks  a lot and sorry for delay!

asarazan:
Is there an ETA on when this will make it into a release? I've got a couple pain points that this will help with.

PR #821, opened 2016-02-16
Add method CharSequence.occurencesOf

cbruegg:
Related KEEP proposal: [KEEP-20](https://github.com/Kotlin/KEEP/issues/20).

The goal of this pull request is to add an extension method `CharSequence.occurencesOf`, taking two parameters:
1. A pattern to look for in the receiver
2. A flag `ignoreCase`.

It returns a lazy sequence of indices at which the pattern occurs in the receiver.

It is based on the evidentially `O(n)` Knuth-Morris-Pratt algorithm, where `n` is the length of the text to search in. Using `indexOf` repeatedly would have runtime `O(nk)`, where `k` is the length of the pattern.

An abstract use case would be finding all occurences of a short pattern of length `k` in a long text of length `n`, which would be `O(nk)` with `indexOf`, but is `O(n)` with `occurencesOf`. A more concrete use case: Suppose we have a large number of texts (perhaps in a search engine) we would like to run a search query against. With `occurencesOf`, we could get all relevant parts of the text very efficiently, maybe get a substring of the occurence 100 chars before and after, obtaining a summary of where the word occurs. 

This PR of course also includes tests.

### DIFF
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1190,6 +1190,8 @@ private fun computePrefixFunction(pattern: CharSequence): IntArray {
  * Returns a list of indices where the pattern occurs in this String. This method
  * searches character by character and thus does not support regular expressions
  * as input for the pattern.
+ * For a pattern longer than the text, an empty sequence is returned. If the pattern
+ * is empty, all indices are matched.
  *
  * @param [pattern] The pattern to look for in this String. Regular expressions
  *                 are not supported
@@ -1200,8 +1202,12 @@ private fun computePrefixFunction(pattern: CharSequence): IntArray {
  */
 public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean = false): Sequence<Int> {
 
-    if (pattern.isEmpty() || pattern.length > this.length) {
+    if (pattern.length > this.length) {
         return emptySequence()
+    }
+
+    if (pattern.isEmpty()) {
+        return indices.asSequence()
     }
 
     if (pattern.length == 1) {
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1187,17 +1187,30 @@ private fun computePrefixFunction(pattern: CharSequence): IntArray {
 }
 
 /**
- * Returns a list of indices where the pattern occurs in this String. This method
+ * Returns a sequence of indices where the char occurs in this CharSequence.
+ *
+ * @param [char] The character to look for in the string
+ * @param [ignoreCase] If true, characters are matched even if one is upper and the other is
+ *                 lower case
+ *
+ * @return A list of indices where the supplied [char] occurs in the text.
+ */
+public fun CharSequence.occurrencesOf(char: Char, ignoreCase: Boolean = false): Sequence<Int> {
+    return indices.asSequence().filter { this[it].equals(char, ignoreCase) }
+}
+
+/**
+ * Returns a sequence of indices where the pattern occurs in this CharSequence. This method
  * searches character by character and thus does not support regular expressions
  * as input for the pattern.
  * For a pattern longer than the text, an empty sequence is returned. If the pattern
- * is empty, all indices are matched.
+ * is empty, all indices plus an empty occurrence at the end are matched.
  *
  * @param [pattern] The pattern to look for in this String. Regular expressions
  *                 are not supported
  * @param [ignoreCase] If true, characters are matched even if one is upper and the other is
  *                 lower case
- * @param [matchOverlapping] If true, also match overlapping occurences.
+ * @param [matchOverlapping] If true, also match overlapping occurrences.
  *
  * @return A list of indices where the supplied [pattern] starts in the text.
  */
@@ -1208,11 +1221,11 @@ public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean
     }
 
     if (pattern.isEmpty()) {
-        return indices.asSequence()
+        return (0..length).asSequence()
     }
 
     if (pattern.length == 1) {
-        return indices.asSequence().filter { this[it].equals(pattern[0], ignoreCase) }
+        return occurrencesOf(pattern[0], ignoreCase)
     }
 
     // Non-trivial pattern matching, perform computation
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1200,7 +1200,7 @@ private fun computePrefixFunction(pattern: CharSequence): IntArray {
  */
 public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean = false): Sequence<Int> {
 
-    if (isEmpty() || pattern.isEmpty()) {
+    if (pattern.isEmpty() || pattern.length > this.length) {
         return emptySequence()
     }
 
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1174,7 +1174,7 @@ private fun computePrefixFunction(pattern: CharSequence): IntArray {
     var matches = 0
     for (i in 1..pattern.length - 1) {
         while (matches > 0 && pattern[matches] != pattern[i]) {
-            matches = resultTable[matches]
+            matches = resultTable[matches - 1]
         }
 
         if (pattern[matches] == pattern[i]) {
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
what about tests for ignore case = true?

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,77 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
...
+ * @param [pattern] The pattern to look for in this String. Regular expressions
+ *                 are not supported
+ * @param [ignoreCase] If true, characters are matched even if one is upper and the other is
+ *                 lower case
+ *
+ * @return A list of indices where the supplied [pattern] starts in the text.
+ */
+public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean = false): Sequence<Int> {
+
+    if (isEmpty() || pattern.isEmpty()) {
```
cy6erGn0m: `if (pattern.isEmpty() || pattern.length > this.length)`  ?

cbruegg:
There's this one: `assertEquals("Abaabaaababac".occurrencesOf("abaa", ignoreCase = true).toList(), listOf(0, 3))`

I'll add some more though. Good suggestion for the early return as well, I'll change it and update this PR. Thanks for the feedback!

cy6erGn0m:
I also believe we need few tests for unicode characters

cy6erGn0m:
Another one could be `assertEquals("abaabaaababac".occurrencesOf("abAa", ignoreCase = true).toList(), listOf(0, 3))`

cbruegg:
I've just added tests for two greek unicode characters and added more tests for the ignoreCase condition.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,77 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
...
+ *                 are not supported
+ * @param [ignoreCase] If true, characters are matched even if one is upper and the other is
+ *                 lower case
+ *
+ * @return A list of indices where the supplied [pattern] starts in the text.
+ */
+public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean = false): Sequence<Int> {
+
+    if (pattern.isEmpty() || pattern.length > this.length) {
+        return emptySequence()
```
ilya-g: I believe this should be consistent with the result of `Regex("").findAll(input = "....")`, i.e. a match in every position.

cbruegg:
What do you mean? If you take a look at the tests, `"abaabaaababac".occurrencesOf("abaa")` returns 0 and 3, so it does match every position. I haven't run the regex you've posted, but it should have the same result.

cbruegg:
Nevermind, I see what you mean. I'll change the PR.

cy6erGn0m:
Should we return *all* occurrences or only non-overlapping? eg
`AAA.occurrencesOf(AA)` = `0, 1` or `0` ?

cbruegg:
I could probably add an additional parameter for that. Its default value
should be to find all occurrences in my opinion.

Sergey Mashkov <notifications@github.com> schrieb am Mi., 17. Feb. 2016
11:47:

cy6erGn0m:
@ilya-g what do you think? what behaviour is better by default? as far as I understand all our api doesn't return overlapping occurrences, isn't it?

cbruegg:
The equivalent regex does not return overlapping occurrences. I'll set the
default parameter value to non-overlapping.

Sergey Mashkov <notifications@github.com> schrieb am Mi., 17. Feb. 2016
12:35:

cbruegg:
Just checking you've noticed the commit - not sure if GitHub sends notifications for commits in PRs or only for comments. If you saw the commit already, please disregard this comment.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,84 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
...
+ * @return A list of indices where the supplied [pattern] starts in the text.
+ */
+public fun CharSequence.occurrencesOf(pattern: CharSequence, ignoreCase: Boolean = false, matchOverlapping: Boolean = false): Sequence<Int> {
+
+    if (pattern.length > this.length) {
+        return emptySequence()
+    }
+
+    if (pattern.isEmpty()) {
+        return indices.asSequence()
```
ilya-g: It should be `(0..length).asSequence()`, because you need to count an empty match before the first character and empty one after the last characater.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,84 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
...
+    }
+
+    if (pattern.length == 1) {
+        return indices.asSequence().filter { this[it].equals(pattern[0], ignoreCase) }
```
ilya-g: We could extract `CharSequence.occurrencesOf(char: Char, ignoreCase)` function from here.

cbruegg:
Thanks for the suggestions, I've incorporated them.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,97 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
  * * Splits this char sequence to a list of lines delimited by any of the following character sequences: CRLF, LF or CR.
  */
 public fun CharSequence.lines(): List<String> = lineSequence().toList()
+
+/**
+ * For each i in [0, length), this function computes
+ * the length of the longest suffix of a substring of pattern from 0 to i
+ * that is also a prefix of the pattern itself.
+ */
+private fun computePrefixFunction(pattern: CharSequence): IntArray {
```
ilya-g: I'm not sure the prefix function is computed correctly. For the following example "ABCDABD" it returns [0,0,0,0,1,2,0], while according to the [reference](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm#Worked_example_of_the_table-building_algorithm) it should be [-1,0,0,0,0,1,2].

Also it loops infinitely on the input "AAABAAB".

libraries/stdlib/test/text/StringTest.kt
```diff
@@ -714,6 +714,43 @@ class StringTest {
         }
     }
 
+    @test fun occurencesOf() {
+        // Multiple overlapping occurances
+        assertEquals("abaabaaababac".occurrencesOf("abaa", matchOverlapping = true).toList(), listOf(0, 3))
```
ilya-g: `expected` value comes first, `actual` value second.

libraries/stdlib/src/kotlin/text/Strings.kt
cbruegg: I've fixed the infinite loop, the prefix function algorithm missed a subtraction that wasn't caught by a test yet, so I've also added a test case for that.

I've never seen -1 in the table before, so KMP as described in the Wikipedia article seems to be a slightly different variation of KMP. My implementation is based on the pseudocode we were given in a lecture on Fundamental Algorithms at the University of Paderborn and it was also proven to be working correct by the lecturer, so this should be fine.

libraries/stdlib/test/text/StringTest.kt
cbruegg: Fixed, thanks.

libraries/stdlib/src/kotlin/text/Strings.kt
ilya-g: Should the `ignoreCase` parameter affect prefix function?

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1162,3 +1162,97 @@ public fun CharSequence.lineSequence(): Sequence<String> = splitToSequence("\r\n
...
+    var matches = 0
+    return generateSequence {
+        while (i < length) {
+            while (matches > 0 && !pattern[matches].equals(this[i], ignoreCase)) {
+                matches = prefixFunction[matches - 1]
+            }
+
+            if (pattern[matches].equals(this[i], ignoreCase)) {
+                matches++
+            }
```
ilya-g: This approach requires two character comparisons per position when `matches > 0`. Could you restructure the loop body? Also applies to `computePrefixFunction`.

libraries/stdlib/src/kotlin/text/Strings.kt
cbruegg: Actually yes, it should. Learning the hard way that even with tests such things can easily be missed, sorry. I've just pushed a fix with another test case as well.

libraries/stdlib/src/kotlin/text/Strings.kt
cbruegg: Done.

ilya-g:
Since we have no plans merging this PR in the standard library (see [this comment](https://github.com/Kotlin/KEEP/issues/20#issuecomment-230317194) in the related KEEP discussion), I'm closing this for now.

Since it's linked to the KEEP, it would be easy to recover it when it will become required.

PR #822, opened 2016-02-17
KT-9183 - Annotation processing isn't processing all field and method annotations unless the names are unique

donaldlittlepie:
Took a stab at fixing what I thought the bug with - https://youtrack.jetbrains.com/issue/KT-9183 - was

I believe there are probably a few issues with kapt processing (i.e incremental builds), but the one which was blocking me was that field names and method names had to be unique across annotations.   

This change broke the original `AnnotationListParseTest.testPlatformStatic` test. I updated the test to, so that it doesn't fail, but I thought I should bring it to your attention.

Hopefully it helps, if only to provide a test which shows my problem.

### DIFF
--- a/libraries/tools/kotlin-annotation-processing/src/main/kotlin/org/jetbrains/kotlin/annotation/AnnotationProcessorWrapper.kt
+++ b/libraries/tools/kotlin-annotation-processing/src/main/kotlin/org/jetbrains/kotlin/annotation/AnnotationProcessorWrapper.kt
@@ -37,18 +37,19 @@ class AnnotatedClassDescriptor(classFqName: String) : AnnotatedElementDescriptor
 }
 
 class AnnotatedMethodDescriptor(classFqName: String, public val methodName: String) : AnnotatedElementDescriptor(classFqName) {
-    override fun equals(other: Any?) = other is AnnotatedMethodDescriptor && methodName == other.methodName
+    override fun equals(other: Any?) = other is AnnotatedMethodDescriptor && methodName == other.methodName && classFqName == other.classFqName
 
-    override fun hashCode() = methodName.hashCode()
+    override fun hashCode() =  methodName.hashCode() + classFqName.hashCode()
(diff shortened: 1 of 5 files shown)


libraries/tools/kotlin-annotation-processing/src/main/kotlin/org/jetbrains/kotlin/annotation/AnnotationProcessorWrapper.kt
```diff
@@ -37,18 +37,19 @@ class AnnotatedClassDescriptor(classFqName: String) : AnnotatedElementDescriptor
 }
 
 class AnnotatedMethodDescriptor(classFqName: String, public val methodName: String) : AnnotatedElementDescriptor(classFqName) {
-    override fun equals(other: Any?) = other is AnnotatedMethodDescriptor && methodName == other.methodName
+    override fun equals(other: Any?) = other is AnnotatedMethodDescriptor && methodName == other.methodName && classFqName == other.classFqName
 
-    override fun hashCode() = methodName.hashCode()
+    override fun hashCode() =  methodName.hashCode() + classFqName.hashCode()
```
naixx: Is it the best way to combine hashcodes?

donaldlittlepie: Probably not. Thanks for pointing this out.

inorichi:
I think methods should also use referential equality, because there can be methods with the same name and different arguments.

```
class SomeClass {

    @MyAnnotation
    fun onEvent(event: A) {}

    @MyAnnotation
    fun onEvent(event: B) {}
}
```

With this example, kapt will process only one of these methods.

yanex:
https://github.com/JetBrains/kotlin/pull/823 was merged instead.
Anyway, thank you for the contribution!

PR #823, opened 2016-02-23
kapt: Support same annotation on Foo.baz and Bar.baz

hansenji:
Handle the case where an elements with the same name in two different classes are annotated with the same annotation;
```
class Foo() {
  @Qux
  var baz: Int
}

class Bar() {
  @Qux
  var baz: Int
}
```
In the current implementation only one of the annotations will be sent to the Processor.
The pull request fixes that issue.

`@PlatformStatic` was removed from the source so the test has been removed.


abreslav:
Could you create/find an issue in [YouTrack](https://youtrack.jetbrains.com/issues/KT) for this request and mention this PR in the comments? Thanks!

hansenji:
https://youtrack.jetbrains.com/issue/KT-9183

yanex:
Commits are in master: https://github.com/JetBrains/kotlin/commit/100252b18b052ddfc07dafd2a2da81a678881551, https://github.com/JetBrains/kotlin/commit/b9dc92d7cf0a1423d4bc510fce6b956f04bda373.

PR #826, opened 2016-02-26
Implement smart enter processors for try, catch (parameters and body) and finally

cypressious:
Fixes #KT-10013


yole:
Thanks for your contributioN!

PR #827, opened 2016-02-27
Add range get from list and list multiplication

MarcinMoskala:
There are to 2 handy features to get elements from list like:
list[0, 1, 2, 5] or list[2..4]

### DIFF
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -246,6 +246,11 @@ operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
             listOf<T>()
 
 /**
+ * Returns the list multiplication.
+ */
+operator fun <T> List<T>.times(factor: Int): List<T> = (1..factor).flatMap { this }
+
+/**
  * Checks that `from` and `to` are in
  * the range of [0..size] and throws an appropriate exception, if they aren't.
  */
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -237,12 +237,11 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
 /**
  * Returns the elements from the list from indexes specified in range.
  */
-public operator fun <T> List<T>.get(indexesRange : IntRange): List<T> = indexesRange.map { get(it) }
-
-/**
- * Returns the elements from the list from indexes specified in arguments.
- */
-public operator fun <T> List<T>.get(vararg indexes : Int): List<T> = indexes.map { get(it) }
+operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if(indexesRange.first<=indexesRange.last)
+            subList(indexesRange.first, indexesRange.last+1)
+        else
+            listOf<T>()
 
 /**
  * Checks that `from` and `to` are in
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -238,10 +238,12 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
  * Returns the elements from the list from indexes specified in range.
  */
 operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
-        if(indexesRange.first < 0 || indexesRange.last > lastIndex)
-            throw IndexOutOfBoundsException()
-        else if(indexesRange.first<=indexesRange.last)
-            subList(indexesRange.first, indexesRange.last+1)
+        if (indexesRange.first < 0)
+            throw IndexOutOfBoundsException("Range beginning (${indexesRange.first}) is less than zero.")
+        else if (indexesRange.last > lastIndex)
+            throw IndexOutOfBoundsException("Range end (${indexesRange.last}) is bigger then the end of table ($lastIndex)")
+        else if (indexesRange.first <= indexesRange.last)
+            subList(indexesRange.first, indexesRange.last + 1)
         else
             listOf<T>()
 
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -238,7 +238,9 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
  * Returns the elements from the list from indexes specified in range.
  */
 operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
-        if(indexesRange.first<=indexesRange.last)
+        if(indexesRange.first < 0 || indexesRange.last > lastIndex)
+            throw IndexOutOfBoundsException()
+        else if(indexesRange.first<=indexesRange.last)
             subList(indexesRange.first, indexesRange.last+1)
         else
             listOf<T>()
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -237,24 +237,26 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
 /**
  * Returns the elements from the list from indexes specified in range.
  */
-operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
-        if (indexesRange.first < 0)
+public operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if (indexesRange.isEmpty())
+            listOf<T>()
+        else if (indexesRange.first < 0)
             throw IndexOutOfBoundsException("Range beginning (${indexesRange.first}) is less than zero.")
         else if (indexesRange.last > lastIndex)
             throw IndexOutOfBoundsException("Range end (${indexesRange.last}) is bigger then the end of table ($lastIndex)")
         else if (indexesRange.first <= indexesRange.last)
-            subList(indexesRange.first, indexesRange.last + 1)
+            subList(indexesRange.first, indexesRange.last + 1).toList()
         else
-            listOf<T>()
+            throw Exception("Unknown exception in getting range of indexes from list")
 
 /**
  * Returns the list multiplication.
  */
-operator fun <T> List<T>.times(factor: Int): List<T> =
-        if (factor < 0)
-            throw ArithmeticException("List factor must be bigger then 0")
-        else
-            (1..factor).flatMap { this }
+public operator fun <T> List<T>.times(factor: Int): List<T> {
+    require(factor >= 0) { "List factor must be bigger then 0" }
+    return (1..factor).flatMap { this }
+}
+
 
 /**
  * Checks that `from` and `to` are in
--- a/libraries/stdlib/src/kotlin/collections/Collections.kt
+++ b/libraries/stdlib/src/kotlin/collections/Collections.kt
@@ -235,6 +235,16 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
 }
 
 /**
+ * Returns the elements from the list from indexes specified in range.
+ */
+public operator fun <T> List<T>.get(indexesRange : IntRange): List<T> = indexesRange.map { get(it) }
+
+/**
+ * Returns the elements from the list from indexes specified in arguments.
+ */
+public operator fun <T> List<T>.get(vararg indexes : Int): List<T> = indexes.map { get(it) }
+
+/**
  * Checks that `from` and `to` are in
  * the range of [0..size] and throws an appropriate exception, if they aren't.
  */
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
Doesn't vararg getter conflict with the regular `get`?

cy6erGn0m:
For me `get` with vararg is unnecessary thing but I like the `IntRange` idea

cy6erGn0m:
But I am not sure about implementation: shouldn't we do `subList(range).toList()` instead?

MarcinMoskala:
There is no conflict in java 
http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2.5

MarcinMoskala:
More like:

operator fun <T> List<T>.get(indexesRange: IntRange): List<T> = subList(indexesRange.first, indexesRange.last)

?

MarcinMoskala:
Let's do it then

MarcinMoskala:
I correct it. Could You take a look on that? :)

MarcinMoskala:
I also added list multiplication. Also usefull feature

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,22 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
 }
 
 /**
+ * Returns the elements from the list from indexes specified in range.
+ */
+operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if(indexesRange.first < 0 || indexesRange.last > lastIndex)
```
cy6erGn0m: What is the value of `last` of empty range and of `lastIndex` of the empty list? Don't they throw an exception? You have to add corresponding tests

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,22 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
+            throw IndexOutOfBoundsException()
+        else if(indexesRange.first<=indexesRange.last)
+            subList(indexesRange.first, indexesRange.last+1)
+        else
+            listOf<T>()
+
+/**
+ * Returns the list multiplication.
+ */
+operator fun <T> List<T>.times(factor: Int): List<T> = (1..factor).flatMap { this }
```
cy6erGn0m: Negative `factor` should throw an exception

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,22 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
 }
 
 /**
+ * Returns the elements from the list from indexes specified in range.
+ */
+operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if(indexesRange.first < 0 || indexesRange.last > lastIndex)
+            throw IndexOutOfBoundsException()
```
cy6erGn0m: It should be more debug information: exact range bounds and the list size

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,22 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
+        else if(indexesRange.first<=indexesRange.last)
```
cy6erGn0m: Formatting

libraries/stdlib/test/collections/CollectionTest.kt
```diff
@@ -854,4 +854,38 @@ class CollectionTest {
...
+            f()
+        } catch(e: Exception){
+            return true
+        }
+        return false
+    }
+
+    @test fun testListTimes(){
+        val shortlistStr = listOf("a", "b")
+        assertTrue(shortlistStr*0 == listOf<String>())
```
cy6erGn0m: `assertEquals(expected, actual) ` is preferred as it dumps diff on failure

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: Done

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: Done

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: Done

libraries/stdlib/test/collections/CollectionTest.kt
MarcinMoskala: Done

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: This is controversive. I was thinking about it and I check how it works in Python ant there for minus factors there is no error but empyt list. But I will change it If you prefer this way.

cy6erGn0m: Yes, our other functions do that so `times` should as well

libraries/stdlib/test/collections/CollectionTest.kt
```diff
@@ -854,4 +854,41 @@ class CollectionTest {
...
         assertEquals("[1, a, null, ${Long.MAX_VALUE.toString()}]", listOf(1, "a", null, Long.MAX_VALUE).toString())
     }
+
+    @test fun testGetRange() {
+        val listStr = listOf("a", "b", "c", "d", "e")
+
+        assertTrue(throwsError{listStr[-1..2]})
+        assertTrue(throwsError{listStr[1..10]})
+
+        assertEquals(listStr[IntRange.EMPTY], listOf<String>())
```
cy6erGn0m: I should be expected value at the first place, the second argument is actual, swap these arguments please

libraries/stdlib/test/collections/CollectionTest.kt
```diff
@@ -854,4 +854,41 @@ class CollectionTest {
         // we need toString() inside pattern because of KT-8666
         assertEquals("[1, a, null, ${Long.MAX_VALUE.toString()}]", listOf(1, "a", null, Long.MAX_VALUE).toString())
     }
+
+    @test fun testGetRange() {
+        val listStr = listOf("a", "b", "c", "d", "e")
```
cy6erGn0m: I believe we also need tests on an empty list: should fail at any range but empty range on empty list should produce empty list

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,28 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
+            subList(indexesRange.first, indexesRange.last + 1)
+        else
+            listOf<T>()
+
+/**
+ * Returns the list multiplication.
+ */
+operator fun <T> List<T>.times(factor: Int): List<T> =
+        if (factor < 0)
+            throw ArithmeticException("List factor must be bigger then 0")
```
cy6erGn0m: You may just do `require(factor >= 0) {"error message… "}`. Also notice that in Java wrong arguments should cause `IllegalArgumentException ` (so it is done inside of the `require` function)

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,28 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
 /**
+ * Returns the elements from the list from indexes specified in range.
+ */
+operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if (indexesRange.first < 0)
+            throw IndexOutOfBoundsException("Range beginning (${indexesRange.first}) is less than zero.")
+        else if (indexesRange.last > lastIndex)
+            throw IndexOutOfBoundsException("Range end (${indexesRange.last}) is bigger then the end of table ($lastIndex)")
+        else if (indexesRange.first <= indexesRange.last)
+            subList(indexesRange.first, indexesRange.last + 1)
```
cy6erGn0m: It should be `toList()` after because we need a copy but `subList` just refers to the original list.

cy6erGn0m:
Generally it is much better now, thank you for your help. Fix the remaining notes and we'll discuss it. I think it is the good one candidate for merge

libraries/stdlib/test/collections/CollectionTest.kt
MarcinMoskala: Done

libraries/stdlib/test/collections/CollectionTest.kt
MarcinMoskala: Done

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: Done

libraries/stdlib/src/kotlin/collections/Collections.kt
MarcinMoskala: Done

MarcinMoskala:
Fixed

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,30 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
 /**
+ * Returns the elements from the list from indexes specified in range.
+ */
+public operator fun <T> List<T>.get(indexesRange: IntRange): List<T> =
+        if (indexesRange.isEmpty())
+            listOf<T>()
+        else if (indexesRange.first < 0)
+            throw IndexOutOfBoundsException("Range beginning (${indexesRange.first}) is less than zero.")
+        else if (indexesRange.last > lastIndex)
+            throw IndexOutOfBoundsException("Range end (${indexesRange.last}) is bigger then the end of table ($lastIndex)")
```
cy6erGn0m: typo: than

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -235,6 +235,30 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
+        else if (indexesRange.first <= indexesRange.last)
+            subList(indexesRange.first, indexesRange.last + 1).toList()
+        else
+            throw Exception("Unknown exception in getting range of indexes from list")
+
+/**
+ * Returns the list multiplication.
+ */
+public operator fun <T> List<T>.times(factor: Int): List<T> {
+    require(factor >= 0) { "List factor must be bigger then 0" }
```
cy6erGn0m: Type: than
Also lets include actual value in the message

cy6erGn0m: Also, in fact the message is not exact correct: factor shouldn't be negative so zero is also included

HappyEmu:
I'd like to throw in the idea to support negative upper range limits as well. The negative upper limit would represent the offset from the size of the list. Example:

`list[2..-1]` would return all elements starting from index 2 until the end (`list.size - 1`) of the list.
`list[3..-2]` would return all elements starting from index 3 until index `list.size - 2`

MarcinMoskala:
Can do it. And wanted to do it. Just don't know if rest agree with that. That is why I prefer to close it like that right now (because it is my first pull request and I care a lot) and add this funcionality in next pull request.

cy6erGn0m:
@MarcinMoskala Another one problem I can see is `list[1..10 step 2]`. Open question is: `list[10 downTo 1] = ?`

cy6erGn0m:
Regarding negative range bounds: not sure it is good idea as it may cause strange consequences. See

```
fun myfun(someRange: IntRange) = myList[someRange]
```

In this case we can potentially get unpredictable results as IntRange could be computed somehow

```
myfun(1..someString.indexOf(something))
```

MarcinMoskala:
That is why 

operator fun <T> List<T>.get(indexesRange: IntRange): List<T> = indexesRange.toList().map { get(it) }

was perfect xP

MarcinMoskala:
With - values:

List.get(indexesRange: IntRange): List = indexesRange.toList().map {if(i<0) i-size else i}. map { get(it) }

cy6erGn0m:
Why do you need `toList()` here?
I still believe it is not good idea to support negative values

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -237,24 +237,26 @@ public fun <T> List<T>.binarySearch(fromIndex: Int = 0, toIndex: Int = size, com
...
 /**
  * Returns the list multiplication.
  */
-operator fun <T> List<T>.times(factor: Int): List<T> =
-        if (factor < 0)
-            throw ArithmeticException("List factor must be bigger then 0")
-        else
-            (1..factor).flatMap { this }
+public operator fun <T> List<T>.times(factor: Int): List<T> {
+    require(factor >= 0) { "List factor must be bigger then 0" }
```
HaasJona: "than", also is 0 itself valid or not?

PR #828, opened 2016-03-03
Implement Smart Enter Processor for adding a lambda block

cypressious:
Fixes #KT-6785


cypressious:
Unfortunately, I haven't figured out how to make the unit tests work. I suspect it has to do with previous fixers only working on the syntax level so the call resolution stuff is broken in the test environment.

cypressious:
Tests work after all if you use your own function instead of stdlib ones like `with`.

cypressious:
Parameter names are not set anymore. Also the lambda is now in the same line.

cypressious:
I simplified the whole logic. The caret is now set directly in the first pass so there is no need for a second pass anymore.

yole:
Looks good now. Thanks a lot for your contribution!

cypressious:
Great. I think I get the hang of the smart enter logic now. When I understood the multi-pass logic, everything became much more clearer.

PR #829, opened 2016-03-04
added javax.annotation.Nonnull

robfletcher:
Since the equivalent `Nullable` annotation is there, I thought it was odd `Nonnull` was missing.


dzharkov:
The annotation is already supported, but because of special semantics (there is an optional parameter determining nullability) it should be treated differently from common `NonNull` annotations.

See https://github.com/robfletcher/kotlin/blob/master/core/descriptor.loader.java/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt#L34 

Please, let me know if annotation was not handled properly in your cases

robfletcher:
I did find some differences in behavior between JetBrains `@NotNull` annotation and `@Nonnull`. See this spek: https://gist.github.com/robfletcher/fe1591ddbc8144289fdf#file-interopnullabilityspec-kt-L21-L51

dzharkov:
@robfletcher Behavior is different because of annotation processor instrumenting bytecode for methods marked as `NotNull`. 
It's not related to Kotlin compiler, you can achieve the same result with raw Java

PR #830, opened 2016-03-06
consistently omit semicolon in js math.kt

egorich239:


ilya-g:
Thanks for PR.
I've decided to reformat js-stdlib instead, including removing of redundant semicolons.

PR #831, opened 2016-03-08
Add new expressions to template

NitroG42:
Added two missing expressions for live-templates, the className(), and
the functionName().

### DIFF
--- a/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/macro/ClassNameMacro.kt
+++ b/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/macro/ClassNameMacro.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2010-2015 JetBrains s.r.o.
+ *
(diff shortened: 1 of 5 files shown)


yole:
The names of these functions conflict with the standard Java live template macro function names. You need to add the "kotlin" prefix to the names to ensure they don't conflict.

NitroG42:
They are only available in the kotlin context, will the conflict still risk to occurs ?

yole:
Yes, the names still need to be distinct to ensure the templates are parsed correctly.

yole:
A test would be nice.

yole:
Rewritten and merged manually.

NitroG42:
Hello, I'm sorry for the lack of news, I planned to make the modifications (I updated the code), but wanted to learn how to make test for macro, and I just didn't have the time to do that right now.
Thank you for the rewriting, I'm glad it could be merged to the project though.

yole:
It turns out that there's no really good way to test a live template macro in the IntelliJ Platform, so I pushed the change without tests.

PR #833, opened 2016-03-13
Add release dates to the ChangeLog

svetlyak40wt:


yole:
March 7th is not the release date for Kotlin 1.0.2, nor any other version of Kotlin.

svetlyak40wt:
I've just used git-blame to see the date, when header 1.0.2 was added. If you know the real date or it is not released yet, then put correct date. This will help community to keep on track and learn the history of the language.

yole:
There are more than enough places that can be used to keep track of the history of Kotlin, including the blog, Twitter and the release notes on GitHub. The changelog file is an internal document that will be used to prepare the official release notes on GitHub, and is not intended as a primary source of information for external users.

PR #836, opened 2016-03-19
Fix for KT-11383

VladRassokhin:


goodwinnk:
@VladRassokhin Do you see an effect from adding '\n' to placeholder text in editor?

VladRassokhin:
@goodwinnk Forgot to test that. Actually there no effect in editor.
Seems something is wrong inside IntelliJ :disappointed:

PR #837, opened 2016-03-19
Formatter: Remove spaces between expression and parentheses

cypressious:
Fixes #KT-11476


goodwinnk:
Great, thank you! Merged with addition modification in SmartEnterTest.

PR #840, opened 2016-03-21
Implement quickfix to remove '.java' when KClass is expected

cypressious:
Fixes #KT-11078

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertClassToKClassFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertClassToKClassFix.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2010-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertClassToKClassFix.kt
```diff
@@ -0,0 +1,82 @@
...
+        if (!super.isAvailable(project, editor, file)) return false
+
+        val expressionType = element.analyze(BodyResolveMode.PARTIAL).getType(element) ?: return false
+        if (!expressionType.isJClass()) return false
+
+        val children = element.children
+        if (children.size != 2) return false
+
+        val firstChild = children.first() as? KtExpression ?: return false
+        val firstChildType = firstChild.analyze(BodyResolveMode.PARTIAL).getType(firstChild) ?: return false
```
yole: You should reuse the BindingContext from the first analyze() call.

idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertClassToKClassFix.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        return firstChildType.isSubtypeOf(type)
+    }
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val lastChild = element.children.lastOrNull() ?: return
+        editor?.document?.replaceString(lastChild.startOffset - 1, lastChild.endOffset, "")
```
yole: The quickfix will do nothing if there is no editor. It's possible to perform the change through the PSI, and it's better to do it that way.

cypressious:
@yole Done

yole:
Merged manually. Thanks a lot!

PR #842, opened 2016-03-25
Implement quickfix for letting type implement interface

cypressious:
Fixes #KT-11404

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/LetImplementInterfaceFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/LetImplementInterfaceFix.kt
@@ -30,12 +30,12 @@ import org.jetbrains.kotlin.psi.KtPsiFactory
 import org.jetbrains.kotlin.resolve.DescriptorToSourceUtils
 import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.TypeUtils
 import org.jetbrains.kotlin.types.typeUtil.isInterface
-import org.jetbrains.kotlin.types.typeUtil.supertypes
 
 private fun KotlinType.renderShort() = IdeDescriptorRenderers.SOURCE_CODE_SHORT_NAMES_IN_TYPES.renderType(this)
 
-private fun KotlinType.getSuperTypesRecursive() = constructor.supertypes.let { it + it.flatMap { it.supertypes() } }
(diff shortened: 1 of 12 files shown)


mglukhikh:
Under investigation

cypressious:
@mglukhikh Any feedback? :)

cypressious:
Rebased, removed reference to type, fixed the logic to consider all indirect supertypes, grouped tests and added one

idea/src/org/jetbrains/kotlin/idea/quickfix/LetImplementInterfaceFix.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.resolve.DescriptorToSourceUtils
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.isInterface
+import org.jetbrains.kotlin.types.typeUtil.supertypes
+
+private fun KotlinType.renderShort() = IdeDescriptorRenderers.SOURCE_CODE_SHORT_NAMES_IN_TYPES.renderType(this)
+
+private fun KotlinType.getSuperTypesRecursive() = constructor.supertypes.let { it + it.flatMap { it.supertypes() } }
```
udalov: This is exponential; please use `TypeUtils#getAllSupertypes` instead

cypressious: If I understand correctly, Mikhail is merging this manually in https://github.com/JetBrains/kotlin/commit/b569e267ac165a2dcb0d5c9670b5df4e6c55ae6f#diff-fd98cb60318ca31d01b7fbc6d9f39f3a

cypressious: Anyway, fixed it.

mglukhikh:
Pushed to master. I don't think additional code with indirect supertypes is really necessary because we will not have any error and thus no quick-fix. Anyway, I added a test for this case. Sorry for slow consideration.

cypressious:
@mglukhikh The problem arrises when you expect generic type `A<X>` but the type already indirectly implements `A<Y>`. Here's a demonstration:

```kotlin

    foo(B()<caret>)
}


fun foo(a: A<Y>) {
}

interface A<T>
open class C : A<X>
class B : C()

class X
class Y
```

PR #844, opened 2016-03-29
#KT-10797 Fixed

mcgee:
Fix for https://youtrack.jetbrains.com/issue/KT-10797


yole:
I don't think that adding the `operator` modifier automatically is a good idea at this time. I think that in Kotlin 1.0 we can rely on the fact that functions intended to be used with the 'in' operator already have the modifier, and simply disable the quickfix if the modifier is not present.

mcgee:
Okay, fixed.

PR #845, opened 2016-03-30
Fix #KT-11675 Fixed - Add Smart Enter processor for 'init'

yarulan:


yole:
Thanks for your contribution!

PR #847, opened 2016-04-02
Prevent NPE when resolving enum entries and their properties

EddieRingle:
Compiler was crashing with the following:

```kotlin
enum class CrashEnum {
  first,
  second,
  name
}
```

i.e., if an enum had entries named `name` or `ordinal`,
the compiler would crash. This also caused many errors
in the IDE as the file was being resolved.

Signed-off-by: Eddie Ringle <eddie@ringle.io>


EddieRingle:
Looked into it more and @orangy is correct, it's definitely a lot more involved than someone new to the codebase like myself is prepared to deal with. Closing this, but I've submitted a PR (#848) that adds a test for [KT-11750](https://youtrack.jetbrains.com/issue/KT-11750).

PR #849, opened 2016-04-04
Fix handling of nullable types in LetImplementInterfaceFix #KT-11404

cypressious:


mglukhikh:
Pushed to master (manually). Many thanks again.

PR #850, opened 2016-04-04
ConvertToForEachFunctionCallIntention: convert 'continue' to 'return@forEach'

cypressious:
Fixed #KT-11764

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToForEachFunctionCallIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToForEachFunctionCallIntention.kt
@@ -17,10 +17,12 @@
 package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.idea.util.CommentSaver
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.contentRange
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
 import org.jetbrains.kotlin.psi.psiUtil.forEachDescendantOfType
 import java.util.*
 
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToForEachFunctionCallIntention.kt
```diff
@@ -35,15 +34,45 @@ class ConvertToForEachFunctionCallIntention : SelfTargetingIntention<KtForExpres
...
+            it.replace(psiFactory.createExpression("return@forEach"))
+        }
 
         commentSaver.restore(result)
     }
+
+    private fun KtElement.getContinuesWithLabel(labelName: String?): List<KtContinueExpression> {
+        val continueElements = ArrayList<KtContinueExpression>()
+
+        forEachDescendantOfType<KtContinueExpression>({ it !is KtLoopExpression }) {
```
yole: I think the code will be cleaner if you merge the two loops into one.

yole: You also need to check for nested lambdas/functions/classes - the `continue` statements inside them must be left alone.

cypressious: @yole 
Regarding your first comment: 
There is a certain logic behind the two loops. The first one never enters nested loops and collects all the unqualified `continue`s while the second one also enters nested loops. If I merged the two, how would I know whether to collect an unqualified `continue`?

Regarding your second comment:
Can you show me a code snippet that would be broken by the current implementation? In my understanding a `continue` inside a lambda/function/class would need to be inside a separate loop in which case the `it !is KtLoopExpression` check would prevent me from collecting them.

yole: OK, agreed regarding the two loops. As for the second comment, intentions are not always invoked on green code, and it's usually best to avoid breaking it even further.

cypressious: Added checks.

cypressious:
Related: https://youtrack.jetbrains.com/issue/KT-11781

yole:
Thanks for your contribution!

PR #852, opened 2016-04-05
Add QuickFix to add 'crossinline' to parameter

cypressious:
Fixes #KT-10844


yole:
Thanks for your contribution!

PR #853, opened 2016-04-06
Fix let implement quickfix label for object expressions

cypressious:
Fixes #KT-11782


mglukhikh:
Merged manually with master and 1.0.2 branches.

PR #855, opened 2016-04-08
Let Implement Quickfix: Launch "Implement Members" dialog after applying

cypressious:
Fixes #KT-11836


mglukhikh:
Merged manually into master and 1.0.2 branches.

PR #859, opened 2016-04-22
Add intrinsic for KCallable.name property

cypressious:
Fixes #KT-11531


compiler/testData/codegen/box/reflection/properties/propertyName.kt
```diff
@@ -0,0 +1,12 @@
...
+    val a = ""
+    val b = ""
+}
+
+fun box(): String {
+    val a = A::a.name
+    if (a != "a") return "Fail $a"
+    val b = A::`b`.name
+    if (b != "b") return "Fail $b"
+    return "OK"
```
udalov: This test doesn't check that no classes were generated for the property reference. I don't think it's possible in a `box` test, please try a bytecode listing test instead (`compiler/testData/codegen/bytecodeListing`).

Also please test constructors and functions

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/KCallableNameProperty.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.jetbrains.org.objectweb.asm.Type
+
+class KCallableNameProperty : IntrinsicPropertyGetter() {
+    override fun generate(resolvedCall: ResolvedCall<*>?, codegen: ExpressionCodegen, returnType: Type, receiver: StackValue): StackValue? {
+        val expressionReceiver = resolvedCall!!.dispatchReceiver as? ExpressionReceiver ?: return null
+        val expression = expressionReceiver.expression as? KtCallableReferenceExpression ?: return null
+        val callableReference = expression.callableReference
+        val descriptor = callableReference.getResolvedCall(codegen.bindingContext)?.resultingDescriptor ?: return null
+
+        val name = if (descriptor is ConstructorDescriptor) {
```
udalov: `descriptor.name.asString()`

compiler/testData/codegen/box/reflection/properties/propertyName.kt
cypressious: It seems classes are indeed generated. Where should I look to prevent that?

cypressious: Should I hook into `ExpressionCodegen#visitSimpleNameExpression` and prevent the anonymous class from being generated or is there an approach where I can remove the generated class afterwards?

cypressious:
Constructor test already exists in `compiler\testData\codegen\box\reflection\constructors\constructorName.kt`

compiler/testData/codegen/box/reflection/properties/propertyName.kt
cypressious: I got it working by modifying `visitSimpleNameExpression`.

udalov:
> Constructor test already exists in `compiler\testData\codegen\box\reflection\constructors\constructorName.kt`

`KCallable#name` itself is already tested, and not only for constructors but for properties and functions as well. I think I was wrong in suggesting you to use `codegen/box` tests, they won't test anything here, so maybe they should be deleted.

cypressious:
I'll remove the tests. Any feedback on the implementation?

udalov:
I think the implementation is fine, could not think of anything substantially better. I've launched a remote run with your changes.

udalov:
Merged (https://github.com/JetBrains/kotlin/commit/4a619db7219e1a09b63c27d28db25b44ae10066c), thank you for the contribution!

PR #860, opened 2016-04-23
Implement Intention to add @JvmOverloads

cypressious:
Fixes #KT-11523


idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmOverloadsIntention.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.jetbrains.kotlin.idea.project.ProjectStructureUtil
+import org.jetbrains.kotlin.idea.util.addAnnotation
+import org.jetbrains.kotlin.idea.util.findAnnotation
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
+
+private val annotationFqName = FqName("kotlin.jvm.JvmOverloads")
+
+class AddJvmOverloadsIntention : SelfTargetingIntention<KtParameterList>(
+        KtParameterList::class.java, "Add '@JvmOverloads' annotation"
```
yole: I think that all intentions that modify the declaration of a function need to be available on the function name. Showing this only on the parameter list is confusing

idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmOverloadsIntention.kt
```diff
@@ -0,0 +1,61 @@
...
+        val parent = element.parent as? KtModifierListOwner ?: return false
+        val target = when (parent) {
+            is KtNamedFunction -> "function '${parent.name}'"
+            is KtPrimaryConstructor -> "primary constructor"
+            is KtSecondaryConstructor -> "secondary constructor"
+            else -> return false
+        }
+        text = "Add '@JvmOverloads' annotation to $target"
+
+        return !ProjectStructureUtil.isJsKotlinModule(element.getContainingKtFile())
```
yole: If a constructor has one parameter with a default value, adding @JvmOverloads won't modify the resulting bytecode in any way, so the intention shouldn't be displayed.

cypressious: Done

idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmOverloadsIntention.kt
cypressious: Done

idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmOverloadsIntention.kt
```diff
@@ -23,38 +23,57 @@ import org.jetbrains.kotlin.idea.util.addAnnotation
...
         }
-        text = "Add '@JvmOverloads' annotation to $target"
+        val parameters = parameterList?.parameters ?: return false
+
+        text = "Add '@JvmOverloads' annotation to $targetName"
 
         return !ProjectStructureUtil.isJsKotlinModule(element.getContainingKtFile())
-               && element.parameters.any { it.hasDefaultValue() }
-               && parent.findAnnotation(annotationFqName) == null
+               && parameters.size > 1
```
yole: This is not exactly the check that I meant. Specifically for the primary constructor, Kotlin will generate a no-argument overload when all arguments have default values, even if the @JvmOverloads annotation is not present. This doesn't happen for regular functions.

cypressious: You're right. I should have thought of that.

cypressious:
I think it should be correct now.

yole:
Thanks for your contribution!

PR #862, opened 2016-04-24
 KT-11780: Fixed incorrect "No cast needed" warning when casting from extension …

Valdemar0204:
…function to regular one and vise versa


Valdemar0204:
Fix for KT-11780

erokhins:
I've merged your changes manually. Thank you!

PR #863, opened 2016-04-25
Make functions which depend on the reified feature available in JS. #KT-11346

ilya-g:
Note: do not merge with GitHub


libraries/stdlib/src/generated/_Arrays.kt
```diff
@@ -1,3 +1,19 @@
+/*
```
ilya-g: It's added by IDEA's pre-commit action. Really shouldn't for generated files.

libraries/stdlib/test/collections/CollectionTest.kt
```diff
@@ -95,6 +95,48 @@ class CollectionTest {
...
+
+        // doesn't distinguish double from int in JS
+//        val doubleValues: List<Double> = values.filterIsInstance<Double>()
+//        assertEquals(listOf(3.0), doubleValues)
+
+        val stringValues: List<String> = values.filterIsInstance<String>()
+        assertEquals(listOf("abc", "cde"), stringValues)
+
+        // is Any doesn't work in JS
+//        val anyValues: List<Any> = values.filterIsInstance<Any>()
```
ilya-g: ok, could be mentioned

ilya-g:
merged manually

PR #864, opened 2016-04-27
buildString with initial capacity for the StringBuilder instance

K0zka:
this method helps save computation time when the developer knows or have a good
educated guess about the size of the generated string and have to use control
structures while filling the string builder

Signed-off-by: Laszlo Hornyak <laszlo.hornyak@gmail.com>


libraries/stdlib/test/text/StringBuilderTest.kt
```diff
@@ -5,6 +5,12 @@ import org.junit.Test as test
 
 class StringBuilderTest {
 
+    @test fun stringBuildWithInitialCapacity() {
+        val s = buildString(123) {
+            assertEquals(123, capacity())
```
ilya-g: This test won't compile in JS because there is no `StringBuilder.capacity` there.
To make this test JVM-only, move it to `StringBuilderJVMTest.kt`

libraries/stdlib/src/kotlin/text/StringBuilder.kt
```diff
@@ -10,6 +10,12 @@ package kotlin.text
 public inline fun buildString(builderAction: StringBuilder.() -> Unit): String = StringBuilder().apply(builderAction).toString()
 
 /**
+ * Builds new string by populating newly created [StringBuilder] initialized with the given capacity using provided [builderAction] and then converting it to [String].
+ */
+@kotlin.internal.InlineOnly
+inline fun buildString(length : Int, builderAction: StringBuilder.() -> Unit): String = StringBuilder(length).apply(builderAction).toString()
```
ilya-g: Constructor parameter is called `capacity` so it should have the same name here.

K0zka:
replaced https://github.com/JetBrains/kotlin/pull/893

PR #865, opened 2016-04-27
Add quickfix to make referenced primary constructor parameter a property

cypressious:
Fixes #KT-12089

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
@@ -454,17 +454,8 @@ fun checkReservedPrefixWord(sink: DiagnosticSink, element: PsiElement, word: Str
     }
 }
 
-fun KtNameReferenceExpression.getPrimaryConstructorParameterWithSameName(): KtParameter? {
-    return getPrimaryConstructorParameterWithName(getReferencedName())
+fun KtElement.nonStaticOuterClasses(): Sequence<KtClass> {
+    return generateSequence(containingClass()) { if (it.isInner()) it.containingClass() else null }
 }
 
-private fun KtElement.getPrimaryConstructorParameterWithName(name: String): KtParameter? {
-    val ktClass = getStrictParentOfType<KtClass>() ?: return null
-    val result = ktClass.getPrimaryConstructor()?.valueParameters?.firstOrNull { it.name == name }
-
-    return when {
-        result != null -> result
-        ktClass.isInner() -> ktClass.getPrimaryConstructorParameterWithName(name)
-        else -> null
-    }
-}
+fun KtElement.containingClass(): KtClass? = getStrictParentOfType<KtClass>()
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
@@ -454,3 +454,17 @@ fun checkReservedPrefixWord(sink: DiagnosticSink, element: PsiElement, word: Str
     }
 }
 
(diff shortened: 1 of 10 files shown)


compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
```diff
@@ -454,3 +454,17 @@ fun checkReservedPrefixWord(sink: DiagnosticSink, element: PsiElement, word: Str
     }
 }
 
+fun KtNameReferenceExpression.getPrimaryConstructorParameterWithSameName(): KtParameter? {
```
yole: I think this is far too specific to live in a general PSI utlilities class; I'd move it into the quickfix file.

compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
```diff
@@ -454,3 +454,17 @@ fun checkReservedPrefixWord(sink: DiagnosticSink, element: PsiElement, word: Str
...
+    return getPrimaryConstructorParameterWithName(getReferencedName())
+}
+
+private fun KtElement.getPrimaryConstructorParameterWithName(name: String): KtParameter? {
```
yole: The name of this function doesn't describe what it does. A more descriptive name would be "findContainingClassConstructorParameterWithName" (which is long enough to hint that it may be better to split the logic: a function to generate a sequence of outer classes for a class and a single-line call to find the parameter of the class by name).

compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
cypressious: My motivation to make this method public was that there's another quickfix, that adds a constructor parameter with the same name that could be deactivated when this method returns a non-null value.

yole: You can still make it public and reference it from that other quickfix.

cypressious: Ok

idea/src/org/jetbrains/kotlin/idea/quickfix/MakeConstructorParameterPropertyFix.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.psiUtil.getAssignmentByLHS
+import org.jetbrains.kotlin.psi.psiUtil.getPrimaryConstructorParameterWithSameName
+
+class MakeConstructorParameterPropertyFix(element: KtParameter, private val kotlinValVar: KotlinValVar) : KotlinQuickFixAction<KtParameter>(element) {
+    override fun getFamilyName() = "Make primary constructor parameter a property"
+    override fun getText() = "Make primary constructor parameter '${element.name}' a property"
```
yole: In the case when we're updating an outer class, it'd be better to show the name of that class in the inspection text.

compiler/frontend/src/org/jetbrains/kotlin/psi/psiUtil/ktPsiUtil.kt
cypressious: What would be a good name for this sequence method? `getNonStaticOuterClassSequence()`?

yole: Simply nonStaticOuterClasses()

yole:

PR #866, opened 2016-04-27
Add quickfix to add 'toString'

cypressious:
Fixes #KT-12121

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/AddToStringFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/AddToStringFix.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2010-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/AddToStringFix.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.createExpressionByPattern
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+
+class AddToStringFix(element: KtExpression, private val nullable: Boolean) : KotlinQuickFixAction<KtExpression>(element) {
+    override fun getFamilyName() = "Add 'toString()' call"
+    override fun getText() = if (nullable) "Add safe '?.toString()' call" else "Add 'toString()' call"
```
HaasJona: I don't think this nullable variant is needed. In Kotlin, it's safe to call `null.toString()`.

yole:
Thanks for your contribution!

PR #870, opened 2016-05-11
KT-12074 Building Kotlin maven projects using a parent pom will siletly fail

cy6erGn0m:
https://youtrack.jetbrains.com/issue/KT-12074

We should consider module's basedir if there is relative source directory specified for the execution

Also classpath logging removed that was accidently added back: it shouldn't be shown as it is too verbose


cy6erGn0m:
1 - yes, 2 - no (maybe "yes" if I configure plugin in parent module as well but it will test one more way to compile, not this particular issue)

PR #871, opened 2016-05-17
KT-12019 Highlighting of redundant 'if' statements

ensirius:
#KT-12019 Fixed

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
@@ -54,8 +54,6 @@ class RedundantIfInspection : AbstractKotlinInspection(), CleanupLocalInspection
     }
 
     private fun getReturnedExpression(expression: KtExpression?) : KtExpression? {
-        if (expression == null) return null
-
         when(expression) {
             is KtReturnExpression -> return expression.returnedExpression
             is KtBlockExpression -> {
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2010-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
```diff
@@ -0,0 +1,81 @@
...
+        if (!KtPsiUtil.isTrueConstant(thenReturn) && KtPsiUtil.isTrueConstant(elseReturn)) {
+            return true
+        }
+
+        return false
+    }
+
+    private fun getReturnedExpression(expression: KtExpression?) : KtExpression? {
+        if (expression == null) return null
+        if (expression !is KtBlockExpression) return null
```
yole: This check isn't necessary: a statement where `return` is not enclosed in braces is no less trivial.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
```diff
@@ -0,0 +1,81 @@
...
+        if (expression.statements.size != 1) return null
+
+        val statement = expression.statements[0]
```
yole: `expression.statements[0] as? KtReturnExpression ?: return null`

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
```diff
@@ -0,0 +1,81 @@
...
+                }
+            }
+        }
+    }
+
+    private fun isRedundant(expression: KtIfExpression): Boolean {
+        val thenReturn = getReturnedExpression(expression.then) ?: return false
+        val elseReturn = getReturnedExpression(expression.`else`) ?: return false
+
+        if (KtPsiUtil.isTrueConstant(thenReturn) && !KtPsiUtil.isTrueConstant(elseReturn)) {
```
yole: I find it somewhat confusing that the KtNodeTypes.BOOLEAN_CONSTANT is separate from the value check. It would be clearer to add the corresponding `isFalseConstant()` method to `KtPsiUtil` and to remove the node type check from `getReturnedExpression()`.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantIfInspection.kt
ensirius: This check required to cast expression to get statements

yole: You need to handle the case when the expression passed to this method is itself a `PsiReturnExpression`: `if (foo) return true else return false`

yole:
Thanks for your contribution!

udalov:
I've [fixed the name](https://github.com/JetBrains/kotlin/commit/68f1cf5f5254d607aa3a3721ab0c4fc601153155) of the inspection description. Otherwise "Can't find inspection description" was reported in the settings

PR #872, opened 2016-05-18
KT-11023 Inspection to highlight unnecessary java usage

ensirius:
#KT-11023 Fixed

### DIFF
--- a/idea/resources/inspectionDescriptions/UnnecessaryJavaUsage.html
+++ b/idea/resources/inspectionDescriptions/UnnecessaryJavaUsage.html
@@ -1,5 +1,5 @@
 <html>
 <body>
-Reports unnecessary java usage.
+Use of Java API that has a Kotlin equivalent.
 </body>
 </html>
--- a/idea/resources/inspectionDescriptions/UnnecessaryJavaUsage.html
+++ b/idea/resources/inspectionDescriptions/UnnecessaryJavaUsage.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+Reports unnecessary java usage.
+</body>
+</html>
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1525,7 +1525,7 @@
                      groupName="Kotlin"
                      enabledByDefault="true"
                      cleanupTool="true"
-                     level="WEAK_WARNING"
+                     level="WEAK WARNING"
                      language="kotlin"
     />
 
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1521,7 +1521,7 @@
     />
 
     <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.UnnecessaryJavaUsageInspection"
-                     displayName="Unnecessary java usage"
+                     displayName="Use of Java API that has a Kotlin equivalent"
                      groupName="Kotlin"
                      enabledByDefault="true"
                      cleanupTool="true"
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1516,7 +1516,7 @@
                      groupName="Kotlin"
                      enabledByDefault="true"
                      cleanupTool="true"
-                     level="WARNING"
+                     level="WEAK_WARNING"
                      language="kotlin"
     />
 
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1520,6 +1520,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.UnnecessaryJavaUsageInspection"
+                     displayName="Unnecessary java usage"
+                     groupName="Kotlin"
+                     enabledByDefault="true"
+                     cleanupTool="true"
+                     level="WARNING"
+                     language="kotlin"
+    />
+
     <referenceImporter implementation="org.jetbrains.kotlin.idea.quickfix.KotlinReferenceImporter"/>
 
     <fileType.fileViewProviderFactory filetype="KJSM" implementationClass="com.intellij.psi.ClassFileViewProviderFactory"/>
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1516,7 +1516,7 @@
                      groupName="Kotlin"
                      enabledByDefault="true"
                      cleanupTool="true"
-                     level="WEAK_WARNING"
+                     level="WARNING"
                      language="kotlin"
     />
 
@@ -1525,7 +1525,7 @@
                      groupName="Kotlin"
                      enabledByDefault="true"
                      cleanupTool="true"
-                     level="WARNING"
+                     level="WEAK_WARNING"
                      language="kotlin"
     />
 
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
@@ -65,9 +65,9 @@ class UnnecessaryJavaUsageInspection : AbstractKotlinInspection(), CleanupLocalI
             val factory = KtPsiFactory(element)
 
             if(pattern.second) {
-                element.replace(factory.createExpressionByPattern(pattern.first, receiver.text, *arguments.toTypedArray()));
+                element.replace(factory.createExpressionByPattern(pattern.first, receiver.text, *arguments.toTypedArray()))
             } else {
-                element.replace(factory.createExpressionByPattern(pattern.first, *arguments.toTypedArray()));
+                element.replace(factory.createExpressionByPattern(pattern.first, *arguments.toTypedArray()))
             }
         }
     }
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
@@ -43,7 +43,7 @@ class UnnecessaryJavaUsageInspection : AbstractKotlinInspection(), CleanupLocalI
 
                 holder.registerProblem(expression,
                                        "Unnecessary java usage",
-                                       ProblemHighlightType.LIKE_UNUSED_SYMBOL,
+                                       ProblemHighlightType.WEAK_WARNING,
                                        UnnecessaryJavaUsageFix(patterns[pattern]!!, value))
             }
         }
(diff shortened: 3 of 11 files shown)


idea/resources/inspectionDescriptions/UnnecessaryJavaUsage.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+Reports unnecessary java usage.
```
yole: This is not clear. A better description would be "Use of Java API that has a Kotlin equivalent" or something like that.

idea/src/META-INF/plugin.xml
```diff
@@ -1520,6 +1520,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.UnnecessaryJavaUsageInspection"
+                     displayName="Unnecessary java usage"
+                     groupName="Kotlin"
+                     enabledByDefault="true"
+                     cleanupTool="true"
+                     level="WARNING"
```
yole: This should be a WEAK_WARNING at most. There's nothing wrong with the original code - it's simply not as concise as it could be.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,64 @@
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitQualifiedExpression(expression: KtQualifiedExpression) {
+                super.visitQualifiedExpression(expression)
+
+                val selectorExpression = expression.selectorExpression ?: return
+                if (selectorExpression !is KtCallExpression) return
+                if (selectorExpression.valueArguments.size != 1) return
+                val value = selectorExpression.valueArguments[0].text
+                val pattern = expression.text.replace(value, "$0")
```
yole: This is not good enough. You need to resolve the call to make sure that it points to the correct method, and not to a method of a different class that also happens to be called `Collections`.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,64 @@
...
+                if (!patterns.containsKey(pattern)) return
+
+                holder.registerProblem(expression,
+                                       "Unnecessary java usage",
+                                       ProblemHighlightType.LIKE_UNUSED_SYMBOL,
```
yole: LIKE_UNUSED_SYMBOL highlighting is applied to code fragments that can be removed entirely. In this case, the quick fix is not removing the code, but is replacing it with different code.

idea/src/META-INF/plugin.xml
```diff
@@ -1520,6 +1520,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.UnnecessaryJavaUsageInspection"
+                     displayName="Unnecessary java usage"
```
yole: This is no more clear than the same text in the inspection description, and also needs to be changed.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,71 @@
...
+                super.visitQualifiedExpression(expression)
+
+                val selectorExpression = expression.selectorExpression ?: return
+                if (selectorExpression !is KtCallExpression) return
+                val value = selectorExpression.valueArguments.singleOrNull() ?: return
+
+                val calleeExpression = selectorExpression.calleeExpression as KtSimpleNameExpression
+                val bindingContext = calleeExpression.analyze(BodyResolveMode.PARTIAL)
+                val target = calleeExpression.mainReference.resolveToDescriptors(bindingContext).singleOrNull() ?: return
+                val pattern = target.fqNameSafe.asString()
```
ensirius: I tryed use ```importableFqName``` but it's was giving null for me. I am a bit confused with this.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -36,15 +40,18 @@ class UnnecessaryJavaUsageInspection : AbstractKotlinInspection(), CleanupLocalI
 
                 val selectorExpression = expression.selectorExpression ?: return
                 if (selectorExpression !is KtCallExpression) return
-                if (selectorExpression.valueArguments.size != 1) return
-                val value = selectorExpression.valueArguments[0].text
-                val pattern = expression.text.replace(value, "$0")
+                val value = selectorExpression.valueArguments.singleOrNull() ?: return
+
+                val calleeExpression = selectorExpression.calleeExpression as KtSimpleNameExpression
```
yole: The calleeExpression is not always a KtSimpleNameExpression, so you should use a safe cast here.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -28,41 +28,47 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
...
-                val pattern = target.fqNameSafe.asString()
-                if (!patterns.containsKey(pattern)) return
-
-                holder.registerProblem(expression,
-                                       "Unnecessary java usage",
-                                       ProblemHighlightType.WEAK_WARNING,
-                                       UnnecessaryJavaUsageFix(patterns[pattern]!!, value.text))
+                val pattern = patterns[target.fqNameSafe.asString()] ?: return
+                val javaUsageFix = UnnecessaryJavaUsageFix(pattern, expression.receiverExpression, arguments.map{it.text})
+                holder.registerProblem(expression, "Unnecessary java usage", ProblemHighlightType.WEAK_WARNING, javaUsageFix)
```
yole: Once again "unnecessary java usage".

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -28,41 +28,47 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 class UnnecessaryJavaUsageInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
 
     val patterns = mapOf(
-            "java.io.PrintStream.println" to "println($0)",
-            "java.io.PrintStream.print" to "print($0)",
-            "java.util.Collections.sort" to "$0.sort()"
+            "java.io.PrintStream.println" to Pair("println($0)", false),
```
yole: This boolean value is quite unclear. Much better to use an enum instead.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -28,41 +28,47 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
...
-                                       UnnecessaryJavaUsageFix(patterns[pattern]!!, value.text))
+                val pattern = patterns[target.fqNameSafe.asString()] ?: return
+                val javaUsageFix = UnnecessaryJavaUsageFix(pattern, expression.receiverExpression, arguments.map{it.text})
+                holder.registerProblem(expression, "Unnecessary java usage", ProblemHighlightType.WEAK_WARNING, javaUsageFix)
             }
         }
     }
 
-    private class UnnecessaryJavaUsageFix(val pattern: String, val value: String) : LocalQuickFix {
+    private class UnnecessaryJavaUsageFix(val pattern: Pair<String, Boolean>,
```
yole: Better to unpack the pair and to use separate arguments for the text and conversion mode.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.references.mainReference
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+
+class UnnecessaryJavaUsageInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+
+    private val patterns = mapOf(
+            "java.io.PrintStream.println" to Pair("println($0)", ConversionType.METHOD),
```
cypressious: What about other usages of  `PrintStream.println` outside of `System.out`, e.g. `System.err`? This will produce a lot of false positives.

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,82 @@
...
+            "java.io.PrintStream.print" to Pair("print($0)", ConversionType.METHOD),
+            "java.util.Collections.sort" to Pair("$0.sort()", ConversionType.METHOD),
+            "java.util.HashMap.put" to Pair("$0[$1] = $2", ConversionType.ASSIGNMENT)
```
cypressious: Why limit to `HashMap`?

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
```diff
@@ -0,0 +1,82 @@
...
+                super.visitQualifiedExpression(expression)
+
+                val selectorExpression = expression.selectorExpression
+                if(selectorExpression !is KtCallExpression) return
+                val args = selectorExpression.valueArguments
+                if(args.isEmpty()) return
+
+                val calleeExpression = selectorExpression.calleeExpression as? KtSimpleNameExpression ?: return
+                val bindingContext = calleeExpression.analyze(BodyResolveMode.PARTIAL)
+                val target = calleeExpression.mainReference.resolveToDescriptors(bindingContext).singleOrNull() ?: return
```
cypressious: You can use `findOriginalTopMostOverriddenDescriptors` e.g. to to get `MutableMap.put` instead of `HashMap.put`

idea/src/org/jetbrains/kotlin/idea/inspections/UnnecessaryJavaUsageInspection.kt
ilya-g: @ensirius  Could you ensure the inspection is only triggered when `put` result isn't used?

PR #874, opened 2016-05-30
Formatter: handle spaces around 'is' keyword

cypressious:
Fixes #KT-12548


idea/formatter/src/org/jetbrains/kotlin/idea/formatter/kotlinSpacingRules.kt
```diff
@@ -228,6 +228,8 @@ fun createSpacingBuilder(settings: CodeStyleSettings, builderUtil: KotlinSpacing
             betweenInside(REFERENCE_EXPRESSION, LAMBDA_ARGUMENT, CALL_EXPRESSION).spaces(1)
             betweenInside(TYPE_ARGUMENT_LIST, LAMBDA_ARGUMENT, CALL_EXPRESSION).spaces(1)
 
+            around(IS_KEYWORD).spaces(1)
```
yole: This is really minor, but still: to make the code easier to follow, I'd move this line next to the rule for the AS keyword.

cypressious: I moved the line and squashed the commits.

yole:
Thanks for your contribution!

PR #875, opened 2016-06-01
Formatter: handle spaces between class name and type parameters

cypressious:
Fixes #KT-12446


yole:
Thanks for your contribution!

PR #877, opened 2016-06-06
Add support for the Android test plugin.

JakeWharton:
[KT-11767](https://youtrack.jetbrains.com/issue/KT-11767)


JakeWharton:
Had a hell of a time trying to get this to compile let alone test locally. Going to rely on your CI here.

libraries/tools/kotlin-gradle-plugin/pom.xml
```diff
@@ -33,7 +33,7 @@
         <dependency>
             <groupId>com.android.tools.build</groupId>
             <artifactId>gradle</artifactId>
-            <version>1.1.0</version>
+            <version>1.3.0</version>
```
JakeWharton: I suppose this isn't strictly needed since the plugin is looked up by name.

tsvtkv:
I've made a minor fix to the test and merged the PR manually.

Thanks for the contribution!

PR #879, opened 2016-06-06
Formatter: handle spaces before value arguments list

cypressious:
Fixes #KT-12634


goodwinnk:
Merged manually with tests modification, thank you!
https://github.com/JetBrains/kotlin/commit/5c68681f962334bf58785346b3f48c909309ffe6

Feedback will be very important on this change because interesting corner cases are affected:
`synchronised (obj) {}`
`with (s) {}`

Maybe some options or advanced logic will be required in future for these cases.

PR #885, opened 2016-06-12
Implement quickfix for wrong primitive literal

cypressious:
Fixes: KT-12251

Not sure if I should disable the other quickfix which adds a call to `.toX()`

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/WrongPrimitiveLiteralFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/WrongPrimitiveLiteralFix.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2010-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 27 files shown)


cypressious:
Rewrote the fix and disabled the `.toX()` fix hen the new is active. Also added some tests.

idea/src/org/jetbrains/kotlin/idea/quickfix/WrongPrimitiveLiteralFix.kt
```diff
@@ -0,0 +1,96 @@
...
+            else if ('.' !in this) {
+                append(".0")
+            }
+        }
+        else {
+            if (constValue is Float || constValue is Double) {
+                append(constValue.toLong())
+            }
+            else {
+                append(element.text.takeWhile { it != 'l' && it != 'L' })
```
yole: I'd rather use `element.trimEnd('l', 'L')`

yole:
Looks good except for the minor style issue.

cypressious:
@yole fixed the style issue

PR #887, opened 2016-06-23
Formatter: Remove spaces before question mark in nullable typs

cypressious:
Fixes #KT-12830


goodwinnk:
Accepted with small modifications in commit message and ChangeLog.md. Thank you!

PR #888, opened 2016-06-23
Add quickfix for "Illegal usage of inline parameter"

cypressious:
Fixes #KT-12838


yole:
Thanks for your contribution!

PR #890, opened 2016-07-03
Fixed encoding problem between daemon and jps in some cases (KT-9728)

soywiz:
This PR tries to fix this issue: https://youtrack.jetbrains.com/issue/KT-9728

I have two comments related to this:
- Maybe the encoding should be placed in a constant shared between JPS and Daemon but don't know where, so I just used literals for now
- Don't know how to test this so left untested automatically, (but verified that the problem is fixed)

For some reason it seems that PrintStream/StringReader uses different default charsets between the daemon and the jps in some machines, so the fix tries to not use the default, but a known charset.

### DIFF
--- a/compiler/daemon/src/org/jetbrains/kotlin/daemon/CompileServiceImpl.kt
+++ b/compiler/daemon/src/org/jetbrains/kotlin/daemon/CompileServiceImpl.kt
@@ -414,8 +414,8 @@ class CompileServiceImpl(
                 compilationsCounter.incrementAndGet()
                 val rpcProfiler = if (daemonOptions.reportPerf) WallAndThreadTotalProfiler() else DummyProfiler()
                 val eventManger = EventMangerImpl()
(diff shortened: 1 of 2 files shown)


bashor:
Good job, @soywiz!

I've already fixed it as KT-9954 ([rr/z/kt9954](https://github.com/JetBrains/kotlin/tree/rr/z/kt9954)), but didn't push it yet because some tests failed on some agents on CI.

I'll take your changes in KotlinCompilerRunner.kt manually and push the fix as soon as possible.

Thanks for your effort!

soywiz:
Woops. Didn't know this was investigated and fixed already. The issues was from last year, and since it still happened on master/1.1 I though that no one investigated it yet. And I finally was able to compile and run master by myself so I decided to look at it.

Cool. Feel free to grab what you need :+1: and to close this when done.

And thank you for looking at this!

bashor:
Looks like the issue and PR obsolete, so I close it.

Anyway, thanks again!

PR #892, opened 2016-07-08
KT-9228 Implement inspection to remove @ used as annotation argument

shiraji:
#KT-9228 Fixed

https://youtrack.jetbrains.com/issue/KT-9228


mglukhikh:
IMO the solution is unnecessary complex. Why you do not want to take KtAnnotatedExpression (and not KtAnnotated) as the basic for your intention? You can mark each such expression separately if 'at' is unnecessary. I think it will be more clear for the user.

shiraji:
I agree with you. When i first tried to use `KtAnnotatedExpression`, no events occur. Maybe some settings were wrong. Let me fix this up.

shiraji:
Could you review again @mglukhikh ? Now use `KtAnnotatedExpression`. Let me know if you want me to squash the commit.

mglukhikh:
By the way: files like IntentionTestGenerated.java (to be precise, *TestGenerated.java) should never be changed manually. To regenerate them, please use "Generate Tests" run configuration (I've already done it for your changes).

shiraji:
> files like IntentionTestGenerated.java (to be precise, *TestGenerated.java) should never be changed manually. To regenerate them, please use "Generate Tests" run configuration (I've already done it for your changes).

i didn't know that. i will run it in next time. Thanks!

mglukhikh:
Refactored and added manually to master and 1.0.4 branches.

shiraji:
Thanks!

mglukhikh:
Thank you for your help

PR #893, opened 2016-07-12
Build string with initial capacity

K0zka:
this method helps save computation time when the developer knows or have a good
educated guess about the size of the generated string and have to use control
structures while filling the string builder

### DIFF
--- a/libraries/stdlib/src/kotlin/text/StringBuilder.kt
+++ b/libraries/stdlib/src/kotlin/text/StringBuilder.kt
@@ -13,7 +13,7 @@ public inline fun buildString(builderAction: StringBuilder.() -> Unit): String =
  * Builds new string by populating newly created [StringBuilder] initialized with the given capacity using provided [builderAction] and then converting it to [String].
  */
 @kotlin.internal.InlineOnly
-inline fun buildString(length : Int, builderAction: StringBuilder.() -> Unit): String = StringBuilder(length).apply(builderAction).toString()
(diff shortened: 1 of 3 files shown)


ilya-g:
Going to merge this one soon.

libraries/stdlib/src/kotlin/text/StringBuilder.kt
```diff
@@ -10,6 +10,12 @@ package kotlin.text
 public inline fun buildString(builderAction: StringBuilder.() -> Unit): String = StringBuilder().apply(builderAction).toString()
 
 /**
+ * Builds new string by populating newly created [StringBuilder] initialized with the given capacity using provided [builderAction] and then converting it to [String].
+ */
```
ilya-g: Will fix formatting upon rebase.

mglukhikh:
I think everything is OK with this PR

ilya-g:
@K0zka I've merged it manually, thanks for PR.

ilya-g:
@mikehearn had some concerns about this overload in slack, I'll quote him here:

> i guess my main concern/thought is that stdlib methods that abbreviate code by only a tiny amount should probably be things that you use all the time. map/filter/forEach/etc are all like this

> pre-sized StringBuilders are usually an optimisation and that's not something you see every few lines

K0zka:
I understand that, at the same time I need optimizations, please decide if you want it in kotlin or I should keep it in my own project.

PR #896, opened 2016-07-14
new function body changed to call the shorter TODO() function

K0zka:
Signed-off-by: Laszlo Hornyak <laszlo.hornyak@gmail.com>


K0zka:
purely cosmetic change, TODO("something") is so much more easy to spot since it means code is not ready, compared to throwing an exception - which in some cases can be the implementation of the function

yole:
Thanks for your contribution!

PR #897, opened 2016-07-15
Debugger: check that label for marked object is a valid java identifier

NataliaUkhorskaya:
#KT-12651 Fixed

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinCodeFragmentFactory.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinCodeFragmentFactory.kt
@@ -43,14 +43,13 @@ import org.jetbrains.eval4j.jdi.asValue
 import org.jetbrains.kotlin.asJava.KtLightClass
 import org.jetbrains.kotlin.idea.KotlinFileType
 import org.jetbrains.kotlin.idea.codeInsight.CodeInsightUtils
+import org.jetbrains.kotlin.idea.core.quoteIfNeeded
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinCodeFragmentFactory.kt
```diff
@@ -245,10 +245,12 @@ class KotlinCodeFragmentFactory: CodeFragmentFactory() {
         fun createCodeFragmentForLabeledObjects(project: Project, markupMap: Map<*, ValueMarkup>): Pair<String, Map<String, Value>> {
             val sb = StringBuilder()
             val labeledObjects = HashMap<String, Value>()
+            val psiNameHelper = PsiNameHelper.getInstance(project)
+
             val entrySet: Set<Map.Entry<*, ValueMarkup>> = markupMap.entries
             for ((value, markup) in entrySet) {
                 val labelName = markup.text
-                if (!Name.isValidIdentifier(labelName)) continue
+                if (!Name.isValidIdentifier(labelName) || !psiNameHelper.isIdentifier(labelName)) continue
```
goodwinnk: Do we need Name.isValidIdentifier() here at all?

NataliaUkhorskaya: You're right. I've updated the commit.

goodwinnk:
Merged manually with slight modifications (in ChangeLog.md and commit message) into master and 1.0.4.

Thank you!

PR #898, opened 2016-07-18
KT-12697 Expand selection action select ": Type"

shiraji:
#KT-12697 Fixed

https://youtrack.jetbrains.com/issue/KT-12697

I'm not sure how to add test cases for "expand selection".

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinFunctionReturnTypeSelectioner.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinFunctionReturnTypeSelectioner.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2010-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.idea.editor.wordSelection
+
+import com.intellij.codeInsight.editorActions.ExtendWordSelectionHandlerBase
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.psi.KtTypeReference
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+class KotlinFunctionReturnTypeSelectioner : ExtendWordSelectionHandlerBase() {
+
+    override fun canSelect(e: PsiElement?): Boolean {
+        return e is KtTypeReference
+    }
+
+    override fun select(e: PsiElement?, editorText: CharSequence?, cursorOffset: Int, editor: Editor): List<TextRange>? {
+        e ?: return null
+        editorText ?: return null
+        val position = getColonPosition(e, editorText)
+        if (position < 0) return null
+        return listOf(TextRange(position, e.endOffset))
+    }
+
+    private fun getColonPosition(e: PsiElement, editorText: CharSequence): Int {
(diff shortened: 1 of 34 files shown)


yole:
The tests for this feature are [here](https://github.com/JetBrains/kotlin/blob/master/idea/tests/org/jetbrains/kotlin/idea/WordSelectionTest.java). Please do add the tests for the changes you're making.

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinFunctionReturnTypeSelectioner.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun select(e: PsiElement?, editorText: CharSequence?, cursorOffset: Int, editor: Editor): List<TextRange>? {
+        e ?: return null
+        editorText ?: return null
+        val position = getColonPosition(e, editorText)
+        if (position < 0) return null
+        return listOf(TextRange(position, e.endOffset))
+    }
+
+    private fun getColonPosition(e: PsiElement, editorText: CharSequence): Int {
+        val colonIndex = editorText.filterIndexed { i, c -> i < e.startOffset }.indexOfLast { it == ':' }
```
yole: Please do not use the text or regular expressions to perform the action. All the necessary information is available in the PSI; please do use it.

shiraji: I changed to use `CharArrayUtil`

yole: As I said, please use the PSI. Go up the tree to find a `KtCallableDeclaration` and use its `getColon` method to locate the colon. This will ensure that the action will work only where needed (and will not work in parameter type declarations, where it's not needed).

shiraji: I didn't know much about PSI class. I will study more about PSI.

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinSuperTypeSelectioner.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+class KotlinSuperTypeSelectioner : ExtendWordSelectionHandlerBase() {
+    override fun canSelect(e: PsiElement?): Boolean {
+        return e is KtSuperTypeList
+    }
+
+    override fun select(e: PsiElement?, editorText: CharSequence?, cursorOffset: Int, editor: Editor): List<TextRange>? {
+        e ?: return null
```
yole: Those parameters are actually non-null. You don't need the null checks here and in the other selectioner.

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinSuperTypeSelectioner.kt
```diff
@@ -0,0 +1,49 @@
...
+        editorText ?: return null
+        val colonPosition = getColonPosition(e)
+        if (colonPosition < 0) return null
+        return listOf(TextRange(colonPosition, e.endOffset))
+    }
+
+    private fun getColonPosition(e: PsiElement): Int {
+        var parent = e.parent
```
yole: You can use the `PsiElement.getParentOfType()` extension function to do this with less code.

shiraji: This makes code simpler! Thanks.

shiraji:
@yole Do you want me to squash the commits?

yole:
Right now the selectioner handles every variable type: for example, if you have `fun foo(bar: Bar)`, it will select `: Bar`. I don't see why this is useful for parameters or local variables. This should work only for supertypes of a class and the return type of a function.

And once again sorry that the review takes so long; I've been busy with other things :(

shiraji:
For local variables, `val foo: String = ""`, developers want to omit ": String" using expand selection.
For parameters, you are right. I will fix this.

I understand you are busy. I had no idea what was going on. I thought I was doing something wrong. Thanks for your response!

shiraji:
I found another issue for this PR

```kotlin
    fun foo(): String {
        foo2(object : <caret>DocumentAdapter(){
        })
        return ""
    }
```

Highlight like this.

```kotlin
    fun foo()<select>: String {
        foo2(object : DocumentAdapter()</select>{
        })
        return ""
    }
```

Looks not good...

yole:

PR #901, opened 2016-07-19
KT-9009 fix

semoro:


yole:
- Please provide tests for the feature you're adding. See [here](https://github.com/JetBrains/kotlin/tree/master/idea/testData/addImport) and [here](https://github.com/JetBrains/kotlin/blob/master/idea/tests/org/jetbrains/kotlin/addImport/AbstractAddImportTest.kt) for existing auto import tests.

- Please avoid copy/pasting existing implementation classes. If you need to reuse code from them, please refactor them so that there is no duplication.

semoro:
There is no tests for auto import quick fix, so how we should test quickfix?

PR #904, opened 2016-07-21
KT-12887 Extend selection should select call's invoked expression

shiraji:
#KT-12887 fixed

This probably conflicts with #898


yole:
Thanks for your contribution!

PR #908, opened 2016-07-28
KT-13011 Inspection for unnecessary lateinit

shiraji:
#KT-13011 Fixed

https://youtrack.jetbrains.com/issue/KT-13011

### DIFF
--- a/idea/resources/inspectionDescriptions/RemoveUnnecessaryLateinit.html
+++ b/idea/resources/inspectionDescriptions/RemoveUnnecessaryLateinit.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection detects lateinit that can be safely removed
+</body>
+</html>
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1273,6 +1273,11 @@
       <category>Kotlin</category>
     </intentionAction>
 
+    <intentionAction>
+      <className>org.jetbrains.kotlin.idea.intentions.RemoveUnnecessaryLateinitIntention</className>
+      <category>Kotlin</category>
+    </intentionAction>
+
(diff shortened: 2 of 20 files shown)


mglukhikh:
You forgot to add RemoveUnnecessaryLateinit.html into idea/resources/inspectionDescriptions

mglukhikh:
Also I would say that 'LIKE_UNUSED_SYMBOL' highlighting would be better for this inspection.

mglukhikh:
Also the method you are using to resolve secondary constructor dependency is not the most optimal (you analyze the same constructor more than once in recursion). Moreover, it's dangerous when some constructor accidentally delegates to itself, AFAIC you will have stack overflow as result.

shiraji:
Do you have suggestion for how to resolve secondary constructors?

shiraji:
> You forgot to add RemoveUnnecessaryLateinit.html into idea/resources/inspectionDescriptions

oops...now I found a big problem with previous PR. I don't remember adding .html for https://github.com/JetBrains/kotlin/pull/895/ . Do you want me to send another PR for adding RemoveAtmarkUsedAsAnnotationArgumentInspection.html ?

shiraji:
> Also I would say that 'LIKE_UNUSED_SYMBOL' highlighting would be better for this inspection (to change this, you can override problemHighlightType)

This mean, for this issue, can I remove Intention? like [UnusedSymbolInspection](https://github.com/JetBrains/kotlin/blob/a0226f3fd812285dccce5c9c279c4da8900396ed/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedSymbolInspection.kt) does

mglukhikh:
Added manually to 1.0.4, will add to master branch tomorrow morning. Thanks for your efforts!

shiraji:
Thanks!

PR #909, opened 2016-08-01
KT-13113 Add inspection to detect single-expression string template

shiraji:
#KT-13113 Fixed

https://youtrack.jetbrains.com/issue/KT-13113

Feature that adds `toString()` to non-String class is discussed at #kontributor channels.

### DIFF
--- a/idea/resources/inspectionDescriptions/RemoveSingleExpressionStringTemplate.html
+++ b/idea/resources/inspectionDescriptions/RemoveSingleExpressionStringTemplate.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection reports single-expression String template that can be safely removed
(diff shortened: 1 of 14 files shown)


mglukhikh:
Added manually to 1.0.4 branch. Will add to master branch today later. BTW: in its original state, intention was suggested on simple strings, like "Hello", I've fixed it. Also, I decided that inspection should be reported only for string types, otherwise we should have only intention (syntax with .toString() is longer, and performance boost got is very questionable).

mglukhikh:
Thanks!

shiraji:
> intention was suggested on simple strings, like "Hello", I've fixed it

Oops. I should have test case for that. Thanks anyway!

PR #910, opened 2016-08-01
Correction dots place

nnsnodnb:


yole:
What are you trying to achieve with this pull request? There is nothing wrong about the position of the dots in the original document, and no visible difference in rendering between the original and the new version.

PR #913, opened 2016-08-03
Allow to evaluate kotlin expression in Java Files

NataliaUkhorskaya:


compiler/frontend/src/org/jetbrains/kotlin/psi/KtCodeFragment.kt
```diff
@@ -67,7 +73,15 @@ abstract class KtCodeFragment(
 
     override fun isValid() = true
 
-    override fun getContext() = context
+    override fun getContext(): PsiElement? {
+        if (fakeContextForJavaFile != null) return fakeContextForJavaFile
+        if (context !is KtElement) {
+            LOG.warn("CodeFragment with non-kotlin context should have fakeContextForJavaFile set: \noriginalContext = ${context?.getElementTextWithContext()}")
```
NataliaUkhorskaya: We shouldn't create kotlin codeFragments for scala files. See KotlinCodeFragmentFactory.isContextAccepted

idea/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinCodeFragmentFactory.kt
```diff
@@ -115,9 +116,73 @@ class KotlinCodeFragmentFactory: CodeFragmentFactory() {
             }
         })
 
+        if (contextElement != null && contextElement !is KtElement) {
```
NataliaUkhorskaya: Now such variables are skipped during text rendering, so it doesn't matter if they are present in resulting map (this map is used only to get values for kotlin variables that are present in resulting text)

idea/src/org/jetbrains/kotlin/idea/debugger/evaluate/KotlinCodeFragmentFactory.kt
```diff
@@ -115,9 +116,73 @@ class KotlinCodeFragmentFactory: CodeFragmentFactory() {
...
     }
 
+    private fun getVisibleLocalVariables(contextElement: PsiElement?, debuggerContext: DebuggerContextImpl): Map<LocalVariable, Value>? {
+        val semaphore = Semaphore()
+        semaphore.down()
+
+        var visibleVariables: Map<LocalVariable, Value>? = null
+
+        val worker = object : DebuggerCommandImpl() {
+            override fun action() {
```
NataliaUkhorskaya: I've replaced it with codeFragment.getContentElement().statements.lastOrNull() to make this line clearer

PR #915, opened 2016-08-09
Better teamcity integration

cy6erGn0m:


cy6erGn0m:
The idea is to throw compiler exception that Teamcity is waiting for (there is a special Teamcity maven agent that can listen for exceptions). We have our own based on the required type with additional information that Teamcity can use in the future.

libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
```diff
@@ -133,15 +138,32 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                        CollectionsKt.map(messageCollector.getCollectedErrors(), new Function1<Pair<CompilerMessageLocation, String>, CompilerMessage>() {
+                            @Override
+                            public CompilerMessage invoke(Pair<CompilerMessageLocation, String> pair) {
+                                String lineContent = pair.getFirst().getLineContent();
+                                int lineContentLength = lineContent == null ? 0 : lineContent.length();
+
+                                return new CompilerMessage(
+                                        pair.getFirst().getPath(),
+                                        CompilerMessage.Kind.ERROR,
+                                        pair.getFirst().getLine(),
```
cy6erGn0m: Are they zero-based `CompilerMessageLocation` ?

libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
```diff
@@ -133,15 +138,32 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
 
         ExitCode exitCode = compiler.exec(messageCollector, Services.EMPTY, arguments);
 
         switch (exitCode) {
             case COMPILATION_ERROR:
-                throw new MojoExecutionException("Compilation error. See log for more details");
             case INTERNAL_ERROR:
-                throw new MojoExecutionException("Internal compiler error. See log for more details");
+                throw new KotlinCompilationFailureException(
+                        CollectionsKt.map(messageCollector.getCollectedErrors(), new Function1<Pair<CompilerMessageLocation, String>, CompilerMessage>() {
```
cy6erGn0m: Isn't compiler exception logged as error?

cy6erGn0m:
it's too difficult to do the same as Teamcity does. Unfortunately see no easy way to do a test

libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
cy6erGn0m: -1 should be 0 so CompileMessage will handle it properly. I've extracted a function `fixLocation` to handle it.

libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
```diff
@@ -133,16 +137,32 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                        @Override
+                        public CompilerMessage invoke(Pair<CompilerMessageLocation, String> pair) {
+                            CompilerMessageLocation location = pair.getFirst();
+                            String message = pair.getSecond();
+                            String lineContent = location.getLineContent();
+                            int lineContentLength = lineContent == null ? 0 : lineContent.length();
+
+                            return new CompilerMessage(
+                                    location.getPath(),
+                                    CompilerMessage.Kind.ERROR,
```
cy6erGn0m: yes, but in this case it looks not so clear that there is -1 and 0 with special meaning

PR #916, opened 2016-08-11
KT-6975 Quickfix for adding "inline" when "reified" is specified

shiraji:
#KT-6975 Fixed

https://youtrack.jetbrains.com/issue/KT-6975


mglukhikh:
Sorry but in this case intention / inspection pair is not appropriate. There is already a compiler error about it: REIFIED_TYPE_PARAMETER_NO_INLINE, so you should not detect it second time. Instead, a quick-fix should be added for this compiler error. You can look at QuickFixRegistrar.kt as binding place for such quick-fixes and ChangeVisibilityOnExposureFactory.kt as possible example.

shiraji:
Ah! Didn't know there is such thing. I will fix this.

shiraji:
Implemented.

mglukhikh:
Thanks, I will merge it today manually.

shiraji:
Thanks

PR #918, opened 2016-08-12
KT-13336 Fix problem with appending literal to variable omitting braces

shiraji:
#KT-13336 Fixed

https://youtrack.jetbrains.com/issue/KT-13336


mglukhikh:
Polished a little bit, added a few tests, merged manually into master (will merge into 1.0.4 today later).

PR #919, opened 2016-08-14
Allow suppressing warnings for secondary constructor

cypressious:
Fixes #KT-12627


goodwinnk:
Please add a test for suppression action availability. https://github.com/JetBrains/kotlin/blob/master/idea/testData/quickfix/suppress/declarationKinds/class.kt file can be used as example.

cypressious:
Done

goodwinnk:
Great, thanks! The fix is now merged into 1.0.4 and master branches.

PR #920, opened 2016-08-14
Highlight use of toString() inside string interpolation as redundant

cypressious:
Fixes #KT-10731


mglukhikh:
I polished it a little bit: changed inspection type to LIKE_UNUSED_SYMBOL; added change of `foo = ${foo.toString()}` just to `foo = $foo` without curly braces + some minor details. Merged manually into master (today later into 1.0.4). Thanks.

PR #921, opened 2016-08-14
Make folding if to return/assignment work with if-else-if-else

cypressious:
Fixes: #KT-13452


mglukhikh:
It handles the following case incorrectly:
```
fun test(n: Int) {
    var a: String = ""
//  v Replace if with assignment: incorrect
    if (n == 1)
        a = "one"
    else if (n == 2)
        a = "two"
}
```

cypressious:
I handle it for return but forgot to copy it over for assignments.

mglukhikh:
It should be enough to check here that last `if` in branches has `else`.

cypressious:
I'll fix it later today.

cypressious:
Done

mglukhikh:
OK thanks, pushed manually into master, will push into 1.0.4 a bit later.

PR #922, opened 2016-08-14
Implement intentions to add/remove braces to/from when entries

cypressious:
Fixes: #KT-12043


mglukhikh:
I've added a few minor details and merged manually into master (today later in 1.0.4). Thanks.

PR #928, opened 2016-08-17
KT-4519 Show first line with content of block comment when collapsed

shiraji:
#KT-4519 Fixed

https://youtrack.jetbrains.com/issue/KT-4519

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
@@ -111,12 +111,11 @@ class KotlinFoldingBuilder : CustomFoldingBuilder(), DumbAware {
     }
 
     private fun getCommentContents(line: String): String {
-        val comment = KotlinCommenter()
         return line.trim()
-                .replace(comment.documentationCommentPrefix!!, "")
-                .replace(comment.blockCommentPrefix!!, "")
-                .replace(comment.blockCommentSuffix!!, "")
-                .replace(comment.documentationCommentLinePrefix!!, "")
+                .replace("/**", "")
+                .replace("/*", "")
+                .replace("*/", "")
+                .replace("*", "")
     }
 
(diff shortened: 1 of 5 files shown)


idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
```diff
@@ -97,12 +97,28 @@ class KotlinFoldingBuilder : CustomFoldingBuilder(), DumbAware {
...
 
+    private fun getFirstLineOfComment(node: ASTNode): String {
+        val targetCommentLine = node.text.split("\n").firstOrNull {
+            getCommentContents(it).isNotEmpty()
+        } ?: return ""
+        return getCommentContents(targetCommentLine)
+    }
+
+    private fun getCommentContents(line: String): String {
+        val comment = KotlinCommenter()
```
yole: There is no need to create a new object here; it's OK to hard-code the token texts.

idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
```diff
@@ -97,12 +97,28 @@ class KotlinFoldingBuilder : CustomFoldingBuilder(), DumbAware {
...
+        return line.trim()
+                .replace(comment.documentationCommentPrefix!!, "")
+                .replace(comment.blockCommentPrefix!!, "")
+                .replace(comment.blockCommentSuffix!!, "")
+                .replace(comment.documentationCommentLinePrefix!!, "")
```
yole: Please make sure that the line has at most one leading and at most one trailing space after you've remove the prefix and suffix

shiraji: How about trimming both of spaces here and add one leading/trailing space at https://github.com/JetBrains/kotlin/pull/928/files#diff-e4fc5e886f7f1903e7f5f67545e46564R110

yole:

shiraji:
Thanks!

PR #933, opened 2016-08-20
Add quickfix for "A type annotation is required on a value parameter" 

cypressious:
https://youtrack.jetbrains.com/issue/KT-12804


mglukhikh:
Merged manually with minor corrections into master. Thanks!

PR #937, opened 2016-08-22
KT-13521 Warning right part of "expression ?: null" is useless

shiraji:
#KT-13521 Fixed

https://youtrack.jetbrains.com/issue/KT-13521


mglukhikh:
Thanks!

shiraji:
That's quick. Thanks!

PR #938, opened 2016-08-23
KT-5193 Make multiline strings foldable

shiraji:
#KT-5193 Fixed

https://youtrack.jetbrains.com/issue/KT-5193


idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
```diff
@@ -99,10 +100,18 @@ class KotlinFoldingBuilder : CustomFoldingBuilder(), DumbAware {
     override fun getLanguagePlaceholderText(node: ASTNode, range: TextRange): String = when {
         node.elementType == KtTokens.BLOCK_COMMENT -> "/${getFirstLineOfComment(node)}.../"
         node.elementType == KDocTokens.KDOC -> "/**${getFirstLineOfComment(node)}...*/"
+        node.elementType == KtNodeTypes.STRING_TEMPLATE -> "\"\"\"${getFirstLineOfString(node)}...\"\"\""
         node.psi is KtImportList -> "..."
         else ->  "{...}"
     }
 
+    private fun getFirstLineOfString(node: ASTNode): String {
+        val targetStringLine = node.text.split("\n").firstOrNull {
```
yole: You can avoid multiple `replace` calls if you use `asSequence().map { it.replace(...) }.firstOrNull { it.isNotBlank }`

idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
```diff
@@ -99,10 +100,18 @@ class KotlinFoldingBuilder : CustomFoldingBuilder(), DumbAware {
...
+            it.replace("\"\"\"", "").isNotBlank()
+        } ?: return ""
+        return " ${targetStringLine.replace("\"\"\"", "").trim()} "
```
yole: It's better to add surrounding spaces in `getLanguagePlaceholderText()`, not in this function. (This applies also to `getFirstLineOfComment()`.)

shiraji: But then if the string or the comment is empty, user will get `"""  ..."""`

idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
shiraji: Make sense. I will fix this.

idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
yole: Yes, you're right that the current solution produces somewhat cleaner results.

yole:
Thanks for your contribution!

shiraji:
Thanks for reviewing this pull request

PR #940, opened 2016-08-30
Add data annotation to JeName for equals/hashCode

grandstaish:
The docs say that this class must obey the equals contract (https://docs.oracle.com/javase/7/docs/api/javax/lang/model/element/Name.html)


grandstaish:
Actually I'm wrong about this, so closing the PR for now.

PR #942, opened 2016-08-30
Kt-12205 fix

semoro:
https://youtrack.jetbrains.com/issue/KT-12205

Problem caused by non-implemented clauses in KotlinInficesHelper, now fixed.


semoro:
Found breaking test, work in progress to fix this problem

semoro:
Fixed all problems, ready for review

compiler/light-classes/src/org/jetbrains/kotlin/asJava/LightClassUtil.kt
```diff
@@ -185,6 +169,15 @@ object LightClassUtil {
         return null
     }
 
+    private fun getWrappingClasses(declaration: KtDeclaration): List<PsiClass> {
```
yole: You can simplify this code by changing the return type to `Sequence<PsiClass>` and using `generateSequence()`

idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/KotlinShortNamesCache.kt
```diff
@@ -85,30 +87,78 @@ class KotlinShortNamesCache(private val project: Project) : PsiShortNamesCache()
...
+                .flatMap { LightClassUtil.getLightClassPropertyMethods(it).allDeclarations }
+                .filter { it.name == name }
+                .map { it as? PsiMethod }
+
+        return (propertyAccessorsPsi + kotlinFunctionsPsi).filterNotNull().toList().toTypedArray()
+    }
+
+    override fun getMethodsByNameIfNotMoreThan(name: String, scope: GlobalSearchScope, maxCount: Int): Array<PsiMethod> {
+        val methods = getMethodsByName(name, scope)
+        if (methods.size > maxCount)
```
yole: This implementation defeats the purpose of this method. You should calculate the candidates lazily and stop evaluating if you've found more than `maxCount` candidates.

idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/KotlinShortNamesCache.kt
```diff
@@ -85,30 +87,78 @@ class KotlinShortNamesCache(private val project: Project) : PsiShortNamesCache()
...
-    override fun getAllMethodNames(): Array<String>
-            = emptyArray()
+    override fun getAllMethodNames(): Array<String> {
+        val functionIndex = KotlinFunctionShortNameIndex.getInstance()
+        val functionNames = functionIndex.getAllKeys(project)
+
+        val propertiesIndex = KotlinPropertyShortNameIndex.getInstance()
+        val globalScope = GlobalSearchScope.allScope(project)
+        val propertyAccessorNames = propertiesIndex.getAllKeys(project)
+                .flatMap { propertiesIndex.get(it, project, globalScope) }
```
yole: This will be extremely slow. It's OK to return names for which `getMethodsByName()` will then not find any methods, but it's not OK to look up every single method in the index.

idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/KotlinShortNamesCache.kt
```diff
@@ -85,30 +87,78 @@ class KotlinShortNamesCache(private val project: Project) : PsiShortNamesCache()
...
-    override fun getFieldsByName(name: String, scope: GlobalSearchScope): Array<PsiField>
-            = emptyArray()
+    override fun getAllFieldNames(): Array<String> {
+        val propertiesIndex = KotlinPropertyShortNameIndex.getInstance()
+        val globalScope = GlobalSearchScope.allScope(project)
 
-    override fun getAllFieldNames(): Array<String>
-            = emptyArray()
+        return propertiesIndex.getAllKeys(project)
+                .flatMap { propertiesIndex.get(it, project, globalScope) }
```
yole: Same here. No property index lookups please.

yole:
Reviewed the code so far, please see comments.

idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/KotlinShortNamesCache.kt
semoro: It is normal that in some cases it will return names of not-existing methods?
Like `setFoo` for 
```kotlin
 val foo = "" 
```

idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/KotlinShortNamesCache.kt
semoro: But we unable to determine is property has an @JvmField and it's backing field available for access without having instance of KtProperty.
Or we should return all possible names without this check?  Or some approach like `KotlinPropertyShortNameIndex.getInstance().getAllValues(project)` is possible?

semoro:
Working on complex test, to cover possible aspects of KotlinShortNamesCache

yole:
The pull request introduces an unfortunate regression: all Kotlin methods are shown twice in the Goto Symbol dialog (one version coming from KotlinGotoSymbolContributor and another from the new KotlinShortNameCache implementation). For methods that have a custom `@JvmName` this is actually fine: it's useful to show them with both their Java and Kotlin name. For those that don't, this should be fixed.

semoro:
@yole `getMethodsByNameIfNotMoreThan`, `getFieldsByNameIfNotMoreThan` now optimized by time consumption in case of elements count <= maxCount are found, may be should optimize it by memory in case found > maxCount instead?

idea/idea-core/src/org/jetbrains/kotlin/idea/core/KotlinIndicesHelper.kt
```diff
@@ -61,6 +66,13 @@ class KotlinIndicesHelper(
 
     private val moduleDescriptor = resolutionFacade.moduleDescriptor
     private val project = resolutionFacade.project
+    private val scopeWithoutKotlin = ScopeWithoutKotlin(scope)
```
yole: You can use `SearchUtil.excludeKotlinSources`

yole: Nevermind, made the change myself.

yole:

PR #945, opened 2016-09-06
KT-13674 Add quickfix for FINAL_UPPER_BOUND

shiraji:
https://youtrack.jetbrains.com/issue/KT-13674


mglukhikh:
Quick-fix you have implemented is not `RemoveFinalUpperBoundFix`. I'd name it like `InlineTypeParameterFix`, because it removes not a bound but a parameter itself, replacing its usages with usages of the bound. I'll integrate it in the source with some renaming & polishing but in fact it's not a solution of KT-13674.

shiraji:
What would be the expected solution for this issue?

mglukhikh:
As stated, it should remove a bound and leave parameter as is.

mglukhikh:
However, I integrated your fix in master & 1.0.5, so thank you, this fix is also helpful.

shiraji:
Thanks. I misunderstand the word "bound".

```
data class DC<T : Int>(val x: T, val y: String)
```

```
data class DC<T>(val x: T, val y: String)
```

Let me work on this, too.

mglukhikh:
OK, with pleasure.

PR #947, opened 2016-09-09
KT-9590 reified: Diagnostic is misleading, quickfix wanted

semoro:
https://youtrack.jetbrains.com/issue/KT-9590

Fixed diagnostic type for this issue
Added QuickFix to add reified for type parameters of function if needed. 

Also fixed strange bug with wrong diagnostic creating argument, so it's origin was wrong for
```Kotlin
fun <T: Iterable<Array<R>>, R> T.flatten(): Array<R> {
    return this.flatMap { it.asIterable() }.toTypedArray<caret>()
}
```
It supposed that `T` is not reified, but really `R` should be `reified`
```Kotlin
fun <T: Iterable<Array<R>>, reified R> T.flatten(): Array<R> {
    return this.flatMap { it.asIterable() }.toTypedArray<caret>()
}
```


mglukhikh:
I'll integrate it in master & 1.0.5 today, so thank you. Some notes: you've forgotten to add 'inline' modifier to a function; the recommended way to create quick-fix test is separate *.kt and *.kt.after files, and stuff for multi-file tests is not needed for them. Also, I've performed some polishing.

semoro:
@mglukhikh 
you've forgotten to add 'inline' modifier to a function;
I'am not forgotten, this is an intentional solution, cause we already has quickfix which add inline if reified are present, but it you think that it should add both reified and inline, than ok)

PR #949, opened 2016-09-10
Fix KtTypeReference.hasParentheses() to check for both LPAR and RPAR

calexHG:
A year old typo had this redundantly checking for LPAR twice, rather than checking once for LPAR and once for RPAR.


yole:
Thanks for your contribution!

PR #950, opened 2016-09-10
KT-13693 Expand selection for array brackets

shiraji:
#KT-13693 Fixed

https://youtrack.jetbrains.com/issue/KT-13693


yole:
Thanks for your contribution!

PR #951, opened 2016-09-10
#KT-13750 fix

semoro:
https://youtrack.jetbrains.com/issue/KT-13750

Notes: Should it add `@JvmStatic` really even if field matches `* static final long serialVersionUID` or preform check for `java.io.Serializable` is implemented
Now it preforms check.
Also topic on discuss: `@JvmStatic` or `@JvmField`

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/Converter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/Converter.kt
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.j2k
 import com.intellij.openapi.project.Project
 import com.intellij.psi.*
 import com.intellij.psi.CommonClassNames.*
+import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.psi.util.InheritanceUtil
 import com.intellij.psi.util.PsiMethodUtil
 import com.intellij.psi.util.PsiTreeUtil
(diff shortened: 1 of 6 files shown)


j2k/src/org/jetbrains/kotlin/j2k/Converter.kt
```diff
@@ -424,6 +425,24 @@ class Converter private constructor(
         }
     }
 
+
+    private fun specialAnnotationPropertyCases(field: PsiField): Annotations {
+        val javaSerializableInterface = PsiType.getTypeByName("java.io.Serializable", project, GlobalSearchScope.allScope(project)).resolve()
```
yole: It's a minor issue, but the correct way is to use `field.getResolveScope()` instead of `allScope` here. Otherwise you may get incorrect results if the project uses multiple JDKs in different modules.

yole: And also `JavaPsiFacade.findClass()` instead of `PsiType.geTypeByName()`. And `CommonClassNames.JAVA_IO_SERIALIZABLE` instead of a hard-coded constant.

yole:
Looks good except for some minor notes regarding correct API usage. Please fix, and I'll merge and backport to 1.0.5. Thanks!

yole:
Thanks for your contribution!

PR #952, opened 2016-09-12
KT-13552 fixed; bug in converting compound case statements with default statement to "else" statement in "when" statement  

zeesh49:
Fixes [KT-13552](https://youtrack.jetbrains.com/issue/KT-13552) bug in SwitchConverter.kt file and added test case.

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/SwitchConverter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/SwitchConverter.kt
@@ -39,7 +39,10 @@ class SwitchConverter(private val codeConverter: CodeConverter) {
                 continue
             }
             pendingSelectors.add(codeConverter.convertStatement(case.label) as WhenEntrySelector)
-            if (case.statements.isNotEmpty()) {
+            if (!case.label.isDefaultCase && case.statements.isEmpty() && cases[i + 1]?.label!!.isDefaultCase){
+                result.add(WhenEntry(pendingSelectors, convertCaseStatementsToBody(cases, i + 1)).assignNoPrototype())
+                pendingSelectors = ArrayList()
+            }else if (case.statements.isNotEmpty()) {
                 result.add(WhenEntry(pendingSelectors, convertCaseStatementsToBody(cases, i)).assignNoPrototype())
                 pendingSelectors = ArrayList()
             }
--- a/j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
+++ b/j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
@@ -1,0 +1,21 @@
+//method
+void foo() {
+	switch (status) {
+	    case "init":
+	    case "dial":
+	    case "transmit":
+		return Color.BLACK;
+
+	    case "ok":
+		return 0xFF006600;
+
+	    case "cancel":
+		return 0xFF666666;
+
+	    case "fail":
+	    case "busy":
+	    case "error":
+	    default:
+		return 0xFF660000;
+	}
+}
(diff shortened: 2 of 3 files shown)


j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
```diff
@@ -0,0 +1,21 @@
+//method
+void foo() {
```
yole: Please try to ensure that testdata files are valid Java code. This file does not compile ('status' is undefined, the types do not match, etc.)

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
```diff
@@ -0,0 +1,21 @@
+//method
```
yole: You need to include the generated test class into your commit.

j2k/src/org/jetbrains/kotlin/j2k/SwitchConverter.kt
```diff
@@ -39,7 +39,10 @@ class SwitchConverter(private val codeConverter: CodeConverter) {
                 continue
             }
             pendingSelectors.add(codeConverter.convertStatement(case.label) as WhenEntrySelector)
-            if (case.statements.isNotEmpty()) {
+            if (!case.label.isDefaultCase && case.statements.isEmpty() && cases[i + 1]?.label!!.isDefaultCase){
```
yole: You are accessing `cases[i+1]` without checking that `i+1` is a valid index in the `cases` collection.

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
```diff
@@ -0,0 +1,21 @@
...
+
+	    case "ok":
+		return 0xFF006600;
+
+	    case "cancel":
+		return 0xFF666666;
+
+	    case "fail":
+	    case "busy":
+	    case "error":
```
yole: Please add a test for the case when 'default' is in the middle of the list of other cases (`case "error": default: case "busy:") and make sure that this test passes.

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
zeesh49: should I create a complete java file with main method etc? like  kotlin/j2k/testData/fileOrElement/switch/comlicatedFallDown.java?
I did not do so because there are other .java files in the with similar structure to the one I created, because I followed that structure, like:
 kotlin/j2k/testData/fileOrElement/switch/caseWithBlock.java

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
zeesh49: ok, just to confirm, below file right?
 kotlin/j2k/tests/org/jetbrains/kotlin/j2k/JavaToKotlinConverterSingleFileTestGenerated.java

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
yole: You don't need a `main` method but you need a file that can be highlighted without errors in the editor.

j2k/testData/fileOrElement/switch/caseWithDefaultblock.java
yole: Yes

goodwinnk:
@zeesh49 It looks it the issue was closed a long ago, so I'm closing this pull request. Thank you for you time and work.

PR #954, opened 2016-09-19
#KT-13931 fixed

qwwdfsad:
More compact (14 vs 19 bytecodes) contains method for IntRange.
Tested versus "All Jvm Backend Tests" run configuration.
No additional tests were added, there is enough coverage for primitive range 'in' (at least 5 new tests fail when implementation of genInIntRange is incorrect).

Verified that new method is almost the same as classic javac version (except additional label, which is not necessary), now both versions are recognized by decompiler as ` 239 <= x && x <= 250`:
```
  public final static intCmp(I)Z
   L0
    LINENUMBER 7 L0
    SIPUSH 239
    ILOAD 0
    IF_ICMPGT L1
    ILOAD 0
    SIPUSH 250
    IF_ICMPGT L1
   L2
    ICONST_1
    GOTO L3
   L1
    ICONST_0
   L3
    IRETURN
   L4
    LOCALVARIABLE x I L0 L4 0
    MAXSTACK = 2
    MAXLOCALS = 1

  public final static inInt(I)Z
   L0
    LINENUMBER 11 L0
    SIPUSH 239
    ILOAD 0
    IF_ICMPGT L1
    ILOAD 0
    SIPUSH 250
    IF_ICMPGT L1
    ICONST_1
    GOTO L2
   L1
    ICONST_0
   L2
    IRETURN
   L3
    LOCALVARIABLE x I L0 L3 0
    MAXSTACK = 2
    MAXLOCALS = 1
}
```

JMH benchmark doesn't show any difference on my laptop between old and new generated methods.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
@@ -3470,38 +3470,35 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
     }
 
     /*
-     * Translates x in a..b (for int and char ranges only)
-     * to a <= x && x >= b same way as javac 1.8.0_91 does.
+     * Translates x in a..b (for int and char ranges only) to a <= x && x <= b
      */
     private void genInIntRange(StackValue leftValue, KtBinaryExpression rangeExpression) {
+        int localVarIndex = myFrameMap.enterTemp(Type.INT_TYPE);
+
         // Load left bound
         gen(rangeExpression.getLeft(), Type.INT_TYPE);
-        // Load argument
+        // Load x into local variable to avoid StackValue#put side-effects
         leftValue.put(Type.INT_TYPE, v);
+        v.store(localVarIndex, Type.INT_TYPE);
+        v.load(localVarIndex, Type.INT_TYPE);
 
-        // If (left > arg) goto L1 (return)
+        // If (x < left) goto L1
         Label l1 = new Label();
         v.ificmpgt(l1);
 
-        // Load argument again
-        leftValue.put(Type.INT_TYPE, v);
-        // Load right bound
+        // If (x > right) goto L1
+        v.load(localVarIndex, Type.INT_TYPE);
         gen(rangeExpression.getRight(), Type.INT_TYPE);
-        // If (x > right) goto L1 (return)
         v.ificmpgt(l1);
 
         Label l2 = new Label();
-        // Push 'true'
         v.iconst(1);
-        // Goto return
         v.goTo(l2);
 
-        // L1: push 'false'
         v.mark(l1);
         v.iconst(0);
-
-        // L2: implicit ireturn
         v.mark(l2);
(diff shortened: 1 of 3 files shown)


dnpetrov:
Please, provide some tests

Simple functional tests for JVM code generation are added as files under compiler/testData/codegen/box. "Box" test should define `fun box(): String` that should return "OK" if everything is correct (or return any other value / throw an exception otherwise).

After that, use "Generate Tests" run configuration in Kotlin project.

For example, the following test fails after your changes:
```
// WITH_RUNTIME

fun box(): String {
    var x = 0
    assert(x++ in 1 .. 1)
    return "OK"
}
```

qwwdfsad:
I will add tests, thank you for spotting the problem and for explanation how to test code generation!

I've found the root cause: it's not always possible to just load value from StackValue twice as it can have side effects~~, so maybe its worth trying to have two paths for bytecode generation~~.
Sadly, for expression `x in 1..2` x is StackValue$Expression (REFERENCE_EXPRESSION), which always has `canHaveSideEffects=true`. Moreover, even for `1 in 1..2` literal one is StackValue$Expression (INTEGER_CONSTANT).

Anyway, I've end up with implementation of contains which is still as compact as `x >= 239 && x <= 250` (and recognized by decompiler as java version) with respect to side effects: I will finish writing tests and verifying that everything is fine and contribute it soon.

Question out of scope of this work:
Can such expressions (reference/constants wrapped in StackValue$Expression) actually have side effects or it's safe to check for it in StackValue$Expression constructor when deciding if it can have side effects? Should it be supported in Expression then?

dnpetrov:
> Can such expressions (reference/constants wrapped in StackValue$Expression) actually have side effects or it's safe to check for it in StackValue$Expression constructor when deciding if it can have side effects? Should it be supported in StackValue$Expression then?

Yes, but, frankly speaking, it is legacy code.
Our current plans for JVM back-end is to build a new version on top of the IR-based infrastructure. This resolves most of our current architectural issues: very limited code reuse between back-ends, poor separation of concerns leading to leaking abstractions (such as StackValues with possible side effects), and so on. We want to develop some more back-ends for Kotlin (native comes first, Jack and Web Assembly are close candidates for the next), and the existing framework is not well-suited for this.
Both IR and new JVM BE are currently in active development and are about to appear in 1.2 and, most likely, replace existing JVM BE in 1.3.

qwwdfsad:
Thank you for detailed answer!

To sum up:
I've fixed issue with side-effects and also added proper support of `!in` operator.
Now both `in` and `!in` have smaller byte-code size and are consistent with javac version:
`x in 239..250` and `x !in 239..250` are translated to `239 <= x && x <= 250` and `239 > x || x > 250` respectively instead of 
```
 boolean result = true;
    if (x < 230) {
      result = false;
    }
    boolean result2 = true;
    if (x > 250) {
      result2 = false;
    }
    return (result & result2);
``` 
and xoring results with `true`.

I've also written JMH benchmark: new versions have the same score as old ones for both C2 and C1 (to somehow 'emulate' android) compilers.

dnpetrov:
Pushed to master. Thanks!

PR #955, opened 2016-09-20
KT-13635: implement quick fix to convert a too long char literal to a string

gitreelike:
Details:
- the resulting string will be a normal string i.e. not a raw string
- any double quotes will be converted to an escaped double quote: 'x"x' -> "x\"x"
- ~~any back slash will be converted to an escaped back slash: 'x\x' -> "x\\\\x"~~ _(edit: not doing this see comments and commits below)_
- have a look at KT-13635

This is my first pull request - hopefully I did nothing dumb. Just tell me if anything is wrong.


idea/src/org/jetbrains/kotlin/idea/quickfix/TooLongCharLiteralToStringFix.kt
```diff
@@ -0,0 +1,40 @@
...
+        val newStringContent = text
+                .slice(1..text.length - 2)
+                .replace("\\", "\\\\")
+                .replace("\"", "\\\"")
+        val newElement = KtPsiFactory(element).createStringTemplate(newStringContent)
+
+        element.replace(newElement)
+    }
+
+    override fun getFamilyName(): String = "Strings"
```
mglukhikh: Usually `= text` is in use here.

idea/src/org/jetbrains/kotlin/idea/quickfix/TooLongCharLiteralToStringFix.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val text = element.text
+        if (!(text.startsWith("'") && text.endsWith("'") && text.length >= 2)) {
+            return
+        }
+
+        val newStringContent = text
+                .slice(1..text.length - 2)
+                .replace("\\", "\\\\")
```
mglukhikh: For me it's suspicious. We can have a literal like `'Bye\n'` and expect to get `"Bye\n"` as result, so I'd not touch these back slashes.

idea/src/org/jetbrains/kotlin/idea/quickfix/TooLongCharLiteralToStringFix.kt
```diff
@@ -0,0 +1,40 @@
...
+                .replace("\"", "\\\"")
```
mglukhikh: Changing double quote to escaped double quote is logical, I think it should be left as is.

mglukhikh:
Mostly everything is Ok. I added some notes for your changes, please fix them, and I'll integrate the result into master branch.

gitreelike:
Thanks, will do that tonight.

gitreelike:
I had been thinking about the back slash and the dollar and wasn't quite sure how to handle them. Thank you for your input on this. I will change the back slash behaviour and document both with a test. I will also address the familyName.

mglukhikh:
Thanks, I added everything to the master and 1.0.5 branches

PR #956, opened 2016-09-21
KT-13146 J2K recursion while converting self-referenced anonymous functions

semoro:
https://youtrack.jetbrains.com/issue/KT-13146

This fix is only small workaround of a big problem

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
@@ -538,6 +538,13 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
         val qualifier = expression.qualifierExpression
 
         var identifier = Identifier.withNoPrototype(referenceName, isNullable)
+
+
+        if (expression.getContainingClass()?.getParentOfType<PsiField>(false).let { it != null && it == expression.qualifierExpression?.reference?.resolve() }) {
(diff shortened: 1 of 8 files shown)


j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -538,6 +538,13 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
         val qualifier = expression.qualifierExpression
 
         var identifier = Identifier.withNoPrototype(referenceName, isNullable)
+
+
+        if (expression.getContainingClass()?.getParentOfType<PsiField>(false).let { it != null && it == expression.qualifierExpression?.reference?.resolve() }) {
```
yole: Why are you checking for `PsiField` specifically? Won't the same problem happen if the anonymous class is used as a local variable initializer?

yole: `x.let { it != null && ... }` is much better expressed as `x?.let { ... }`

semoro: It is used in if check
Which better
`if(x.let{ it != null && ........................long_check.....................................})`
or
`if(x?.let{  ........................long_check..................................... } ?: false)`

semoro: I should use `PsiVariable` to cover fields and local variables?

yole:
Merged into master and 1.0.5. Thanks for your contribution!

PR #957, opened 2016-09-22
KT-11990 J2K `native` to `external` conversion

semoro:
https://youtrack.jetbrains.com/issue/KT-11990

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/ClassBodyConverter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/ClassBodyConverter.kt
@@ -152,14 +152,13 @@ class ClassBodyConverter(private val psiClass: PsiClass,
     ): Member? {
         when (member) {
             is PsiMethod -> {
-                if (!member.hasModifierProperty(PsiModifier.NATIVE))
-                    memberToPropertyInfo[member]?.let { propertyInfo ->
-                        if (propertyInfo.field != null) return null // just drop the method, property will be generated when converting the field
-                        return if (member == propertyInfo.getMethod || propertyInfo.getMethod == null)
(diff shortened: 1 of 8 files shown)


j2k/testData/fileOrElement/function/nativeMethods.java
```diff
@@ -0,0 +1,6 @@
+public class Foo {
+    private native final void nativeMethod()
```
yole: Please avoid irrelevant syntax errors in testdata. Java does require semicolons :)

j2k/src/org/jetbrains/kotlin/j2k/ClassBodyConverter.kt
```diff
@@ -152,13 +152,14 @@ class ClassBodyConverter(private val psiClass: PsiClass,
...
         when (member) {
             is PsiMethod -> {
-                memberToPropertyInfo[member]?.let { propertyInfo ->
-                    if (propertyInfo.field != null) return null // just drop the method, property will be generated when converting the field
-                    return if (member == propertyInfo.getMethod || propertyInfo.getMethod == null)
-                        converter.convertProperty(propertyInfo, classKind)
-                    else
-                        null // drop the method, property will be generated when converting the field or the getter
-                }
+                if (!member.hasModifierProperty(PsiModifier.NATIVE))
```
yole: This doesn't seem to be a correct check. This is valid Kotlin:

``` kotlin
val foo: Int = 0
    external get
```

semoro:
Why we must initialize val with external get?

yole:
I think it's a bug in the validator; there is no other reason

yole:
Thanks for your contribution!

PR #958, opened 2016-09-22
KT-13744 Add "Copy concatention text to clipboard" intension

shiraji:
#KT-13744 Fixed

https://youtrack.jetbrains.com/issue/KT-13744


idea/src/org/jetbrains/kotlin/idea/intentions/copyConcatenatedStringToClipboard/ConcatenatedStringGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+class ConcatenatedStringGenerator {
+    fun create(element: KtBinaryExpression): String {
+        val binaryExpression = getBinaryParentExpression(element)
+        val stringBuilder = StringBuilder()
+        create(binaryExpression, stringBuilder)
+        return stringBuilder.toString()
+    }
+
+    private fun getBinaryParentExpression(element: KtBinaryExpression): KtBinaryExpression {
+        var binaryExpression = element.getStrictParentOfType<KtBinaryExpression>() ?: return element
```
mglukhikh: There is `getTopmostParentOfTypes`, I think it should work here

idea/src/org/jetbrains/kotlin/idea/intentions/copyConcatenatedStringToClipboard/ConcatenatedStringGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+        while (binaryExpression != null) {
+            binaryExpression = binaryExpression.getStrictParentOfType<KtBinaryExpression>() ?: return element
+        }
+        return element
+    }
+
+    private fun create(element: KtBinaryExpression, sb: StringBuilder) {
```
mglukhikh: I'd declare it as something like `private fun KtBinaryExpression.appendTo(sb: StringBuilder)`. And the same for the next function.

mglukhikh:
Mostly everything is Ok, so thank you. I added some notes into your code. Please do the requested changes, and I integrate them into master.

shiraji:
Do you think `idea/testData/stringgenerator` is good name of test data
directory? I could not come up better name for this new unit test.

2016年9月22日(木) 20:03 Mikhail Glukhikh <notifications@github.com>:

mglukhikh:
May be `concatenatedStringGenerator` would be a bit better.

shiraji:
OK, I will fix these asap.

2016年9月22日(木) 20:32 Mikhail Glukhikh <notifications@github.com>:

idea/src/org/jetbrains/kotlin/idea/intentions/copyConcatenatedStringToClipboard/ConcatenatedStringGenerator.kt
shiraji: Ah! That's what I need it. Thanks.

idea/src/org/jetbrains/kotlin/idea/intentions/copyConcatenatedStringToClipboard/ConcatenatedStringGenerator.kt
shiraji: I am still not used to use "extension method". Thanks for pointing this and giving me a chance to learn this!

mglukhikh:
Thanks, I'll squash and merge it.

shiraji:
OK, Thanks. If you don't mind I can squash the commits? (You need to do code review, again, if I did it. That's why you are the one who squash, I guess)

mglukhikh:
It's not necessary, done already. BTW, for some reason your testNewLine does not work on my PC. Now I'm trying to investigate it.

shiraji:
`testNewLine`'s .result file has space after 1. The auto code format trims the space.

mglukhikh:
Yes but I got an exception "Failed to create expression from text". At this moment it's unclear, why. However, this case works in IDE, so I integrated your PR without this test, will investigate the situation a bit later.

PR #959, opened 2016-09-24
KT-13945 Add intention to replace Math.max/min with coerceAtLeast/coerceAtMost

shiraji:
#KT-13945 Fixed

https://youtrack.jetbrains.com/issue/KT-13945

### DIFF
--- a/idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
+++ b/idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
+</body>
+</html>
--- a/idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
+++ b/idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
@@ -1,5 +1,0 @@
-<html>
-<body>
-This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
-</body>
-</html>
--- a/idea/resources/inspectionDescriptions/ReplaceMathMinWithCoerceAtMost.html
+++ b/idea/resources/inspectionDescriptions/ReplaceMathMinWithCoerceAtMost.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This intention detects 'Math.min' calls that can be safely replaced with 'coerceAtMost'
+</body>
+</html>
--- a/idea/resources/inspectionDescriptions/ReplaceMathMinWithCoerceAtMost.html
+++ b/idea/resources/inspectionDescriptions/ReplaceMathMinWithCoerceAtMost.html
@@ -1,5 +1,0 @@
-<html>
-<body>
-This intention detects 'Math.min' calls that can be safely replaced with 'coerceAtMost'
-</body>
-</html>
--- a/idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
+++ b/idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
@@ -1,5 +1,5 @@
 <html>
 <body>
-    This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
+    This intention replaces 'Math.max' calls with 'coerceAtLeast'
 </body>
 </html>
--- a/idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
+++ b/idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+    This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
+</body>
+</html>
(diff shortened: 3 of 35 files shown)


idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
```
mglukhikh: inspection

idea/resources/inspectionDescriptions/ReplaceMathMinWithCoerceAtMost.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention detects 'Math.min' calls that can be safely replaced with 'coerceAtMost'
```
mglukhikh: inspection

idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+    This intention detects 'Math.max' calls that can be safely replaced with 'coerceAtLeast'
```
mglukhikh: This intention replaces 'Math.max' calls with safe 'coerceAtLeast' ... or something like. Inspection detects, but inspection does some change.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMaxWithCoerceAtLeastIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+        val newExpression = KtPsiFactory(element).createExpression("${valueArguments[0].text}.coerceAtLeast(${valueArguments[1].text})")
+        target.replaced(newExpression)
+    }
+
+    override fun isApplicableTo(element: KtCallExpression) = isMaxMethod(element)
+
+    private fun isMaxMethod(element: KtCallExpression) =
+            element.calleeExpression?.text == "max" && element.valueArguments.size == 2 && element.isMethodCall("java.lang.Math.max")
+
+    private fun KtCallExpression.isMethodCall(fqMethodName: String): Boolean {
```
cypressious: You have the same method in `ReplaceMathMinWithCoerceAtMostIntention`. It would make more sense to pull it out in a Utils class.

mglukhikh: Already exists in ReplaceSingleLineLetIntention. Should be extracted to some Utils file, e.g. intentions/Utils.kt.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMaxWithCoerceAtLeastIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameUnsafe
+
+class ReplaceMathMaxWithCoerceAtLeastInspection : IntentionBasedInspection<KtCallExpression>(ReplaceMathMaxWithCoerceAtLeastIntention::class)
+
+class ReplaceMathMaxWithCoerceAtLeastIntention() : SelfTargetingOffsetIndependentIntention<KtCallExpression>(
```
mglukhikh: I think abstract class should be extracted from this intention and the next one. Too many similar code inside

idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
mglukhikh: I think we should implement only intentions here and not inspections. `Math.max(1, 2)` for many cases is as good as `1.coerceAtLeast(2)`.

idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
mglukhikh: Moreover, the word 'safe' is not needed here. Looks like replacement is possible in any case.

idea/resources/inspectionDescriptions/ReplaceMathMaxWithCoerceAtLeast.html
cypressious: Agreed.

shiraji: OK, I will remove both inspections

idea/resources/intentionDescriptions/ReplaceMathMaxWithCoerceAtLeastIntention/description.html
shiraji: I may misunderstand the role of inspection/intention...Thanks.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMaxWithCoerceAtLeastIntention.kt
shiraji: Thanks! I will refactor them.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMaxWithCoerceAtLeastIntention.kt
shiraji: Could you suggest me the name of abstract class? `ReplaceMathMethodsWithCoerceAt`? uh...doesn't make sense...

mglukhikh:
I think it's OK with changes, so I'll integrate your PR in master, thanks.

shiraji:
Thanks both of you guys, @mglukhikh and @cypressious

PR #960, opened 2016-09-27
KT-13974 Fix 'in' operator intention for Strings

takahirom:
Thank you for awesome programming language 👍 
I would like to contribute.

KT-13974 Fixed
https://youtrack.jetbrains.com/issue/KT-13974

Currently 'in' operator intention checks parameter by `SingleValueParameter` . But `CharSequence.contains` has two parameters. (First parameter has no default value. And Second parameter has default value)
https://github.com/JetBrains/kotlin/blob/0d7819e011031bac447923247e73544c1dfb14e6/libraries/stdlib/src/kotlin/text/Strings.kt#L1000

So, I created `FirstNoDefaultAndDefaultParameters` and I use it.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/conventionNameCalls/ReplaceContainsIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/conventionNameCalls/ReplaceContainsIntention.kt
@@ -48,7 +48,8 @@ class ReplaceContainsIntention : SelfTargetingRangeIntention<KtDotQualifiedExpre
         if (!element.isReceiverExpressionWithValue()) return null
 
         val functionDescriptor = getFunctionDescriptor(element) ?: return null
-        if (!functionDescriptor.isOperator || !OperatorChecks.check(functionDescriptor).isSuccess) return null
+
+        if (!functionDescriptor.isOperator) return null
 
         return element.callExpression!!.calleeExpression!!.textRange
     }
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/conventionNameCalls/ReplaceContainsIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/conventionNameCalls/ReplaceContainsIntention.kt
@@ -49,7 +49,7 @@ class ReplaceContainsIntention : SelfTargetingRangeIntention<KtDotQualifiedExpre
 
         val functionDescriptor = getFunctionDescriptor(element) ?: return null
 
-        if (!functionDescriptor.isOperator) return null
+        if (!functionDescriptor.isOperator && !OperatorChecks.check(functionDescriptor).isSuccess) return null
 
         return element.callExpression!!.calleeExpression!!.textRange
     }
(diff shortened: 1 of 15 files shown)


yole:
The current rules for checking the 'in' operator are implemented intentionally, and also intentionally suppressed for `CharSequence.contains`. Instead of changing the language rules, you should change the logic in ReplaceContainsIntention.kt:51 so that the intention is always enabled when the target function is an operator.

takahirom:
@yole Thanks! I will change.

takahirom:
@yole 
Done.

idea/src/org/jetbrains/kotlin/idea/intentions/conventionNameCalls/ReplaceContainsIntention.kt
```diff
@@ -48,7 +48,8 @@ class ReplaceContainsIntention : SelfTargetingRangeIntention<KtDotQualifiedExpre
         if (!element.isReceiverExpressionWithValue()) return null
 
         val functionDescriptor = getFunctionDescriptor(element) ?: return null
-        if (!functionDescriptor.isOperator || !OperatorChecks.check(functionDescriptor).isSuccess) return null
+
+        if (!functionDescriptor.isOperator) return null
```
yole: This will break the use of the intention with Java methods, which don't use an explicit `operator` modifier but still can be used as operators. The correct fix is to replace || with && in the original check.

takahirom: I fixed it. But I have a little problem that test failed.
https://github.com/takahirom/kotlin/blob/145ee2c92f909706272922e2f3eab03d64d853ec/idea/testData/intentions/conventionNameCalls/replaceContains/withoutOperatorModifier.kt

Result:

```
java.lang.RuntimeException: java.lang.AssertionError: isAvailable() for class org.jetbrains.kotlin.idea.intentions.conventionNameCalls.ReplaceContainsIntention should return false
```

---

So I delete `IS_APPLICABLE: false` comment. And I created `withoutOperatorModifier.kt.after`.

Result:

```
java.lang.RuntimeException: org.junit.ComparisonFailure: All actual errors should be mentioned in test data with // ERROR: directive. But no unnecessary errors should be me mentioned 
Expected :<empty>
Actual   :'operator' modifier is required on 'contains' in 'test.Test'
```

---

So I added  `// ERROR: 'operator' modifier is required on 'contains' in 'test.Test'` to `withoutOperatorModifier.kt`

Result:

```
java.lang.RuntimeException: org.junit.ComparisonFailure: All actual errors should be mentioned in test data with // ERROR: directive. But no unnecessary errors should be me mentioned 
Expected :'operator' modifier is required on 'contains' in 'test.Test'
Actual   :<empty>
```

---

Probably Because of  `'operator' modifier is required on 'contains' in 'test.Test'` Error is  transformed syntax error by intention.

Do you have any idea to solve this?

takahirom: I'm thinking about removing test.

takahirom: I removed this test.
https://github.com/takahirom/kotlin/blob/145ee2c92f909706272922e2f3eab03d64d853ec/idea/testData/intentions/conventionNameCalls/replaceContains/withoutOperatorModifier.kt
Please check .

yole:
Sorry for the delay. I've merged your changes and added some extra logic that allowed to preserve the withoutOperatorModifier test (we now make a distinction between Java and Kotlin methods).

Thanks for your contribution!

PR #961, opened 2016-10-01
KT-13937 Inspection/Intention replace size check with 'isNotEmpty' and 'isEmpty'

shiraji:
#KT-13937 Fixed

https://youtrack.jetbrains.com/issue/KT-13937


mglukhikh:
I've added it into master and 1.0.5, thanks.

PR #962, opened 2016-10-02
KT-7931 fixed

qwwdfsad:
Optimized code generation for iteration over string, results are added to initial benchmark (https://gist.github.com/qwwdfsad/69616e8a42a74def31e48d774c622379): new version is as fast as java version.

I've decided not to generate intrinsic for iteration over CharSequence because end user may not expect such code transformation for his custom string.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
@@ -966,15 +966,9 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
             v.iconst(0);
             v.store(indexVar, Type.INT_TYPE);
 
-            StackValue value = gen(forExpression.getLoopRange());
-            if (value instanceof StackValue.Local && value.type.equals(JAVA_STRING_TYPE)) {
-                stringVar = ((StackValue.Local) value).index;
-            }
-            else {
-                stringVar = createLoopTempVariable(JAVA_STRING_TYPE);
-                value.put(JAVA_STRING_TYPE, v);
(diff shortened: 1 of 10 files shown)


dnpetrov:
This test fails:

```
// WITH_RUNTIME

fun useBoxedChar(ch: Char?) = ch!!

fun box(): String {
    var s = ""
    for (ch: Char? in "OK") {
        s += useBoxedChar(ch)
    }

    return s
}
```

qwwdfsad:
Eh, same problem as in KT-12582, shame to miss it :(
Don't you mind if I extract logic about `loopParameter()` from `AbstractForInProgressionOrRangeLoopGenerator` to `AbstractForLoopGenerator`? I will not try to refactor `ForInArrayLoopGenerator`  to use it, but at least will reuse it in `StringForLoopGenerator`

dnpetrov:
Yes, that would do the trick.

qwwdfsad:
Fixed

dnpetrov:
This test fails:
```
// WITH_RUNTIME

fun box(): String {
    var clist = listOf('1', '2', '3', '4')
    var res1 = ""
    for (ch in clist) {
        res1 += ch
        clist = listOf()
    }

    var s = "1234"
    var res2 = ""
    for (ch in s) {
        res2 += ch
        s = ""
    }

    return if (res1 == res2) "OK" else "'$res1' != '$res2'"
}
```

Exception:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 1
	at java.lang.String.charAt(String.java:686)
	at ForInStringLocalVarKt.box(forInStringLocalVar.kt:13)
```

qwwdfsad:
Fixed. This problem also uncovered KT-14589, but it will be fixed separately

PR #963, opened 2016-10-05
KT-10871, KT-12625 Add inspection/intention for redundant calls of conversion methods

shiraji:
#KT-10871 Fixed
 #KT-12625 Fixed

https://youtrack.jetbrains.com/issue/KT-10871
https://youtrack.jetbrains.com/issue/KT-12625

One thing I could do this is test for `MutableMap` and `Map`. The following code does not let me pass the test whereas it works well on IDEA

```kotlin
// WITH_RUNTIME
val foo = mutableMapOf(1, 2, 3).toMutableMap()
```

```kotlin
// WITH_RUNTIME
val foo = mapOf(1, 2, 3).toMap()
```

It gives me `Unresolved reference` error. It seems the test case cannot find kotlin.collections.Mapskt file


idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.jetbrains.kotlin.idea.inspections.IntentionBasedInspection
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
+import org.jetbrains.kotlin.psi.KtConstantExpression
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.calls.callUtil.getType
+import java.util.*
+
+class RemoveRedundantCallsOfConversionMethodsInspection : IntentionBasedInspection<KtDotQualifiedExpression>(RemoveRedundantCallsOfConversionMethodsIntention::class)
```
mglukhikh: I think these redundant calls should be highlighted as unused symbols.

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
```diff
@@ -0,0 +1,69 @@
...
+                                       "toInt()" to Int::class.qualifiedName,
+                                       "toChar()" to Char::class.qualifiedName,
+                                       "toShort()" to Short::class.qualifiedName,
+                                       "toByte()" to Byte::class.qualifiedName)
+
+
+    override fun applyTo(element: KtDotQualifiedExpression, editor: Editor?) {
+        element.replaced(element.receiverExpression)
+
+        mutableMapOf(1 to 1).toMutableMap()
```
mglukhikh: What is it? Looks like it's not needed here :)

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    override fun isApplicableTo(element: KtDotQualifiedExpression): Boolean {
```
mglukhikh: Looks like we can apply this intention for all qualified expressions, not only unsafe ones.

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
mglukhikh: Also I think that only selectors should be highlighted. And probably, even the intention should be active on selector only (imagine that receiver is some long expression).

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
```diff
@@ -0,0 +1,69 @@
...
+        val selectorExpressionText = element.selectorExpression?.text
+        val qualifiedName = targetClassMap[selectorExpressionText] ?: return false
+        val receiverExpression = element.receiverExpression
+        return when (receiverExpression) {
+            is KtStringTemplateExpression -> String::class.qualifiedName
+            is KtConstantExpression -> receiverExpression.getType(receiverExpression.analyze())?.getJetTypeFqName(false)
+            else -> receiverExpression.getResolvedCall(receiverExpression.analyze())?.candidateDescriptor?.returnType?.getJetTypeFqName(false)
+        } == qualifiedName
```
mglukhikh: I'm not sure here what will be happened if somebody defines custom `toType()` function. Probably it's not too important, because it's expected that `toType()` has result of `Type` anyway.

mglukhikh:
I think also that this case with maps can be ignored in tests.

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
shiraji: Ahhh!!! I accidentally paste the test code...I will remove this.

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
shiraji: Could you explain more detail about unsafe ones?

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
shiraji: This intention only detects `targetClassMap`'s Type. I don't think the custom `toType()` is detected by this intention. I don't think this intention should detects custom ones. Do you think this intention find custom one as well?

shiraji:
Could you explain why I could not create test cases for `Maps`?

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
mglukhikh: I think the best way here is to check it.

mglukhikh:
About `Maps` I do not know details, but probably they are not included into runtime library which is in use inside tests. It's fixable but it's not very important.

shiraji:
Got it. Thanks for explanation!

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
shiraji: Let me clarify what you suggest here. 

You say if there is following code

```kotlin
bar().toCustomType() 
```

If bar() return type name is `package.name.CustomType` and `toCustomType()`'s return type name is also `package.name.CustomType`, this inspection detect as redundant.

I made `targetClassMap` because I want this inspection to work for only `toType()` method we know about. 

Yes it actually is much easier to implement that but does it lead false positive?

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
shiraji: Make sense! Thanks.

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt
mglukhikh: Ok, probably you're right.

shiraji: Ok then, i will take care of KtSafeQualifiedExpression

shiraji:
Implemented it.

mglukhikh:
Thank you for your efforts. I'll integrate this change into master and (probably) 1.0.5 on Monday

PR #965, opened 2016-10-08
Fix Boolean.TYPE does not compile

takahirom:
#KT-13550 Fixed
https://youtrack.jetbrains.com/issue/KT-13550

Thank you for your advice in https://youtrack.jetbrains.com/issue/KT-13550 .
I was suggested the javaPrimitiveType through the KT-13550.
However, maybe if you are not using the Java class (for example java.lang.Boolean), javaPrimitiveType and java is the same.
Not used Java classes as a result of J2K in the current implementation.
So, I changed to use .java.


takahirom:
I reconsider. If you might want to teach that it is Primitive types of Java to the user, This changes is not good.

PR #966, opened 2016-10-09
KT-5435 Surround with try/catch should generate more Kotlin-style code

shiraji:
#KT-5435

https://youtrack.jetbrains.com/issue/KT-5435


yole:
If you're converting an existing Java file to Kotlin, please do it as a sequence of the following commits:
* Rename the file from .java to .kt and commit without any changes (this ensures the history is preserved)
* Rename the file back, run J2K and make the minimum number of changes required for the code to compile.
* Clean up the code to convert it into idiomatic Kotlin.

After that, make the changes that you need to fix the issue as a separate commit. Right now I can't review your changes because I don't see what you changed compared to the original Java code. Also it's not OK that you left the old Java class around.

shiraji:
@yole 

Do you think other surrounding functions can be kotlin-style? If so, I will do what you suggested. I wasn't too sure that they can be kotlin-style or not.

At this point, `if`, `if/else`, `{ }`, and `try`'s, are the one that can use the new helper. If the following code is the target codes

```kotlin
        val foo = bar()
        println(foo)
```

then, new helper will generate

`if`

```kotlin
        val foo = if (<caret>) {
            bar()
        }
        println(foo)
```

`if/else`

```kotlin
        val foo = if (<caret>) {
            bar()
        } else {
        }
        println(foo)
```

`{ }`

```kotlin
        val foo = run {
            bar()
        }
        println(foo)
```

Is this sounds good for you?

yole:
Yes, let's try to apply this to the other statements as well.

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/order/valAndClass.kt.after
```diff
@@ -3,9 +3,9 @@ fun foo() {
         fun test() {}
     }
 
-    val d: A
+    val d =
     if (<caret>) {
-        d = A()
+        A()
```
shiraji: I don't think this new line is the problem with this pull request.

If I have following case

```kotlin
fun foo() {
    <selection>class A {
        fun test() {}
    }
    
    
    
    val d: A = A()</selection>

    d.test()
    A()
}
```

For current kotlin plugin converts them into 

```kotlin
class A {
        fun test() {
        }
    }

    val d: A


    if () {
        d = A()
    }
```

shiraji:
Implemented it. Could you review this PR, again?

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
```diff
@@ -16,29 +16,26 @@
...
+import org.jetbrains.kotlin.idea.util.IdeDescriptorRenderers
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 import org.jetbrains.kotlin.types.KotlinType
-
-import java.util.ArrayList
+import java.util.*
 
 object MoveDeclarationsOutHelper {
```
yole: This `object` serves no purpose; the contents of the file should simply be top-level functions

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
```diff
@@ -83,6 +83,35 @@ object MoveDeclarationsOutHelper {
         return PsiUtilCore.toPsiElementArray(resultStatements)
     }
 
+    private fun addSymbolAfterDeclaration(container: PsiElement, declaration: KtProperty, isLastStatemnt: Boolean) {
```
yole: typo: statemnt -> statement

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
```diff
@@ -83,6 +83,35 @@ object MoveDeclarationsOutHelper {
...
+        if (isLastStatemnt) {
+            container.addAfter(KtPsiFactory(declaration).createEQ(), declaration)
+        }
+        else {
+            container.addAfter(KtPsiFactory(declaration).createNewLine(), declaration)
```
yole: It should never be necessary to create newlines from text. All newlines should be inserted by the formatter.

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
```diff
@@ -52,16 +52,16 @@ object MoveDeclarationsOutHelper {
             val scope = LocalSearchScope(container)
             val lastStatementOffset = statements[statements.size - 1].textRange.endOffset
 
-            for (statement in statements) {
+            statements.forEachIndexed { i, statement ->
                 if (needToDeclareOut(statement, lastStatementOffset, scope)) {
                     if (statement is KtProperty && statement.initializer != null) {
-                        var declaration = createVariableDeclaration(statement, generateDefaultInitializers)
+                        val isLastStatement = i == statements.size - 1
+                        var declaration = createVariableAssignment(statement, generateDefaultInitializers, isLastStatement)
```
yole: I don't think this code is structured well enough. You have three functions that do three sets of entirely different things based on a value of a boolean parameter. I think it's clearer to pull the check out to the top level and to avoid introducing the functions altogether.

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/val/fullQualifiedType.kt.after
```diff
@@ -1,7 +1,6 @@
 fun foo() {
-    val a: kotlin.test.Asserter?
-    if (<caret>) {
-        a = null
+    val a = if (<caret>) {
```
yole: It's not OK that you're losing the variable type here.

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/order/valAndClass.kt.after
yole: I do think it's a problem that should be fixed if possible.

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
shiraji: Fixed it

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
shiraji: Fixed it

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/val/fullQualifiedType.kt.after
shiraji: Right. I think if there is type declaration, this intention should keep them.

```kotlin
val a: kotlin.test.Asserter? = if (<caret>) {
    null
}
```

should be the expected result. I will work on that.

idea/src/org/jetbrains/kotlin/idea/codeInsight/surroundWith/MoveDeclarationsOutHelper.kt
shiraji: Removed this line.

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/val/fullQualifiedType.kt.after
shiraji: Implemented it

idea/testData/codeInsight/surroundWith/if/moveDeclarationsOut/order/valAndClass.kt.after
shiraji: I think this is quite hard to fix...it's problem with locating `dummyFirstStatement` and inserting generated statement. I'm not even sure that it is possible to fix this.

shiraji:
Sorry not to mention this but I think this pull request is ready for re-review.

goodwinnk:
Hello! Your changes were merged manually after some heavy refactoring: https://github.com/JetBrains/kotlin/commit/06dad5f45baa960d80fbdc476513d676347bf91e. Please try to do movement refactorings and Java to Kotlin conversion beforehand to make relevant changes more obvious. 

Thank you!

PR #970, opened 2016-10-11
make Kotlin plugin compatible with IDEA 2017.1 builds

chashnikov:
We are creating stable branch for IDEA 2016.3 releases and 'master' branch will have 171.\* build numbers.


yole:
Merged manually into branch for 2016.3.

PR #972, opened 2016-10-13
KT-14329  Do not report inspection "Remove empty class body" for anonymous objects

shiraji:
#KT-14329 Fixed

https://youtrack.jetbrains.com/issue/KT-14329

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveEmptyClassBodyIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveEmptyClassBodyIntention.kt
@@ -21,6 +21,8 @@ import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.openapi.editor.Editor
 import org.jetbrains.kotlin.idea.inspections.IntentionBasedInspection
 import org.jetbrains.kotlin.psi.KtClassBody
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 
 class RemoveEmptyClassBodyInspection : IntentionBasedInspection<KtClassBody>(RemoveEmptyClassBodyIntention::class), CleanupLocalInspectionTool {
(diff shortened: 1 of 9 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/RemoveEmptyClassBodyIntention.kt
```diff
@@ -30,5 +32,10 @@ class RemoveEmptyClassBodyInspection : IntentionBasedInspection<KtClassBody>(Rem
 class RemoveEmptyClassBodyIntention : SelfTargetingOffsetIndependentIntention<KtClassBody>(KtClassBody::class.java, "Remove empty class body") {
     override fun applyTo(element: KtClassBody, editor: Editor?) = element.delete()
 
-    override fun isApplicableTo(element: KtClassBody) = element.text.replace("{", "").replace("}", "").isBlank()
+    override fun isApplicableTo(element: KtClassBody): Boolean {
+        element.getStrictParentOfType<KtObjectDeclaration>()?.let {
```
cypressious: Won't
```
object Foo {
    class Bar {}
}
```
cause a false positive here?

shiraji: Added test cases for following cases

```kotlin
// IS_APPLICABLE: false

object Foo {
    fun test() {
        object {}<caret>
    }
}

```

```kotlin
// IS_APPLICABLE: true

object Foo {
    class Bar {}<caret>
}
```

shiraji: Ah!

```kotlin
object Foo {
    object {}<caret>
}
```

cypressious: Is 
```kotlin
val foo = object {
    class Foo {}
}
```

valid?

yole:
Thanks for your contribution!

PR #973, opened 2016-10-13
Add quickfixes for lateinit-related errors KT-14342

cypressious:
Fixes #KT-14342


mglukhikh:
We already have fixes for both required actions: `ChangeVariableMutabilityFix` and `RemoveNullableFix`, so it's enough just to create factories for these actions. Nevertheless, thank you for your efforts. Added into master & 1.0.5 with changes.

PR #974, opened 2016-10-13
Add intention to convert top level val with object expression to object declaration

cypressious:
https://youtrack.jetbrains.com/issue/KT-14137

### DIFF
--- a/idea/resources/intentionDescriptions/ValToObjectIntention/after.kt.template
+++ b/idea/resources/intentionDescriptions/ValToObjectIntention/after.kt.template
@@ -3,7 +3,7 @@ interface B {
 }
 
 
-<spot>object A : B</spot> {
+<spot>object a : B</spot> {
     override fun c() {
     }
 }
--- a/idea/resources/intentionDescriptions/ValToObjectIntention/after.kt.template
+++ b/idea/resources/intentionDescriptions/ValToObjectIntention/after.kt.template
@@ -1,0 +1,9 @@
+interface B {
+    fun c()
+}
+
+
+<spot>object A : B</spot> {
+    override fun c() {
+    }
+}
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.idea.intentions
+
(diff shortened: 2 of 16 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+        val o = element.initializer as? KtObjectLiteralExpression ?: return
+        val od = o.objectDeclaration
+        val superTypeList = od.getSuperTypeList() ?: return
+        val body = od.getBody() ?: return
+
+        val prefix = modifier?.text?.plus(" ") ?: ""
+        val replacementText = "${prefix}object $name: ${superTypeList.text} ${body.text}"
+        val replaced = element.replaced(KtPsiFactory(element).createDeclarationByPattern<KtObjectDeclaration>(replacementText))
+
+        RenameProcessor(element.project, replaced, name.capitalize(), false, false).doRun()
```
yole: I think that automatically changing the name without user input is not a very good idea. It's better to trigger a rename refactoring where the user inputs the name, initialized with the capitalized version of the old name.

idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+        if (element.isVar) return false
+        if (!element.isTopLevel) return false
+        if (element.initializer !is KtObjectLiteralExpression) return false
+        if (element.getter != null) return false
+
+        return true
+    }
+
+    override fun applyTo(element: KtProperty, editor: Editor?) {
+        val modifier = element.visibilityModifier()
```
yole: You're losing all annotations on the original property. (I think that it would be best not to offer the intention if the property has any, because the semantics of annotations on a `val` and an `object` is significantly different.)

idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+        val name = element.name ?: return
+        val o = element.initializer as? KtObjectLiteralExpression ?: return
+        val od = o.objectDeclaration
+        val superTypeList = od.getSuperTypeList() ?: return
```
yole: I think you should check the supertype list in `isApplicableTo` as well, so that you don't end up with an intention which is enabled and does nothing.

yole:
And one more thing: this will break usages from Java because with a `val` they can refer to the property name directly, and with an `object` they require the `.INSTANCE` qualifier.

cypressious:
What would you suggest regarding the Java problem?

yole:
After the rename is complete, you can search for usages of the new declaration in Java files and update them as needed.

idea/src/org/jetbrains/kotlin/idea/intentions/ValToObjectIntention.kt
cypressious: Can you point me to how I start an inline renaming? All I could find was the one with the dialog.

cypressious:
@yole I've decided to remove the renaming part (maybe until later) and also disable the intention if it has non-Kotlin usages. On the one hand this is a bit lazy, on the other hand it can make sense because the usefulness of the intention comes from the fact that it only simplifies the declaration site but doesn't change the call sites, but this is only true for Kotlin code, not Java.

idea/resources/intentionDescriptions/ValToObjectIntention/after.kt.template
```diff
@@ -0,0 +1,9 @@
+interface B {
+    fun c()
+}
+
+
+<spot>object A : B</spot> {
```
yole: This should be `object a`

yole:

PR #976, opened 2016-10-16
KT-7718

nd:


yole:
The intended formatting for this code is to wrap before the colon, not after it.

goodwinnk:
@yole Wrapping is out of the scope of KT-7718. The fix looks valid to me, I going to test it on buildserver to see are there any accidental test fails in other test groups.

yole:
Sure, just pointing out that the formatting is not optimal in any case

goodwinnk:
Merged manually into 1.0.5 and master branches with slight modification in commit message and test.
Thank you!

PR #977, opened 2016-10-17
KT-14131

nd:


nd:
Noticed some failed tests, will fix them and recreate a PR.

PR #978, opened 2016-10-19
KT-14131

nd:
Judging by examples in documentation, 1 space between label and expression is expected. Other tests (e.g. PrimaryConstructor.after.kt) suggest that spaces between label and comment should be left as they are.


goodwinnk:
The spacing between label and comment is modfied when there're several lines. Please test it the the following code:

`fun test1() {
    loop@    while (true) {

    }
}

fun test2() {
    listOf(1).forEach lit@    {
        if (it == 0) return@lit
        print(it)
    }
}

fun test3() {
    block@   /*comment*/ {

    }
}

fun test3() {
    block@


    /*comment*/ {

    }
}

fun test4() {
    block@




    //eol comment
    {

    }
}`

I would suggest to use 'betweenInside(LABEL_QUALIFIER, EOL_COMMENT, LABELED_EXPRESSION).spacing(0, Int.MAX_VALUE, 0, true, settings.KEEP_BLANK_LINES_IN_CODE)' or remove custom rules for the comment cases.

goodwinnk:
Merged into master, thank you!

PR #980, opened 2016-10-27
avoid deprecated usage of InspectionProjectProfileManager

develar:
I didn't compile it, but should be ok :)


yole:
Pushed a sligthly different fix (compatible with IJ 2016.2) in 38a448ebbdfe2d39e452ca83c25db7b46ff7e669

PR #981, opened 2016-10-27
KT-14390 stop "Remove redundant '.let' call" when receiver is used

shiraji:
#KT-14390 Fixed

https://youtrack.jetbrains.com/issue/KT-14390


mglukhikh:
Thanks, I've integrated this in master with minor style fixes.

PR #982, opened 2016-10-27
Use "Generate Tests" to regenerate `testLabeledExpression`

shiraji:
It keeps showing the diff when I run "Generate Tests". I think this pull request https://github.com/JetBrains/kotlin/pull/978/files generates test cases without running "Generate Tests".


mglukhikh:
As far as I can see now, these two tests are in the same order in master branch?

PR #984, opened 2016-10-29
KT-14575 Implement "Redundant if" quickfix's negations

shiraji:
#KT-14575 Fixed

https://youtrack.jetbrains.com/issue/KT-14575


mglukhikh:
BTW, I don't think that `LIKE_UNUSED_SYMBOL` highlighting is so gut for this inspection (usually, such a highlighting is in use when something can just be deleted).

mglukhikh:
Pushed into master & 1.0.5 with some refactoring (also changed highlighting style to WEAK_WARNING), thanks for your efforts

PR #988, opened 2016-11-01
KT-14396 Implement "Remove redundant '.let' call" binary operator support

shiraji:
#KT-14396 Fixed

https://youtrack.jetbrains.com/issue/KT-14396

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
@@ -50,7 +50,7 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
                     is KtDotQualifiedExpression -> left.replaceFirstReceiver(factory, receiver, parent.operationSign == KtTokens.SAFE_ACCESS)
                     else -> receiver
                 }
-                val newExpression = factory.createExpression("${newLeft.text}${operationReference.text}${right?.text ?: ""}")
+                val newExpression = factory.createExpressionByPattern("$0$1$2", newLeft.text, operationReference.text, right?.text ?: "")
                 parent.replace(newExpression)
             }
             else -> {
@@ -58,7 +58,7 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
                     is KtDotQualifiedExpression -> left.deleteFirstReceiver().text
                     else -> ""
                 }
-                val newExpression = factory.createExpression("$newLeftExpressionTemplate${operationReference.text}${right?.text ?: ""}")
+                val newExpression = factory.createExpressionByPattern("$0$1$2", newLeftExpressionTemplate, operationReference.text, right?.text ?: "")
                 element.replace(newExpression)
             }
         }
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
@@ -56,7 +56,7 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
             else -> {
                 val newLeftExpressionTemplate = when (left) {
                     is KtDotQualifiedExpression -> left.deleteFirstReceiver().text
-                    else -> ""
+                    else -> KtTokens.THIS_KEYWORD.value
                 }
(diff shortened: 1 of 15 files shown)


idea/testData/intentions/replaceSingleLineLetIntention/letWithThisShortBinaryExpression.kt.after
```diff
@@ -0,0 +1,6 @@
+// WITH_RUNTIME
+// IS_APPLICABLE: true
+
+fun Int.foo() {
+    +1
```
mglukhikh: This transformation does not look correct for me. I would expect `this + 1` as a result here,

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
```diff
@@ -31,18 +31,49 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
...
+        val factory = KtPsiFactory(element.project)
+        val parent = element.parent
+        when (parent) {
+            is KtQualifiedExpression -> {
+                val receiver = parent.receiverExpression
+                val newLeft = when (left) {
+                    is KtDotQualifiedExpression -> left.replaceFirstReceiver(factory, receiver, parent.operationSign == KtTokens.SAFE_ACCESS)
+                    else -> receiver
+                }
+                val newExpression = factory.createExpression("${newLeft.text}${operationReference.text}${right?.text ?: ""}")
```
mglukhikh: I think `createExpressionByPattern` should be better here and below. I'm trying to avoid using `text`, when possible.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
```diff
@@ -59,17 +90,49 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
...
         val parameterName = lambdaExpression.getParameterName() ?: return false
-        val receiverExpression = dotQualifiedExpression.getLeftMostReceiverExpression()
-        if (receiverExpression.text != parameterName) return false
-        dotQualifiedExpression.selectorExpression?.let {
-            if (it.anyDescendantOfType<KtLambdaExpression>()) return false
+        val bodyExpression = lambdaExpression.bodyExpression?.children?.singleOrNull() ?: return false
+
+        return when (bodyExpression) {
+            is KtBinaryExpression -> bodyExpression.isApplication(parameterName)
+            is KtDotQualifiedExpression -> bodyExpression.isApplicable(parameterName)
```
mglukhikh: Why is the first `isApplication`, but the second `isApplicable`?

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
```diff
@@ -59,17 +90,49 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
...
     }
 
+    private fun KtBinaryExpression.isApplication(parameterName: String): Boolean {
+        val left = left ?: return false
+
+        if (when (left) {
+            is KtNameReferenceExpression -> left.text != parameterName
+            is KtDotQualifiedExpression -> !left.isApplicable(parameterName)
+            else -> return false
+        }) return false
```
mglukhikh: This construction is hard to read. I'd rewrite it like

```
when (...) {
    is ... -> if (...) return false
}
```

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
shiraji: OK. I will use `createExpressionByPattern` next time.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
shiraji: Typo! Fix this

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
shiraji: Ah...that seems much simpler to read. Thanks.

mglukhikh:
Thanks, I'll integrate this PR today.

shiraji:
Thanks!

PR #989, opened 2016-11-06
Make the Kotlin Ant task more flexible

keyoffecka:
Allow to run the Kotlin Ant task with dependency libraries with non-standard names and located in non-standard locations.

### DIFF
--- a/ant/src/org/jetbrains/kotlin/ant/KotlinAntTaskUtil.kt
+++ b/ant/src/org/jetbrains/kotlin/ant/KotlinAntTaskUtil.kt
@@ -25,39 +25,96 @@ import java.lang.ref.SoftReference
 import java.net.JarURLConnection
 
 internal object KotlinAntTaskUtil {
+    const val COMPILER_JAR_NAME = "kotlin-compiler.jar"
+    const val RUNTIME_JAR_NAME = "kotlin-runtime.jar"
+    const val REFLECT_JAR_NAME = "kotlin-reflect.jar"
+
     private var classLoaderRef = SoftReference<ClassLoader?>(null)
 
-    private val libPath: File by lazy {
+    val libPath: File? by lazy {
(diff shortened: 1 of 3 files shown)


udalov:
Could you please explain why such change is needed? In what circumstances is the Kotlin Ant task launched without the compiler and/or the runtime beside it?

keyoffecka:
As the comment to the pull request says, this will make the Kotlin Ant task more flexible, meaning that you will be able to support custom project layouts. Please consider our project layout:

**ext** - bootstrap jars which we cannot fetch automatically from common repositories (this is where your ant jar and preloader are located)
**tool** - external tools we run to build the project, this may be pre-processors, source code generators or what so ever (this is where your compiler and reflect jars are located, hopefully we can fetch them from the Net)
**runtime** - jars not needed at the compile time but needed at runtime, used to run the application or system
**provided** - jars needed at both compile and runtime but are considered to be a part of a runtime environment or container
**compile** - jars needed at both compile and runtime but are considered to be a part of the application or  system  itself

You may consider this project layout good or bad, but this is just one of millions and millions examples of custom layouts that projects other than yours may use.

```
There are more things in heaven and earth, Horatio, 
Than are dreamt of in your philosophy. 
- Hamlet (1.5.167-8), Hamlet to Horatio
```

yole:
Given your last comment in https://youtrack.jetbrains.com/issue/KT-14654, I assume you won't be needing this extra flexibiity any more.

keyoffecka:
Don't worry. You may spend another million years discussing whether you need it or not, meanwhile it has already been deployed.

http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22kotlin-ant%22

But yes, you are totally right, after the truth about Kotlin comes out, unfortunately, all the made efforts just make no sense any more.

PR #990, opened 2016-11-08
KT-14589 fixed

qwwdfsad:
Fixed by removing local variable optimization, I haven't found the way to preserve this optimization for `val`s only


qwwdfsad:
Whoa, looks like I've messed with branch, will reopen properly

PR #839, opened 2016-03-20
KT-7930

voddan:
Added  extensions`String.toIntOrNull` etc; Optimased basing on openJDK implementations

PR #991, opened 2016-11-18
Sample Implementation of KT-7930

06needhamt:
Add tryParse For Numerical Primitives


ilya-g:
This approach is quite inefficient. The same functionality is to be covered by PR #839.

06needhamt:
@ilya-g I didn't notice it was done already it was marked as unassigned on the issue tracker

PR #993, opened 2016-11-26
Intention "Specify type explicitly" should offer all bounds for platform types

cypressious:
Fixes #KT-12633

Not sure how to test this.


mglukhikh:
Thanks for your efforts. I've moved your change into the intention itself to avoid behaviour changes for other callees. Merged manually into master & 1.0.6.

PR #994, opened 2016-11-27
KT-14890 Fix the problem with RemoveEmptyClassBodyIntention for nested class followed by a secondary constructor

shiraji:
#KT-14890 Fixed

https://youtrack.jetbrains.com/issue/KT-14890


mglukhikh:
Thanks for your efforts. Merged manually into master & 1.0.6 branches.

PR #995, opened 2016-11-29
Kapt3: Fix parameter names in overloaded methods (KT-15024)

grandstaish:


plugins/kapt3/src/org/jetbrains/kotlin/kapt3/stubs/parseParameters.kt
```diff
@@ -46,7 +45,7 @@ internal fun MethodNode.getParametersInfo(containingClass: ClassNode): List<Para
     for (index in startParameterIndex..parameterTypes.lastIndex) {
         val type = parameterTypes[index]
         var name = parameters.getOrNull(index - startParameterIndex)?.name
-                   ?: localVariables.getOrNull(index + (if (isStatic) 0 else 1))?.name
+                   ?: localVariables.getOrNull(index)?.name
                    ?: "p${index - startParameterIndex}"
```
grandstaish: @yanex Removing this offset fixes https://youtrack.jetbrains.com/issue/KT-15024, but I'm not sure exactly what the offset was for in the first place. Could you let me know what it's for and I can update this PR if needed.

grandstaish:
Closing as this appears to have been fixed

PR #996, opened 2016-12-05
KT-15030 Remove redundant calls of conversion methods: false positive for 'toList()'

shiraji:
#KT-15030 Fixed

https://youtrack.jetbrains.com/issue/KT-15030


mglukhikh:
OK. If we do not invent some better solution in a few days, I will integrate this one into master. Thanks for your efforts!

shiraji:
Yea, but it's still easy to add it back.

mglukhikh:
Integrated into master. Will integrate in 1.0.6 probably tomorrow

PR #997, opened 2016-12-08
KT-15142 Remove redundant calls of the conversion method wrongly shown for Any?.toString

shiraji:
#KT-15142 Fixed

https://youtrack.jetbrains.com/issue/KT-15142


mglukhikh:
Merged into master & 1.0.6, many thanks.

PR #999, opened 2016-12-16
remove comment line that is not relevant anymore

oshai:
the comment line is not relevant since this commit: https://github.com/JetBrains/kotlin/commit/54977ee09ad24cddc6fd889e2076071f8e85f0f7


ilya-g:
@oshai Can you explain why it's not relevant anymore? We still cannot delegate implementation of abstract  methods inherited from an abstract class to another instance of that class.

oshai:
Maybe it is just my mistake. I thought that when you added "by lazy" it made the comment not relevant. maybe it was just confusing me so you are welcome to just close the PR if its not the case.

ilya-g:
The code around that place is going to be refactored and the comment is likely to be gone. See PR #1008.

PR #1000, opened 2016-12-23
KT-6217 Add inspection that reports unused equals expression

shiraji:
#KT-6217 Fixed

https://youtrack.jetbrains.com/issue/KT-6217

### DIFF
--- a/idea/resources/inspectionDescriptions/UnusedEquals.html
+++ b/idea/resources/inspectionDescriptions/UnusedEquals.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection reports unused equals expression.
+</body>
+</html>
--- a/idea/resources/inspectionDescriptions/UnusedEquals.html
+++ b/idea/resources/inspectionDescriptions/UnusedEquals.html
@@ -1,5 +1,5 @@
 <html>
 <body>
-This inspection reports unused equals expression.
+This inspection reports unused equals expressions.
 </body>
 </html>
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedEqualsInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedEqualsInspection.kt
@@ -24,6 +24,7 @@ import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtVisitorVoid
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
 import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsResultOfLambda
 
 class UnusedEqualsInspection : AbstractKotlinInspection() {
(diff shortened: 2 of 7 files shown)


idea/resources/inspectionDescriptions/UnusedEquals.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection reports unused equals expression.
```
cypressious: expression**s**

idea/src/org/jetbrains/kotlin/idea/inspections/UnusedEqualsInspection.kt
```diff
@@ -0,0 +1,71 @@
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitBinaryExpression(expression: KtBinaryExpression) {
+                super.visitBinaryExpression(expression)
+                if (expression.operationToken != KtTokens.EQEQ) return
+                val parent = expression.parent as? KtBlockExpression ?: return
+                val lastBlockStatement = parent.lastBlockStatementOrThis()
+                when {
+                    expression.evaluatesTo(lastBlockStatement) && expression.getLineNumber() == lastBlockStatement.getLineNumber() ->
+                        expression.getStrictParentOfType<KtLambdaExpression>()?.visitLambdaExpression(holder, expression) ?: holder.registerUnusedEqualsProblem(expression)
```
mglukhikh: All of this looks too nasty and unsafe. Please use `KtExpression.isUsedAsExpression` from `BindingContextUtils.kt` to determine whether your `equals` result is used or not.

idea/src/org/jetbrains/kotlin/idea/inspections/UnusedEqualsInspection.kt
```diff
@@ -0,0 +1,71 @@
...
+                        if (!expression.evaluatesTo(lastBlockStatementOrThis)) holder.registerUnusedEqualsProblem(expression)
+                    }
+                    else -> holder.registerUnusedEqualsProblem(expression)
+                }
+            }
+
+            private fun ProblemsHolder.registerUnusedEqualsProblem(expression: KtBinaryExpression) {
+                registerProblem(expression,
+                                "Unused equals expression",
+                                ProblemHighlightType.LIKE_UNUSED_SYMBOL)
```
mglukhikh: Should we provide some quick fix? I'd say that in simple cases with local variables or constants binary expression could be completely deleted, but in cases like `foo() == bar()` we should get

```
foo()
bar()
```

instead

idea/src/org/jetbrains/kotlin/idea/inspections/UnusedEqualsInspection.kt
```diff
@@ -0,0 +1,71 @@
...
+                    KotlinBuiltIns.isBoolean(lambdaTypeArguments[0].type) -> {
+                        val lastBlockStatementOrThis = bodyExpression?.lastBlockStatementOrThis() ?: return
+                        if (!expression.evaluatesTo(lastBlockStatementOrThis)) holder.registerUnusedEqualsProblem(expression)
+                    }
+                    else -> holder.registerUnusedEqualsProblem(expression)
+                }
+            }
+
+            private fun ProblemsHolder.registerUnusedEqualsProblem(expression: KtBinaryExpression) {
+                registerProblem(expression,
```
mglukhikh: I'd say it's better to provide also text range here and highlight `==` only as unused.

elizarov:
I'd suggest that in cases like "a == something" where a is some modifiable thing the most likely explanation is that it was a typo and a fix is to remove an extra =

cypressious:
Maybe all the quick fixes are a bit to speculative and not really
necessary. Having the warning is the most important thing.

Kirill Rakhman

Am 23.12.2016 12:27 nachm. schrieb "Roman Elizarov" <
notifications@github.com>:

elizarov:
Good point.

mglukhikh:
Ok, quick-fixes are really controversial here so we can omit them.

shiraji:
Done. Thanks for pointing them. 

At the first I was thinking a quick fix and notice there are so many ways to fix the expressions. As @cypressious https://github.com/JetBrains/kotlin/pull/1000#issuecomment-268977235, I decided not to provide them.

mglukhikh:
`isUsedAsResultOfLambda` is not the only case. What about something like `val eq = if (...) a == b else c == d`, or the same with braces: `val eq = if (...) { a == b } else { c == d }`?

shiraji:
Done for the case that the binary expression is used as an expression.

mglukhikh:
Merged manually into master

PR #1003, opened 2016-12-27
#KT-7940 J2K: convert Integer.MAX_VALUE to Int.MAX_VALUE. Fix

aTrask:
Any comments - welcome


j2k/src/org/jetbrains/kotlin/j2k/SpecialFiled.kt
```diff
@@ -0,0 +1,208 @@
...
+package org.jetbrains.kotlin.j2k
+
+import com.intellij.psi.CommonClassNames.*
+import com.intellij.psi.PsiField
+import org.jetbrains.kotlin.builtins.PrimitiveType
+import org.jetbrains.kotlin.j2k.ast.Expression
+import org.jetbrains.kotlin.j2k.ast.Identifier
+import org.jetbrains.kotlin.j2k.ast.QualifiedExpression
+
+enum class SpecialFiled(val qualifiedClassName: String?, val fieldName: String) {
```
semoro: SpecialField

j2k/src/org/jetbrains/kotlin/j2k/SpecialFiled.kt
```diff
@@ -0,0 +1,208 @@
...
+
+    BYTE_MAX_VALUE(JAVA_LANG_BYTE, "MAX_VALUE") {
```
semoro: You should not hard-code all possible constants.

Proper way is to check if class mapped to kotlin class, and if mapped and kotlin class has constant with such name and type, then preform convert

semoro:
It is possible to check if Java class has mapping to Kotlin using `JavaToKotlinClassMap.INSTANCE.mapJavaToKotlin(fqName)`, and then check if it has similar constant, with such value, then preform conversion

AlexTrask:
@semoro Thx for your help comment. It will help me apply your review notes

goodwinnk:
@aTrask Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

There were number of design issues in the old Java to Kotlin converted that forced us to make a major rewrite for the whole module. New converter can be found in nk2k directory in the repository and already enabled by default in 1.3.50-eap2.

When we had started this rewrite we didn't know the final list of issues that is going to be fixed or unresolved at the end.

KT-7940 is still reproducible in the new Java to Kotlin converter that is especially sad because code from this pull request can't be reused directly to fix the issue. Unfortunately fixing should be started from the ground.

PR #1004, opened 2016-12-27
group-and-fold implementation

ilya-g:
An implementation of https://github.com/Kotlin/KEEP/issues/23  proposal

### DIFF
--- a/libraries/stdlib/common/src/generated/_Strings.kt
+++ b/libraries/stdlib/common/src/generated/_Strings.kt
@@ -423,6 +423,13 @@ public header inline fun <K, M : MutableMap<in K, MutableList<Char>>> CharSequen
 public header inline fun <K, V, M : MutableMap<in K, MutableList<V>>> CharSequence.groupByTo(destination: M, keySelector: (Char) -> K, valueTransform: (Char) -> V): M
 
 /**
+ * Creates a [Grouping] source from a char sequence to be used later with one of group-and-fold operations
+ * using the specified [keySelector] function to extract a key from each character.
+ */
+@SinceKotlin("1.1")
+public header inline fun <K> CharSequence.groupingBy(crossinline keySelector: (Char) -> K): Grouping<Char, K>
+
+/**
  * Returns a list containing the results of applying the given [transform] function
  * to each character in the original char sequence.
  */
--- a/libraries/stdlib/src/kotlin/collections/Grouping.kt
+++ b/libraries/stdlib/src/kotlin/collections/Grouping.kt
@@ -24,36 +24,38 @@ public interface Grouping<T, out K> {
 }
 
 /**
- * Groups elements from the [Grouping] source by key and aggregates elements of each group with the specified [operation].
+ * Groups elements from the [Grouping] source by key and applies [operation] to the elements of each group sequentially,
+ * passing the previously accumulated value and the current element as arguments, and stores the results in a new map.
  *
  * The key for each element is provided by the [Grouping.keyOf] function.
  *
  * @param operation function is invoked on each element with the following parameters:
- *  - `key`: the key of a group this element belongs to;
- *  - `value`: the current value of the accumulator of a group, can be `null` if it's first `element` encountered in the group;
+ *  - `key`: the key of the group this element belongs to;
+ *  - `accumulator`: the current value of the accumulator of the group, can be `null` if it's the first `element` encountered in the group;
  *  - `element`: the element from the source being aggregated;
- *  - `first`: indicates whether it's first `element` encountered in the group.
+ *  - `first`: indicates whether it's the first `element` encountered in the group.
  *
  * @return a [Map] associating the key of each group with the result of aggregation of the group elements.
  */
 @SinceKotlin("1.1")
 public inline fun <T, K, R> Grouping<T, K>.aggregate(
-        operation: (key: K, value: R?, element: T, first: Boolean) -> R
+        operation: (key: K, accumulator: R?, element: T, first: Boolean) -> R
 ): Map<K, R> {
     return aggregateTo(mutableMapOf<K, R>(), operation)
 }
 
 /**
- * Groups elements from the [Grouping] source by key and aggregates elements of each group with the specified [operation]
- * to the given [destination] map.
+ * Groups elements from the [Grouping] source by key and applies [operation] to the elements of each group sequentially,
+ * passing the previously accumulated value and the current element as arguments,
+ * and stores the results in the given [destination] map.
  *
  * The key for each element is provided by the [Grouping.keyOf] function.
  *
  * @param operation a function that is invoked on each element with the following parameters:
- *  - `key`: the key of a group this element belongs to;
- *  - `accumulator`: the current value of the accumulator of the group, can be `null` if it's first `element` encountered in the group;
+ *  - `key`: the key of the group this element belongs to;
+ *  - `accumulator`: the current value of the accumulator of the group, can be `null` if it's the first `element` encountered in the group;
  *  - `element`: the element from the source being aggregated;
- *  - `first`: indicates whether it's first `element` encountered in the group.
+ *  - `first`: indicates whether it's the first `element` encountered in the group.
  *
  * If the [destination] map already has a value corresponding to some key,
  * then the elements being aggregated for that key are never considered as `first`.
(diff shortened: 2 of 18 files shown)


ilya-g:
TODO
- add SinceKotlin
- add headers for platform specialized functions.

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,278 @@
...
+ * - [Sequence.groupingBy]
+ * - [Array.groupingBy]
+ * - [CharSequence.groupingBy]
+ *
+ * For the list of group-and-fold operations available, see the [extension functions](#extension-functions) for `Grouping`.
+ */
+@SinceKotlin("1.1")
+public interface Grouping<T, out K> {
+    /** Returns an [Iterator] which iterates through the elements of the source. */
+    fun elementIterator(): Iterator<T>
```
cy6erGn0m: @{18f15b45-4413-42a3-82e7-d057983d1e65,Ilya Gorbunov} `sourceIterator()` or just `source()` ? Need to be documented: whether it creates iterator every time or keep the same instance. If the behaviour is undefined we have to notice it here.

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,278 @@
...
+ *  - `element`: the element from the source being aggregated;
+ *  - `first`: indicates whether it's first `element` encountered in the group.
+ *
+ * @return a [Map] associating the key of each group with the result of aggregation of the group elements.
+ */
+@SinceKotlin("1.1")
+public inline fun <T, K, R> Grouping<T, K>.aggregate(
+        operation: (key: K, value: R?, element: T, first: Boolean) -> R
+): Map<K, R> {
+    val result = mutableMapOf<K, R>()
```
cy6erGn0m: Why don't we use `aggregateTo` instead?

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * If the [destination] map already has a value corresponding to some key,
+ * then the elements being aggregated for that key are never considered as `first`.
+ *
+ * @return the [destination] map associating the key of each group with the result of aggregation of the group elements.
+ */
+@SinceKotlin("1.1")
+public inline fun <T, K, R, M : MutableMap<in K, R>> Grouping<T, K>.aggregateTo(
+        destination: M,
+        operation: (key: K, accumulator: R?, element: T, first: Boolean) -> R
```
cy6erGn0m: `value` vs `accumulator`, `v` vs `acc`

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,278 @@
...
+ * @return a [Map] associating the key of each group with the count of element in the group.
+ */
+@SinceKotlin("1.1")
+@JvmVersion
+public inline fun <T, K> Grouping<T, K>.eachSumOf(valueSelector: (T) -> Int): Map<K, Int> =
+        // fold(0) { acc, e -> acc + valueSelector(e)} optimized for boxing
+        fold(
+                initialValueSelector = { k, e -> kotlin.jvm.internal.Ref.IntRef() },
+                operation = { k, acc, e -> acc.apply { element += valueSelector(e) } })
+        .mapValues { it.value.element }
```
cy6erGn0m: as we already use internal API hack we could do even more: we can create replace `IntRef` to `Int` inplace and then do unchecked cast.

libraries/stdlib/common/src/generated/_Strings.kt
```diff
@@ -423,6 +423,13 @@ public header inline fun <K, M : MutableMap<in K, MutableList<Char>>> CharSequen
 public header inline fun <K, V, M : MutableMap<in K, MutableList<V>>> CharSequence.groupByTo(destination: M, keySelector: (Char) -> K, valueTransform: (Char) -> V): M
 
 /**
+ * Creates a [Grouping] source from a char sequence to be used later with one of group-and-fold operations
+ * using the specified [keySelector] function to extract a key from each character.
+ */
+@SinceKotlin("1.1")
+public header inline fun <K> CharSequence.groupingBy(crossinline keySelector: (Char) -> K): Grouping<Char, K>
```
cy6erGn0m: why there is specialization here? the implementations look the same

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: I believe `elementIterator` being a `fun` conveys the intent well. If that was the same instance every time, it would be a property.

libraries/stdlib/common/src/generated/_Strings.kt
ilya-g: It's how the code is generated now.

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: I wouldn't call it internal API hack, because that API is used publicly by the compiler in the generated bytecode.
`mapInPlace` is possible as JVM-only optimization, otherwise it would make the code non-portable to platforms, where generics are reified.

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: It was called `source` in the beginning, that I renamed it to `iterator` and then to `elementIterator`

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: I wouldn't call it internal API hack, because that API is used publicly by the compiler in the generated bytecode.
mapInPlace is possible as JVM-only optimization, otherwise it would make the code non-portable to platforms, where generics are reified.

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: After some consideration, `sourceIterator` seems to look better.

ilya-g:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:35`|
|---|
I'm not sure how could it be possible without wrapping each value. And wrapping each value is very ineffective in terms of number of allocations.

ilya-g:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:35`|
|---|
I'm not sure how could it be possible without wrapping each value. And wrapping each value is very ineffective in terms of number of allocations.

yole:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:35`|
|---|
True. I don't have any better suggestion to improve this, so let's keep it as is.

yole:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:47`|
|---|
You're using inconsistent wording - "aggregates with" or "accumulates with" - and neither of those is a standard usage of the corresponding verbs with the preposition. I think you need to provide a more detailed explanation of what happens here - something like applies [operation] to the elements of each group sequentially, passing the previously accumulated value and the current element as arguments, and stores the results in the given [destination] map". This applies to other functions as well.

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,282 @@
+package kotlin.collections
+
+/**
+ * Represents a source of elements with a [keyOf] function, which can be applied to each element to get its key.
+ *
+ * A [Grouping] structure serves as an intermediate step in group-and-fold operations:
+ * they group elements by their keys and then fold each group with some aggregating operation.
+ *
+ * It is created by attaching `keySelector: (T) -> K` function to a source of elements.
+ * To get an instance of [Grouping] use one of `groupingBy` extension functions:
```
yole: |`line 32`|
|---|
"a group" -> "the group" here and below

yole:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:213`|
|---|
count of element_s_

yole:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:219`|
|---|
Why the space after the open parenthesis?

libraries/stdlib/src/kotlin/collections/Grouping.kt
```diff
@@ -0,0 +1,282 @@
...
+ * - [Iterable.groupingBy]
+ * - [Sequence.groupingBy]
```
yole: |`line 61`|
|---|
Any chance for some `@sample`'s?

yole:
|`libraries/stdlib/src/kotlin/collections/Grouping.kt:240`|
|---|
Why is this function called `eachSumOf` and not `eachSumBy`, which would be consistent with `Iterable.sumBy`?

ilya-g:
What does this comment relate to?

ilya-g:
> Why is this function called eachSumOf and not eachSumBy, which would be consistent with Iterable.sumBy?

We're going to rename `sumBy` and `sumByDouble` to `sumOf` for Iterables, however it's blocked by KT-11265. Maybe we should not introduce `eachSumOf` until then.

ilya-g:
I've decided to hold it back for a while.

libraries/stdlib/src/kotlin/collections/Grouping.kt
ilya-g: `aggregate` and `aggregateTo` are not intended to be used much in user code, they are here more to implement other folding functions when `fold` or `reduce` do not fit for some reason.
Other functions (fold, reduce, eachCount, etc) will get samples eventually.

PR #1006, opened 2016-12-27
KT-11208 - readLine() shouldn't use BufferedReader

lukaszwojtow:


libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -135,7 +135,7 @@ public inline fun println() {
 // Since System.in can change its value on the course of program running,
 // we should always delegate to current value and cannot just pass it to InputStreamReader constructor.
 // We could use "by" implementation, but we can only use "by" with interfaces and InputStream is abstract class.
-private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(object : InputStream() {
+private val stdin: InputStreamReader by lazy { InputStreamReader(object : InputStream() {
```
ilya-g: If we're not going to buffer contents, `stdin` wrapper isn't required anymore.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -171,11 +171,21 @@ private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(obj
...
 
 /**
  * Reads a line of input from the standard input stream.
  *
  * @return the line read or `null` if the input stream is redirected to a file and the end of file has been reached.
  */
-public fun readLine(): String? = stdin.readLine()
+public fun readLine(): String? {
+    tailrec fun readChars(oldBuffer: StringBuffer? = null): StringBuffer? {
+        val c = mystdin.read()
```
ilya-g: What is `mystdin` here?

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -171,11 +171,21 @@ private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(obj
...
+})}
 
 /**
  * Reads a line of input from the standard input stream.
  *
  * @return the line read or `null` if the input stream is redirected to a file and the end of file has been reached.
  */
-public fun readLine(): String? = stdin.readLine()
+public fun readLine(): String? {
+    tailrec fun readChars(oldBuffer: StringBuffer? = null): StringBuffer? {
```
ilya-g: Why do you choose `StringBuffer` instead of `StringBuilder`?

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -171,11 +171,21 @@ private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(obj
...
+        val c = mystdin.read()
+        if(c < 0) return oldBuffer
+        val buffer = oldBuffer ?: StringBuffer()
+        val ch = c.toChar()
+        if(ch == '\r' || ch == '\n') return buffer
```
ilya-g: We should support "\n" and "\r\n" line delimiters.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -171,11 +171,21 @@ private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(obj
...
+        return readChars(buffer.append(ch))
+    }
+    return readChars()?.toString()
+}
```
ilya-g: Overall, `buildString` + imperative loop might be more clear than tail recursion approach.

PR #1008, opened 2016-12-28
KT-11208 - readLine() shouldn't use BufferedReader

lukaszwojtow:


libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +132,34 @@ public inline fun println() {
...
+
+public fun readLine(): String? {
+    val buffer = StringBuilder()
+    var c = System.`in`.read()
+    if(c < 0) return null
+    buffer.append(c.toChar())
+    do {
+        c = System.`in`.read()
+        if(c < 0) return buffer.toString()
+        val ch = c.toChar()
```
ilya-g: Unfortunately, this will work only for US-ASCII encoding.

ilya-g:
By the way, there is no need to close pull request and submit another one, when you want to add some edits. You can push you commits to the same branch (or even force push them), and they will be attached to the same review.

lukaszwojtow:
Actually, it turns out we need this 'stdin' wrapper as it causes some funny errors otherwise. `readLine()` needs just `read():Int` from `InputStream` but I left others untouched for now.

PR #1011, opened 2017-01-06
Convert function to property shouldn't insert explicit type if it was inferred previously

cypressious:
Fixes #KT-14820

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertFunctionToPropertyIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertFunctionToPropertyIntention.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2015 JetBrains s.r.o.
+ * Copyright 2010-2017 JetBrains s.r.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertFunctionToPropertyIntention.kt
```diff
@@ -88,8 +85,10 @@ class ConvertFunctionToPropertyIntention : SelfTargetingIntention<KtNamedFunctio
             function.setName(newName)
 
             val property = originalFunction.replace(psiFactory.createProperty(function.text)) as KtProperty
-            if (needsExplicitType) {
-                elementsToShorten.add(property.typeReference!!)
+
+            if (!hasExplicitType) {
+                val linebreak = property.getter?.node?.treePrev
```
yole: Ouch. This should be handled by a formatter rule, not by AST manipulations.

cypressious: I agree this isn't pretty. Currently, the formatter allows both, getters on the same line and on the next line. Forcing one of them could be to strict.

Maybe the `KtPsiFactory` could get an option to create a property with the getter on the same line. What do you think @yole?

yole: Actually yes, the way this intention replaces the internal child nodes of `KtFunction` is quite ugly. I think it would be much nicer to use `KtPsiFactory.CallableBuilder` to construct the property.

cypressious: I think the reason for the current way is because it preserves annotations, modifiers, receiver type and other things I probably forgot. If we redid it, we would have to manually take care of it.

cypressious:
@yole I reworked the code and added some more tests. I think it's cleaner now.

PR #1012, opened 2017-01-06
Fix Quickfix for missing operator adds infix modifier to created function

cypressious:
Fixes #KT-14907


yole:

PR #1013, opened 2017-01-06
Add quickfix for changing primitive cast to primitive conversion method

cypressious:
Fixes #KT-15564


mglukhikh:
Merged into master, thanks.

PR #1014, opened 2017-01-06
This repository should include Scala License

kmizu:
since kotlinc inherits scalac

e.g. https://github.com/JetBrains/kotlin/blob/master/compiler/cli/bin/kotlinc#L17


yole:
The Kotlin repository contains no code with Lightbend copyright. The scripts were copied in 2011, before Lightbend existed, so the appropriate Scala license to include is the older one. I've added it in af1ed78e171065cc4faa46c22cfe0ceaef517f62

yole:
And thanks for bringing this issue to our attention!

PR #1015, opened 2017-01-06
"Add names to call arguments" shouldn't appear when the only argument is a trailing lambda

cypressious:
Fixes #KT-15501


mglukhikh:
Merged into master, thanks

PR #1017, opened 2017-01-07
Fixes KT-15540

austinv11:
This commit fixes NoSuchElementExceptions being thrown when a KFunction has `findAnnotation` called when it does not contain the annotation. Instead it will return null (as it seems to have been intended due to the return type being nullable).


udalov:
Thanks, I've also changed the declaration in core/reflection.jvm/src/kotlin/reflect/KAnnotatedElements.kt and added a test, will push soon.

PR #1018, opened 2017-01-12
KT-15406 Fix the problem with ConvertPrimaryConstructorToSecondaryIntention for enum

shiraji:
#KT-15406 Fixed

https://youtrack.jetbrains.com/issue/KT-15406


mglukhikh:
Ok, thank you, I'll changed this a bit and merged into 1.1 & 1.0.7.

PR #1019, opened 2017-01-13
KT-15553 Support compile constant for "Copy concatenation text to clipboard"

shiraji:
#KT-15553 Fixed

https://youtrack.jetbrains.com/issue/KT-15553


mglukhikh:
Thank you, merged it into master.

PR #1021, opened 2017-01-18
merge Keplin project scripting into the core of Kotlin

apatrida:
as per discussions in the slack scripting channel...

* Merge Keplin project scripting code into Kotlin core.
* Overhauls the scripting layers (GenericRepl and related, and JSR223 and related)
* Adds reset back in time (reset back to a given code line, dropping the top of the history stack off)
* Adds repeating modes (none, repeat only latest eval'd line, or repeat any previously executed again in random order)
* Also adds better thread-safe IO capturing, default imports, SimpleRepl wrapper, more unit tests
* Adds other JSR223 engines
* fixed the Ivocable interface for Jsr223 which wasn't quite right.  Spec says that previously defined functions can be glumped together into an interface.  Instead, the current code just tried to cast the `Line_*` class as the interface which is not the same thing at all.  Now it delegates to the previous lines (in reverse order) to find the function matching the signature of the interface, via a proxy.  Also method / function calls also traverse previous lines to find the first available matching function to call.
* Fuller test coverage (in script-utils)


apatrida:
Also note:  invocable for Jsr223 wasn't quite right.  It says that previously defined functions can be glumped together into an interface.  Instead, the current code just tried to cast the Line_* class as the interface which is not the same thing at all.  Now it delegates to the previous lines (in reverse order) to find the function matching the signature.

apatrida:
The interfaces in `ReplApi.kt` are broken into parts because I was trying to limit what functionality had to be present for things like remote daemon REPL compiler.  They might be able to be combined back into fewer but currently act as a list of small Traits each part can implement in part.

apatrida:
Maybe after 1.1 we can look at `ReplInterpreter.kt` and unifying it with using the `Generic*` classes

raulraja:
Is this still happening or has this effort been abandoned? 
We are having issues with the current Script engine impl where it just hangs forever trying to resolve scoped variables.

ligee:
The most parts from this PR is already merged into kotlin long time ago.
The most important omission is dealing with readers/writers in JSR-223, should be handled separately https://youtrack.jetbrains.com/issue/KT-29501

PR #1023, opened 2017-01-22
Add inspection for private primary constructors in data classes

cypressious:
Fixes #KT-15709


mglukhikh:
Merged into master, thank you.

PR #1024, opened 2017-02-04
KT-12183 Implement an intention converting several calls with same receiver to with/apply/run

shiraji:
#KT-12183 Fixed

https://youtrack.jetbrains.com/issue/KT-12183

### DIFF
--- a/idea/resources/intentionDescriptions/ConvertToApplyIntention/after.kt.template
+++ b/idea/resources/intentionDescriptions/ConvertToApplyIntention/after.kt.template
@@ -1,0 +1,7 @@
+fun foo() {
+    <spot>val a = MyClass().apply {
+        a.setFoo(1)
+        a.setBar(2)
+        a.setBaz(3)
+    }</spot>
+}
--- a/idea/resources/intentionDescriptions/ConvertToApplyIntention/before.kt.template
+++ b/idea/resources/intentionDescriptions/ConvertToApplyIntention/before.kt.template
@@ -1,0 +1,6 @@
+fun foo() {
+    <spot>val a = MyClass()
+    a.setFoo(1)
+    a.setBar(2)
+    a.setBaz(3)</spot>
+}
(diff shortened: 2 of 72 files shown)


mglukhikh:
What about chained calls like `a.foo().bar()` or `a.x.bar()`? There are no such tests, and I'm not sure they are handled correctly here.

mglukhikh:
Also, seems important to distinguish `apply()` from two other functions here because it generates its result differently. I'd say that by default `apply()` should have less priority or even should not be shown if we do not want to use its result. `rcv.run()` and `with(rcv)` looks more logical in these situations.

shiraji:
OK, then I would just remove intention for `apply()`. Then later, we can make it back if someone find a way to implement what @cypressious said https://youtrack.jetbrains.com/issue/KT-12183#comment=27-1866609 . What do you think?

mglukhikh:
Yes, it's possible way. I just do not like an idea to have three effectively identical intentions.

shiraji:
I have done a lot of changes here. Could you review this PR again, please?

mglukhikh:
Sorry please, will merge it today and tomorrow :). We're in release rush now.

shiraji:
Of course!

mglukhikh:
I have to say that the code is not clean enough, because large part exists in both ConvertToScopeIntention and ConvertToApplyIntention. This should be solved before merging into master. I see that they have different base element types, but really it's not an obstacle. You can try to solve the problem yourself if you wish, or I can do it myself when we're out of release rush -- your choice.

shiraji:
Let me work on that.

Are you suggesting they all have the same base element type, or create a Util class for delegation?

shiraji:
OK, now `ConvertToScopeIntention` is the parent of all 3 intentions.

mglukhikh:
So OK, I have finally merged it, thanks.

shiraji:
Thanks! And congra to releasing 1.1!!!

PR #1028, opened 2017-02-09
PrimitiveType.BITS and PrimitiveType.BYTES

pcarrier:


ilya-g:
Relates to https://youtrack.jetbrains.com/issue/KT-8247

ilya-g:
We have introduced these constants as `SIZE_BITS` and `SIZE_BYTES` in each companion object of the integer primitives, i.e. excluding `Float` and `Double` types. See this commit: a2eb62d8f1370b2f152f21f6c9cd605d9f85758c

PR #1029, opened 2017-02-17
Add intention to add missing components to destructuring assignment

cypressious:
Fixes #KT-16258


mglukhikh:
Slightly polished and merged into master, thanks!

PR #1031, opened 2017-03-02
Fix two Rename related bugs

cypressious:
https://youtrack.jetbrains.com/issue/KT-14401
https://youtrack.jetbrains.com/issue/KT-16605

I manually tested it, but I can't get the automated tests to work.

I tried `SimpleNameReferenceRenameTest` but it always gets the parent element instead of the one behind the caret.

Then I tried `InplaceRenameTest` but it always throws an exception.


asedunov:
Thank you for your contribution! I'll add some tests and do a manual merge afterwards

PR #1033, opened 2017-03-06
Add extension Iterable.indexOfOrNull()

pawegio:
In my opinion such an extension method is missing in stdlib.


udalov:
Thanks for the contribution, but I don't think this is the right place for such change. `addToStdlib.kt` is a part of the Kotlin compiler code. We only declare something there which is used in the compiler code and is generic enough to be a good candidate for the standard library. Until a function is used in the compiler code, there's no need to declare it there. We try not to keep unused declarations around in our project.

If you're expressing an opinion that a function would be useful for the standard library, please file a [YouTrack issue](https://kotl.in/issue) instead. If you mean to actually add a function to the standard library, it would require a bit more than adding its implementation: it should be documented and tested as well.

PR #1035, opened 2017-03-06
KT-16585: Use gradle's built in input task checking

JLLeitschuh:
This is the beginning of a fix for KT-16585.
The up to date when block was trying to replicate gradle's directory update check.

There are other problems that are detailed in KT-16585 that need to be fixed.

**Contributor gripe**
I been able to actually compile this code. I found setting up my environment to test this code was far more complicated than I expected.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/KotlinGradlePluginIT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/KotlinGradlePluginIT.kt
@@ -428,4 +428,37 @@ class KotlinGradleIT: BaseGradleIT() {
             assertSuccessful()
         }
     }
+
+    @Test
+    fun testGradleJavaIcWorking() {
+        val project = Project("kotlinJavaProject", "2.14.1")
+        project.setupWorkingDir()
+
+        val buildScript = File(project.projectDir, "build.gradle")
+
+        buildScript.modify { "$it\n" + "compileJava.options.incremental = true" }
+        project.build("build") {
+            assertSuccessful()
+        }
+
+        // Then modify a Java source and check that compileJava is incremental:
+        File(project.projectDir, "src/main/java/demo/HelloWorld.java").modify { "$it\n" + "class NewClass { }" }
+        project.build("build") {
+            assertSuccessful()
+            assertNotContains("not incremental")
+            assertContains("Incremental compilation")
+        }
+
+        // Then modify a Kotlin source and check that Gradle sees that Java is not up-to-date:
+        File(project.projectDir, "src/main/kotlin/helloWorld.kt").modify {
+            it.trim('\r', '\n').trimEnd('}') + "\nval z: Int = 0 }"
+        }
+        project.build("build") {
+            assertSuccessful()
+            assertContains(":compileKotlin")
+            assertNotContains(":compileJava UP-TO-DATE")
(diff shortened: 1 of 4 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -551,8 +552,22 @@ internal open class KotlinAndroidPlugin(
...
+        if (!gradleSupportsJavaIcWithClassesDirs && kotlinClassesCompiled) {
+            logger.info("Marking $task out of date, because kotlin classes are changed")
+            return@upToDateWhen false
+        }
+        true
+    }
+
     // Make Gradle check if the javaTask up-to-date based on the Kotlin classes
-    javaTask.inputs.dir(kotlinTask.destinationDir!!)
+    javaTask.inputs.dir(kotlinTask.destinationDir)
```
tsvtkv: I think you should also include kapt1 annotations file:
https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/AnnotationProcessingManager.kt#L139

In kapt1 AP is performed by java task, so if this file is not included, AP may be skipped when only source annotations are modified.
Kapt3 should probably work as intended.
I think the test should be added for this case (Java IC + Kapt 1/3 + change only Kotlin source annotations).

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -551,8 +552,22 @@ internal open class KotlinAndroidPlugin(
 }
 
 private fun configureJavaTask(kotlinTask: KotlinCompile, javaTask: AbstractCompile, logger: Logger) {
+    // Gradle Java IC in older Gradle versions (before 2.14) cannot check .class directories updates.
+    // To make it work, reset the up-to-date status of compileJava with this flag.
+    kotlinTask.anyClassesCompiled = false
+    javaTask.outputs.upToDateWhen { task ->
+        val gradleSupportsJavaIcWithClassesDirs = ParsedGradleVersion.parse(javaTask.project.gradle.gradleVersion)
```
tsvtkv: I think that version checking should be performed before `upToDateWhen` callback is added

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/KotlinGradlePluginIT.kt
```diff
@@ -428,4 +428,37 @@ class KotlinGradleIT: BaseGradleIT() {
             assertSuccessful()
         }
     }
+
+    @Test
+    fun testGradleJavaIcWorking() {
+        val project = Project("kotlinJavaProject", "2.14.1")
```
tsvtkv: What about newer Gradle versions?
Btw, it would be interesting to test if Java compile avoidance works in this case with the Gradle 3.4 (in other words that Java is not compiled if Kotlin public API is not changed).

JLLeitschuh:
This looks good. Thanks for updating this @h0tk3y.

h0tk3y:
Closing because it has been merged manually. Thanks!

PR #1037, opened 2017-03-10
KT-16131 Add quickfix for 'Cannot access member'

shiraji:
#KT-16131 Fixed

https://youtrack.jetbrains.com/issue/KT-16131


mglukhikh:
There is `ChangeVisibilityFix` already, there is no need to re-invent it.

mglukhikh:
Also, changing visibility to `internal` or `protected` may be not enough, and ideally should not be suggested when it cannot help.

mglukhikh:
See also `MakeVisibleFactory`.

shiraji:
Could you check this again please?

mglukhikh:
Really situation is even more simple because we already have `MakeVisibleFactory`. I'll do all integration needed today

shiraji:
I was thinking `MakeVisibleFactory` is already used for `INVISIBLE_MEMBER`. https://github.com/shiraji/kotlin/blob/da720bcfc416e31d5286157d93ff109f1197fe42/idea/src/org/jetbrains/kotlin/idea/quickfix/QuickFixRegistrar.kt#L173-L175

Thanks for your help.

mglukhikh:
https://github.com/JetBrains/kotlin/commit/27e1462b006915ce375540467f5ff05bb3a988c0

PR #1038, opened 2017-03-20
KT-16764 Extend Java source directory set instead of copying srcDirs

Minecrell:
See https://youtrack.jetbrains.com/issue/KT-16764

Starting with Gradle 3.2, Gradle supports defining file collections with task dependencies as source directory for `SourceDirectorySet` (e.g. to generate some source before compiling the project). Normally, Gradle will automatically run them before compiling the sources.

The Kotlin plugin breaks this by copying the plain source directories (as `File`s) to the Kotlin source set.

Gradle has the `SourceDirectorySet.source(SourceDirectorySet)` method as a solution for this which allows extending another source directory set. It's backward compatible all the way back to Gradle 1.0 and will also handle task dependencies properly in newer Gradle versions. Gradle will lazily evaluate the extended source directory set when using this method so it is no longer necessary to apply it in the `afterEvaluate` block.

I've added a new Gradle plugin integration test that will only pass with the fix. The existing tests for extra custom source directories in the Java source set also pass successfully with my changes.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/SimpleKotlinGradleIT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/SimpleKotlinGradleIT.kt
@@ -55,6 +55,13 @@ class SimpleKotlinGradleIT : BaseGradleIT() {
     }
 
     @Test
+    fun testSrcDirTaskDependency() {
+        Project("srcDirTaskDependency", "3.3").build("clean", "build") {
+            assertSuccessful()
+        }
+    }
+
+    @Test
     fun testLanguageVersion() {
         Project("languageVersion", GRADLE_VERSION).build("build") {
             assertFailed()
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/srcDirTaskDependency/build.gradle
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/srcDirTaskDependency/build.gradle
@@ -1,0 +1,41 @@
+buildscript {
+    repositories {
+        mavenLocal()
+    }
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    }
+}
+
+apply plugin: 'kotlin'
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+}
+
+task generateSources {
+    outputs.dir('generated')
+
+    doLast {
+        def file = new File('generated/test/TestClass.java')
+        file.parentFile.mkdirs()
+        file.text = """
+            package test;
+
+            public class TestClass {
+
+            }
+        """
+    }
+}
+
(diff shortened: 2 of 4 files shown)


h0tk3y:
The suggested fix doesn't work with Gradle 2.14.1 and below, failing with the following exception:
```
org.gradle.internal.typeconversion.UnsupportedNotationException: Cannot convert the provided notation to a File or URI: task ':generateSources'.
The following types/formats are supported:
  - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.
  - A String or CharSequence URI, for example 'file:/usr/include'.
  - A File instance.
  - A URI or URL instance.
	at org.gradle.internal.typeconversion.ErrorHandlingNotationParser.parseNotation(ErrorHandlingNotationParser.java:56)
	at org.gradle.api.internal.file.AbstractFileResolver.convertObjectToFile(AbstractFileResolver.java:118)
	at org.gradle.api.internal.file.BaseDirFileResolver.doResolve(BaseDirFileResolver.java:71)
        ...
        at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSrcDirTrees(DefaultSourceDirectorySet.java:147)
	at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSrcDirs(DefaultSourceDirectorySet.java:76)
	at org.jetbrains.kotlin.gradle.tasks.FilteringSourceRootsContainer.add(SourceRoots.kt:80)
	at org.jetbrains.kotlin.gradle.tasks.KotlinCompile.source(Tasks.kt:313)
	at org.jetbrains.kotlin.gradle.plugin.Kotlin2JvmSourceSetProcessor$doTargetSpecificProcessing$1.execute(KotlinPlugin.kt:150)
```
I will probably add a version check to fall back to the original behavior with older versions.

Minecrell:
@h0tk3y The error looks like you are attempting to use my test code on Gradle 2.14.1. Defining a task as source directory only works in Gradle 3.2+. For older versions, only regular string/file instances can be used.

Minecrell:
@h0tk3y I've checked again and I'm quite sure everything is working as intended on Gradle 2, otherwise the [`testKotlinExtraJavaSrc` integration test](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/SimpleKotlinGradleIT.kt#L49-L54) wouldn't complete successfully with my changes using Gradle 2.10.

As said above, the task dependency as a Java source directory is only supported in Gradle 3.2+. This PR will only fix the Kotlin tasks to respect the task dependency, it doesn't patch the Gradle core to make it work on older Gradle versions.

Minecrell:
Fixed the conflicts, no functional change otherwise.

@h0tk3y Any news here? [KT-16764](https://youtrack.jetbrains.com/issue/KT-16764) is still broken with Kotlin 1.1.4-3 and Gradle 4.1. The single line change in this PR fixes it and still maintains compatibility with older Gradle versions. What's preventing it from getting merged? Thanks!

h0tk3y:
@Minecrell 

Thanks for your contribution!
The change you proposed is going to be included in 1.1.60 along with [another improvement](https://youtrack.jetbrains.com/issue/KT-17564) in source directories configuration [(commit)](https://github.com/JetBrains/kotlin/commit/e4969084fb1aed6b119a2513889b0c1601e32cdb). 

I've slightly reworked the test case from this PR, because in general we try to avoid creating additional test projects when there's already a similar one and a minor customization is enough.

PR #1039, opened 2017-03-20
KT-16722 KtReference should respect its scope when resolving declarations

Minecrell:
Suggested fix for: https://youtrack.jetbrains.com/issue/KT-16722

Currently, declarations for `KtReference` are searched globally in the project. Normally, this is not a problem, but as soon as multiple different Kotlin libraries exist in the project (e.g. when using Kotlin Gradle build scripts) the wrong declaration might be returned. This is especially annoying if the other declaration doesn't have any sources attached (e.g. for Gradle's Kotlin distribution).

This changes `KtReference` to search declarations only in the resolve scope of its element, which will usually refer to only the libraries of the current module.

### DIFF
--- a/idea/idea-analysis/src/org/jetbrains/kotlin/idea/codeInsight/DescriptorToSourceUtilsIde.kt
+++ b/idea/idea-analysis/src/org/jetbrains/kotlin/idea/codeInsight/DescriptorToSourceUtilsIde.kt
@@ -18,6 +18,7 @@ package org.jetbrains.kotlin.idea.codeInsight
 
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
+import com.intellij.psi.search.GlobalSearchScope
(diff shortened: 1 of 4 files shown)


pTalanov:
Hello.

First of all, thank you for your effort. 
Unfortunately we won't be accepting your PR. Similar changes (using different API) were just introduced into master as an effort to fix https://youtrack.jetbrains.com/issue/KT-15084 . 
Sadly I wasn't aware of your work on KT-16722 until today. I checked master branch against your example https://github.com/Minecrell/kotlintest2 and the problem is partially gone. Navigation works correctly for all declarations except for builtins which are somewhat of a special case ( Array.iterator() is a builtin). I'll be testing another patch to cover this case also.

Minecrell:
@pTalanov Thank you for your comment. Please let me know when there is a patch I can test.

PR #1040, opened 2017-03-21
WIP: KT-14109/	KT-16645 support for inlay param hints and type hints

fitermay:
This is a work in progress. Works against the master_171 branch. Submitting it for review and feedback. Also, I have some questions:

In no particular order:

1)  What should the default blacklist be?
2)  What other options should we provide on top of the ones provided in this impl?
3)  How should varargs be presented?
4)  It seems that the inherited blacklist is provided as a global settings.  Should we support a blacklist for JS modules vs Java modules? If so can it be done against the current API?
5) The seems to be an issue with the intentions being slow to toggle when the enable/disable option is toggled. Any plans to fix that on the IDEA end?


fitermay:
@yole Please review at your convinience

yole:
I've started a branch based on your work and made some additional changes (still work in progress):
https://github.com/JetBrains/kotlin/tree/rr/yole/filtermay-inlays

You're welcome to send feedback regarding these changes. And thanks a lot for your contribution! The core implementation is very good; I'm mostly adding new logic rather than rewriting your code.

yole:
Merged to master manually.

PR #1041, opened 2017-03-22
its --> it's in Char.kt

akkatracker:


yole:
Thanks!

udalov:
Other primitive classes also had this typo, I've fixed it in https://github.com/JetBrains/kotlin/commit/43f1c843ce09d4fa834559df30def9f05029b82f

akkatracker:
@udalov , thanks. I ran a quick search however I find github search is rather ineffective for these kinds of cases.

PR #1042, opened 2017-03-31
Configure the standard library and jvmTarget automatically

DanielThomas:
Configure the standard library and jvmTarget automatically based on the Java `sourceCompatibility` convention (it's implicitly set by Gradle, based on the JDK being used for compilation). This removes the boilerplate of declaring the standard library, and uses the JDK specific variant if appropriate, and sets the `jvmTarget` to match.

I'm not sure what your plans are in this space, but we've found this to be incredibly useful, and would reduce a default Kotlin project to:

```
plugins {
  id "org.jetbrains.kotlin.jvm" version "<version>"
}
```

Wanted to run this by someone on the Kotlin team before I spend time writing tests for this.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
@@ -8,10 +8,7 @@ import com.android.build.gradle.internal.variant.BaseVariantOutputData
 import com.android.build.gradle.internal.variant.TestVariantData
 import com.android.builder.model.SourceProvider
 import groovy.lang.Closure
-import org.gradle.api.InvalidUserDataException
-import org.gradle.api.Plugin
-import org.gradle.api.Project
-import org.gradle.api.ProjectConfigurationException
+import org.gradle.api.*
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.logging.Logger
 import org.gradle.api.logging.Logging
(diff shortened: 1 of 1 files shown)


runningcode:
On Android or on projects which require backward compatibility with older sdks, sourceCompatibility is set to 1.7 or 1.8 but the compiled source must not use JDK 1.8 classes. This would cause a runtime crash on older devices on android.

DanielThomas:
@runningcode I was taken by surprise by the change in Kotlin 1.1.2 to require JDK 1.8 without also also addressing this, and requiring that the older JDK classes be available for compilation to avoid JDK 8 classes leaking into compiled classes.

My suggestion would be that if the current JDK differs from the `jvmTarget`, the build should fail unless compatible classes are provided via an environment variable, or explicit configuration.

sithsupersu:
Couldn't  you also flag compatable classes??..

DanielThomas:
Our solution for `targetCompatibility` was a plugin that configures `jvmTarget` automatically:

https://github.com/nebula-plugins/gradle-java-cross-compile-plugin

h0tk3y:
We have implemented JVM target mapping for Android as per https://youtrack.jetbrains.com/issue/KT-21030
An equivalent for non-Android JVM projects is also planned: https://youtrack.jetbrains.com/issue/KT-29470
As to the standard library dependency, we may provide automatic dependency configuration in the future which will be valid for Kotlin/JS and Multiplatform projects as well. Another important point is that we need a way to cancel this automatic dependency configuration for some special use cases such as compiling the standard library itself, and this implementation does not allow that. This PR also uses obsolete artifact IDs for kotlin-stdlib JDK 7 & 8 packages. 
@DanielThomas thanks for your effort! I'm closing this PR for now, but we'll consider reusing it in further work.

PR #1044, opened 2017-04-04
Added the substring and setLength missing methods to the StringBuilder of the JS stdlib

albertoruibal:
They were available only in the StringBuilder of the JVM stdlib (because it extends the java.lang.StringBuilder), and not when compiling for JS

### DIFF
--- a/js/js.libraries/src/core/text.kt
+++ b/js/js.libraries/src/core/text.kt
@@ -33,9 +33,17 @@ public class StringBuilder(content: String = "") : Appendable, CharSequence {
     override val length: Int
         get() = string.asDynamic().length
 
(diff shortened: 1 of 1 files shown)


js/js.libraries/src/core/text.kt
```diff
@@ -33,10 +33,18 @@ public class StringBuilder(content: String = "") : Appendable, CharSequence {
     override val length: Int
         get() = string.asDynamic().length
 
+    fun setLength(newLength : Int) {
+        string = string.substring(0, newLength)
```
ilya-g: This implementation doesn't conform to the `setLength` method contract.

PR #1045, opened 2017-04-05
Use Java compiler JCTree.*-s and Symbol.*-s for Kotlin files analysis and reuse them to compile Java files.

baratynskiy:


build.xml
```diff
@@ -3,7 +3,7 @@
     <property file="resources/kotlinManifest.properties"/>
 
     <!-- Set to false to disable proguard run on kotlin-compiler.jar. Speeds up the build -->
-    <property name="shrink" value="true"/>
+    <property name="shrink" value="false"/>
```
yanex: Why? :(

compiler/javac/src/org/jetbrains/kotlin/components/JavacVirtualFileFinder.kt
```diff
@@ -0,0 +1,42 @@
...
+        if (javaClass.outerClass != null) {
+            // For nested classes we get a file of the containing class, to get the actual class file for A.B.C,
+            // we take the file for A, take its parent directory, then in this directory we look for A$B$C.class
+            file = file.parent!!.findChild(classFileName(javaClass) + ".class").sure { "Virtual file not found for $javaClass" }
+        }
+
+        return KotlinBinaryClassCache.getKotlinBinaryClass(file)
+    }
+
+    inline fun <T : Any> T?.sure(message: () -> String): T = this ?: throw AssertionError(message())
```
mglukhikh: This function already exists (anyway you will not need it after processing the previous comment)

compiler/javac/src/org/jetbrains/kotlin/components/JavacVirtualFileFinder.kt
```diff
@@ -0,0 +1,42 @@
...
+
+abstract class JavacVirtualFileFinder : VirtualFileFinder() {
+
+    override fun findKotlinClass(javaClass: JavaClass): KotlinJvmBinaryClass? {
+        var file = javaClass.computeClassId()?.let(this::findVirtualFileWithHeader) ?: return null
+
+        if (javaClass.outerClass != null) {
+            // For nested classes we get a file of the containing class, to get the actual class file for A.B.C,
+            // we take the file for A, take its parent directory, then in this directory we look for A$B$C.class
+            file = file.parent!!.findChild(classFileName(javaClass) + ".class").sure { "Virtual file not found for $javaClass" }
```
mglukhikh: Again, it's better not to repeat the same code. Extract some protected function from `VirtualFileFinder`

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.jetbrains.kotlin.wrappers.trees.JCClass
+import org.jetbrains.kotlin.wrappers.trees.JCPackage
+import java.io.Closeable
+import java.io.File
+import javax.lang.model.element.TypeElement
+import javax.tools.JavaFileManager
+import javax.tools.JavaFileObject
+import javax.tools.StandardLocation
+
+class Javac(javaFiles: Collection<File>,
```
mglukhikh: May be `JavaCompilerWrapper`?

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+    private val elements = JavacElements.instance(context)
+    private val fileObjects = fileManager.getJavaFileObjectsFromFiles(javaFiles).toJavacList()
+    private val compilationUnits: JavacList<JCTree.JCCompilationUnit> = fileObjects.map(javac::parse).toJavacList()
+
+    private val javaClasses = compilationUnits
+            .flatMap { unit -> unit.typeDecls
+                    .flatMap { JCClass(it as JCTree.JCClassDecl, trees.getPath(unit, it), this).withInnerClasses() }
+            }
+            .map { it.fqName to it }
+            .toMap()
```
mglukhikh: There is `keysToMap` in Kotlin project itself & also `associateBy` in stdlib. I'm not sure it's quite important but they may be faster (one internal container instead of two) and definitely shorter.

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+
+    private val javaPackages = compilationUnits
+            .map { FqName(it.packageName.toString()) to JCPackage(it.packageName.toString(), this) }
+            .toMap()
```
mglukhikh: See above

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+        scope.contains(AnyJavaSourceVirtualFile) -> javaPackages[fqName]
+        else -> findPackageInSymbols(fqName.asString()) ?: javaPackages[fqName]
+    }
+
+    fun findSubPackages(fqName: FqName) = symbols.packages
+                                                  .filter { (k, _) -> k.toString().startsWith("$fqName.") }
+                                                  .map { JavacPackage(it.value, this) } +
+                                          javaPackages
+                                                  .filter { it.key.isSubpackageOf(fqName) && it.key != fqName }
+                                                  .map { it.value }
```
mglukhikh: There is `filterValues`.

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    override fun close() {
+        fileManager.close()
+        javac.close()
+    }
+
+    fun findClass(fqName: FqName, scope: GlobalSearchScope = EverythingGlobalScope()) = when {
+        scope is EverythingGlobalScope -> javaClasses[fqName] ?: findClassInSymbols(fqName.asString())
+        scope.contains(AnyJavaSourceVirtualFile) -> javaClasses[fqName]
```
mglukhikh: I don't quite understand the idea with this object. As I can see, it's used in `contains` only. Who does put it into the scope?

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
```diff
@@ -0,0 +1,227 @@
...
+    private fun JavaClass.withInnerClasses(): List<JavaClass> = listOf(this) + innerClasses.flatMap { it.withInnerClasses() }
+
+}
+
+private object AnyJavaSourceVirtualFile : VirtualFile() {
+    override fun refresh(asynchronous: Boolean, recursive: Boolean, postRunnable: Runnable?) {}
+
+    override fun getLength() = 0L
+
+    override fun getFileSystem() = throw UnsupportedOperationException()
```
mglukhikh: May be (if you need this object at all) it's better to provide some message like `"Should never be called"`.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCField.kt
```diff
@@ -0,0 +1,54 @@
...
+                                             javac: Javac) : JCMember<T>(tree, treePath, containingClass, javac), JavaField {
+
+    override val name
+        get() = Name.identifier(tree.name.toString())
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = if (containingClass.isInterface) true else tree.modifiers.isStatic
```
mglukhikh: `containingClass.isInterface || tree.modifiers.isStatic`

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,118 @@
...
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
+
+    override val isFinal
+        get() = tree.modifiers.isFinal
+
+    override val visibility
+        get() = tree.modifiers.visibility
```
mglukhikh: Probably false if class is nested in interface (see JCField)

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,118 @@
...
+    override val annotations: Collection<JavaAnnotation>
+        get() = emptyList()
+
+    override fun findAnnotation(fqName: FqName) = annotations.find { it.classId?.asSingleFqName() == fqName }
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
```
mglukhikh: Probably false if class is nested in interface (see JCField)

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCConstructor.kt
```diff
@@ -0,0 +1,53 @@
...
+                                                 javac: Javac) : JCMember<T>(tree, treePath, containingClass, javac), JavaConstructor {
+
+    override val name
+        get() = Name.identifier(tree.name.toString())
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
```
mglukhikh: Constructor never can be static

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCConstructor.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override val isFinal
+        get() = tree.modifiers.isFinal
```
mglukhikh: Also it never can be overridable. Not sure should it be considered as `final` or not.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCConstructor.kt
```diff
@@ -0,0 +1,53 @@
...
+class JCConstructor<out T : JCTree.JCMethodDecl>(tree: T,
+                                                 treePath: TreePath,
+                                                 containingClass: JavaClass,
+                                                 javac: Javac) : JCMember<T>(tree, treePath, containingClass, javac), JavaConstructor {
+
+    override val name
+        get() = Name.identifier(tree.name.toString())
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
```
mglukhikh: Constructor never can be abstract

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
```diff
@@ -0,0 +1,60 @@
...
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
+
+    override val isFinal
+        get() = tree.modifiers.isFinal
+
+    override val visibility
```
mglukhikh: This can be extracted into `JCMember`

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.jetbrains.kotlin.load.java.structure.JavaType
+import org.jetbrains.kotlin.name.Name
+
+class JCMethod<out T : JCTree.JCMethodDecl>(tree: T,
+                                            treePath: TreePath,
+                                            containingClass: JavaClass,
+                                            javac: Javac) : JCMember<T>(tree, treePath, containingClass, javac), JavaMethod {
+
+    override val name
+        get() = Name.identifier(tree.name.toString())
```
mglukhikh: This can be extracted into `JCMember`

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
```
mglukhikh: Non-static methods in interface are `abstract` by default. However in Java 1.8+ they are not abstract if `default` modifier exists. Here we have complex logic, so be accurate, please.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
+
+    override val isFinal
+        get() = tree.modifiers.isFinal
```
mglukhikh: I think interface methods are never `final`

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCPackage.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import org.jetbrains.kotlin.javac.Javac
+import org.jetbrains.kotlin.load.java.structure.JavaPackage
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+
+class JCPackage(val name: String, val javac: Javac) : JavaPackage {
+
+    override val fqName: FqName
+        get() = FqName(name)
```
mglukhikh: Consider making it an initializer and remove getter

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCValueParameter.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.jetbrains.kotlin.load.java.structure.JavaValueParameter
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+
+class JCValueParameter<out T : JCTree.JCVariableDecl>(tree: T,
+                                                      treePath: TreePath,
+                                                      javac: Javac) : JCElement<T>(tree, treePath, javac), JavaValueParameter {
+
+    override val annotations: Collection<JavaAnnotation>
+        get() = emptyList()
```
mglukhikh: Value parameter in Java can have annotations

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMember.kt
```diff
@@ -0,0 +1,39 @@
...
+                                        treePath: TreePath,
+                                        override val containingClass: JavaClass,
+                                        javac: Javac) : JCElement<T>(tree, treePath, javac), JavaMember {
+
+    override val isDeprecatedInJavaDoc = false
+
+    override val annotations
+        get() = emptyList<JavaAnnotation>()
+
+    override fun findAnnotation(fqName: FqName) = annotations.firstOrNull { it.classId?.asSingleFqName() == fqName }
```
mglukhikh: Again, what do you want to find in the empty list?

udalov:
FYI `lazy { ... }` is very expensive here, in my opinion. Both in terms of memory (bytecode size and memory used by the program) and execution performance. Most of the methods in `JavaElement` and subclasses are invoked once or twice, so there's usually no need to cache anything. Please refer to the PSI-based implementation (`JavaElementImpl` and subclasses).

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/K2JVMCompiler.kt
```diff
@@ -186,6 +187,14 @@ class K2JVMCompiler : CLICompiler<K2JVMCompilerArguments>() {
                 val environment = createEnvironmentWithScriptingSupport(rootDisposable, configuration, arguments, messageCollector)
                                   ?: return COMPILATION_ERROR
 
+                if (arguments.useJavac) {
```
baratynskiy: It cannot be done because `environment` is used in this if / else statement.

compiler/javac/src/org/jetbrains/kotlin/javac/Javac.kt
baratynskiy: In this case `contains` function just checks a file extension. This is why we can use this mock object here.

compiler/javac/src/org/jetbrains/kotlin/wrappers/symbols/JavacType.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.jetbrains.kotlin.name.FqName
+import javax.lang.model.type.TypeKind
+import javax.lang.model.type.TypeMirror
+
+open class JavacType<out T : TypeMirror>(val typeMirror: T,
+                                         val javac: Javac) : JavaType, JavaAnnotationOwner {
+
+    companion object {
+        fun <TM : TypeMirror> create(t: TM, javac: Javac) = when {
+            t.kind.isPrimitive || t.toString() == "void" -> JavacPrimitiveType(t, javac)
```
baratynskiy: `PrimitiveType` includes `boolean`, `byte`, `short`, `int`, `long`, `char`, `float`, and `double`.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,118 @@
...
+            .joinToString(separator = ".") {
+                (it as? JCTree.JCCompilationUnit)?.packageName?.toString()
+                ?: (it as JCTree.JCClassDecl).name
+            }
+            .let(::FqName)
+
+    override val supertypes
+        get() = arrayListOf<JavaClassifierType>().apply {
+            fun JCTree.mapToJavaClassifierType() = when {
+                this is JCTree.JCExpression -> JCClassifierType(this, TreePath(treePath, this), javac)
```
baratynskiy: For example, `List<String>` and just `List`. The first is `JCTree.JCTypeApply`, while the second is `JCTree.JCExpression`.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
baratynskiy: It cannot be done because `JCMember` type parameter is `<out T : JCTree>`. JCTree doesn't have a `name` field. At the same time `JCTree.JCMethodDecl` and `JCTree.JCVariableDecl` don't have a common ancestor class except for `JCTree`.

compiler/javac/src/org/jetbrains/kotlin/wrappers/symbols/JavacReferenceAnnotationArgument.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.jetbrains.kotlin.javac.Javac
+import org.jetbrains.kotlin.load.java.structure.JavaEnumValueAnnotationArgument
+import org.jetbrains.kotlin.name.FqName
+import javax.lang.model.element.VariableElement
+
+class JavacReferenceAnnotationArgument(val element: VariableElement,
+                                       javac: Javac) : JavacAnnotationArgument(FqName(element.simpleName.toString()), javac),
+        JavaEnumValueAnnotationArgument {
+
+    override fun resolve() = JavacField(element, javac)
```
baratynskiy: There is no mistake here. Maybe there is a mistake in the NetBeans plugin itself.

compiler/javac/src/org/jetbrains/kotlin/wrappers/symbols/JavacTypeParameter.kt
```diff
@@ -0,0 +1,35 @@
...
+
+class JavacTypeParameter<out T : TypeParameterElement>(element: T,
+                                                       javac: Javac) : JavacClassifier<T>(element, javac), JavaTypeParameter {
+
+    override val name
+        get() = SpecialNames.safeIdentifier(element.simpleName.toString())
+
+    override val upperBounds: Collection<JavaClassifierType>
+        get() = listOf(JavacClassifierType((element.asType() as TypeVariable).upperBound, javac))
+
```
baratynskiy: We don't need these methods to be overridden because we inherit them from `JavacElement`. In NB plugin we should do it because there is no `ElementHandle` for type parameters.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,118 @@
...
+    override val name
+        get() = SpecialNames.safeIdentifier(tree.simpleName.toString())
+
+    override val annotations: Collection<JavaAnnotation>
+        get() = emptyList()
+
+    override fun findAnnotation(fqName: FqName) = annotations.find { it.classId?.asSingleFqName() == fqName }
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
```
baratynskiy: There is another property for interfaces called `isInterface`.

compiler/javac/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
baratynskiy: It will return `false` for an interface method.

compiler/javac-wrapper/src/org/jetbrains/kotlin/components/JavacBasedClassFinder.kt
```diff
@@ -0,0 +1,51 @@
...
+        javac = JavacWrapper.getInstance(proj)
+        CodeAnalyzerInitializer.getInstance(proj).initialize(trace, codeAnalyzer.moduleDescriptor, codeAnalyzer)
+    }
+
+    override fun findClass(classId: ClassId) = javac.findClass(classId.asSingleFqName(), javaSearchScope)
+
+    override fun findPackage(fqName: FqName) = javac.findPackage(fqName, javaSearchScope)
+
+    override fun knownClassNamesInPackage(packageFqName: FqName) = javac.findClassesFromPackage(packageFqName)
+            .mapNotNull {
```
mglukhikh: Here you can use `mapNotNullTo` without `toSet` at the end, it may be a bit faster

compiler/javac-wrapper/src/org/jetbrains/kotlin/components/JavacBasedJavaResolverCache.kt
```diff
@@ -0,0 +1,40 @@
...
+
+class JavacBasedJavaResolverCache(resolveSession: ResolveSession) : AbstractJavaResolverCache(resolveSession) {
+
+    override fun recordMethod(method: JavaMethod, descriptor: SimpleFunctionDescriptor) {}
+
+    override fun recordConstructor(element: JavaElement, descriptor: ConstructorDescriptor) {}
+
+    override fun recordField(field: JavaField, descriptor: PropertyDescriptor) {}
+
+    override fun recordClass(javaClass: JavaClass, descriptor: ClassDescriptor) {}
```
mglukhikh: I'm not sure these four methods should be empty. This place should be discussed; I'm afraid we can lose some IDE features here. Should be checked attentively who uses the slices `CONSTRUCTOR` / `VARIABLE` / `CLASS` / `FUNCTION`

compiler/javac-wrapper/src/org/jetbrains/kotlin/components/JavacBasedPropertyInitializerEvaluator.kt
```diff
@@ -0,0 +1,66 @@
...
+    private fun JCField<*>.constant(descriptor: PropertyDescriptor) = value?.let {
+        if (isCompileTimeConstant() && it is JCTree.JCLiteral) {
+            ConstantValueFactory(descriptor.builtIns).createConstantValue(it.value)
+        } else null
+    }
+
+    private fun JavacField<*>.isCompileTimeConstant() = value?.let {
+        val typeMirror = type.typeMirror
+
+        (typeMirror.kind.isPrimitive || typeMirror.toString() == "java.lang.String")
```
mglukhikh: It would be better to avoid `toString()` call here, if possible. Does this `typeMirror` have something like `name` or `fqName`?

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/symbols/JavacClass.kt
```diff
@@ -0,0 +1,106 @@
...
+    override val isInterface
+        get() = element.kind == ElementKind.INTERFACE
+
+    override val isAnnotationType
+        get() = element.kind == ElementKind.ANNOTATION_TYPE
+
+    override val isEnum
+        get() = element.kind == ElementKind.ENUM
+
+    override val lightClassOriginKind = null
```
mglukhikh: Possibly it can break something in IDE

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,124 @@
...
+    override fun findAnnotation(fqName: FqName) = annotations.find { it.classId?.asSingleFqName() == fqName }
+
+    override val isDeprecatedInJavaDoc
+        get() = findAnnotation(FqName("java.lang.Deprecated")) != null
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
+
+    override val isStatic
+        get() = tree.modifiers.isStatic
```
mglukhikh: What about interface member classes? (they're always static as I know)

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    override val isFinal
+        get() = tree.modifiers.isFinal
+
+    override val visibility
+        get() = tree.modifiers.visibility
```
mglukhikh: What about interface member classes? (they're always public as I know)

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCMethod.kt
```diff
@@ -0,0 +1,59 @@
...
+class JCMethod<out T : JCTree.JCMethodDecl>(tree: T,
+                                            treePath: TreePath,
+                                            containingClass: JavaClass,
+                                            javac: JavacWrapper) : JCMember<T>(tree, treePath, containingClass, javac), JavaMethod {
+
+    override val name
+        get() = Name.identifier(tree.name.toString())
+
+    override val isAbstract
+        get() = tree.modifiers.isAbstract
```
mglukhikh: What about interface members? (not sure it's important)

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt
```diff
@@ -161,6 +162,14 @@ object KotlinToJVMBytecodeCompiler {
                 ProgressIndicatorAndCompilationCanceledStatus.checkCanceled()
                 writeOutput(state.configuration, state.factory, null)
             }
+
+            if (chunk.size == 1) {
+                val javacWrapper = JavacWrapper.getInstance(environment.project)
+                if (projectConfiguration[JVMConfigurationKeys.USE_JAVAC]!!) {
```
udalov: @baratynskiy Please use `getBoolean(...)` instead of `get(...)!!`

compiler/testData/compileJavaAgainstKotlin/method/Any.kt
```diff
@@ -1,4 +1,4 @@
 package test
 
 // extra parameter is to preserve generic signature
-fun anyany(a: Any, ignore: java.util.List<String>) = a
+fun anyany(a: kotlin.Any, ignore: java.util.List<String>) = a
```
mglukhikh: This question looks still actual

compiler/testData/compileJavaAgainstKotlin/method/Int.kt
```diff
@@ -1,3 +1,3 @@
 package test
 
-fun lll(a: Int) = a
+fun lll(a: kotlin.Int) = a
```
mglukhikh: Here we have `kotlin.` added too

compiler/testData/compileJavaAgainstKotlin/method/IntArray.kt
```diff
@@ -1,4 +1,4 @@
 package test
 
 // extra parameter is to make sure generic signature is not erased
-fun doNothing(array: IntArray, ignore: java.util.List<String>) = array
+fun doNothing(array: kotlin.IntArray, ignore: java.util.List<String>) = array
```
mglukhikh: And here we have `kotlin.` too

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/utils.kt
```diff
@@ -68,19 +68,19 @@ fun TreePath.resolve(javac: JavacWrapper): Pair<FqName, JavaClass?> {
...
             .map {
-                javac.findClassesFromPackage(it.qualifiedIdentifier.toString()
-                                                     .substringBeforeLast(".").let(::FqName))
-                        .find { it.name.identifier == name }
-                        ?.let { it.fqName!! to it }
+                javac.findClass("${it.qualifiedIdentifier.toString().substringBefore("*")}$name"
+                                .let(::FqName))
             }
             .firstOrNull()
+            ?.let { it.fqName!! to it }
```
mglukhikh: This commit is no more actual (superceded by another one)

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/utils.kt
```diff
@@ -72,37 +70,53 @@ private fun resolveImport(name: String,
...
+                else -> emptyList<JCTree.JCTypeParameter>()
+            }
+        }
+        .find { it.toString().substringBefore(" ") == leaf.toString() }
+        ?.let {
+            JCTypeParameter(it,
+                            javac.getTreePath(it, compilationUnit),
+                            javac)
+        }
+
```
mglukhikh: This one is also not actual

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/utils.kt
```diff
@@ -80,12 +80,25 @@ private fun resolveImport(name: String,
...
                           compilationUnit: JCTree.JCCompilationUnit,
                           javac: JavacWrapper): JavaClassifier? {
     // try to find an inner class
-    (parentPath.find { it is JCTree.JCClassDecl } as? JCTree.JCClassDecl)?.let { outerClass ->
-        javac.findClass(FqName("${compilationUnit.packageName}.${outerClass.name}.$name"))
+    (parentPath.find { it is JCTree.JCClassDecl
+                       && it.extending != leaf
+                       && !it.implementing.contains(leaf) } as? JCTree.JCClassDecl)?.let { outerClass ->
+        javac.findClass(FqName("${compilationUnit.packageName}.${outerClass.name}"))
+                ?.tryToResolveInner(name, javac)
```
mglukhikh: This commit is probably also not actual

compiler/javac-wrapper/src/org/jetbrains/kotlin/javac/KotlinClassifiersCache.kt
```diff
@@ -18,75 +18,87 @@ package org.jetbrains.kotlin.javac
...
 
     private val classifiers = hashMapOf<FqName, JavaClass>()
 
     fun getKotlinClassifier(fqName: FqName) = classifiers[fqName] ?: createClassifier(fqName)
 
-    private fun createClassifier(fqName: FqName) = object : JavaClass {
-        override val isAbstract: Boolean
-            get() = false
+    private fun createClassifier(fqName: FqName): JavaClass? {
+        if (!kotlinClasses.contains(fqName)) return null
```
mglukhikh: This is no more actual change

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/utils.kt
```diff
@@ -119,6 +115,12 @@ private fun TreePath.tryToResolve(name: String,
     return javac.getKotlinClassifier(FqName(name))
 }
 
+private fun TreePath.findOuterClass(javac: JavacWrapper) = filterIsInstance<JCTree.JCClassDecl>()
+        .filter { it.extending != leaf && !it.implementing.contains(leaf) }
+        .reversed()
+        .joinToString(separator = ".", prefix = "${compilationUnit.packageName}.") { it.simpleName }
+        .let { javac.findClass(FqName(it)) }
+
```
mglukhikh: This commit looks also not actual

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCClass.kt
```diff
@@ -48,7 +48,7 @@ class JCClass<out T : JCTree.JCClassDecl>(tree: T,
         get() = tree.modifiers.isAbstract
 
     override val isStatic
-        get() = outerClass?.isInterface ?: tree.modifiers.isStatic
+        get() = if (outerClass?.isInterface ?: false) true else tree.modifiers.isStatic
```
mglukhikh: It's better to write here

```
(outerClass?.isInterface ?: false) || tree.modifiers.isStatic
```

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCType.kt
```diff
@@ -0,0 +1,55 @@
...
+            is JCTree.JCArrayTypeTree -> JCArrayType(tree, TreePath(treePath, tree), javac)
+            is JCTree.JCWildcard -> JCWildcardType(tree, TreePath(treePath, tree), javac)
+            is JCTree.JCTypeApply -> JCClassifierTypeWithTypeArgument(tree, TreePath(treePath, tree), javac)
+            is JCTree.JCExpression -> JCClassifierType(tree, TreePath(treePath, tree), javac)
+            else -> throw UnsupportedOperationException("Unsupported type: $tree")
+        }
+    }
+
+    override val annotations
+        get() = emptyList<JCAnnotation>()
```
mglukhikh: Are you sure this list should be empty? In `JavaTypeImpl` it's not so

compiler/javac-wrapper/src/org/jetbrains/kotlin/wrappers/trees/JCClassifierType.kt
```diff
@@ -0,0 +1,74 @@
...
+        get() = emptyList()
+
+    override val isRaw
+        get() = (classifier as? MockKotlinClassifier)?.hasTypeParameters
+                ?: (classifier as? JavaClass)?.typeParameters?.isNotEmpty()
+                ?: false
+
+}
+
+class JCClassifierTypeWithTypeArgument<out T : JCTree.JCTypeApply>(tree: T,
```
mglukhikh: May be it's not the best name. For me it's more or less obvious, that `JCClassifierTypeWithTypeArgument` is an inheritor of `JCClassifierType`, but in fact it's not so. I'd change one of these two names.

compiler/javac-wrapper/src/org/jetbrains/kotlin/components/JavacBasedJavaResolverCache.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.SimpleFunctionDescriptor
+import org.jetbrains.kotlin.load.java.components.AbstractJavaResolverCache
+import org.jetbrains.kotlin.load.java.structure.JavaClass
+import org.jetbrains.kotlin.load.java.structure.JavaElement
+import org.jetbrains.kotlin.load.java.structure.JavaField
+import org.jetbrains.kotlin.load.java.structure.JavaMethod
+import org.jetbrains.kotlin.resolve.lazy.ResolveSession
+
+class JavacBasedJavaResolverCache(resolveSession: ResolveSession) : AbstractJavaResolverCache(resolveSession) {
```
mglukhikh: May be we should name it like `StubJavaResolverCache`. The thing here is obviously not javac based.

udalov:
Please move packages `components` and `wrappers` in the module `javac-wrapper` from `org.jetbrains.kotlin` inside `org.jetbrains.kotlin.javac`

build.xml
```diff
@@ -375,6 +382,7 @@
                 <zipfileset src="${dependencies}/jline.jar"/>
                 <zipfileset src="${dependencies}/javaslang-2.0.6.jar"/>
                 <zipfileset src="${protobuf.jar}"/>
+                <zipfileset src="${tools.jar}"/>
```
mglukhikh: This string must be removed due to licensing problems

mglukhikh:
Everything except the last commit (controversial one about JPS plugin) was merged into master. Hope we do not break anything. Thanks for everybody who helped!

PR #1047, opened 2017-04-10
Completion of keywords (like `constructor`) doesn't insert parentheses/braces if they're already there

cypressious:
Fixes #KT-15603
Fixes #KT-13524

### DIFF
--- a/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
+++ b/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2015 JetBrains s.r.o.
+ * Copyright 2010-2017 JetBrains s.r.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import com.intellij.openapi.editor.Document
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiWhiteSpace
 import com.intellij.psi.codeStyle.CodeStyleManager
 import org.jetbrains.kotlin.idea.completion.KeywordLookupObject
 import org.jetbrains.kotlin.idea.core.moveCaret
@@ -33,6 +34,8 @@ import org.jetbrains.kotlin.psi.KtBlockStringTemplateEntry
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtPsiFactory
 import org.jetbrains.kotlin.psi.psiUtil.canPlaceAfterSimpleNameEntry
+import org.jetbrains.kotlin.psi.psiUtil.getNextSiblingIgnoringWhitespace
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
 
 fun surroundWithBracesIfInStringTemplate(context: InsertionContext): Boolean {
     val startOffset = context.startOffset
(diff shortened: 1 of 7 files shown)


idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
```diff
@@ -144,8 +146,14 @@ fun createKeywordConstructLookupElement(
                     val beforeCaret = tailBeforeCaret.indentLinesAfterFirst(newIndent)
                     val afterCaret = tailAfterCaret.indentLinesAfterFirst(newIndent)
 
-                    insertionContext.document.insertString(offset, beforeCaret + afterCaret)
-                    insertionContext.editor.moveCaret(offset + beforeCaret.length)
+                    val element = insertionContext.file.findElementAt(offset)?.getNextSiblingIgnoringWhitespaceAndComments(true)
```
goodwinnk: Why getNextSiblingIgnoringWhitespaceAndComments() is needed? I expect that it is for 'for' keyword completion, but at least one test needed for it.
I also think that there shouldn't be comments skipping.

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
```diff
@@ -144,8 +146,14 @@ fun createKeywordConstructLookupElement(
...
+
+                    if (element != null && beforeCaret.trimStart().startsWith(insertionContext.document.getText(TextRange.from(element.startOffset, 1)))) {
```
goodwinnk: I think it will be more safe to use == instead of startsWith()

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
goodwinnk: The following a bit artificial snippet can be used as an example of unexpected behaviour:

```
fun other() {
  try<caret>

  // Some
  {
    println()
  }()
}
```

goodwinnk: More natural example:

```
fun other(a: Any) {
  if<caret>
  (a as Int).let { println(a + 12) }
}
```

goodwinnk:
Hello @cypressious! Thank you for the contribution! 

While it's nice to have a common solution for all affected keywords, I think the behavior should be restricted a to avoid unexpected behavior. 

Will you update the request?

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
cypressious: The reason I used `startsWith()` is that some keywords insert more than a single character, e.g. `init` where `beforeCaret.trimStart()` will be equal to "{\n".

See https://github.com/JetBrains/kotlin/blob/b5db50f42996c09de12a1bddb98adb960a32bc01/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KeywordCompletion.kt#L72-L72

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
cypressious: @goodwinnk Thanks for the review.

`getNextSiblingIgnoringWhitespaceAndComments()` is needed because otherwise the following case will be handled wrongly:

```kotlin
class Foo cons<caret> () //note the space before the primary constructor
```

I agree that comments shouldn't be skipped. For that I could introduce a new extension function `getNextSiblingIgnoringWhitespace()` and use it instead.

Regarding your examples where my logic would lead to arguably wrong behavior: How about I restrict my logic to cases where there isn't a line break between the keyword and the brace/paran?

So this 

```kotlin
fun other(a: Any) {
  if<caret>
  (a as Int).let { println(a + 12) }
}
```

will become

```kotlin
fun other(a: Any) {
  if (<caret>)
  (a as Int).let { println(a + 12) }
}
```

but this

```kotlin
fun other(a: Any) {
  whil<caret> ()
}
```

will become
```kotlin
fun other(a: Any) {
  while (<caret>)
}
```

cypressious:
@goodwinnk I commented on your review. I think this PR shouldn't be closed yet, should it?

goodwinnk:
Sorry, I thought it can be reopened from both sides :)

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
goodwinnk: I think proposed fixes are great.

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/handlers/handlerUtils.kt
goodwinnk: I see, thank you for the explanation.

cypressious:
@goodwinnk I have updated the PR.

goodwinnk:
I merged it to master with a very small change in `handlerUtils.kt`, we have a code guide rule to format if-else as
```
if (condition) {
}
else {
}
```

Thank you!

PR #1048, opened 2017-04-13
KT-16298: Fix plugin trying to remove cache already removed by clean

JLLeitschuh:
Not totally sure if this will fix the issue (or even compile). I think this should ensure that the cache isn't flushed if the directory no longer exists.

(Not sure where you can find a copy of the JDK 1.6 these days for OSX (a link on the readme would be really useful)).


tsvtkv:
@JLLeitschuh thank you for your contribution! 

The change is going to be included in 1.1.3.
The PR is merged manually, so I'm closing it now.

PR #1049, opened 2017-04-14
Deprecated setting was used for import optimization on the fly.

Alefas:
New IDEA per-project setting was completely ignored by Kotlin.


yole:
Thanks!

PR #1050, opened 2017-04-14
Smart convert to body enter handler feature.

Alefas:
smartenterkotlin


idea/src/org/jetbrains/kotlin/idea/editor/KotlinEditorOptions.java
```diff
@@ -33,6 +33,7 @@
 public class KotlinEditorOptions implements PersistentStateComponent<KotlinEditorOptions> {
     private boolean donTShowConversionDialog = false;
     private boolean enableJavaToKotlinConversion = true;
+    private boolean enableSmartEditing = true;
```
yole: I don't like the idea of adding a preference for this behavior. First of all, the name of the preference says nothing at all to the user. Second, if anyone becomes annoyed by the behavior, it's quite hard for them to discover how to change it.

PR #1051, opened 2017-04-16
Update usage of deprecated sourceMappingURL format

ajalt:
The usage of `//@` in sourceMappingURL pragmas was deprecated in favor of `//#`. 
The deprecation happened in 2013, and was motivated by incompatibilities in IE.

The Chromium and Mozilla issues tracking the change:

https://bugs.chromium.org/p/chromium/issues/detail?id=239660
https://bugzilla.mozilla.org/show_bug.cgi?id=870361

And the Mozilla documentation on the warning that is issued as a result of using the deprecated syntax:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_source_map_pragma


bashor:
Thanks, @ajalt!

PR #1052, opened 2017-04-18
KT-16786 Add Intention to add "open" modifier to a non-private method or property in an open class

shiraji:
#KT-16786 Fixed

https://youtrack.jetbrains.com/issue/KT-16786


.idea/saveactions_settings.xml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="SaveActionSettings">
+    <option name="actions">
+      <set>
+        <option value="organizeImports" />
+        <option value="reformat" />
+        <option value="reformatChangedCode" />
+      </set>
+    </option>
+  </component>
+</project>
```
mglukhikh: Not needed

idea/src/org/jetbrains/kotlin/idea/intentions/AddOpenModifierIntention.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class AddOpenModifierIntention : SelfTargetingIntention<KtDeclaration>(KtDeclaration::class.java, "Make open"), LowPriorityAction {
+    override fun isApplicableTo(element: KtDeclaration, caretOffset: Int): Boolean {
+        if (element.hasModifier(KtTokens.OPEN_KEYWORD) || element.hasModifier(KtTokens.PRIVATE_KEYWORD)) return false
+        when (element) {
+            is KtProperty -> if (element.isLocal) return false
+            is KtFunction -> if (element.isLocal) return false
+        }
+        val ktClass = element.getNonStrictParentOfType<KtClass>() ?: return false
+        return ktClass.hasModifier(KtTokens.OPEN_KEYWORD)
```
mglukhikh: Things are not so easy here. First of all, there is `KtDeclaration.implicitModality()`, please use it. For example, interface functions with body are implicitly open and do not require explicit open modifier. Also, you need `containingClassOrObject` and not `getNonStrictParentOfType<KtClass>`, you are risking to miss e.g. companion object in the middle this way. Remember also that object / interface members should never have `open` added, as well as top-level guys. And at last, class can have `open` members if it's `open`, `abstract` or `sealed` itself, I'd not limit it to just open classes as described in an issue.

mglukhikh: Also, element can easily be `KtClassOrObject`, please consider this case. The easiest way here is to limit this intention only to properties and functions, this way you need `KtCallableDeclaration`. Remember also nice cases like `KtParameter` or `KtDestructuringDeclarationEntry`, both of them should never be open.

mglukhikh:
Also I think this intention should not be suggested for abstract members, either explicitly or implicitly, because making them `open` will likely make code red.

shiraji:
Could you review this PR, again?

mglukhikh:
Yes, I will merge it today, thanks :)

mglukhikh:
Merged manually. Intention forbidden for everything not `KtProperty` or `KtNamedFunction`, e.g. for `KtConstructor`. Intention allowed for `enum class` members.

PR #1054, opened 2017-04-24
KT-17053 Add inspection to detect use of single callable reference as a lambda body

shiraji:
#KT-17053 Fixed

Issue: https://youtrack.jetbrains.com/issue/KT-17053
Related issue: https://youtrack.jetbrains.com/issue/KT-17051


mglukhikh:
So thanks, I merged it into master, with some refactorings. Probably I should mention that `IntentionBasedInspection` is no more recommended as base class for inspections. You may look at the final commit as possible example: https://github.com/JetBrains/kotlin/commit/0eceef1519653d37cf2f6177da5d5aa059375436

shiraji:
Got it. Thanks! I will use `AbstractKotlinInspection` in next time.

PR #1056, opened 2017-04-24
Implement intetion to convert string concatenation to raw string

cypressious:
Fixes #KT-17503

Not sure if tests are required. I'm basicly using two existing intentions that are already heavily tested.


mglukhikh:
Thanks, I've merged it. Added one test from the issue just to be sure.

PR #1057, opened 2017-04-24
Intention to name anonymous parameter

nd:
#KT-17191 Fixed


mglukhikh:
Thank you. I will extend this intention, because there are some cases that are not considered in your code. Probably will add it into master today.

mglukhikh:
So thanks, it was finally merged :)

PR #1058, opened 2017-04-26
Do not suggest 'Convert to secondary constructor' for annotation classes

nd:
#KT-17408 Fixed


mglukhikh:
Thank you, will merge it today.

PR #1060, opened 2017-04-26
optimization: stateless properties

uujava:
Please consider proposed optimization:
    Delegate to statically initialized fields if delegate object doesn't have state associated with an object instance.

That might be naive implementation and it misses test so far. 

Please check https://github.com/uujava/kotlin-stateless-properties for example and more details.


uujava:
feature request in youtrack: KT-17637

uujava:
Just wonder If anybody could review this request or KT-17637. Is this feasible approach? I could rework per review if needed.

dnpetrov:
Closing as obsolete.

PR #1061, opened 2017-04-28
Change visibility for annotation class constructor

shiraji:
https://youtrack.jetbrains.com/issue/KT-17599
https://youtrack.jetbrains.com/issue/KT-17600


mglukhikh:
Thanks, merged into master

PR #1062, opened 2017-04-28
KT-16419 - expand Lock#withLock to semaphore

jonnyzzz:


ilya-g:
It's interesting that `Lock.withLock()` acquires the lock uninterruptibly and on the other hand `Semaphore.withAcquired` acquires the semaphore interruptibly. Is there a reason for this difference besides the default method is used in each case?

jonnyzzz:
My logic was to mimic same behavior as in Java API. I'll update javadoc to
state that explicitly.

Do you think it worth to add the second method or default parameter to
control interruptability explicitly?

PR #1063, opened 2017-04-29
Convert 'if' with 'is' check to 'as?' with safe call

nd:
#KT-17054 Fixed


mglukhikh:
Just to document: I'm investigating it now. We have two very similar intentions (IfThenToSafeAccess, IfThenToElvis) and I'm trying to make them working in the similar way.

mglukhikh:
Thank you, I've finally added it into master (with the set of commits making IfThenToSafeAccess / IfThenToElvis working in the same manner).

PR #1064, opened 2017-05-01
Detect redundant 'is' check

nd:
#KT-14187 Fixed

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/Errors.java
+++ b/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/Errors.java
@@ -825,6 +825,8 @@ public interface Errors {
     DiagnosticFactory0<KtExpression> ALWAYS_NULL = DiagnosticFactory0.create(WARNING);
 
     DiagnosticFactory0<KtNullableType> USELESS_NULLABLE_CHECK = DiagnosticFactory0.create(WARNING, NULLABLE_TYPE);
+    DiagnosticFactory1<KtElement, Boolean> USELESS_IS_CHECK = DiagnosticFactory1.create(WARNING);
+
 
     // Properties / locals
 
--- a/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java
+++ b/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java
@@ -395,6 +395,7 @@ public class DefaultErrorMessages {
         MAP.put(IS_ENUM_ENTRY, "'is' over enum entry is not allowed, use comparison instead");
         MAP.put(ENUM_ENTRY_AS_TYPE, "Use of enum entry names as types is not allowed, use enum type instead");
         MAP.put(USELESS_NULLABLE_CHECK, "Non-null type is checked for instance of nullable type");
+        MAP.put(USELESS_IS_CHECK, "Check for instance is always ''{0}''", TO_STRING);
         MAP.put(WRONG_SETTER_PARAMETER_TYPE, "Setter parameter type must be equal to the type of the property, i.e. ''{0}''", RENDER_TYPE, RENDER_TYPE);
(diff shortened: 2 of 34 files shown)


compiler/frontend/src/org/jetbrains/kotlin/diagnostics/Errors.java
```diff
@@ -824,6 +824,8 @@
     DiagnosticFactory0<KtExpression> ALWAYS_NULL = DiagnosticFactory0.create(WARNING);
 
     DiagnosticFactory0<KtNullableType> USELESS_NULLABLE_CHECK = DiagnosticFactory0.create(WARNING, NULLABLE_TYPE);
+    DiagnosticFactory0<KtElement> USELESS_IS_CHECK = DiagnosticFactory0.create(WARNING);
+    DiagnosticFactory0<KtElement> USELESS_IS_NOT_CHECK = DiagnosticFactory0.create(WARNING);
```
zarechenskiy: I'd propose to use one diagnostic with parameter: 
```
DiagnosticFactory1<KtElement, Boolean> USELESS_IS_CHECK = DiagnosticFactory1.create(WARNING);
```

compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java
```diff
@@ -394,6 +394,8 @@ public static DiagnosticRenderer getRendererForDiagnostic(@NotNull Diagnostic di
         MAP.put(IS_ENUM_ENTRY, "'is' over enum entry is not allowed, use comparison instead");
         MAP.put(ENUM_ENTRY_AS_TYPE, "Use of enum entry names as types is not allowed, use enum type instead");
         MAP.put(USELESS_NULLABLE_CHECK, "Non-null type is checked for instance of nullable type");
+        MAP.put(USELESS_IS_CHECK, "'is' check is always true");
+        MAP.put(USELESS_IS_NOT_CHECK, "'!is' check is always false");
```
zarechenskiy: Then here:
```
MAP.put(USELESS_IS_CHECK, "Check for instance is always ''{0}''", TO_STRING);
```

compiler/frontend/src/org/jetbrains/kotlin/types/expressions/PatternMatchingTypingVisitor.kt
```diff
@@ -439,6 +443,13 @@ class PatternMatchingTypingVisitor internal constructor(facade: ExpressionTyping
             context.trace.report(Errors.USELESS_NULLABLE_CHECK.on(element as KtNullableType))
         }
         checkTypeCompatibility(context, targetType, subjectType, typeReferenceAfterIs)
+        if (!subjectType.containsError() && !targetType.containsError() && subjectType.isSubtypeOf(targetType)) {
```
zarechenskiy: Note that this isn't correct for platform types (see https://kotlinlang.org/docs/reference/java-interop.html#null-safety-and-platform-types)

Currently there will be warning for case like this:
```
class A {
  static A create() { return null; }
}
--
fun test() {
  A.create() is A // false warning 
}
```

zarechenskiy: Also this check doesn't use data flow info:
```
fun test(x: Any) {
  x as String
  x is String 
}
```

zarechenskiy: Please check method `castIsUseless` or `isKnownToBeNotNull` for an example of `DataFlowAnalyzer`

zarechenskiy:
Thanks for your PR!

Please correct comments and add test to check that there aren't any warnings for error types:
```
fun foo() = unresolved
fun test() {
  foo() is Any?
} 
```

nd:
@zarechenskiy please take a look, I hope I got the dataflow/platform types right

zarechenskiy:
LGTM, thanks for your contribution! I'll merge it soon manually.

PR #1065, opened 2017-05-01
Fix List.takeLast for lists without RandomAccess

langara:


ilya-g:
Thanks for spotting this. I'll merge this after all tests have run.

PR #1067, opened 2017-05-02
* exclude showing of KotlinWithLibraryConfigurators in Kobalt based p…

dmitry-zhuravlev:
Kobalt have its own Kotlin libs configurator: https://github.com/cbeust/kobalt-intellij-plugin/blob/master/src/com/beust/kobalt/intellij/configurator/KotlinKobaltProjectConfigurator.kt
Doesn't need to show up Java/JS default configurators to not confuse users.


goodwinnk:
It looks with the change in this line https://github.com/JetBrains/kotlin/blob/f6b805a973cd35eef695efdc8f985b7befc75371/idea/idea-jvm/src/org/jetbrains/kotlin/idea/configuration/KotlinWithLibraryConfigurator.kt#L330, this is all that left from this pull request https://github.com/JetBrains/kotlin/commit/e36b1037bbbabf84cba5083772f7e0eabb0c0e29. 

@dmitry-zhuravlev Can you please confirm that it will be enough?

goodwinnk:
I have pushed the commit to master: https://github.com/JetBrains/kotlin/commit/ab259c2d6faca987009e6fccd6b43e71b118977e.

PR #1068, opened 2017-05-02
add inspection for calls to a data class' copy method without named arguments.

cypressious:
Fixes #KT-17660

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/CopyWithoutNamedArgumentsInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/CopyWithoutNamedArgumentsInspection.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


cypressious:
@mglukhikh I haven't found a solution for hiding the duplicate intention without using `IntentionBasedInspection`, yet.

mglukhikh:
OK, I'll try to find the solution myself :)

idea/src/org/jetbrains/kotlin/idea/inspections/CopyWithoutNamedArgumentsInspection.kt
```diff
@@ -0,0 +1,63 @@
...
+class CopyWithoutNamedArgumentsInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitCallExpression(expression: KtCallExpression) {
+                super.visitCallExpression(expression)
+
+                val reference = expression.referenceExpression() as? KtNameReferenceExpression ?: return
+                if (reference.getReferencedNameAsName() != DataClassDescriptorResolver.COPY_METHOD_NAME) return
+                if (expression.valueArguments.none { !it.isNamed() }) return
```
mg6maciej: `.all { it.isNamed() }` is easier to read.

idea/testData/inspections/copyWithoutNamedArguments/test.kt
```diff
@@ -0,0 +1,15 @@
+data class Foo(val a: String) {
+    fun copy(i: Int) {}
+}
+
+class Foo2() {
+    fun copy(i: Int) {}
+}
+
```
mg6maciej: I'd add an example that clearly shows all should be named, like this:
```
data class SomeName(a: Int, b: Int, c: String)
f.copy(2, c = "")
```

mg6maciej: And maybe another where `copy` is called on `this`:
```
fun SomeName.func() = copy(0)
```

mglukhikh:
I'll improve this inspection and merge it into master myself.

PR #1069, opened 2017-05-03
KT-16851 Quickfix adding label to unallowed 'return' in closures

shiraji:
#KT-16851 Fixed

https://youtrack.jetbrains.com/issue/KT-16851


mglukhikh:
Merged into master, thank you.

PR #1072, opened 2017-05-06
Detect redundant semicolon before empty import list

nd:
#KT-16577 Fixed


yole:
Thanks for your contribution!

PR #1073, opened 2017-05-07
Stricter type checks in 'convert to range' intention

nd:
#KT-17079 #KT-17762 Fixed

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertTwoComparisonsToRangeCheckIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertTwoComparisonsToRangeCheckIntention.kt
@@ -24,9 +24,15 @@ import org.jetbrains.kotlin.idea.inspections.IntentionBasedInspection
 import org.jetbrains.kotlin.idea.intentions.branchedTransformations.evaluatesTo
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.DescriptorUtils
 import org.jetbrains.kotlin.resolve.calls.callUtil.getType
 import org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator
 import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.builtIns
+import org.jetbrains.kotlin.types.typeUtil.immediateSupertypes
+import org.jetbrains.kotlin.types.typeUtil.isConstructedFromClassWithGivenFqName
+import org.jetbrains.kotlin.types.typeUtil.isPrimitiveNumberType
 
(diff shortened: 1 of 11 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertTwoComparisonsToRangeCheckIntention.kt
```diff
@@ -124,21 +128,71 @@ class ConvertTwoComparisonsToRangeCheckIntention : SelfTargetingOffsetIndependen
         // To avoid possible side effects
         if (!min.isSimple() || !max.isSimple()) return null
 
+        val valContext = value.analyze()
```
yole: You don't need to call `analyze()` multiple times; calling it once will return a `BindingContext` with information about all the expressions.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertTwoComparisonsToRangeCheckIntention.kt
```diff
@@ -124,21 +128,71 @@ class ConvertTwoComparisonsToRangeCheckIntention : SelfTargetingOffsetIndependen
...
     }
 
-    private fun KotlinType?.isValidTypeForIncrementDecrementByOne(): Boolean {
-        this ?: return false
+    private fun getDoubleConstant(intExpr: KtExpression, type: KotlinType, context: BindingContext): String? {
+        val intConst = ConstantExpressionEvaluator.getConstant(intExpr, context)?.getValue(type) ?: return null
+        return (intConst as? Number)?.toDouble()?.toString()
+    }
+
+    private fun KotlinType.isComparable(): Boolean {
```
yole: `DescriptorUtils.isSubtypeOfClass(this, expression.builtIns.getComparable())`

idea/testData/intentions/convertTwoComparisonsToRangeCheck/compareToDifferentTypes.kt
```diff
@@ -0,0 +1,9 @@
+// IS_APPLICABLE: false
+// WITH_RUNTIME
+class A(val _value: Int) {
+    operator fun compareTo(other: Int) = _value.compareTo(other)
+}
+
+fun test(a: A): Boolean {
+    return a >= 0 && a <= 100<caret>
```
yole: My suggested change to KotlinType.isComparable() will break this test, but I think it's OK because this test should be red, and we make no guarantees regarding the behavior of intentions on red code.

nd: The class defines compareTo(Int), so the code is green, but intention makes it red. Is it too esoteric to detect?

nd:
Updated it, please take a look.

idea/testData/intentions/convertTwoComparisonsToRangeCheck/compareToDifferentTypes.kt
nd: Actually with changed KotlinType.isComparable() test still works

yole:
Thanks!

PR #1074, opened 2017-05-08
KT-7848: Literal copy/paste processor for Kotlin

fitermay:


idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+
+private fun PsiElement.findContainingTemplate(): PsiElement {
+    val parent = this.parent
+    @Suppress("IfThenToElvis")
+    return if (parent is KtStringTemplateEntry) parent.parent else parent
+}
+
+private fun PsiFile.getTemplateIfAtLiteral(offset: Int): KtStringTemplateExpression? {
+    val at = this.findElementAt(offset) ?: return null
+    return when (at.node?.elementType) {
```
yole: Why not simply `at.getNonStrictParentOfType<KtStringTemplateExpression>`()?

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+        val endTp = file.getTemplateIfAtLiteral(selectionModel.selectionEnd) ?: return text
+        if (beginTp.isSingleQuoted() != endTp.isSingleQuoted()) {
+            return text
+        }
+
+
+        return if (beginTp.isSingleQuoted()) {
+            TemplateTokenizer(text).map {
+                when (it) {
+                    is LiteralChunk -> StringUtil.escaper(true, "\$\"").`fun`(it.text)
```
yole: I think that it's better to create a single `StringBuilder` to hold the entire preprocessed contents, and to use the overload of `escapeStringCharacters` that takes a `StringBuilder` as a parameter.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+                        tripleQuoteRe.replace(escapedDollar) { "\"\"" + "\${'\"'}".repeat(it.value.count() - 2) }
+                    }
+
+                    is EntryChunk -> chunk.text
+                    is NewLineChunk -> "\n"
+
+                }
+            }.joinToString(separator = "")
+        }
+
```
yole: Please reformat the files to remove unnecessary blank lines.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+
+}
+
+private sealed class TemplateChunk
+private data class LiteralChunk(val text: String) : TemplateChunk()
+private data class EntryChunk(val text: String) : TemplateChunk()
+private class NewLineChunk : TemplateChunk()
```
yole: Can be `object NewLineChunk`

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+
+private class TemplateTokenizer(private val inputString: String) : Sequence<TemplateChunk> {
```
yole: I think this naming is kind of confusing; a tokenizer is something that returns a sequence, not something that is a sequence. I'd call it `TemplateTokenSequence`

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+
+    private fun String.guessIsTemplateEntryStart(): Boolean =
+            if (this.startsWith("\${")) {
+                true
+            }
+            else if (this.length > 1 && this[0] == '$'){
+                KotlinLexer().apply { start(this@guessIsTemplateEntryStart.substring(1)) }.tokenType == KtTokens.IDENTIFIER
```
yole: Better to declare the result of `substring` as a variable outside of the `apply` block.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+            while (to < inputString.length) {
+                val c = inputString[to]
+                if (c == '\\') {
+                    to += 1
+                    if (to < inputString.length)
+                        to += 1
+                    continue
+
+                }
+                when (c) {
```
yole: No reason to use `when` here; `if` would be cleaner.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
```diff
@@ -0,0 +1,285 @@
...
+        }
+        return buildIterator {
+            var from = 0
+            var to = 0
+            while (to < inputString.length) {
+                val c = inputString[to]
+                if (c == '\\') {
+                    to += 1
+                    if (to < inputString.length)
+                        to += 1
```
yole: I think this assumes that all escapes are single-character? That's not true; there are also `\uxxxx` escape sequences.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
yole: Also would be great to have some unit tests for `TemplateTokenizer` (separate from the overall functional tests).

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
fitermay: This one is intentional.  If the cursor is __inside__ a long entry or short entry we don't want to consider it.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
fitermay: >ok

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
fitermay: Good point

fitermay:
Thanks for the review. All of these are indeed problems except for "getTemplateIfAtLiteral". I'm pretty sure I'm doing the correct thing there.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
yole: OK, I see. Then this is fine.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinLiteralCopyPasteProcessor.kt
fitermay: Actually, after giving it some though I don't think it will be a problem because in this case we only care about escape sequences which apply to escaping '$'

yole:
Thanks again!

fitermay:
That was not the final push. I just pushed it so I could work off another machine. Just finished now, with a few more changes.

PR #1075, opened 2017-05-08
KT-7848 -- unmerged changes

fitermay:


fitermay:
@yole Some umerged commits from KT-7848 (Tests for the  TemplateTokenSequence and a fix)

yole:
Thanks!

PR #1076, opened 2017-05-08
KT-17708 - Remove "Move to class body" intention for annotation parameters

nphumbert:
As I understand, annotation classes cannot have a body. Consequently, the "Move to class body" intention should not be available for annotation parameters.

#KT-17708 Fixed


yole:

PR #1077, opened 2017-05-09
Take control flow into account for checking fall-through switch branches

nd:
#KT-16133 Fixed


semoro:
Thank you for the contribution a lot! :slightly_smiling_face:

PR #1078, opened 2017-05-09
Detect escaped backslashes

nd:
#KT-15761 Fixed


semoro:

PR #1080, opened 2017-05-11
KT-17580: Enable 'when' as statement exhaustiveness check for sealed classes

fitermay:
This feature was already in the code. Just needed to enable for statements.


mglukhikh:
Sorry, but NON_EXHAUSTIVE_WHEN is intentionally not reported on sealed classes, it's a very controversial warning :(. About an issue, we need here an intention only, not an inspection or compiler warning.

mglukhikh:
You can here, for example, take `AddWhenRemainingBranchesFix` and convert it to an intention, it's relatively easy. Then `NON_EXHAUSTIVE_WHEN` for enums can use this intention as quick-fix via `IntentionWrapper`.

fitermay:
A sealed class is a type of enumeration, I'm surprised it is controversial. I will convert to intention.

fitermay:
Hi,
I made the changes but in a slightly different way than suggested. Hope it's ok

mglukhikh:
Yes, I think this way is also possible. Will check everything manually and merge.

PR #1081, opened 2017-05-11
KT-17853: Kotlin.js switched parameters of Math.atan2

gzoritchak:
https://youtrack.jetbrains.com/issue/KT-17853
The current exposition of the JavaScript [Math object] switched the parameters names: x and y are inverted which is confusing. 
The correct version is `atan2(y,x)` and not the current `atan2(x,y)`


ilya-g:
Thanks for the fix!

PR #1082, opened 2017-05-12
Detect recursive property accessors

nd:
#KT-17221 Fixed

### DIFF
--- a/idea/resources/inspectionDescriptions/RecursivePropertyAccessor.html
+++ b/idea/resources/inspectionDescriptions/RecursivePropertyAccessor.html
@@ -1,0 +1,23 @@
+<html>
+<body>
+Reports recursive property accessor calls which can end up with StackOverflowError
+<p>
+  For example:
+  <code><pre>
+    <b>class</b> A {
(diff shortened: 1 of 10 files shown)


mglukhikh:
@nd Looks good, but I'd say the second case like below (duplicating comment in the issue) is even more important. Not sure should it be the same inspection, probably the other one is better.

```
// JavaInterface.java
public interface JavaInterface {
    String getSmth();
}
// Test.kt
class Test {
    val smth: String
    fun createJavaInterface() = object : JavaInterface {
        override fun getSmth() = smth // expected this@smth, but synthetic property is used instead!
    }
}
```

nd:
@mglukhikh sure, will try writing one.

mglukhikh:
@nd At the moment I've almost done it myself. It can work almost in the same way as the first part. So thanks for your help

mglukhikh:
Also, I've added a fix for the first part (done by you) which replaces recursive access with `field`.

mglukhikh:
Merged into master and 1.0.3, thank you

PR #1083, opened 2017-05-14
Report cases when class member can be private

nd:


idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.kotlin.psi.addRemoveModifier.addModifier
+import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.isInheritable
+import org.jetbrains.kotlin.psi.psiUtil.isOverridable
+
+class MemberVisibilityCanPrivateInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitProperty(property: KtProperty) {
```
yole: This should check val/var parameters defined in the primary constructor, too (KtParameter)

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+                    registerProblem(holder, function)
+                }
+            }
+        }
+    }
+
+
+    private fun canBePrivate(declaration: KtDeclaration): Boolean {
+        val modifiers = declaration.modifierList
+        if (modifiers != null && (modifiers.hasModifier(KtTokens.PRIVATE_KEYWORD) || modifiers.hasModifier(KtTokens.OVERRIDE_KEYWORD))) return false
```
yole: You can use `declaration.hasModifier()` directly

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+        val klass = (declaration.parent as? KtClassBody)?.parent as? KtClass ?: return false
+        if (!klass.isInheritable() && modifiers?.hasModifier(KtTokens.PROTECTED_KEYWORD) ?: false) return false //reported by ProtectedInFinalInspection
+        if (declaration.isOverridable()) return false
+        val allUsages = ReferencesSearch.search(declaration, declaration.useScope).count()
```
yole: This is very inefficient: usages search is an expensive operation and most of the cost is wasted. You should iterate over all usages and stop as soon as you find one outside the local search scope.

See also the [corresponding Java inspection](https://github.com/JetBrains/intellij-community/blob/master/java/java-analysis-impl/src/com/intellij/codeInspection/visibility/AccessCanBeTightenedInspection.java#L143) for inspiration.

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private fun registerProblem(holder: ProblemsHolder, declaration: KtDeclaration) {
+        holder.registerProblem(declaration,
+                               "Can have 'private' visibility",
+                               ProblemHighlightType.WEAK_WARNING,
+                               MakePrivateFix())
+    }
+
+    class MakePrivateFix : LocalQuickFix {
```
yole: You can reuse `AddModifierFix`

nd:
Thanks, updated it, please take a look.

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.isInheritable
+import org.jetbrains.kotlin.psi.psiUtil.isOverridable
+
+class MemberVisibilityCanPrivateInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitProperty(property: KtProperty) {
+                super.visitProperty(property)
+                if (canBePrivate(property)) {
```
yole: This is now reported on local variables (which are also KtProperty instances).

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+                    registerProblem(holder, parameter)
+                }
+            }
+        }
+    }
+
+
+    private fun canBePrivate(declaration: KtDeclaration): Boolean {
+        if (declaration.hasModifier(KtTokens.PRIVATE_KEYWORD) || declaration.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return false
+        val klass = declaration.containingClass() ?: return false
```
yole: Wouldn't hurt to handle members of `object` declarations too.

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+                true
+            }
+        })
+        return inClassUsageFound && !otherUsageFound
+    }
+
+    private fun registerProblem(holder: ProblemsHolder, declaration: KtDeclaration) {
+        val modifierListOwner = declaration.getParentOfType<KtModifierListOwner>(false) ?: return
+        holder.registerProblem(declaration,
+                               "Can have 'private' visibility",
```
yole: Better to include name of declaration in the message, otherwise it's hard to understand results of a batch inspection.

nd:
What is easier to review, new commits to this branch or a squashed force-pushed commit with new implementation?

nd:
Updated it, please take a look

idea/src/META-INF/plugin.xml
```diff
@@ -2180,6 +2180,14 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.MemberVisibilityCanPrivateInspection"
+                     displayName="Can have 'private' visibility"
```
yole: Better to show a more full name here: "Class member can have 'private' visibility"

idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -0,0 +1,91 @@
...
+        return inClassUsageFound && !otherUsageFound
+    }
+
+    private fun registerProblem(holder: ProblemsHolder, declaration: KtDeclaration) {
+        val modifierListOwner = declaration.getParentOfType<KtModifierListOwner>(false) ?: return
+        val member = when (declaration) {
+            is KtNamedFunction -> "Function"
+            else -> "Property"
+        }
+        holder.registerProblem(declaration,
```
yole: This highlights the entire declaration with a wavy underline, which is very intrusive. Should highlight only the name identifier.

yole:
This was in fact merged a while ago; forgot to close the pull request.

PR #1084, opened 2017-05-14
Semicolon after loop with empty body is not redundant

nd:
#KT-12805 Fixed


mglukhikh:
I think it's OK, thank you. Will merge soon.

PR #1085, opened 2017-05-17
Update ReadMe.md, mention slash must be included in path

jankotek:
If forward slash is not included, build fails. Update readme to mention that.


yole:
How exactly does the build fail for you? I have paths without slashes in my environment configuration and everything works correctly.

jankotek:
I tried again, it builds without `/` at end.

PR #1088, opened 2017-05-18
Updates ReadMe.md to name the Sublime Text Package

lucamtudor:
Changes the link name for the Sublime Text Package from "TextMate" to "Sublime Text"


goodwinnk:
Thank you for the update.

PR #1089, opened 2017-05-18
kt16016

alexanderdr:
Changes to turn the issue in kt16016 into a compiler diagnostic error rather than a
stack trace.

https://youtrack.jetbrains.com/issue/KT-16016


erokhins:
Merged manually

PR #1090, opened 2017-05-20
Fix nullability quick fix

shiraji:
https://youtrack.jetbrains.com/issue/KT-17726


mglukhikh:
Thank you. I'll try to take best from both this PR and https://github.com/JetBrains/kotlin/pull/1091

mglukhikh:
Partially merged into master

PR #1091, opened 2017-05-21
Fix nullability quick-fixes with implicit receiver

nd:
#KT-17726

### DIFF
--- a/idea/testData/quickfix/nullables/unsafeInfixCall/unsafeInvokeWithImplicitReceiver.kt
+++ b/idea/testData/quickfix/nullables/unsafeInfixCall/unsafeInvokeWithImplicitReceiver.kt
@@ -1,0 +1,15 @@
+// "Replace with safe (?.) call" "false"
+// ACTION: Add non-null asserted (!!) call
+// ACTION: Convert to expression body
+// ACTION: Replace with safe (this?.) call
+// ACTION: Wrap with '?.let { ... }' call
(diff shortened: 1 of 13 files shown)


idea/testData/quickfix/nullables/unsafeInfixCall/unsafeInvokeWithImplicitReceiver.kt
```diff
@@ -0,0 +1,15 @@
...
+// ERROR: Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type A?
+
+class A {
+    fun foo() {
+    }
+}
+
+fun A?.bar() {
+    <caret>foo()
+}
```
mglukhikh: Why does this test have so strange name? I see no `invoke` here, either safe or unsafe

mglukhikh:
Thank you. I'll try to take best from both this PR and https://github.com/JetBrains/kotlin/pull/1090

idea/testData/quickfix/nullables/unsafeInfixCall/unsafeInvokeWithImplicitReceiver.kt
nd: Right, it should be unsafeCallWithImplicitReceiver.kt..

mglukhikh:
Merged into master

PR #1093, opened 2017-05-23
KotlinCommonIntentionActionsFactory extensions for IDEA-85507 and KT-11906

xiexed:
`createAddMethodAction` and `createAddBeanPropertyActions` common intention actions added

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/KotlinCommonIntentionActionsFactory.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/KotlinCommonIntentionActionsFactory.kt
@@ -26,7 +26,7 @@ import org.jetbrains.kotlin.psi.KtModifierListOwner
 import org.jetbrains.uast.UDeclaration
 
 
-class KotlinCommonModifications : JvmCommonIntentionActionsFactory() {
+class KotlinCommonIntentionActionsFactory : JvmCommonIntentionActionsFactory() {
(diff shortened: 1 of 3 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/KotlinCommonIntentionActionsFactory.kt
```diff
@@ -0,0 +1,147 @@
...
+            override fun getFamilyName(): String = "Add method"
+
+            private val text = "Add method '$methodName' to '${uClass.name}'"
+
+            override fun getText(): String = text
+
+            override fun invoke(project: Project, file: PsiFile, editor: Editor?, startElement: PsiElement, endElement: PsiElement) {
+                val visibilityStr = javaVisibilityMapping.getValue(visibilityModifier)
+                val psiFactory = KtPsiFactory(uClass)
+                val function = psiFactory.createFunction("$visibilityStr fun $methodName($paramsStr)$returnTypeString{}")
```
yole: You should use `KtPsiFactory.CallableBuilder` instead of building the text manually.

xiexed: Updated. But such a strange thing this `CallableBuilder`, why not convert to string only in `asString` method to not depend on builder functions call order? And...  hm... what do you think, does it have sense to make a similar language independent builder, and make it a `createAddMethodAction` parameter?

PR #1096, opened 2017-05-24
Don't suggest cast when !! is possible

nd:
#KT-18033 Fixed


mglukhikh:
Merged into master, thanks.

PR #1097, opened 2017-05-24
Don't overwrite plugin classpaths #KT-18022

mrginglymus:
When the kotlin maven compiler plugin detects dependencies it overwrites `arguments.pluginClasspaths` with the classpaths of the detected dependencies. However, `arguments.pluginClasspaths` may have already been set by the compiler plugin earlier on if the `kapt` goal is enabled.

This pull request addresses this by merging the kapt classpaths and the detected classpaths, rather than having one overwrite the other.

#KT-18022 fixed.


cy6erGn0m:
The main question here is what if the same dependency exists in both lists? Is it safe to have duplicate dependencies in the plugin classpath?

yanex:
Merged manually. Thank you!

PR #1099, opened 2017-05-25
#KT-17895 Add inspection to replace rangeTo or .. call with until

geralt-encore:
A lot of code for this inspection is duplicated from `ReplaceRangeToWithUntilIntention`. I don't really know how to share logic properly between two of this since I haven't done it before. If you have any suggestions on how to improve it I will be happy to fix it. Also, if more tests are needed I will be happy to add them as well.

https://youtrack.jetbrains.com/issue/KT-17895
#KT-17895 fixed


ilya-g:
Note that you could just force push commits to your branch and they will be updated in the related PR immediately. There's no need to close it and create another one.

geralt-encore:
Yeah, I know it just didn't update after I amended my commit and I thought that it would be easier just to create a new one.

PR #1100, opened 2017-05-25
#KT-17895 Add inspection to replace rangeTo or .. call with until

geralt-encore:
A lot of code for this inspection is duplicated from ReplaceRangeToWithUntilIntention. I don't really know how to share logic properly between two of this since I haven't done it before. If you have any suggestions on how to improve it I will be happy to fix it. Also, if more tests are needed I will be happy to add them as well.

https://youtrack.jetbrains.com/issue/KT-17895
#KT-17895 fixed

### DIFF
--- a/idea/resources/inspectionDescriptions/ReplaceRangeToWithUntil.html
+++ b/idea/resources/inspectionDescriptions/ReplaceRangeToWithUntil.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection reports calls to 'rangeTo' or the '..' operator instead of calls to 'until'.
+</body>
+</html>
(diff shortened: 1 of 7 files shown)


mglukhikh:
I think in this particular case the best way is to delete existing intention, because inspection duplicates it. You can easy use extension functions from this file making it internal or public instead of private. In case we would need both intention and inspection (e.g. inspection reports only some subset of cases coverable by intention), we should add a property of intention type into inspection and use `isApplicable` function from it to find problems. I'll do all necessary actions myself today.

geralt-encore:
Thank you for your explanation! I just wasn't sure if it is safe to remove existing intention or not. Let me know if I can add something to the PR.

mglukhikh:
Merged into master, thanks.

PR #1101, opened 2017-05-26
KT-18006 Bring behavior of Kotlin literal copy/paste in line with Java's after IDEA-172984

fitermay:


yole:

PR #1103, opened 2017-05-28
JS: Avoid declaration skeleton body generation in case of external class or object

Pozo:
When you are working on a Kotlin JS project and you would like to tell Kotlin that a certain declaration is written in pure JavaScript, you should mark it with external modifier. According to the [reference](https://kotlinlang.org/docs/reference/js-interop.html) you should omit bodies of external declarations.

Currently when you're using "Create member function" quick fix on an external class or object, Idea generates you the function body as well, which is syntactically incorrect, and you have to remove these every time. This can be painfully especially when you're working on a interop library.

This pull request address this by extending "declaration skeleton" generation logic take into account three cases.

- creating member function on an external class declaration
- creating member function on an external object declaration
- creating member function on an companion object within an external class declaration (and only class declaration since 'companion' modifier is not applicable inside an object)


yole:

PR #1104, opened 2017-05-29
Convert to primary constructor preserves custom accessors

nd:


mglukhikh:
Thanks, I will merge it.

PR #1105, opened 2017-05-29
Add quickfix for @receiver annotation being applied to extension member instead of extension type

cypressious:
Fixes #KT-14648


idea/src/org/jetbrains/kotlin/idea/quickfix/MoveReceiverAnnotationFix.kt
```diff
@@ -0,0 +1,59 @@
...
+
+            val declaration = entry.getParentOfType<KtCallableDeclaration>(true) ?: return null
+            if (declaration.receiverTypeReference == null) return null
+
+            return MoveReceiverAnnotationFix(entry)
+        }
+    }
+}
+
+fun String.foo() {
```
mglukhikh: Looks like not-deleted trash. I'll remove it.

mglukhikh:
Thanks, I'll merge it.

PR #1106, opened 2017-05-31
Fix KT-16725; Remove spaces before array indices

neworld:
I am not sure is this good fix, because I needed to modify another test, which actually conflicts with request of [KT-16725](https://youtrack.jetbrains.com/issue/KT-16725)


idea/tests/org/jetbrains/kotlin/formatter/FormatterTestGenerated.java
```diff
@@ -596,6 +596,12 @@ public void testSpaceBeforeFunctionLiteral() throws Exception {
             doTest(fileName);
         }
 
+        @TestMetadata("SpaceBeforeIndices.after.kt")
+        public void testSpaceBeforeIndices() throws Exception {
+            String fileName = KotlinTestUtils.navigationMetadata("idea/testData/formatter/SpaceBeforeIndices.after.kt");
+            doTest(fileName);
+        }
+
         @TestMetadata("SpacedInsideParans.after.kt")
```
Sweetnesss1: Lenor

yole:
The fix is good; the test modification you did is fine.

yole:
Fixed one additional test and merged manually. Thanks for your contribution!

PR #1107, opened 2017-05-31
#KT-13099 Formatting

geralt-encore:
https://youtrack.jetbrains.com/issue/KT-13099
#KT-13099 fixed


yole:

PR #1108, opened 2017-06-01
KT-18148 Incorrect, not working quickfix - final and can't be overridden

fitermay:


mglukhikh:
Thank you, merged into master.

PR #1113, opened 2017-06-02
KT-16558 Add space before parentheses config

neworld:
https://youtrack.jetbrains.com/issue/KT-16558


idea/formatter/src/org/jetbrains/kotlin/idea/formatter/kotlinSpacingRules.kt
```diff
@@ -212,11 +212,12 @@ fun createSpacingBuilder(settings: CodeStyleSettings, builderUtil: KotlinSpacing
             afterInside(LT, TYPE_ARGUMENT_LIST).spaces(0)
             beforeInside(GT, TYPE_ARGUMENT_LIST).spaces(0)
 
-            betweenInside(FOR_KEYWORD, LPAR, FOR).spacing(1, 1, 0, false, 0)
-            betweenInside(IF_KEYWORD, LPAR, IF).spacing(1, 1, 0, false, 0)
-            betweenInside(WHILE_KEYWORD, LPAR, WHILE).spacing(1, 1, 0, false, 0)
-            betweenInside(WHILE_KEYWORD, LPAR, DO_WHILE).spacing(1, 1, 0, false, 0)
-            betweenInside(WHEN_KEYWORD, LPAR, WHEN).spacing(1, 1, 0, false, 0)
+            betweenInside(FOR_KEYWORD, LPAR, FOR).lineBreakOrForceSpace(false, kotlinCommonSettings.SPACE_BEFORE_FOR_PARENTHESES)
```
yole: There is never a line break in these locations, so you should better use `spaceIf`.

neworld: I noticed that `spaceIf` doesn't join lines and other tests start failing, [like this one](https://github.com/JetBrains/kotlin/blob/master/idea/testData/formatter/When.kt#L25-L26)

output is:

```kotlin
    when
    (true) {

    }
```

expected:

```kotlin
    when (true) {

    }
```

yole: I think this is as it should be. The "Keep line breaks in code" option tells the formatter to keep all line breaks, including weird ones like that, so it preserves them. In real-life cases, the important thing is to add or remove the space when the statement is on a single line.

neworld: Make sense. I make `spaceIf()`. Later I will test and fix if needed "Keep when formatting -> Line breaks"

neworld:
Fixed. `KEEP_LINE_BREAKS` works

yole:

PR #1115, opened 2017-06-06
KT-17823 Intention "Make private" and friends should respect modifier order

t-kameyama:
Fix for KT-17823
https://youtrack.jetbrains.com/issue/KT-17823


yole:

PR #1118, opened 2017-06-07
Eliminate some unnecessary CHECKCAST operations

scottkennedy:
In certain circumstances, an ArrayList would be cast to a Collection and then back to a List.

This eliminates CHECKCAST operations to weaker types.

#KT-18311 Fixed


udalov:
Hi, thanks for the pull request. The main problem with this approach is that the classes loaded during the execution of the compiler might differ significantly from the classes that are being compiled. For example, since the Kotlin compiler is used to compile itself, it won't be easy to remove a supertype from one of the classes in the compiler itself (because with these changes, the compiler would still think that the supertype is there, leading to bad code generated). In general, it's incorrect to draw any information about the code being compiled from the JVM process that is running the compiler.

Another problem is that `Class.forName` leads to initialization of the class, which may lead to undesired side effects during compilation.

Lastly, I suspect that trying to load a class every time in `StackValue.coerce` causes the class loader to go look for it, involving a noticeable amount of I/O operations, which may lead to worse performance here.

PR #1120, opened 2017-06-08
KT-17026 Replace explicit parameter should not be shown on destructuring declaration

t-kameyama:
Fix for KT-17026
https://youtrack.jetbrains.com/issue/KT-17026


mglukhikh:
Thank you, I'll integrate this.

PR #1122, opened 2017-06-09
KT-18158 Expand selection should select the comment after expression getter on the same line

t-kameyama:
Fix for https://youtrack.jetbrains.com/issue/KT-18158


yole:
While this is a good change and I'll accept it, this is not what the original issue was about. The original issue asks to change the behavior of Edit | Extend Selection so that it includes an extra step for the declaration with comment. See https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinDeclarationSelectioner.kt for related code.

t-kameyama:
@yole 
I added for original issue.

yole:

PR #1124, opened 2017-06-10
KT-15368: Add Boolean? == const to elvis conversion and vice versa

Dimach:
Convert elvis to equality check implemented as inspection due
code style guide


mglukhikh:
Edited and merged into master, thank you!

PR #1125, opened 2017-06-10
KT-13676 Add better quickfix for scope functions

shiraji:
#KT-13676 Fixed

https://youtrack.jetbrains.com/issue/KT-13676


mglukhikh:
Edited and merged into master. Receiver analysis by text replaced with analysis by descriptors. Thank you.

PR #1126, opened 2017-06-11
KT-18236 Add inspection for potentially wrongly placed unary operators

shiraji:
#KT-18236 Fixed

https://youtrack.jetbrains.com/issue/KT-18236


cypressious:
Why not also unary plus? It could be overloaded.

shiraji:
Good point. I will add that.

shiraji:
done

mglukhikh:
Thanks, merged into master

PR #1127, opened 2017-06-11
KT-18407: Do not show "Move to constructor" intention for properties declared in interfaces

vasily-kirichenko:
#KT-18407 Fixed


yole:
Please provide tests for this change. See [here](https://github.com/JetBrains/kotlin/blob/master/idea/testData/intentions/addConstModifier/inapplicableToConst.kt) for an example of a test that verifies that an intention is not available.

vasily-kirichenko:
@yole done.

yole:
Merged manually. Thanks a lot for your contribution! Please take a look at the additional commit: https://github.com/JetBrains/kotlin/commit/a6d80b78de012dfe569b55557c4d56383da44673

PR #1128, opened 2017-06-12
"Make <modifier>" intention inserts modifier between annotation and class keywords

t-kameyama:
Fix for
https://youtrack.jetbrains.com/issue/KT-14092
https://youtrack.jetbrains.com/issue/KT-14093


cypressious:
@t-kameyama While you're at it, you could also fix https://youtrack.jetbrains.com/issue/KT-18425

yole:

PR #1129, opened 2017-06-12
KT-18424: Add Kotlin Data Class file template

pavlospt:
Just like having a file template for `Class`, there are times that we need to generate a `Data Class` instead of generating a `Class` and modifying it to be a `Data Class`. 

Related feature request: https://youtrack.jetbrains.com/issue/KT-18424

### DIFF
--- a/idea/resources/fileTemplates/internal/Kotlin Data Class.kt.ft
+++ b/idea/resources/fileTemplates/internal/Kotlin Data Class.kt.ft
@@ -1,0 +1,5 @@
+#if (${PACKAGE_NAME} && ${PACKAGE_NAME} != "")package ${PACKAGE_NAME}
+
+#end
+#parse("File Header.java")
+data class ${NAME}()
--- a/idea/resources/fileTemplates/internal/Kotlin Data Class.kt.html
+++ b/idea/resources/fileTemplates/internal/Kotlin Data Class.kt.html
@@ -1,0 +1,14 @@
+<html>
+<body>
+<table border="0" cellpadding="2" cellspacing="0" style="border-collapse: collapse">
+  <tr>
+    <td colspan="3">
+      <font face="verdana" size="-1">
+        This is a built-in template used by <b>IDEA</b> each time you create a
+        Kotlin data class
+      </font>
+    </td>
+  </tr>
+</table>
+</body>
+</html>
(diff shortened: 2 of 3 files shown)


pavlospt:
I see that some have been assigned to it! Should I close this PR or there is a chance it will be merged?

Alefas:
There are lots of chances to have it merged. Sorry for delays, just give me a bit more time before I start working on bunch of pull requests we have now in repo. It's wrong to have so much open PRs, so I'm really sorry and we will try to fix it more or less soon.

pavlospt:
@Alefas thanks for updating! Please take your time :) Appreciate the info!

Alefas:
> @Alefas is there any update with this :) ?

Oh, sorry, I completely missed it among lots of GitHub notifications (good reason for me to finally read https://blog.github.com/2017-07-18-managing-large-numbers-of-github-notifications/ and fix it). We are debating on it if we want one more item in this dialog, it's already too much and big question if we have a way to solve this usability problem differently. We'll try to solve it somehow before 1.3.20, or accept this PR.

Alefas:
> @Alefas What is weird though is that since Kotlin can support many different types of classes (e.g.: `enum` `sealed` `data`) why have that dialog only support the classic `class` and `enum` for example? :)

To select right thing in dialog you have to use mouse or press 'down' too many times in some cases. Most common case could be to create just plain file and then to add what you actually need like "data class" and so on.

I have idea to support "data class Foo" in name field, that would choose data class template, but it's hardly discoverable to people. Anyway, we will try to investigate it a bit to find proper UX for lots of cases.

pavlospt:
But you can also write the initial letters of the menu item you want and not just use your mouse or `down` key.

Alefas:
> But you can also write the initial letters of the menu item you want and not just use your mouse or `down` key.

In any case I initiated consultation with our UX team, probably they will give us better solution, I just don't know. As a last resort, we can actually add data class, it's not a big deal. I just want to be sure, that we don't have better way to do it.

vladimirdolzhenko:
This PR requires some changes
- use i18n for the name
- I think it is reasonable to add `sealed class` and `annotation` as part of [KT-18424](https://youtrack.jetbrains.com/issue/KT-18424)

If you don't mind I can do it for you and merge after that.

pavlospt:
@vladimirdolzhenko Sure please go ahead 😄  Since I have changed laptops from the time the PR was opened, I would probably need to setup everything from scratch to refresh this PR 😄

vladimirdolzhenko:
Thanks a lot for the contribution. 

It is merged.

PR #1130, opened 2017-06-13
KT-18425 Make <modifier> intention inserts modifier at wrong position for sealed class

t-kameyama:
Fix for https://youtrack.jetbrains.com/issue/KT-18425


mglukhikh:
Thanks, will merge it today

PR #1131, opened 2017-06-13
REPL type inference support

khud:
I am working on new Kotlin REPL (as part of Apache Spark integration) based on GenericRepl and need to have type inference information available from GenericRepl. This PR provides minimum necessary information without substantial changes in the compiler. Only last inferred type is required. It allows to have `:type` command easily for new shell (like Scala and Haskell), moreover it simplifies support for unnamed REPL results (i.e. resN).


compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/repl/ReplApi.kt
```diff
@@ -83,7 +84,9 @@ sealed class ReplCompileResult : Serializable {
                           val mainClassName: String,
                           val classes: List<CompiledClassData>,
                           val hasResult: Boolean,
-                          val classpathAddendum: List<File>) : ReplCompileResult()
+                          val classpathAddendum: List<File>) : ReplCompileResult() {
+        val type = LastInferredTypeHolder.inferredType.get()
```
udalov: It would be easier to pass the expression type explicitly where `ReplCompileResult.CompiledClasses` is created at `GenericReplCompiler.compile` with:
```
val expression = psiFile.getChildOfType<KtExpression>()
compilerState.analyzerEngine.trace.bindingContext.getType(expression)
```
(+ handle the case when there's no expression in a REPL line)

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/repl/ReplApi.kt
```diff
@@ -111,7 +114,9 @@ interface ReplEvalAction {
 
 sealed class ReplEvalResult : Serializable {
     class ValueResult(val value: Any?) : ReplEvalResult() {
-        override fun toString(): String = "Result: $value"
+        val type = LastInferredTypeHolder.inferredType.get()
```
udalov: I think you can obtain this type from `compileResult` at `GenericReplEvaluator.eval` and pass to `ValueResult`'s constructor here

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/repl/ReplApi.kt
khud: Partially done, but not yet. Need some time to check.

khud:
Found a suitable way to get expression from psiFile. It looks strait-forward, but correct. Any other suggestions?

udalov:
I don't have any other comments at this point.

@ligee Please review as well

ligee:
@khud it seems that passing our internal `KotlinType` class to the REPL API is not a good idea. Could you, please, change it so the `ValueResult` contains a string with type instead. 
@udalov please, provide an example, how to render the type into a string properly.

udalov:
To render a `KotlinType` to a string, use `DescriptorRenderer.renderType`. You can either get the pre-declared renderer instance from `DescriptorRenderer`'s companion object (`FQ_NAMES_IN_TYPES` and `SHORT_NAMES_IN_TYPES` are some of the most commonly used ones), or construct a `DescriptorRendererImpl` with custom `DescriptorRendererOptions` if none of the pre-declared ones suit your needs.

khud:
What is wrong with KotlinType::toString()? Do we really need all machinery with `DescriptorRenderer.renderType`?

udalov:
> What is wrong with KotlinType::toString()? Do we really need all machinery with DescriptorRenderer.renderType?

The difference between the two is that we use `toString` for debug and tests only, and `DescriptorRenderer` -- for presenting types and declarations to the user (e.g. in the IDE, in reflection, etc.). So, the former has no requirements on what it should return, but the latter is relatively much more stable and tested. You may notice that some KotlinType implementations use `DescriptorRenderer.DEBUG` in their `toString`, which outputs some additional debug information. Some have their own ad-hoc implementation that can change at any moment. Some can even return things like `"<Not computed yet>"` in their `toString`.

khud:
@udalov good point.  @ligee Done. `FQ_NAMES_IN_TYPES` seems to be more robust solution. I believe that to pass an instance of `DescriptorRenderer` as argument (with default value) is too much for any reasonable case.  Isn't it?

ligee:
@khud Thank you for the PR. I cleaned up unused imports, squashed commits into the single one and merged into the master. (https://github.com/JetBrains/kotlin/commit/d165ea9ea75d9a4ee278db16b8257b56e83cb528)

PR #1132, opened 2017-06-13
Support implicit receiver in if-then intentions

nd:


mglukhikh:
Thank you, merged into master

PR #1134, opened 2017-06-14
KT-17920 Add inspection removing redundant spread operator for arrayOf call

shiraji:
#KT-17920 Fixed

https://youtrack.jetbrains.com/issue/KT-17920


mglukhikh:
Thank you, merged into master

PR #1136, opened 2017-06-14
Add argument list parentheses wrapping format settings; part of KT-9423

neworld:
image

These settings allow easier follow [long class header convention](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting) and such formatting is heavily used in my team.

Ported from Java. I am not sure about single argument wrapping. I suppose it is a bug and don't belong in this feature because it happens of wrapper settings in [KotlinCommonBlock.kt](https://github.com/JetBrains/kotlin/blob/master/idea/formatter/src/org/jetbrains/kotlin/idea/formatter/KotlinCommonBlock.kt).


yole:
I actually just commited a set of changes implementing the same options - I needed an implementation of those for my work on the Kotlin style guide. Sorry about the duplicate work you've had to do!

PR #1137, opened 2017-06-15
KT-18074 Suggestion in Intention 'Specify return type explicitly' doesn't support generic type parameter

t-kameyama:
Fix for https://youtrack.jetbrains.com/issue/KT-18074

### DIFF
--- a/idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
+++ b/idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
@@ -19,6 +19,7 @@
 package org.jetbrains.kotlin.idea.util
 
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.TypeParameterDescriptor
 import org.jetbrains.kotlin.idea.imports.canBeReferencedViaImport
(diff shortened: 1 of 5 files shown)


yole:
I think it's better to fix the logic so that type parameters can be checked correctly, and not to suppress the check.

t-kameyama:
I fixed to check type parameter.

idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
```diff
@@ -97,7 +98,15 @@ fun KotlinType.isResolvableInScope(scope: LexicalScope?, checkTypeParameters: Bo
 
     val descriptor = constructor.declarationDescriptor
     if (descriptor == null || descriptor.name.isSpecial) return false
-    if (!checkTypeParameters && descriptor is TypeParameterDescriptor) return true
+
+    if (descriptor is TypeParameterDescriptor) {
+        if (checkTypeParameters) {
+            val owner = descriptor.containingDeclaration
+            if (owner is FunctionDescriptor && owner.typeParameters.any { it == descriptor }) return true
```
mglukhikh: any { it == descriptor } is equivalent to just contains(descriptor)

idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
```diff
@@ -97,7 +98,15 @@ fun KotlinType.isResolvableInScope(scope: LexicalScope?, checkTypeParameters: Bo
...
+        } else {
```
mglukhikh: FYI: in Kotlin project if...else is formatted in a different way:
```
if () {
}
else {
}
```

idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
t-kameyama: Fixed

idea/ide-common/src/org/jetbrains/kotlin/idea/util/TypeUtils.kt
t-kameyama: Fixed

mglukhikh:
Please pay attention to the fact that function you've changed `KotlinType.isResolvableInScope` is used in different places, not only in the intention you are trying to fix. That's why some logic was broken.

t-kameyama:
I changed to check type parameters in SpecifyTypeExplicitlyIntention.
Please review again.

mglukhikh:
Thank you, I've merged it.

PR #1139, opened 2017-06-16
Apply De Morgan's laws to non-negated binary expressions

nd:
#KT-18460 Fixed


cypressious:
If you rename the intention, you should also rename the description files in https://github.com/JetBrains/kotlin/tree/master/idea/resources/intentionDescriptions/ConvertNegatedExpressionWithDemorgansLawIntention

mglukhikh:
Thanks, looks great. After your fix, intention will not be available on `!` (which was true before), but looks like it's not a problem.

mglukhikh:
Merged manually

PR #1140, opened 2017-06-16
Fixed Slack link

Miha-x64:
http://kotlinslackin.herokuapp.com/ is down, replaced with http://slack.kotlinlang.org/


yole:
Thanks!

PR #1141, opened 2017-06-16
Fix for KT-13458 and KT-13436

t-kameyama:
Fix for
https://youtrack.jetbrains.com/issue/KT-13458
https://youtrack.jetbrains.com/issue/KT-13436


mglukhikh:
Mostly OK, thanks, will merge into master

mglukhikh:
Merged manually

PR #1142, opened 2017-06-16
Disabling obsolete "Final Kotlin class or function with Spring annotation" inspection 

xiexed:
Because Uast-based "Final declaration can't be overridden at runtime" inspection covers it's functionality


asedunov:
Thank you!

PR #1143, opened 2017-06-17
Intention to replace add/addAll on a mutable collection with +=

nd:


yole:
This intention should not be offered when the receiver is a `var`, otherwise there will be an overload resolution ambiguity after you change to += (see https://youtrack.jetbrains.com/issue/KT-5488).

```
fun foo() {
    var f = mutableListOf(1)
    f.addAll(listOf(2   ))
}
```

nd:
Thanks! fixed

yole:
Thanks!

PR #1144, opened 2017-06-19
Inspection detecting redundant Unit return type

nd:


yole:
There's already RemoveExplicitTypeIntention which provides the same action as your intention on the `fun bar(): Unit = println("ok")` example. We should avoid showing such duplicated actions. Also, you don't need to reimplement the logic for deleting the type specification; simply `.setTypeReference(null)` should work.

nd:
Reworked it so that RemoveExplicitTypeIntention also handles Unit return type in function with body

yole:
Now the inspection highlights any return type of a function with an expression body as redundant, even if it is not Unit:

fun shazam(): Int = 1

nd:
Thanks, fixed

yole:
Thanks!

PR #1145, opened 2017-06-19
evaluate project build dir lazily (fix KT-10537)

nskvortsov:
Gradle plugin should access project model as late as possible to support modification of properties after `apply plugin` instruction was process. This particular patch implements this for Kotlin-related task output directory.


h0tk3y:
Merged manually [(commit)](https://github.com/JetBrains/kotlin/commit/b2cdb0d63d39670aa7ad386d24a26ba455a565e0). Thanks!

PR #1147, opened 2017-06-19
Add toHexString extension function for ByteArray

MarcinMoskala:
As response to discussion on Slack


MarcinMoskala:
Hey, is function placed correctly? Do you want me to also add Byte.toHex extension function?

ilya-g:
Nice function, though there are some implementation decisions that need to be discussed:
- what should be the default case of characters and whether should it be customizable?
- do we need an optional byte separator, like space?
- do we need to cover the use case of separating longer groups of bytes, for example to have a newline after each 40 bytes?

Perhaps it's worth to write a KEEP to have a place for discussion, or at least an issue in the tracker.

qwwdfsad:
We do plan to provide a full-blown design and implementation from scratch, along with KEEP and alternatives. Closing this as abandoned

PR #1148, opened 2017-06-20
a fix for gradle kotlin dsl project config

alexvas:
Gradle plugin gradle-script-kotlin became gradle-kotlin-dsl, so
KotlinBuildScript now lives inside gradle-kotlin-dsl<skipped>.jar

Adding one more prefix to classpath search, so both old
(gradle-script-kotlin) and new (gradle-kotlin-dsl) versions of
Gradle plugin are supported. See
https://youtrack.jetbrains.com/issue/KT-18557


yole:
We've already committed the same change internally (https://github.com/JetBrains/kotlin/commit/2e823a4816c13076efc89c96c567d666acd9203b). Thanks anyway!

PR #1150, opened 2017-06-21
KT-14799: added inspection to simplify successive null checks.

Dimach:


mglukhikh:
Looks like you must change all qualified expression' components on stability, otherwise offered transformation is not semantically equivalent. For example, class mutable property is not stable, because it can return different results on different reads. As an example, use `KtExpression.isStable` from `IfThenUtils`, however, it works correctly not for all expression types. Full set of stability-related functions is available in `DataFlowValeFactory`.

mglukhikh:
More exactly, if you're converting to `a?.b?.c?.d != null`, three components `a`, `a.b`, `a.b.c` must be checked. Looks like last component `a.b.c.d` may not be checked, because it's read just once in either case.

Dimach:
Done.

idea/src/org/jetbrains/kotlin/idea/inspections/NullChecksToSafeCallInspection.kt
```diff
@@ -78,7 +81,7 @@ private fun collectNullCheckExpressions(expression: KtBinaryExpression): Triple<
 
 private fun isNullChecksToSafeCallFixAvailable(expression: KtBinaryExpression): Boolean {
     val (lte, rte) = collectNullCheckExpressions(expression) ?: return false
-    return getDuplicates(lte, rte)?.parent as? KtDotQualifiedExpression != null
+    return lte.isChainStable() && getDuplicates(lte, rte)?.parent as? KtDotQualifiedExpression != null
```
mglukhikh: At least one or two tests required here

Dimach: I don't know  how to write tests in this case because if i run test on code with unstable chain like this `fun main(args: Array<String>) {
    val a: Testtt? = Testtt()
    if (a?.a<caret> != null && a?.a.a != null) {
 
    }
}

class Testtt {
    val a: Testtt? = null
        get() = field
}`
compiler throws a error
`Smart cast to 'Testtt' is impossible, because 'a.a' is a property that has open or custom getter`.  What i must do in this case?

mglukhikh: You can check the case when `a` is nullable extension property. Also, you can modify last check to `a?.a?.a != null`.

Dimach: If i modify last check to `a?.a?.a != null` code will compile, yes, but it didnt test code that checking stability of expression because i explicitly check that is right part of expression is dot, safecall and !! makes intention applicable  `getDuplicates(lte, rte)?.parent as? KtDotQualifiedExpression != null`

mglukhikh:
Mostly it's OK now. Will merge it today. BTW, a counter-example found:

```
val a: Testtt? = Testtt()
    if (a<caret> != null && a.a != null) {

    }
}
class Testtt
val Testtt?.a: Testtt? get() = null
```

Your version of inspection suggest to make replacement, but it's not equivalent here (already fixed by me).

mglukhikh:
Also, this test is buggy:

```
    val a: Testtt? = Testtt()
    if (a<caret> != null && a.a.a != null) {

    }
}
class Testtt {
    val a: Testtt = this
}
```

You're replacing it now with `a?.a.a != null`, which is certainly not true (dot call is unsafe here).

mglukhikh:
Excellent false-positive (extracted from Kotlin project):

```
class KotlinType
class KotlinTypeInfo(val type: KotlinType?) {
    fun foo(other: KotlinTypeInfo) {
        if (type != null && other.type != null) {

        }
    }
}
```

Here you erroneously suggest to apply fix, and we get an exception.

mglukhikh:
So merged manually, but with some fixes. BTW I eliminated cases like `x != null && x.y.z != null`, thus allowing only one additional selector. Anyway, thank you.

PR #1152, opened 2017-06-21
Add reverse-domain version of JS Gradle plugin.

JakeWharton:
It was the only one missing this recommended naming form.


h0tk3y:
We did not add the name to the existing `kotlin2js` plugin intentionally: we planned to make some breaking changes to the JS plugin, so we may later release the updated plugin as `org.jetbrains.kotlin.js` (and short name `kotlin-js`), and deprecate `kotlin2js`, leaving it as it is. Therefore closing the request.

JakeWharton:
Got it. Thanks for the info.

PR #1153, opened 2017-06-22
KT-18534 Quick-fix to add empty brackets after primary constructor

t-kameyama:
Fix for https://youtrack.jetbrains.com/issue/KT-18534


mglukhikh:
Fix looks correct, thank you

mglukhikh:
Merged manually

PR #1154, opened 2017-06-22
Do not suggest adding !! on null constant

nd:


mglukhikh:
FYI: in Kotlin project we use non-standard formatting of if-else:

```
if (something) {
}
else if (smthOther) {
}
else {
}
```

mglukhikh:
Also, your fix is correct but it may be extended to situation when expression is not a null constant but exactly null. E.g.

```
fun foo(arg: String?) {
    if (arg == null) {
        val x: String = arg // Don't add !! here, it's obviously KNPE
    }    
}
```

I can think about it myself or you can do it, your choice. To do it, we should extract KotlinTypeInfo for `arg` expression in RHS using `EXPRESSION_TYPE_INFO` slice, then read `dataFlowInfo` from it and extract `arg` nullability.

mglukhikh:
However, I do not think this quick fix should be in cleanup / migration because this transformation is not a migration, just a fix of syntax error. I will remove it from there.

nd:
Thanks, will try implementing the non-constant case

nd:
I should have noticed the formatting issue: in this case it is obviously different from the surrounding code, but in code written from scratch it might be not so obvious. Maybe it is worth suggesting to run 'Reformat code' in the README contributing section..

mglukhikh:
OK, I will merge it

mglukhikh:
Merged manually, thank you

PR #1155, opened 2017-06-22
Spring dependency-generations actions: `runWriteAction` moved to the proper place

xiexed:
A fix for KT-18385


asedunov:
Merged, thank you!

PR #1156, opened 2017-06-23
First semicolon in enum with no entries is not redundant

nd:


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantSemicolonInspection.kt
```diff
@@ -57,6 +58,16 @@ class RedundantSemicolonInspection : AbstractKotlinInspection(), CleanupLocalIns
 
         if (semicolon.parent is KtEnumEntry) return false
 
+        (semicolon.parent.parent as? KtClass)?.let {
+            if (it.isEnum() && it.getChildrenOfType<KtEnumEntry>().isEmpty()) {
+                if (semicolon.prevLeaf { it !is PsiWhiteSpace && it !is PsiComment && !it.isLineBreak() }?.node?.elementType == KtTokens.LBRACE &&
+                    semicolon.nextLeaf { it !is PsiWhiteSpace && it !is PsiComment && !it.isLineBreak() }?.node?.elementType != KtTokens.RBRACE) {
```
mglukhikh: I thinks it's more precise just to check here that `getDeclarations()` isn't empty, instead of doing these operations with `nextLeaf`.

nd: I've replaced the RBRACE check as you suggested. Looks like the LBRACE check is still needed: without it the inspection treats all semicolons in enum with no entries as non-redundant, instead of only the first one. Maybe there is a better way to check that there is no declarations before the semicolon?

mglukhikh: No, I don't know a better way, and yes, LBRACE check is still needed.

mglukhikh:
Merged manually

mglukhikh:
Thank you

PR #1157, opened 2017-06-23
KT-17815 Quick-fix "Replace with safe call & elvis"

t-kameyama:
Fix for https://youtrack.jetbrains.com/issue/KT-17815


mglukhikh:
This fix looks correct, thank you

mglukhikh:
Merged manually

PR #1158, opened 2017-06-25
KT-18501: Fix to find KDoc even if KDoc is in ModifierList

takahirom:
https://youtrack.jetbrains.com/issue/KT-18501


semoro:

PR #1159, opened 2017-06-26
KT-14093 "Make <modifier>" intention available only on modifier when declaration already have a visibility modifier

t-kameyama:
FIx for https://youtrack.jetbrains.com/issue/KT-14093


mglukhikh:
I think it's OK, thank you.

mglukhikh:
Merged manually

PR #1160, opened 2017-06-27
Fix String capitalize/decapitalize sample references

stkent:


ilya-g:
Thanks, I've merged it manually.

PR #1161, opened 2017-06-27
Fix decapitalize sample name

stkent:
Combines with #1160 to fix both the capitalize and decapitalize samples!


ilya-g:
I've merged it with #1160 manually.

PR #1162, opened 2017-06-28
Add inspection for simplifying 'if's with constant condition

cypressious:
Fixes #KT-17919

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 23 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -0,0 +1,120 @@
...
+        override fun getFamilyName() = "Simplify expression"
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val ifExpression = descriptor.psiElement as? KtIfExpression ?: return
+            val caretModel = ifExpression.findExistingEditor()?.caretModel
+
+            val branch = ifExpression.branch(conditionValue)?.unwrapBlockOrParenthesis() ?: return
+
+            val lastExpression =
+                    if (branch !is KtBlockExpression) {
```
xiexed: Probably with `when` it will be more readable

mglukhikh:
NB: in future, please write both GroupName and GroupPath in inspection description inside plugin.xml.

mglukhikh:
Thank you, I've merged it with minor changes. I recommend using LocalInspectionTestGenerated for inspection fix tests, it requires less text than QuickFixTestGenerated (and can check more).

PR #1165, opened 2017-07-01
KT-6676 Show enum ordinal in quickdoc

Knize:


idea/src/org/jetbrains/kotlin/idea/KotlinQuickDocumentationProvider.kt
```diff
@@ -175,6 +175,23 @@ class KotlinQuickDocumentationProvider : AbstractDocumentationProvider() {
                 // element is not an KtReferenceExpression, but KtClass of enum
                 return renderEnum(element, originalElement, quickNavigation)
             }
+            else if (element is KtEnumEntry && !quickNavigation) {
+                val desc = element.resolveToDescriptorIfAny()
+                val ordinal =
```
semoro: It's better to calculate ordinal with `val ordinal = element.containingClassOrObject?.run { getChildrenOfType<KtEnumEntry>().indexOf(element) }`

semoro:

PR #1166, opened 2017-07-02
Improve a reported message if -jdk-home path does not exist.

vedun-z:
Fix https://youtrack.jetbrains.com/issue/KT-18794


udalov:
Thank you!

PR #1167, opened 2017-07-02
[KT-9570] Ranges should override sum, max, min

MGaetan89:
Related issue: https://youtrack.jetbrains.com/issue/KT-9570

### DIFF
--- a/js/js.libraries/src/core/generated/_CollectionsJs.kt
+++ b/js/js.libraries/src/core/generated/_CollectionsJs.kt
@@ -2169,7 +2169,7 @@ public fun Iterable<Short>.sum(): Int {
 @kotlin.jvm.JvmName("sumOfInt")
 public fun Iterable<Int>.sum(): Int {
     if (this is IntProgression) {
-        return (this.first + this.last) * this.count() / 2
+        return (this.first + this.last) * (this.count() / 2)
     }
     var sum: Int = 0
     for (element in this) {
@@ -2184,7 +2184,7 @@ public fun Iterable<Int>.sum(): Int {
 @kotlin.jvm.JvmName("sumOfLong")
 public fun Iterable<Long>.sum(): Long {
     if (this is LongProgression) {
-        return (this.first + this.last) * this.count() / 2
+        return (this.first + this.last) * (this.count() / 2)
     }
     var sum: Long = 0L
     for (element in this) {
--- a/js/js.libraries/src/core/generated/_CollectionsJs.kt
+++ b/js/js.libraries/src/core/generated/_CollectionsJs.kt
@@ -1385,6 +1385,11 @@ public inline fun <T> Iterable<T>.any(predicate: (T) -> Boolean): Boolean {
  * Returns the number of elements in this collection.
  */
 public fun <T> Iterable<T>.count(): Int {
+    when (this) {
+        is CharProgression -> return (this.last - this.first) / this.step + 1
+        is IntProgression -> return (this.last - this.first) / this.step + 1
+        is LongProgression -> return (this.last - this.first) / this.step + 1
+    }
     var count = 0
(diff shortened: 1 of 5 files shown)


js/js.libraries/src/core/generated/_CollectionsJs.kt
```diff
@@ -2165,6 +2183,9 @@ public fun Iterable<Int>.sum(): Int {
  */
 @kotlin.jvm.JvmName("sumOfLong")
 public fun Iterable<Long>.sum(): Long {
+    if (this is LongProgression) {
+        return (this.first + this.last) * this.count() / 2
```
HaasJona: Is this as overflow safe as the old approach? `(this.first + this.last) * this.count()` might overflow when `(this.first + this.last) * (this.count() / 2)` wouldn't.

MGaetan89: Nice catch. I updated the code.

HaasJona:
That was just a question. I think the new solution is also bad because now you are rounding down because of the Integer division when the count is uneven.

MGaetan89:
Right. I just quickly read the notification, without thinking too much about it.
I'll check this as soon as I can.

ilya-g:
This pull request in making no progress so I'm going to close it for now. I'll merge however doc fixes for `getProgressionLastElement` function.

jablko:
If I reopen and rebase this, and use the following to prevent overflows and rounding, is there any chance/interest in merging it?
```Kotlin
  val averageTimesTwo = this.first + this.last
  return if (averageTimesTwo % 2 == 0) averageTimesTwo / 2 * this.count()
  else this.count() / 2 * averageTimesTwo

**Comment by ilya-g** [comment]
@jablko Using `this.count()` in the implementation doesn't solve the original problem described in https://youtrack.jetbrains.com/issue/KT-9570

**Comment by jablko** [comment]
@ilya-g Thanks for the fast reply!

> @jablko Using `this.count()` in the implementation doesn't solve the original problem described in https://youtrack.jetbrains.com/issue/KT-9570

I think it does run in constant time (solves KT-9570) because this PR specializes `count()`, unless I misunderstand you or am overlooking something?
https://github.com/JetBrains/kotlin/pull/1167/files#diff-d4bb22410815c8bb30b15cfcfedca943c3271129318e4edbe0c1e1d2bce12b1bR1397-R1402

**Comment by ilya-g** [comment]
Ah, ok, that implementation is fast, but also susceptible to overflows. Note that it's not possible to represent the size of some `IntProgression`s (not even talking about `LongProgression`) with an `Int` number. You can see the concerns regarding `size`/`count()` for progressions in this PR: https://github.com/JetBrains/kotlin/pull/4095#issuecomment-796706836

**Comment by jablko** [comment]
Good point. The following prevents those overflows by replacing `this.count()` (`Int`) with `sizeMinusOne` (`UInt`/`ULong`). Works for `(Int.MIN_VALUE..Int.MAX_VALUE).sum()` and `(Long.MIN_VALUE..Long.MAX_VALUE).sum()` (playground: https://pl.kotl.in/5tO2BKYaN):
```Kotlin
  val averageTimesTwo = this.first + this.last
  val sizeMinusOne = (this.last.toUInt() - this.first.toUInt()) / this.step.toUInt() // ULong for LongProgression
  return if (averageTimesTwo % 2 == 0) averageTimesTwo / 2 * (sizeMinusOne.toInt() + 1)
  else ((sizeMinusOne / 2u).toInt() + 1) * averageTimesTwo
```
`sizeMinusOne` because `Int.MAX_VALUE.toUInt() - Int.MIN_VALUE.toUInt() + 1u` overflows (whereas `Int.MAX_VALUE.toUInt() - Int.MIN_VALUE.toUInt()` doesn't).

PR #1168, opened 2017-07-02
[KT-15962] `orElse` extension that is a functional version of the Elvis operator

MGaetan89:
Related issue: https://youtrack.jetbrains.com/issue/KT-15962


qwwdfsad:
We still haven't decided whether this is a suitable addition to Kotlin's standard library

PR #1169, opened 2017-07-02
[KT-15695] String.nullIfEmpty

MGaetan89:
This PR adds the following extensions functions:
- `Array.nullIfEmpty()`
- `CharSequence.nullIfEmpty()`
- `Collection.nullIfEmpty()`
- `String.nullIfEmpty()`

It also fixes a couple of warning when running `./gradlew :tools:kotlin-stdlib-gen:run`.

Related issue: https://youtrack.jetbrains.com/issue/KT-15695


MGaetan89:
I'm using the same sample for every kind of `Array`. Do you prefer a dedicated sample for each kind?

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
```diff
@@ -187,5 +185,26 @@ fun arrays(): List<GenericFunction> {
...
+            else if (f == Collections)
+                "@sample samples.collections.Collections.nullIfEmpty"
+            else
+                ""
+            }
+            """
+        }
+        returns("SELF?")
+        body {
+            "return if (this == null || this.isEmpty()) null else this"
```
ilya-g: `this == null` 
^ expression is always false

libraries/tools/kotlin-stdlib-gen/src/templates/Elements.kt
```diff
@@ -14,7 +14,7 @@ fun elements(): List<GenericFunction> {
         doc { f -> "Returns `true` if [element] is found in the ${f.collection}." }
         typeParam("@kotlin.internal.OnlyInputTypes T")
         returns("Boolean")
-        body(Iterables) { f ->
+        body(Iterables) {
```
ilya-g: Please extract unrelated changes, such as formatting improvement in a separate PR.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
MGaetan89: You're right. I mixed it with the `String`/`CharSequence` implementation.
Should I unify both to apply on non-nullable type?
I will also revert the formatting changes.

MGaetan89:
@ilya-g I've remove all `this == null` checks, and made `nullIfEmpty()` extension available only on non-nullable `CharSequence` and `String`.

ilya-g:
> I'm using the same sample for every kind of Array. 

That's enough to give the usage example.

MGaetan89:
Is there something else to do for this to get merged?

ilya-g:
It would be nice to cover this with some tests, despite the fact they might sound trivial.
To get this merged, we should first discuss this function on a design meeting, and it could take some time to complete, so please be patient.

pun-ky:
Adding these methods is surprising for me because Kotlin rather encourages to use non-null types so that having that methods seems for me a little bit not a way to go 😉

ilya-g:
@pun-ky While Kotlin encourages working with non-nullable values it also provides expressive ways of dealing with nullable ones, e.g. safe call `?.` and elvis `?:` operators. It might be advantageous to turn an empty collection into `null` and then use these operators:

```
val keywords = getFilterKeywords().nullIfEmpty() ?: return
...
```

So there's nothing wrong with this feature request. 
Though in Kotlin 1.3 the issue [KT-15695](https://youtrack.jetbrains.com/issue/KT-15695) was implemented in a different way: as a more general purpose function `ifEmpty`, so that `ifEmpty { null }` could be used instead of `nullOfEmpty()`.

pun-ky:
Seems legit 👍 especially when ifEmpty is accepting value provider / lambda. Cool!

PR #1171, opened 2017-07-04
Add org.gradle.api.Nullable to NULLABLE_ANNOTATIONS list

eskatos:
See https://github.com/gradle/gradle/blob/master/subprojects/base-services/src/main/java/org/gradle/api/Nullable.java

It predates `javax.annotation` and is used in the Gradle Java API.
Having this annotation acknowledged by Kotlin will make working with the [Gradle Kotlin DSL](https://github.com/gradle/kotlin-dsl) nicer.

The change is pretty trivial, please tell me if some coverage is needed and where to add it.


eskatos:
Closing in favor of: https://github.com/gradle/gradle/issues/2382

PR #1172, opened 2017-07-04
Report MemberVisibilityCanBePrivate on visibility modifier if present

nd:


yole:
Please don't forget to rerun affected tests when you make a change; this change breaks the quickfix tests for this inspection. (I've fixed them.)_

PR #1173, opened 2017-07-05
KT-12594: Add inspection to detect redundant step 1

Knize:


idea/src/META-INF/plugin.xml
```diff
@@ -2278,6 +2278,14 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.RedundantProgressionStepInspection"
+                     displayName="Redundant progression step"
+                     groupName="Kotlin"
```
yole: Please provide a group path and group name for the inspection according to https://github.com/JetBrains/kotlin/commit/a09b31b5680952c56bb3ea5bbcb7939084e9f027

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantProgressionStepInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+
+class RedundantProgressionStepInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitBinaryExpression(expression: KtBinaryExpression) {
+                if (isStepOneCall(expression)) {
+                    holder.registerProblem(expression,
+                                           "Iteration step is redundant. Recommended to remove it.",
```
yole: No need to say "Recommended..." in the problem description; "redundant" implies that already.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantProgressionStepInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        fun isStepOneCall(expression: KtBinaryExpression): Boolean {
+            val left = expression.left as? KtBinaryExpression ?: return false
+            if (left.operationToken != KtTokens.RANGE) return false
+            if (expression.operationReference.text != "step") return false
+            val right = expression.right as? KtConstantExpression ?: return false
+            val rightConst = right.firstChild as? LeafPsiElement ?: return false
+            if (rightConst.elementType != KtTokens.INTEGER_LITERAL) return false
+
+            if (rightConst.text !in setOf("1", "1L")) return false
```
yole: Please use ConstantExpressionEvaluator.getConstant() instead.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantProgressionStepInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+            if (!isStepOneCall(expression)) return
+            expression.replaced(expression.left!!)
+        }
+    }
+
+    companion object {
+        private val STEP_FUNCTION_FQ_NAME = "kotlin.ranges.step"
+
+        fun isStepOneCall(expression: KtBinaryExpression): Boolean {
+            val left = expression.left as? KtBinaryExpression ?: return false
```
yole: You should use KtPsiUtil.deparenthesize() to handle the case when the range expression is in parentheses.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantProgressionStepInspection.kt
```diff
@@ -0,0 +1,89 @@
...
+
+
+    class EliminateStepOneFix : LocalQuickFix {
+        override fun getName() = "Eliminate redundant iteration step size"
+
+        override fun getFamilyName() = name
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            if (!FileModificationService.getInstance().preparePsiElementForWrite(descriptor.psiElement)) return
+            val expression = descriptor.psiElement as? KtBinaryExpression ?: return
```
yole: This only handles "step" called as an infix function; would be better to handle regular "step" calls too.

yole:
Just wanted to note that non-infix calls are still not handled. Please add support for them, and I'll merge.

vladimirdolzhenko:
Sorry for quite long delay in response to this PR.

Could you please update it according to the latest code base and add i18n support for messages ?

vladimirdolzhenko:
Shall we consider this PR as abandon ?

PR #1176, opened 2017-07-07
Initial commit for Boolean expression is const inspection.

vedun-z:
See: https://youtrack.jetbrains.com/issue/KT-18897

### DIFF
--- a/idea/resources/inspectionDescriptions/BooleanExpressionIsAlwaysConstant.html
+++ b/idea/resources/inspectionDescriptions/BooleanExpressionIsAlwaysConstant.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection reports boolean expressions that are always false or true.
+</body>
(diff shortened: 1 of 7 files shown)


mglukhikh:
Kotlin project already includes control-flow analysis tool which is located in `compiler/frontend/src/cfg` and used at this moment for initialization analysis and usage analysis. I think it is the perfect way to implement analysis like yours. Implementing the same analysis in an inspection is bad because you are repeating the same logic there (with less coverage of PSI elements) and also spending more time.

mglukhikh:
I will close this PR. If you have interest, you can try to implement the same feature using our CFA tool, it's relatively easy. If you have questions about this tool, you can ask me in comments to https://youtrack.jetbrains.com/issue/KT-18897 or directly by e-mail (mikhail.glukhikh@jetbrains.com).

PR #1177, opened 2017-07-07
Better quick-fix text when member has visibility modifier

nd:


idea/src/org/jetbrains/kotlin/idea/inspections/MemberVisibilityCanPrivateInspection.kt
```diff
@@ -112,9 +112,18 @@ class MemberVisibilityCanPrivateInspection : AbstractKotlinInspection() {
             else -> "Property"
         }
         val nameElement = (declaration as? PsiNameIdentifierOwner)?.nameIdentifier ?: return
-        holder.registerProblem(declaration.visibilityModifier() ?: nameElement,
+        val visibilityModifier = declaration.visibilityModifier()
+        val fix = object : AddModifierFix(modifierListOwner, KtTokens.PRIVATE_KEYWORD) {
```
yole: I would do this as a general fix to AddModifierFix (which already has special logic for handling modality modifiers), not as a single-case override here.

yole:
Already fixed in https://github.com/JetBrains/kotlin/commit/1ed1751b63dec3ef78bc9e1f69059e95c12648c0, sorry.

PR #1179, opened 2017-07-11
KT-18723 Correct "after" sample for intention Convert to apply

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18723


mglukhikh:
Thank you! (merged manually)

PR #1181, opened 2017-07-12
KT-18822 Do not report "property can be private" on properties of private classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18822


mglukhikh:
Thank you, will merge it

mglukhikh:
Merged

PR #1185, opened 2017-07-12
KT-18965 add quick-fix for USELESS_IS_CHECK

neworld:
Fix of: https://youtrack.jetbrains.com/issue/KT-18965

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveUselessIsCheckFixForIf.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveUselessIsCheckFixForIf.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.idea.quickfix
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.kotlin.diagnostics.Diagnostic
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtIsExpression
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.KtWhenEntry
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+
+class RemoveUselessIsCheckFixForIf(element: KtIsExpression) : KotlinQuickFixAction<KtIsExpression>(element) {
+    override fun getFamilyName() = "Remove useless is check"
+
+    override fun getText(): String = familyName
+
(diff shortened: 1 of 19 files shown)


idea/testData/quickfix/expressions/removeUselessIsCheckInWhen.kt
```diff
@@ -0,0 +1,6 @@
+// "Remove useless is check" "false"
+fun foo(a: String) {
+    when (1) {
+        <caret>is Int -> { }
```
mglukhikh: Well, it's an exceptional case but here we can also do something :). If a condition in when branch is definitely true, we can replace the condition with `else` and delete all subsequent branches because they are unreachable anyway. If the condition is definitely false (add test for this case), we can delete this when branch.

neworld: I would like to try by myself :) But I have bad feeling. For `if` case we are modifying only condition. Like
```diff
-if (1 is Boolean) { print() }
+if (false) { print() }
```
but not whole branch:
```diff
-if (1 is Boolean) { print() }
+
```

What are you asking is to change whole `when`. It could be optimized totally:
```diff
-when (1) {
-  is Boolean -> print()
-}
+
```
and 
```diff
-when (1) {
-  is Int -> print()
-}
+ print()
```

mglukhikh: I recommend never delete the whole when. For your examples, we should get

```
// false condition
when (1) {
}

// true condition
when (1) {
    else -> print()
}
```

I don't think it's something dangerous because user can always undo the optimization.

Keep in mind also when without subject case. Here we can just replace conditions into `true` and `false`, as for simple conditions.

neworld:
Sorry for being late. Please review.

idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveUselessIsCheckFixForWhen.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val condition = element ?: return
+        val whenEntry = condition.parent as KtWhenEntry
+        val whenExpression = whenEntry.parent as KtWhenExpression
+
+        if (condition.isNegated) {
+            condition.parent.delete()
+        } else {
+            whenExpression.entries.forEach { it.delete() }
```
mglukhikh: You should not delete all entries here, but only subsequent ones. E.g.

```
val something: Base = ...
when (something) {
    is Derived -> forDerived()
    is Base /* always true */ -> forBase()
    is Any /* unreachable */ -> forAny()
    is SomethingWrong /* also unreachable */ forWrong()
    else -> forElse()
}
```

Here `is Base` branch must be left intact.

idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveUselessIsCheckFixForIf.kt
```diff
@@ -18,13 +18,15 @@ package org.jetbrains.kotlin.idea.quickfix
...
+import com.intellij.psi.util.PsiTreeUtil
 import org.jetbrains.kotlin.diagnostics.Diagnostic
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtIsExpression
 import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.KtWhenEntry
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
 
-class RemoveUselessIsCheckFix(element: KtIsExpression) : KotlinQuickFixAction<KtIsExpression>(element) {
+class RemoveUselessIsCheckFixForIf(element: KtIsExpression) : KotlinQuickFixAction<KtIsExpression>(element) {
```
mglukhikh: Not a good name, may be it's better to leave it as before. This action is applicable also for when without subject, or just for code like `val a = b is Any`.

mglukhikh:
Merged, thank you.

PR #1186, opened 2017-07-12
KT-18979 Add val to proper place for 'Add val/var' intent

neworld:
Fix of https://youtrack.jetbrains.com/issue/KT-18979


asedunov:
Merged. Thank you for the fix

PR #1187, opened 2017-07-13
`KotlinCommonIntentionActionsFactory` is able to remove constructors parameters

xiexed:
a fix for EA-104621

### DIFF
--- a/idea/tests/org/jetbrains/kotlin/idea/quickfix/CommonIntentionActionsTest.kt
+++ b/idea/tests/org/jetbrains/kotlin/idea/quickfix/CommonIntentionActionsTest.kt
@@ -168,6 +168,20 @@ class CommonIntentionActionsTest : LightPlatformCodeInsightFixtureTestCase() {
         """.trim().trimMargin(), true)
     }
 
+    fun testRemoveConstructorParameters() {
+        myFixture.configureByText("foo.kt", """
+        |class <caret>Foo(i: Int) {
+        |}
+        """.trim().trimMargin())
+
+        myFixture.launchAction(codeModifications.createAddCallableMemberActions(MethodInsertionInfo.constructorInfo(
+                atCaret<UClass>(myFixture), makeParams())).findWithText("Remove 1st parameter from method 'Foo'"))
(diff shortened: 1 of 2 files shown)


idea/tests/org/jetbrains/kotlin/idea/quickfix/CommonIntentionActionsTest.kt
```diff
@@ -168,6 +168,20 @@ class CommonIntentionActionsTest : LightPlatformCodeInsightFixtureTestCase() {
...
     }
 
+    fun testRemoveConstructorParameters() {
+        myFixture.configureByText("foo.kt", """
+        |class <caret>Foo(i: Int) {
+        |}
+        """.trim().trimMargin())
+
+        myFixture.launchAction(codeModifications.createAddCallableMemberActions(MethodInsertionInfo.constructorInfo(
+                atCaret<UClass>(myFixture), makeParams())).findWithText("Remove 1st parameter from method 'Foo'"))
```
yole: It's better to change the intention text so that it says "constructor" and not "method".

xiexed: It is not an easy task, because it uses `ChangeMethodSignatureFromUsageFix` from Java, which is hardcoded to "method". It is possible to extend it of course in Kotlin and reimplement functionality. But it sounds like another task.

yole: I think it's worth fixing PsiFormatUtil to handle constructors correctly, but that's a separate task indeed.

xiexed: Ok. So could I now cherry-pick it to `1.1.4_172`-branch? Or what are the relationships between `master_172` and `1.1.4_172` now?

yole: Yes, please do

PR #1189, opened 2017-07-13
KT-13870 Wrong caption "Change to property access" for Quick Fix to convert class instantiation to object reference

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13870


yole:

PR #1192, opened 2017-07-14
KT-12195 Quickfix @JvmStatic on main() method in an object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12195

### DIFF
--- a/idea/resources/intentionDescriptions/AddJvmStaticIntention/after.kt.template
+++ b/idea/resources/intentionDescriptions/AddJvmStaticIntention/after.kt.template
@@ -1,0 +1,3 @@
+<spot>@JvmStatic</spot>
+fun main(args: Array<String>) {
+}
--- a/idea/resources/intentionDescriptions/AddJvmStaticIntention/after.kt.template
+++ b/idea/resources/intentionDescriptions/AddJvmStaticIntention/after.kt.template
@@ -1,3 +1,5 @@
-<spot>@JvmStatic</spot>
-fun main(args: Array<String>) {
+object Test {
+    <spot>@JvmStatic</spot>
+    fun main(args: Array<String>) {
+    }
 }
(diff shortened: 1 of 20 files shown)


idea/resources/intentionDescriptions/AddJvmStaticIntention/after.kt.template
```diff
@@ -0,0 +1,3 @@
+<spot>@JvmStatic</spot>
```
yole: The template shows a top-level method, not a method inside an `object`. Please update it to show what the intention action actually does.

idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmStaticIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+class AddJvmStaticIntention : SelfTargetingIntention<KtNamedFunction>(
+    KtNamedFunction::class.java,
+    "Add '@JvmStatic' annotation"
+), LowPriorityAction {
+
+    private val annotationFqName = FqName("kotlin.jvm.JvmStatic")
+
+    override fun isApplicableTo(element: KtNamedFunction, caretOffset: Int): Boolean {
+        if (element.findAnnotation(annotationFqName) != null) return false
+        if (!UnusedSymbolInspection.isEntryPoint(element)) return false
```
yole: See `KotlinRunLineMarkerContributor` for the correct way to check that a function is `main`.

idea/src/org/jetbrains/kotlin/idea/intentions/AddJvmStaticIntention.kt
```diff
@@ -0,0 +1,47 @@
...
+        element.parent.parent.run {
```
yole: Please use `KtDeclaration.containingClassOrObject`.

t-kameyama:
I fixed. Please review.

yole:

PR #1193, opened 2017-07-14
KT-12504 Intention to make open class with only private constructors sealed

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12504

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
@@ -18,9 +18,9 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
-import com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.lexer.KtTokens.*
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtPrimaryConstructor
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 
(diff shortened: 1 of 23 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
```diff
@@ -0,0 +1,65 @@
...
+
+class ConvertClassToSealedClassIntention : SelfTargetingRangeIntention<KtClass>(KtClass::class.java, "Convert to sealed class") {
+
+    override fun applicabilityRange(element: KtClass): TextRange? {
+        if (element.modifierList == null) return null
+        if (!element.hasModifier(OPEN_KEYWORD) && !element.hasModifier(ABSTRACT_KEYWORD)) return null
+
+        val constructors = listOfNotNull(element.primaryConstructor) + element.secondaryConstructors
+        if (constructors.isEmpty()) return null
+        if (constructors.any { !it.hasModifier(PRIVATE_KEYWORD) }) return null
```
yole: Please use `none` instead of `any` with a negative condition.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
```diff
@@ -0,0 +1,65 @@
...
+        return TextRange(element.startOffset, nameIdentifier.endOffset)
+    }
+
+    override fun applyTo(element: KtClass, editor: Editor?) {
+        val newElement = element.copy() as KtClass
+
+        newElement.modifierList?.run {
+            getModifier(OPEN_KEYWORD)?.delete()
+            getModifier(ABSTRACT_KEYWORD)?.delete()
+            newElement.addModifier(SEALED_KEYWORD)
```
yole: It's better to move this line out of the `run` block.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
```diff
@@ -0,0 +1,65 @@
...
+        val constructors = listOfNotNull(element.primaryConstructor) + element.secondaryConstructors
+        if (constructors.isEmpty()) return null
+        if (constructors.any { !it.hasModifier(PRIVATE_KEYWORD) }) return null
+
+        val nameIdentifier = element.nameIdentifier ?: return null
+        return TextRange(element.startOffset, nameIdentifier.endOffset)
+    }
+
+    override fun applyTo(element: KtClass, editor: Editor?) {
+        val newElement = element.copy() as KtClass
```
yole: It doesn't seem necessary to create a copy; you should be able to apply all the changes to the original `KtClass` directly.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        newElement.modifierList?.run {
+            getModifier(OPEN_KEYWORD)?.delete()
+            getModifier(ABSTRACT_KEYWORD)?.delete()
+            newElement.addModifier(SEALED_KEYWORD)
+        }
+
+        newElement.primaryConstructor?.run {
+            modifierList?.getModifier(PRIVATE_KEYWORD)?.delete()
+            getConstructorKeyword()?.delete()
```
yole: The `constructor` keyword can only be deleted if there are no annotations or other modifiers on the constructor. Please add a test for the case with annotations.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertClassToSealedClassIntention.kt
```diff
@@ -0,0 +1,65 @@
...
+            if (newElement.secondaryConstructors.isEmpty() && valueParameters.isEmpty()) valueParameterList?.delete()
+            (newElement.nameIdentifier?.prevSibling as? PsiWhiteSpace)?.delete()
```
yole: It should never be necessary to do that manually; the formatter should take care of whitespaces.

idea/testData/intentions/convertClassToSealedClass/abstract.kt.after
```diff
@@ -0,0 +1,3 @@
+// INTENTION_TEXT: "Convert to sealed class"
```
yole: Since you aren't changing the intention text dynamically, there's no need to include INTENTION_TEXT in testdata files.

t-kameyama:
I fixed. Please review.

yole:

PR #1194, opened 2017-07-14
KT-13552.

Dimach:


j2k/src/org/jetbrains/kotlin/j2k/SwitchConverter.kt
```diff
@@ -38,17 +37,27 @@ class SwitchConverter(private val codeConverter: CodeConverter) {
...
         var pendingSelectors = ArrayList<WhenEntrySelector>()
+        var defaultSelector: WhenEntrySelector? = null
+        var defaultEntry: WhenEntry? = null
         for ((i, case) in cases.withIndex()) {
             if (case.label == null) { // invalid switch - no case labels
                 result.add(WhenEntry(listOf(ValueWhenEntrySelector(Expression.Empty).assignNoPrototype()), convertCaseStatementsToBody(cases, i)).assignNoPrototype())
                 continue
             }
-            pendingSelectors.add(codeConverter.convertStatement(case.label) as WhenEntrySelector)
+            (codeConverter.convertStatement(case.label) as WhenEntrySelector).run {
```
semoro: I'd extracted selector to a variable, then smart-cast on #54 will calculate properly.

PR #1195, opened 2017-07-15
KT-18855 Convert 'Remove @ from annotation argument' to compiler error & quick-fix

neworld:
https://youtrack.jetbrains.com/issue/KT-18855


compiler/frontend/src/org/jetbrains/kotlin/resolve/AnnotationResolverImpl.java
```diff
@@ -141,13 +141,28 @@ public static void checkAnnotationType(
         }
     }
 
+    @Nullable
+    private static <T extends PsiElement> T findAncestorOfType(@NotNull PsiElement element, Class<T> ancestorType) {
```
mglukhikh: There is already `getParentOfType`, please use it here

neworld:
I fixed. Please review.

mglukhikh:
Merged manually, thank you.

PR #1196, opened 2017-07-16
KT-18722 Correct "before" sample in description for intention Convert to enum class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18722


yole:
Thanks!

PR #1197, opened 2017-07-18
Do not report "property can be private" on JvmField properties

nd:


nd:
Kotlin plugin started from Idea works correctly, but for some reason test fails to resolve the 'kotlin.jvm.JvmField' class. @mglukhikh do you know why it might happen?

mglukhikh:
Rewritten and added to master. About resolve of `JvmField`, it's some inspection test artifact. I recommend to use tests from `inspectionsLocal` group because they are easier to debug and more flexible. Also, the function to detect `JvmField` annotation already exists.

PR #1198, opened 2017-07-18
KT-18738 Misleading quick fix message for an 'open' modifier on an interface member

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18738


mglukhikh:
Thank you!

PR #1200, opened 2017-07-20
KT-18518

Dimach:
Rewritten polyadic expression converter for correct type handling, KT-18518 fixed.

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
@@ -48,7 +48,7 @@ interface SpecialExpressionConverter {
 }
 
 fun ExpressionConverter.withSpecialConverter(specialConverter: SpecialExpressionConverter): ExpressionConverter {
-    return object: ExpressionConverter {
+    return object : ExpressionConverter {
         override fun convertExpression(expression: PsiExpression, codeConverter: CodeConverter)
                 = specialConverter.convertExpression(expression, codeConverter) ?: this@withSpecialConverter.convertExpression(expression, codeConverter)
     }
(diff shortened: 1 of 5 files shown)


j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -241,7 +241,8 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
         }
         val name = if (primitiveType != null) {
             "javaPrimitiveType"
-        } else {
+        }
+        else {
```
lynnfield: It is wrong according to Java Coding Conventions

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -867,7 +912,8 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                 if (parameters.size == 1 && !isKotlinFunctionType) {
                     // for lambdas all parameters with types should be present
                     emptyList()
-                } else {
+                }
+                else {
```
piotrek1543: Please restore previous code formatting - actual `if-else` statements and spaces are against accepted conceptions.

yole: @piotrek1543 Please let the Kotlin team decide on the coding conventions used in the Kotlin project. Thanks.

goodwinnk:
@Dimach Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

KT-18518 is still reproducible in the new Java to Kotlin converter that is especially sad because code from this pull request can't be reused directly to fix the issue. Unfortunately fixing should be started from the ground.

PR #1201, opened 2017-07-20
CharSequence#split documentation corrected

koral--:
Newline appears as no character in rendered documentation. Changed to space like in the rest of `split` methods.

### DIFF
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1176,8 +1176,7 @@ public fun CharSequence.split(vararg delimiters: Char, ignoreCase: Boolean = fal
 /**
  * Splits this char sequence around matches of the given regular expression.
  *
- * @param limit Non-negative value specifying the maximum number of substrings to return.
- * Zero by default means no limit is set.
+ * @param limit Non-negative value specifying the maximum number of substrings to return. Zero by default means no limit is set.
  */
 @kotlin.internal.InlineOnly
 public inline fun CharSequence.split(regex: Regex, limit: Int = 0): List<String> = regex.split(this, limit)


ilya-g:
Could you explain why this fix is required? As far as I can see here https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/split.html there's no problems with this particular overload.

koral--:
In Android Studio it is rendered without any spacing:
image

ilya-g:
Then it seems to be a bug in Kotlin IDE plugin, please file it to http://kotl.in/issue

koral--:
OK, will do. However two little questions:
1. On the page you linked, the 1st overload (using newline in sources) looks like that:
image
While the last one (using space):
image
Is it another bug, if so where it should be reported/fixed?

2. All other overloads use newline, shouldn't style be unified?

ilya-g:
That's strange, 'cause I don't see a newline in the `limit` parameter in the sources of the first overload (they could be navigated to right from the docs page). Do you?

We should investigate where that new paragraph comes from, when the page is built.

koral--:
>  I don't see a newline in the limit parameter in the sources of the first overload (they could be navigated to right from the docs page). Do you?

Indeed there is no newline, sorry, those 2 overloads confused me.

So to sum up:
Here: https://github.com/JetBrains/kotlin/blob/1.1.3/libraries/stdlib/src/kotlin/text/Strings.kt#L1180 is the newline which is stripped in IDE but looks OK on the page: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/split.html#split (the last overload). It is reported as KT-19153.

Here: https://github.com/JetBrains/kotlin/blob/1.1.3/libraries/stdlib/src/kotlin/text/Strings.kt#L1147 is the space which looks OK in IDE but on the page (first overload), it causes unneeded extra line break: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/split.html#split

Note that the line above problematic one: https://github.com/JetBrains/kotlin/blob/1.1.3/libraries/stdlib/src/kotlin/text/Strings.kt#L1146 has one more character and is displayed without issues. So the line length seems to be unrelated.
On the other hand, only that one overload broken on the page has a blank line and additional text below the last parameter: https://github.com/JetBrains/kotlin/blob/1.1.3/libraries/stdlib/src/kotlin/text/Strings.kt#L1149 may this is the cause?

BTW shouldn't that text https://github.com/JetBrains/kotlin/blob/1.1.3/libraries/stdlib/src/kotlin/text/Strings.kt#L1149-L1151 be located above parameters?

koral--:
Fixed in https://youtrack.jetbrains.com/issue/KTIJ-6702 and formatting looks ok in stdlib 1.5.30.

PR #1202, opened 2017-07-21
KT-18852 "Lift return out of when" does not work for exhaustive when without else

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18852


mglukhikh:
Thank you, it looks good.

mglukhikh:
Merged manually

PR #1205, opened 2017-07-21
"Use destructuring declaration" should not be available for top-level and class properties

cypressious:
Fixes #KT-19167


yole:

PR #1206, opened 2017-07-22
Add inspection for when with only else

cypressious:
Fixes #KT-12523

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/WhenWithOnlyElseInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/WhenWithOnlyElseInspection.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/WhenWithOnlyElseInspection.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.psi.KtVisitorVoid
+import org.jetbrains.kotlin.psi.KtWhenExpression
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
+
+class WhenWithOnlyElseInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+            override fun visitWhenExpression(expression: KtWhenExpression) {
+                super.visitWhenExpression(expression)
```
yole: You don't actually need the `super` call here.

idea/src/org/jetbrains/kotlin/idea/inspections/WhenWithOnlyElseInspection.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    private class SimplifyFix(
+            private val isUsedAsExpression: Boolean
+    ) : LocalQuickFix {
+        override fun getFamilyName() = name
+
+        override fun getName() = "Simplify expression"
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val whenExpression = descriptor.psiElement as? KtWhenExpression ?: return
```
yole: Please don't forget `FileModificationService.getInstance().preparePsiElementForWrite()`

cypressious:
@yole fixed after review and rebased

PR #1208, opened 2017-07-24
KT-18826 INAPPLICABLE_LATEINIT_MODIFIER should have a quickfix to remove initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18826


idea/src/org/jetbrains/kotlin/idea/quickfix/RemovePartsFromPropertyFix.java
```diff
@@ -135,4 +136,23 @@ public static KotlinSingleIntentionActionFactory createFactory() {
             }
         };
     }
+
+    public static KotlinSingleIntentionActionFactory createLateInitFactory() {
```
yole: Please convert this file to Kotlin if you're adding substantial new code to it. (Rename to .kt, commit, then rename back to .java, run J2K, make any changes needed for the project to compile, commit again, then make your changes.)

idea/src/org/jetbrains/kotlin/idea/quickfix/RemovePartsFromPropertyFix.java
```diff
@@ -135,4 +136,23 @@ public static KotlinSingleIntentionActionFactory createFactory() {
...
+        return new KotlinSingleIntentionActionFactory() {
+            @Override
+            public KotlinQuickFixAction<KtProperty> createAction(@NotNull Diagnostic diagnostic) {
+                PsiElement element = Errors.INAPPLICABLE_LATEINIT_MODIFIER.cast(diagnostic).getPsiElement();
+                KtProperty property = PsiTreeUtil.getParentOfType(element, KtProperty.class);
+                if (property == null) return null;
+
+                boolean hasInitializer = property.hasInitializer();
+                boolean hasGetter = property.getGetter() != null && property.getGetter().getBodyExpression() != null;
```
yole: The getter and setter checks are not needed. This quickfix should apply to any lateinit property; _late_init means that it's initialized somewhere in the class, and this applies to regular properties with default getters and setters too.

t-kameyama: I think the getter and setter checks are needed. This quickfix apply to only lateinit property because the element get from INAPPLICABLE_LATEINIT_MODIFIER.

t-kameyama:
Fixed. Please review.

yole:
Not sure why you think the getter and setter checks are needed. I want the quickfix to appear in the most basic case:

```
class Foo {
    lateinit var s = ""
}
```

The quickfix should add an explicit property type and remove the initializer.

t-kameyama:
I got it. Is it not necessary to delete not allowed getter or setter?
```
class Foo {
    lateinit var str: String
        get() = "foo"
        set(value) {}
}
```

yole:

PR #1209, opened 2017-07-24
Full Date class for JS

rostopira:
I think that it's essential part of any language. How can you code without Date class?

### DIFF
--- a/js/js.libraries/src/core/date.kt
+++ b/js/js.libraries/src/core/date.kt
@@ -4,5 +4,109 @@ package kotlin.js
  * Exposes the [Date API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) to Kotlin.
  */
 public external class Date() {
-    public fun getTime(): Double
+
+    constructor(timestamp: Long)
+    constructor(dateString: String)
+    constructor(year: Int,
+                month: Int = definedExternally,
+                date: Int = definedExternally,
+                hours: Int = definedExternally,
+                minutes: Int = definedExternally,
+                seconds: Int = definedExternally,
+                millis: Int = definedExternally)
+
+    public fun getDate(): Int = definedExternally
+    public fun getDay(): Int = definedExternally
+    public fun getFullYear(): Int = definedExternally
+    public fun getHours(): Int = definedExternally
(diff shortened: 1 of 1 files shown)


ilya-g:
While we agree that having date/time API is essential, we don't think that JS Date is an example of a good API design.

We have plans to provide typed Kotlin definitions for https://www.npmjs.com/package/js-joda. This way one will be able to use the same date/time API in JS and JVM.

bashor:
@ilya-g, I think, it would be nice to improve existing declaration anyway.

js/js.libraries/src/core/date.kt
```diff
@@ -4,5 +4,110 @@ package kotlin.js
...
+                              options: dynamic = definedExternally): String = definedExternally
+    public fun toLocaleTimeString(locales: dynamic = definedExternally, /* String | Array<String> */
+                                  options: dynamic = definedExternally): String = definedExternally
+    public override fun toString(): String = definedExternally
+    public fun toTimeString(): String = definedExternally
+    public fun toUTCString(): String = definedExternally
+    public fun valueOf(): Long = definedExternally
+
+    companion object {
+        fun now(): Long = definedExternally
```
bashor: I think return type should not be `Long` here since it's not native JS type and it's emulated in Kotlin. Probably it should be `Number` or `Double`.

bashor: Similarly, all usages of Long in this file should be reviewed. Also, sometimes it can be `Int`

rostopira: Returns cannot be Int because, according to docs, all returns is a timestamp since 1970 in milliseconds. Also, it's bad to treat it is as Double, because you can't than convert it to Date without ugly cast .toLong(). Same thing with Number, ugly casts all over the code

ilya-g:
The most of the missing members of JS date were introduced in the commit 43302ce00

devulex:
Hi! Why is this commit not included in release 1.2.0?
Now in the release there is only a getTime() method.
```
public external class Date() {
    public fun getTime(): Double
}
```

bashor:
@devulex the changes will be available since 1.2.20

devulex:
Thanks!

PR #1210, opened 2017-07-24
Script-runtime 

pTalanov:


core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
```diff
@@ -20,6 +20,15 @@ package kotlin.script.dependencies
 
 import java.io.File
 
+// discuss
+
+// Currently File is used to disambiguate between plain strings and paths
+// One of the ideas is to use java.nio.Path but this requires JDK 7
+// that means that script templates would require higher JDK (but since script are run by calling koltinc it seems ok to me after consideration)
+// Andrey expressed the idea that File (or Path) does not support other protocols, should we use URL/URI? (is it viable to support non-file protocols in compiler and IDE?)
```
udalov: > File (or Path) does not support other protocols

Path _does_ support other protocols AFAIK. You can create a custom `java.nio.file.FileSystem` and resolve paths within that file system.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -21,6 +21,24 @@ package kotlin.script.dependencies
...
 import kotlin.script.dependencies.DependenciesResolver.ResolveResult
 
+//discuss
+
+// Provides api to discover dependencies of scripts
+// Dependencies can be content-dependent
+//
+// Some concerns on naming:
+// Environment -> ScriptEnvironment (top level class with too common a name)
+// ResolveResult -> ResolutionResult
```
udalov: +1

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
```diff
@@ -20,6 +20,15 @@ package kotlin.script.dependencies
 
 import java.io.File
 
+// discuss
+
+// Currently File is used to disambiguate between plain strings and paths
+// One of the ideas is to use java.nio.Path but this requires JDK 7
+// that means that script templates would require higher JDK (but since script are run by calling koltinc it seems ok to me after consideration)
```
udalov: I think requiring JDK 8 in scripts is fine

core/script.runtime/src/kotlin/script/templates/standard/templates.kt
```diff
@@ -16,18 +16,25 @@
 
 package kotlin.script.templates.standard
 
+// discuss
+//
+// These are some 'basic' script templates
+// Should we keep them here?
```
udalov: I think we should do the following with these:
* `SimpleScriptTemplate` -> remove (script with args can be always used instead of it)
* `kotlin.script.templates.standard.ScriptTemplateWithArgs` -> rename and move to `kotlin.script.templates.StandardScriptTemplate`
* `kotlin.script.templates.standard.ScriptTemplateWithBindings` -> move to `kotlin.script.templates.ScriptTemplateWithBindings`

core/script.runtime/src/kotlin/script/templates/annotations.kt
```diff
@@ -22,13 +22,28 @@ import kotlin.reflect.KClass
...
+// @file: ExtraCompilerArgs("-enableWhatever")
+
 const val DEFAULT_SCRIPT_FILE_PATTERN = ".*\\.kts"
 
+// classes annotated with ScriptTemplateDefinition become script templates.
+// examples: https://github.com/JetBrains/kotlin/blob/5faad493b4cf7bf33bf82475e966a99a8e835720/compiler/tests/org/jetbrains/kotlin/scripts/ScriptTemplateTest.kt#L564-L601
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class ScriptTemplateDefinition(val resolver: KClass<out ScriptDependenciesResolver> = NoDependencies::class,
                                           val scriptFilePattern: String = DEFAULT_SCRIPT_FILE_PATTERN)
```
goodwinnk: Do we need `scriptFilePattern` in this restricted form at all? 
- IDE uses `ScriptTemplatesProvider.filePattern` that allows to determine scripts without loading script definition annotation.
- There might be scripts of different kind that differentiate by full path or maybe content.

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
```diff
@@ -16,6 +16,11 @@
 
 package kotlin.script.extensions
 
+// discuss
+
+// Is this an appropriate place to put this class?
```
udalov: This doesn't look like an appropriate place. If I understand correctly, it has nothing to do with scripts, so putting it into package `kotlin.script.extensions` is risky. As soon as we decide to publish it as a supported compiler plugin, we'll have to move it to another package and artifact, causing incompatibilities, so maybe it'd be better to do this earlier.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -47,18 +66,23 @@ interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesReso
...
+// ScriptReport -> ScriptDiagnostic
 data class ScriptReport(val message: String, val severity: Severity = ScriptReport.Severity.ERROR, val position: Position? = null) {
     data class Position(val startLine: Int, val startColumn: Int, val endLine: Int? = null, val endColumn: Int? = null)
     enum class Severity { ERROR, WARNING, INFO, DEBUG }
 }
 
-fun ScriptDependencies.asSuccess(): ResolveResult.Success = ResolveResult.Success(this)
\ No newline at end of file
+// should we expose this helper?
+fun ScriptDependencies.asSuccess(): ResolutionResult.Success = ResolutionResult.Success(this)
```
udalov: I think, we should not

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -47,18 +66,23 @@ interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesReso
     }
 }
 
+// is File a could type to use here?
+// No way to get script name if file is not present, should add another property (val fileName: String)
 interface ScriptContents {
     val file: File?
     val annotations: Iterable<Annotation>
```
udalov: Maybe use `Collection` here?

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -47,18 +66,23 @@ interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesReso
...
     val text: CharSequence?
```
udalov: Why not `String?`?

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -32,6 +50,7 @@ interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesReso
 
     sealed class ResolveResult {
         abstract val dependencies: ScriptDependencies?
+        // reports -> diagnostics
```
udalov: +1 for `val diagnostics: List<ScriptDiagnostic>`

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -32,6 +50,7 @@ interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesReso
 
     sealed class ResolveResult {
```
udalov: I'd simplify this by assuming that the result is "success" iff there are no errors:
```
data class ResolutionResult(
    val dependencies: ScriptDependencies, // ScriptDependencies.Empty in case of failure
    val diagnostics: List<ScriptDiagnostic>
) {
    val isSuccess: Boolean get() = diagnostics.none { it.severity == ERROR } // maybe an extension
}
```

core/script.runtime/src/kotlin/script/templates/annotations.kt
```diff
@@ -22,13 +22,28 @@ import kotlin.reflect.KClass
 import kotlin.script.dependencies.DependenciesResolver.NoDependencies
 import kotlin.script.dependencies.ScriptDependenciesResolver
 
+// discuss
+
+// One of the requests is to allow to share compiler args via script templates: https://youtrack.jetbrains.com/issue/KT-19120
+// Making it a property of ScriptTemplateDefinition is a viable options but it doesn't allow to changes compiler arguments based on script content.
+// For example, it could be possible to allow file level annotation and use it like:
+// @file: ExtraCompilerArgs("-enableWhatever")
```
udalov: I think it's fine if we don't support extra compiler arguments based on script content now. If this becomes really needed, we can support it in the future, e.g. with an annotation like you mentioned here.

core/script.runtime/src/kotlin/script/templates/annotations.kt
```diff
@@ -22,13 +22,28 @@ import kotlin.reflect.KClass
 import kotlin.script.dependencies.DependenciesResolver.NoDependencies
 import kotlin.script.dependencies.ScriptDependenciesResolver
 
+// discuss
```
udalov: Maybe move this file to package `kotlin.script.annotations`?

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
udalov: Also maybe rename to `fileAnnotations` to highlight that those are only `@file:` use-site-targeted annotations, but I'm not sure if that's necessary

core/script.runtime/src/kotlin/script/templates/annotations.kt
```diff
@@ -22,13 +22,28 @@ import kotlin.reflect.KClass
...
 const val DEFAULT_SCRIPT_FILE_PATTERN = ".*\\.kts"
 
+// classes annotated with ScriptTemplateDefinition become script templates.
+// examples: https://github.com/JetBrains/kotlin/blob/5faad493b4cf7bf33bf82475e966a99a8e835720/compiler/tests/org/jetbrains/kotlin/scripts/ScriptTemplateTest.kt#L564-L601
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class ScriptTemplateDefinition(val resolver: KClass<out ScriptDependenciesResolver> = NoDependencies::class,
                                           val scriptFilePattern: String = DEFAULT_SCRIPT_FILE_PATTERN)
 
+// usage example: https://github.com/JetBrains/kotlin/blob/88652154c96402475d42ae0496ab0b423cc0a2b2/libraries/tools/kotlin-script-util/src/main/kotlin/org/jetbrains/kotlin/script/util/resolve.kt#L42-L42
```
udalov: I'm confused, why is this needed? Can't the DependenciesResolver implementation just filter out the unneeded annotations from ScriptContents.annotations?

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
```diff
@@ -20,6 +20,15 @@ package kotlin.script.dependencies
...
+// Andrey expressed the idea that File (or Path) does not support other protocols, should we use URL/URI? (is it viable to support non-file protocols in compiler and IDE?)
+//
+// Explicitly references javaHome, what if it's not a jvm targeted script?
+// Currently it's convenient for IDE code and for the user because including jdk classes in classpath is a mess
 data class ScriptDependencies(
         val javaHome: File? = null,
```
udalov: What is `javaHome`: is it required to be a path to the JDK, or a path to the JRE is enough?

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
pTalanov: Okay, say there that resolver has failed with an exception (reports an error and returns ScriptDependencies.Empty). Should we update script highlighting in IDE in that case? We should display an error for sure but is marking all references as unresolved actually helpful? I believe that API you propose doesn't allow to handle this scenario.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
```diff
@@ -20,6 +20,15 @@ package kotlin.script.dependencies
...
         val classpath: List<File> = emptyList(),
```
zarechenskiy: Can we use here directories/jars/individual files? 
Maybe it's worth to document it somewhere

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
```diff
@@ -20,6 +20,15 @@ package kotlin.script.dependencies
...
+// discuss
+
+// Currently File is used to disambiguate between plain strings and paths
+// One of the ideas is to use java.nio.Path but this requires JDK 7
+// that means that script templates would require higher JDK (but since script are run by calling koltinc it seems ok to me after consideration)
+// Andrey expressed the idea that File (or Path) does not support other protocols, should we use URL/URI? (is it viable to support non-file protocols in compiler and IDE?)
+//
+// Explicitly references javaHome, what if it's not a jvm targeted script?
+// Currently it's convenient for IDE code and for the user because including jdk classes in classpath is a mess
 data class ScriptDependencies(
```
udalov: What's the difference between `sources` and `scripts`?

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
pTalanov: Honestly, everything should be documented. We can reference jars and directories, not sure about files.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
pTalanov: `sources` means source roots. Enables navigation in IDE.
`scripts` means some script files this script depends on. Not sure what it means exactly. It isn't supported in IDE (nor compiler I think).

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
udalov: You're right, we shouldn't mark everything as unresolved in this case. But I still think we should avoid hierarchies here. For example, we can make `isSuccess` a boolean property
```
data class ResolutionResult(
    val isSuccess: Boolean,
    val dependencies: ScriptDependencies,
    val diagnostics: List<ScriptDiagnostic>
)
```

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
pTalanov: `annotations` are actually lazily loaded (we need reflection to do this so it's relatively expensive). `Iterable` indicates that there is no container yet that holds all the instances. I think `Sequence` is better suited to express `lazy` and `no-container` ideas.
I like `filleAnnotations` btw. We just have to make sure that we have no plans to expand the meaning.

udalov: Users are likely to call `.isEmpty` on it, in which case it's still probably fully computed eagerly.

I'm fine with either `Collection` or `Sequence` or `Stream`. `Iterable` looks weird though

core/script.runtime/src/kotlin/script/templates/annotations.kt
pTalanov: It can. That's why i'm not sure if that's needed at all. I think we have to wait for @ligee to comment on what's the initial purpose of this. 
From IDE point of view I can say that since it has to construct ScriptContents object synchronously before sending (potentially) async request we can't achieve true laziness when loading annotations since the file may have changed by the time client calls ScriptContents::annotations (moreover we can't read file content from the thread that actually handles the request).

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
pTalanov: Not sure about users calling `isEmpty`. I think it's more likely that users will ignore this property or search for annotations that their logic depends on (`annotations.filterIsInstance<UserAnnotation>()`).

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
pTalanov: That somewhat maps to current API in a following way:
```
data class ResolutionResult(
    val dependencies: ScriptDependencies?,
    val diagnostics: List<ScriptDiagnostic>
) {
    val isSuccess get() = dependencies != null
}
```

Not sure why we should avoid hierarchies. Can you explain your reasoning a bit?

udalov: > That somewhat maps to current API in a following way

This way is fine too but in (the most useful) case of success, you have to use `dependencies!!`, which is unfortunate.

> Not sure why we should avoid hierarchies. Can you explain your reasoning a bit?

Because the two subclasses here are not very different, and IMHO they add more complexity than value. A data class is easier to read and use than a sealed hierarchy of two nested sealed classes. Compare
```
val result = resolver.resolve(...)
if (result is DependenciesResolver.ResolutionResult.Success) {
    ...
}
```

```
val result = resolver.resolve(...)
if (result.isSuccess) {
    ...
}
```

pTalanov: My main concern with your proposal is that the user has to pass both `isSuccess` flag and notnull `dependencies` in both cases. While `ScriptDependencies.Empty` is a completely valid value for a successful resolution, it's not clear that it's meant to be ignored if `isSuccessful == false`. That could be explained in the docs of course.
I agree that using sealed class makes API a bit cumbursome. What I find problematic in the current API is that it's not clear that successful resolution can actually contain error diagnostics and that may be confusing for implementors of the API. 
I would not put a lot of emphasis on ease of consuming the API (since it's gonna be us who'll be doing that) but rather focus on the API being easy to implement.
In my mind those snippets would look like (with the current API):
```
when (result) {
    is Success -> ...
    is Failure -> ...
}
```
or 
```
val dependencies = result.dependencies ?: ScriptDependencies.Empty
val diagnostics = result.diagnostics
...
```

pTalanov: Talked to @udalov in person. We seem to agree that: 
```
data class ResolutionResult(
    val dependencies: ScriptDependencies?,
    val diagnostics: List<ScriptDiagnostic>
)
```
is simpler and covers all the required cases. It needs some documentation around what `dependencies == null` means but so do all other considered variations.

pTalanov: `val diagnostics: List<ScriptDiagnostics> = emptyList()` would be better?

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
```diff
@@ -21,6 +21,24 @@ package kotlin.script.dependencies
...
+// but rather as a dependency for a module that contains DependenciesResolver implementation so maybe this is a non-problem
+//
+// ResolveResult contains resulting dependencies and reports (diagnostics?)
+// reports may contain errors (regardless of result being Success or Failure) which is gonna lead to compiler error and script not run/error highlighting in IDE.
+// Is this semantic reasonable? Are Success and Failure misleading names?
+// Main idea behind Failure is to be able to distinguish between scenarios where resolver could or could not return meaningful ScriptDependencies object.
+// For example, IDE can avoid repainting all external references as errors when Resolver threw an exception or is in inconsistent state.
 typealias Environment = Map<String, Any?>
 
 interface DependenciesResolver : @Suppress("DEPRECATION") ScriptDependenciesResolver {
```
ilya-g: `DependencyResolver` instead of `DependenciesResolver`

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
ligee: Perhaps yes. The environment in which we'd compile script using e.g. the daemon running under java 8, but run it under java 6 is theoretically possible but practically seems too complicated to be useful.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
ligee: It is unused yet but meant for support of multifile scripts, some form of `use` or `include` functionality.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
ligee: Is the meaning of "classpath" in java already nailed somewhere? I guess we're using it in exactly the same sence.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: How is "ResolutionResult" better than "ResolveResult"? In my understanding - we have an action named "resolve" and the result of this action is "ResolveResult". With "ResolutionResult" one has to make an additional (albeit grammatic) mapping.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
udalov: I suggest to remove `scripts` and support it in the future when the requirements become clear

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: Why? There are potentially many of them.
We're using plural form everywhere now.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
udalov: Noun+noun is more correct grammatically and is wider used in our project (for example, `AnalysisResult`, `CompilationResult`, `DeadCodeEliminationResult`, `FileProcessingResult`, etc.)

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: I believe that we better avoid additional documentation here. Maybe something like this:
```
data class ResolutionResult(
    val dependencies: ScriptDependencies,
    val diagnostics: List<ScriptDiagnostic>
    val isSuccess get() = diagnostics.none { it.severity == ScriptDiagnostic.Severity.ERROR }
)
```

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: +1

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: I'm still confused by the fact that `Sequence` is not `Iterable`, but otherwise `Sequence` is a good variant for me.

About `annotations` - the fact that we're able to use only file annotations here is an unfortunate limitation of the current parser, that I believe could be lifted later. So I don't think that this limitation should be reflected in the API.

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
ligee: Should it go to the runtime/stdlib then?

core/script.runtime/src/kotlin/script/templates/annotations.kt
ligee: Fine with me, although this one doesn't seem critical to me.

core/script.runtime/src/kotlin/script/templates/annotations.kt
ligee: I thought about it as well, but it looked too flexible. But we can consider something like
```
val scriptMatcher: KClass<out ScriptMatcher> = StandardScriptFileMatcher::class
```
instead of `scriptFilePattern`

core/script.runtime/src/kotlin/script/templates/annotations.kt
ligee: The main reason is not to construct annotations unless the resolver needs them. And potentially more type safety.

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
udalov: I would prefer if this annotation was removed completely and [the same mechanism](https://kotlinlang.org/docs/reference/compiler-plugins.html#how-to-use-all-open-plugin) as in allopen and noarg plugins was used: a configuration parameter that specifies which annotations enable the SamWithReceiver behavior

core/script.runtime/src/kotlin/script/templates/standard/templates.kt
ligee: Since the (manual) construction of the script requires actual arguments array, the simple no-arg variant could be useful in some applications. So, id'leave it, may be renaming to `...NoParameters` (although I like `Simple` better).
The rest - I agree.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
ligee: Changing API here is very painful. Maybe that is something to address as well with this review.

The requirements are quite clear. People want to be able to write something like
```
@file:use("other-script.kts")
```
and use all definitions from the included scripts after that. It is not clear how we can implement it, but from the API point of view - it is clear.
That's why I'd leave it as is, and tried to implement on the next iteration.

core/script.runtime/src/kotlin/script/templates/annotations.kt
udalov: There aren't many annotations usually so the benefits of this laziness are not clear to me. I suggest to remove this and reintroduce in the future only if someone meets real performance problems.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
pTalanov: @ligee See the beginning of this thread.

core/script.runtime/src/kotlin/script/templates/standard/templates.kt
udalov: > the (manual) construction of the script requires actual arguments array
Why would that be a problem?

core/script.runtime/src/kotlin/script/templates/annotations.kt
pTalanov: I think we should think ahead if it makes sense to support this. Because if we provide only the simplest API right now and decide to add 'content-based' API later we'll end up with two APIs that do the same thing and have problems merging their results later. 
The snippet I came up with just illustrutates that the user can potentially want to decide to change compiler arguments based on ScriptContents with some custom logic.

API proposal: expand `ScriptDependencies` class with `val extraCompilerArguments: List<String> = emptyList()` so this can managed uniformly with other scipt content depent stuff.

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
ligee: The problem is that now the script template is the only configuration transfer channel for gradle kotlin DSL, therefore the annotation `SamWithReceiverAnnotations` is a natural way to convey the info. If we'll drop it, we'll need to introduce another channel, which is quite unfortunate.

pTalanov: We should consider if implementing https://youtrack.jetbrains.com/issue/KT-19120 is the way to go.

core/script.runtime/src/kotlin/script/templates/annotations.kt
ligee: Seems we're talking about two things here:
1. The laziness of the annotation creation
2. Additional filter in the form of `@AcceptedAnnotations`

I generally disagree with the "pessimization" approach, especially when the optimized one is already in place. Usually, nobody meets real performance problem in one clearly visible bottleneck, but the problem is usually distributed evenly. Skipping the job that nobody needs is usually quite straightforward and implementation here is quite simple. 
On the other hand, if the laziness will become problematic (as @pTalanov mentioned above), the second mechanism of filtering will eliminate the need of constructing unneeded annotations (and in most cases will eliminate the construction altogether, since there are not many annotation-processing resolvers yet written).
Another consideration is that the annotation creation code (https://github.com/JetBrains/kotlin/blob/master/compiler/frontend.script/src/org/jetbrains/kotlin/script/scriptAnnotationsPreprocessing.kt#L35) is not simple, and is relying on some frontend magic, and may need some more logic on to of what is already written, so I don't think that the performance is something we can easily discount here.
And finally, the filtering annotation could be used as an additional type check, ensuring that the resolver receiving the annotations it is able to process.

core/script.runtime/src/kotlin/script/templates/standard/templates.kt
ligee: The simple variant is easier to grasp by someone who is trying understand the samples or to use the code, no?

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
ligee: Yes, that would be a generic solution for such problems, I like it. But then we should implement it before dropping this annotation from here.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: ok, fine with me.

core/script.runtime/src/kotlin/script/extensions/samWithReceiver.kt
udalov: Yes, I think once KT-19120 is implemented, users can just pass something like `-Xplugin:kotlin-sam-with-receiver.jar -P plugin:sam-with-receiver:annotation= org.gradle.api.HasImplicitReceiver`

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
ligee: Because `CharSequence` is more generic and should be enough for the most of the actual use cases - searching or parsing. But I'm not really against making it a `String`, provided that we have a reason.

core/script.runtime/src/kotlin/script/templates/standard/templates.kt
udalov: Of course it is, but it isn't an argument as to whether we should keep it :) The string array is a trivial concept and we require everyone starting Kotlin programming to understand that the entrypoint is always `fun main(args: Array<String>)` anyway. We didn't introduce `fun main()` just because it's easier, there was no demand. So I hope that almost everyone will immediately understand why scripts need arguments, and if a script doesn't need arguments, it can just ignore them.

core/script.runtime/src/kotlin/script/dependencies/resolvers.kt
udalov: I believe it's easier for the client code to work with this text as a string, there's no real reason for the implementor to return any other char sequence except string, and `CharSequence`s are rarely found in good APIs anyway.

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
orangy: I think non-JVM scripts is what we should anticipate. Kotlin/Native specifically. So dependencies infra should be flexible enough, e.g. no javaHome & classpath, but instead may be native libs. Native scripting is also potentially used in game dev, like Lua.

ligee: This sounds nice and interesting, but how far we are from providing compilation facilities on non-JVM platforms?

core/script.runtime/src/kotlin/script/dependencies/dependencies.kt
pTalanov: Let's think about what we can do instead. Something like:
```
data class ScriptDependencies {
   val sdkBinaries: List<File> = emptyList(),
   val dependencyBinaries: List<File> = emptyList(),
   val dependencySources: List<File> = emptyList(),
   val scriptDependencies: List<File> = emptyList(),
  ...
  extraCompilerArguments, imports
}
```

Maybe just changing the naming so it would makes sense on other platforms is enough.

udalov: Expanding on the idea proposed in https://github.com/JetBrains/kotlin/pull/1210#discussion_r129553174, another option would be to drop this class altogether and just use the compiler arguments list instead of `ScriptDependencies`. So, on JVM I'd pass `-java-home` and `-classpath`, on JS -- `-libraries`, etc.

core/script.runtime/src/kotlin/script/templates/annotations.kt
pTalanov: Wanna stress that this isn't a trivial issue. For now scripts files must have ".kts" extension I believe. And we may or may not decide to lift this restriction later. But determining script type by it's content is something else entirely. What're the potential usecases?
Problem arises when we allow both non-standard extensions and content-based matches. Say we have some file "script.stk" and there is custom matcher that says that it is a script of specific type based on it's content so we must load it's content even before we know how to parse this file (and parsing decision should not involve any complex logic).

ligee:
The PR itself is obsolete, but the comments are useful and shall be reviewed during the next round of scripting API reviews. https://youtrack.jetbrains.com/issue/KT-28528

PR #1211, opened 2017-07-25
KT-4748

Dimach:
Inspection and tests.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/DoubleNegationInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/DoubleNegationInspection.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/DoubleNegationInspection.kt
```diff
@@ -0,0 +1,68 @@
...
+                        return
+                    }
+                    var parent = expression.parent
+                    while (parent is KtParenthesizedExpression) {
+                        parent = parent.parent
+                    }
+                    if (parent is KtPrefixExpression && parent.operationToken == KtTokens.EXCL) {
+                        holder.registerProblem(expression,
+                                               "Redundant double negation",
+                                               ProblemHighlightType.WEAK_WARNING,
```
mglukhikh: For future: use here `GENERIC_ERROR_OR_WARNING` if you want to get highlight type depending on inspection severity level (by default, from plugin.xml, or configured by user). Explicitly given `WEAK_WARNING` enforces this severiry level making it unconfigurable by user (will fix myself).

mglukhikh:
Looks good, I will test this inspection.

idea/src/org/jetbrains/kotlin/idea/inspections/DoubleNegationInspection.kt
Dimach: Ok, thanks for tip.

mglukhikh:
Also, inspection description was forgotten, added (hint: run InspectionDescriptionsTest). Merged manually into master, thank you.

PR #1212, opened 2017-07-26
KT-13985 "Add remaining branches" action does not use back-ticks correctly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13985


mglukhikh:
Merged after minor edit. Thank you

PR #1213, opened 2017-07-26
KT-12613 "Make abstract" on member of open or final class should make abstract both member and class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12613

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/AddModifierFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/AddModifierFix.kt
@@ -33,6 +33,7 @@ import org.jetbrains.kotlin.lexer.KtModifierKeywordToken
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.lexer.KtTokens.*
 import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.containingClass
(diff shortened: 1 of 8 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/AddModifierFix.kt
```diff
@@ -56,6 +57,13 @@ open class AddModifierFix(
 
     override fun invoke(project: Project, editor: Editor?, file: KtFile) {
         element?.addModifier(modifier)
+
+        if (modifier == KtTokens.ABSTRACT_KEYWORD && (element is KtProperty || element is KtNamedFunction)) {
+            element?.containingClass()?.run {
+                if (!hasModifier(KtTokens.ABSTRACT_KEYWORD) && !hasModifier(KtTokens.SEALED_KEYWORD)) addModifier(KtTokens.ABSTRACT_KEYWORD)
+                if (hasModifier(KtTokens.OPEN_KEYWORD)) removeModifier(KtTokens.OPEN_KEYWORD)
```
mglukhikh: Should work without this line, at least in theory. Will check it myself.

mglukhikh:
Merged after minor fix

idea/src/org/jetbrains/kotlin/idea/quickfix/AddModifierFix.kt
mglukhikh: Really works (add abstract also removes other modality modifiers, if any)

PR #1215, opened 2017-07-26
KT-19232 Replace Math.min with coerceAtMost intention is broken

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19232

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMethodsWithKotlinNativeMethodsIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMethodsWithKotlinNativeMethodsIntention.kt
@@ -18,18 +18,17 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import org.jetbrains.kotlin.idea.core.replaced
-import org.jetbrains.kotlin.psi.KtCallExpression
-import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
-import org.jetbrains.kotlin.psi.KtPsiFactory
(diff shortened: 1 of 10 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMathMethodsWithKotlinNativeMethodsIntention.kt
```diff
@@ -18,18 +18,17 @@ package org.jetbrains.kotlin.idea.intentions
...
 abstract class ReplaceMathMethodsWithKotlinNativeMethodsIntention(
-        text: String, val replacedMethodName: String, val mathMethodName: String
+        text: String, private val replacedMethodName: String, private val mathMethodName: String
 ) : SelfTargetingOffsetIndependentIntention<KtCallExpression>(KtCallExpression::class.java, text) {
 
     override fun applyTo(element: KtCallExpression, editor: Editor?) {
-        val target = element.getStrictParentOfType<KtDotQualifiedExpression>() ?: element
+        val target = element.getStrictParentOfType<KtDotQualifiedExpression>()?.takeIf {
+            (it.receiverExpression as? KtNameReferenceExpression)?.getReferencedName() == "Math"
+        } ?: element
```
mglukhikh: It's also not the best way, you can meet something like

```
Math.max(max(1, 3), max(2, 4))
```

and will have the same problem. Will fix it myself.

mglukhikh: Suppose `getQualifiedExpressionForSelectorOrThis()` should be better here

mglukhikh:
Fixed & merged manually, thank you.

PR #1218, opened 2017-07-28
KT-18822 Do not report "can be private" on members of private / local classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18822


mglukhikh:
Thank you, merged (added also EffectiveVisibility.LOCAL).

PR #1219, opened 2017-07-29
KT-18832: Fix incompatibility of kotlin-gradle-plugin with Java 9

marcphilipp:
Use latest 2.x versions of Commons IO/Lang to fix incompatibility of with versioning scheme of Java 9.


marcphilipp:
Using 2.5 does not help after all.

PR #1221, opened 2017-07-30
KT-17379: Converting multiline expressions creates dangling operations fixed

Dimach:


compiler/frontend/src/org/jetbrains/kotlin/psi/KtPsiUtil.java
```diff
@@ -513,6 +513,16 @@ public static boolean areParenthesesNecessary(
             return false;
         }
 
+        if (innerExpression instanceof KtBinaryExpression) {
+            PsiElement expr = innerExpression.getFirstChild();
+            while (expr != null) {
+                if (expr instanceof PsiWhiteSpace && (expr.textContains('\n') || expr.textContains('\r'))) {
```
yole: IntelliJ always normalizes text of files to use \n line breaks, the \r will never be there. Please remove it from the check.

j2k/testData/fileOrElement/kotlinApiAccess/ObjectMembers.kt
```diff
@@ -4,8 +4,8 @@ internal class C {
     fun foo(): Int {
         KotlinObject.property1 = 1
         KotlinObject.property2 = 2
-        return KotlinObject.foo() +
+        return (KotlinObject.foo() +
```
yole: The parentheses are only required if the line break is before the operator. In this case they should not be added.

Dimach:
Done.

j2k/testData/fileOrElement/kotlinApiAccess/CorrectNullabilityDetected.kt
```diff
@@ -3,13 +3,13 @@ import kotlinApi.*
 //TODO: Formatter works incorrectly
```
piotrek1543: After changes this `TODO` should be removed.

j2k/testData/fileOrElement/kotlinApiAccess/CorrectNullabilityDetected.kt
```diff
@@ -3,13 +3,13 @@ import kotlinApi.*
...
     fun foo(c: KotlinClass): Int {
-        return c.nullableProperty!!.length
-        +c.property.length
-        +KotlinClass.nullableStaticVar!!
-        +KotlinClass.staticVar
-        +KotlinClass.nullableStaticFun(1)!!
-        +KotlinClass.staticFun(1)
-        +nullableGlobalFunction("")!!.length
-        +globalFunction("").length
+        return (c.nullableProperty!!.length
```
piotrek1543: After changes `TODO` comment should be removed

j2k/testData/fileOrElement/strings/formattedString.kt
```diff
@@ -4,7 +4,7 @@ object A {
             "text3"
 
 
-    val TEXT2 = "text1\n"
-    + "text2\n"
-    + "text3"
+    val TEXT2 = ("text1\n"
```
piotrek1543: If I were you, I would reduce space from 2 to 1 line. I think one is enough to make this object readable.

j2k/testData/fileOrElement/kotlinApiAccess/CorrectNullabilityDetected.kt
piotrek1543: After changes TODO comment above your code should be removed. You've already fixed it.

j2k/testData/fileOrElement/strings/formattedString.kt
Dimach: You write comment to proper file? I dont see any problem here. Maybe you mean kt-17379.kt?

Dimach:
I improved new lines recognition (diff you can see in orWithNewLine.kt) but i not fix problem with "end of line comments" that produce spare newlines in this PR, its different task as i think.

semoro:
Merged manually. 
Thanks! 🙂

PR #1222, opened 2017-07-31
KT-19299 Quickfix to correct overriding function signature keeps java NotNull annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19299


mglukhikh:
Merged. However, I removed yet `JAVAX_NONNULL_ANNOTATION` due to its unclear status. May be it should be restored a bit later.

PR #1223, opened 2017-07-31
KT-19273 Do not report "property can be private" in annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19273


mglukhikh:
Thank you, merged

PR #1224, opened 2017-07-31
KT-19126 Add replace property initializer with getter in interfaces

neworld:
https://youtrack.jetbrains.com/issue/KT-19126

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/KtPsiFactory.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/KtPsiFactory.kt
@@ -261,6 +261,15 @@ class KtPsiFactory @JvmOverloads constructor(private val project: Project, val m
         return getter
     }
 
+    fun createPropertySetter(expression: KtExpression): KtPropertyAccessor {
+        val property = createProperty("val x get() = 1\nset(value) = TODO()")
+        val setter = property.setter!!
+        val bodyExpression = setter.bodyExpression!!
+
+        bodyExpression.replace(expression)
+        return setter
+    }
+
     fun createDestructuringDeclaration(text: String): KtDestructuringDeclaration {
         return (createFunction("fun foo() {$text}").bodyExpression as KtBlockExpression).statements.first() as KtDestructuringDeclaration
     }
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceInitializerWithGetterFix.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.psi.PsiNameIdentifierOwner
+import org.jetbrains.kotlin.diagnostics.Diagnostic
+import org.jetbrains.kotlin.idea.core.deleteElementAndCleanParent
+import org.jetbrains.kotlin.psi.KtDeclaration
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+
+/**
```
yole: Please don't add `@author` tags to the code

idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceInitializerWithGetterFix.kt
```diff
@@ -0,0 +1,58 @@
...
+ * @author Andrius Semionovas
+ * @since 2017-07-31
+ */
+class ReplaceInitializerWithGetterFix(element: KtProperty): KotlinQuickFixAction<KtProperty>(element) {
```
yole: We already have `ConvertPropertyInitializerToGetterIntention` which is offered as an intention action in this context; you should be able to reuse it directly.

idea/testData/quickfix/replaceInitializerWithGetter/simple.kt
```diff
@@ -0,0 +1,7 @@
+// "Replace initializer with getter" "true"
+
+fun String.foo() = "bar"
+
+interface A {
+    val name = <caret>"The quick brown fox jumps over the lazy dog".foo()
```
yole: Neither the existing intention action nor your suggested quickfix handle `var` properties correctly. They should create a stub for the setter with no implementation.

neworld:
Updated

yole:

PR #1226, opened 2017-08-01
KT-18482: "Move lambda argument to parenthesis" action generate uncompilable code fixed

Dimach:


idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
```diff
@@ -70,7 +72,11 @@ fun KtLambdaArgument.moveInsideParenthesesAndReplaceWith(
     val newCallExpression = oldCallExpression.copy() as KtCallExpression
 
     val psiFactory = KtPsiFactory(project)
-    val argument = if (newCallExpression.getValueArgumentsInParentheses().any { it.isNamed() }) {
+
+    val argument = if (newCallExpression.getValueArgumentsInParentheses().run {
```
yole: I think this nested `run` expression is far too complex. Better to extract this logic into a method.

idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
```diff
@@ -354,11 +360,11 @@ fun KtTypeParameterListOwner.addTypeParameter(typeParameter: KtTypeParameter): K
     val list = KtPsiFactory(this).createTypeParameterList("<X>")
     list.parameters[0].replace(typeParameter)
     val leftAnchor = when (this) {
-        is KtClass -> nameIdentifier ?: getClassOrInterfaceKeyword()
-        is KtNamedFunction -> funKeyword
-        is KtProperty -> valOrVarKeyword
-        else -> null
-    } ?: return null
+                         is KtClass -> nameIdentifier ?: getClassOrInterfaceKeyword()
```
piotrek1543: There is a lot of unnecessary code formatting, which make harder to find what exactly author of PR changed. @Dimach, did was intentional and done by you or automatically changed by IntelliJ?

idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
piotrek1543: I agree @yole. This part is really good example what should be extracted to method.

idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
Dimach: I use reformat code action (alt+shift+l) so idea format my changes and all file. I can remove all format changes from PR if you want.

yole: It's OK to keep the formatting changes.

piotrek1543: @Dimach I just wondered the reason of this changes. Now, it's ok.;-)

Dimach:
Done.

idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
```diff
@@ -92,6 +95,14 @@ fun KtLambdaArgument.moveInsideParenthesesAndReplaceWith(
     return oldCallExpression.replace(newCallExpression) as KtCallExpression
 }
 
+private fun shouldLambdaParameterNamed(args: List<ValueArgument>, callExpr: KtCallExpression): Boolean {
+    return args.any { it.isNamed() } ||
```
yole: Now that it's a separate function, there's no need to cramp it all into a single expression.

```
if (args.any { it.isNamed () }) return true
val callee = ...resolve as? KtFunction ?: return true
return callee.valueParameters.any { ... }
```

Also the grammatically correct name would be `shouldLambdaParameterBeNamed`.

yole:

PR #1227, opened 2017-08-03
KT-18083: Detect valid extension main functions

cypressious:
Fixes #KT-18083


cypressious:
@piotrek1543 I've already commented with the link to the PR on the youtrack issue, so the PR shouldn't go unnoticed, but I've also changed the name.

yole:

PR #1228, opened 2017-08-04
KT-18232: Kotlin code converter misses annotations fixed

Dimach:


semoro:
Fixed more correctly in #1238. Thank you anyway! 🙂

PR #1230, opened 2017-08-05
Add assertSame and assertNotSame methods.

JakeWharton:


JakeWharton:
@ilya-g

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -10,6 +10,12 @@ class BasicAssertionsTest {
     }
 
     @Test
+    fun testAssertSame() {
```
piotrek1543: `test` prefix reminds me old days of `JUnit`, where it was obligatory to run any test :-)

Could you tell me why there is doubled `kotlin` and `test` packages in path: `kotlin/libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/`? Just wondering.

ilya-g: `kotlin.test/common` is the project directory, then `src/test/kotlin` is the source root for tests, then `kotlin/test/tests` is the package name for tests of `kotlin.test` :)

ilya-g:
I've merged it with minor edits: https://github.com/JetBrains/kotlin/compare/485ca10~2...485ca10

JakeWharton:
Thanks!

PR #1232, opened 2017-08-05
KT-9669: Join Lines should add semicolon when joining statements into the same line

voddan:
#KT-1234 Fixed: introduced `JoinStatementAddSemicolonHandler`; registered the handler in `plugin.xml`; added test coverage

### DIFF
--- a/idea/testData/joinLines/addSemicolon/WhileAndACall.kt
+++ b/idea/testData/joinLines/addSemicolon/WhileAndACall.kt
@@ -1,0 +1,4 @@
+fun foo() {
+    <caret>while (true) println()
+    println()
+}
--- a/idea/testData/joinLines/addSemicolon/WhileAndACall.kt.after
+++ b/idea/testData/joinLines/addSemicolon/WhileAndACall.kt.after
@@ -1,0 +1,3 @@
+fun foo() {
+    while (true) println();<caret> println()
+}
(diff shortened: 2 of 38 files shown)


idea/testData/joinLines/addSemicolon/WhileAndACall.kt.after
```diff
@@ -0,0 +1,3 @@
+fun foo() {
+    while (true) println();<caret> println()
```
piotrek1543: why did you put there semicolon?

idea/testData/joinLines/addSemicolon/WhileAndACall.kt
```diff
@@ -0,0 +1,4 @@
+fun foo() {
+    <caret>while (true) println()
```
piotrek1543: how do you want to interrupt this infinite loop statement?

voddan: Not sure what you mean. Why should I want to interrupt it?

yole:

voddan:
Funny thing is I had to learn half of the Kotlin compiler before I found the simpler solution. Took me long enough 😄

voddan:
BTW, the build process for the project is just horrible. 

I spent a day and a half in the beginning before I could compile tests without issues. Even worse, IDE stoped compiling after I rebased to the current master. Deleting the folder and downloading everything from scratch did not help. 

Please do something about it. Migrating everything to a proper Maven project could definitely help.

yole:
We're in the process of migrating the build of Kotlin to gradle-kotlin-dsl.

voddan:
Glad to hear that! 

Please also migrate the downloading of the dependancies, as well as the build targets 🙏  Updating dependancies was a nightmare over unstable internet, I had to hack the script and download archives by hand. Build targets, dependent on the IJ configuration, is a very unstable solution. The ide is not really up to reconstructing builds.

PR #1233, opened 2017-08-07
KT-19282 Support "flip equals" intention for String.equals extension from stdlib

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19282

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/SwapStringEqualsIgnoreCaseIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/SwapStringEqualsIgnoreCaseIntention.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/SwapStringEqualsIgnoreCaseIntention.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    override fun applyTo(element: KtDotQualifiedExpression, editor: Editor?) {
+        val callExpression = element.callExpression ?: return
+        val offset = (editor?.caretModel?.offset ?: 0) - (callExpression.calleeExpression?.startOffset ?: 0)
+        val receiverExpression = element.receiverExpression
+        val valueArguments = callExpression.valueArguments.takeIf { it.size == 2 } ?: return
+        val newElement = KtPsiFactory(element).createExpressionByPattern(
+                "$0.equals($1, $2)",
+                valueArguments[0].text,
```
mglukhikh: It's better to use here expressions and not their text

t-kameyama: Fixed

mglukhikh:
Merged manually

PR #1234, opened 2017-08-10
KT-19393 long == <int_literal> in Java converted to red code in Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19393


semoro:
Thanks! 🙂

PR #1235, opened 2017-08-10
KT-19390 Character and string concatenation in Java is converted to code with multiple type errors in Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19390


j2k/src/org/jetbrains/kotlin/j2k/CodeConverter.kt
```diff
@@ -115,7 +115,21 @@ class CodeConverter(
         if (needConversion(actualType, expectedType)) {
             val expectedTypeStr = expectedType.canonicalText
             if (expression is PsiLiteralExpression) {
-                if (expectedTypeStr == "float" || expectedTypeStr == "double") {
+                if (actualType.canonicalText == "char" && expression.parent !is PsiExpressionList) {
+                    when (expectedTypeStr) {
```
semoro: I think that's here should be a better way. 
Better to move such code may be to type converter or so. 
Because it looks like something similar required here #1200

semoro:
Sorry for delay

j2k/src/org/jetbrains/kotlin/j2k/CodeConverter.kt
t-kameyama: Fixed

semoro:
Thanks! 🙂

PR #1236, opened 2017-08-10
KT-10375, KT-19523 J2K number conversion issues fixed.

Dimach:


j2k/testData/fileOrElement/switch/KT-13552.kt
```diff
@@ -73,10 +73,10 @@ object SwitchDemo {
         val status = ""
         when (status) {
             "init", "dial", "transmit" -> return 0x111111
-            "ok" -> return 0xFF006600.toInt()
-            "cancel" -> return 0xFF666666.toInt()
-            "fail", "busy", "error" -> return 0xFF660000.toInt()
-            else -> return 0xFF660000.toInt()
+            "ok" -> return -16751104
```
semoro: Please, write out these numbers in hex, if input in hex.

j2k/testData/fileOrElement/literalExpression/hex.kt
```diff
@@ -2,12 +2,13 @@ internal class Test {
...
-        val i3 = -1
+        val i3 = -0x1
         val i4 = 0XCCCC
         val i5 = 0x6fffffff
-        val l1: Long = -1
+        val l1: Long = -0x1
         val l2 = 0xffffffffL
         val l3 = 0xffffffffffL
-        val l4 = -1
+        val l4 = -0x1
```
semoro: `L` is missing after conversion, so result type of `l4` would be `Int` instead of `Long`

semoro:
Thank you! 🙂

PR #1237, opened 2017-08-11
KT-19614 Quickfix for INVISIBLE_MEMBER doesn't offer to make member protected if referenced from subclass

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19614

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/MakeVisibleFactory.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/MakeVisibleFactory.kt
@@ -44,7 +44,7 @@ object MakeVisibleFactory  : KotlinIntentionActionsFactory() {
 
         val module = DescriptorUtils.getContainingModule(descriptor)
         val targetVisibilities = when (descriptor.visibility) {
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/MakeVisibleFactory.kt
```diff
@@ -44,7 +44,7 @@ object MakeVisibleFactory  : KotlinIntentionActionsFactory() {
 
         val module = DescriptorUtils.getContainingModule(descriptor)
         val targetVisibilities = when (descriptor.visibility) {
-            PRIVATE, INVISIBLE_FAKE -> if (module != usageModule) listOf(PUBLIC) else listOf(PUBLIC, INTERNAL)
+            PRIVATE, INVISIBLE_FAKE -> if (module != usageModule) listOf(PUBLIC) else listOf(PUBLIC, INTERNAL, PROTECTED)
```
mglukhikh: It's not so easy, because `protected` should not be offered if member is referenced not from a subclass. Please add necessary checks here and such a test.

t-kameyama: Fixed

mglukhikh:
Merged manually, thank you

PR #1238, opened 2017-08-11
KT-18232 Kotlin code converter misses annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18232


semoro:
Merged manually. Thanks! 🙂

PR #1241, opened 2017-08-11
Promote @Test and @Ignore into kotlin.test.* in common module.

JakeWharton:
#KT-19494 Fixed

### DIFF
--- a/libraries/kotlin.test/common/src/main/kotlin/org/junit/Ignore.kt
+++ b/libraries/kotlin.test/common/src/main/kotlin/org/junit/Ignore.kt
@@ -1,0 +1,5 @@
+package org.junit
+
+@Deprecated("Use 'Ignore' from kotlin.test package", replaceWith = ReplaceWith("Ignore", imports = "kotlin.test.Ignore"))
+@Suppress("HEADER_WITHOUT_IMPLEMENTATION")
+header annotation class Ignore
--- a/libraries/kotlin.test/common/src/main/kotlin/org/junit/Test.kt
+++ b/libraries/kotlin.test/common/src/main/kotlin/org/junit/Test.kt
@@ -1,4 +1,5 @@
 package org.junit
 
+@Deprecated("Use 'Test' from kotlin.test package", replaceWith = ReplaceWith("Test", imports = "kotlin.test.Test"))
 @Suppress("HEADER_WITHOUT_IMPLEMENTATION")
 header annotation class Test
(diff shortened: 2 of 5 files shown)


libraries/kotlin.test/common/src/main/kotlin/org/junit/Ignore.kt
```diff
@@ -0,0 +1,5 @@
+package org.junit
+
+@Deprecated("Use 'Ignore' from kotlin.test package", replaceWith = ReplaceWith("Ignore", imports = "kotlin.test.Ignore"))
+@Suppress("HEADER_WITHOUT_IMPLEMENTATION")
+header annotation class Ignore
```
JakeWharton: I'm working on not needing this...

JakeWharton: I think this is blocked by https://youtrack.jetbrains.com/issue/KT-19656.

I can't create `@Ignore` and alias to `org.junit.Ignore` without mirroring its full set of properties, despite the fact that on the JVM they all have defaults.

JakeWharton: And I don't want to "simply" mirror the properties because honoring them in the JS runners is a non-trivial task that I don't want to scope as part of this PR.

libraries/kotlin.test/common/src/main/kotlin/org/junit/Test.kt
```diff
@@ -1,4 +1,5 @@
 package org.junit
 
+@Deprecated("Use 'Test' from kotlin.test package", replaceWith = ReplaceWith("Test", imports = "kotlin.test.Test"))
```
jaredsburrows: 👍

jaredsburrows:
LGTM.

JakeWharton:
This PR conflicts with the `@Ignore` that was added directly to the `js/` module instead of being put in the `common/` one. `@Before` and `@After` were also added to `js/` but should have been added to `common/` as well.

I can update this PR to include `@Before` and `@After`, although it will suffer the same problems as commented above.

ilya-g:
Please see this comment about the plan for multiplatform test annotations: https://youtrack.jetbrains.com/issue/KT-19696#comment=27-2419074

ilya-g:
Thanks for PR anyway, I think I could partially use it later.

JakeWharton:
Yep! Happy to see forward progress no matter how it gets done.

PR #1242, opened 2017-08-14
KT-19428 Add inspection for redundant overrides that only call the super method

takahirom:
https://youtrack.jetbrains.com/issue/KT-19428

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.*
+import com.intellij.openapi.project.Project
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+
+class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession) =
+            object : KtVisitorVoid() {
+                override fun visitNamedFunction(function: KtNamedFunction) {
+                    super.visitNamedFunction(function)
+                    if (!function.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return
+
+                    val bodyExpression = function.bodyExpression
+                    bodyExpression ?: return
+
+                    val body = if (bodyExpression is KtDotQualifiedExpression) {
+                        bodyExpression
+                    }
+                    else {
+                        if (bodyExpression.children.size != 1) {
+                            return
+                        }
+                        bodyExpression.children[0]
+                    }
+
+                    val qualifiedExpression = if (body is KtReturnExpression) {
+                        val returnedExpression = body.returnedExpression
+                        returnedExpression ?: return
+                        returnedExpression as? KtDotQualifiedExpression ?: return
+                    }
+                    else {
+                        body as? KtDotQualifiedExpression ?: return
+                    }
+
+                    if (qualifiedExpression.receiverExpression !is KtSuperExpression) return
+
+                    val superSelectorExpression = qualifiedExpression.selectorExpression
+                    val superCallElement = superSelectorExpression as? KtCallElement ?: return
+
+                    if (!isSameFunctionName(superSelectorExpression, function)) return
+
+                    if (!isSameArguments(superCallElement, function)) return
+
+                    holder.registerProblem(function,
+                                           "Redundant override",
+                                           ProblemHighlightType.LIKE_UNUSED_SYMBOL,
+                                           RedundantOverrideFix())
+
+
+                }
+            }
+
+    private fun isSameArguments(superCallElement: KtCallElement, function: KtNamedFunction): Boolean {
+        val argumentList = superCallElement.valueArgumentList ?: return false
+        val parameterList = function.valueParameterList ?: return false
+        val diffArguments = argumentList.arguments.filter { argument ->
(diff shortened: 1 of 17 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+                    bodyExpression ?: return
+
+                    val body = if (bodyExpression is KtDotQualifiedExpression) {
+                        bodyExpression
+                    }
+                    else {
+                        if (bodyExpression.children.size != 1) {
+                            return
+                        }
+                        bodyExpression.children[0]
```
cypressious: You could use `singleOrNull() ?: return` here.

takahirom: I don't know it! Thanks :)

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,99 @@
...
+                    else {
+                        body as? KtDotQualifiedExpression ?: return
+                    }
+
+                    if (qualifiedExpression.receiverExpression !is KtSuperExpression) return
+
+                    val superSelectorExpression = qualifiedExpression.selectorExpression
+                    val superCallElement = superSelectorExpression as? KtCallElement ?: return
+
+                    if (!isSameFunctionName(superSelectorExpression, function)) return
```
takahirom: I can use `superCallElement` for the argument. I will fix it.

idea/testData/inspectionsLocal/redundantOverride/arguments.kt
```diff
@@ -0,0 +1,10 @@
+open class Foo {
+    open fun arguments(arg1: Int, arg2: Long) {
+    }
+}
+
+class Bar : Foo() {
+    override <caret>fun arguments(arg1: Int, arg2: Long) {
```
mglukhikh: I should mention that overridden function is not redundant if it has some other modifiers. For example, sometimes is reasonable to add `final` here, so function itself is not changed, but derived classes are made unable to override it. Also, I think that any annotation here makes the function not redundant. The same for any visibility modifier when you make overridden function more visible then original one. Please consider these cases, fix inspection, and add tests for them.

mglukhikh:
Also, please add test when `super` function is called in expression body format:

```
override fun foo() = super.foo()
```

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -69,6 +68,15 @@ class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLoc
                 }
             }
 
+    private fun KtNamedFunction.containsOtherModifier(): Boolean {
+        MODIFIER_EXCLUDE_OVERRIDE.forEach {
```
cypressious: You can simplify this to `return MODIFIER_EXCLUDE_OVERRIDE.any { hasModifier(it) }`

takahirom:
@mglukhikh 
Done!

```
Also, please add test when super function is called in expression body format:

override fun foo() = super.foo()
```

Probably test of expression body format is already contained.
https://github.com/JetBrains/kotlin/pull/1242/files#diff-9a733327a84bcf25fd5047bae854a384R7

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
takahirom: Thanks! I fixed it.

takahirom:
I fixed conflicts.

mglukhikh:
I think mostly it's Ok, I will recheck it, slightly correct style, and merge.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,102 @@
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession) =
+            object : KtVisitorVoid() {
+                override fun visitNamedFunction(function: KtNamedFunction) {
+                    super.visitNamedFunction(function)
+                    if (!function.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return
+                    if (function.containsOtherModifier()) return
+                    if (function.annotationEntries.size != 0) return
+
+                    val bodyExpression = function.bodyExpression
+                    bodyExpression ?: return
```
mglukhikh: This and previous can be written in one line

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+
+class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession) =
+            object : KtVisitorVoid() {
+                override fun visitNamedFunction(function: KtNamedFunction) {
+                    super.visitNamedFunction(function)
+                    if (!function.hasModifier(KtTokens.OVERRIDE_KEYWORD)) return
+                    if (function.containsOtherModifier()) return
```
mglukhikh: I'd inline this function

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,102 @@
...
+                    if (function.annotationEntries.size != 0) return
```
mglukhikh: isNotEmpty() ?

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -0,0 +1,102 @@
...
+                                           RedundantOverrideFix())
+                }
+            }
+
+    private fun KtNamedFunction.containsOtherModifier(): Boolean = MODIFIER_EXCLUDE_OVERRIDE.any { hasModifier(it) }
+
+    private fun isSameArguments(superCallElement: KtCallElement, function: KtNamedFunction): Boolean {
+        val argumentList = superCallElement.valueArgumentList ?: return false
+        val parameterList = function.valueParameterList ?: return false
+        val diffArguments = argumentList.arguments.filter { argument ->
```
mglukhikh: This can be rewritten shorter using e.g. `none`

mglukhikh: Also, seems to me this logic works incorrectly if arguments are replaced.

mglukhikh:
Merged manually

takahirom:
Thank you for comments. I can learn it.

PR #1244, opened 2017-08-15
Effect system + CFG (technical PR)

mglukhikh:


mglukhikh:
Please do not merge

compiler/resolution/src/org/jetbrains/kotlin/effectsystem/functors/FunctorsUtils.kt
```diff
@@ -0,0 +1,68 @@
...
+internal fun foldConditionsWithOr(list: List<ESClause>): ESBooleanExpression? =
+        if (list.isEmpty())
+            null
+        else
+            list.map { it.condition }.reduce { acc, condition -> acc.or(condition) }
+
+/**
+ * Places all clauses that equal to `firstModel` into first list, and all clauses that equal to `secondModel` into second list
+ */
+internal fun List<ESClause>.strictPartition(firstModel: ESEffect, secondModel: ESEffect): Pair<List<ESClause>, List<ESClause>> {
```
mglukhikh: May be it's better to return here just lists of conditions, not lists of clauses. Then you can accept list of conditions in `foldConditionsWithOr`.

compiler/resolution/src/org/jetbrains/kotlin/effectsystem/adapters/InfoCollector.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    override fun visitLambda(esLambda: ESLambda): MutableContextInfo {
+        throw IllegalStateException("InfoCollector shouldn't visit non-boolean values")
+    }
+
+    private fun <R> inverted(block: () -> R): R {
+        isInverted = isInverted.not()
+        val result = block()
+        isInverted = isInverted.not()
```
mglukhikh: Ugly construction. Are you sure it's really needed?

compiler/frontend/src/org/jetbrains/kotlin/effectsystem/adapters/ValueIDs.kt
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.jetbrains.kotlin.effectsystem.adapters
+
+import org.jetbrains.kotlin.effectsystem.structure.ConstantID
+import org.jetbrains.kotlin.effectsystem.structure.ESValueID
+import org.jetbrains.kotlin.resolve.calls.smartcasts.DataFlowValue
+
+class DataFlowValueID(val dfv: DataFlowValue) : ESValueID {
```
mglukhikh: Not quite sure this wrapper is needed. Can `DataFlowValue` itself serve you here?

compiler/testData/diagnostics/tests/effects/controlflow/flowInlining/unreachableCode.kt
```diff
@@ -0,0 +1,11 @@
+// !DIAGNOSTICS: -INVISIBLE_MEMBER -INVISIBLE_REFERENCE -UNUSED_PARAMETER
+// !LANGUAGE: +CalledInPlaceEffect
+
+import kotlin.internal.*
+
+inline fun myRun(@CalledInPlace(InvocationCount.EXACTLY_ONCE) block: () -> Unit) = block()
+
+fun throwInLambda(): Int {
+    val x = myRun { throw java.lang.IllegalArgumentException() }
+    return <!TYPE_MISMATCH!>x<!>
```
dsavvinov: Sorry, that's erroneous overwrite of previous commit, i.e. this test data is wrong and test fails. `UNREACHABLE_CODE` is indeed reported here, and this diff shouldn't be in commit with introduction of language features.
As for type mismatch, we currently have the same behavior for ordinary lambdas.

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/CallCompleter.kt
```diff
@@ -398,4 +401,13 @@ class CallCompleter(
...
         return expressionType != null && TypeUtils.isNullableType(expressionType)
     }
+
+    private fun MutableResolvedCall<*>.updateResultDataFlowInfoUsingEffects(bindingTrace: BindingTrace) {
+        val moduleDescriptor = DescriptorUtils.getContainingModule(this.resultingDescriptor?.containingDeclaration ?: return)
+
+        val resultDFIfromES = effectSystem.getResultDataFlowInfo(this, bindingTrace, moduleDescriptor)
+        this.dataFlowInfoForArguments.updateResultInfo(resultDFIfromES)
+
+        effectSystem.recordDefiniteInvocations(this, bindingTrace, moduleDescriptor)
```
dsavvinov: Introduction of separate language features required some refactoring of EffectSystem endpoints to make them more granular (previously, lambdas invocations were written during inference of resulting data flow info), thus requiring users of Effect System now ask for recording of definite invocations explicitly. Probably, commit message needs a bit of rewording.

compiler/frontend/src/org/jetbrains/kotlin/effectsystem/adapters/ValueIDs.kt
dsavvinov: The intent is to abstract effect system away from PSI, so that we could transparently change one wrapper into another as soon as we will have IR.

compiler/resolution/src/org/jetbrains/kotlin/effectsystem/adapters/InfoCollector.kt
dsavvinov: Potentially, this can be fixed by using visitor with data and passing around boolean flag `isInverted` as this data.

compiler/frontend/src/org/jetbrains/kotlin/effectsystem/adapters/CallTreeBuilder.kt
```diff
@@ -0,0 +1,194 @@
...
+    override fun visitLabeledExpression(expression: KtLabeledExpression, data: Unit): CTNode = expression.baseExpression?.accept(this, data) ?: UNKNOWN_CALL
+
+    override fun visitBinaryExpression(expression: KtBinaryExpression, data: Unit): CTCall {
+        tryGetCachedCall(expression)?.let { return it }
+
+        val leftNode = expression.left?.accept(this, data) ?: return UNKNOWN_CALL
+        val rightNode = expression.right?.accept(this, data) ?: return UNKNOWN_CALL
+
+        val functor = when (expression.operationToken) {
+            KtTokens.EQEQ, KtTokens.EQEQEQ -> {
```
dsavvinov: I think, everything is working correctly now, but only due to the very limited amount of cases where equality actually supported (only `Boolean` to `true`/`false` and comparisons with `null`). Still, it can potentially backfire if we will decide to cover more cases, so I think you're right, we should treat identity comparison more carefully (with separate functor, probably)

compiler/frontend/src/org/jetbrains/kotlin/types/expressions/ControlStructureTypingVisitor.java
```diff
@@ -105,8 +105,19 @@ public KotlinTypeInfo visitIfExpression(KtIfExpression ifExpression, ExpressionT
 
         LexicalWritableScope thenScope = newWritableScopeImpl(context, LexicalScopeKind.THEN, components.overloadChecker);
         LexicalWritableScope elseScope = newWritableScopeImpl(context, LexicalScopeKind.ELSE, components.overloadChecker);
-        DataFlowInfo thenInfo = components.dataFlowAnalyzer.extractDataFlowInfoFromCondition(condition, true, context).and(conditionDataFlowInfo);
-        DataFlowInfo elseInfo = components.dataFlowAnalyzer.extractDataFlowInfoFromCondition(condition, false, context).and(conditionDataFlowInfo);
+        DataFlowInfo basicThenInfo = components.dataFlowAnalyzer.extractDataFlowInfoFromCondition(condition, true, context).and(conditionDataFlowInfo);
+        DataFlowInfo thenInfoFromES = components.effectSystem.getConditionalInfoForThenBranch(
```
dsavvinov: Language features checks are executed in Effect System endpoints. Is it better to lift them to call-sites?

libraries/stdlib/src/kotlin/util/Preconditions.kt
```diff
@@ -82,7 +89,8 @@ public inline fun <T:Any> checkNotNull(value: T?): T = checkNotNull(value) { "Re
  * @sample samples.misc.Preconditions.failCheckWithLazyMessage
  */
 @kotlin.internal.InlineOnly
-public inline fun <T:Any> checkNotNull(value: T?, lazyMessage: () -> Any): T {
+@kotlin.internal.Returns
+public inline fun <T:Any> checkNotNull(@kotlin.internal.Equals(kotlin.internal.ConstantValue.NOT_NULL) value: T?, lazyMessage: () -> Any): T {
```
dsavvinov: In the current implementation, we don't have a way to express that function always throws (mainly because we were focused on practical applications)

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/model/MutableDataFlowInfoForArguments.java
```diff
@@ -47,6 +48,9 @@ public void updateInfo(@NotNull ValueArgument valueArgument, @NotNull DataFlowIn
             throw new IllegalStateException();
         }
 
+        @Override
+        public void updateResultInfo(@NotNull DataFlowInfo dataFlowInfo) { }
```
dsavvinov: This is actually reachable during resolution of calls with callable references. Though, it currently reachable only with `EMPTY` info, so it can be fixed with additional check in `CallCompleter.updateResultDataFlowInfoUsingEffects`. Nevertheless, I want to discuss this with @erokhins later.

libraries/stdlib/src/kotlin/util/Standard.kt
```diff
@@ -28,60 +28,60 @@ public inline fun TODO(reason: String): Nothing = throw NotImplementedError("An
  * Calls the specified function [block] and returns its result.
  */
 @kotlin.internal.InlineOnly
-public inline fun <R> run(block: () -> R): R = block()
+public inline fun <R> run(@kotlin.internal.CalledInPlace(kotlin.internal.InvocationCount.EXACTLY_ONCE) block: () -> R): R = block()
```
dsavvinov: Currently, we don't have a way to express that function always throws.

compiler/frontend/src/org/jetbrains/kotlin/cfg/ControlFlowInformationProvider.kt
```diff
@@ -367,7 +368,8 @@ class ControlFlowInformationProvider private constructor(
         // Do not consider top-level properties
         if (containingDeclarationDescriptor is PackageFragmentDescriptor) return false
         var parentDeclaration = getElementParentDeclaration(writeValueInstruction.element)
-        while (true) {
+
+        loop@ while (true) {
```
mglukhikh: I recommend some self-describing name for this label

compiler/frontend/src/org/jetbrains/kotlin/cfg/PseudocodeTraverser.kt
```diff
@@ -195,6 +195,7 @@ enum class TraversalOrder {
     BACKWARD
 }
 
+
```
mglukhikh: Remove me

compiler/frontend/src/org/jetbrains/kotlin/cfg/pseudocode/ControlFlowInstructionsGenerator.kt
```diff
@@ -77,13 +78,18 @@ class ControlFlowInstructionsGenerator : ControlFlowBuilderAdapter() {
...
+        super.exitSubroutine(subroutine, invocationKind)
         delegateBuilder.exitBlockScope(subroutine)
         val worker = popBuilder()
         if (!builders.empty()) {
             val builder = builders.peek()
-            builder.declareFunction(subroutine, worker.pseudocode)
+
+            if (invocationKind == null) {
+                builder.declareFunction(subroutine, worker.pseudocode)
+            } else {
```
mglukhikh: Formatting

compiler/frontend/src/org/jetbrains/kotlin/effectsystem/adapters/ValueIDs.kt
mglukhikh: OK, understood

compiler/frontend/src/org/jetbrains/kotlin/types/expressions/ControlStructureTypingVisitor.java
mglukhikh: No, I think it can be done in either way. But please check your logic for statements like `if (1 == 2) { ... } ` does not work in 1.1.

mglukhikh:
My review is finished at this point

mglukhikh:
Probably deprecated

PR #1246, opened 2017-08-15
KT-19634 <boxed_value> == <primitive_value> converted to reference equality in Kotlin

a2kaido:
https://youtrack.jetbrains.com/issue/KT-19634


semoro:
Thanks! 🙂

PR #1248, opened 2017-08-16
keep braces in redundant cascade if

sckm:
#KT-19704 Fixed

https://youtrack.jetbrains.com/issue/KT-19704


idea/testData/inspections/constantConditionIf/test.kt
```diff
@@ -25,4 +25,14 @@ fun bar() {
         baz("a")
         1
     } else 2)
+
+    if(s ==null) {
```
lynnfield: forgot space after equals

PR #1249, opened 2017-08-16
Improve Boolean infix functions' documentation

aoriani:
I and my coworker @nicholaspark09 notice some differences in the generated bytecode when using the `&&` operator and the `and` infix function. 

For
```kotlin
fun andOperator(a: Boolean, b: Boolean) = a && b
```
the generated bytecode is : 

```
   L0
    LINENUMBER 3 L0
    ILOAD 0
    IFEQ L1
    ILOAD 1
    IFEQ L1
   L2
    ICONST_1
    GOTO L3
   L1
    ICONST_0
   L3
    IRETURN
   L4
```

Whereas for 
```kotlin
  fun andInfixFun(a: Boolean, b: Boolean) = a and b
```
the bytecode is : 
```
L0
    LINENUMBER 5 L0
    ILOAD 0
    ILOAD 1
    IAND
    IRETURN
```

In other words,  the Kotlin `&&` operates in similar fashion to Java `&&` operator performing short-circuit evaluation, whereas the `and` infix function -- like Java `&` operator with `Boolean`  args -- does not.  I think that is an important difference in behavior and therefore it should be denoted on the documentation.

### DIFF
--- a/core/builtins/native/kotlin/Boolean.kt
+++ b/core/builtins/native/kotlin/Boolean.kt
@@ -27,12 +27,14 @@ public class Boolean private constructor() : Comparable<Boolean> {
     public operator fun not(): Boolean
 
     /**
-     * Performs a logical `and` operation between this Boolean and the [other] one.
+     * Performs a logical `and` operation between this Boolean and the [other] one. Unlike the `&&` operator,
+     * this function does not perform short-circuit evaluation. Both `this` and [other] will always be evaluated.
      */
     public infix fun and(other: Boolean): Boolean
 
     /**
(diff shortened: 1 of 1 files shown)


yole:
Thanks for your contribution!

yole:
@piotrek1543 Could you please stop marking Kotlin pull requests as approved by you? GitHub doesn't make it obvious whether an approval comes from a contributor to the project, and your approvals may confuse the authors of pull requests and lead them to think that their changes have been approved by the Kotlin team, which is not the case.

PR #1250, opened 2017-08-17
KT-19555 Unnecessary String copy constructor call in Java is converted to red code in Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19555


j2k/testData/fileOrElement/methodCallExpression/stringMethods.kt
```diff
@@ -7,7 +7,7 @@ internal class A {
     @Throws(Exception::class)
     fun constructors() {
         String()
-        // TODO: new String("original");
+        "original"
```
piotrek1543: In `stringMethods` class constructor you used `String("original")`, here you put only its parameter. I am not sure if putting directly `"original"` here is a proper thing.

yole: @piotrek1543 If you can suggest a different way to fix the underlying issue so that J2K does not generate red code, you're welcome to do so. Leaving a comment saying that you're unsure whether something is the right fix is not helpful; please don't.

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -501,6 +501,14 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                 }
             }
 
+            if (type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                val argument = expression.argumentList?.expressions?.singleOrNull()
+                if (argument != null && argument.type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                    result = codeConverter.convertExpression(argument)
```
yole: See my comment to https://youtrack.jetbrains.com/issue/KT-19555

j2k/testData/fileOrElement/methodCallExpression/stringMethods.kt
piotrek1543: @yole yes, you're right. Saying that I'm not sure about a provided solution without providing any helpful clues or suggestions was inappropriate; I won't.

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
t-kameyama: Fixed

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -501,6 +501,18 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                 }
             }
 
+            if (type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                val argument = expression.argumentList?.expressions?.singleOrNull()
+                if (argument != null && argument.type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                    result = if ((argument.reference?.resolve() as? PsiVariable)?.hasModifierProperty(PsiModifier.FINAL) == true) {
```
yole: This is not what I meant as "non-constant strings". Simply check whether the argument is a `PsiLiteralExpression`.

j2k/testData/fileOrElement/methodCallExpression/stringMethods.kt
```diff
@@ -4,10 +4,18 @@ import java.nio.charset.Charset
...
+
     @Throws(Exception::class)
     fun constructors() {
         String()
-        // TODO: new String("original");
+
+        val LOCAL_CONSTANT = "local_constant"
+        val s1 = FIELD_CONSTANT
+        val s2 = LOCAL_CONSTANT
+        val s3 = java.lang.String("original") as String
```
yole: The copy constructor is not needed here.

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
t-kameyama: Fixed

j2k/testData/fileOrElement/methodCallExpression/stringMethods.kt
t-kameyama: Fixed

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -501,6 +501,18 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                 }
             }
 
+            if (type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                val argument = expression.argumentList?.expressions?.singleOrNull()
+                if (argument != null && argument.type?.canonicalText == CommonClassNames.JAVA_LANG_STRING) {
+                    result = if (argument is PsiLiteralExpression) {
+                        codeConverter.convertExpression(argument)
+                    } else {
+                        LiteralExpression("${CommonClassNames.JAVA_LANG_STRING}(${argument.text}) as String").assignPrototype(expression)
```
semoro: Isn't it better to create valid AST here? Via `MethodCallExpression`, `TypeCastExpression` and so on?

t-kameyama:
Fixed

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -522,7 +522,15 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                     result = if (argument is PsiLiteralExpression) {
                         codeConverter.convertExpression(argument)
                     } else {
-                        LiteralExpression("${CommonClassNames.JAVA_LANG_STRING}(${argument.text}) as String").assignPrototype(expression)
+                        val argumentList = ArgumentList.withNoPrototype(listOf(codeConverter.convertExpression(argument)))
+                        val methodCall =
+                            MethodCallExpression.buildNonNull(null, CommonClassNames.JAVA_LANG_STRING, argumentList).assignNoPrototype()
```
semoro: Looks like prototype should be assigned here, instead of type-cast

t-kameyama:
Fixed

goodwinnk:
@t-kameyama Thank you for your and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported. 

There were number of design issues in the old Java to Kotlin converted that forced us to make a major rewrite for the whole module. New converter can be found in `nk2k` directory in the repository and already enabled by default in 1.3.50-eap2. 

When we had started this rewrite we didn't know the final list of issues that is going to be fixed or unresolved at the end. 

KT-19555 is still reproducible in the new Java to Kotlin converter that is especially sad because code from this pull request can't be reused directly to fix the issue. Unfortunately fixing should be started from the ground.

PR #1251, opened 2017-08-18
KT-19351 Repeated annotation @Deprecated after J2K for method with annotation and javadoc

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19351


yole:

PR #1253, opened 2017-08-21
KT-19644 @deprecated with multi-line message in Javadoc converted to @Deprecated("long ugly string")

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19644


yole:

PR #1255, opened 2017-08-21
KT-19337 Counter loop in Java is transformed to less effective loop in Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19337

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/ForConverter.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/ForConverter.kt
@@ -209,7 +209,7 @@ class ForConverter(
         return if (reversed)
             DownToExpression(startConverted, convertBound(bound, if (inclusiveComparison) 0 else +1))
         else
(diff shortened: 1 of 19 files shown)


j2k/src/org/jetbrains/kotlin/j2k/ForConverter.kt
```diff
@@ -209,7 +209,7 @@ class ForConverter(
         return if (reversed)
             DownToExpression(startConverted, convertBound(bound, if (inclusiveComparison) 0 else +1))
         else
-            RangeExpression(startConverted, convertBound(bound, if (inclusiveComparison) 0 else -1))
+            UntilExpression(startConverted, convertBound(bound, if (inclusiveComparison) +1 else 0))
```
yole: I don't think this is correct. `until` should be used only in the case where the original code used `-1`. Cases where `0..9` was generated should remain unchanged.

t-kameyama: Fixed

yole:

PR #1257, opened 2017-08-22
KT-19651 Java class with static-only methods can contain 'protected' members

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19651


semoro:
Thanks! 🙂

PR #1258, opened 2017-08-23
KT-19574 Code with inferred default parameters and parameter vs property name clashes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19574

### DIFF
--- a/j2k/src/org/jetbrains/kotlin/j2k/OverloadReducer.kt
+++ b/j2k/src/org/jetbrains/kotlin/j2k/OverloadReducer.kt
@@ -188,7 +188,14 @@ fun Converter.convertParameterList(
                 val parameter = params[i]
                 val defaultValue = overloadReducer?.parameterDefault(method, i)
                 val defaultValueConverted = if (defaultValue != null)
-                    deferredElement { codeConverter -> codeConverter.correctCodeConverter().convertExpression(defaultValue, parameter.type) }
+                    deferredElement { codeConverter ->
+                        val expr = codeConverter.correctCodeConverter().convertExpression(defaultValue, parameter.type)
+                        val refExpr = (defaultValue as? PsiReferenceExpression)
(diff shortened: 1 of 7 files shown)


j2k/src/org/jetbrains/kotlin/j2k/OverloadReducer.kt
```diff
@@ -188,7 +188,14 @@ fun Converter.convertParameterList(
                 val parameter = params[i]
                 val defaultValue = overloadReducer?.parameterDefault(method, i)
                 val defaultValueConverted = if (defaultValue != null)
-                    deferredElement { codeConverter -> codeConverter.correctCodeConverter().convertExpression(defaultValue, parameter.type) }
+                    deferredElement { codeConverter ->
+                        val expr = codeConverter.correctCodeConverter().convertExpression(defaultValue, parameter.type)
+                        val refExpr = (defaultValue as? PsiReferenceExpression)
```
semoro: It will not work when there are some operators around the expression.
```
boolean isA();
void foo() {
 foo(!isA())
}
void foo(boolean isA);
```

And when property used not in default value for clash position
```
boolean isA();
void foo() {
 foo(false, isA())
}
void foo(boolean isA, boolean isB);
```

semoro: Please fix it and add to test case 🙂

t-kameyama: Fixed

goodwinnk:
@t-kameyama Thank you for this pull request. I'm going to close this pull request without merge as it meant to fix the issue in the old Java To Kotlin converter that is now unsupported. Please check [this comment](https://github.com/JetBrains/kotlin/pull/1250#issuecomment-517478133) for the explanation.

KT-19574 is still unresolved in the new Java to Kotlin converter. 

Again thank you for pull request and sorry it is closed like this.

PR #1259, opened 2017-08-24
Extension points for serialization plugin in JS translator

sandwwraith:
* Plugins for JS CLI compiler now can be loaded via -XPlugin option
* New extension point in project
* JS translator modified to accept synthetic declarations and call
 extension.
* Some functions made public to create a small API

### DIFF
--- a/idea/src/META-INF/extensions/common.xml
+++ b/idea/src/META-INF/extensions/common.xml
@@ -14,6 +14,9 @@
         <extensionPoint name="syntheticResolveExtension"
                         interface="org.jetbrains.kotlin.resolve.extensions.SyntheticResolveExtension"
                         area="IDEA_PROJECT"/>
+        <extensionPoint name="jsSyntheticTranslateExtension"
+                        interface="org.jetbrains.kotlin.js.translate.extensions.JsSyntheticTranslateExtension"
+                        area="IDEA_PROJECT"/>
         <extensionPoint name="findUsagesHandlerDecorator"
                         interface="org.jetbrains.kotlin.plugin.findUsages.handlers.KotlinFindUsagesHandlerDecorator"
                         area="IDEA_PROJECT"/>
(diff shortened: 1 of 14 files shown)


idea/src/META-INF/extensions/common.xml
```diff
@@ -14,6 +14,9 @@
         <extensionPoint name="syntheticResolveExtension"
                         interface="org.jetbrains.kotlin.resolve.extensions.SyntheticResolveExtension"
                         area="IDEA_PROJECT"/>
+        <extensionPoint name="jsSyntheticTranslateExtension"
```
bashor: I think it can be moved to `kotlin2js.xml`

js/js.translator/src/org/jetbrains/kotlin/js/translate/declaration/ClassTranslator.kt
```diff
@@ -83,11 +89,30 @@ class ClassTranslator private constructor(
...
+
+        val companionDescriptor = descriptor.companionObjectDescriptor
+
+        // add non-declared (therefore, not traversed) synthetic companion object
+        if (companionDescriptor is SyntheticClassOrObjectDescriptor) {
+            bodyVisitor.generateClassOrObject(companionDescriptor.syntheticDeclaration, nonConstructorContext, true)
+        }
+
+        // synthetic nested classes
+        descriptor
```
bashor: what for do we to generate any other synthetic nested class?

bashor:
Merged manually with cosmetic changes. 
Thanks!

js/js.translator/src/org/jetbrains/kotlin/js/translate/declaration/ClassTranslator.kt
sandwwraith: In JVM, they are used to generate annotation interfaces synthetic implementation. In JS, they're currently not used (because annotations translated to usual class), but I've made this call to extension for consistency, because it's possible for any extension to declare such classes.

PR #1260, opened 2017-08-24
Support for gradle subplugins in Kotlin2JS compilation

sandwwraith:
Because ServiceLoader can't check generic argument of
KotlinGradleSubplugin in runtime, we have to manually check the type in
`isApplicable`.


h0tk3y:
LGTM, and it has now passed the remote run. Merging.

PR #1262, opened 2017-08-25
KT-16260 Add intention to specify all types explicitly in destructuring assignment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16260


idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyInDestructuringAssignmentIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+        if (element.containingFile is KtCodeFragment) return null
+        val entries = element.noTypeReferenceEntries()
+        if (entries.isEmpty()) return null
+        if (entries.any { SpecifyTypeExplicitlyIntention.getTypeForDeclaration(it).isError }) return null
+        return TextRange(element.startOffset, element.initializer?.let { it.startOffset - 1 } ?: element.endOffset)
+    }
+
+    override fun applyTo(element: KtDestructuringDeclaration, editor: Editor?) {
+        val entries = element.noTypeReferenceEntries()
+        if (editor != null && element.initializer != null)
```
yole: This behaves differently for a destructuring assignment and a destructuring used in a loop:

```
        val map = mapOf(1 to "two")
        for ((key, value) in map) { 
        }
```

The behavior should be the same in both cases.

idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyInDestructuringAssignmentIntention.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.kotlin.types.isError
+
+class SpecifyTypeExplicitlyInDestructuringAssignmentIntention : SelfTargetingRangeIntention<KtDestructuringDeclaration>(
+        KtDestructuringDeclaration::class.java, "Specify all types explicitly in destructuring assignment"
+), LowPriorityAction {
+
+    override fun applicabilityRange(element: KtDestructuringDeclaration): TextRange? {
+        if (element.containingFile is KtCodeFragment) return null
+        val entries = element.noTypeReferenceEntries()
+        if (entries.isEmpty()) return null
```
yole: This is redundant; the next line will also return if `entries` is empty.

idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyInDestructuringAssignmentIntention.kt
t-kameyama: If `&& element.initializer != null` is deleted, the behavior will be same:
```
val map = mapOf(1 to "two")
for ((key, value)<caret> in map) {
}
```

However, destructuring in lambda will not work correctly:
```
val map = mapOf(1 to "two")
map.forEach { (key, value)<caret> ->
}
```
I think the cause is `typeRef.isValid` returns false:
https://github.com/t-kameyama/kotlin/blob/827c57d6ed7492b5f41fbc4ac87b196deebde7bf/idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyIntention.kt#L159
I also confirmed that `typeRef.text` is `Any`.

Is there a way for `typeRef.isValid` to become `true`?

idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyInDestructuringAssignmentIntention.kt
t-kameyama: This is not redundant. `entries.any {}` returns false if `entries` is empty.

t-kameyama: Fixed

idea/src/org/jetbrains/kotlin/idea/intentions/SpecifyTypeExplicitlyInDestructuringAssignmentIntention.kt
t-kameyama: Fixed

yole:
Merged manually (with some minor wording changes). Thanks a lot for your contribution!

PR #1263, opened 2017-08-25
KT-16033 Automatically static import the enum value name when "Add remaining branches" on an enum from another class/file

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16033


yole:
I'm not sure this is actually a good idea. There is no rule or code style requirement saying that enum constants should be always automatically imported, and it's not true that importing them always makes code more readable. I'd rather fix the original issue by providing a separate, explicit action to replace all qualified references to members of a certain class in a file with imports.

t-kameyama:
Added "Add remaining branches with import" quick fix

mglukhikh:
Merged after some polishing. For example, you do not require separate quick-fix class for this purpose.

RyeGordo:
How do we disable this new behaviour?
I can't find a "Add remaining branches with import" available for configuration.

PR #1264, opened 2017-08-25
Fix minor grammar inconsistency: trailing underscores in numeric literals

havrikov:
This change prohibits such illegal numeric literals as `13_37_` or `0xCAFE_BABE_` (note the trailing underscores).


yole:
Thanks!

PR #1265, opened 2017-08-27
KT-19666 Fix replacedBaseClause for selector which has implicit receiver

sckm:
https://youtrack.jetbrains.com/issue/KT-19666

Should i do something instead of error at IfThenUtils L179?
https://github.com/JetBrains/kotlin/compare/master...sckm:KT-19666?expand=1#diff-d15fa721b5c3a5e0d3e1166ae36958ceR179


yole:

PR #1266, opened 2017-08-27
KT-19629 "Convert to primary constructor" quick fix should not move `init{...}` section down

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19629

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertSecondaryConstructorToPrimaryIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertSecondaryConstructorToPrimaryIntention.kt
@@ -18,6 +18,7 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
+import com.intellij.psi.search.searches.ReferencesSearch
 import org.jetbrains.kotlin.descriptors.*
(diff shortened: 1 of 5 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertSecondaryConstructorToPrimaryIntention.kt
```diff
@@ -181,10 +181,15 @@ class ConvertSecondaryConstructorToPrimaryIntention : SelfTargetingRangeIntentio
         with (constructorInClass.replace(constructor)) {
             constructorCommentSaver.restore(this)
         }
-        element.delete()
 
-        if ((initializer.body as? KtBlockExpression)?.statements?.isNotEmpty() ?: false) {
-            klass.addDeclaration(initializer)
+        if ((initializer.body as? KtBlockExpression)?.statements?.isNotEmpty() == true) {
```
yole: Why not simply add after `element` and then delete `element`?

t-kameyama: I think property used in initializer needs to be defined before initializer.

mglukhikh:
I think generally it's good now, but not exactly sure about `ReferencesSearch` usage (it can affect performance). I'll think a little bit and accept it.

mglukhikh:
Finally I merged it. Thank you and sorry for so long reviewing!

PR #1267, opened 2017-08-28
KT-19648 "Redundant setter parameter type" range includes whole parameter declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19648


yole:

PR #1276, opened 2017-08-30
kaptGenerateStubs now depends on kapt build dependencies #KT-20001 fixed

pivotal-vladimir:
https://youtrack.jetbrains.com/issue/KT-20001


yanex:
Merged manually. Thanks!

PR #1280, opened 2017-09-01
KT-19943 Redundant 'toInt' after converting explicit Integer#intValue call

a2kaido:
https://youtrack.jetbrains.com/issue/KT-19943


semoro:
Thanks! 🙂 Merged manually.

PR #1281, opened 2017-09-04
KT-5878 Quickfix for "variable initializer is redundant" (VARIABLE_WITH_REDUNDANT_INITIALIZER)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-5878


yole:

PR #1282, opened 2017-09-04
KT-19514 Redundant getter / setter inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19514


yole:

PR #1284, opened 2017-09-06
KT-17322 Intentions to generate a getter and a setter for a property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17322


mglukhikh:
Thank you, merged after minor refactoring.

PR #1285, opened 2017-09-07
KT-18742 Add quick-fix for CANNOT_CHECK_FOR_ERASED

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18742


idea/src/org/jetbrains/kotlin/idea/quickfix/MakeTypeParameterReifiedAndFunctionInlineFix.kt
```diff
@@ -0,0 +1,72 @@
...
+            AddModifierFix(typeParameter, KtTokens.REIFIED_KEYWORD).invoke(project, editor, file)
+        }
+        else {
+            inlineFix.invoke(project, editor, file)
+            val psiFactory = KtPsiFactory(project)
+            val typeParameterName = KotlinNameSuggester.suggestNameByName("T") { name ->
+                function.typeParameters.none { it.name == name }
+            }
+            function.addTypeParameter(psiFactory.createTypeParameter("reified $typeParameterName"))
+            element.replace(psiFactory.createExpression(typeParameterName))
```
mglukhikh: All this else branch together with examples looks very dangerous to me. Looks like these actions change function semantics so I'd not recommend to perform them. I'd prefer to use this quick-fix only if type parameter exists.

t-kameyama: Fixed

mglukhikh:
Thank you, merged into master.

PR #1286, opened 2017-09-08
Add Quickfix for type mismatch that converts Sequence/Array/Iterable/Collection/(Mutable)List

cypressious:
Fixes #KT-19735

https://youtrack.jetbrains.com/issue/KT-19735

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertCollectionFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertCollectionFix.kt
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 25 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/ConvertCollectionFix.kt
```diff
@@ -0,0 +1,88 @@
...
+
+class ConvertCollectionFix(element: KtExpression, val type: CollectionType) : KotlinQuickFixAction<KtExpression>(element) {
+    override fun getFamilyName(): String = "Convert to ${type.displayName}"
+    override fun getText() = "Convert expression to '${type.displayName}' by inserting '.${type.functionCall}'"
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val expression = element ?: return
+        val factory = KtPsiFactory(expression)
+
+        val replaced = expression.replaced(factory.createExpressionByPattern("$0.$1", expression.text, type.functionCall))
```
mglukhikh: Just `expression` is better here than `expression.text`

mglukhikh: Also, I suppose you may need to take expression into brackets sometimes. E.g. for a call like `functionOnList(a + b)`.

mglukhikh: Full example. A bit exotic but quite possible:
```
fun foo(a: List<String>, b: List<String>) {
    bar(a + b) // expected: bar((a + b).toTypedArray()), got erroneous bar(a + b.toTypedArray())
}
fun bar(a: Array<String>) {}
```

cypressious:
@mglukhikh fixed

mglukhikh:
Merged manually

PR #1289, opened 2017-09-11
KT-18765 Move incremental compilation message from Gradle's warning to info logging level

gildor:
https://youtrack.jetbrains.com/issue/KT-18765


h0tk3y:
Thanks for your contribution!

PR #1252, opened 2017-08-19
Surround with try/catch(/finally) and if/else for expressions

cypressious:
https://youtrack.jetbrains.com/issue/KT-14175
https://youtrack.jetbrains.com/issue/KT-19782

PR #1290, opened 2017-09-11
Implement "Surround with try/catch(/finally)" and "Surround with if/else" for expressions

cypressious:
https://youtrack.jetbrains.com/issue/KT-14175
https://youtrack.jetbrains.com/issue/KT-19782

Continued from #1252


goodwinnk:
I will merge the changes after a remote run, thank you!

I also did some follow-up refactorings, please, take a look at https://github.com/JetBrains/kotlin/compare/rri/nk

PR #1291, opened 2017-09-12
KT-19642 Javadoc comment placement is wrong after default parameters inference in J2K

tommykw:
https://youtrack.jetbrains.com/issue/KT-19642

### DIFF
--- a/j2k/testData/fileOrElement/comments/commentsForConstructors.kt
+++ b/j2k/testData/fileOrElement/comments/commentsForConstructors.kt
@@ -8,8 +8,7 @@ internal class A// this is a primary constructor
 
     // this is a secondary constructor 2
     constructor(s: String) : this(s.length) {} // end of secondary constructor 2 body
-}// this is a secondary constructor 1
-// end of secondary constructor 1 body
+}
 
 internal class B// this constructor will disappear
 (private val x: Int) // end of constructor body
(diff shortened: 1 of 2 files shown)


j2k/testData/fileOrElement/comments/commentsForConstructors.kt
```diff
@@ -8,8 +8,7 @@ internal class A// this is a primary constructor
 
     // this is a secondary constructor 2
     constructor(s: String) : this(s.length) {} // end of secondary constructor 2 body
-}// this is a secondary constructor 1
-// end of secondary constructor 1 body
```
semoro: Not sure that it nice to silently drop the comment. 
It's better to place it after comment for declaration where default parameters inferred now.

tommykw: Thanks, will fix.

goodwinnk:
@tommykw Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

KT-19642 is only partially fixed in the new Java to Kotlin converter but code from this pull request can't be directly used for fixing the rest. Unfortunately fixing should be started from the ground.

PR #1296, opened 2017-09-15
KT-20261 Incorrect "Redundant Unit return type" inspection for Nothing-typed expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20261


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantUnitReturnTypeInspection.kt
```diff
@@ -43,6 +44,7 @@ class RedundantUnitReturnTypeInspection : AbstractKotlinInspection() {
                     if (!function.hasBlockBody()) {
                         val bodyExpression = function.bodyExpression
                         if (bodyExpression != null) {
+                            if (bodyExpression is KtThrowExpression) return
```
mglukhikh: It's not the correct way. `KtThrowException` is not the only one that has `Nothing` type, e.g. `null!!` has the same type. You should check `bodyExpression` type directly.

t-kameyama: Fixed

mglukhikh:
Thank you, merged manually

PR #1297, opened 2017-09-15
Character literals in arithmetic expression in Java converted to red code in Kotlin

paplorinc:
Please review [commit-by-commit](https://github.com/JetBrains/kotlin/pull/1297/commits)

---------

Fixes: [#KT-19359](https://youtrack.jetbrains.com/issue/KT-19359), [#KT-19363](https://youtrack.jetbrains.com/issue/KT-19363)
Related to [#KT-18046](https://youtrack.jetbrains.com/issue/KT-18046)


paplorinc:
Hey @semoro, would you like me to change anything for this to becomes accepted?

j2k/testData/fileOrElement/issues/operationsWithChar.kt
```diff
@@ -42,6 +42,69 @@ internal class Test {
...
+    }
+
+    fun operationsWithCharLiterals() {
+        b('A'.toInt() > '0'.toInt())
+        b('A'.toInt() >= '0'.toInt())
+        b('A'.toInt() < '0'.toInt())
+        b('A'.toInt() <= '0'.toInt())
+
+        b('0'.toInt() == 'A'.toInt())
+        b('0'.toInt() != 'A'.toInt())
```
semoro: This `toInt` calls are redundant. 
It should be possible to not generate it after 2be27d53800c56e2555d75420336904fe02161a8

Please, merge the latest master, and try to work around it.

paplorinc: Sure thing @semoro, doing it now! :)
How do I know where the `toInt` is needed currently, as it seems the comparisons and the `minus` method don't need it, the rest does?

paplorinc: Done

j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt
```diff
@@ -328,6 +319,13 @@ class DefaultExpressionConverter : JavaElementVisitor(), ExpressionConverter {
                 text = text.replace("\\\\([0-3]?[0-7]{1,2})".toRegex()) {
                     String.format("\\u%04x", Integer.parseInt(it.groupValues[1], 8))
                 }
+                val parent = expression.parent
+                if (parent is PsiBinaryExpression) {
+                    val castNeeded = (parent.operationTokenType !in COMPARISON_OPS) || !isChar(parent.rOperand)
+                    if (castNeeded) {
+                        text += ".${Char::toInt.name}()"
+                    }
+                }
```
semoro: Is it possible to check if cast required in `getOperandExpectedType`? 
And pass correct expected type, to let `CodeConverter.convertExpression` handle type conversions.

paplorinc: Thanks @semoro for your patience, was very busy lately.
Let me see what I can do here :)

paplorinc: Hey @semoro, I'm not exactly sure what you're requesting, could you help me out?

semoro: Yep, Now `toInt` added only while converting literal, but it's not sufficient:
```
public class TestArithmWithChars {
    public int foo(char x) {
        return '0' >> x;
    }
}
```
Converted to: 
```
class TestArithmWithChars {
    fun foo(x: Char): Int {
        return '0'.toInt() shr x //.toInt missing
    }
}
```

This problem caused by incorrect type passed to `CodeConverter.convertExpresion`[`CodeConverter.kt:97`](https://github.com/paplorinc/kotlin/blob/e0cf221a18bb9bb6bc33f8df1158ec9592fb9275/j2k/src/org/jetbrains/kotlin/j2k/CodeConverter.kt#L97)

This type incorrectly computed in `ExpressionConverter.visitBinaryExpression` [`ExpressionConverter.kt:113`](https://github.com/paplorinc/kotlin/blob/e0cf221a18bb9bb6bc33f8df1158ec9592fb9275/j2k/src/org/jetbrains/kotlin/j2k/ExpressionConverter.kt#L113)

paplorinc: Fixed, you were right, the fix should have been in `getOperandExpectedType` instead.
I like this version a lot better than the original, thanks again for the hints @semoro! :)

I pushed the new version, would you like me to rebase the changes so that it doesn't contain the wrong modifications?

semoro: I'll rebase it myself on merging, thanks! Will review again on days

goodwinnk:
@paplorinc Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

When we started this rewrite we didn't know the final list of issues that are going to be fixed or unresolved at the end.

[KT-19363](https://youtrack.jetbrains.com/issue/KT-19363) became obsolete after shipping new converter, while [KT-19359](https://youtrack.jetbrains.com/issue/KT-19359) and [KT-18046](https://youtrack.jetbrains.com/issue/KT-18046) are still reproducible. Unfortunately, code from this pull request can't be reused directly to fix those issues and fixing should be started from the ground.

PR #1298, opened 2017-09-15
Improve information regarding running tests that load resources

JLLeitschuh:
Figuring out how to run isolated generated tests is not imidiately straightforward.
This clears up confusion by throwing a more helpful error in the tests and adding a section to the ReadMe.

https://youtrack.jetbrains.com/issue/KT-20271

Assigned @yanex


ReadMe.md
```diff
@@ -148,6 +148,10 @@ You can now run the various Run/Debug Configurations such as
 * All Compiler Tests
 * All IDEA Plugin Tests
 
+### Running specific generated tests
+
+If you need to debug a specific generated test ensure that you have the `Working directory` your IntellJ run configuration set
```
paplorinc: > generated test ensure that

* generated test`,` ensure that

> you have the \`Working directory\` your IntellJ

* you have the \`Working directory\` `in` your Intell`i`J

> If you don't every test

If you don't`,` every test

compiler/tests-common/org/jetbrains/kotlin/test/KotlinTestUtils.java
```diff
@@ -430,7 +431,18 @@ public static String doLoadFile(String myFullDataPath, String name) throws IOExc
     }
 
     public static String doLoadFile(@NotNull File file) throws IOException {
-        return FileUtil.loadFile(file, CharsetToolkit.UTF8, true);
+        try {
+            return FileUtil.loadFile(file, CharsetToolkit.UTF8, true);
+        }
+        catch (FileNotFoundException fileNotFoundException) {
+            // Display the absolute path of the file that wasn't loaded
```
paplorinc: You're explaining the code behavior in a comment.
The code you wrote is clear enough in that regard, no need for the comment :)

ReadMe.md
```diff
@@ -148,6 +148,10 @@ You can now run the various Run/Debug Configurations such as
...
+to the root directory of this project. If you don't every test you try to run will fail with a `No such file or directory` exception.
```
paplorinc: Basically change the test run config from `$MODULE_DIR$` to `$PROJECT_DIR$`, right?
Could we maybe fix that, instead of explaining, i.e. by setting the `JUnit` default to it?

paplorinc: Hmm, the default seems to be stored in the `workspace.xml`, as
```xml
<option name="WORKING_DIRECTORY" value="%MODULE_WORKING_DIR%" />
```

Changing that to
```xml
<option name="WORKING_DIRECTORY" value="$PROJECT_DIR$" />
```
also solves the problem.

Is there a way to fix that automatically?

ReadMe.md
JLLeitschuh: Thanks, will fix those. I knew what I had written didn't sound quite right. COMMAS!

ReadMe.md
JLLeitschuh: I'm not sure. Do you have a suggestion?

paplorinc: No, just questions :)
Maybe we can ask someone from the IntelliJ team on whether it can be fixed properly.

compiler/tests-common/org/jetbrains/kotlin/test/KotlinTestUtils.java
JLLeitschuh: This sound better?
```
/*
 * Unfortunately, the FileNotFoundException will only show the relative path in it's exception message.
 * This clarifies the exception by showing the full path.
 */
```

paplorinc: A lot better, you're explaining the reason for the change, not the code itself! :)

JLLeitschuh:
All of your comments should be resolved except for the one regarding `workspace.xml` which I'm not really sure how to solve.

JLLeitschuh:
I have no idea about `test.ranges.RangeTest`. Maybe you have to open that project from within the embedded version of IntelliJ that the project ships with?

JLLeitschuh:
Thank you all. Great work!

PR #1299, opened 2017-09-17
KT-20259 Show warning if arrays are compared by '!='

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20259


mglukhikh:
Thank you, I will merge it.

PR #1301, opened 2017-09-18
KT-16402 AssertionError on completing expression after template in string literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16402

### DIFF
--- a/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
+++ b/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
@@ -332,7 +332,6 @@ class KotlinCompletionContributor : CompletionContributor() {
                 psiDocumentManager.commitAllDocuments()
 
                 val token = getToken(context.file, document.charsSequence, startOffset)
-                assert(token.node.elementType == KtTokens.IDENTIFIER || token.node.elementType == KtTokens.THIS_KEYWORD)
                 val nameRef = token.parent as KtNameReferenceExpression
 
                 document.insertString(nameRef.startOffset, "{")
--- a/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
+++ b/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
@@ -331,8 +331,7 @@ class KotlinCompletionContributor : CompletionContributor() {
                 val psiDocumentManager = PsiDocumentManager.getInstance(context.project)
                 psiDocumentManager.commitAllDocuments()
 
-                assert(startOffset > 1 && document.charsSequence[startOffset - 1] == '.')
-                val token = context.file.findElementAt(startOffset - 2)!!
+                val token = getToken(context.file, document, startOffset)
                 assert(token.node.elementType == KtTokens.IDENTIFIER || token.node.elementType == KtTokens.THIS_KEYWORD)
                 val nameRef = token.parent as KtNameReferenceExpression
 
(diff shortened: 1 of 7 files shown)


idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
```diff
@@ -344,6 +343,15 @@ class KotlinCompletionContributor : CompletionContributor() {
 
                 super.handleInsert(context)
             }
+
+            private fun getToken(file: PsiFile, document: Document, startOffset: Int): PsiElement {
```
paplorinc: Minor note: the second parameter could probably be generalized to `document.charsSequence`

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
```diff
@@ -344,6 +343,15 @@ class KotlinCompletionContributor : CompletionContributor() {
...
+                assert(startOffset > 1 && document.charsSequence[startOffset - 1] == '.')
+                val token = file.findElementAt(startOffset - 2)!!
+                return if (token.node.elementType == KtTokens.IDENTIFIER || token.node.elementType == KtTokens.THIS_KEYWORD)
+                    token
+                else
+                    getToken(file, document, token.startOffset + 1)
```
paplorinc: Question: how exhaustive is the stopping condition here, i.e. can you imagine a scenario where this would cause a stack overflow or a `NPE` at line `350`?

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
t-kameyama: Thank you, fixed

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
t-kameyama: Stopping condition is line 348

paplorinc: asserts are usually disabled in production. But even if they are not, this will throw an exception, if not met, so it's basically the same as a stack overflow, right?
```java
/**
 * Throws an [AssertionError] if the [value] is false
 * and runtime assertions have been enabled on the JVM using the *-ea* JVM option.
 */
@kotlin.internal.InlineOnly
public inline fun assert(value: Boolean) {
    assert(value) { "Assertion failed" }
}
```

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KotlinCompletionContributor.kt
```diff
@@ -331,8 +331,7 @@ class KotlinCompletionContributor : CompletionContributor() {
                 val psiDocumentManager = PsiDocumentManager.getInstance(context.project)
                 psiDocumentManager.commitAllDocuments()
 
-                assert(startOffset > 1 && document.charsSequence[startOffset - 1] == '.')
-                val token = context.file.findElementAt(startOffset - 2)!!
+                val token = getToken(context.file, document, startOffset)
                 assert(token.node.elementType == KtTokens.IDENTIFIER || token.node.elementType == KtTokens.THIS_KEYWORD)
```
paplorinc: you may want to put this assertion inside the method, to make it self-validating

semoro:
Merged manually. Thanks! 🙂

PR #1302, opened 2017-09-18
Incorrect reversed() results of Long/Int progression with Long.MIN_VALUE and Int.MIN_VALUE steps

paplorinc:
Fixes: [#KT-17176](https://youtrack.jetbrains.com/issue/KT-17176)

### DIFF
--- a/core/builtins/src/kotlin/Progressions.kt
+++ b/core/builtins/src/kotlin/Progressions.kt
@@ -32,6 +32,7 @@ public open class CharProgression
     ) : Iterable<Char> {
     init {
         if (step == 0) throw kotlin.IllegalArgumentException("Step must be non-zero")
+        else if (step == Int.MIN_VALUE) throw kotlin.IllegalArgumentException("Step must have an inverse")
     }
 
     /**
@@ -86,6 +87,7 @@ public open class IntProgression
     ) : Iterable<Int> {
     init {
         if (step == 0) throw kotlin.IllegalArgumentException("Step must be non-zero")
+        else if (step == Int.MIN_VALUE) throw kotlin.IllegalArgumentException("Step must have an inverse")
     }
 
     /**
@@ -140,6 +142,7 @@ public open class LongProgression
     ) : Iterable<Long> {
     init {
         if (step == 0L) throw kotlin.IllegalArgumentException("Step must be non-zero")
+        else if (step == Long.MIN_VALUE) throw kotlin.IllegalArgumentException("Step must have an inverse")
     }
 
     /**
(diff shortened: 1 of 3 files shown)


generators/src/org/jetbrains/kotlin/generators/builtins/progressions.kt
```diff
@@ -70,7 +81,7 @@ public open class $progression
     /**
      * The last element in the progression.
      */
-    public val last: $t = getProgressionLastElement(start.to$incrementType(), endInclusive.to$incrementType(), step).to$t()
+    public val last: $t = getProgressionLastElement(start$cast, endInclusive$cast, step).to$t()
```
paplorinc: should I eliminate the ending cast also?

paplorinc:
Hey @ilya-g, would you like me to change anything for this to becomes accepted?

ilya-g:
@paplorinc no, it's all good here, I think no changes are needed. But we need to discuss this behavior change in the language committee, so it may take some time to be accepted.

generators/src/org/jetbrains/kotlin/generators/builtins/progressions.kt
```diff
@@ -47,6 +53,10 @@ class GenerateProgressions(out: PrintWriter) : BuiltInsSourceGenerator(out) {
             else -> throw IllegalArgumentException()
         }
 
+        val cast = when (kind) {
+            CHAR -> ".toInt()"
+            else -> ""
+        }
```
ilya-g: The cast like `Int.toInt` is a no-op, so it isn't worth to complicate generator code to avoid it.

core/builtins/src/kotlin/Progressions.kt
```diff
@@ -32,6 +32,7 @@ public open class CharProgression
     ) : Iterable<Char> {
     init {
         if (step == 0) throw kotlin.IllegalArgumentException("Step must be non-zero")
+        else if (step == Int.MIN_VALUE) throw kotlin.IllegalArgumentException("Step must have an inverse")
```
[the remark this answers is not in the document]
paplorinc: This is what I meant:
```
println(Int.MIN_VALUE == -Int.MIN_VALUE)
true
```

generators/src/org/jetbrains/kotlin/generators/builtins/progressions.kt
paplorinc: I think it was needed a year ago, when I did the PR

paplorinc:
@ilya-g, do you need me to change anything else?

ilya-g:
I've addressed the remaining issues and merged it into 1.3-M2 branch.
Thanks for the PR.

PR #1304, opened 2017-09-20
KT-15769 Join lines could "convert to expression body"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15769


yole:

PR #1305, opened 2017-09-20
KT-18674 Join Lines should join strings

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18674


idea/testData/joinLines/stringTemplate/simple.kt
```diff
@@ -0,0 +1,4 @@
+fun test1() {
```
paplorinc: Some questions:
* does this work for more than 2 lines, e.g. 
```kotlin
val s = "one"
+ ", two"
+ ", three"
```
* does it work for non-strings, e.g.
```java
val s = "x=" +
0
```

idea/testData/joinLines/stringTemplate/simple.kt.after
```diff
@@ -0,0 +1,3 @@
+fun test1() {
+    val s = <caret>"foobar"
```
paplorinc: should the caret maybe placed to the same place, i.e.
```kotlin
"foo<caret>bar"
```

yole: Yes, the caret needs to be in the same place.

idea/testData/joinLines/stringTemplate/simple2.kt.after
```diff
@@ -0,0 +1,4 @@
+fun test2() {
+    val foo = "1"
+    val s = <caret>"${foo}bar"
```
yole: I think this is surprising behavior. Lines should be joined into a single expression only if both the left hand side and the right hand side of the concatenation are constants.

idea/testData/joinLines/stringTemplate/simple.kt
yole: The Join Lines action joins the line where the caret is with the next line. It always acts on exactly two lines.

paplorinc: Thanks

idea/testData/joinLines/stringTemplate/simple.kt.after
t-kameyama: Fixed

idea/testData/joinLines/stringTemplate/simple2.kt.after
t-kameyama: Fixed

idea/testData/joinLines/stringTemplate/simple.kt
t-kameyama: Fixed

yole:
Thanks a lot for your contribution, and sorry for the delay with the review!

PR #1307, opened 2017-09-21
KT-20091 "Convert object literal to class" should create inner class if necessary

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20091


asedunov:
Merged, thank you!

PR #1308, opened 2017-09-21
Allow to use Gradle property to configure JDK paths

gildor:
Currently, to build Kotlin we should set environment variables with paths to different versions of JDK, but it's not always useful because for example on Mac it's hard to set a global variable and Gradle uses the daemon, so you cannot just add an environment variable to user's bash config.

This change allows to get JDK paths from Gradle project properties, it allows to set paths to global gradle config or pass it using Gradle command line argument `-D` without passing environment variables to gradle command explicitly or configure IDE.

Properties have higher priority than env variables, to allow override it for particular project or particular gradle run

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -40,6 +40,8 @@ In order to build Kotlin distribution you need to have:
 
 For local development, if you're not working on bytecode generation or the standard library, it's OK to have only JDK 8 installed, and to point all of the environment variables mentioned above to your JDK 8 installation.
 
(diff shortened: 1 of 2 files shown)


gildor:
Additional note. I didn't update README of the project, because not sure that Kotlin team wants to push such approach to contributors, but it should be useful in some cases.
But I would happy to add note to readme about alternative way to set JDKs

gildor:
Autodetection doesn't work for my JDKs installed through `sdkman` on Mac
Okay, I will update Readme

gildor:
Added Readme note and rebased

ligee:
Cherrypicked into master. Thanks!

PR #1309, opened 2017-09-21
KT-17204 Add `Assign to property quickfix`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17204


idea/src/org/jetbrains/kotlin/idea/quickfix/AssignToPropertyFix.kt
```diff
@@ -0,0 +1,60 @@
...
+            val expression = diagnostic.psiElement as? KtNameReferenceExpression ?: return null
+
+            val containingClass = expression.containingClass() ?: return null
+            val right = (expression.parent as? KtBinaryExpression)?.right ?: return null
+            val type = expression.analyze().getType(right) ?: return null
+            val name = expression.getReferencedNameAsName()
+
+            val hasAssignableProperty = containingClass.getProperties().any {
+                it.nameAsName == name &&
+                it.isVar &&
```
mglukhikh: I think there is a reason to improve this case for secondary constructors. Inside them, `var` is not required if property does not have its own initializer and is not declared in primary constructor. Code like `foo = foo` is high probable in this case.

idea/src/org/jetbrains/kotlin/idea/quickfix/AssignToPropertyFix.kt
```diff
@@ -0,0 +1,60 @@
...
+class AssignToPropertyFix(element: KtNameReferenceExpression) : KotlinQuickFixAction<KtNameReferenceExpression>(element) {
+
+    override fun getText() = "Assign to property"
+
+    override fun getFamilyName() = text
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val element = element ?: return
+        val psiFactory = KtPsiFactory(element)
+        element.replace(psiFactory.createExpressionByPattern("this.$0", element))
```
mglukhikh: Please consider the case with ambiguous `this`, e.g. ` with (...) { ... }`. NB: this case is relatively rare, so if it appears to be too hard to solve it correctly, leave it as is (may be I think about it myself).

t-kameyama: Fixed

idea/src/org/jetbrains/kotlin/idea/quickfix/AssignToPropertyFix.kt
t-kameyama: Fixed

t-kameyama: Also added a fix for property defined in primary constructor.
```kotlin
class Test(var foo: Int) {
    fun test(foo: Int) {
        <caret>foo = foo
    }
}
```

mglukhikh:
Ok in general. Will polish and merge it.

PR #1310, opened 2017-09-21
KT-14695 Simplify comparison intention produces meaningless statement for assert()

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14695


mglukhikh:
Fixed & merged manually. Now both intention & quick-fix support this feature.

PR #1311, opened 2017-09-21
Implement using() as possible solution for [KT-1475]

FelixEngl:
Here is a possible solution for: [https://youtrack.jetbrains.com/issue/KT-1475](https://youtrack.jetbrains.com/issue/KT-1475)


ilya-g:
Sorry, but the size of this proposal isn't justified by the problem it tries to solve.

PR #1313, opened 2017-09-23
build.gradle.kts: use listOfNotNull() instead of listOf().filterNotNull()

cypressious:


mglukhikh:
Merged manually, thank you

PR #1314, opened 2017-09-25
KT-19134 IntelliJ Color Scheme editor - allow changing color of colons and double colons

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19134


yole:

PR #1315, opened 2017-09-25
Converting property getter to block body must insert explicit return type

cypressious:
Fixes #KT-20417

https://youtrack.jetbrains.com/issue/KT-20417


cypressious:
I was unable to actually run the test, please tell me if it fails.

mglukhikh:
I will check the test and merge it

PR #1316, opened 2017-09-26
`KotlinUAnnotation.psi` returns `PsiAnnotation`

xiexed:


xiexed:
It will be a new version of this change soon.

PR #1317, opened 2017-09-26
Fix assignments as expressions #KT-19346 #KT-19573 #KT-19358 Fixed

dgrachev28:


dgrachev28:
I tried to do this in more general way than it had been done before for multiple assignments.
But there are still some problems, which I haven't solve yet:
1. assignments in loops:
```java
while ((str = nextLine()) != null) {}
```
2. assignments which shouldn't be executed:
```java
if (false && (a = 5) == 5) {}
```

The second one is really subtle because it changes semantics without failing compilation

semoro:
Such cases can be resolved with something like that: 
`if (false && 5.also { a = it } == 5)`
`while(nextLine().also { str = it } != null)`

goodwinnk:
@dgrachev28 Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

All issues ([KT-19346](https://youtrack.jetbrains.com/issue/KT-19346), [KT-19573](https://youtrack.jetbrains.com/issue/KT-19573) and [KT-19358](https://youtrack.jetbrains.com/issue/KT-19358)) are now fixed in new Java to Kotlin converter.

PR #1319, opened 2017-09-27
Avoid ISUB in kotlin.repeat

fvasco:


ilya-g:
Thanks for the PR, I've merged it. 
It really results in less code being produced both in JVM and JS!

fvasco:
RegExp is your friend ;)

    0\.\.\w+ - 1

You can find something like:

    public val Collection<*>.indices: IntRange
        get() = 0..size - 1

and many others

dnpetrov:
Note that lack of `until` in `indices` doesn't really cause much hassle, because `indices` itself is treated as a compiler intrinsic where appropriate. Still, yes, it's a good idea to run ReplaceRangeToWithUntil on stdlib sources.

fvasco:
Hi @ilya-g
I found the same issue in 1.1.60, a slightly different version.

Methods are:

* Iterable.drop(Int)
* List.takeLast(Int)

Updated regexp: `\w+\s*\.\.\s*\w+\s*-\s*1`

ilya-g:
@fvasco Done, I've just used the inspection, that checks that `rangeTo` can be replaced with `until`. Here are the results (preliminary): 81aaf6f1cd05a77d49376f23f8e57080cd6588e0

PR #1320, opened 2017-09-28
KT-16636 Remove parentheses after deleting the last unused constructor parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16636

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedSymbolInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedSymbolInspection.kt
@@ -386,8 +386,15 @@ class SafeDeleteFix(declaration: KtDeclaration) : LocalQuickFix {
     override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
         val declaration = descriptor.psiElement.getStrictParentOfType<KtDeclaration>() ?: return
         if (!FileModificationService.getInstance().prepareFileForWrite(declaration.containingFile)) return
+        val primaryConstructorParameterList = (declaration.parent as? KtParameterList)?.takeIf { it.parent is KtPrimaryConstructor }
         ApplicationManager.getApplication().invokeLater(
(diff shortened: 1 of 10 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/UnusedSymbolInspection.kt
```diff
@@ -386,8 +386,15 @@ class SafeDeleteFix(declaration: KtDeclaration) : LocalQuickFix {
     override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
         val declaration = descriptor.psiElement.getStrictParentOfType<KtDeclaration>() ?: return
         if (!FileModificationService.getInstance().prepareFileForWrite(declaration.containingFile)) return
+        val primaryConstructorParameterList = (declaration.parent as? KtParameterList)?.takeIf { it.parent is KtPrimaryConstructor }
         ApplicationManager.getApplication().invokeLater(
-                { SafeDeleteHandler.invoke(project, arrayOf(declaration), false) },
+                {
+                    SafeDeleteHandler.invoke(project, arrayOf(declaration), null, false, null) {
+                        if (primaryConstructorParameterList != null && primaryConstructorParameterList.parameters.isEmpty()) {
```
cypressious: can be shortened to `if (primaryConstructorParameterList?.parameters?.isEmpty() == true)`

t-kameyama: Thank you, I will fix

t-kameyama: Fixed

mglukhikh:
Accepted. The only point is we should not delete primary constructor for `expect class` in any case. But this is multi-platform stuff and it's easier for me to fix it myself. Thank you.

mglukhikh:
Merged, thanks

PR #1322, opened 2017-09-29
`KotlinUastLanguagePlugin` catching exceptions in conversions

xiexed:
KT-20448, KT-20450


xiexed:
Actually same things done in 85254838f66d3787d4be93fe9206aeedfa955b70

PR #1324, opened 2017-09-30
fix conversion of assignment expression in method call

sckm:
#KT-19573 Fixed

https://youtrack.jetbrains.com/issue/KT-19573


goodwinnk:
@sckm Thank you for your time and effort you put to this pull request, and I'm sorry that we can't merge it now as it does the fix in the previous version of the Java to Kotlin converter that is now unsupported.

When we started this rewrite we didn't know the final list of issues that are going to be fixed or unresolved at the end.

[KT-19573](https://youtrack.jetbrains.com/issue/KT-19573) was one of the issues that is now fixed after this major converter rewrite. 

PR #1325, opened 2017-10-02
KT-20409 useless warning "Remove curly braces" for Chinese character

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20409


mglukhikh:
Ok, thank you, I will test it once more and merge it.

PR #1326, opened 2017-10-02
Support for collections literals in LightAnnotations (KT-20543)

xiexed:


xiexed:
Pushed into **master** manually

PR #1328, opened 2017-10-04
KT-20231 False positive 'Redundant override' when delegated member hides super type override

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20231

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
@@ -19,11 +19,17 @@ package org.jetbrains.kotlin.idea.inspections
 import com.intellij.codeInspection.*
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.idea.actions.generate.findDeclaredFunction
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.descriptorUtil.getSuperInterfaces
 
 class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession) =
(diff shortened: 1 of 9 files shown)


mglukhikh:
Looks good in general. Will look at it once more in IDEA and merge.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -98,4 +105,21 @@ class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLoc
...
+        val superType = (entry as? KtDelegatedSuperTypeEntry)?.typeReference?.let { context[BindingContext.TYPE, it] }
+        val superTypeDescriptor = superType?.constructor?.declarationDescriptor as? ClassDescriptor
+        superTypeDescriptor?.let { isDefinedIn(listOf(it)) } ?: false
+    } ?: false
+}
+
+private fun KtNamedFunction.isDefinedIn(classDescriptors: List<ClassDescriptor>): Boolean {
+    val functionName = name ?: return false
+    return classDescriptors.any {
+        it.findDeclaredFunction(functionName, false) { it.valueParameters == valueParameters } != null
```
mglukhikh: I don't understand how `==` works here, because we have PSI on one side and descriptor on the other side

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -98,4 +105,21 @@ class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLoc
...
+        || isDefinedIn(it.getSuperInterfaces())
```
mglukhikh: Here it's also not clear why super interfaces only and not the super class

mglukhikh:
Discovered a few important things, see comments.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
mglukhikh: Got it! It works in your test just because both lists are empty.

mglukhikh:
@t-kameyama You can either try to fix problems or wait for my fix.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
t-kameyama: I will fix.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
t-kameyama: I think that checking super classes is unnecessary. A class delegation can only use interface and an interface can not inherit from a class.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
t-kameyama: I FIxed and added test case.

mglukhikh:
At this moment I already have a solution.

mglukhikh:
Added to master. Thank you for good tests!

PR #1329, opened 2017-10-04
Delete "Running specific generated tests" from README

shiraji:
This section is valid before kotlin is built by gradle. Since there is
no working directory setting in run configuration, we need to delete
this document


yole:
Thanks!

PR #1331, opened 2017-10-05
KT-16394 "Convert reference to lambda" generates wrong code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16394


mglukhikh:
Thank you, merged

PR #1332, opened 2017-10-06
Fix Different Default and Fallback on Compiler Execution Strategy

yayaa:
Having default and fallback values are different creates some problem for us that in order to solve we need to do some workaround :)

**Problem:**
We have ["Error unmarshaling return header"](https://youtrack.jetbrains.com/issue/KT-17601) and in order to fix, we tried to set `-Dkotlin.compiler.execution.strategy=in-process` but then we got into ["ClassLoader"](https://youtrack.jetbrains.com/issue/KT-20233) issue. Then we realised if you set something different then "in-process" (by accident) it works!

Digging into source code, lead us to figure out that if you don't set the strategy at all, default strategy is "daemon" but if you set anything at all then it goes into either `compileInProcess` or `compileOutOfProcess`. But we cannot set "in-process" because of mentioned issue. 

Right now, our work around is to set "out-of-process" to skip `compileWithDaemon` and fallback to `compileInProcess`. 

**Fix:**
Having daemon check with strategy in same time and decide if it is required. Have it `true` as default in order not to change behaviour. But still make sure, when `org.gradle.daemon` is false; we can set `kotlin.compiler.execution.strategy` to `in-process` to actually run `compileInProcess`.

**Credits:**
Thanks to [Daniel Hartwich](https://github.com/dhartwich1991) and [Svyatoslav Chatchenko](https://github.com/MyDogTom) to looking into and trying to figure out source of the problem.

**Related Issues:**
https://youtrack.jetbrains.com/issue/KT-17601
https://youtrack.jetbrains.com/issue/KT-20233

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
@@ -150,8 +150,9 @@ internal class GradleCompilerRunner(private val project: Project) : KotlinCompil
             kotlinDebug { "Kotlin compiler args: ${argsArray.joinToString(" ")}" }
         }
 
+        val gradleDaemonUsed = System.getProperty("org.gradle.daemon")?.let(String::toBoolean) ?: true
         val executionStrategy = System.getProperty(KOTLIN_COMPILER_EXECUTION_STRATEGY_PROPERTY) ?: DAEMON_EXECUTION_STRATEGY
-        if (executionStrategy == DAEMON_EXECUTION_STRATEGY) {
+        if (executionStrategy == DAEMON_EXECUTION_STRATEGY && gradleDaemonUsed) {
             val daemonExitCode = compileWithDaemon(compilerClassName, compilerArgs, environment)
 
             if (daemonExitCode != null) {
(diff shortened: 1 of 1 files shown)


nguyenhuy:
@yayaa  Thanks for submitting this! We're hitting the same problems described above. In our case, the only strategy that works is "out-of-process". This, our CI gradle command includes `--no-daemon -Dkotlin.compiler.execution.strategy="out-of-process" -Dorg.gradle.daemon=true`. 

Per the code [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt#L166), both`-Dkotlin.compiler.execution.strategy="out-of-process"` and `-Dorg.gradle.daemon=true` are needed to force Kapt to compile in a separate process. As mentioned in Gradle documentation [here](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:ways_to_disable_gradle_daemon), `--no-daemon` has higher precedence over `org.gradle.daemon`, so gradle daemon is not used, which is recommended for CI environments. (Well, to be 100% correct, a daemon's started at the beginning of the build but stopped at the end)

Anyways, I generally agree with your fix and has one small question. Is there a better way to detect if the gradle daemon is running? As you can see above, `--no-daemon` and `--daemon` options can actually defeat the check for `org.gradle.daemon` system property in Kapt.

yayaa:
Hey @nguyenhuy, 

From `--no-daemon` and `org.gradle.daemon` clash perspective i totally agree, and i don't know what would be the better way to find it out instead of reading `org.gradle.daemon` directly. Maybe there is another way to read also `--no-daemon` request and combine those 2. But again, right now, i don't know.

This fix actually in order to cover the loophole caused by default value and fallback value being different. So with this fix, problem might be more visible. And i hope it might take more attention then, because right now, since this loophole exists people might even ignore the real problem by thinking they found a solution. 

I hope we get a quick action on this so we can move further and found a real solution.

yayaa:
Looks like "ClassLoader" problem is solved -> https://youtrack.jetbrains.com/issue/KT-20233
So it's perfect time to merge this! 😊 Then we can benefit it properly!

PR #1333, opened 2017-10-06
Introduce License FIle

dhartwich1991:
Since Kotlin is running under Apache 2.0 license it makes sense to reflect that also in the repository which is why I generated this License file.


yole:
This is already reflected in https://github.com/JetBrains/kotlin/blob/master/license/LICENSE.txt; no need to introduce another copy.

PR #1336, opened 2017-10-10
KT-20484 Add quick fix to add required target to annotation used on a type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20484


mglukhikh:
I'm sorry for long delay. Will process all feature PRs when I have time.

mglukhikh:
Slightly fixed and merged manually. In particular, I forbid to suggest this quick-fix for library annotations or annotations from read only files.

PR #1337, opened 2017-10-11
KT-20349 Convert lambda to reference for trailing lambda inserts parameter names for all arguments if at least one named argument was passed

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20349


mglukhikh:
Looks pretty good, will merge it today.

mglukhikh:
Merged, thank you.

PR #1341, opened 2017-10-12
KT-20631 Inspection to detect use of Unit as a standalone expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20631


mglukhikh:
Merged, thank you

mglukhikh:
I found a bug here: `Unit` should not be reported as redundant when inside lambda, because it can influence type inference. `return Unit` or `return@labeled Unit` can be reported even in this case. I will fix it.

PR #1345, opened 2017-10-18
KT-20492 Offer "Simplify" intention for 'when' expression where only one branch is known to be true

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20492


mglukhikh:
Merged, thanks. BTW I converted it to inspection, it looks more logical.

PR #1347, opened 2017-10-18
Spring-Kotlin: Gutter repaired (KT-20550, KT-20566)

xiexed:
Platform now allows Gutter to be placed only on leafs elements, this patch fixes it for Spring-Kotlin.


asedunov:
Merged, thanks

PR #1348, opened 2017-10-18
KT-20777 idea KDoc: show sample code containing type parameters correctly

sckm:
#KT-20777 Fixed


semoro:
Accepted manually. Thank you!

PR #1351, opened 2017-10-19
Provide samples for stdlib functions (KT-20357)

ltongues:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -43,12 +43,11 @@ class Collections {
         }
 
         @Sample
-        @Suppress("UselessCallOnNotNull")
         fun collectionOrEmpty() {
             val nullCollection: Collection<Any>? = null
             assertPrints(nullCollection.orEmpty(), "[]")
 
-            val collection = listOf('a', 'b', 'c')
+            val collection: List<Char>? = listOf('a', 'b', 'c')
             assertPrints(collection.orEmpty(), "[a, b, c]")
         }
 
@@ -183,12 +182,11 @@ class Collections {
         }
 
         @Sample
-        @Suppress("UselessCallOnNotNull")
         fun listOrEmpty() {
             val nullList: List<Any>? = null
             assertPrints(nullList.orEmpty(), "[]")
 
-            val list = listOf('a', 'b', 'c')
+            val list: List<Char>? = listOf('a', 'b', 'c')
             assertPrints(list.orEmpty(), "[a, b, c]")
         }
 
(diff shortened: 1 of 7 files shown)


ltongues:
@ilya-g Thanks, all fixed. Please have another look.

In addition, I removed the inverted insertion point from the last three binary search samples to keep them simple. Also I made the comparator in `binarySearchByKeyWithComparator()` more interesting to make it sufficiently different from `binarySearchBySelector()` sample.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -23,14 +23,224 @@ import kotlin.test.*
...
+            val boxes = numbers.map { Box(it) }
+            assertPrints(boxes.binarySearchBy(10) { it.value }, "3")
+        }
+
+        @Sample
+        fun binarySearchByKeyWithComparator() {
+            data class Box(val value: Int)
+            val comparator = Comparator<Box> { x, y -> y.value - x.value }
+
+            val numbers = listOf(12, 10, 7, 3, 1)
```
ilya-g: It might be unclear from this example that the list must be sorted so that the sign of `comparison` is monotonically increasing.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -23,14 +23,224 @@ import kotlin.test.*
...
+            val boxes = numbers.map { Box(it) }
+            assertPrints(boxes.binarySearch { comparator.compare(it, Box(10)) }, "1")
```
ilya-g: When one has a `Box` instance to compare with, he could use another overload. This overload might be useful when you have just a `value` (which is not comparable itself or you want to employ some custom comparator for it) and you want to find the matching `Box` instance.

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -71,4 +71,11 @@ class Strings {
         assertPrints(proteins.take(5).toList(), "[Isoleucine, Arginine, Glycine, Arginine, Glutamine]")
     }
 
+    @Sample
+    fun stringToByteArray() {
+        val charset = Charsets.UTF_8
+        val byteArray = "Hello".toByteArray(charset)
```
ilya-g: We can also print the contents of `byteArray` here.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -23,14 +23,224 @@ import kotlin.test.*
...
+            val collection = listOf('a', 'b', 'c')
+            assertTrue(collection.isNotEmpty())
+        }
+
+        @Sample
+        fun collectionOrEmpty() {
+            val nullCollection: Collection<Any>? = null
+            assertPrints(nullCollection.orEmpty(), "[]")
+
+            val collection: List<Char>? = listOf('a', 'b', 'c')
```
ilya-g: Should be `Collection<Char>?` otherwise another overload is called.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -23,14 +23,225 @@ import kotlin.test.*
...
+
+        @Sample
+        fun binarySearchWithComparisonFunction() {
+            data class Box(val value: Int)
+
+            val numbers = listOf(12, 10, 7, 3, 1)
+            val boxes = numbers.map { Box(it) }
+
+            // `boxes` is sorted according to the comparison function
+            assertPrints(boxes.binarySearch { 10 - it.value }, "1")
```
ltongues: I hope this is what you had in mind for this overload. I had a hard time coming up with an example for some reason :)

ltongues:
@ilya-g Thanks, fixed.

ilya-g:
Thank you for the contribution, I've merged this PR manually with some corrections on top of it.

PR #1352, opened 2017-10-20
KT-20622 Remove “Remove explicit type specification” intention for an annotated type case

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20622


mglukhikh:
Merged, thank you

PR #1353, opened 2017-10-20
UastKotlinPsiVariable `psiParent` made lazy to eliminate recursion wh…

xiexed:
…en building UAST parents


xiexed:
Should target `master` branch

PR #1354, opened 2017-10-20
Fix for SOE in Uast

xiexed:


plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/psi/UastKotlinPsiVariable.kt
```diff
@@ -27,7 +23,7 @@ class UastKotlinPsiVariable private constructor(
         name: String,
         typeProducer: () -> PsiType,
         val ktInitializer: KtExpression?,
-        val psiParent: PsiElement?,
+        psiParentProvider: () -> PsiElement?,
```
yole: Minor nit: we're using 'producer' and 'provider' for two parameters that work in exactly the same way. Better be consistent.

PR #1356, opened 2017-10-21
provide samples for Comparisons (KT-20357)

sckm:


ltongues:
@ilya-g 

> General notes:

This is useful, thanks. Maybe we should put this text inside `libraries/stdlib/samples/test/samples/README.md` ?

ilya-g:
> Maybe we should put this text inside libraries/stdlib/samples/test/samples/README.md

@ltongues Sure, I'm going to write one soon.

sckm:
Thank you for the review, I'll try to improve samples.

sckm:
@ilya-g sorry for the late, please take a look.

In addition, at some samples like `thenBy`, a sorting result with comparator before applying the function is added. I hope this helps to understand functions. What do you think about this?

ilya-g:
Very well, now it's much better. I'll just change `thenComparator` sample not to use `listOf(*array)`, since `sorted*` functions always create a copy of list, then squash all these commits and merge them.

ilya-g:
Squashed and merged, thanks!

PR #1357, opened 2017-10-21
Remove stream buffering in readLine (KT-11208)

Meztihn:
[KT-11208](https://youtrack.jetbrains.com/issue/KT-11208)


ilya-g:
Nice, it would be interesting to compare this implementation with PR #1008.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,10 +132,9 @@ public inline fun println() {
     System.out.println()
 }
 
-// Since System.in can change its value on the course of program running,
-// we should always delegate to current value and cannot just pass it to InputStreamReader constructor.
+// Since System.in can change its value on the course of program running, we should always delegate to current value.
 // We could use "by" implementation, but we can only use "by" with interfaces and InputStream is abstract class.
-private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(object : InputStream() {
+private val stdin: InputStream by lazy { object : InputStream() {
```
ilya-g: Since we're not wrapping it in a reader anymore, we could call `read` methods directly on `System.in` and remove this val entirely.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -171,11 +170,36 @@ private val stdin: BufferedReader by lazy { BufferedReader(InputStreamReader(obj
...
+            previous = current
+            current = inputStream.read()
+        }
+
+        if (!(previous == carriageReturn && current == lineFeed)) outputStream.write(previous)
+
+        outputStream.toByteArray()
+    }
+
+    return ByteArrayInputStream(bytes).reader().use { it.readText() }
```
ilya-g: We could just convert these bytes to string with the `ByteArray.toString(encoding)` function.

By the way, I wonder which encoding is used to encode characters in `System.in`?

cy6erGn0m: Unfortunately it is not going to work with UTF-8 as it could contain multi-byte characters with 0x0d or 0x0a inside. For example http://www.fileformat.info/info/unicode/char/020d/index.htm or even http://www.fileformat.info/info/unicode/char/040d/index.htm

libraries/stdlib/src/kotlin/io/Console.kt
cy6erGn0m: I personally like this particular change however isn't it a breaking change?

libraries/stdlib/src/kotlin/io/Console.kt
ilya-g: Perhaps we could use [this solution](https://stackoverflow.com/questions/17227331/decode-bytes-to-chars-one-at-a-time) to decode chars one at a time in the given encoding, though I'm concerned with how much garbage it would produce.

libraries/stdlib/src/kotlin/io/Console.kt
Meztihn: It won't break anything because readLine is the only place where it is used. But leaving it can be useful in future, considering [KT-11207](https://youtrack.jetbrains.com/issue/KT-11207). Anyway I should try to simplify this declaration.

libraries/stdlib/src/kotlin/io/Console.kt
Meztihn: @ilya-g Thanks!
Default Charset in InputStreamReader is platform dependent but `InputStream.reader()` in Kotlin uses `Charsets.UTF_8` as default. For using platform dependent charset much more complicated algorithm is needed. I'm not sure it's worth it so I'll use UTF-8 for now.
I don't think it's necessary to read chars right away. We always can translate needed chars to bytes and work with them. Especially considering that we always have to check `read()` output for -1.
@cy6erGn0m If I understand UTF-8 encoding correctly no other char can contain `\r` or `\n` chars. Their binary representations are `0000 1101` and `0000 1010` respectively and any bytes in multi byte char should start with at least one `1`.
The site with examples is down so I can't check them for now.

ilya-g: >  I'm not sure it's worth it so I'll use UTF-8 for now.

I don't think it's a matter of choice. Making `readLine` only work with UTF-8 default encoding will lead to simply incorrect chars being read from Windows console for example.

cy6erGn0m: @Meztihn to work properly with any encoding you could try to adopt something like this: https://gist.github.com/cy6erGn0m/464453a2059c11f9b31b949aacc3b8ce (feel free to copy it if needed)

Also it is important to write tests: you can use old-school PipedInputStream/PipedOutputStream to verify corner-cases

PR #1358, opened 2017-10-22
KT-20591 Show annotations in parameter info

shiraji:
#KT-20591 Fixed

https://youtrack.jetbrains.com/issue/KT-20591

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinFunctionParameterInfoHandler.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinFunctionParameterInfoHandler.kt
@@ -255,7 +255,7 @@ abstract class KotlinParameterInfoWithCallHandlerBase<TArgumentList : KtElement,
             if (named) append("[")
 
             parameter.annotations.getAllAnnotations().forEach {
-                it.annotation.fqName?.shortName()?.asString()?.let { append("@$it ") }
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/parameterInfo/KotlinFunctionParameterInfoHandler.kt
```diff
@@ -254,6 +254,10 @@ abstract class KotlinParameterInfoWithCallHandlerBase<TArgumentList : KtElement,
         return buildString {
             if (named) append("[")
 
+            parameter.annotations.getAllAnnotations().forEach {
+                it.annotation.fqName?.shortName()?.asString()?.let { append("@$it ") }
```
semoro: Please, extract fqName to variable to reduce safe-call count

shiraji: How about this?

```kotlin
it.annotation.fqName?.let { append("@${it.shortName().asString()} ") }
```

semoro: Cool!

shiraji:
There you go. Thanks!

semoro:
Thanks! 🙂

PR #1362, opened 2017-10-23
KT-20714 Inspection for self-assigment of properties

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20714

### DIFF
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -2622,6 +2622,16 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.SelfAssignmentInspection"
+                     displayName="Self assignment"
+                     groupPath="Kotlin"
+                     groupName="Probable bugs"
+                     enabledByDefault="true"
+                     cleanupTool="true"
+                     level="WARNING"
+                     language="kotlin"
+    />
+
     <referenceImporter implementation="org.jetbrains.kotlin.idea.quickfix.KotlinReferenceImporter"/>
 
     <fileType.fileViewProviderFactory filetype="KJSM" implementationClass="com.intellij.psi.ClassFileViewProviderFactory"/>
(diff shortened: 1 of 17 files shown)


idea/src/META-INF/plugin.xml
```diff
@@ -2622,6 +2622,16 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.SelfAssignmentInspection"
+                     displayName="Self assignment"
+                     groupPath="Kotlin"
+                     groupName="Probable bugs"
+                     enabledByDefault="true"
+                     cleanupTool="true"
```
mglukhikh: Please do not set `cleanupTool = true` if you are not absolutely sure that your inspection should be included into "Code Cleanup" action.

idea/src/org/jetbrains/kotlin/idea/inspections/SelfAssignmentInspection.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.kotlin.psi.KtVisitorVoid
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+
+class SelfAssignmentInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+
+            override fun visitBinaryExpression(expression: KtBinaryExpression) {
+                super.visitBinaryExpression(expression)
+
```
mglukhikh: You do not check that expression is an assignment

idea/src/org/jetbrains/kotlin/idea/inspections/SelfAssignmentInspection.kt
```diff
@@ -0,0 +1,65 @@
...
+                val left = expression.left ?: return
+                val right = expression.right ?: return
```
mglukhikh: Expressions like `foo() = foo()` also can be highlighted by this inspection

mglukhikh: Also it is invoked for `x = other.x`

mglukhikh:
Also I suspect it can be invoked simultaneously with `AssignToPropertyFix`.

PR #1364, opened 2017-10-23
Clean up test skips introduced in KT-17137

egor-n:
We need to clean these up since primitive array `is` checks should work now.


anton-bannykh:
Merged as https://github.com/JetBrains/kotlin/commit/715d5e90ba406d83708fc7ef694e6b9acec30ecc

Sorry for the wait - couldn't merge, because those tests were still being used with TypedArray's enabled up until  https://github.com/JetBrains/kotlin/commit/4a5e9bbc7f2a91f9c1cfb130ba61a6c6d69b6d56

anton-bannykh:
Thanks!

PR #1366, opened 2017-10-23
Provide samples for Iterables (KT-20357)

tommykw:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/iterables.kt
+++ b/libraries/stdlib/samples/test/samples/collections/iterables.kt
@@ -24,16 +24,9 @@ class Iterables {
 
     @Sample
     fun iterable() {
-        val iterator = buildIterator {
-            val list = intArrayOf(1, 2, 3).asIterable().zip((4..6), { a, b ->
-                if (a < b) a else b
-            })
-            yieldAll(list)
-        }
-
-        assertPrints(iterator.next(), "1")
-        assertPrints(iterator.next(), "2")
-        assertPrints(iterator.next(), "3")
+        val iterable = Iterable { buildIterator { yieldAll(1..3) } }
+        val result = iterable.mapIndexed { index, i -> index + i }
+        assertPrints(result, "[1, 3, 5]")
     }
 
     @Sample
--- a/libraries/stdlib/samples/test/samples/collections/iterables.kt
+++ b/libraries/stdlib/samples/test/samples/collections/iterables.kt
@@ -24,7 +24,13 @@ class Iterables {
 
     @Sample
     fun iterable() {
-        val iterator = Iterable { buildIterator { yieldAll(listOf(1, 2, 3)) } }.iterator()
+        val iterator = buildIterator {
+            val list = intArrayOf(1, 2, 3).asIterable().zip((4..6), { a, b ->
+                if (a < b) a else b
+            })
+            yieldAll(list)
+        }
+
         assertPrints(iterator.next(), "1")
         assertPrints(iterator.next(), "2")
         assertPrints(iterator.next(), "3")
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/collections/iterables.kt
```diff
@@ -0,0 +1,43 @@
...
+package samples.collections
+
+import samples.*
+import kotlin.test.*
+
+class Iterables {
+
+    @Sample
+    fun iterable() {
+        val iterator = Sequence { arrayOf(1, 2, 3).iterator() }.iterator()
```
ilya-g: This sample doesn't use the function it is attached to.
The function may be useful for turning into `Iterable` something that has iterator or is an `Iterator`, but isn't `Iterable` itself, for example `buildIterator { }`

libraries/stdlib/samples/test/samples/collections/iterables.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import samples.*
+import kotlin.coroutines.experimental.buildIterator
+import kotlin.test.*
+
+class Iterables {
+
+    @Sample
+    fun iterable() {
+        val iterator = buildIterator { yieldAll(listOf(1, 2, 3)) }
```
ilya-g: This sample still doesn't call `Iterable` function for which it is provided.

tommykw:
Thanks! Could you check this PR, again?

libraries/stdlib/samples/test/samples/collections/iterables.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import samples.*
+import kotlin.coroutines.experimental.buildIterator
+import kotlin.test.*
+
+class Iterables {
+
+    @Sample
+    fun iterable() {
+        val iterator = Iterable { buildIterator { yieldAll(listOf(1, 2, 3)) } }.iterator()
```
ilya-g: Hmm, now it's confusing why do we get an iterator immediately from that `Iterable`: if one wanted to have an iterator, he could use directly that one returned by `buildIterator`.

Usually we want to wrap something in `Iterable` to pass it where `Iterable` is expected. For example there are a lot of extension functions taking `Iterable` as a receiver: `map`, `filter` and so on.

ilya-g: Also list of consecutive numbers here can be replaced with a range `1..3`

tommykw: Thanks, fixed.

libraries/stdlib/samples/test/samples/collections/iterables.kt
```diff
@@ -24,7 +24,13 @@ class Iterables {
 
     @Sample
     fun iterable() {
-        val iterator = Iterable { buildIterator { yieldAll(listOf(1, 2, 3)) } }.iterator()
+        val iterator = buildIterator {
```
ilya-g: Sorry, I don't see `Iterable` function call here again. Have you forgotten for which function this sample is? :)

What I've meant is something like:
```
val someIterable = Iterable { buildIterator { ... } }

// now you can call extensions on the Iterable
val result: List<String> = someIterable.mapIndexed { index, value -> "$index => $value" }
assertPrints(result, "...")
```

tommykw: Really sorry!! I will try to fix samples.

tommykw: Thanks, fixed.

ilya-g:
I've merged it manually with some improvements, thanks.

PR #1370, opened 2017-10-24
Add inspections for implicit and explicit 'this'

cypressious:
Fixes #KT-4580
https://youtrack.jetbrains.com/issue/KT-4580


mglukhikh:
Excellent counter-example found:
```
class Your(val x: Int)
fun Your.foo() = <caret>this.x // Add explicit 'this'
```

after it we get `fun Your.foo() = (this.this).x` together with NPE provoked by broken PSI :)

mglukhikh:
Fixed, slightly simplified, and merged manually. Thank you. NB: I was not sure about `analyzeFully` vs `analyze` here, but looks like `analyze` also works.

cypressious:
@mglukhikh I'm never sure about the difference between the two. Maybe someone should document this for external kontributers.

mglukhikh:
It's too unclear to document it. I can only say you that `analyzeFully()` should have results equivalent to full compilation, and `analyze()` has no such guarantees. E.g., as I can remember `analyze()` will not produce CFG information.

PR #1371, opened 2017-10-25
KT-20763 Wrong resulting code for "add star projection" quick-fix for inner class with generic outer one

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20763


mglukhikh:
Looks good, I will merge it, thank you.

mglukhikh:
Done, thank you again.

PR #1373, opened 2017-10-26
Fix the doc for String.Companion.format

colriot:
Fixes https://youtrack.jetbrains.com/issue/KT-20968

### DIFF
--- a/libraries/stdlib/src/kotlin/text/StringsJVM.kt
+++ b/libraries/stdlib/src/kotlin/text/StringsJVM.kt
@@ -123,22 +123,22 @@ public inline fun String.toCharArray(destination: CharArray, destinationOffset: 
 public inline fun String.format(vararg args: Any?): String = java.lang.String.format(this, *args)
 
 /**
- * Uses this string as a format string and returns a string obtained by substituting the specified arguments,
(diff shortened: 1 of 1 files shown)


libraries/stdlib/src/kotlin/text/StringsJVM.kt
```diff
@@ -137,8 +137,8 @@ public inline fun String.Companion.format(format: String, vararg args: Any?): St
 public inline fun String.format(locale: Locale, vararg args : Any?) : String = java.lang.String.format(locale, this, *args)
 
 /**
- * Uses this string as a format string and returns a string obtained by substituting the specified arguments,
- * using the default locale.
+ * Uses this string as a format string and returns a string obtained by substituting the specified arguments, using
```
ilya-g: I've spotted another copy-paste related issue: it uses not this string, but the provided [format] string.
Could you improve these two overloads' docs as well?

colriot: Sure thing! Should I squash commits?

colriot: Done

PR #1375, opened 2017-10-26
Function sortedSetOf should return SortedSet

colriot:
Fixes https://youtrack.jetbrains.com/issue/KT-20972
At the same time, I'm not sure whether I was supposed to change Public API refs or not. Looking for comments.


h0tk3y:
This change is not binary compatible. Changing the return type of a function has the same effect on binary compatibility as combined removing an existing function and adding a new one with the new return type. This will break compatibility of existing binaries compiled against current or older versions of kotlin-stdlib with the newer stdlib versions. This is not something to be done right away, if ever.

colriot:
Thanks for the link to the whole strategy.
I understand that this is a binary incompatible change, but can't this fix be considered for inclusion in a major version change to 2.0?

PR #1376, opened 2017-10-26
Fix Gradle plugin task name in readme.

artem-zinnatullin:
```console
$ ./gradlew gradlePluginsTest
…
FAILURE: Build failed with an exception.

* What went wrong:
Task 'gradlePluginsTest' not found in root project 'kotlin'. Some candidates are: 'gradlePluginTest'.
```

Correct task name: https://github.com/JetBrains/kotlin/blob/55416341871259c6aef9912fce51fba74f1202ac/build.gradle.kts#L324


ilya-g:
Thanks for spotting!

PR #1377, opened 2017-10-28
Change 'elementAtOrElse' implementation to use 'getOrElse'

colriot:
Fixes https://youtrack.jetbrains.com/issue/KT-20993


ilya-g:
I'm closing this for now according to the reason in the issue — until the compiler is able to eradicate that redundant bytecode during inlining.

colriot:
Got it. Is there any issue tracking this compiler behavior?

PR #1378, opened 2017-10-28
Propose more idiomatic code for FileReadWrite

colriot:
A small rewrite of several functions to use already defined functions and remove duplicated logic.

Also had a question about adding the second parameter to `outputStream()` (with default value) to open for append, and using this function to remove explicit `FileOutputStream` constructors. 
WDYT?

### DIFF
--- a/libraries/stdlib/src/kotlin/io/FileReadWrite.kt
+++ b/libraries/stdlib/src/kotlin/io/FileReadWrite.kt
@@ -50,7 +50,7 @@ public inline fun File.printWriter(charset: Charset = Charsets.UTF_8): PrintWrit
  *
  * @return the entire content of this file as a byte array.
  */
-public fun File.readBytes(): ByteArray = FileInputStream(this).use { input ->
+public fun File.readBytes(): ByteArray = inputStream().use { input ->
     var offset = 0
     var remaining = this.length().let {
         if (it > Int.MAX_VALUE) throw OutOfMemoryError("File $this is too big ($it bytes) to fit in memory.") else it
@@ -129,19 +129,16 @@ public fun File.forEachBlock(action: (buffer: ByteArray, bytesRead: Int) -> Unit
  */
 public fun File.forEachBlock(blockSize: Int, action: (buffer: ByteArray, bytesRead: Int) -> Unit): Unit {
     val arr = ByteArray(blockSize.coerceAtLeast(MINIMUM_BLOCK_SIZE))
-    val fis = FileInputStream(this)
 
-    try {
+    inputStream().use {
         do {
-            val size = fis.read(arr)
+            val size = it.read(arr)
             if (size <= 0) {
                 break
             } else {
                 action(arr, size)
             }
         } while (true)
-    } finally {
-        fis.close()
     }
 }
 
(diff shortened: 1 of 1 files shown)


libraries/stdlib/src/kotlin/io/FileReadWrite.kt
```diff
@@ -129,19 +129,16 @@ public fun File.forEachBlock(action: (buffer: ByteArray, bytesRead: Int) -> Unit
  */
 public fun File.forEachBlock(blockSize: Int, action: (buffer: ByteArray, bytesRead: Int) -> Unit): Unit {
     val arr = ByteArray(blockSize.coerceAtLeast(MINIMUM_BLOCK_SIZE))
-    val fis = FileInputStream(this)
 
-    try {
+    inputStream().use {
         do {
-            val size = fis.read(arr)
+            val size = it.read(arr)
```
ilya-g: Better to add a more descriptive name instead of `it`, as it is inside a multiline lambda, for example `stream` or `input`.

libraries/stdlib/src/kotlin/io/FileReadWrite.kt
```diff
@@ -196,6 +192,5 @@ public fun File.readLines(charset: Charset = Charsets.UTF_8): List<String> {
  * @param charset character set to use. By default uses UTF-8 charset.
  * @return the value returned by [block].
  */
-@RequireKotlin("1.2", versionKind = RequireKotlinVersionKind.COMPILER_VERSION, message = "Requires newer compiler version to be inlined correctly.")
```
ilya-g: This annotation originates from `use` function and is contagious for all inline functions in the standard library, so it would be incorrect to remove it here.

ilya-g:
Regarding the proposed `append` parameter of `outputStream` -- it should be possible to add it (better with an overload in this case), but this proposal should be backed with use cases, which are not covered by `File.appendText` and `appendBytes`. If you know some, feel free to submit the proposal to YouTrack.

libraries/stdlib/src/kotlin/io/FileReadWrite.kt
colriot: I got it. Removed this because `use` was changed to `useLines` and the latter has the proper annotation on itself.

libraries/stdlib/src/kotlin/io/FileReadWrite.kt
colriot: Will change it to `input`.

colriot:
Let's leave `outputStream` as is for now - I don't have a strong need myself, `appendText` and `appendBytes` cover my use cases.

About other changes: I wanted to make the code that doesn't repeat itself and uses provided functions instead of explicit Java constructors. Because now different parts of the code that works with streams look inconsistent to each other. Let's clean it up :)

libraries/stdlib/src/kotlin/io/FileReadWrite.kt
piotrek1543: Agree with @ilya-g, but also wonder if it would be better when we would use in line 56 `length` or eventually `size` instead of mystical `it`.

ilya-g:
@colriot @piotrek1543 Thank you guys for the PR and other ideas.
I've merged it partially and named the length variable in `readBytes`.

PR #1379, opened 2017-10-29
Correcting link on libraries/ReadMe

Megamiun:
Correcting a readme link that doesn't link to the correct place.


yole:
This README is entirely out of date. Libraries are now built with Gradle and not Maven; there's no need to build the compiler separately, because libraries are part of the main project now. I've filed https://youtrack.jetbrains.com/issue/KT-20995 to update the README.

PR #1381, opened 2017-10-30
KT-20981 False positive for "redundant super" in data class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20981


mglukhikh:
Merged, thank you

PR #1382, opened 2017-10-30
KT-13378 Provide ability to configure highlighting for !! in expressions and ? in types

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13378


yole:

PR #1383, opened 2017-11-01
KT-21023 Inspection to highlight variables / functions with implicit `Nothing?` type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21023


t-kameyama:
Fixed
```kotlin
var topVar = null
val topVal = null // no warning
fun topFun() = null // no warning

fun test() {
    var localVar = null
    val localVal = null // no warning
    fun localFun() = null // no warning
}

open class OpenTest {
    open var overridableVar = null
    open val overridableVal = null
    open fun overridableFun() = null

    var notOverridableVar = null
    val notOverridableVal = null // no warning
    fun notOverridableFun() = null // no warning
}
```

mglukhikh:
Merged manually.

PR #1384, opened 2017-11-02
KT-13702 Issue a warning when equals is called recursively within itself

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13702


idea/src/META-INF/plugin.xml
```diff
@@ -2633,6 +2633,16 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.RecursiveEqualsCallInspection"
+                     displayName="Recursive equals call"
+                     groupPath="Kotlin"
+                     groupName="Probable bugs"
+                     enabledByDefault="true"
+                     cleanupTool="true"
```
mglukhikh: Please, do not mark everything as `cleanupTool=true`

idea/src/org/jetbrains/kotlin/idea/inspections/RecursiveEqualsCallInspection.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.containingClass
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.resolve.DescriptorUtils
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.typeUtil.isBoolean
+import org.jetbrains.kotlin.types.typeUtil.isNullableAny
+
+class RecursiveEqualsCallInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
```
mglukhikh: And here, `CleanupLocalInspectionTool` is also not needed

idea/src/org/jetbrains/kotlin/idea/inspections/RecursiveEqualsCallInspection.kt
```diff
@@ -0,0 +1,75 @@
...
+                val classDescriptor = expr.containingClass()?.descriptor ?: return
+
+                val functionDescriptor = expr.getNonStrictParentOfType<KtNamedFunction>()?.descriptor as? FunctionDescriptor ?: return
+                if (functionDescriptor.name.asString() != "equals" ||
+                    !DescriptorUtils.isOverride(functionDescriptor) ||
+                    functionDescriptor.valueParameters.singleOrNull()?.type?.isNullableAny() == false ||
+                    functionDescriptor.returnType?.isBoolean() == false) return
+
+                val context = expr.analyze(BodyResolveMode.PARTIAL)
+                if (expr.operationToken != KtTokens.EQEQ) return
```
mglukhikh: I'd check it from the beginning to avoid more expensive unnecessary checks

t-kameyama:
Fixed

mglukhikh:
Additional checks revealed many false positives here. Really we should report "recursive equals" *only* for checks like `this == other`, with `this` and nothing else on the left and `other` (`equals` parameter) and nothing else on the right. On this stage it's easier for me to fix everything myself, but it's better for you also to know it.

mglukhikh:
Merged with some significant fixes

PR #1386, opened 2017-11-06
Provide samples for Iterators (KT-20357)

tommykw:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/iterators.kt
+++ b/libraries/stdlib/samples/test/samples/collections/iterators.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package samples.collections
+
+import samples.Sample
+import samples.assertPrints
+import java.util.*
+
+class Iterators {
+
+    @Sample
+    fun iteratorForEnumeration() {
+        val list = arrayListOf<String>()
+        val vector = Vector<String>().apply {
+            add("RED")
+            add("GREEN")
+            add("BLUE")
+        }
+        val iterator = vector.elements().iterator()
+        while (iterator.hasNext()) {
+            list.add(iterator.next())
+        }
+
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/collections/iterators.kt
```diff
@@ -0,0 +1,71 @@
...
+            list.add(iterator.next())
+        }
+
+        assertPrints(list.joinToString(":"), "RED:GREEN:BLUE")
+    }
+
+    @Sample
+    fun iterator() {
+        val intIterator = IntProgression.fromClosedRange(1, 3, 1).iterator()
+        val iterator = intIterator.iterator()
```
ilya-g: There's no point in calling `iterator()` when you already have an iterator. Usually this extension is used to satisfy `for` operator convention for an Iterator. For example you can get a MutableIterator from a MutableCollection, then iterate it with `for` operator and do something with it at the same time, namely remove some items.

libraries/stdlib/samples/test/samples/collections/iterators.kt
```diff
@@ -0,0 +1,71 @@
...
+    @Sample
+    fun iteratorForEnumeration() {
+        val list = arrayListOf<String>()
+        val vector = Vector<String>().apply {
+            add("RED")
+            add("GREEN")
+            add("BLUE")
+        }
+        val iterator = vector.elements().iterator()
+        while (iterator.hasNext()) {
```
ilya-g: If you want to get a List from Enumeration, you can just call `Enumeration.toList()`. Usually this `iterator` extension is to safisfy `for` operator convention, so that you could use `for` on `Enumeration`: `for (e in vector.elements()) { ... }`

libraries/stdlib/samples/test/samples/collections/iterators.kt
```diff
@@ -0,0 +1,71 @@
...
+        assertPrints(iterator.next(), "1")
+        assertPrints(iterator.next(), "2")
+        assertPrints(iterator.next(), "3")
+    }
+
+    @Sample
+    fun withIndexIterator() {
+        var result = 0
+        val iterator = (1..3).iterator()
+        for ((index) in iterator.withIndex()) {
```
ilya-g: The value is not used here. 
This sample may be hard to understand, because this iterator itself returns integers and indices are also integers, so they can be easily confused.

libraries/stdlib/samples/test/samples/collections/iterators.kt
```diff
@@ -0,0 +1,71 @@
...
+            result += index
+        }
+
+        assertPrints(result, "3")
+    }
+
+    @Sample
+    fun forEachIterator() {
+        var result = 0
+        (1..3).iterator().forEach {
```
ilya-g: It's worth to extract `iterator` into a variable, otherwise it's unclear why can't one call `forEach` directly on the range.
We also can do something with `iterator` before calling `forEach` on it, for example skip some entries.

tommykw:
Thank you for the review. I will try to fix samples.

tommykw:
Please take another look

ilya-g:
Thank you, I've merged it with some corrections (dca23e339a)

PR #1387, opened 2017-11-09
Added samples for map filtering operators (KT-20357)

vlad-kasatkin:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/maps.kt
+++ b/libraries/stdlib/samples/test/samples/collections/maps.kt
@@ -143,10 +143,10 @@ class Maps {
             val originalMap = mapOf("key1" to 1, "key2" to 2, "key3" to 3)
             val destinationMap = mutableMapOf("key40" to 40, "key50" to 50)
 
-            val filteredToMap = originalMap.filterTo(destinationMap) { it.value < 3 }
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/collections/maps.kt
```diff
@@ -104,7 +104,104 @@ class Maps {
...
+
+        @Sample
+        fun filterTo() {
+            val originalMap = mapOf("key1" to 1, "key2" to 2, "key3" to 3)
+            val destinationMap = mutableMapOf("key40" to 40, "key50" to 50)
+
+            val filteredToMap = originalMap.filterTo(destinationMap) { it.value < 3 }
+
+            //destination map is updated with filtered items from the original map
+            assertEquals(destinationMap, filteredToMap)
```
ilya-g: Better to assert `assertTrue(destinationMap === filteredMap)` to emphasize it's the same instance. Same in `filterNotTo` example.

pivotal-vladimir: Thank you for your feedback! I have updated samples. Will add a couple of more samples if I have some free time next week :)

ilya-g:
Thank you. I've squashed the commits and merged them to master. They should get into the documentation with the release of 1.2.20.

PR #1389, opened 2017-11-10
Change run icon in gutter to default one from IJ Platform (KT-15254)

bulenkov:
Kotlin plugin adds very attractive and distractive icon into the gutter for main methods. I propose to use well-known platform icon for that which is suitable and has familiar metaphor instead of just Kotlin logo.


sashache:
Ref: https://youtrack.jetbrains.com/issue/KT-15254

yole:
@piotrek1543 I understand your good intentions, but really this isn't helpful. @bulenkov is a JetBrains engineer. The Kotlin project does not have any policy on pull request titles and does not use them in any part of the process.

PR #1390, opened 2017-11-10
More entry points for serialization compiler plugin

sandwwraith:
* In Kotlin/JS JPS compilation
* In kotlin-common CLI compiler
* In kotlin-common gradle build

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/metadata/K2MetadataCompiler.kt
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/metadata/K2MetadataCompiler.kt
@@ -27,6 +27,7 @@ import org.jetbrains.kotlin.cli.common.messages.OutputMessageUtil
 import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.cli.jvm.config.addJvmClasspathRoots
+import org.jetbrains.kotlin.cli.jvm.plugins.PluginCliParser
 import org.jetbrains.kotlin.codegen.CompilationException
 import org.jetbrains.kotlin.config.CommonConfigurationKeys
 import org.jetbrains.kotlin.config.CompilerConfiguration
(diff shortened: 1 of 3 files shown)


compiler/cli/src/org/jetbrains/kotlin/cli/metadata/K2MetadataCompiler.kt
```diff
@@ -53,6 +54,9 @@ class K2MetadataCompiler : CLICompiler<K2MetadataCompilerArguments>() {
     ): ExitCode {
         val collector = configuration.getNotNull(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY)
 
+        val plugLoadResult = PluginCliParser.loadPluginsSafe(arguments, configuration)
```
udalov: `val pluginLoadResult` please

compiler/cli/src/org/jetbrains/kotlin/cli/metadata/K2MetadataCompiler.kt
```diff
@@ -53,6 +54,9 @@ class K2MetadataCompiler : CLICompiler<K2MetadataCompilerArguments>() {
     ): ExitCode {
         val collector = configuration.getNotNull(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY)
 
+        val pluginLoadResult = PluginCliParser.loadPluginsSafe(arguments, configuration)
+        if (pluginLoadResult != ExitCode.OK) return pluginLoadResult
```
bashor: I think would be better to report something here.

cc @udalov

sandwwraith: Some errors reported in `loadPluginsSafe` function:

https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/plugins/PluginCliParser.kt#L43

udalov: Looks fine to me

alexey-tsvetkov:
The changes themselves look good to me

compiler/cli/src/org/jetbrains/kotlin/cli/metadata/K2MetadataCompiler.kt
bashor: ok

sandwwraith:
@alexey-tsvetkov commit message itself tells about "entry points for compiler plugins", I've just changed PR title to be more precise. Sorry for misleading.

udalov:
@sandwwraith I suppose this can be closed?

PR #1391, opened 2017-11-14
KT-20894 "Add type" quick fix does not take into account `vararg` modifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20894


mglukhikh:
Thank you, merged

PR #1392, opened 2017-11-14
KT-21104 Do not propose to make local lateinit var immutable

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21104


mglukhikh:
Thank you, merged.

PR #1393, opened 2017-11-14
Change UNINITIALIZED_VALUE from object to Any()

LouisCAD:
This shouldn't change any behavior as `UNINITIALIZED_VALUE` is only used to check for reference equality.
However, this change saves a class since `object` creates a single use class, by using instead a field.

### DIFF
--- a/libraries/stdlib/src/kotlin/util/Lazy.kt
+++ b/libraries/stdlib/src/kotlin/util/Lazy.kt
@@ -105,7 +105,7 @@ public enum class LazyThreadSafetyMode {
 }
 
 
-private object UNINITIALIZED_VALUE
+private val UNINITIALIZED_VALUE = Any()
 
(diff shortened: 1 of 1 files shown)


ilya-g:
Though it was noted that making `UNINITIALIZED_VALUE` just an instance of `Any` may complicate debugging, because instead of the name of the class `UNINITIALIZED_VALUE@5305068a` you'll get something like `java.lang.Object@5305068a` when evaluating/watching over lazy instance fields.

The proposed solution is to introduce internal enum for all such constant-like objects in the standard library.

LouisCAD:
@ilya-g Could as well be just a `String` wrapper that prints its value in `toString()`, which could be reused in all parts of the stdlib. Cleaner than the enum IMO because the purpose of the objects stays where it's used, and not in a god place.

PR #1396, opened 2017-11-17
Provide samples for ReversedViews (KT-20357)

tommykw:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/reversedViews.kt
+++ b/libraries/stdlib/samples/test/samples/collections/reversedViews.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


tommykw:
Thank you for the review, please take a look.

libraries/stdlib/samples/test/samples/collections/reversedViews.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Sample
+    fun asReversedMutableList() {
+        val original = mutableListOf(1, 2, 3, 4, 5)
+        val reversed = original.asReversed()
+
+        assertPrints(original, "[1, 2, 3, 4, 5]")
+        assertPrints(reversed, "[5, 4, 3, 2, 1]")
+
+        original.add(6)
```
ilya-g: I thinks it is enough to show only two examples of modification: structural like `add` or `remove`, and non-structural like `set`.

ilya-g: It should be shown how changes of `reversed` affect the `original` list.

libraries/stdlib/samples/test/samples/collections/reversedViews.kt
```diff
@@ -0,0 +1,60 @@
...
+import samples.assertPrints
+
+class ReversedViews {
+    @Sample
+    fun asReversedList() {
+        val original = ('a'..'e').toList()
+        val reversed = original.asReversed()
+
+        assertPrints(original, "[a, b, c, d, e]")
+        assertPrints(reversed, "[e, d, c, b, a]")
```
ilya-g: It should be shown, how changes in `original` affect `reversed` list in case if `original` is a mutable list.
Like:
```
val original = mutableListOf(...)
val originalReadOnly = original as List<...>
val reversed = originalReadOnly.asReversed()
// reversed is read-only list, but can be changed by modifying original one
```

tommykw:
Done. Thank you for the review, please take a look.

ilya-g:
Thanks, I've merged the samples with some minor edits

PR #1399, opened 2017-11-21
KT-11023 Inspection to highlight usages of Collections.sort() and replace them with .sort() method from Kotlin stdlib

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11023


idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,104 @@
...
+                (valueArgsSize == 1 || valueArgsSize == 2) && isMutableList(valueArgs[0])
+            else ->
+                false
+        }
+    }
+
+    private fun isMutableList(arg: KtValueArgument?): Boolean {
+        if (arg == null) return false
+        val expression = arg.getArgumentExpression() ?: return false
+        val context = expression.analyze(BodyResolveMode.PARTIAL)
```
cypressious: You are analyzing the same code twice. You should reuse the `context` from above.

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    private fun canReplaceWithStdLib(fqName: String, callExpression: KtCallExpression, context: BindingContext): Boolean {
+        if (!fqName.startsWith("java.util.Collections.")) return false
+        val valueArgs = callExpression.valueArguments
+        val valueArgsSize = valueArgs.size
+        if (valueArgsSize < 1) return false
+        if (!isMutableList(valueArgs[0], context)) return false
+        return when (fqName) {
```
yole: The `fill()` and `shuffle()` methods have been added in Kotlin 1.2. You need to check the API level for the current module, and offer these replacements only if the API level is at least 1.2.

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,113 @@
...
+            "java.util.Collections.reverse" -> valueArgsSize == 1
+            "java.util.Collections.shuffle" -> valueArgsSize == 1 || valueArgsSize == 2
+            "java.util.Collections.sort" -> valueArgsSize == 1 || valueArgsSize == 2
+            else -> false
+        }
+    }
+
+    private fun isMutableList(arg: KtValueArgument?, context: BindingContext): Boolean {
+        val expression = arg?.getArgumentExpression() ?: return false
+        return expression.getType(context)?.nameIfStandardType?.asString() == "MutableList"
```
yole: Please don't use `nameIfStandardType` (this is a JS frontend function, and should not be used in common code), and please use `KotlinBuiltIns.mutableList`.

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+private class ReplaceWithSortFix(private val fqName: String) : LocalQuickFix {
```
yole: The name of this class is no longer appropriate now that you've extended it to support other functions.

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,113 @@
...
+    override fun getName() = "Replace with stdlib"
```
yole: This should show the specific replacement.

t-kameyama:
Fixed. Please review again.

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,123 @@
...
+            "java.util.Collections.reverse" -> size == 1
+            "java.util.Collections.shuffle" -> checkApiVersion(ApiVersion.KOTLIN_1_2, expression) && (size == 1 || size == 2)
+            "java.util.Collections.sort" -> size == 1 || size == 2
+            else -> false
+        }
+    }
+
+    private fun checkApiVersion(requiredVersion: ApiVersion, expression: KtDotQualifiedExpression): Boolean {
+        val module = ModuleUtilCore.findModuleForPsiElement(expression) ?: return true
+        val version = getRuntimeLibraryVersion(module) ?: return true
```
yole: Please use `module.languageVersionSettings.apiVersion` instead of this process (this honors the API version specified in the Gradle build or facet settings).

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,123 @@
...
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val expression = descriptor.psiElement as? KtDotQualifiedExpression ?: return
+        val callExpression = expression.callExpression ?: return
+        val valueArguments = callExpression.valueArguments
+        val arg1 = valueArguments.getOrNull(0)?.getArgumentExpression()
+        val arg2 = valueArguments.getOrNull(1)?.getArgumentExpression()
+        val factory = KtPsiFactory(project)
+        val newExpression = when (methodName) {
```
yole: Please unify common cases so there's less duplication in this code

idea/src/org/jetbrains/kotlin/idea/inspections/JavaCollectionsStaticMethodCallInspection.kt
```diff
@@ -0,0 +1,123 @@
...
+                arg1 != null -> factory.createExpressionByPattern("$0.reverse()", arg1)
+                else -> null
+            }
+            "shuffle" -> when {
+                arg1 != null && arg2 != null -> factory.createExpressionByPattern("$0.shuffle($1)", arg1, arg2)
+                arg1 != null -> factory.createExpressionByPattern("$0.shuffle()", arg1)
+                else -> null
+            }
+            "sort" -> when {
+                arg1 != null && arg2 != null -> factory.createExpressionByPattern("$0.sortWith(Comparator $1)", arg1, arg2.text)
```
yole: This syntax works only if the argument to `Collections.sort()` is a lambda, and I don't see any check where you make sure that it is one.

t-kameyama:
Fixed. Please review.

yole:

PR #1402, opened 2017-11-22
Remove redundant let supports for "in" operator

shiraji:
#KT-21373 Fixed

https://youtrack.jetbrains.com/issue/KT-21373

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
@@ -18,6 +18,7 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import org.jetbrains.kotlin.idea.inspections.IntentionBasedInspection
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.anyDescendantOfType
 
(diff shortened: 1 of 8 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSingleLineLetIntention.kt
```diff
@@ -102,6 +103,7 @@ class ReplaceSingleLineLetIntention : SelfTargetingOffsetIndependentIntention<Kt
         return when (right) {
             is KtNameReferenceExpression -> right.text != parameterName
             is KtDotQualifiedExpression -> !right.hasLambdaExpression() && !right.nameUsed(parameterName)
+            is KtBinaryExpression -> right.operationToken == KtTokens.RANGE
```
mglukhikh: Don't quite understand why only `RANGE` is allowed here. Please consider other binary operators. It seems to me (may be I miss something) that it should be allowed for most of them.

shiraji: I wasn't too sure which case can be like this format.
```
Expression BinaryOperator KtBinaryExpression
```
In case, if there is unexpected syntax, it would break this intention.

---

I found the current implementation cannot catch this 

```
listOf(1,2,3).filter { it.let { it in IntRange(1, 10) } }
```

Let me fix this.

shiraji:
@mglukhikh Could you review this, again? Thanks

mglukhikh:
Looks good. I will test it and then merge.

mglukhikh:
Slightly changed and merged, thank you.

PR #1403, opened 2017-11-22
KT-15941 "Convert to secondary constructor" produces invalid code for generic property with default value

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15941


mglukhikh:
Don't quite understand the situation, will recheck myself. But probably Ok, thank you.

mglukhikh:
PSI problem, understood. Commit message changed, merged, thank you again. BTW it could affect some other IDE actions using this part of PSI factory.

PR #1404, opened 2017-11-24
KT-20435 False "function is never used" warning

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20435


idea/src/org/jetbrains/kotlin/idea/inspections/UnusedSymbolInspection.kt
```diff
@@ -270,6 +270,8 @@ class UnusedSymbolInspection : AbstractKotlinInspection() {
             return false
         }
 
+        if (!ReferencesSearch.search(declaration, useScope).forEach(::checkReference)) return true
```
mglukhikh: It's Ok but I don't like that we now invoke references search twice, it can affect performance. I'd consider here to do this only for particular case with `JvmName`. Will fix myself.

mglukhikh:
Fixed and merged, thank you

PR #1405, opened 2017-11-24
KT-20429 False-positive 'Unused return value of a function with lambda expression body' in enum constant constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20429


idea/src/org/jetbrains/kotlin/idea/inspections/UnusedLambdaExpressionBodyInspection.kt
```diff
@@ -59,7 +60,8 @@ class UnusedLambdaExpressionBodyInspection : AbstractKotlinInspection() {
         }
     }
 
-    private fun KtExpression.used(context: BindingContext): Boolean = context[BindingContext.USED_AS_EXPRESSION, this] ?: true
+    private fun KtExpression.used(context: BindingContext): Boolean =
+            context[BindingContext.USED_AS_EXPRESSION, this] ?: true || getStrictParentOfType<KtValueArgument>() != null
```
yole: This is not a very good fix. A value argument may contain a lambda with a lot of code, and your check would consider all expressions in the lambda to be used, which could hide many problems. I think the right fix there would be to understand why USED_AS_EXPRESSION is not recorded in this case, and to fix it if possible.

t-kameyama: Fixed. Please review again.

mglukhikh:
Edited and merged, thank you

PR #1407, opened 2017-11-28
Fix for KT-19188

ramonwirsch:
Fixes nondeterministic Default Method Order under existance of generic
type parameters.
SubstitutingScope did not respect source code method order by not using
a LinkedHashSet


dzharkov:
I've squashed your commits and pushed them manually.

PR #1408, opened 2017-11-28
Recognize injections in annotations (KT-13636)

bendowski:
Recognize patterns from both Java (psiMethod) and Kotlin
(kotlinParameter) that match the annotation argument.


xiexed:
already done. but only for *IDEA 173* and should already work with released Kotlin plugins https://github.com/JetBrains/kotlin/pull/1254/files

bendowski:
I'm sorry for being a bit confused by the process :) I got the email below
but can't see the comment in neither GitHub nor Youtrack. I'm also not sure
where I should leave my comments...

Anyway: so it seems this has been fixed in in September. Does it mean we
can close the bug and count on this change shipping in the next Kotlin
version for Android Studio?

xiexed:
Sorry, It was me, I've posted a comment about some similar work done towards this and then I've deleted it.

Yes, this issue is kind of fixed in idea 173, but probably you PR still will be merged in some way. There are some concerns about automatic injections of all java-patterns into Kotlin because there are some known issues about it. So we need some time to decide what is better to do.

And yes Github is quite ok for discussions.

goodwinnk:
Thank you for the contribution! Your commit was merged manually and now is in master:

https://github.com/JetBrains/kotlin/commit/6f350c032d51a995028eb7fb695d82a5cfe51638

PR #1409, opened 2017-11-29
KT-19977 Convert Lambda to reference produces red code when wrong implicit receiver is in scope

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19977


mglukhikh:
Looks good, will merge it

mglukhikh:
Merged, thank you

PR #1410, opened 2017-11-30
KT-18308 'Remove braces from else statement' intention breaks code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18308

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveBracesIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveBracesIntention.kt
@@ -20,7 +20,9 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
+import org.jetbrains.kotlin.KtNodeTypes
 import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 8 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/RemoveBracesIntention.kt
```diff
@@ -61,8 +63,16 @@ class RemoveBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class
...
+        val factory = KtPsiFactory(block)
+
         if (construct is KtDoWhileExpression) {
-            newElement.parent!!.addAfter(KtPsiFactory(block).createNewLine(), newElement)
+            newElement.parent!!.addAfter(factory.createNewLine(), newElement)
+        }
+
+        if (construct is KtIfExpression &&
+            container.node.elementType == KtNodeTypes.ELSE &&
+            construct.getStrictParentOfType<KtDotQualifiedExpression>() != null) {
```
mglukhikh: Why do you consider just this case? What if we have (e.g.) KtSafeQualifiedExpression instead?

mglukhikh: I'd say any `KtExpression` at next level should work here.

mglukhikh:
Fixed and merged Thanks.

PR #1414, opened 2017-12-01
KT-21544 "Add type" quick fix incorrectly processes `vararg` modifier with primitive type array initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21544


yole:

PR #1415, opened 2017-12-01
Uast: returning of secondary constructors bodies (KT-21575)

xiexed:


plugins/uast-kotlin/testData/SuperCalls.log.txt
```diff
@@ -47,6 +47,11 @@ UFile (package = )
         UAnnotationMethod (name = A)
             UParameter (name = str)
                 UAnnotation (fqName = org.jetbrains.annotations.NotNull)
+            UBlockExpression
+                UBlockExpression
```
xiexed: Probably should be unwrapped also as in secondary ones

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUClass.kt
```diff
@@ -139,9 +139,21 @@ class KotlinConstructorUMethod(
                 initializers.forEach {
```
xiexed: `initializers` should be here only for primary constructors

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUClass.kt
```diff
@@ -128,27 +128,37 @@ class KotlinConstructorUMethod(
...
+                bodyExpressions.forEach {
+                    addIfNotNull(languagePlugin.convertOpt(it, uastParent))
                 }
             }
         }
     }
+
+    private fun getBodyExpressions(): List<KtExpression> {
+        if (isPrimary) return getInitializers()
+        val bodyExpression = (psi.kotlinOrigin as? KtFunction)?.bodyExpression ?: return emptyList()
```
yole: I think that, if a class has no primary constructor, the body of each secondary constructor should include the list of initializer blocks prepended to it. This will match the behavior of the JVM backend.

xiexed: but then it is ambiguity on who will be the parent of these elements when building tree from down-to-top

xiexed: and also if in `init` block there will be something to report then it will be reported number of secondary constructors times. So it is not a good idea to copy initializer blocks.

yole: Then maybe copy it only under the first secondary constructor?

xiexed: it will go against the truth, but I agree maybe it is better than just skip it... Ok lets go this way

PR #1416, opened 2017-12-03
.gitignore: remove duplicates

YusukeHosonuma:
Remove duplicate `.idea/modules.xml` from `.gitignore`.

Another declaration is in L33.
https://github.com/JetBrains/kotlin/pull/1416/files#diff-a084b794bc0759e7a6b77810e01874f2L33


yole:
Thanks!

PR #1419, opened 2017-12-05
KT-21603 "Join declaration and assignment" should remove 'lateinit' for 'var'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21603


mglukhikh:
Ok, will merge it

PR #1420, opened 2017-12-05
Kotlin Uast decoupled from Java Uast

xiexed:


plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -71,16 +77,53 @@ abstract class AbstractKotlinUVariable(givenParent: UElement?)
 
     override fun getContainingFile(): PsiFile = unwrapFakeFileForLightClass(psi.containingFile)
 
-    override val annotations: List<UAnnotation> by lz { psi.annotations.map { JavaUAnnotation(it, this) } }
+    override val annotations by lz {
+        val sourcePsi = sourcePsi ?: return@lz psi.annotations.map { WrappedUAnnotation(it, this) }
+        SmartList<UAnnotation>(KotlinNullabilityUAnnotation(sourcePsi, this)).also { annotations ->
```
yole: I wouldn't use `also()` here

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -71,16 +77,53 @@ abstract class AbstractKotlinUVariable(givenParent: UElement?)
...
+            (sourcePsi as? KtModifierListOwner)?.annotationEntries?.
```
yole: Please wrap before `?.`

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -110,18 +153,33 @@ class KotlinUVariable(
...
 ) : AbstractKotlinUVariable(givenParent), UParameter, PsiParameter by psi {
 
-    override val javaPsi = unwrap<UParameter, PsiParameter>(psi)
+    final override val javaPsi = unwrap<UParameter, PsiParameter>(psi)
 
     override val psi = javaPsi
 
+    private val isLightConstructorParam by lz { psi.getParentOfType<PsiMethod>(true)?.isConstructor }
+
+    private val isKtConstructorParam by lz { sourcePsi?.getParentOfType<KtCallableDeclaration>(true)?.let { it is KtPrimaryConstructor } }
```
yole: `isKtPrimaryConstructorParam`?

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -110,18 +153,33 @@ class KotlinUVariable(
...
+
+    override fun acceptsAnnotationTarget(target: AnnotationUseSiteTarget?): Boolean {
+        if (sourcePsi !is KtParameter) return false
+        if (isKtConstructorParam == isLightConstructorParam && target == null) return true
+        when (target) {
+            AnnotationUseSiteTarget.CONSTRUCTOR_PARAMETER -> return isLightConstructorParam == true
+            AnnotationUseSiteTarget.SETTER_PARAMETER -> return isLightConstructorParam != true
+            else -> return false
```
yole: Looks like this won't accept `@param:Foo` annotation on regular (non-constructor) parameters.

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -139,6 +197,48 @@ open class KotlinUParameter(
...
+        get() = null
+    override val javaPsi: PsiElement?
+        get() = null
+    override val sourcePsi: PsiElement?
+        get() = null
+    override val qualifiedName: String?
+        get() = when (nullability) {
+            TypeNullability.NOT_NULL -> NotNull::class.qualifiedName
+            TypeNullability.NULLABLE -> Nullable::class.qualifiedName
+            TypeNullability.FLEXIBLE -> "FlexibleNullability"
```
yole: I don't think it makes any sense to report such annotations; this name isn't defined anywhere.

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -139,6 +197,48 @@ open class KotlinUParameter(
...
+            annotatedElement.typeReference?.getType()?.let { return it }
+        }
+        if (annotatedElement is KtProperty) {
+            annotatedElement.initializer?.let { it.getType(it.analyze()) }?.let { return it }
+            annotatedElement.delegateExpression?.let { it.getType(it.analyze())?.arguments?.firstOrNull()?.type }?.let { return it }
+        }
+        annotatedElement.getParentOfType<KtProperty>(false)?.let {
+            it.typeReference?.getType() ?: it.initializer?.let { it.getType(it.analyze()) }
+        }?.let { return it }
+        return null
```
yole: This will always return null for expression body methods with inferred return type.

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -139,6 +197,48 @@ open class KotlinUParameter(
...
+            TypeNullability.NULLABLE -> Nullable::class.qualifiedName
+            TypeNullability.FLEXIBLE -> "FlexibleNullability"
+            null -> null
+        }
+
+    override fun findAttributeValue(name: String?): UExpression? = null
+
+    override fun findDeclaredAttributeValue(name: String?): UExpression? = null
+
+    override fun resolve(): PsiClass? = null
```
yole: This can be implemented

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -139,6 +197,48 @@ open class KotlinUParameter(
...
+    val nullability by lz { getTargetType(annotatedElement)?.nullability() }
+
+    override val attributeValues: List<UNamedExpression>
+        get() = emptyList()
+    override val psi: PsiElement?
+        get() = null
+    override val javaPsi: PsiElement?
+        get() = null
+    override val sourcePsi: PsiElement?
+        get() = null
```
yole: Wouldn't it be better to report the type reference (if any) as the source PSI, so that inspection reports could be attached to some element?

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
yole: Also please don't forget about the `@receiver` target; it also maps to parameters.

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
```diff
@@ -149,6 +249,10 @@ open class KotlinUField(
 
     override val psi = javaPsi
 
+    override fun acceptsAnnotationTarget(target: AnnotationUseSiteTarget?): Boolean =
+            target == AnnotationUseSiteTarget.FIELD ||
+            (sourcePsi is KtProperty) && (target == null || target == AnnotationUseSiteTarget.PROPERTY)
```
yole: The `@delegate:` target also maps to fields.

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
xiexed: But can `@param:Foo` be used anywhere else except primary constructor? for `@receiver` fixed

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
xiexed: But we don't annotate methods nullability... but probably we should of course

plugins/uast-kotlin/src/org/jetbrains/uast/kotlin/declarations/KotlinUVariable.kt
yole: Yes, you can specify the use site target anywhere even if it's redundant.

xiexed: > '@param:' annotations could be applied only to primary constructor parameters

Idea says to me

xiexed:
Merged manually

PR #1421, opened 2017-12-05
Recognize Checker Framework *declaration* annotations

cpovirk:
We are migrating Guava to use these annotations rather than jsr305's `@Nullable`.
We can't use the Checker Framework's _`@Nullable`_ yet because we promise compatibility with Java 7, which doesn't support type annotations.
This is related to but distinct from https://youtrack.jetbrains.com/issue/KT-21408, which is about a different jsr305 annotation we use, `@ParametersAreNonnullByDefault`.

I've also updated some docs to mention Kotlin's existing support for the Checker Framework _`@NonNull`_.

NOTE: This commit does not yet run the new tests (though I've found a hacky way to confirm that they will indeed pass). I am unsure how to update files like `ForeignAnnotationsTestGenerated.java` that say they are generated with `org.jetbrains.kotlin.generators.tests.TestsPackageorg.jetbrains.kotlin.generators.tests.TestsPackage` (and so `testAllFilesPresentInTests` is failing for them). Please advise me how to update those files. Thanks.

### DIFF
--- a/compiler/testData/foreignAnnotations/tests/checkerFramework.kt
+++ b/compiler/testData/foreignAnnotations/tests/checkerFramework.kt
@@ -1,0 +1,33 @@
+// !DIAGNOSTICS: -UNUSED_VARIABLE -UNUSED_PARAMETER
+// FILE: A.java
+
+import org.checkerframework.checker.nullness.compatqual.*;
+
+public class A {
+    @NullableDecl public String field = null;
+
+    @NullableDecl
+    public String foo(@NonNullDecl String x, @NullableDecl CharSequence y) {
+        return "";
+    }
+
+    @NonNullDecl
+    public String bar() {
+        return "";
+    }
+
+}
+
+// FILE: main.kt
+
+fun main(a: A) {
+    a.foo("", null)?.length
+    a.foo("", null)<!UNSAFE_CALL!>.<!>length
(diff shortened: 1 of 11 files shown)


udalov:
To update `*Generated` tests, you can run `./gradlew generateTests`

cpovirk:
Thanks! I've updated the generated tests.

cpovirk:
We have a version of Guava coming out that uses the Checker Framework declaration annotations, so we're eager to see this go in. Let me know if there's anything else I can do to help things along.

udalov:
Could you please [report a separate issue](http://kotl.in/issue) describing your use case, so that we could track it properly, it'd end up in the changelog in the correct place, etc. Thanks!

cpovirk:
Thanks. I've filed [KT-21982](https://youtrack.jetbrains.com/issue/KT-21982).

udalov:
Merged to [master](https://github.com/JetBrains/kotlin/commit/ac87ad422d0e9a9d4dbc952ee33fe471bd5c1d3a) and [1.2.20](https://github.com/JetBrains/kotlin/commit/d202cf53147d4dbe0f0af9236040dc44daa98bcc). Thank you for the contribution!

cpovirk:
And thank you!

PR #1422, opened 2017-12-06
KT-21612 The "Remove redundant getter" inspection removes the type specifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21612


mglukhikh:
Merged, thank you.

PR #1423, opened 2017-12-07
KT-21560 Inspection to sort modifiers

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21560


yole:

PR #1426, opened 2017-12-12
KT-20357 - Added samples for property delegates.

AdamMc331:
Adding some standard lib samples to go with https://youtrack.jetbrains.com/issue/KT-20357.

### DIFF
--- a/libraries/stdlib/samples/test/samples/properties/delegates.kt
+++ b/libraries/stdlib/samples/test/samples/properties/delegates.kt
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-package samples.delegates
+package samples.properties
 
+import kotlin.properties.Delegates
 import samples.*
 import kotlin.test.*
 
@@ -38,7 +39,7 @@ class Delegates {
     @Sample
     fun throwVetoableDelegate() {
         var max: Int by Delegates.vetoable(0) { property, oldValue, newValue ->
-            return if (newValue > oldValue) true else throw(Throwable("New value must be larger than old value."))
+            if (newValue > oldValue) true else throw(Throwable("New value must be larger than old value."))
         }
 
         assertPrints(max, "0")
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/properties/delegates.kt
```diff
@@ -0,0 +1,67 @@
...
+package samples.delegates
+
+import samples.*
+import kotlin.test.*
+
+class Delegates {
+    @Sample
+    fun vetoableDelegate() {
+        var max: Int by Delegates.vetoable(0) { property, oldValue, newValue ->
+            newValue > oldValue
```
ilya-g: Another common use case of `vetoable` is to throw an exception before the new value is assigned if it is invalid. It could be illustrated in a separate sample.

libraries/stdlib/samples/test/samples/properties/delegates.kt
```diff
@@ -0,0 +1,67 @@
...
+        max = 10
+        assertPrints(10, max)
+    }
+
+    @Sample
+    fun notNullAccessError() {
+        var max: Int by Delegates.notNull()
+
+        assertFailsWith<IllegalStateException> { println(max) }
+    }
```
ilya-g: `nonNullDelegate` and `notNullAccessError` can be merged in one sample.

libraries/stdlib/samples/test/samples/properties/delegates.kt
AdamMc331: Added this in throwVetoableDelegate() sample.

libraries/stdlib/samples/test/samples/properties/delegates.kt
```diff
@@ -0,0 +1,77 @@
...
+        assertPrints(max, "10")
+
+        max = 5
+        assertPrints(max, "10")
+    }
+
+    @Sample
+    fun throwVetoableDelegate() {
+        var max: Int by Delegates.vetoable(0) { property, oldValue, newValue ->
+            if (newValue > oldValue) true else throw(Throwable("New value must be larger than old value."))
```
ilya-g: Usually `IllegalArgumentException` is used in this case. Also there are redundant parentheses here.

ilya-g:
Thanks, I've squashed and merged it with minor edits (`assertTrue`/`assertFalse` instead of `assertPrints` for booleans)

PR #1430, opened 2017-12-12
Remove unnecessary parameter null checks in JvmStatic delegate methods.

denisvnukov:
Remove unnecessary non-null parameter checks inside static delegate methods
created for @JvmStatic companion object methods. Allows function generation
strategy decide if such checks need to be injected.

 #KT-7188 Fixed


dnpetrov:
Rebased & pushed manually. Thanks!

PR #1431, opened 2017-12-13
KT-21726 "arrayOf can be replaced with literal" inspection quick fix produces incompilable result in presence of spread operator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21726

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceArrayOfWithLiteralInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceArrayOfWithLiteralInspection.kt
@@ -44,7 +44,10 @@ class ReplaceArrayOfWithLiteralInspection : AbstractKotlinInspection() {
 
                 val parent = expression.parent
                 when (parent) {
-                    is KtValueArgument -> parent.parent.parent as? KtAnnotationEntry ?: return
+                    is KtValueArgument -> {
+                        if (parent.parent.parent !is KtAnnotationEntry) return
+                        if (parent.getSpreadElement() != null && !parent.isNamed()) return
+                    }
                     is KtParameter -> {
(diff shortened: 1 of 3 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceArrayOfWithLiteralInspection.kt
```diff
@@ -44,7 +44,10 @@ class ReplaceArrayOfWithLiteralInspection : AbstractKotlinInspection() {
 
                 val parent = expression.parent
                 when (parent) {
-                    is KtValueArgument -> parent.parent.parent as? KtAnnotationEntry ?: return
+                    is KtValueArgument -> {
+                        if (parent.parent.parent !is KtAnnotationEntry) return
+                        if (parent.getSpreadElement() != null && !parent.isNamed()) return
```
mglukhikh: Why so? Quick-fix can still be applied, but `*` should be preserved!

idea/testData/inspectionsLocal/replaceArrayOfWithLiteral/unnamedVararg.kt
```diff
@@ -0,0 +1,7 @@
+// LANGUAGE_VERSION: 1.2
+// PROBLEM: none
+
+annotation class Some(vararg val strings: String)
+
+@Some(*<caret>arrayOf("alpha", "beta", "omega"))
```
mglukhikh: This can be converted into `@Some(*["alpha", "beta", "omega"])`

mglukhikh:
After some thinking I decided there is no much reason to suggest "convert to array literal" in this case, because `*` is anyway redundant and after application of its fix `arrayOf` or `[]` will be anyway removed. So I will merge it in the current state, thank you.

PR #1432, opened 2017-12-13
KT-21727 "Redundant spread operator" inspection does not report array literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21727


mglukhikh:
Should be Ok, will merge it

mglukhikh:
Merged, thank you

PR #1434, opened 2017-12-15
KT-21770 Pasting $this into an interpolated string shouldn't escape $

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21770


yole:

PR #1438, opened 2017-12-19
KT-21741 Inspection to detect is checks for object types

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21741

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantObjectTypeCheckInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantObjectTypeCheckInspection.kt
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2010-2017 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantObjectTypeCheckInspection.kt
```diff
@@ -0,0 +1,97 @@
...
+    override fun getName() = "Remove 'is' operator"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val element = descriptor.psiElement as? KtWhenConditionIsPattern ?: return
+        val typeReference = element.typeReference ?: return
+        val factory = KtPsiFactory(project)
+        val newElement = factory.createExpression(typeReference.text)
+        element.replace(newElement)
```
mglukhikh: Here you break PSI structure: KtWhenCondition must be here, not a KtExpression

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantObjectTypeCheckInspection.kt
```diff
@@ -0,0 +1,97 @@
...
+class RedundantObjectTypeCheckInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return object : KtVisitorVoid() {
+
+            override fun visitIsExpression(expression: KtIsExpression) {
+                super.visitIsExpression(expression)
+                val typeReference = expression.typeReference ?: return
+                if (!typeReference.isObject()) return
+                registerProblem(expression, ReplaceWithEqualityFix(expression.isNegated))
```
mglukhikh: Here and below, I think better report problem just on `is`, not on a whole expression

mglukhikh:
Fixed and merged. Thank you.

PR #1439, opened 2017-12-20
KT-21698 `Create interface` shouldn't suggest to declare it inside a class which implements it

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21698


yole:

PR #1440, opened 2017-12-21
Added samples to Strings.kt to showcase stdlib string functions

Samaritan1011001:


Samaritan1011001:
Any update on this pull request?

.idea/runConfigurations/Generate_Compiler_Tests.xml
```diff
@@ -18,7 +18,6 @@
       </option>
       <option name="vmOptions" value="" />
     </ExternalSystemSettings>
-    <extension name="coverage" enabled="false" merge="false" sample_coverage="true" runner="idea" />
```
ilya-g: This change is not related to samples, better to revert

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,196 @@ class Strings {
         assertPrints(byteArray.toString(charset), "Hello")
     }
 
+
+    //Indices is IntRange
+    @Sample
+    fun slice() {
+        assertPrints("abcdef".slice(IntRange(1,2)),"bc")
+        assertPrints("abcdef".slice(IntRange(0,0)),"a")
```
ilya-g: Usually `IntRange` is constructed with the `..` operator: `1..2`

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,196 @@ class Strings {
...
+    }
+
+    //Without the startIndex
+    @Sample
+    fun startsWith(){
+        assertPrints("abcd".startsWith("ab",true),"true")
+        assertPrints("abcd".startsWith("Ab",true),"true")
+        assertPrints("Abcd".startsWith("Ab",false),"true")
+        assertPrints("abcd".startsWith("ab",false),"true")
```
ilya-g: Better to use named parameter to pass `ignoreCase` for clarity

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,196 @@ class Strings {
...
+    }
+
+    //delimiter: Char
+
+    @Sample
+    fun substringBefore(){
+        assertPrints("abcd".substringBefore("c"),"ab")
+        assertPrints("abcd".substringBefore("8"),"abcd")
+        assertPrints("kotlinot".substringBefore("ot"),"k")
+        assertPrints("kotlin".substringBefore('k'),"")
```
ilya-g: I think it'd be ok to merge first and third sample. 
Better to use some real use case to illustrate these functions, for example to take a part of a dot-delimitied identifier, or a slash-delimited path.

ilya-g: Could you also show the usage of these functions with `missingDelimiterValue` parameter specified?

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,196 @@ class Strings {
...
+    fun substringAfterLast(){
+        assertPrints("abcd".substringAfterLast("c"),"d")
+        assertPrints("abcd".substringAfterLast("8"),"abcd")
+        assertPrints("kotlinot".substringAfterLast("o"),"t")
+        assertPrints("kotlin".substringAfterLast('o'),"tlin")
+    }
+
+    @Sample
+    fun toBoolean(){
+        assertPrints("true".toBoolean(),"true")
```
ilya-g: Please use `assertTrue`/`assertFalse` assertions instead

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,196 @@ class Strings {
...
+        assertPrints("  a bcd  ".trimEnd(),"  a bcd")
+        assertPrints("  abcd  ".trimEnd(),"  abcd")
+
+    }
+
+    @Sample
+    fun trimStartString(){
+
+        //with no arguments
+        assertPrints("  a bcd  ".trimStart(),"a bcd  ")
```
ilya-g: These trailing spaces will be invisible when printed.

ilya-g:
@Samaritan1011001 sorry, I've made some review comments, but forgot to submit the review, so they all were left pending.

libraries/stdlib/samples/test/samples/text/strings.kt
ilya-g: Also in `1..2` range, the end index 2 can be confused with the range length which also happens to be 2.

Samaritan1011001:
@ilya-g I have made the changes asked for. Please do check it. Also sorry for the late reply.

ilya-g:
Good, but some problems are still unaddressed.

-  Please reformat samples according to the [Kotlin style guide](https://kotlinlang.org/docs/reference/coding-conventions.html) (Ctrl+Alt+L could help to do that)
- String-to-number conversion samples should be extracted to a separate class with samples.
- The second argument of `assertPrints` should be a literal string
- function and local variable names should be camel-cased
- trailing spaces in `trim*` samples will be invisible when printed

Samaritan1011001:
Done.

Samaritan1011001:
@ilya-g Do I need to make other changes?

Samaritan1011001:
Are we done here? @ilya-g

PR #1441, opened 2017-12-21
KT-15180 Incorrect quickfix 'Specify type explicitly'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15180


mglukhikh:
Merged, thank you

PR #1442, opened 2017-12-21
README: IntelliJ IDEA 2017.3 is not early access anymore

educhastenier:
update README.md with direct download link to IntelliJ IDEA latest version


yole:
Thanks!

PR #1444, opened 2017-12-22
KT-21929 Inappropriate quick fix for a sealed class instantiation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21929


yole:

PR #1445, opened 2017-12-23
KT-18396 Bad quickfix for wrong nested classes in inner class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18396


mglukhikh:
Thank you, merged

PR #1446, opened 2017-12-23
KT-21214: Fix lambda function selection

joscha-alisch:
Fixes https://youtrack.jetbrains.com/issue/KT-21214

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
@@ -21,14 +21,13 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.psi.KtCallExpression
-import org.jetbrains.kotlin.psi.KtLambdaArgument
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 
 class KotlinCallExpressionWithLambdaSelectioner : ExtendWordSelectionHandlerBase() {
 
     override fun canSelect(e: PsiElement): Boolean
-            = e is KtCallExpression && e.lastArgumentIsLambda() && e.lambdaIsOutsideBrackets()
+            = e is KtCallExpression && e.hasLambda()
 
     override fun select(e: PsiElement?, editorText: CharSequence?, cursorOffset: Int, editor: Editor?): List<TextRange>? {
         if (e !is KtCallExpression) return null
@@ -37,9 +36,7 @@ class KotlinCallExpressionWithLambdaSelectioner : ExtendWordSelectionHandlerBase
         return listOf(TextRange(e.startOffset, endOffset))
     }
 
-    private fun KtCallExpression.lastArgumentIsLambda(): Boolean
-            = valueArguments.isNotEmpty() && valueArguments.last() is KtLambdaArgument
+    private fun KtCallExpression.hasLambda(): Boolean
+            = lambdaArguments.isNotEmpty()
 
-    private fun KtCallExpression.lambdaIsOutsideBrackets(): Boolean
-            = valueArgumentList?.children?.size != valueArguments.size
 }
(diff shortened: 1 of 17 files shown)


idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
```diff
@@ -0,0 +1,45 @@
...
+        if (e !is KtCallExpression) return null
+
+        val endOffset = e.valueArgumentList?.endOffset ?: return null
+        return listOf(TextRange(e.startOffset, endOffset))
+    }
+
+    private fun KtCallExpression.lastArgumentIsLambda(): Boolean
+            = valueArguments.isNotEmpty() && valueArguments.last() is KtLambdaArgument
+
+    private fun KtCallExpression.lambdaIsOutsideBrackets(): Boolean
```
goodwinnk: This is probably unneeded. If something is parsed as `KtLambdaArgument` it won't be in `valueArgumentList`.

goodwinnk:
Hello, thank you for the contribution!

I would like to accept the change, but could you please merge two commits into one. It's proved to be useful when main changes and corresponding tests are in the same commit.

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
joscha-alisch: Are you sure about that? If I remember correctly from debugging this in the beginning, the behaviour is as follows:

If a lambda is inside the brackets, it's part of valueArgumentList:
` foo( { "my lambda" } )`

If it's outside, it's not in valueArgumentList:
```
foo {
    "my lambda"
}
```

Might be wrong about this though, can't check it right now.

joscha-alisch:
Thanks for reviewing and happy holidays @goodwinnk, I squashed both commits into one :)

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
goodwinnk: You're right about including/excluding logic for `valueArgumentList`, but it's still enough to check that `KtLambdaArgument` is present.

`KtValueArgumentList` contains only nodes with type `VALUE_ARGUMENT`, and if there's a `KtLambdaArgument` psi element, it means that there's `LAMBDA_AGRUMENT`, and lists sizes would differ. 

Please take a look to `KtCallExpression.getValueArguments()` implementation. 

`https://github.com/JetBrains/kotlin/blob/78e278ec4c2a6c012897910d3829eb0ddead65ab/compiler/frontend/src/org/jetbrains/kotlin/psi/KtCallExpression.java#L78`

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override fun select(e: PsiElement?, editorText: CharSequence?, cursorOffset: Int, editor: Editor?): List<TextRange>? {
+        if (e !is KtCallExpression) return null
+
+        val endOffset = e.valueArgumentList?.endOffset ?: return null
+        return listOf(TextRange(e.startOffset, endOffset))
+    }
+
+    private fun KtCallExpression.lastArgumentIsLambda(): Boolean
+            = valueArguments.isNotEmpty() && valueArguments.last() is KtLambdaArgument
```
goodwinnk: I have understood that it can be implemented as `lambdaArguments().isNotEmpty()` after `lambdaIsOutsideBrackets()` discussion.

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
joscha-alisch: Ah, got it.. Makes perfect sense, thanks :)

idea/src/org/jetbrains/kotlin/idea/editor/wordSelection/KotlinCallExpressionWithLambdaSelectioner.kt
joscha-alisch: Nice, makes the whole thing way cleaner :)

goodwinnk:
I have merged commits once again and added a directive for YouTrack to commit message so it could link the issue with the commit.

The final commit is now in master: 809cc220edd5316bee37880584041d41461391b4.

Thank you!

PR #1447, opened 2017-12-24
KT-12814 Specify type explicitly produces erroneous code when platform type overrides not-null type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12814


idea/testData/intentions/specifyTypeExplicitly/classNameClashing.kt
```diff
@@ -1,4 +1,4 @@
-// WITH_RUNTIME
+// RUNTIME_WITH_FULL_JDK
```
mglukhikh: Don't understand why it's needed

mglukhikh:
Merged with minor fixes. Thank you.

PR #1448, opened 2017-12-24
KT-11208 Remove stream buffering

Meztihn:
[KT-11208](https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FKT-11208)

### DIFF
--- a/libraries/stdlib/src/kotlin/io/Console.kt
+++ b/libraries/stdlib/src/kotlin/io/Console.kt
@@ -139,22 +139,14 @@ public inline fun println() {
 private const val BUFFER_SIZE: Int = 32
 private const val LINE_SEPARATOR_MAX_LENGTH: Int = 2
 
-private var cachedDecoder: CharsetDecoder? = null
-
-private fun decoderFor(charset: Charset): CharsetDecoder {
-    return cachedDecoder?.takeIf { cached ->
-        cached.charset() == charset
-    } ?: charset.newDecoder().also { newDecoder ->
-        cachedDecoder = newDecoder
-    }
-}
+private val decoder: CharsetDecoder by lazy { Charset.defaultCharset().newDecoder() }
 
 /**
  * Reads a line of input from the standard input stream.
  *
  * @return the line read or `null` if the input stream is redirected to a file and the end of file has been reached.
  */
-fun readLine(charset: Charset = Charset.defaultCharset()): String? = readLine(System.`in`, decoderFor(charset))
+fun readLine(): String? = readLine(System.`in`, decoder)
 
 internal fun readLine(inputStream: InputStream, decoder: CharsetDecoder): String? {
     require(decoder.maxCharsPerByte() <= 1) { "Encodings with multiple chars per byte are not supported" }
@@ -165,7 +157,7 @@ internal fun readLine(inputStream: InputStream, decoder: CharsetDecoder): String
 
     var read = inputStream.read()
     if (read == -1) return null
-    while (read != -1) {
+    do {
         byteBuffer.put(read.toByte())
         if (decoder.tryDecode(byteBuffer, charBuffer, false)) {
             if (charBuffer.containsLineSeparator()) {
@@ -176,7 +168,7 @@ internal fun readLine(inputStream: InputStream, decoder: CharsetDecoder): String
             }
         }
         read = inputStream.read()
-    }
+    } while (read != -1)
 
     with(decoder) {
         tryDecode(byteBuffer, charBuffer, true) // throws exception if undecoded bytes are left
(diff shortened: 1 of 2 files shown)


libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,89 @@ public inline fun println() {
...
-    public override fun read(): Int {
-        return System.`in`.read()
-    }
+private const val BUFFER_SIZE: Int = 32
 
-    public override fun reset() {
-        System.`in`.reset()
-    }
+// Since System.in can change its value on the course of program running, we should always delegate to current value.
+private val stdin: InputStream
```
ilya-g: This property seems no longer required, `System.in` can be used directly instead.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,89 @@ public inline fun println() {
...
+}
 
-    public override fun mark(readlimit: Int) {
-        System.`in`.mark(readlimit)
+/**
+ * Reads a line of input from the standard input stream.
+ *
+ * @return the line read or `null` if the input stream is redirected to a file and the end of file has been reached.
+ */
+fun readLine(lineSeparator: String = System.lineSeparator(), charset: Charset = Charset.defaultCharset()): String? =
```
ilya-g: It should work with both types of line separators automatically, and there's no requirement to work with custom ones, so the `lineSeparator` parameter is not needed.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,91 @@ public inline fun println() {
...
+internal fun readLine(inputStream: InputStream, decoder: CharsetDecoder): String? {
+    require(decoder.maxCharsPerByte() <= 1) { "Encodings with multiple chars per byte are not supported" }
+
+    val byteBuffer = ByteBuffer.allocate(BUFFER_SIZE)
+    val charBuffer = CharBuffer.allocate(LINE_SEPARATOR_MAX_LENGTH)
+    val stringBuilder = StringBuilder()
+
+    var read = inputStream.read()
+    if (read == -1) return null
+    while (read != -1) {
```
ilya-g: Always true, this check can be moved to the end of loop, just replace `while () { }` with `do { } while ()`

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,91 @@ public inline fun println() {
...
+}
 
-    public override fun close() {
-        System.`in`.close()
+/**
+ * Reads a line of input from the standard input stream.
+ *
+ * @return the line read or `null` if the input stream is redirected to a file and the end of file has been reached.
+ */
+fun readLine(charset: Charset = Charset.defaultCharset()): String? = readLine(System.`in`, decoderFor(charset))
```
ilya-g: Due to compatibility reasons we cannot introduce an overload in a patch release. So if you don't mind, I'd make it a function without parameters for now.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,83 @@ public inline fun println() {
...
-    }
+private fun CharBuffer.containsLineSeparator(): Boolean {
+    return get(1) == '\n' || get(0) == '\n'
+}
 
-    public override fun read(b: ByteArray, off: Int, len: Int): Int {
-        return System.`in`.read(b, off, len)
-    }
-}))}
+private fun Buffer.flipBack(): Buffer = apply {
```
ilya-g: The return value is not used, so it can be just `flipBack() { ... }`

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,83 @@ public inline fun println() {
...
+        reset()
     }
 
-    public override fun close() {
-        System.`in`.close()
+    with(charBuffer) {
+        val length = position()
+        val first = get(0)
+        val second = get(1)
+        flip()
```
ilya-g: The buffer is not used after, so this `flip` doesn't affect anything.

libraries/stdlib/src/kotlin/io/Console.kt
```diff
@@ -132,50 +136,83 @@ public inline fun println() {
...
+    val positionBefore = charBuffer.position()
+    byteBuffer.flip()
+    with(decode(byteBuffer, charBuffer, isEndOfStream)) {
+        if (isError) throwException()
     }
-
-    public override fun available(): Int {
-        return System.`in`.available()
+    return (charBuffer.position() > positionBefore).also { isDecoded ->
+        byteBuffer.apply { if (isDecoded) clear() else flipBack() }
```
ilya-g: minor: `apply` here doesn't add clarity.

ilya-g:
Ok, I've compared the results with the reference implementation and added more test cases, all seems fine.
I'll make all remaining minor edits by myself and merge it.

ilya-g:
I've finally merged this PR.

@Meztihn, Thanks for your effort and patience.

PR #1449, opened 2017-12-25
KT-20620 Replace operator with function call breaks code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20620


mglukhikh:
Looks good, but how does it behave in a situation like

```
fun foo(x: String?, y: String(?)) {
    if (x != y) { ... }
}
```

Will check and fix if necessary.

mglukhikh:
Merged, thank you

PR #1450, opened 2017-12-25
Provide samples for ranges (KT-20357)

tommykw:


tommykw:
Thank you for the review, please take a look.

ilya-g:
Thank you, I've merged the samples and added some operations on these constructed ranges there.

PR #1451, opened 2017-12-26
KT-20357 Add sample codes for coerceAtLeast, coerceAtMost, coerceIn

shiraji:
https://youtrack.jetbrains.com/issue/KT-20357#comment=27-2650771

### DIFF
--- a/libraries/stdlib/src/generated/_Ranges.kt
+++ b/libraries/stdlib/src/generated/_Ranges.kt
@@ -899,6 +899,7 @@ public fun <T: Comparable<T>> T.coerceIn(range: ClosedFloatingPointRange<T>): T 
  * Ensures that this value lies in the specified [range].
  * 
  * @return this value if it's in the [range], or `range.start` if this value is less than `range.start`, or `range.endInclusive` if this value is greater than `range.endInclusive`.
+ * @sample samples.comparisons.ComparableOps.coerceIn
  */
 public fun <T: Comparable<T>> T.coerceIn(range: ClosedRange<T>): T {
     if (range is ClosedFloatingPointRange) {
@@ -916,6 +917,7 @@ public fun <T: Comparable<T>> T.coerceIn(range: ClosedRange<T>): T {
  * Ensures that this value lies in the specified [range].
  * 
  * @return this value if it's in the [range], or `range.start` if this value is less than `range.start`, or `range.endInclusive` if this value is greater than `range.endInclusive`.
+ * @sample samples.comparisons.ComparableOps.coerceIn
  */
 public fun Int.coerceIn(range: ClosedRange<Int>): Int {
     if (range is ClosedFloatingPointRange) {
@@ -933,6 +935,7 @@ public fun Int.coerceIn(range: ClosedRange<Int>): Int {
  * Ensures that this value lies in the specified [range].
  * 
  * @return this value if it's in the [range], or `range.start` if this value is less than `range.start`, or `range.endInclusive` if this value is greater than `range.endInclusive`.
+ * @sample samples.comparisons.ComparableOps.coerceIn
  */
 public fun Long.coerceIn(range: ClosedRange<Long>): Long {
     if (range is ClosedFloatingPointRange) {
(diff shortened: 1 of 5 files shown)


libraries/stdlib/samples/test/samples/number/int.kt
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package samples.number
+
+import samples.*
+
+class Int {
```
ilya-g: I don't expect there will be distinct samples for each primitive type. Also naming this class the same as the built-in `Int` looks dangerous.

So if you're going to improve further the samples for `coerce*` functions, this can be named as `Coercion`. 
Also it's worth to find a more general package name, since not only numbers could be coerced in range, but also any Comparable type could.

libraries/stdlib/src/generated/_Ranges.kt
```diff
@@ -677,6 +677,7 @@ public fun Short.coerceAtLeast(minimumValue: Short): Short {
 
 /**
  * Ensures that this value is not less than the specified [minimumValue].
+ * @sample samples.number.Int.coerceAtLeast
```
ilya-g: These functions are generated from a template. I think it'd be ok to provide the single sample for all overloads, so this sample reference can be placed in that template. You can find this template here: https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-stdlib-gen/src/templates/Comparables.kt#L27

libraries/stdlib/samples/test/samples/number/int.kt
shiraji: > I don't expect there will be distinct samples for each primitive type

OK, then for the class name, how about `ComparableOps`? In that case we could add samples that are not `coercion` (e.g. `minOf`, `maxOf`)

I'm not sure for the package name. Could you add the sample code under `sample` package? I thought I need to have at least one package.

libraries/stdlib/samples/test/samples/number/int.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    @Sample
+    fun coerceAtLeast() {
+        assertPrints(10.coerceAtLeast(5), "10")
+        assertPrints(10.coerceAtLeast(20), "20")
```
shiraji: Do you want to add more sample code for other primitives? @ilya-g

libraries/stdlib/samples/test/samples/number/int.kt
ilya-g: `samples.comparisons.ComparableOps` is fine too.

libraries/stdlib/samples/test/samples/number/int.kt
ilya-g: I don't think it would significantly improve the sample. On the other hand we could add a separate sample for some comparable types other than numbers.

shiraji:
@ilya-g Updated it.

ilya-g:
Have you tried to regenerate code from templates and see what happens? 
Adding sample for a template itself doesn't have any effect. You need to place the sample reference in the `doc { }` block.

The command to regenerate stdlib sources is here: https://github.com/JetBrains/kotlin/tree/master/libraries/tools/kotlin-stdlib-gen

libraries/stdlib/samples/test/samples/number/int.kt
ilya-g: Have you succeeded with this? I mean a sample for some comparable type.

ilya-g:
Do you have plans for other `coerce*` function samples?

shiraji:
Sorry about that. I didn't know how the template works. Let me fix that.

I'm not going to add sample in this pull request.

I thought this issue is for new kontributors. I have contributed kotlin a lot. So, if you really want me to add sample code, I would do that in other pull requests. But I really want new kontributors to implement other samples. What do you think?

ilya-g:
I think it would be just inconsistent only to provide `coerceAtLeast` with samples leaving other `coerce*` (at least `coerceAtMost`) functions without ones.

shiraji:
Sure. Then, let me add `coerceAtMost` and `coerceIn`

ilya-g:
Thanks, I've merged it manually and added missing samples for comparable types and `coerceIn` with  `ClosedFloatingPointRange` parameter.

PR #1452, opened 2017-12-28
KT-22038 Inspection to replace the usage of Java Collections methods on subtypes of MutableList with the methods from Kotlin stdlib

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22038


mglukhikh:
Simplified a bit and merged, thank you

PR #1453, opened 2017-12-29
KT-21780 Wrong redundant setter inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21780


yole:

PR #1454, opened 2017-12-31
KT-15320 Live templates: please add function which returns the "outer" class name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15320


yole:

PR #1455, opened 2018-01-01
KT-14670 Support kotlinPackageName() macro in live templates

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14670

### DIFF
--- a/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/macro/KotlinPackageNameMacro.kt
+++ b/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/macro/KotlinPackageNameMacro.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/macro/KotlinPackageNameMacro.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.intellij.psi.PsiDocumentManager
+import org.jetbrains.kotlin.psi.KtFile
+
+class KotlinPackageNameMacro : KotlinMacro() {
+    override fun getName() = "kotlinPackageName"
+    override fun getPresentableName() = "kotlinPackageName()"
+
+    override fun calculateResult(params: Array<Expression>, context: ExpressionContext): Result? {
+        val project = context.project
+        val document = context.editor?.document ?: return null
```
yole: Please use `context.psiElementAtStartOffset.containingFile as? KtFile` instead.

t-kameyama: Fixed. Please review.

yole:

PR #1457, opened 2018-01-04
KT-21974 Editor color scheme option for Kotlin typealias names

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21974


yole:

PR #1458, opened 2018-01-04
KT-21949 Please add a separate Color Scheme settings for properties synthesized from Java accessors

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21949


yole:

PR #1459, opened 2018-01-05
KT-21213 multiline kdoc - intellij joins lines together without space

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21213


idea/src/org/jetbrains/kotlin/idea/kdoc/KDocRenderer.kt
```diff
@@ -163,7 +164,10 @@ object KDocRenderer {
         // Avoid wrapping the entire converted contents in a <p> tag if it's just a single paragraph
         val maybeSingleParagraph = markdownNode.children.singleOrNull { it.type != MarkdownTokenTypes.EOL }
         return if (maybeSingleParagraph != null && !allowSingleParagraph) {
-            maybeSingleParagraph.children.joinToString("") { it.toHtml() }
+            val lineSeparator = LineSeparator.getSystemLineSeparator().separatorString
```
yole: Source code in IntelliJ always uses `\n` as a line separator. No need to access the system line separator here.

t-kameyama:
Fixed. Please review.

yole:
You can also remove the unused import.

t-kameyama:
Removed it.

PR #1460, opened 2018-01-05
KT-16955 KDoc: Links with alternative text are not highlighted

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16955


yole:
This is actually an incomplete fix: it fixes the highlighting but not other features (target validation, navigation etc.). It would be much better to fix parsing so that KDocTokens.MARKDOWN_INLINE_LINK is also parsed into a KDocLink, and to fix the KDocLink implementation so that it handles links with alternative text too.

t-kameyama:
@yole Fixed. Please review.

compiler/testData/psi/kdoc/HttpLink.txt
```diff
@@ -9,7 +9,14 @@ KtFile: HttpLink.kt
...
       PsiElement(KDOC_TEXT)(' ')
-      PsiElement(KDOC_MARKDOWN_INLINE_LINK)('[Kotlin](http://kotlinlang.org)')
+      KDOC_MARKDOWN_INLINE_LINK
+        PsiElement(LBRACKET)('[')
+        PsiElement(IDENTIFIER)('Kotlin')
+        PsiElement(RBRACKET)(']')
+        PsiElement(LPAR)('(')
+        PsiElement(IDENTIFIER)('http')
+        PsiElement(COLON)(':')
+        PsiComment(EOL_COMMENT)('//kotlinlang.org)')
```
yole: This is really unfortunate. We shouldn't be using the Kotlin lexer to scan the text between parentheses, which is not supposed to have any defined format at all. The text between LPAR and RPAR should be a single token. (Maybe the right way to fix this is not to pass `KotlinLanguage.INSTANCE` to the `MARKDOWN_INLINE_LINK` constructor, but to create a separate language with its own lexer that will produce Kotlin tokens.)

t-kameyama: `[http://kotlinlang.org]` is not parsed as `MARKDOWN_LINK`, but `[Kotlin](http://kotlinlang.org)` is parsed as `MARKDOWN_INLINE_LINK`. I think that `[Kotlin](http://kotlinlang.org)` does not need to be parsed as `MARKDOWN_INLINE_LINK` like `[http://kotlinlang.org]`.

semoro: `[http://kotlinlang.org]` is not a link according to Markdown [spec](https://spec.commonmark.org/0.23/#links)

But `[Kotlin](http://kotlinlang.org)` is actually an inline link

PR #1461, opened 2018-01-07
Add Sets.emptySet sample

akperkins:
As a part of the work needed for the [Provide samples for all standard library functions issue](https://youtrack.jetbrains.com/issue/KT-20357), I have added a sample for the `Sets.emptySet()` and the `Sets.setOf()` functions.


ilya-g:
Thank you for the sample.
I've merged it, but moved the Sets class into the samples.collections.Collections class.

PR #1462, opened 2018-01-07
Add support for closing and deleting raw strings.

Flamenco:


idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringBackspaceHandler.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.psi.SmartPsiElementPointer
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+import org.jetbrains.kotlin.psi.psiUtil.createSmartPointer
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+class KotlinRawStringBackspaceHandler : BackspaceHandlerDelegate() {
+    var ele: SmartPsiElementPointer<KtStringTemplateExpression>? = null
```
yole: You don't need a smart pointer here. If you need to keep track of a range in a document which is updated when the document changes, use `Document.createRangeMarker()`.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringBackspaceHandler.kt
```diff
@@ -0,0 +1,83 @@
...
+            return
+        }
+        if (file !is KtFile) {
+            return
+        }
+        if (editor != null) {
+            val offset = editor.caretModel.offset
+            val psiElement = file.findElementAt(offset) ?: return
+
+            val possibleString: KtStringTemplateExpression = getStringTemplate(psiElement) ?: return
```
yole: You can simply use `getParentOfType()`.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringBackspaceHandler.kt
```diff
@@ -0,0 +1,83 @@
...
+            if (possibleString.text == "\"\"\"\"\"\"") {
```
yole: This is inconsistent with how backspace works for regular string literals. With a normal string literal, if the caret is between the quotes, both quotes are deleted, and if it is after the quotes, only the trailing quote is deleted. Similarly, this handler should delete the quotes only if the caret is between the two triple quotes.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringBackspaceHandler.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+import org.jetbrains.kotlin.psi.psiUtil.createSmartPointer
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+class KotlinRawStringBackspaceHandler : BackspaceHandlerDelegate() {
+    var ele: SmartPsiElementPointer<KtStringTemplateExpression>? = null
+
+    override fun beforeCharDeleted(c: Char, file: PsiFile?, editor: Editor?) {
```
yole: You don't need the nullability marks here. The `file` and `editor` parameters in `BackspaceHandlerDelegate` and `TypedHandlerDelegate` methods are annotated as `@NotNull`.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringTypedHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.codeInsight.editorActions.TypedHandlerDelegate
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.fileTypes.FileType
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtStringTemplateExpression
+
+class KotlinRawStringTypedHandler : TypedHandlerDelegate() {
```
yole: Please add tests for the logic you're adding. See TypedHandlerTest and BackspaceHandlerTestGenerated.

yole:
Also please add a reference to https://youtrack.jetbrains.com/issue/KT-2582 to the commit description.

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringBackspaceHandler.kt
Flamenco: Nice to know!  At least I learned about smart pointers... 🥇

idea/src/org/jetbrains/kotlin/idea/editor/KotlinRawStringTypedHandler.kt
Flamenco: I did not see tests for the single quote to base this off of.  My project is not building so I can't run the tests on my end.  Specifically, should the delete test just return:

```
<cursor>
```

yole:
Looks good, but one issue is still not addressed: "This is inconsistent with how backspace works for regular string literals. With a normal string literal, if the caret is between the quotes, both quotes are deleted, and if it is after the quotes, only the trailing quote is deleted. Similarly, this handler should delete the quotes only if the caret is between the two triple quotes."

Flamenco:
> This is inconsistent with how backspace works

I thought that was fixed in the latest PR.  It works exactly like the single quote case.

yole:
I may miss something but I don't see any code in KotlinRawStringBackspaceHandler which checks the caret offset.

Flamenco:
When I refactored the code, I narrowed down the parentOfType logic.

yole:
The `parentOfType` method is missing from the pull request as it's currently submitted; the code doesn't compile.

Flamenco:
I'm not really sure how the import was removed, but I added it back in.

yole:
The issue is still not addressed, and it can't be addressed using `parentOfType()`, you actually need to check the caret offset. To make it clear what issue I'm talking about, I've recorded a screencast:
https://www.screencast.com/t/iFrnhG0fM

The triple quotes should be deleted only when the caret is in the first location shown (between the two triple quotes), not when it is moved backwards.

Also while testing this I noticed that insertion of paired triple quotes does not work at the very end of the file.

If you'd like to, I can fix the remaining issues myself.

Flamenco:
@yole I understand the issue now.  Please let me know if I need to do anything else.

Flamenco:
@yole Are we all set with this?  I thought https://github.com/JetBrains/kotlin/pull/1462/commits/4f9af87a28027672302c2ae7a6be98ce12865c33 addressed the remaining issues.

Flamenco:
@yole Any chance you can take a look at my last update?

Montesdodgers13:
I got

goodwinnk:
I'have squashed and merged your commit manually: https://github.com/JetBrains/kotlin/commit/8252b35ed5f5bb457b3c0744219bf2a5f5e27e47

There were several issues with the change, that I've decided to fix right away with https://github.com/JetBrains/kotlin/commit/af930b8364e99a84c1f51a5ae09427bde16974f9:
 - Backspace insertion tests didn't start. Please use "Generate All Tests" to generate test methods.
 - Closing triple-quote was added in this scenario: `val test = """"""<caret>`.
 - `testAutoCloseRawString` didn't work.

Thank you, I'm really glad this old issue is finally closed!

Flamenco:
@goodwinnk Thanks.  I look forward to checking out your modifications.  It ended up being more work than I estimated, but I learned much in the process!

PR #1464, opened 2018-01-09
KT-20439 Add intentions to add/remove labeled return to last expression in a lambda

shiraji:
https://youtrack.jetbrains.com/issue/KT-20439

### DIFF
--- a/idea/resources/intentionDescriptions/RemoveLabeledReturnInLambdaIntention/description.html
+++ b/idea/resources/intentionDescriptions/RemoveLabeledReturnInLambdaIntention/description.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This intention removes labeled return to last expression in a lambda.
+</body>
+</html>
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.intentions
+
+import com.intellij.codeInsight.intention.LowPriorityAction
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.*
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.types.typeUtil.isUnit
+
+class AddLabeledReturnInLambdaIntention :
+        SelfTargetingRangeIntention<KtLambdaExpression>(
+                KtLambdaExpression::class.java,
+                "Add labeled return to last expression in a lambda"
+        ),
+        LowPriorityAction {
+    override fun applicabilityRange(element: KtLambdaExpression): TextRange? {
+        if (!isApplicableTo(element)) return null
+        val labelName = createLabelName(element) ?: return null
+        text = "Add return@$labelName"
+        return element.bodyExpression?.statements?.last()?.textRange
+    }
+
+    override fun applyTo(element: KtLambdaExpression, editor: Editor?) {
+        if (!isApplicableTo(element)) return
+        val labelName = createLabelName(element) ?: return
+        val lastStatement = element.bodyExpression?.statements?.last() ?: return
+        val newExpression = KtPsiFactory(element.project).createExpression("return@$labelName ${lastStatement.text}")
+        lastStatement.replace(newExpression)
+    }
+
+    private fun isApplicableTo(element: KtLambdaExpression): Boolean {
+        val block = element.bodyExpression ?: return false
+        val lastStatement = block.statements.last()
+        return lastStatement !is KtReturnExpression && lastStatement.isUsedAsExpression(lastStatement.analyze())
+    }
+
+    private fun createLabelName(element: KtLambdaExpression): String? {
+        val block = element.bodyExpression ?: return null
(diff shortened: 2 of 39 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
```diff
@@ -0,0 +1,44 @@
...
+        val block = element.getStrictParentOfType<KtBlockExpression>() ?: return false
+        if (block.statements.lastOrNull() != element) return false
+        val callExpression = element.getStrictParentOfType<KtCallExpression>() ?: return false
+        val index = callExpression.valueArguments.indexOfFirst {
+            val lambda = it.getArgumentExpression() as? KtLambdaExpression
+            lambda?.bodyExpression == block
+        }
+        if (index < 0) return false
+        val resolvedCall = callExpression.getResolvedCall(callExpression.analyze()) ?: return false
+        if (resolvedCall.candidateDescriptor.valueParameters[index].type.arguments.last().type.isUnit()) return false
```
mglukhikh: Why this case with `Unit` result is important? Why can't we write something like
```
1.let {
    return@let Unit
}
```

mglukhikh: Anyway, test should be created for this case

idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
```diff
@@ -0,0 +1,44 @@
...
+        val labelName = callExpression.getCallNameExpression()?.text ?: return false
```
mglukhikh: What if the lambda is labeled? E.g.

```
    listOf(1, 2).find label@{
        true
    }
```

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveLabeledReturnInLambdaIntention.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+
+class RemoveLabeledReturnInLambdaIntention :
+        SelfTargetingIntention<KtReturnExpression>(KtReturnExpression::class.java, "Remove labeled return to last expression in a lambda"),
+        LowPriorityAction {
+    override fun isApplicableTo(element: KtReturnExpression, caretOffset: Int): Boolean {
+        val labelName = element.getLabelName() ?: return false
+        val callExpression = element.getStrictParentOfType<KtCallExpression>() ?: return false
+        val callName = callExpression.getCallNameExpression()?.text ?: return false
```
mglukhikh: Here the case with labeled lambda (see above) also should be considered.

idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.getCallNameExpression
+import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.types.typeUtil.isUnit
+
+class AddLabeledReturnInLambdaIntention :
+        SelfTargetingIntention<KtExpression>(KtExpression::class.java, "Add labeled return to last expression in a lambda"),
+        LowPriorityAction {
+    override fun isApplicableTo(element: KtExpression, caretOffset: Int): Boolean {
```
mglukhikh: Not sure it's good to call `isApplicableTo` on any `KtExpression`. I'd consider deriving from `SelfTargetingRangeIntention`  and using something like `KtLambdaExpression` as a base type.

idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
shiraji: Wow. I didn't know there is such syntax in Kotlin. Let me re-think about this.

idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
shiraji: I don't want to add `return@forEach` in the following case.

```
list(1,2).forEach {
    <caret>1
}
```

```
list(1,2).forEach {
    return@forEach 1 // this would be compile error
}
```

shiraji: @mglukhikh what should I do for these case? Do you want to add `return@forEach` at the end of lambda or ignore the case?

mglukhikh: @shiraji You can try `KtExpression.isUsedAsExpression` to control whether this `1` is really lambda result or not. In the second case, `isUsedAsExpression` should return false.

idea/src/org/jetbrains/kotlin/idea/intentions/AddLabeledReturnInLambdaIntention.kt
mglukhikh: Don't forget about this point please

shiraji: I'll ping you as soon as this PR is ready.

shiraji:
@mglukhikh Finally, it's ready now. Could you review this PR? Thanks

idea/resources/intentionDescriptions/RemoveLabeledReturnInLambdaIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention removes labeled return to last expression in a lambda.
```
mglukhikh: from last expression

mglukhikh:
Thank you, I will test it and fix the remaining things myself.

PR #1465, opened 2018-01-09
Fixes for generating overrides of suspend functions

cypressious:
Fixes #KT-21600

https://youtrack.jetbrains.com/issue/KT-21600

### DIFF
--- a/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
+++ b/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
@@ -221,6 +221,7 @@ interface DescriptorRendererOptions {
     var renderConstructorKeyword: Boolean
     var renderUnabbreviatedType: Boolean
     var includeAdditionalModifiers: Boolean
+    var includeSuspendModifier: Boolean
     var parameterNamesInFunctionalTypes: Boolean
     var renderFunctionContracts: Boolean
 }
--- a/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
+++ b/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
@@ -221,7 +221,6 @@ interface DescriptorRendererOptions {
     var renderConstructorKeyword: Boolean
     var renderUnabbreviatedType: Boolean
     var includeAdditionalModifiers: Boolean
-    var includeSuspendModifier: Boolean
     var parameterNamesInFunctionalTypes: Boolean
     var renderFunctionContracts: Boolean
 }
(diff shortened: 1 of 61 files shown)


udalov:
This will likely change some test data that uses .txt files with rendered Kotlin declarations. Please run all compiler tests with `./gradlew cleanTest :compiler:test`

cypressious:
@udalov thx for the heads-up. Any tips on how to bulk-fix it? Or is there only the manual way?

cypressious:
I fixed ~100 tests

cypressious:
I pigybacked the fix for https://youtrack.jetbrains.com/issue/KT-22200 on this PR as well.

core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRendererImpl.kt
```diff
@@ -630,6 +634,8 @@ internal class DescriptorRendererImpl(
 
                 if (includeAdditionalModifiers) {
                     renderAdditionalModifiers(function, builder)
+                } else if (includeSuspendModifier) {
```
udalov: Why else?

core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
```diff
@@ -221,6 +221,7 @@ interface DescriptorRendererOptions {
     var renderConstructorKeyword: Boolean
     var renderUnabbreviatedType: Boolean
     var includeAdditionalModifiers: Boolean
+    var includeSuspendModifier: Boolean
```
udalov: Are there any cases when this option should be false?

core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRendererImpl.kt
cypressious: `includeAdditionalModifiers` implies `includeSuspendModifier`. If I left out the `else`, `suspend` would be rendered twice.

core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRenderer.kt
cypressious: I'm not sure, to be honest. `OverridesCompletion.PRESENTATION_RENDERER` used to omit `suspend` but that's what the PR fixes. I could render `suspend` unconditionally if you think that's the correct behavior.

udalov: Yes, I think it'd be best to always render `suspend` because I don't see why it would be useful not to do that

cypressious: I removed the property so that `suspend` is always rendered when applicable.

udalov:
I've fixed the remaining tests, squashed everything into two commits, and merged to master. Thank you for the contribution!

PR #1466, opened 2018-01-11
KT-22171 Add Intention for single character substring

tommykw:
https://youtrack.jetbrains.com/issue/KT-22171


cypressious:
Shouldn't this be an inspection rather than an intention?

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSubstringWithIndexingOperationIntention.kt
```diff
@@ -0,0 +1,56 @@
...
+        if (element.isSubstringMethod()) {
+            return applicabilityRangeInner(element)
+        }
+
+        return null
+    }
+
+    override fun applicabilityRangeInner(element: KtDotQualifiedExpression): TextRange? {
+        val arguments = element.callExpression?.valueArguments ?: return null
+        if (arguments.size != 2 || !element.isFirstArgumentZero() || !element.isSecondArgumentOne()) return null
```
mglukhikh: You check for very special case here, it's not too much reason to add a new inspection for just `(0, 1)` case. I think `c.substring(25, 26)` can also be replaced with `c[25]`. Also, `c.substring(n, n+1)` can be replaced with `c[n]`.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSubstringWithIndexingOperationIntention.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    private fun KtDotQualifiedExpression.isSubstringMethod(): Boolean {
+        val resolvedCall = toResolvedCall(BodyResolveMode.PARTIAL) ?: return false
+        return (resolvedCall.resultingDescriptor.fqNameUnsafe.asString() == "kotlin.text.substring")
+    }
+
+    private fun KtDotQualifiedExpression.isSecondArgumentOne(): Boolean {
+        val bindingContext = analyze()
+        val resolvedCall = callExpression.getResolvedCall(bindingContext) ?: return false
+        val expression = resolvedCall.call.valueArguments[1].getArgumentExpression() as? KtConstantExpression ?: return false
```
mglukhikh: Try to rewrite this. First, it's too complex, and second, you are using `!!` which can result in an exception

mglukhikh:
About an idea to make it an inspection: it's possible but harder. We must remember that `substring(0, 1)` has a type of `String` and `[0]` has a type of `Char`, so changing one with another will sometimes provoke a compilation error. Such behaviour of intention is IMHO acceptable but for inspection it's absolutely not possible.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSubstringWithIndexingOperationIntention.kt
tommykw: Fixed.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSubstringWithIndexingOperationIntention.kt
tommykw: Fixed.

tommykw:
Sorry for the late, it seems difficult for me to convert it to inspection.

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceSubstringWithIndexingOperationIntention.kt
```diff
@@ -0,0 +1,55 @@
...
+        return (resolvedCall.resultingDescriptor.fqNameUnsafe.asString() == "kotlin.text.substring")
+    }
+
+    private fun KtDotQualifiedExpression.getValueArgument(index: Int): Int? {
+        val bindingContext = analyze()
+        val resolvedCall = callExpression.getResolvedCall(bindingContext) ?: return null
+        val expression = resolvedCall.call.valueArguments[index].getArgumentExpression() as? KtConstantExpression ?: return null
+        val constant = ConstantExpressionEvaluator.getConstant(expression, bindingContext) ?: return null
+        val constantType = bindingContext.getType(expression) ?: return null
+        return constant.getValue(constantType) as Int
```
mglukhikh: Much better to have safe cast (`as?`) here, otherwise you can get CCE in strange cases like `x.substring("alpha")`. Of course they are erroneous, but it's quite possible user write them during editing. Will fix myself

mglukhikh:
So thank you, merged it. Small fix added

PR #1467, opened 2018-01-11
Log standard error only on exit status error

cuihtlauac:
Do not store process standard error messages into execution
log if the process exit code does not indicate an error. Such
messages are warning or important information, not errors, they
should not trigger failures.

FIX KT-22241


udalov:
This would hide compilation warnings in tests that extend `KotlinIntegrationTestBase` in case there are no errors. Which means that we'll get no information if some warning is suddenly reported in some complex test, and the whole point of integration tests is to check that things like this don't happen. I'm afraid this is not something that we'd want.

cuihtlauac:
Thanks for this review. Don't you think there are warnings that should 
not trigger a failure (see for instance what I reported in KT-22241). I 
believe such warnings shall be either filtered out by runProcess() or 
ignored by check() ?

yole:
I'd rather fix the issue by adding code to remove the JAVA_TOOL_OPTIONS line in `KotlinIntegrationTestBase.normalizeOutput().`

@vak280974 Please refrain from reviewing pull requests on this project; your reviews do not add any information and only confuse contributors.

cuihtlauac:
Implemented as requested.

yole:

cuihtlauac:
Thanks for merging!

PR #1468, opened 2018-01-11
KT-22042

qwwdfsad:
Performance fix for KT-22042.
Performance data on android (emulated by `-XX:TieredStopAtLevel=1`) looks impressive,

Before
```
Benchmark                                      Mode  Cnt     Score    Error   Units
AnyOfBenchmark.indexOfAny                      avgt   20    36.251 ±  0.697   ns/op
AnyOfBenchmark.indexOfAny:·gc.alloc.rate       avgt   20   656.478 ± 12.139  MB/sec
AnyOfBenchmark.indexOfAny:·gc.alloc.rate.norm  avgt   20    24.944 ±  0.002    B/op
```

After
```
Benchmark                                      Mode  Cnt   Score    Error   Units
AnyOfBenchmark.indexOfAny                      avgt   20  20.041 ±  1.723   ns/op
AnyOfBenchmark.indexOfAny:·gc.alloc.rate       avgt   20  ≈ 10⁻³           MB/sec
AnyOfBenchmark.indexOfAny:·gc.alloc.rate.norm  avgt   20  ≈ 10⁻⁵             B/op
```

Benchmark 
```
    private val haystack = "haystack"
    private val needles = listOf("_", "!")

    @Benchmark
    fun indexOfAny(): Int {
        return haystack.lastIndexOfAny(needles, 0)
    }
```

I didn't touch `findAnyOf` which receives `Collection<String>` as parameter because it already looks complicated enough due to `instance of String` check.
One of the solutions may be to inline `findAnyOf` into `indexOfAny`/`lastIndexOfAny`, eliminate pair allocation and eliminate range allocation without switches because `last` will be inlined as well, but it will introduce additional code duplication. If you think it's okay I can fix it in this branch.

I've ran only tests for stdlib, which seems to fully cover changed methods, but I'm not sure whether it's enough

### DIFF
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -807,12 +807,25 @@ private fun CharSequence.findAnyOf(chars: CharArray, startIndex: Int, ignoreCase
         return if (index < 0) -1 else index
     }
 
-    val indices = if (!last) startIndex.coerceAtLeast(0)..lastIndex else startIndex.coerceAtMost(lastIndex) downTo 0
-    for (index in indices) {
-        val charAtIndex = get(index)
-        val matchingCharIndex = chars.indexOfFirst { it.equals(charAtIndex, ignoreCase) }
-        if (matchingCharIndex >= 0)
-            return index
+    // Split it to two loops to avoid implicit IntRange allocation
+    when (last) {
+        true -> {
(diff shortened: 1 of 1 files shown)


ilya-g:
Thanks for idea, I've merged an alternative way to fix that.

qwwdfsad:
Yup, this one looks better, thanks

qwwdfsad:
By the way, same thing can be done for indexOfAny/lastIndexOfAny which receives array of strings instead as well

PR #1469, opened 2018-01-11
Add parameter for the starting index for indexOf and lastIndexOf.

JakeWharton:
KT-7067

### DIFF
--- a/js/js.libraries/src/core/generated/_ArraysJs.kt
+++ b/js/js.libraries/src/core/generated/_ArraysJs.kt
@@ -1237,18 +1237,63 @@ public fun CharArray.getOrNull(index: Int): Char? {
     return if (index >= 0 && index <= lastIndex) get(index) else null
 }
 
+@Deprecated("Use indexOf(T, Int)", level = DeprecationLevel.HIDDEN)
+public fun <@kotlin.internal.OnlyInputTypes T> Array<out T>.indexOf(element: T): Int {
+    return indexOf(element, 0)
+}
+
+@Deprecated("Use indexOf(T, Int)", level = DeprecationLevel.HIDDEN)
+public fun ByteArray.indexOf(element: Byte): Int {
+    return indexOf(element, 0)
+}
+
+@Deprecated("Use indexOf(T, Int)", level = DeprecationLevel.HIDDEN)
+public fun ShortArray.indexOf(element: Short): Int {
+    return indexOf(element, 0)
+}
+
+@Deprecated("Use indexOf(T, Int)", level = DeprecationLevel.HIDDEN)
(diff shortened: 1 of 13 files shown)


js/js.libraries/src/core/generated/_ArraysJs.kt
```diff
@@ -1767,18 +1812,63 @@ public inline fun CharArray.last(predicate: (Char) -> Boolean): Char {
...
+@Deprecated("Use lastIndexOf(T, Int)", level = DeprecationLevel.HIDDEN)
+public fun CharArray.lastIndexOf(element: Char): Int {
+    return lastIndexOf(element, 0)
+}
+
 /**
  * Returns last index of [element], or -1 if the array does not contain element.
  */
-public fun <@kotlin.internal.OnlyInputTypes T> Array<out T>.lastIndexOf(element: T): Int {
+public fun <@kotlin.internal.OnlyInputTypes T> Array<out T>.lastIndexOf(element: T, startIndex: Int = 0): Int {
```
ilya-g: In string extensions `lastIndexOf` begins search from `startIndex` position (which is `lastIndex` by default) and goes down to zero.

libraries/tools/kotlin-stdlib-gen/src/templates/Elements.kt
```diff
@@ -89,21 +102,33 @@ object Elements : TemplateGroupBase() {
...
     }
 
     val f_lastIndexOf = fn("lastIndexOf(element: T)") {
         include(Iterables, Sequences, ArraysOfObjects, ArraysOfPrimitives, Lists)
+    } builder {
+        typeParam("@kotlin.internal.OnlyInputTypes T")
+        specialFor(Lists) {
+            annotation("""@Suppress("EXTENSION_SHADOWED_BY_MEMBER") // false warning, extension takes precedence in some cases""")
+        }
+        annotation("""@Deprecated("Use lastIndexOf(T, Int)", level = DeprecationLevel.HIDDEN)""")
```
ilya-g: Use `deprecate(forBinaryCompatibility)` instead.

JakeWharton:
Yeah sorry I've been meaning to get back to this.

ilya-g:
Questions to discuss: 
- should there be overloads with `startIndex` for `Iterable` and `Sequence`, given that they do not have efficient way to skip elements.
- should we leave overloads like `List<T>.indexOf(element: T)`, those that delegate search to the list itself, which may be more efficient.
- how to implement `List.indexOf(element, startIndex)` better:
  - using indexed access
  - using `listIterator(startIndex)`
  - using `subList().indexOf(element)`

PR #1470, opened 2018-01-12
KT-15176 Remove "Create type alias" intention when called on java class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15176


asedunov:
Merged! Thank you

PR #1471, opened 2018-01-12
KT-21881 Replace "If" with safe access intention false positive

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21881


mglukhikh:
Merged, thank you

PR #1474, opened 2018-01-14
Sample for setOf method

SwapnilSankla:
Sample for setOf method

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -271,6 +271,14 @@ class Collections {
             assertTrue(set == other, "Empty sets are equal")
             assertPrints(set, "[]")
         }
+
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -271,6 +271,14 @@ class Collections {
             assertTrue(set == other, "Empty sets are equal")
             assertPrints(set, "[]")
         }
+
+        @Sample
+        fun readOnlySet() {
+            val emptySet: Set<Int> = setOf()
```
ilya-g: `setOf` is already covered by the sample above. Take a look at `readOnlyList` and `singletonReadOnlyList` samples for an example.

ilya-g:
I've improve this sample a bit and merged it manually, thanks for the PR.

PR #1475, opened 2018-01-14
KT-16661

qwwdfsad:
Performance improvements of `Strings#split`.
For detailed evaluation see [KT-22248](https://youtrack.jetbrains.com/issue/KT-22248) (which is now marked as duplicate of KT-16661)

After 25c6cb0 Kotlin version became consistently faster than Java version for any arguments.

```
Benchmark                 (needle)   Mode  Cnt   Score   Error   Units
SplitBenchmark.splitJava         ,  thrpt   10  57.818 ± 1.740  ops/us
SplitBenchmark.splitJava         a  thrpt   10   6.687 ± 1.353  ops/us
SplitBenchmark.splitKt           ,  thrpt   10  77.459 ± 1.855  ops/us
SplitBenchmark.splitKt           a  thrpt   10  11.378 ± 0.299  ops/us

```

### DIFF
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1222,10 +1222,8 @@ private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: In
         if (!isLimited || result.size < limit - 1) {
             result.add(substring(currentOffset, nextIndex))
             currentOffset = nextIndex + delimiter.length
-            nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
-        } else {
-            result.add(substring(currentOffset, length))
-            return result
+            // Do not search for next occurrence if we're reaching limit
+            nextIndex = if (result.size == limit - 1) -1 else indexOf(delimiter, currentOffset, ignoreCase)
         }
     }
 
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1218,19 +1218,14 @@ private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: In
 
     val isLimited = limit > 0
     val result = ArrayList<String>(if (isLimited) limit.coerceAtMost(10) else 10)
-    while (nextIndex != -1) {
-        if (!isLimited || result.size < limit - 1) {
-            result.add(substring(currentOffset, nextIndex))
-            currentOffset = nextIndex + delimiter.length
-            // Do not search for next occurrence if we're reaching limit
-            nextIndex = if (result.size == limit - 1) -1 else indexOf(delimiter, currentOffset, ignoreCase)
-        }
+    while (nextIndex != -1 && (!isLimited || result.size < limit - 1)) {
+        result.add(substring(currentOffset, nextIndex))
+        currentOffset = nextIndex + delimiter.length
+        // Do not search for next occurrence if we're reaching limit
+        nextIndex = if (result.size == limit - 1) break else indexOf(delimiter, currentOffset, ignoreCase)
     }
 
-    if (!isLimited || result.size < limit) {
-        result.add(substring(currentOffset, length))
-    }
-
+    result.add(substring(currentOffset, length))
     return result
 }
 
--- a/libraries/stdlib/src/kotlin/text/Strings.kt
+++ b/libraries/stdlib/src/kotlin/text/Strings.kt
@@ -1167,7 +1167,7 @@ public fun CharSequence.split(vararg delimiters: String, ignoreCase: Boolean = f
     if (delimiters.size == 1) {
         val delimiter = delimiters[0]
         if (!delimiter.isEmpty()) {
-            return split(delimiters[0], ignoreCase, limit)
+            return split(delimiter, ignoreCase, limit)
         }
     }
 
@@ -1212,10 +1212,12 @@ private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: In
 
     var currentOffset = 0
     var nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
-    val isLimited = limit > 0
-    // This condition cannot be moved inside of ctor call because it will kill laziness of internal array allocation
-    val result = if (isLimited) ArrayList<String>(limit.coerceAtMost(10)) else ArrayList<String>()
+    if (nextIndex == -1) {
+        return listOf(this.toString())
+    }
 
+    val isLimited = limit > 0
+    val result = ArrayList<String>(if (isLimited) limit.coerceAtMost(10) else 10)
     while (nextIndex != -1) {
         if (!isLimited || result.size < limit - 1) {
             result.add(substring(currentOffset, nextIndex))
@@ -1223,16 +1225,8 @@ private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: In
             nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
         } else {
             result.add(substring(currentOffset, length))
-            currentOffset = length
-            break
+            return result
         }
-    }
-
-    if (currentOffset == 0) {
-        // If no matches found return single element list
-        // instead of triggering allocation of underlying array in 'result'
-        // Speedup is ~15% for small string
-        return listOf(this.toString())
     }
 
     if (!isLimited || result.size < limit) {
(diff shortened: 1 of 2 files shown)


libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1183,8 +1191,53 @@ public fun CharSequence.splitToSequence(vararg delimiters: Char, ignoreCase: Boo
...
+ *
+ * @param delimiter String used as delimiter
+ * @param ignoreCase `true` to ignore character case when matching a delimiter. By default `false`.
+ * @param limit The maximum number of substrings to return.
+ */
+private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: Int): List<String> {
+    var currentOffset = 0
+    var nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
+    val isLimited = limit > 0
+    val result = ArrayList<String>()
```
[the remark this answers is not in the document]
ilya-g: Agree, please do, but only for `isLimited` case.
Better to use `minOf`, it's available without import.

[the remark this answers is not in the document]
qwwdfsad: Why not in both cases?
10 (not 16 actually) is default array list size in both openjdk and android.

`result = ArrayList<String>(minOf(limit, 10))` looks simpler than `result = if (isLimited) ... else ...`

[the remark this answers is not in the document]
ilya-g: `minOf(limit, 10)` is 0 when the limit is 0. Is the capacity intended to be so?
Actually it would be better to use `limit.coerceAtMost(10)` here to communicate the intent more clearly.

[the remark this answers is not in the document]
qwwdfsad: Yes. "0" indicates that default size should be used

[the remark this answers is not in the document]
ilya-g: As far as I can see in the sources, `ArrayList()` and `ArrayList(0)` constructors initialize the list state differently, the latter grows its capacity more often, so it may result in more reallocations.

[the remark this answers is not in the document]
qwwdfsad: Yes, you're right

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1163,8 +1163,16 @@ public fun CharSequence.splitToSequence(vararg delimiters: String, ignoreCase: B
  * the beginning to the end of this string, and matches at each position the first element in [delimiters]
  * that is equal to a delimiter in this instance at that position.
  */
-public fun CharSequence.split(vararg delimiters: String, ignoreCase: Boolean = false, limit: Int = 0): List<String> =
-        rangesDelimitedBy(delimiters, ignoreCase = ignoreCase, limit = limit).asIterable().map { substring(it) }
+public fun CharSequence.split(vararg delimiters: String, ignoreCase: Boolean = false, limit: Int = 0): List<String> {
+    if (delimiters.size == 1) {
+        val delimiter = delimiters[0]
+        if (!delimiter.isEmpty()) {
+            return split(delimiters[0], ignoreCase, limit)
```
ilya-g: `delimiters[0]` value is already in `delimiter` local val.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1183,8 +1191,56 @@ public fun CharSequence.splitToSequence(vararg delimiters: Char, ignoreCase: Boo
...
+    // This condition cannot be moved inside of ctor call because it will kill laziness of internal array allocation
+    val result = if (isLimited) ArrayList<String>(limit.coerceAtMost(10)) else ArrayList<String>()
+
+    while (nextIndex != -1) {
+        if (!isLimited || result.size < limit - 1) {
+            result.add(substring(currentOffset, nextIndex))
+            currentOffset = nextIndex + delimiter.length
+            nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
+        } else {
+            result.add(substring(currentOffset, length))
```
ilya-g: Looks like we can return result right after this.

ilya-g: Also if we get here, it means that `nextIndex` search at the previous iteration is unused.

qwwdfsad: Simplified it a bit more, though `nextIndex` is still unused. It doesn't make any difference (in my small-medium string size benchmarks, though I'm sure it's still possible to show the opposite), but makes code less readable

ilya-g: Consider the case when you want to split strings like `key=very_long_value` with limit=2. In this case the function will search for a second `=` to the end of a long string, though it could have stopped already.

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1183,8 +1191,50 @@ public fun CharSequence.splitToSequence(vararg delimiters: Char, ignoreCase: Boo
...
+ * @param delimiter String used as delimiter
+ * @param ignoreCase `true` to ignore character case when matching a delimiter. By default `false`.
+ * @param limit The maximum number of substrings to return.
+ */
+private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: Int): List<String> {
+    require(limit >= 0, { "Limit must be non-negative, but was $limit." })
+
+    var currentOffset = 0
+    var nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
+    if (nextIndex == -1) {
```
ilya-g: Looks like we can add ` || limit == 1` because it should return the same single item list.

ilya-g: Though there isn't a lot of use cases for limit=1, thus we can avoid introducing a special path for this case.

libraries/stdlib/src/kotlin/text/Strings.kt
qwwdfsad: Point taken, fixed (and simplified again)

libraries/stdlib/src/kotlin/text/Strings.kt
```diff
@@ -1222,10 +1222,8 @@ private fun CharSequence.split(delimiter: String, ignoreCase: Boolean, limit: In
         if (!isLimited || result.size < limit - 1) {
             result.add(substring(currentOffset, nextIndex))
             currentOffset = nextIndex + delimiter.length
-            nextIndex = indexOf(delimiter, currentOffset, ignoreCase)
-        } else {
-            result.add(substring(currentOffset, length))
-            return result
+            // Do not search for next occurrence if we're reaching limit
+            nextIndex = if (result.size == limit - 1) -1 else indexOf(delimiter, currentOffset, ignoreCase)
```
ilya-g: `break` instead of `-1` would be more clear

ilya-g:
All tests have passed, so I'm going to merge this soon. Would you mind if I squash some commits?

Could you also share the code of benchmarks? A comment here or a gist would be enough.

qwwdfsad:
>Would you mind if I squash some commits?

Sure, I don't mind

>Could you also share the code of benchmarks?

I was playing with parameters in this one
```
@Fork(value = 1)//, jvmArgsAppend = arrayOf("-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"))  //, jvmArgsAppend = arrayOf("-XX:TieredStopAtLevel=1"))
@Measurement(iterations = 10)
@Warmup(iterations = 10)
@State(Scope.Benchmark)
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
open class SplitBenchmark {

    @Param("haystack")
    lateinit private var haystack: String

    @Param("a", ",")
    lateinit private var needle: String

    @Benchmark
    fun splitJava(): List<String> {
        return (haystack as java.lang.String).split(needle).toList()
    }

    @Benchmark
    fun splitKt(): List<String> {
        return haystack.split(needle)
    }

    @Benchmark
    fun splitKtLimit(): List<String> {
        return haystack.split(needle, limit = 2)
    }
}

```

ilya-g:
I've simplified it a bit more and merged, see the commits https://github.com/JetBrains/kotlin/compare/2805371bdcfa05bd2e72f86077f3a54518fd89ae~1...6b9520ec73c8a6c0f83d4ec3afd2a295364ed54a

Thank you for the PR.

PR #1476, opened 2018-01-15
Uast: `getArgumentForParameter(i: Int)` implementation for Kotlin  (IDEA-184046)

xiexed:


xiexed:
check  spread operators call `*`
arrays as single vararg parameter (if it could be distinguished on user side)
check varargs on Java api and probably introduce UVarargExpression

PR #1477, opened 2018-01-16
KT-22298 - Added documentation for Arrays copyOf(size) method

Kerooker:
[KT-22298 Array extension function copyOf(newSize: Int) should document that the copied values are copied to the start of the new array, and that the new array might use truncating and padding to resize the array](https://youtrack.jetbrains.com/issue/KT-22298)


libraries/tools/kotlin-stdlib-gen/src/templates/SpecialJVM.kt
```diff
@@ -373,7 +373,22 @@ object CommonArrays : TemplateGroupBase() {
         include(ArraysOfPrimitives, PrimitiveType.defaultPrimitives)
         include(InvariantArraysOfObjects)
     } builder {
-        doc { "Returns new array which is a copy of the original array, resized to the given [newSize]." }
+        doc {
+            """
+            Returns new array which is a copy of the original array, resized to the given [newSize], truncating or padding
+            with nulls if necessary. For all indices that are valid in both the original array and the copy, the two arrays
```
ilya-g: Primitive arrays are padded with the default values of the respective primitives rather than `null`.

Perhaps this `doc` block should be only for `InvariantArraysOfObjects`, and `ArraysOfPrimitives` should get its own version.

libraries/tools/kotlin-stdlib-gen/src/templates/SpecialJVM.kt
```diff
@@ -373,7 +373,22 @@ object CommonArrays : TemplateGroupBase() {
...
+            contents will be identical. For all indices that are valid in the copy, but not in the original, the copy will
+            contain null.
+
+            Examples:
```
ilya-g: It's good that you've provided some samples, but please use `@sample` tag. Check this [short guide](https://github.com/JetBrains/kotlin/tree/master/libraries/stdlib/samples) about sample authoring.

libraries/tools/kotlin-stdlib-gen/src/templates/SpecialJVM.kt
Kerooker: Should I commit the sources from templates too? I thought that the build process would do it, so I didn't commit the files

ilya-g: Yes, the generated sources are in vcs, so you need to commit them.

ilya-g:
Please do not merge master branch into your branch. Instead _rebase_ your branch on top of master.

ilya-g:
@Kerooker Do you need any help with this?

Kerooker:
The project simply doesn't work on my machine, so I gave up. The last error was "Unexpected lock protocol found in lock file. Expected 1, found 0.", but previously I had a problem for it being windows with WindowsRegistry.

I spent over 6 hours trying to fix the project, so I just gave up helping in any way

PR #1478, opened 2018-01-17
Filter out "Picked Up" messages from stderr

cuihtlauac:
KT-22340 Fixed


cuihtlauac:
Recreated as: https://github.com/JetBrains/kotlin/pull/1489

PR #1479, opened 2018-01-17
KT-22159 "Replace return with 'if' expression" should not place return before expressions of type Nothing

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22159


mglukhikh:
Thank you, merged

PR #1480, opened 2018-01-18
KT-21921 Extension for concatenating java.util.stream.Stream

aerovulpe:
https://youtrack.jetbrains.com/issue/KT-21921


ilya-g:
Looks good. Does it make sense to provide overloads for specialized streams as well?

aerovulpe:
@ilya-g Yeah, I think so. I've added the specialized overloads. Thanks :)

MukjepScarlet:
Is this still active?

PR #1482, opened 2018-01-19
KT-22097 Redundant Unit inspection false positive for single expression function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22097

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantUnitExpressionInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantUnitExpressionInspection.kt
@@ -9,29 +9,57 @@ import com.intellij.codeInspection.*
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElementVisitor
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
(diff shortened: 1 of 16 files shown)


mglukhikh:
Thank you, polished & merged

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantUnitExpressionInspection.kt
```diff
@@ -9,30 +9,58 @@ import com.intellij.codeInspection.*
...
 
-            // Do not report just 'Unit' in function literals (return@label Unit is OK even in literals)
-            if (parent is KtBlockExpression && parent.getParentOfType<KtFunctionLiteral>(strict = true) != null) return
+        if (this == parent.lastBlockStatementOrThis()) {
+            val prev = this.previousStatement() ?: return true
+            if (prev.isUnitLiteral()) return true
+            if (prev.getResolvedCall(analyze())?.resultingDescriptor?.returnType?.isUnit() == true) return true
+            if (prev is KtDeclaration) {
+                return if (prev is KtFunction)
+                    parent.parent?.parent?.let { it is KtIfExpression || it is KtWhenExpression } != true
```
mglukhikh: Try to avoid such constructions in future, they are too complex to understand something

PR #1483, opened 2018-01-22
KT-22374 "Join lines" works incorrectly in case of line containing more than one string literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22374


idea/src/org/jetbrains/kotlin/idea/joinLines/JoinToStringTemplateHandler.kt
```diff
@@ -35,16 +35,66 @@ class JoinToStringTemplateHandler : JoinRawLinesHandlerDelegate {
...
-
-        val offset = left.endOffset - 1
-        expression.replace(ConvertToStringTemplateIntention.buildReplacement(expression))
-        return offset
+        val binaryExpr = (plus.parent.parent as? KtBinaryExpression) ?: return -1
+        if (!binaryExpr.joinable()) return -1
+
+        val lineCount = binaryExpr.getLineCount()
+        val targetBinaryExpr = binaryExpr.expand(lineCount + 1)
+        val (left, rightText) = targetBinaryExpr.collect(lineCount)
```
goodwinnk: In general, it seems to work fine. But can you come up with the better name for `expand` and `collect` functions? Also I would suggest rewriting `expand` and `collect` without recursion.

t-kameyama: Fixed

goodwinnk:
Thank you! I have merged both commits into one: 4232ad8dfe2b6cef732be6a31cbbce1fa5b93b75.

PR #1484, opened 2018-01-23
KT-19194 Some Live Templates should probably be enabled also for "expressions" not only "statements"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19194

### DIFF
--- a/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/KotlinTemplateContextType.java
+++ b/idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/KotlinTemplateContextType.java
@@ -163,16 +163,8 @@ public abstract class KotlinTemplateContextType extends TemplateContextType {
 
         @Override
         protected boolean isInContext(@NotNull PsiElement element) {
-            PsiElement parentStatement = PsiTreeUtil.findFirstParent(element, e -> {
-                if (!(e instanceof KtExpression)) return false;
-                PsiElement parent = e.getParent();
-                ASTNode node = parent.getNode();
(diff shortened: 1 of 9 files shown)


idea/idea-live-templates/src/org/jetbrains/kotlin/idea/liveTemplates/KotlinTemplateContextType.java
```diff
@@ -162,11 +163,15 @@ public Statement() {
...
-                @Override
-                public boolean value(PsiElement element) {
-                    return element instanceof KtExpression && (element.getParent() instanceof KtBlockExpression);
-                }
+            PsiElement parentStatement = PsiTreeUtil.findFirstParent(element, e -> {
+                if (!(e instanceof KtExpression)) return false;
+                PsiElement parent = e.getParent();
+                ASTNode node = parent.getNode();
+                return parent instanceof KtBlockExpression ||
+                       parent instanceof KtContainerNodeForControlStructureBody ||
```
goodwinnk: I've found `KtPsiUtil.isStatementContainer` as on of the usages for `KtContainerNodeForControlStructureBody`, can it be used here? 
Are checks for `THEN` and `ELSE` needed? It looks like `parent instanceof KtContainerNodeForControlStructureBody` should work for them too.

t-kameyama: Fixed

goodwinnk:
Thank you!

I have merged commits into c06aaf6128c8d51211a05c075554534c077fb35f.

PR #1486, opened 2018-01-23
Add missing signature

cuihtlauac:
Private method calculateCapacity() was missing in test data.

FIX: KT-21047


cuihtlauac:
FIX: https://youtrack.jetbrains.com/issue/KT-21047

dzharkov:
Thank you for suggestion, but this fix may not help because how the test works it really depends on what minor version of JDK is being used (`calculateCapacity` is only present in the newest versions).
I'm considering to avoid rendering private inherited members in tests like this or just remove the one.

PR #1481, opened 2018-01-18
Add support for proxies to scripts

cuihtlauac:
Update jcabi-aether dependency from 0.10.1 to 1.0-SNAPSHOT. In order to
do that, add https://oss.sonatype.org/content/repositories/snapshots to
this list of repositories :-(

Script dependencies are resolved using jcabi-aether. Unfortunately "stable"
release (0.10.1) does not support proxies. Therefore, script dependencies
are not pulled behind a proxy. Issue is fixed in SNAPSHOT release of
jcabi-aether.

FIXED: KT-22363

PR #1488, opened 2018-01-24
Add support for proxies to scripts

cuihtlauac:
Update jcabi-aether dependency from 0.10.1 to 1.0-SNAPSHOT. In order to
do that, add https://oss.sonatype.org/content/repositories/snapshots to
this list of repositories :-(

Script dependencies are resolved using jcabi-aether. Unfortunately "stable"
release (0.10.1) does not support proxies. Therefore, script dependencies
are not pulled behind a proxy. Issue is fixed in SNAPSHOT release of
jcabi-aether.

FIXED: KT-22363


cuihtlauac:
Replace https://github.com/JetBrains/kotlin/pull/1481

build.gradle.kts
```diff
@@ -14,7 +14,8 @@ buildscript {
     val repos = listOfNotNull(
             bootstrapKotlinRepo,
             "https://jcenter.bintray.com/",
-            "https://plugins.gradle.org/m2")
+            "https://plugins.gradle.org/m2",
+            "https://oss.sonatype.org/content/repositories/snapshots")
```
ligee: Please, add the snapshots repo directly to the `kotlin-script-util` build script. I think we don't want (yet) to add it to all projects.

ligee:
Merged manually - https://github.com/JetBrains/kotlin/commit/a9280cdbcda6dcbe8349b1f4c73b58f1a1d53501

PR #1489, opened 2018-01-24
Filter out "Picked Up" messages from stderr

cuihtlauac:
KT-22340 Fixed


cuihtlauac:
Replace: https://github.com/JetBrains/kotlin/pull/1478
FIX: https://youtrack.jetbrains.com/issue/KT-22340

udalov:
I've replaced `Regex("...")` to `"...".toRegex()` which is considered more idiomatic Kotlin, and pushed to master. Thanks!

PR #1490, opened 2018-01-25
KT-22409 Intention for changing property setter accessibility

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22409


mglukhikh:
Thank you, merged.

PR #1491, opened 2018-01-25
KT-22374 "Join lines" works incorrectly in case of line containing more than one string literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22374#comment=27-2681388


goodwinnk:
Great, thank you!

The commit is now in master: 7e417272b2bfd6c89f9f6cc0377c6f6e0a378fa2.

PR #1492, opened 2018-01-29
KT-22167 "Add annotation target" quick fix does nothing and disappears from menu

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22167


asedunov:
Merged! Thank you for your contribution

PR #1493, opened 2018-01-29
KT-22364 Redundant setter is not reported for overridden fields

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22364


mglukhikh:
Fixed & merged manually.

PR #1494, opened 2018-01-31
KT-22620 TestNG support in kotlin.test

neonailol:
Added module to kotlin.test with bindings to testng

### DIFF
--- a/libraries/kotlin.test/testng/src/main/kotlin/Annotations.kt
+++ b/libraries/kotlin.test/testng/src/main/kotlin/Annotations.kt
@@ -1,0 +1,14 @@
+/*
+ * Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+
+@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+@file:JvmPackageName("kotlin.test.annotations")
+package kotlin.test
+
+public actual typealias Test = org.testng.annotations.Test
+public actual typealias Ignore = org.testng.annotations.Ignore
+public actual typealias BeforeTest = org.testng.annotations.BeforeClass
+public actual typealias AfterTest = org.testng.annotations.AfterClass
--- a/libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
+++ b/libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
@@ -11,8 +11,7 @@ import kotlin.test.*
 class TestNGContributor : AsserterContributor {
     override fun contribute(): Asserter? {
         for (stackFrame in currentStackTrace()) {
-            @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-            val className = stackFrame.className as java.lang.String
+            val className = stackFrame.className
 
             if (className.startsWith("org.testng.") || className.startsWith("testng.")) {
                 return TestNGAsserter
--- a/libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
+++ b/libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package kotlin.test.testng
+
+import org.testng.*
+import kotlin.test.*
+
+class TestNGContributor : AsserterContributor {
+    override fun contribute(): Asserter? {
+        for (stackFrame in currentStackTrace()) {
+            @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+            val className = stackFrame.className as java.lang.String
+
+            if (className.startsWith("org.testng.") || className.startsWith("testng.")) {
+                return TestNGAsserter
+            }
+        }
+
+        return null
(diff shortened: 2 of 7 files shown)


libraries/kotlin.test/testng/src/main/kotlin/Annotations.kt
```diff
@@ -0,0 +1,14 @@
...
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+
+@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+@file:JvmPackageName("kotlin.test.annotations")
+package kotlin.test
+
+public actual typealias Test = org.testng.annotations.Test
+public actual typealias Ignore = org.testng.annotations.Ignore
```
ilya-g: Interesting, I hadn't noticed this annotation in javadoc of testng before, seems that it's rather new. What is the testng version it's available from?

libraries/kotlin.test/testng/src/main/resources/META-INF/services/kotlin.test.AsserterContributor
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+# that can be found in the license/LICENSE.txt file.
+#
+
+kotlin.test.junit.TestNGContributor
```
ilya-g: Package name seems incorrect

libraries/kotlin.test/testng/src/test/kotlin/TestNGContributorTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+        Thread {
+            q.put(asserter)
+        }.start()
+
+        Assert.assertEquals("DefaultAsserter", q.take().`class`.simpleName)
+    }
+
+    @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+    private val asserter: java.lang.Object
```
ilya-g: `asserter` is a public top-level property, it can be obtained directly without reflection.

libraries/kotlin.test/testng/src/main/kotlin/Annotations.kt
neonailol: I suppose you talking about `@Ignore`, i believe it was added in latest version 6.13

neonailol:
realized i attached wrong issue to PR, and did not find something similar in youtrack, created new issue KT-22620 for this pr

libraries/kotlin.test/testng/src/main/resources/META-INF/services/kotlin.test.AsserterContributor
ilya-g: And a copyright here is not required

neonailol:
@ilya-g thanks for review, changed code accordingly. 
should i create new PR without merge commit?

libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
```diff
@@ -11,8 +11,7 @@ import kotlin.test.*
 class TestNGContributor : AsserterContributor {
     override fun contribute(): Asserter? {
         for (stackFrame in currentStackTrace()) {
-            @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-            val className = stackFrame.className as java.lang.String
+            val className = stackFrame.className as String
```
ilya-g: isn't `className` already String?

libraries/kotlin.test/testng/src/test/kotlin/TestNGContributorTest.kt
```diff
@@ -5,24 +5,21 @@ import org.testng.*
 import java.util.concurrent.*
 
 class TestNGContributorTest {
+
     @Test
     fun smokeTest() {
-        Assert.assertEquals("TestNGAsserter", asserter.`class`.simpleName)
+        Assert.assertEquals("TestNGAsserter", kotlin.test.asserter)
```
ilya-g: This won't pass most likely, you're comparing `Asserter` object with a String

ilya-g:
@neonailol You can just force push to your branch, and this review will be updated accordingly.

libraries/kotlin.test/testng/src/main/kotlin/TestNGSupport.kt
neonailol: yes it is

libraries/kotlin.test/testng/src/test/kotlin/TestNGContributorTest.kt
neonailol: i run `./gradlew kotlin-test:kotlin-test-testng:test` and `./gradlew kotlin-test:kotlin-test-testng:test` before committing, everything was good

neonailol:
@ilya-g implemented suggested changes

neonailol:
@ilya-g is everything ok, or more changes needed?

ilya-g:
The code seems fine now, next I'll run this branch on our build server.

Just one question's left: do we need to specify something in `build.gradle` to trigger `testng` runner for test?

neonailol:
@ilya-g yes indeed, forgot about whis part, in maven it's automatically picked up, but in gradle we need to specify this explicitly

ilya-g:
Thanks, I've merged the PR manually into master and 1.2.30 branches, see the commit range https://github.com/JetBrains/kotlin/compare/ad7701922c9c083875b759a30a8522ced734178a~1...0f130a0a5aa384393d5393166cfab8a9182d7fe1

PR #1495, opened 2018-02-01
KT-22110 "Can be joined with assignment" inspection underlining extends into comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22110


mglukhikh:
I think it's Ok, thank you. The only point: may be using modifier list and/or val/var keyword looks a bit nicer and operates a bit faster than `getNextSiblingIgnoringWhitespacesAndComments`. I'll try to refactor your code this way.

mglukhikh:
Merged

PR #1496, opened 2018-02-01
Remove unnecessary dependencies from AbstractBoundedValue

mikaelpeltier:
- Bounded value is a low-level abstraction for bounded ranges that is
used to generate specialized contains checks and for loop. It only
needs to contains StackValue that will be used to generate code.
- SimpleBoundedValue already have a constructor supporting StackValue
but having only one constructor avoid to pass extra argument that
will be used to only compute another information.


mikaelpeltier:
Sorry, I forgot to commit some modified files, I will add them.

mikaelpeltier:
This PR was tested by:
./gradlew dist
./gradlew jvmCompilerTest

The following test failed with my PR but also without, thus it seems not related to my change.
org.jetbrains.kotlin.jvm.compiler.CompileKotlinAgainstJavaTestGenerated.testListImpl

dnpetrov:
Merged manually, thanks!

PR #1498, opened 2018-02-02
KT-22334 Specialized loops using range such as integer..array.size-1

mikaelpeltier:
- Into PrimitiveNumberRangeLiteralRangeValue modifies how bounded
value are created by checking if the high bound range can be modified
to be exclusive instead of inclusive such as the generated code will
be optimized.
- Add black box tests checking that the specialization does not fail
on any kind of arrays.
- Add a bytecode test checking that the code is correctly optimized.

Fix of https://youtrack.jetbrains.com/issue/KT-22334

### DIFF
--- a/compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
+++ b/compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
@@ -1,79 +1,79 @@
 // WITH_RUNTIME
 
-fun checkByteArray():Boolean {
+fun checkByteArray(): Boolean {
     val byteArray = byteArrayOf(1, 2, 3)
     var sum = 0
-    for (i in 0..byteArray.size-1) {
+    for (i in 0..byteArray.size - 1) {
         sum += byteArray[i]
     }
     if (sum != 6) return false
     return true
 }
 
-fun checkShortArray():Boolean {
+fun checkShortArray(): Boolean {
     val shortArray = shortArrayOf(1, 2, 3)
     var sum = 0
-    for (i in 0..shortArray.size-1) {
+    for (i in 0..shortArray.size - 1) {
         sum += shortArray[i]
     }
     if (sum != 6) return false
     return true
 }
 
-fun checkCharArray():Boolean {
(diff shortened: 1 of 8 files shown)


mikaelpeltier:
This PR was tested by:
./gradlew dist
./gradlew jvmCompilerTest

The following test failed with my PR but also without, thus it seems not related to my change.
org.jetbrains.kotlin.jvm.compiler.CompileKotlinAgainstJavaTestGenerated.testListImpl

compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
```diff
@@ -0,0 +1,143 @@
+// WITH_RUNTIME
+
+fun checkByteArray():Boolean {
```
jshvarts: here and in other functions below, add space before Boolean return type (`fun checkByteArray(): Boolean`)

compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
```diff
@@ -0,0 +1,143 @@
...
+    var start = 0
+    var sum = 0
+    for (i in (start..intArray.size-1).reversed()) {
+        sum += intArray[i]
+    }
+    if (sum != 6) return false
+    return true
+}
+
+fun box() : String {
```
jshvarts: spacing fun box(): String

compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
mikaelpeltier: Done, I reformatted the test

compiler/testData/codegen/box/controlStructures/forInArray/forInArraySpecializedToUntil.kt
mikaelpeltier: Done, I reformatted the test

mikaelpeltier:
Pushed to master by dnpetrov

PR #1500, opened 2018-02-05
KT-18731 Use reference equality to compare enums

mikaelpeltier:
Fix of https://youtrack.jetbrains.com/issue/KT-18731


mikaelpeltier:
This PR was tested by:
./gradlew dist
./gradlew jvmCompilerTest

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3107,6 +3108,18 @@ private StackValue generateEquals(
             );
         }
 
+        if ((opToken == KtTokens.EQEQ || opToken == KtTokens.EXCLEQ) &&
+            (isEnumClass(bindingContext.getType(left).getConstructor().getDeclarationDescriptor()) ||
```
dnpetrov: LHS should also be non-null. Use `TypeUtils.isNullableType` to check that type is (non-)nullable (it implements detailed check for possibly parameterized type.

The following test should pass:
```
enum class Test { A, B }

fun eq(x: Test?, y: Test) = x == y

fun box() = if (!eq(null, A)) "OK" else "Fail"
```

dnpetrov:
Pushed to master.

PR #1501, opened 2018-02-05
Added components.xml to automatically compile kotlin maven projects

clarkperkins:
I've added a components.xml file to the maven plugin, overriding the default lifecycle mapping for `jar`, `ejb`, `war`, and `rar` packaging types.  With this file, you can enable extensions on the kotlin-maven-plugin, and the `kapt`, `compile`, `test-kapt`, and `test-compile` executions will automatically be added to your build, bound to their appropriate lifecycle phases.

```xml
<build>
    <plugins>
        <plugin>
            <groupId>org.jetbrains.kotlin</groupId>
            <artifactId>kotlin-maven-plugin</artifactId>
            <version>1.2-SNAPSHOT</version>  <!-- of course you'd want a released version here eventually -->
            <extensions>true</extensions>
            <!-- No need for the executions section here -->
        </plugin>
    </plugins>
</build>
```


goodwinnk:
I have created an issue about this proposal, with the list of additional connected task: https://youtrack.jetbrains.com/issue/KT-29346.

bdemers:
Please!

I came here to open a similar issue :) after hacking up a similar `components.xml`.

I also needed to add a Mojo to configure the source roots correctly out of the box (as maven defaults to `/src/main/java`)
https://gist.github.com/bdemers/53203fe313776932a7113a8e91ee5911

The `components.xml` phases would need to execute this goal too:
```xml
<initialize>
    ${project.groupId}:${project.artifactId}:${project.version}:configure
</initialize>
```

I'm a bit of a Kotlin noob, so I'm not sure this was caused by me doing something wrong, or just how the current kotlin-maven-plugin works.

bdemers:
Awesome!! Thank you for including this!

CherepanovAleksei:
@bdemers , @clarkperkins, in such case, how do you configure kapt? For e.g. how do you pass annotationProcessorPaths or other options?

bdemers:
You can still configure the plugin(s) as you normally would, the extension usage here, binds the appropriate plugin goals to the maven phases. Basically, you should be able to remove a lot of `<execution>` blocks from your pom.

You would still need any relevant `<configuration>` though.

PR #1502, opened 2018-02-06
KT-22714 Avoid to generate unecessary checkcast

mikaelpeltier:
- StackValue already avoid to generate checkcast from a type or an
array to java.lang.Object. Add a new case to avoid to generate a
checkcast from an array to an array of java.lang.Object when arrays
have the same dimensions.

Fix of https://youtrack.jetbrains.com/issue/KT-22714

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/StackValue.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/StackValue.java
@@ -395,7 +395,15 @@ public abstract class StackValue {
             }
         }
         else if (toType.getSort() == Type.ARRAY) {
-            v.checkcast(toType);
+            if (fromType.getSort() != Type.ARRAY) {
+                v.checkcast(toType);
+            }
+            else if (toType.getDimensions() != fromType.getDimensions()) {
+                v.checkcast(toType);
+            }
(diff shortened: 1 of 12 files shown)


max-kammerer:
@mikaelpeltier Why do you need 'JVM_TARGET: 1.8' in tests? Seems both tests could be moved to  codegen/box and codegen/bytecodeText

mikaelpeltier:
It is perhaps not mandatory, following a discussion with @dnpetrov, I used JVM_TARGET 1.8 to check that my change does not break something with the stack frame generation. Nevertheless, if you estimate that it is not needed, I can move my test into `codegen/box` and `codegen/bytecodeText`. 
Tell me what you prefer.

max-kammerer:
Yes, smth could be broken, but:
1. I didn't see anything in your test that could do it (and i can't  come up with smth)
2. we also run common tests with switched default jvm target to 1.8 (see :compiler:codegenTarget8Jvm8Test)
so, pls move tests

mikaelpeltier:
Tests moved and PR tested by:
./gradlew dist
./gradlew jvmCompilerTest

max-kammerer:
Thanks! Merged manually

PR #1504, opened 2018-02-08
KT-22719 Incorrect warning 'Redundant semicolon' when having method call before lambda expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22719


mglukhikh:
Merged without changes, thank you.

PR #1506, opened 2018-02-09
KT-22492 "Specify explicit lambda signature" intention is available only on lambda braces

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22492


mglukhikh:
I think it's Ok, but now it's logical to derive from `SelfTargetingOffsetIndependentIntention`. I'll add the relevent commit. Thank you.

mglukhikh:
Merged

PR #1509, opened 2018-02-10
KT-22428 Create member function from usage shouldn't present type parameters as options

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22428


yole:

PR #1513, opened 2018-02-12
Do not use hardcoded plugin ID for update checks

stigger:


goodwinnk:
Would you mind merging it into one commit?

stigger:
@goodwinnk Done.

goodwinnk:
What exceptions are meant in commit message? Are they for Kotlin plugin? 

I think "Do not use hardcoded plugin ID for update checks" would be a better name, and some explanation why this is needed would be nice.

stigger:
@goodwinnk Done

goodwinnk:
Merged manually as 53946456cfe90100f1ffc4af9a0eefe5a60b2a35. Thank you!

PR #1514, opened 2018-02-12
Fix import of `expectedBy` scope dependencies inside composite builds

nskvortsov:
Gradle's subproject name/path is only a part of a module ID, when composite build is imported. Included build's name is missing.
It is easy to obtain it, as long as `expectedBy` dependency does not cross gradle build boundary.


asedunov:
Merged! Thank you

PR #1519, opened 2018-02-15
KT-20357: Add samples for any, all and none

yusaka39:

Add samples for `any`, `all` and `none` and regenerate stdlib codes.

### DIFF
--- a/js/js.libraries/src/core/generated/_ArraysJs.kt
+++ b/js/js.libraries/src/core/generated/_ArraysJs.kt
@@ -9491,6 +9491,8 @@ public infix fun CharArray.union(other: Iterable<Char>): Set<Char> {
 
 /**
  * Returns `true` if all elements match the given [predicate].
+ * 
+ * @sample samples.collections.Collections.Aggregates.all
  */
 public inline fun <T> Array<out T>.all(predicate: (T) -> Boolean): Boolean {
     for (element in this) if (!predicate(element)) return false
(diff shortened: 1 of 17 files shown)


libraries/stdlib/samples/test/samples/aggregates/aggregates.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package samples.aggregates
+
+import samples.Sample
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+
+class Aggregates {
```
ilya-g: Better to place it as a nested class of `samples.collections.Collections`

libraries/stdlib/samples/test/samples/aggregates/aggregates.kt
```diff
@@ -0,0 +1,99 @@
...
+    @Sample
+    fun all() {
+        val isEven: (Int) -> Boolean = { it % 2 == 0 }
+        val zeroToTen = 0..10
+        assertFalse {
+            zeroToTen.all(isEven)
```
ilya-g: Though we can definitely use `all`/`any`/`none` with a callable reference, the most common usage is with a lambda function: `.all { isEven(it) }`. Probably, we can show both options.

YutaSakata:
@ilya-g
Thank you for your review. I updated and squashed the branch.

yusaka39:
@ilya-g
Thank you for your review. I updated and squashed the branch.

ilya-g:
Thank you, I've merged this PR manually with minor edits.

PR #1520, opened 2018-02-15
KT-22869: asserting exception message in the assertFailsWith method

adikm:
https://youtrack.jetbrains.com/issue/KT-22869

### DIFF
--- a/libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
+++ b/libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
@@ -20,6 +20,29 @@ class BasicAssertionsTest {
     }
 
     @Test
+    fun testAssertFailsWithMessage() {
+        assertFailsWith<NoSuchElementException>("This message is asserted") {
+            throw NoSuchElementException("This message is asserted")
+        }
+    }
+
+    @Test
+    fun testAssertFailsWithMessageFails() {
+        assertTrue(true) // at least one assertion required for qunit
+
+        withDefaultAsserter run@ {
+            try {
+                assertFailsWith<IllegalArgumentException>("message") {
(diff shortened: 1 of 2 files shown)


libraries/kotlin.test/jvm/src/main/kotlin/AssertionsImpl.kt
```diff
@@ -28,8 +28,10 @@ private fun <T : Throwable> assertFailsWithImpl(exceptionClass: Class<T>, messag
         block()
     } catch (e: Throwable) {
         if (exceptionClass.isInstance(e)) {
-            @Suppress("UNCHECKED_CAST")
-            return e as T
+            if (message == null || message == e.message) {
```
JakeWharton: This should be replaced with a proper assertion and failure. If the message doesn't match you'll get a cryptic failure saying "Expected an exception of type RuntimeException to be thrown, but was RuntimeException".

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -19,6 +19,29 @@ class BasicAssertionsTest {
...
+    fun testAssertFailsWithMessageFails() {
+        assertTrue(true) // at least one assertion required for qunit
+
+        withDefaultAsserter run@ {
+            try {
+                assertFailsWith<IllegalArgumentException>("message") {
+                    throw IllegalArgumentException("another message")
+                }
+            } catch (e: AssertionError) {
+                return@run
```
JakeWharton: You should assert on the message here actually indicating a mis-matched message. This also obviates the need for the dummy assertTrue.

ilya-g:
Closing this, according to the issue comment: https://youtrack.jetbrains.com/issue/KT-22869#comment=27-2709680

If you want to submit the docs improvement, please reopen this PR or open a new one.

PR #1521, opened 2018-02-16
KT-16332 Add braces to 'if' statement intention does not put end-of-line comment properly into braces

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16332


mglukhikh:
It's mostly good. However, we have a special class `CommentSaver` which is intended to save and restore comments during code transformations. I will try to change your code to use this saver and check results on your tests.

mglukhikh:
So it can work with `CommentSaver` but comments still have to be deleted manually. You can look at my commit if you are interested (it's added to yours): https://github.com/JetBrains/kotlin/commit/dd106cb65c4ccbb96a0c02f28ba44e9f9a69f96b

PR #1522, opened 2018-02-16
KT-15858 "Replace with a 'foreach' function call" intention breaks code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15858


mglukhikh:
Fixed and merged. Fixing commit (again, `CommentSaver` was in use): https://github.com/JetBrains/kotlin/commit/bec28c8388d050d371c333391bc0b3f81cf35a7a. Thank you

PR #1524, opened 2018-02-17
Log Diagnostic.Kind.NOTE messages at INFO level

joshallenit:
javax.tools.Diagnostic.Kind.NOTE messages from libraries should be equivalent to INFO level, not WARNING level. I think WARNING was used because by default gradle does not output info level (you need to add the `--info` flag).

With the latest version of the Kotlin plugin (org.jetbrains.kotlin:kotlin-gradle-plugin:1.2.21) our builds broke because Kapt was logging  `javax.tools.Diagnostic.Kind.NOTE` messages from annotation processors (in our case `Realm`) as warnings, instead of as info messages which would make more sense. Our build script essentially looks for `warnings found and -Werror specified` messages from kapt output. 

The build script looks something like this:

```
subprojects { subproject ->
    afterEvaluate {
        tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all { task ->
            if (task.name.endsWith("UnitTestKotlin")) {
                kotlinOptions.suppressWarnings = true
            } else {
                kotlinOptions.allWarningsAsErrors = true
            }
        }
        /*
         * Check tasks that are affected by allWarningsAsErrors but are not failed on warnings, such as kapt* tasks.
         * Modified from https://stackoverflow.com/a/33153700
         */
        tasks.all { task ->
            def standardOutput = new StringBuilder()
            def listener = new StandardOutputListener() {
                @Override
                void onOutput(CharSequence output) {
                    standardOutput.append(output)
                }
            }
            doFirst {
                getLogging().addStandardErrorListener(listener)
            }
            doLast {
                getLogging().removeStandardErrorListener(listener)
                if (standardOutput =~ "warnings found and -Werror specified") {
                    throw new GradleException(
                            "warnings found and -Werror specified, from task " + subproject.name + ":" + task.name)
                }
            }
        }
    }
}
```

### DIFF
--- a/plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/javac/KaptJavaLog.kt
+++ b/plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/javac/KaptJavaLog.kt
@@ -239,7 +239,7 @@ class KaptJavaLog(
 
                 val errWriter = makeWriter(ERROR)
                 val warnWriter = makeWriter(STRONG_WARNING)
-                val noticeWriter = makeWriter(WARNING)
+                val noticeWriter = makeWriter(INFO)
 
                 KaptJavaLog(
                     kaptContext.project, newContext, errWriter, warnWriter, noticeWriter,


yanex:
A number of users requested the opposite: they don't see messages from annotation processors in terminal unless the build is ran with the `--info` or `--debug` option specified.
So I'd not merge this until there will be a solution that wouldn't harm both sides.

Josh-Allen-ConnectedLab:
@yanex The change you made to make NOTE level messages WARNINGS is 100% wrong. For example, turn on  "treat warnings as errors", and you will suddenly see red lines that show up in your build output because of NOTE messages from third party libraries. There is a feature to see info level messages as you mentioned, using the `--info` flag. Please revert your erroneous commit.

yanex:
There will be a toggle in 1.2.60: https://youtrack.jetbrains.com/issue/KT-24985.

Josh-Allen-ConnectedLab:
Awesome, thanks.

yanex:
I'll close this as KT-24985 is done and will be available in 1.2.60.

Josh-Allen-ConnectedLab:
@yanex Is there any documentation on the toggle or how to set it? 

Nothing is mentioned on the kapt page.
https://kotlinlang.org/docs/reference/kapt.html

Josh-Allen-ConnectedLab:
@yanex Bump ^^ Is there any documentation on the toggle or how to set it?

Nothing is mentioned on the kapt page or in issue KT-24985

yanex:
@Josh-Allen-ConnectedLab Kapt now has a [`infoAsWarnings`](https://github.com/JetBrains/kotlin/blob/d6459e6c49227be11323ce59310407b17005c117/plugins/kapt3/kapt3-cli/src/KaptCliOption.kt#L158) option. If it's enabled, kapt translates 'info' messages as warnings.
The option is disabled by default.

PR #1525, opened 2018-02-17
KT-22861 "Add annotation target" quick fix is not available on annotation with use site target

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22861
https://youtrack.jetbrains.com/issue/KT-22862

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/AnnotationChecker.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/AnnotationChecker.kt
@@ -221,24 +221,15 @@ class AnnotationChecker(
             }.toSet()
         }
 
-        fun getDeclarationSiteActualTargetList(
-            annotated: KtElement,
-            descriptor: ClassDescriptor?,
-            trace: BindingTrace,
-            all: Boolean = false
-        ):
+        fun getDeclarationSiteActualTargetList(annotated: KtElement, descriptor: ClassDescriptor?, trace: BindingTrace):
                 List<KotlinTarget> {
-            val actualTargetList = getActualTargetList(annotated, descriptor, trace)
-            return if (all)
-                actualTargetList.defaultTargets + actualTargetList.onlyWithUseSiteTarget + actualTargetList.canBeSubstituted
-            else
-                actualTargetList.defaultTargets
+            return getActualTargetList(annotated, descriptor, trace).defaultTargets
         }
 
         private fun DeclarationDescriptor?.hasBackingField(bindingTrace: BindingTrace) =
             (this as? PropertyDescriptor)?.let { bindingTrace.get(BindingContext.BACKING_FIELD_REQUIRED, it) } ?: false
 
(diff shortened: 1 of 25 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/AddAnnotationTargetFix.kt
```diff
@@ -81,14 +84,30 @@ private fun KtAnnotationEntry.getRequiredAnnotationTargets(annotationClass: KtCl
     }.flatten().toSet()
 
     val annotationTargetValueNames = AnnotationTarget.values().map { it.name }
-    return (requiredTargets + otherReferenceRequiredTargets).filter { it.name in annotationTargetValueNames }
+    return (requiredTargets + otherReferenceRequiredTargets).toSet().filter { it.name in annotationTargetValueNames }
```
mglukhikh: Probably `distinct()` is better here

compiler/frontend/src/org/jetbrains/kotlin/resolve/AnnotationChecker.kt
```diff
@@ -221,9 +221,18 @@ class AnnotationChecker(
             }.toSet()
         }
 
-        fun getDeclarationSiteActualTargetList(annotated: KtElement, descriptor: ClassDescriptor?, trace: BindingTrace):
+        fun getDeclarationSiteActualTargetList(
```
mglukhikh: I don't like an idea to change front-end code in this commit: as result we get an additional parameter used only from IDE code. I think it's better to call `getActualTargetList` from fix (make it public for this), and then take necessary targets from this list (effectively inlining your function here). As alternative idea, you can write separate function in IDE for a case with `all = true`.

idea/src/org/jetbrains/kotlin/idea/quickfix/AddAnnotationTargetFix.kt
```diff
@@ -81,14 +84,30 @@ private fun KtAnnotationEntry.getRequiredAnnotationTargets(annotationClass: KtCl
...
+
+    if (useSiteTarget == null) {
+        return AnnotationChecker.getDeclarationSiteActualTargetList(annotatedElement, null, BindingTraceContext())
+
+    } else {
+        val target = KotlinTarget.USE_SITE_MAPPING[useSiteTarget?.getAnnotationUseSiteTarget()] ?: return emptyList()
+
+        val actualTargetList =
+            AnnotationChecker.getDeclarationSiteActualTargetList(annotatedElement, null, BindingTraceContext(), all = true)
+        if (actualTargetList.none { it == target }) return emptyList()
```
mglukhikh: `if (target !in actualTargetList)`

t-kameyama:
Fixed

mglukhikh:
I think now it's Ok, I will recheck and merge it. Thank you.

mglukhikh:
Polished a bit and merged, thank you.

PR #1526, opened 2018-02-17
add samples for CharJVM.kt

KirillMakarov:
I choose to use `assertPrints` instead of `assertTrue`, because 
`println('a'.isDigit()) // false`
 is more readable than 
`println("'a'.toDigit() is ${'a'.isDigit()}") // false`

### DIFF
--- a/libraries/stdlib/samples/test/samples/text/chars.kt
+++ b/libraries/stdlib/samples/test/samples/text/chars.kt
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package samples.text
+
+import samples.*
+
+class Chars {
+
+    @Sample
+    fun isDefined() {
+        assertPrints('$'.isDefined(), "true")
+    }
+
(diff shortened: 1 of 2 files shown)


ilya-g:
> I choose to use assertPrints instead of assertTrue

It might be more readable when a sample contains just one assertion, but when there are several of them, you'd get the following output if you actually run sample:
```
true
true
false
false
```

So it would be hard to correlate which expression yield which result.

KirillMakarov:
I think blank line between true and false statements in output will help, because samples are grouped by true/false. How we can add blank line in samples? Maybe we should add `assertPrints("","")` between true and false statements? What do you think?

KirillMakarov:
It is easy to fix, but we should choose the best option

ilya-g:
Another option to get more compact but not less expressive samples for `is*` extensions is the following:
```
val chars = listOf('a', 'β', '+', '1')
val (letters, notLetters) = chars.partition { it.isLetter() }

assertPrints(letters, "[a, β]")
assertPrints(notLetters, "[+, 1]")
```

WDYT?

KirillMakarov:
Ok, great idea, I will change

ilya-g:
We have implemented this idea in PR #1738

PR #1527, opened 2018-02-19
KT-22011 Inspection to report the usage of Java Collections methods on immutable Kotlin Collections

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22011


mglukhikh:
Well, I like the code but I don't like the diagnostic messages. I will change them and merge this.

PR #1528, opened 2018-02-20
KT-20357 Add samples for take* and drop* functions

gzoritchak:
Add samples for 
take, takeLast, takeWhile, takeLastWhile,
drop, dropLast, dropWhile, dropLastWhile


ilya-g:
Thanks, I've reformatted samples a bit and merged them with squash.

gzoritchak:
Ok. 

When there is quite a lot of functions I tend to keep them in a alphabetic order but as `drop` and `last` are so close in their role it make sense to keep them next to the the other.

PR #1529, opened 2018-02-20
KT-21743 Method reference not correctly moved into parentheses

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21743

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/MoveSuspiciousCallableReferenceIntoParenthesesInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/MoveSuspiciousCallableReferenceIntoParenthesesInspection.kt
@@ -13,9 +13,11 @@ import com.intellij.psi.PsiElementVisitor
 import org.jetbrains.kotlin.builtins.getReturnTypeFromFunctionType
 import org.jetbrains.kotlin.builtins.isBuiltinFunctionalType
 import org.jetbrains.kotlin.builtins.isBuiltinFunctionalTypeOrSubtype
+import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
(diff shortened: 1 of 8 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/MoveSuspiciousCallableReferenceIntoParenthesesInspection.kt
```diff
@@ -43,26 +45,36 @@ class MoveSuspiciousCallableReferenceIntoParenthesesInspection : AbstractKotlinI
...
             val callableReference = callableReferenceExpression.callableReference
-            val resolvedCall = callableReference.resolveToCall(BodyResolveMode.FULL)
-            val receiverValue = resolvedCall?.let { it.extensionReceiver ?: it.dispatchReceiver }
-            return (receiverValue?.let { "${it.type}" } ?: "") + "::${callableReference.text}"
+            val receiverExpression = callableReferenceExpression.receiverExpression
+            val receiver = if (receiverExpression == null) {
+                ""
+            } else {
+                val descriptor = receiverExpression.getCallableDescriptor()
+                if (descriptor == null || descriptor is ValueParameterDescriptor)
```
mglukhikh: This is suspicious place for me. Could you please explain why `ValueParameterDescriptor`, and only this descriptor, should be handled separately here?

t-kameyama: It is for lambda parameter. If remove ` || descriptor is ValueParameterDescriptor`, following tests fail.
- `defaultParameter.kt`
- `it.kt`
- `parameter.kt`
- `parameter2.kt`

mglukhikh: It should be not just "some" parameter, but this lambda parameter. See this fixing commit: https://github.com/JetBrains/kotlin/commit/b739db3c3573fb24daf1e97c42746d617dea4692.

mglukhikh:
Thank you, I fixed & merged it.

PR #1530, opened 2018-02-20
KT-22933 Add intention/inspection to convert Pair constructor to `to` function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22933


idea/src/META-INF/plugin.xml
```diff
@@ -1554,6 +1554,11 @@
       <category>Kotlin</category>
     </intentionAction>
 
+    <intentionAction>
```
mglukhikh: Really inspection is enough here. Since IDEA 2016.3, we have `INFORMATION` level for inspection (= No highlighting, only fix), so separate intentions are not necessary in this situation.

idea/src/META-INF/plugin.xml
```diff
@@ -2710,6 +2715,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.intentions.ConvertPairConstructorToToFunctionInspection"
+                     displayName="Convert Pair constructor to 'to' function"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="WEAK WARNING"
```
mglukhikh: Not quite sure but I'd prefer `INFORMATION` level here. I think these two kinds of syntax are more or less equivalent. May be `to` looks a bit nicer, but not in every case.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPairConstructorToToFunctionIntention.kt
```diff
@@ -0,0 +1,53 @@
...
+                )
+            }
+        }
+    }
+}
+
+private fun KtCallExpression.isPairConstructorCall(): Boolean {
+    if (valueArguments.size != 2) return false
+    if (valueArguments.mapNotNull { it.getArgumentExpression() }.size != 2) return false
+    return getCallableDescriptor()?.containingDeclaration?.fqNameSafe == FqName("kotlin.Pair")
```
mglukhikh: Please move declaration of this `getCallableDescriptor` to `Utils` file. It's normal place for such functions, if they have multiple use-sites.

t-kameyama:
Fixed

mglukhikh:
It should work, thank you.

PR #1535, opened 2018-02-24
KT-23017: Created EnumRange to make ClosedRanges of Enums Iterable

GrahamLea:
See: https://youtrack.jetbrains.com/issue/KT-23017


GrahamLea:
Note that the `@SinceKotlin("1.2")` annotations probably need modification, but 1.3 didn't work so looking for advice on what to do there...?

PR #1537, opened 2018-02-26
KT-14258 Optimize accesses to private properties defined into companion

mikaelpeltier:
KT-14258 Optimize accesses to private properties defined into companion

- Use direct access to property defined into companion object when
it is possible rather than always use an accessor to access the property.
- Use direct access will speedup runtime performance.
- Avoid to generate accessors when they are not used to generate smaller code.
- Adapt some tests that expect getter and setter to be generated.

Fix of https://youtrack.jetbrains.com/issue/KT-14258


compiler/backend/src/org/jetbrains/kotlin/codegen/PropertyCodegen.java
```diff
@@ -208,8 +215,16 @@ private boolean isAccessorNeeded(
...
         if (declaration.hasDelegate() || declaration.getReceiverTypeReference() != null) return true;
 
-        // Companion object properties always should have accessors, because their backing fields are moved/copied to the outer class
-        if (isCompanionObject(descriptor.getContainingDeclaration())) return true;
+        // Companion object properties should have accessors for non-private properties because these properties can be referenced
+        // via getter/setter. But these accessors getter/setter are not required for private properties that have a default getter
+        // and setter, in this case, the property can always be accessed through the accessor 'access<property name>$cp' and avoid some
+        // useless indirection by using others accessors.
+        if (isCompanionObject(descriptor.getContainingDeclaration())) {
+            if (Visibilities.isPrivate(descriptor.getVisibility()) && isDefaultGetterAndSetter) {
```
udalov: I'm a bit confused by this. Why is the fact that _both_ property accessors are default so important that it defines whether or not should we generate each of those accessors? The common sense is that the getter should be generated only if the getter is non-default, and the setter should be generated only if the setter is non-default. If only one of them is non-default, then that should be generated, and the other should not.

For example, in this case:
```
class A {
    companion object {
        private var x: String = ""
            get() = field
    }
}
```
I'd expect `A.Companion.getX` but _no setter_ to be generated.

I also couldn't find any tests on this behavior, since after dropping these lines (and relying on the accessor visibility check below in this method) all codegen tests still pass.

compiler/testData/asJava/lightClasses/compilationErrors/TraitClassObjectField.kt
```diff
@@ -4,5 +4,7 @@ interface TraitClassObjectField {
     companion object {
         const val x: String? = ""
         private val y: String? = ""
+            // Custom getter is needed, otherwise no need to generate getY
```
udalov: Please remove the accessor from the `.java` file in this test instead of this change

compiler/testData/asJava/lightClasses/nullabilityAnnotations/ClassObjectField.kt
```diff
@@ -4,5 +4,7 @@ class ClassObjectField {
     companion object {
         val x: String? = ""
         private val y: String? = ""
+            // Custom getter is needed, otherwise no need to generate getY
```
udalov: Please remove the accessor from the `.java` file in this test instead of this change

compiler/testData/codegen/box/companion/kt14258.kt
```diff
@@ -0,0 +1,12 @@
+// Checks that methods 'access$getMy$p', 'access$getMy$cp' and 'getMy' are not generated and
```
udalov: I think this box test is unnecessary since it doesn't really check what's described here, it only checks that private properties in companion objects work, which is done elsewhere, so please delete this test

compiler/testData/codegen/bytecodeText/companion/kt14258.kt
```diff
@@ -0,0 +1,19 @@
...
+
+class My {
+    companion object {
+        private val my: String = "OK"
+    }
+
+    fun getMyValue() = my
+}
+
+fun box() = My().getMyValue()
```
udalov: `fun box()` is not needed in bytecode text tests

compiler/testData/codegen/bytecodeText/companion/kt14258.kt
```diff
@@ -0,0 +1,19 @@
+// Checks that methods 'access$getMy$p', 'access$getMy$cp' and 'getMy' are not generated and
```
udalov: Why have you deleted the `kt-14258.kt` test from your earlier commits where a lot more cases were covered? If you think that test was getting too big to understand, you can always split it into several independent tests, e.g. `kt14258_1.kt`, `kt14258_2.kt`, ...

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1957,7 +1961,10 @@ private CodegenContext getBackingFieldContext(
             fieldAccessorKind = FieldAccessorKind.LATEINIT_INTRINSIC;
         }
         else if (isBackingFieldInClassCompanion &&
-            (forceField || propertyDescriptor.isConst() && Visibilities.isPrivate(propertyDescriptor.getVisibility()))) {
+                 (forceField ||
+                  (Visibilities.isPrivate(propertyDescriptor.getVisibility()) &&
+                   (propertyDescriptor.isConst() ||
```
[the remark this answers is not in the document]
mikaelpeltier: Indeed it is not needed, removed.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1986,6 +1993,10 @@ else if ((syntheticBackingField &&
         boolean skipPropertyAccessors;
 
         PropertyDescriptor originalPropertyDescriptor = DescriptorUtils.unwrapFakeOverride(propertyDescriptor);
+        boolean directAccessToGetterProperty = couldUseDirectAccessToProperty(propertyDescriptor, true, isDelegatedProperty, context,
```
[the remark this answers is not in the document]
mikaelpeltier: Done

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1997,8 +2008,9 @@ else if ((syntheticBackingField &&
             ownerDescriptor = propertyDescriptor;
         }
         else if (fieldAccessorKind == FieldAccessorKind.IN_CLASS_COMPANION || fieldAccessorKind == FieldAccessorKind.FIELD_FROM_LOCAL) {
-            boolean isInlinedConst = propertyDescriptor.isConst() && state.getShouldInlineConstVals();
-            skipPropertyAccessors = isInlinedConst || !isPrivateProperty || skipAccessorsForPrivateFieldInOuterClass;
+            // Do not use accessor 'access<property name>$cp' when the property can be accessed directly by his backing field.
```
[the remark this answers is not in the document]
mikaelpeltier: Done

compiler/backend/src/org/jetbrains/kotlin/codegen/JvmCodegenUtil.java
```diff
@@ -199,15 +199,27 @@ public static boolean couldUseDirectAccessToProperty(
         if (KotlinTypeMapper.isAccessor(property)) return false;
 
         CodegenContext context = contextBeforeInline.getFirstCrossInlineOrNonInlineContext();
-        // Inline functions can't use direct access because a field may not be visible at the call site
-        if (context.isInlineMethodContext()) {
+        // Inline functions or inline classes can't use direct access because a field may not be visible at the call site
+        if (context.isInlineMethodContext() || (context.getEnclosingClass() != null && context.getEnclosingClass().isInline())) {
```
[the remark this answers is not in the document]
mikaelpeltier: It is mandatory otherwise test `testReferToPropertyInCompanionObjectOfInlineClass` will failed
because when `couldUseDirectAccessToProperty` is used all conditions are satisfied but later the code accessing the private property into this test is moved into a class suffixed by `$Erased` and it is not longer legal to use a direct access if the access is move into another context.

compiler/backend/src/org/jetbrains/kotlin/codegen/JvmCodegenUtil.java
```diff
@@ -199,15 +199,27 @@ public static boolean couldUseDirectAccessToProperty(
...
             return false;
         }
 
+        DeclarationDescriptor propertyContainingDeclaration = property.getContainingDeclaration();
+        boolean isCompanionPropBackedInOuter = DescriptorUtils.isCompanionObject(propertyContainingDeclaration) &&
```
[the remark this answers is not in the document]
mikaelpeltier: I removed the code and update `isCallInsideSameClassAsDeclared` which seems a better way to solve the problem.

compiler/backend/src/org/jetbrains/kotlin/codegen/JvmCodegenUtil.java
```diff
@@ -199,15 +199,27 @@ public static boolean couldUseDirectAccessToProperty(
...
+                                               JvmAbi.isPropertyWithBackingFieldInOuterClass(property);
         if (!isCallInsideSameClassAsDeclared(property, context)) {
             if (!isDebuggerContext(context)) {
-                // Unless we are evaluating expression in debugger context, only properties of the same class can be directly accessed
-                return false;
+                // Property with backing field into an outer class could be used directly when access is done into the context of the outer
+                // class if the conditions further afield are respected, otherwise accessors must be used.
+                if (isCompanionPropBackedInOuter) {
```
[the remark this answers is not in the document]
mikaelpeltier: I removed the code and update `isCallInsideSameClassAsDeclared` which seems a better way to solve the problem.

compiler/backend/src/org/jetbrains/kotlin/codegen/JvmCodegenUtil.java
```diff
@@ -199,15 +199,27 @@ public static boolean couldUseDirectAccessToProperty(
...
+                    if (propertyContainingDeclaration.getContainingDeclaration() !=
```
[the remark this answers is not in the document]
mikaelpeltier: 1. Because `propertyContainingDeclaration` is `ClassDescriptor` and `context.getContextDescriptor()` is a `FunctionDescriptor` and what I want to do is to check if the access is done into the same class than the owner of the property. Moreover `propertyContainingDeclaration` is the class containing the declaration but not the outer class containing the backed field.
2. `isCallInsideSameClassAsDeclared` does not take into account property with backed field into the outer class, the check is done with where the property is declared and not where the property is generated, which is not the same place for a property that has a backed field into the outer class.

Thanks to your comments, I think that the cleaner way is to remove code that I added and to update `isCallInsideSameClassAsDeclared` to take into account property with backed field.

compiler/testData/asJava/lightClasses/compilationErrors/TraitClassObjectField.kt
mikaelpeltier: Done

compiler/testData/asJava/lightClasses/nullabilityAnnotations/ClassObjectField.kt
mikaelpeltier: Done

compiler/testData/codegen/box/companion/kt14258.kt
mikaelpeltier: Remove, you are right, the runtime behavior is already check by others tests.

compiler/testData/codegen/bytecodeText/companion/kt14258.kt
mikaelpeltier: Yes, the test was too complicated as your recommend I split it into several.

compiler/testData/codegen/bytecodeText/companion/kt14258.kt
mikaelpeltier: `fun box()` removed

compiler/testData/codegen/bytecodeText/constProperty/accessorsForPrivateConstants.kt
```diff
@@ -9,9 +9,12 @@ class Foo {
...
         LOCAL_PRIVATE
+        // Access to the property requires an invokestatic
         OUTER_PRIVATE
     }
 }
 
-// 2 INVOKESTATIC
\ No newline at end of file
+// 1 INVOKESTATIC
+// 1 PUTSTATIC
```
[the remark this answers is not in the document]
mikaelpeltier: I added the check on `GETSTATIC` instructions. They are two `GETSTATIC`, one into the foo method to access `LOCAL_PRIVATE` and one into the accessors to access `OUTER_PRIVATE`.

compiler/testData/codegen/bytecodeText/staticFields/classObjectSyntheticAccessor.kt
```diff
@@ -1,6 +1,8 @@
 class A {
     companion object {
-        private var r: Int = 1;
+        private var r: Int = 1
+            // Custom getter is needed, otherwise no need to generate getTest
```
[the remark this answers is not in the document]
mikaelpeltier: Test deleted

compiler/backend/src/org/jetbrains/kotlin/codegen/PropertyCodegen.java
mikaelpeltier: Totally agree with you, it was my first attempt but solving this problem seems intrusive into the compiler, thus I go back to concentrate on the fix for KT-14258. 
My plan is to file another bug because as you mentioned, I think that the compiler can do again a better job by separating what is did for `getter` and `setter`, but it seems easier to work on it in a separate way to avoid to block the resolution of all cases without custom getter and setter.
One of the problem comes from the fact that the access to a property will not be always do in the same way, it depends on the context, nevertheless this behavior is not compliant with the usage of `accessors` map into `CodegenContext` indeed the key does not take into account if getter or setter accessors are required.

udalov:
I've squashed the commits into one, combined the commit messages and merged manually. Thank you for the contribution!

PR #1538, opened 2018-02-26
Fix wrong example on the docs of @PurelyImplements

yukuku:


dzharkov:

PR #1539, opened 2018-02-27
KT-22970 Add explicit this intention/inspection missing for lambda invocation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22970

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ImplicitThisInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ImplicitThisInspection.kt
@@ -23,6 +23,7 @@ import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ImplicitThisInspection.kt
```diff
@@ -53,7 +54,10 @@ class ImplicitThisInspection : AbstractKotlinInspection() {
             val scope = reference.getResolutionScope(context) ?: return
 
             val descriptor = context[BindingContext.REFERENCE_TARGET, reference] as? CallableDescriptor ?: return
-            val receiverDescriptor = descriptor.extensionReceiverParameter ?: descriptor.dispatchReceiverParameter ?: return
+            val receiverDescriptor = descriptor.extensionReceiverParameter
+                    ?: descriptor.dispatchReceiverParameter
+                    ?: expression.getCallableDescriptor()?.extensionReceiverParameter
```
mglukhikh: Should be Ok. I'll check whether `descriptor` can be just changed with `getCallableDescriptor()` here.

mglukhikh:
Yes, such a replace is enough, so I did it. Thanks.

PR #1541, opened 2018-02-28
KT-22954 "Sort modifiers" quick fix works incorrectly when method is annotated

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22954#u=1519128944445


mglukhikh:
I think it's Ok. I'll try to find easier solution and merge it.

PR #1544, opened 2018-03-01
KT-20282 'Move statement up' works incorrectly for statement after 'finally' block if 'try' block contains closure

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20282


goodwinnk:
Thank you and sorry for the delay in review!

I have merged your commit as 9d6b36c2496c6b3816166b5185e24f4a114cb357.

You should have probably converted the file to Kotlin first :)

PR #1545, opened 2018-03-01
Long.toBigDecimal(): avoid new BigDecimal instance

elifarley:
...by calling BigDecimal.valueOf(this), which doesn't create a new BigDecimal instance if the argument falls in the range [0..10].
(there's a `private static final BigDecimal[] zeroThroughTen` cache in BigDecimal.java with 11 BigDecimal instances)

Here's the bytecode diff for the Int receiver:
```patch
--- inline fun Int.toBigDecimal()
+++ inline fun Int.toBigDecimal()
@@ -4,10 +4,12 @@
   @Lkotlin/internal/InlineOnly;() // invisible
    L0
     LINENUMBER 92 L0
-    NEW java/math/BigDecimal
+    ILOAD 0
+    I2L
+    INVOKESTATIC java/math/BigDecimal.valueOf (J)Ljava/math/BigDecimal;
     DUP
-    ILOAD 0
-    INVOKESPECIAL java/math/BigDecimal.<init> (I)V
+    LDC "BigDecimal.valueOf(this.toLong())"
+    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkExpressionValueIsNotNull (Ljava/lang/Object;Ljava/lang/String;)V
     ARETURN
    L1
     LOCALVARIABLE $receiver I L0 L1 0
```

---
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/kotlin/1545)

### DIFF
--- a/libraries/stdlib/src/kotlin/util/BigDecimals.kt
+++ b/libraries/stdlib/src/kotlin/util/BigDecimals.kt
@@ -105,7 +105,7 @@ public inline fun Int.toBigDecimal(mathContext: MathContext): BigDecimal = BigDe
  */
 @SinceKotlin("1.2")
 @kotlin.internal.InlineOnly
-public inline fun Long.toBigDecimal(): BigDecimal = BigDecimal(this)
+public inline fun Long.toBigDecimal(): BigDecimal = BigDecimal.valueOf(this)
 
 /**
  * Returns the value of this [Long] number as a [BigDecimal].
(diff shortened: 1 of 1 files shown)


elifarley:
I'll change method `Int.toBigDecimal()` as well, if the current change looks ok.

libraries/stdlib/src/kotlin/util/BigDecimals.kt
```diff
@@ -89,7 +89,7 @@ public inline operator fun BigDecimal.dec(): BigDecimal = this.subtract(BigDecim
  */
 @SinceKotlin("1.2")
 @kotlin.internal.InlineOnly
-public inline fun Int.toBigDecimal(): BigDecimal = BigDecimal(this)
+public inline fun Int.toBigDecimal(): BigDecimal = BigDecimal.valueOf(this)
```
ilya-g: There's no such overload of `valueOf`. Please ensure it compiles.

elifarley: Sorry, fixed that.

PR #1548, opened 2018-03-03
KT-21413 Missing inspection: parentheses can be deleted when the only constructor parameter is a function not existing

tommykw:
https://youtrack.jetbrains.com/issue/KT-21413


idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantConstructorParenthesesInspection.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.psi.*
+
+class RemoveRedundantConstructorParenthesesInspection : AbstractKotlinInspection() {
```
mglukhikh: I suspect it should be done other way. First of all, it's quite unclear why only constructors should be handled here, and why only case with one value parameter should be considered. And second, there is `MoveLambdaOutsideParenthesesIntention` which solves the same task. Probably the best we can do here is to convert it to the inspection using `AbstractApplicabilityBasedInspection`. I checked, and intention works in your case too.

idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantConstructorParenthesesInspection.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return callExpressionVisitor { expression ->
+            val valueArguments = expression.valueArgumentList ?: return@callExpressionVisitor
+            if (valueArguments.arguments.size != 1) return@callExpressionVisitor
+            valueArguments.arguments.mapNotNull { it.getArgumentExpression() }.singleOrNull() as? KtLambdaExpression ?: return@callExpressionVisitor
+
+            holder.registerProblem(
+                    expression,
```
mglukhikh: I think we should report it on value argument and not on the whole call expression.

tommykw:
Thank you for your review. Could you review this, again?

mglukhikh:
@tommykw Now it looks good for me, thank you.

PR #1550, opened 2018-03-05
MPP wizard enhancements

mglukhikh:
https://youtrack.jetbrains.com/issue/KT-23097


idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleMultiplatformWizardStep.kt
```diff
@@ -41,6 +41,8 @@ class KotlinGradleMultiplatformWizardStep(
     private val wizardContext: WizardContext
 ) : ModuleWizardStep() {
 
+    private val rootIsCommonCheckBox: JCheckBox =
```
yole: I don't think people will actually understand what this means. I'd change this into two radio buttons: "Create top-level empty module with common and platform modules as children" and "Make platform modules children of common module".

ilya-g: But in either case the root project should not be common.

PR #1551, opened 2018-03-06
KT-23109 "Remove redundant 'if' statement" inspection breaks code with labeled return

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23109


mglukhikh:
Merged, thank you.

PR #1553, opened 2018-03-06
KT-23126 Intention "Add labeled return to last expression in lambda" shouldn't work for `suspend` lambdas

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23126


mglukhikh:
Merged, thank you

PR #1554, opened 2018-03-07
KT-14270 Intentions "Add/Remove braces" should be applied to the statement where caret is if there several nested statements one into another

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14270


mglukhikh:
I merged it but have some doubts about suggesting "Add braces" and "Remove braces" at the same point. Especially if they're suggested as "Add braces to if" and "Remove braces to if" (for the first, upper if is in mind). Not quite sure how it can be done best way.

PR #1556, opened 2018-03-09
KT-22949 NPE on conversion of `run`/`apply` with explicit lambda signature to `let`/`also`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22949
https://youtrack.jetbrains.com/issue/KT-22950


mglukhikh:
Merged, thank you.

PR #1557, opened 2018-03-09
KT-22860 - Added `Add Annotation Target` support to search for Annotations in Java source files

fg123:
https://youtrack.jetbrains.com/issue/KT-22860

This is my first PR so I'm not 100% sure if everything is done right. I've added a test but I don't think the current test generating system supports multi-file quick fix tests. (There are "multifileQuickFix" but I believe that is to generate a test from each file, rather than a test with multiple files included as a project). 

I tried adding it into the multi-file intention tests but `AddAnnotationTargetFix` does not have a no parameter constructor so the reflective constructor invocation failed. Perhaps we need to add a new type of generated test?

I _did_ run it to verify that it was generating the code, so the test should pass if I can figure out where/how to generate it.


mglukhikh:
Mostly it's good, thank you. In future, I recommend you to squash all commits before creating PR. Also, I do not recommend to declare in compiler front-end functions which are used entirely in IDE. I'll fix this details myself.

mglukhikh:
@fg123 Please pay some attention to the resulting commit: https://github.com/JetBrains/kotlin/commit/95392121808d5ab22dd56e56a25adbf8f1c1f944. In future, don't forget to regenerate automatic tests (Generate All Tests run configuration). Also, your tests should belong to quick-fix group (see `withJava.before.Main.kt`, `withJava.after.kt`, `withJava.before.UsageJava.java` in the resulting commit). Anyway, thank you for your contribution.

fg123:
@mglukhikh Thanks so much for your help! :+1:

PR #1559, opened 2018-03-09
Add support for nullability annotations in androidx package.

JakeWharton:


dzharkov:
Merged manually with minor fixes.
FYI: For new generated tests you can run `./gradlew generateTests -x :ultimate:generateTests`

PR #1561, opened 2018-03-10
Standardize and improve descriptions of intentions/inspections

ltongues:


mglukhikh:
Thank you, so much work. We accept you change.

PR #1563, opened 2018-03-12
KT-14788 "Unwrap/Remove" action in Kotlin suggest line removal instead of unwrapping statement under the cursor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14788


goodwinnk:
Thank you! Your commit was merged manually as https://github.com/JetBrains/kotlin/commit/c1f43558b98b383220dd75f3c088d96387c548da.

PR #1564, opened 2018-03-12
KT-23215 "Add function to supertype" quick fix works incorrectly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23215


mglukhikh:
Thank you, merged.

PR #1565, opened 2018-03-12
Is the Interface with out the f a type?

jeis2497052:
This seems like a bug ?  Can a subject matter expert review please ?


ltongues:
I'm not an expert, but this is definitely a typo. It's still an UNRESOLVED_REFERENCE, but for a different reason :)

jeis2497052:
Thanks, it didn't look correct to me.

HTH

bashor:
Merged manually.

Sorry for the delay and thank you for your contribution!

PR #1567, opened 2018-03-13
KT-23227 "Add annotation target" quick fix is not suggested for `field:` use-site target

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23227


mglukhikh:
Merged, thank you.

PR #1568, opened 2018-03-14
KT-12226 "Convert concatenation to template" does not process dollar sign as a Char

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12226


mglukhikh:
Thank you, merged

PR #1569, opened 2018-03-14
KT-19871 Add intentions for specifying use-site targets for an annotation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19871


mglukhikh:
Tested & merged it, thank you!

PR #1570, opened 2018-03-15
KT-20357 Add sample code for Regex.find

shiraji:


ilya-g:
I've reworked the sample a bit and merged it, thanks!

PR #1574, opened 2018-03-22
KT-23152 Incorrect warning 'Redundant semicolon' for a semicolon placed before a call of lambda extension function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23152


mglukhikh:
Merged, thank you

PR #1576, opened 2018-03-22
KT-23045 AE “Failed to create expression from text” on concatenating string with broken quote mark char literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23045
https://youtrack.jetbrains.com/issue/KT-23046


mglukhikh:
Merged, thank you

PR #1578, opened 2018-03-23
Optional plugin dependency for kotlin gradle plugin 'java' subsystem dependent features

vladsoroka:
IntelliJ gradle plugin 'java' subsystem dependant features have been extracted into a separate module. (Since `182+` versions)
The change allows to install the gradle support into non-java IDEs like CLion.

The 'java'/'groovy' dependant features of the kotlin gradle integration should be defined using related optional dependency.
It is required for the Kotlin/Native/Gradle support of the CLion.


4u7:
Merged manually

PR #1579, opened 2018-03-23
KT-23137 Intention for converting to block comment and vise versa

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23137


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertBlockCommentToLineCommentIntention.kt
```diff
@@ -0,0 +1,49 @@
...
+package org.jetbrains.kotlin.idea.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiComment
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiWhiteSpace
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.KtPsiFactory
+
+class ConvertBlockCommentToLineCommentIntention : SelfTargetingIntention<PsiComment>(
```
mglukhikh: Should be probably not applicable to documentation comments:

```
/**
 * ...
 */
```

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertBlockCommentToLineCommentIntention.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun isApplicableTo(element: PsiComment, caretOffset: Int): Boolean {
+        return element.isBlockComment()
+    }
+
+    override fun applyTo(element: PsiComment, editor: Editor?) {
+        val psiFactory = KtPsiFactory(element)
+
+        val prevSibling = element.prevSibling
+        val indent = if (prevSibling is PsiWhiteSpace) {
+            val space = prevSibling.text.reversed().takeWhile { it == ' ' }
```
mglukhikh: What about `'\t'`? Of course it's rare case but it's possible

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertBlockCommentToLineCommentIntention.kt
mglukhikh: it's enough to check that beginning of comment is `/**`

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLineCommentToBlockCommentIntention.kt
```diff
@@ -0,0 +1,67 @@
...
+        return element.isEndOfLineComment()
+    }
+
+    override fun applyTo(element: PsiComment, editor: Editor?) {
+        var firstComment = element
+        while (true) {
+            firstComment = firstComment.prevComment() ?: break
+        }
+
+        val indent = (firstComment.prevSibling as? PsiWhiteSpace)?.text?.reversed()?.takeWhile { it == ' ' } ?: ""
```
mglukhikh: Here tabulation is also possible

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertBlockCommentToLineCommentIntention.kt
mglukhikh: Please check how this and other intention work with situation like below (multiple new-lines between comment and declaration).

```
/* 
   block comment 
*/


fun foo() {}
```

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertBlockCommentToLineCommentIntention.kt
t-kameyama: I added `docComment.kt` test. `isBlockComment()` function returns false, if element is KDoc.

t-kameyama:
Fixed

mglukhikh:
Thank you, merged.

PR #1580, opened 2018-03-23
KT-23303 "Might be const" inspection does not check explicit type specification

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23303


mglukhikh:
Looks great

mglukhikh:
Merged, thank you.

PR #1581, opened 2018-03-24
KT-23321 Intention to move type to separate file shouldn't be available for sealed classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23321


mglukhikh:
Merged (with minor polishing), thank you

PR #1582, opened 2018-03-24
KT-23320 Quick fix to add constructor invocation doesn't work for sealed classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23320

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/SuperClassNotInitialized.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/SuperClassNotInitialized.kt
@@ -24,6 +24,7 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ConstructorDescriptor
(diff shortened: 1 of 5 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/SuperClassNotInitialized.kt
```diff
@@ -61,21 +62,29 @@ object SuperClassNotInitialized : KotlinIntentionActionsFactory() {
 
         val superClass = (type.constructor.declarationDescriptor as? ClassDescriptor) ?: return emptyList()
         val classDescriptor = classOrObjectDeclaration.resolveToDescriptorIfAny(BodyResolveMode.FULL) ?: return emptyList()
-        val constructors = superClass.constructors.filter { it.isVisible(classDescriptor) }
+        val inSameDeclaration = superClass.containingDeclaration == classDescriptor.containingDeclaration
```
mglukhikh: It's not enough. Both declarations must be either top-level or class inside super-class (the last case is covered by isVisible). Otherwise you cannot inherit sealed class.

mglukhikh: I fix it myself

PR #1583, opened 2018-03-24
KT-23377 Simplify boolean expression produces incorrect results when mixing nullable and non-nullable variables

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23377


mglukhikh:
Merged, thank you

PR #1584, opened 2018-03-24
KT-23133 "Remove redundant calls of the conversion method" wrongly shown for Boolan to Int conversion

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23133


mglukhikh:
Merged, thank you

PR #1585, opened 2018-03-24
KT-23346 Lift Assignment quick fix incorrectly processes block assignments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23346


mglukhikh:
Looks well

mglukhikh:
Merged, thank you. NB: it's much better to separate reformat from valuable actions (do them in different commits).

PR #1587, opened 2018-03-25
KT-5206 Support code folding for standard library collection factory functions.

aerovulpe:
https://youtrack.jetbrains.com/issue/KT-5206

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/KotlinFoldingBuilder.kt
@@ -22,21 +22,70 @@ import com.intellij.lang.folding.CustomFoldingBuilder
 import com.intellij.lang.folding.FoldingDescriptor
 import com.intellij.openapi.editor.Document
 import com.intellij.openapi.project.DumbAware
+import com.intellij.openapi.project.DumbService
 import com.intellij.openapi.util.TextRange
 import com.intellij.openapi.util.text.StringUtil
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiWhiteSpace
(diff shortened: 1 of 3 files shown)


goodwinnk:
Thank you for the first contribution to Kotlin project! 
I have accepted your commit manually: https://github.com/JetBrains/kotlin/commit/3a83983bfcc2f96acf1867c5ff3ebb06c83d82a9

I'm sorry for the long review process. I'd decided that it's unfair to ask for additional changes after almost a month from the pull request so I did review commit myself: https://github.com/JetBrains/kotlin/commit/b6699d7522fc6844a57ab2b6fd5950de1a0ed6d5.

Your commit was great and fixed the issue, but  there're still several important things that are worth mentioning for future pull-requests:
 - `analyze()` call can be very time-consuming, and should be performed when all other light-weight checks (psi, context) are done.
 - We're trying to avoid direct dealing with `BindingContext`. Usages of `KtReference.resolveToDescriptors()`, `KtReference.resolve()` or methods from `resolutionApi.kt` are preferable.
 - It's better to have tests for different non-trivial cases that you've faced during writing the code.

I have also changed logic of the fix a bit. An ideal solution would be to enable folding arguments for all single-parameter vararg calls. But I've decided to leave only predefined set of names for now, to get a quick check and avoid analyze in folding feature for most calls. We will see if there are requests for other functions.

Again, thank you for the contribution!

aerovulpe:
Thanks Nikolay! I've taken a look at your commit - a lot to learn from there. 

Yeah, the single-parameter vararg approach seems ideal and I had considered it; but wasn't sure what performance cost it might have had. I'll also keep an eye out on YouTrack for further requests.

I'm happy to contribute. Thank you for all the hard work you and your team puts into making Kotlin such an awesome language!

PR #1590, opened 2018-03-27
KT-23435 False redundant companion reference inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23435


mglukhikh:
Merged, thank you

PR #1592, opened 2018-03-31
More correct NOTHING_TO_INLINE description

Miha-x64:
Strictly speaking, they are not 'lambda parameters', they are 'functional parameters' or 'parameters of functional types'.
Sorry for my pedantry.


udalov:
Thanks, but I'm afraid this makes the message unnecessarily long. Maybe we should drop the "i.e. <...>" part?

Miha-x64:
Ideas:
* First part. 'expected' and 'can be' are both not very strong, we may omit one:
  * Expected performance impact of inlining ''{0}'' is insignificant.
  * Expecting insignificant performance impact of inlining ''{0}''.
  * Performance impact of inlining ''{0}'' can be insignificant.
* Second part.
    * Inlining works best for higher-order functions. — minimalistic
    * Inlining works best for functions with functional parameters. (like () -> Unit). — exemplified
    * Inlining works best for functions which accept/take/consume functions. — humanistic, not technical, not sure in terminology

PR #1596, opened 2018-04-03
Fix two exceptions in "Convert concatenation to template"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23045#comment=27-2793220
https://youtrack.jetbrains.com/issue/KT-23046#comment=27-2793217


Montesdodgers13:
Good work

mglukhikh:
I'm afraid I already fixed it different way (not yet in master). I will think once more which way is better and push the chosen fix into master, and may be 1.2.40 too.

mglukhikh:
Own fix used

PR #1598, opened 2018-04-05
KT-13420 Extend Selection: lambda: whole literal with braces is selected after parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13420


goodwinnk:
Merged as  https://github.com/JetBrains/kotlin/commit/5aa0b7d2aaca26defb4c173c755ae02dd95b2a93, thank you!

PR #1599, opened 2018-04-05
KT-12330 Improve bytecode for data class equals/hashCode methods

mikaelpeltier:
- Do not longer call hashCode on java/lang/Object but do this
on statically checked type.
- Do not longer call Intrinsics.areEqual but use directly
equals method on statically checked type.

Fix of https://youtrack.jetbrains.com/issue/KT-12330


compiler/backend/src/org/jetbrains/kotlin/codegen/ImplementationBodyCodegen.java
```diff
@@ -574,8 +575,20 @@ else if (asmType.getSort() == Type.DOUBLE) {
...
                 }
                 else {
-                    StackValue value = genEqualsForExpressionsOnStack(KtTokens.EQEQ, StackValue.onStack(asmType), StackValue.onStack(asmType));
-                    value.put(Type.BOOLEAN_TYPE, iv);
+                    if (isPrimitive(asmType)) {
+                        StackValue value = StackValue.cmp(KtTokens.EQEQ, asmType, StackValue.onStack(asmType), StackValue.onStack(asmType));
+                        value.put(Type.BOOLEAN_TYPE, iv);
+                    } else {
+                        if (DescriptorUtils.isInterface(propertyDescriptor.getType().getConstructor().getDeclarationDescriptor())) {
+                            iv.invokeinterface(asmType.getInternalName(), "equals",
```
udalov: I think it's fine to continue calling `invokevirtual java/lang/Object.equals` for interfaces, since it's not immediately clear that `invokeinterface I.equals` is better. So I propose to replace this if with:
```
Type owner =
        DescriptorUtils.isInterface(propertyDescriptor.getType().getConstructor().getDeclarationDescriptor())
        ? AsmTypes.OBJECT_TYPE
        : asmType;
iv.invokevirtual(
        owner.getInternalName(), "equals", Type.getMethodDescriptor(Type.BOOLEAN_TYPE, AsmTypes.OBJECT_TYPE), false
);
```

compiler/backend/src/org/jetbrains/kotlin/codegen/AsmUtil.java
```diff
@@ -516,7 +516,12 @@ static void genHashCode(MethodVisitor mv, InstructionAdapter iv, Type type, JvmT
             }
         }
         else if (type.getSort() == Type.OBJECT) {
-            iv.invokevirtual("java/lang/Object", "hashCode", "()I", false);
+            if (isInterface) {
```
udalov: Similarly to the `equals` case, I propose to replace this with something like
```
iv.invokevirtual((isInterface ? AsmTypes.OBJECT_TYPE : type).getInternalName(), "hashCode", "()I", false);
```

compiler/testData/codegen/box/dataClasses/kt12330.kt
```diff
@@ -0,0 +1,14 @@
+// WITH_RUNTIME
```
udalov: Does this box test check something which is not covered by current tests in `codegen/box/dataClasses`? If no, I propose to delete it

udalov:
Merged manually. Thanks!

PR #1600, opened 2018-04-05
KT-20357 Add sample code Strings.toUpperCase

katsutomu:


libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -79,4 +79,10 @@ class Strings {
         assertPrints(byteArray.toString(charset), "Hello")
     }
 
+    @Sample
+    fun stringToUpperCase() {
+        assertPrints("Hello".toUpperCase(), "HELLO")
+        assertPrints("Hello".toUpperCase(Locale.ENGLISH), "HELLO")
```
ilya-g: I wonder, is there such an example that shows why locale could be used for toUpper/LowerCase? I mean so that it would result in a different result compared to no locale overload.

katsutomu: @ilya-g 
Thank you for review.
I understood that the following is unnecessary.
After excluding the following points,I will request a review again.
> assertPrints("Hello".toUpperCase(Locale.ENGLISH), "HELLO")

katsutomu:
@ilya-g 
Fixed review.
I would appreciate it if you could review again.

ilya-g:
Thanks, but we already have samples for `String.toUpperCase` and `String.toLowerCase`, see PR #1762. Only the jvm-specific overloads with `Locale` are left uncovered with samples.
If you're willing to provide them, please note, that a sample should illustrate why one would like to provide a `Locale` when calling `toUpperCase`/`toLowerCase`, for example by showing a difference between calling them with and without `Locale` argument.

Feel free to reopen this PR or create a new one, if you're going to provide samples for these overloads.

PR #1601, opened 2018-04-05
Add context to generateSynthetic* in SyntheticResolveExtension

jimgoog:
Previously, `generateSyntheticClasses` took in a `LazyClassContext`, but `generateSyntheticMethods` and `generateSyntheticProperties` did not.  We found it desirable to have the a binding context available when generating synthetic methods and properties.  This change makes the API more consistent by making context available in all three generate methods.

cc @elizarov @sandwwraith @yanex @erokhins


jimgoog:
Yep, makes sense.  My original draft actually passed in `BindingContext` (which is substantially less powerful than `LazyClassContext`), but then I realized that `generateSyntheticClasses` was passing in `LazyClassContext`, so I did the same for consistency.  `BindingContext` would be perfectly adequate for my use case, as the only thing I do with `LazyClassContext` is pull off the `BindingContext`.

jimgoog:
Just let me know what you guys decide; I'm happy to update accordingly.

sandwwraith:
The reason `generateSyntheticClasses` uses `LazyClassContext` is that `SyntheticClassOrObjectDescriptor` needs it as a constructor argument. Since you don't have to create synthetic classes (I suppose) in `generateSyntheticMethods`, etc imho it is okay to pass only context to them and there will be no inconsistency.

jimgoog:
Ok, sounds like you guys are leaning in favor of `BindingContext`.  PR has been updated accordingly.

compiler/frontend/src/org/jetbrains/kotlin/resolve/extensions/SyntheticResolveExtension.kt
```diff
@@ -62,17 +63,19 @@ interface SyntheticResolveExtension {
                 // todo revert
                 override fun generateSyntheticMethods(
                     thisDescriptor: ClassDescriptor, name: Name,
+                    ctx: BindingContext,
```
udalov: Would be nice to name this `bindingContext` or `context`, since we don't usually use abbreviations for identifier names in our project

erokhins: I will rename it.

PR #1602, opened 2018-04-06
KT-23634 'Add use-site target' intention drops annotation arguments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23634


mglukhikh:
Merged, thank you

PR #1603, opened 2018-04-06
KT-23608 AE “Failed to create expression from text” after applying quick fix “Convert too long character literal to string”

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23608


mglukhikh:
Merged, thank you (may be other similar cases possible, but yet I have not enough time to check it attentively).

PR #1604, opened 2018-04-06
KT-23620 False positive `Redundant Companion reference` on calling object from companion

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23620


mglukhikh:
Merged, thank you

PR #1605, opened 2018-04-06
KT-23419 Intention to replace vararg with array and vice versa

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23419


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertArrayParameterToVarargIntention.kt
```diff
@@ -0,0 +1,53 @@
...
+        if (element.isVarArg) return false
+
+        val type = element.descriptor?.type ?: return false
+        return when {
+            KotlinBuiltIns.isPrimitiveArray(type) ->
+                true
+            KotlinBuiltIns.isArray(type) -> {
+                val typeArguments = typeReference.typeElement?.typeArgumentsAsTypes
+                val typeProjection = typeArguments?.firstOrNull()?.parent as? KtTypeProjection
+                typeProjection?.hasModifier(KtTokens.IN_KEYWORD) == false
```
mglukhikh: You are right about `in`, but we can also have problems when no variance is stated. E.g.

```
fun bar(arr: Array<String>) {
    arr[0] = "" // arr cannot be made vararg here
}
```

With `out` variance, this assignment is not possible, but with no variance, we have to check function body for a possible assignment to array elements :(

idea/resources/intentionDescriptions/ConvertArrayParameterToVarargIntention/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This intention converts an array parameter to a vararg parameter.
```
mglukhikh: Here and below, please select `vararg` with bold.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertArrayParameterToVarargIntention.kt
```diff
@@ -0,0 +1,53 @@
...
+            else ->
+                false
+        }
+    }
+
+    override fun applyTo(element: KtParameter, editor: Editor?) {
+        val typeReference = element.getChildOfType<KtTypeReference>() ?: return
+        val type = element.descriptor?.type ?: return
+        val newType = (if (KotlinBuiltIns.isPrimitiveArray(type)) {
+            PrimitiveType.values().firstOrNull { it.arrayTypeName == type.nameIfStandardType }?.typeName?.asString()
```
mglukhikh: Probably `getPrimitiveArrayElementType` from built-ins should work here.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertArrayParameterToVarargIntention.kt
mglukhikh: I think that for performance reasons it's not a good idea to implement such a check. So I suggest to change intention message in case no variance is stated, adding something like `(may break code)` at the end.

t-kameyama:
Fixed

mglukhikh:
Thank you, merged

PR #1608, opened 2018-04-09
KT-7822 Convert foreach to for loop should place caret on the variable declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-7822


mglukhikh:
Thank you, merged

PR #1609, opened 2018-04-09
KT-23617 Add inspection: redundant internal in local anonymous object / class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23617


t-kameyama:
Fixed

PR #1610, opened 2018-04-09
KT-16085 Inspection "main should return Unit"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16085


idea/resources/inspectionDescriptions/MainFunctionReturnUnit.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection reports a main function should return <b>Unit</b> type.
```
mglukhikh: I'd rephrase this a little bit, like "a main function with incorrect return type (should be Unit)" or something like.

idea/src/org/jetbrains/kotlin/idea/inspections/MainFunctionReturnUnitInspection.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.namedFunctionVisitor
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+class MainFunctionReturnUnitInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return namedFunctionVisitor(fun(function: KtNamedFunction) {
+
+            val descriptor = function.descriptor as? FunctionDescriptor ?: return
```
mglukhikh: Please perform some PSI-based check before (e.g. compare function name with `main`). Converting to descriptor in the very first line of inspection can be performance consuming.

mglukhikh:
Thank you, merged

PR #1612, opened 2018-04-09
Removed Unused Braces For Companion Objects (String, Enum)

ISkylakeS:
As far as we know, braces aren't needed for empty classes, constructors, interfaces, objects, etc. So why there are braces for companion objects of String and Enum classes?


udalov:
If you'd try to compile these files after the change, you'd know the answer, which is: according to the Kotlin grammar, the next word token after the words `companion object` is parsed as its name, and so the word `public` a few lines below would actually be parsed as the name of the companion object, not as the function modifier. I think there would be a parser error about the fact that one declaration (`companion object public`) is not separated from the next (`fun ...`) with `;`.

PR #1614, opened 2018-04-10
Added isNotNullOrEmpty() and isNotNullOrBlank()

JonahStarling:
Added `isNotNullOrEmpty()` and `isNotNullOrBlank()` to the `Strings.kt`. The functions `isBlank()` and `isEmpty()` already have the opposite state functions implemented and I found that it would be useful to also have the opposite state for the two functions `isNullOrEmpty()` and `isNullOrBlank()`.


JonahStarling:
Got it. That makes sense. I feel like it should be obvious that it is `isNot(NullOrEmpty)` but I can see where the confusion would arise.

PR #1615, opened 2018-04-11
KT-14556 Quickfix to suggest use of spread operator does not work with mapOf

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14556


mglukhikh:
Thank you, merged.

PR #1616, opened 2018-04-11
KT-23721 Emit an error when 'tools.jar' is not into plugin classpath

mikaelpeltier:
- Since warnings are discarded when there are others errors, replace the
current warning by an error to always display the message and inform the
user that its annotation processor is badly configured.

Fix of https://youtrack.jetbrains.com/issue/KT-23721

### DIFF
--- a/plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/Kapt3Plugin.kt
+++ b/plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/Kapt3Plugin.kt
@@ -229,7 +229,7 @@ class Kapt3ComponentRegistrar : ComponentRegistrar {
         try {
             Class.forName(JAVAC_CONTEXT_CLASS)
         } catch (e: ClassNotFoundException) {
-            logger.warn("'$JAVAC_CONTEXT_CLASS' class can't be found ('tools.jar' is absent in the plugin classpath). Kapt won't work.")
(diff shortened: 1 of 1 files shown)


plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/Kapt3Plugin.kt
```diff
@@ -229,7 +229,7 @@ class Kapt3ComponentRegistrar : ComponentRegistrar {
         try {
             Class.forName(JAVAC_CONTEXT_CLASS)
         } catch (e: ClassNotFoundException) {
-            logger.warn("'$JAVAC_CONTEXT_CLASS' class can't be found ('tools.jar' is absent in the plugin classpath). Kapt won't work.")
+            logger.error("'$JAVAC_CONTEXT_CLASS' class can't be found ('tools.jar' is absent in the plugin classpath). Kapt won't work.")
```
zarechenskiy: Maybe it's better to use strong warning here? See `CompilerMessageSeverity.STRONG_WARNING`

yanex: Well, it's the other possible solution, but annotation processing phase will be skipped then, and the user likely get an error such as "Unresolved reference <...>" later.
So I think this is closer to error.

PR #1618, opened 2018-04-12
KT-12168 "Remove explicit type specification" intention produce incompilable code in case of function type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12168


mglukhikh:
I think the same is possible for functions with expression body

mglukhikh:
Extended for functions with expression bodies and merged.

PR #1620, opened 2018-04-13
KT-23511 "Remove parameter" quick fix makes generic function call incompilable when type could be inferred from removed parameter only

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23511


asedunov:
Merged manually. Thnak you!

PR #1623, opened 2018-04-13
KT-23559: Fix hint text for ReplaceWithOperatorAssignmentInspection

Mon-chi:


Mon-chi:
https://youtrack.jetbrains.com/issue/KT-23559

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithOperatorAssignmentInspection.kt
```diff
@@ -55,7 +55,7 @@ class ReplaceWithOperatorAssignmentInspection : AbstractApplicabilityBasedInspec
     override val defaultFixText = "Replace with operator-assignment"
 
     override fun fixText(element: KtBinaryExpression) =
-            "Replace with '${element.operationReference.operationSignTokenType?.value}='"
+        "Replace with '${(element.right as KtBinaryExpression).operationReference.operationSignTokenType?.value}='"
```
mglukhikh: Just for safety, I'd use `as?` instead of `as` here

mglukhikh: Also, some test should be provided. Please look at `idea/testData/inspectionsLocal/replaceWithOperatorAssignment` for examples. You should use "Generate All Tests" run configuration to regenerate automatic tests (`LocalInspectionTestGenerated` in our case).

mglukhikh: Also I recommend to run other local inspection tests to be sure that nothing is broken

Mon-chi:
I'm sorry for late reply. I was too busy to work on this.

What should be happen when element.right !as ktBinaryExpression?
And I can't use Generate All Tests due to the bug with Windows. What should I do?

mglukhikh:
> What should be happen when element.right !as ktBinaryExpression?

Nothing special, just insert safe calls in this expression (fixText) everywhere.

mglukhikh:
> And I can't use Generate All Tests due to the bug with Windows. What should I do?

Yes, we had an error at this point with a message like "too long command line". However, it's already fixed. You can rebase your branch on master and everything should work properly (I use Windows myself).

Mon-chi:
Thank you for your support. I've updated.

idea/testData/inspectionsLocal/replaceWithOperatorAssignment/validAddition.kt
```diff
@@ -0,0 +1,4 @@
+fun foo() {
```
mglukhikh: This test does not test your change yet. You should add comment above (to both this file and after file):

```
// FIX: expected fix text
```

to check that fix text is correct.

Mon-chi:
Updated. Is this alright?

mglukhikh:
Yes, many thanks. PR accepted, I'll merge it this week.

mglukhikh:
Merged, thank you

PR #1624, opened 2018-04-14
KT-23798 Deprecate Preconditions.error(msg) for illegalState(msg)

patrickcousins:
This is a suggestion to rename the `error(message)` function in Preconditions.kt. 

This PR is a first step fix for https://youtrack.jetbrains.com/issue/KT-23798 by deprecating the error method and creating a new, nearly identical method named `illegalState(message: Any)`

`error(message)` simply throws an `IllegalStateException(message)` with the message passed to it. 

However, it pollutes the global namespace and prevents others from creating extension methods for their projects (such as sending an error to Android `Log.e`, or their own custom logging / error handling). 

`error(message)`, as it is currently named, can easily be imported accidentally.  Because it throws a runtime exception, it could cause an unexpected crash.

Suggest removing it entirely or renaming it to something more descriptive like: `throwIllegalState(message: Any)` or `illegalState(message: Any)`

(This is my first PR here so please feel free to point me wherever I need to be if I missed anything)

### DIFF
--- a/libraries/stdlib/samples/test/samples/misc/preconditions.kt
+++ b/libraries/stdlib/samples/test/samples/misc/preconditions.kt
@@ -10,7 +10,7 @@ class Preconditions {
     fun failWithError() {
         val name: String? = null
 
-        assertFailsWith<IllegalStateException> { val nonNullName = name ?: error("Name is missing") }
+        assertFailsWith<IllegalStateException> { val nonNullName = name ?: illegalState("Name is missing") }
     }
 
     @Sample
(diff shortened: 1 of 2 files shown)


ilya-g:
This PR seems to be too quick, given that your ultimate goal is to remove `error` function entirely following the deprecation cycle, the issue must be discussed first by the language committee.

Therefore I'm closing this for now, but I'll link this PR there in case if the issue would be approved.

patrickcousins:
Sounds good @ilya-g Thank you, apologies for jumping the gun, I just wanted to have this is a way to generate the discussion.  Thanks for considering it!

PR #1625, opened 2018-04-14
KT-23753 "Remove variable" quick fix should not remove comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23753


mglukhikh:
Merged, thank you

PR #1626, opened 2018-04-14
KT-20357 Add sample for Iterable.zip

kakueki61:


kakueki61:
@ilya-g Thank you for your review!
I changed the templates and generates the code.

ilya-g:
Thanks, I've added the remaining samples for Strings and Sequences and merged it manually.

PR #1628, opened 2018-04-16
KT-11740 Invert if condition intention should not remove line breaks

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11740


mglukhikh:
Looks good, thank you. May be I will polish code a little bit

mglukhikh:
Just as comment: it's generally a **good idea** to do reformat in a separate commit, as done in this PR. It makes much easier to understand purpose and changes of each commit.

PR #1630, opened 2018-04-17
KT-23801 "Convert to constructor" (IntelliJ) quick fix uses wrong use-site target for annotating properties

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23801


mglukhikh:
Merged, thank you

PR #1631, opened 2018-04-17
Stable order of synthesized bridges.

denisvnukov:
In some cases when several bridge methods are generated, their order
is undetermenistic. For example for class ClassLowerWithContext in the
following example

```
interface IrElement
class IrClassContext

interface IrElementVisitor<out R, in D> {
    fun visitElement(element: IrElement, data: D): R
}

interface IrElementTransformer<in D> : IrElementVisitor<IrElement, D> {
    override fun visitElement(element: IrElement, data: D): IrElement =
            element.also { throw RuntimeException() }
}

abstract class ClassLowerWithContext : IrElementTransformer<IrClassContext?>
```

kotlin compiler synthesizes two additional bridges:

```
  public IrElement visitElement(IrElement, IrClassContext);
  [bridge] public Object visitElement(IrElement, Object);
  [bridge] public IrElement visitElement(IrElement, Object);
```

Unfortunately the behavior is not deterministic and not easy to reproduce.

### DIFF
--- a/compiler/backend-common/src/org/jetbrains/kotlin/backend/common/bridges/bridges.kt
+++ b/compiler/backend-common/src/org/jetbrains/kotlin/backend/common/bridges/bridges.kt
@@ -56,7 +56,7 @@ fun <Function : FunctionHandle, Signature> generateBridges(
 
     val implementation = findConcreteSuperDeclaration(function)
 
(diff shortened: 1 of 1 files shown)


udalov:
This looks great and probably fixes https://youtrack.jetbrains.com/issue/KT-23707. Please also add a test case to `MethodOrderTest` just in case (even if it doesn't reproduce on your machine)

udalov:
I've added the test case and merged these changes. Thank you!

PR #1632, opened 2018-04-18
Fix problem with trimEnd wiping out valid spans

jt-gilkeson:
If I pass in a CharSequence with different spans (ImageSpans, colors, etc), the original trimEnd using substring incorrectly wiped out the spans.
changing to subSequence correctly preserves the span data


jt-gilkeson:
Test Case:
val trimmed = SpannableStringBuilder().append("SomeText").append("    ").trimEnd()

substring method trimmed is converted to String 
subSequence preserves trimmed as a Spannable

PR #1633, opened 2018-04-18
KT-11156 Incorrect highlighting for nested class in "Redundant SAM-constructor" inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11156


mglukhikh:
Merged, thank you

PR #1635, opened 2018-04-19
KT-23465 False positive `suspicious callable reference` on lambda invoke with parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23465


mglukhikh:
Merged, thank you

PR #1636, opened 2018-04-19
Add correct type parameter information to synthetic declarations

sandwwraith:
Needed by kapt due to https://github.com/Kotlin/kotlinx.serialization/issues/76

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/psi/synthetics/SyntheticClassOrObjectDescriptor.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/psi/synthetics/SyntheticClassOrObjectDescriptor.kt
@@ -50,13 +50,15 @@ class SyntheticClassOrObjectDescriptor(
     val syntheticDeclaration: KtPureClassOrObject = SyntheticDeclaration(parentClassOrObject, name.asString())
 
     private val thisDescriptor: SyntheticClassOrObjectDescriptor get() = this // code readability
+    var typeParameters: List<TypeParameterDescriptor> = emptyList()
+
     private val typeConstructor = SyntheticTypeConstructor(c.storageManager)
     private val resolutionScopesSupport = ClassResolutionScopesSupport(thisDescriptor, c.storageManager, c.languageVersionSettings, { outerScope })
     private val syntheticSupertypes =
         mutableListOf<KotlinType>().apply { c.syntheticResolveExtension.addSyntheticSupertypes(thisDescriptor, this) }
     private val unsubstitutedMemberScope =
         LazyClassMemberScope(c, SyntheticClassMemberDeclarationProvider(syntheticDeclaration), this, c.trace)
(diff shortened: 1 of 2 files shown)


compiler/frontend/src/org/jetbrains/kotlin/psi/synthetics/SyntheticClassOrObjectDescriptor.kt
```diff
@@ -50,13 +50,15 @@ class SyntheticClassOrObjectDescriptor(
     val syntheticDeclaration: KtPureClassOrObject = SyntheticDeclaration(parentClassOrObject, name.asString())
 
     private val thisDescriptor: SyntheticClassOrObjectDescriptor get() = this // code readability
+    var typeParameters: List<TypeParameterDescriptor> = emptyList()
```
udalov: TBH I'd prefer this to be `private lateinit var` + an `init` method that takes type parameters as a parameter, and require everyone to call `init` after construction explicitly, for clarity. What do you think?

compiler/frontend/src/org/jetbrains/kotlin/psi/synthetics/SyntheticClassOrObjectDescriptor.kt
```diff
@@ -50,13 +50,15 @@ class SyntheticClassOrObjectDescriptor(
...
+
     private val typeConstructor = SyntheticTypeConstructor(c.storageManager)
     private val resolutionScopesSupport = ClassResolutionScopesSupport(thisDescriptor, c.storageManager, c.languageVersionSettings, { outerScope })
     private val syntheticSupertypes =
         mutableListOf<KotlinType>().apply { c.syntheticResolveExtension.addSyntheticSupertypes(thisDescriptor, this) }
     private val unsubstitutedMemberScope =
         LazyClassMemberScope(c, SyntheticClassMemberDeclarationProvider(syntheticDeclaration), this, c.trace)
-    private val unsubstitutedPrimaryConstructor = createUnsubstitutedPrimaryConstructor(constructorVisibility)
+    private val _unsubstitutedPrimaryConstructor by lazy { createUnsubstitutedPrimaryConstructor(constructorVisibility) }
```
udalov: Please use `c.storageManager.createLazyValue` instead of `lazy` in compiler code to prevent possible deadlocks in case this code becomes less trivial

compiler/frontend/src/org/jetbrains/kotlin/psi/synthetics/SyntheticClassOrObjectDescriptor.kt
sandwwraith: In most cases, `emptyList()` is a reasonable default and allows to use existent code without modifications. Personally, I think that such `init` methods are error-prone because it's easy to forget them. 
However, I can see that a lot of descriptor entities in compiler have `initialize` method so I think it's a preferred code style and your comment is fair.

PR #1637, opened 2018-04-20
KT-6364 Incorrect quick-fixes are suggested for nullable extension function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-6364


mglukhikh:
Thank you, merged and closed. NB: it's better to use `createExpressionByPattern` when possible, see here: https://github.com/JetBrains/kotlin/commit/84d71e5a8ca145e23adf634e5814ceb7163c8e29

PR #1638, opened 2018-04-20
Ensure deterministic insertion of checkExpressionValueIsNotNull

denisvnukov:
Kotlinc source’s file DescriptorUtils.kt demonstarted non-deterministic
insertion of checkExpressionValueIsNotNull for value returned by
CallableDescriptor::getOriginal(). It was difficult to reproduce
this behavior on a smaller example, but I added a test which was
failing once in 5-10 times while I was testing manually.

I believe this bug is close to KT-23704.

This PR addresses non-determinism to a degree when I can run 120
compilations with './gradlew dist' and get same classes in all
jars in 'dist'.

NOTE that thew fact that insertion of checkExpressionValueIsNotNull may
depend on order of the types seems suspicios. This CL only addresses
non-determinism part, but I believe it’s worth looking into this more
from semantics point of view.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/types/CommonSupertypes.java
+++ b/compiler/frontend/src/org/jetbrains/kotlin/types/CommonSupertypes.java
@@ -100,7 +100,7 @@ public class CommonSupertypes {
     @NotNull
     private static SimpleType commonSuperTypeForInflexible(@NotNull Collection<SimpleType> types, int recursionDepth, int maxDepth) {
         assert !types.isEmpty();
-        Collection<SimpleType> typeSet = new HashSet<>(types);
+        Collection<SimpleType> typeSet = new LinkedHashSet<>(types);
 
         // If any of the types is nullable, the result must be nullable
         // This also removed Nothing and Nothing? because they are subtypes of everything else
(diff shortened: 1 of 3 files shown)


compiler/frontend/src/org/jetbrains/kotlin/types/CommonSupertypes.java
```diff
@@ -201,7 +201,7 @@ private static String renderTypeConstructorVerboseDebugInformation(TypeConstruct
 
         List<TypeConstructor> order = null;
         for (SimpleType type : types) {
-            Set<TypeConstructor> visited = new HashSet<>();
```
udalov: As far as I see, we never iterate over this set, so it can remain a `HashSet`

compiler/frontend/src/org/jetbrains/kotlin/types/CommonSupertypes.java
```diff
@@ -213,7 +213,7 @@ private static String renderTypeConstructorVerboseDebugInformation(TypeConstruct
         }
         assert order != null;
 
-        Set<TypeConstructor> notSource = new HashSet<>();
```
udalov: This one can also remain a `HashSet`

compiler/frontend/src/org/jetbrains/kotlin/types/CommonSupertypes.java
```diff
@@ -381,7 +381,7 @@ private static void markAll(@NotNull TypeConstructor typeConstructor, @NotNull S
                     @Override
                     public boolean beforeChildren(SimpleType current) {
                         Set<SimpleType> instances =
-                                constructorToAllInstances.computeIfAbsent(current.getConstructor(), k -> new HashSet<>());
```
udalov: This one can also remain a `HashSet`

udalov:
Thank you! I've [simplified](https://github.com/JetBrains/kotlin/commit/f59963ef7344099ab07cfbc58226525d2e903db7) the test case and merged manually.

Also I've investigated this a bit and I think there should be an assertion in the bytecode, so I've reported an issue: https://youtrack.jetbrains.com/issue/KT-24210

PR #1640, opened 2018-04-22
KT-23931 Avoid infinite loop into RedundantGotoMethodTransformer

mikaelpeltier:
- Follow-up of 9fb0f5981339d4a4da687fa9722afe3a71a682f5 to avoid
infinite loop during redundant goto otpimization.

Fix of https://youtrack.jetbrains.com/issue/KT-23931


erokhins:
Thank you for contribution! 

Merged manually: https://github.com/JetBrains/kotlin/commit/b1ef670818dea6112a422d0a987fb55dfe59c911

PR #1641, opened 2018-04-23
KT-12042 "Merge with next when" is not applicable when the statements delimited by semicolon or comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12042


mglukhikh:
Merged, thank you.

PR #1643, opened 2018-04-24
KT-9943 Quick fix/Intention to indent a raw string

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-9943

### DIFF
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -1592,6 +1592,11 @@
       <category>Kotlin</category>
     </intentionAction>
 
+    <intentionAction>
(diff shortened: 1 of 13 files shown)


idea/src/META-INF/plugin.xml
```diff
@@ -1592,6 +1592,11 @@
       <category>Kotlin</category>
     </intentionAction>
 
+    <intentionAction>
+      <className>org.jetbrains.kotlin.idea.intentions.IndentRawStringIntention</className>
+      <category>Kotlin</category>
+    </intentionAction>
+
```
mglukhikh: For future PRs: we have now "branch" files `plugin.xml.172` etc. that also should be modified if you are modifying `plugin.xml`.

mglukhikh:
Thank you, merged

PR #1644, opened 2018-04-24
StringNumberConversionsJVM: Make toBoolean() work with nullable Strings

sschuberth:
java.lang.Boolean.parseBoolean() is explicity documented to accept a
null argument, so there is no good reason this extension function should
not.


libraries/stdlib/jvm/src/kotlin/text/StringNumberConversionsJVM.kt
```diff
@@ -49,10 +49,10 @@ public inline fun Int.toString(radix: Int): String = java.lang.Integer.toString(
 public inline fun Long.toString(radix: Int): String = java.lang.Long.toString(this, checkRadix(radix))
 
 /**
- * Returns `true` if the contents of this string is equal to the word "true", ignoring case, and `false` otherwise.
+ * Returns `true` if the this string is not `null` and its content equals the word "true", ignoring case, and `false` otherwise.
```
mnonnenmacher: Remove "the" from "if the this string".

sschuberth:
@JakeWharton Any rationale for the downvote? Can I improve something?

ilya-g:
Related issue: https://youtrack.jetbrains.com/issue/KT-14119

sschuberth:
Like what? I find

    System.getenv("foo").toBoolean()

to be far more readable than e.g.

    System.getenv("foo")?.let { it == "true" } ?: false

JakeWharton:
Well not only are those not equivalent but you're not even playing fair. The comparison is
```kotlin
System.getenv("foo").toBoolean()
```
to
```kotlin
System.getenv("foo")?.toBoolean() ?: false
```

sschuberth:
True, that's better. Let's see how others think about this.

PR #1645, opened 2018-04-25
KT-14779 Inspection to replace String.format with string templates

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14779


4u7:
Sorry for late responce.
Could you please add tests with import `import java.lang.String.format`?

t-kameyama:
Added test and fixed.

PR #1646, opened 2018-04-26
KT-23977 wrong hint Unit redundant

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23977


mglukhikh:
Could you please check that the case from https://youtrack.jetbrains.com/issue/KT-24066 also works after your fix?

t-kameyama:
Added fix for https://youtrack.jetbrains.com/issue/KT-24066

mglukhikh:
In general it's Ok but it seems to me we can done both these things in more simple way. I'll make a few experiments and merge it (as it is or with some polishing).

mglukhikh:
Ok, I merged it. You can look at fixing / polishing commit: https://github.com/JetBrains/kotlin/commit/e07f83ddbc0cb1f558caef1f6d0e1db548fe9698. In particular: you took the first argument of function type, but should take the last one; it's better to use argument mapping to determine corresponding value parameter; it's better to use `getTargetFunctionDescriptor` to determine which function we are returning from.

PR #1647, opened 2018-04-27
KT-24001 "Suspicious combination of == and ===" false positive

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24001


mglukhikh:
Thank you, finally merged.

PR #1648, opened 2018-04-27
Get resource repositories from the manager

bendowski:
As the concrete repo implementations will be made package-private in AS 3.2 canary 14


bendowski:
Please let me know if this is the best branch to create a PR against.

yanex:
Looks like `ResourceRepositoryManager` is already used in the `as32` bunch:
https://github.com/JetBrains/kotlin/blob/master/plugins/android-extensions/android-extensions-idea/src/org/jetbrains/kotlin/android/model/impl/AndroidModuleInfoProviderImpl.kt.as32#L39

So I will close this. Please ping me if this is still actual for some reason.

PR #1649, opened 2018-04-30
KT-23897 Inspections: report extension functions declared in same class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23897

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedReceiverParameterInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/UnusedReceiverParameterInspection.kt
@@ -59,7 +59,8 @@ class UnusedReceiverParameterInspection : AbstractKotlinInspection() {
                 if (callableDeclaration.isOverridable() ||
                     callableDeclaration.hasModifier(KtTokens.OVERRIDE_KEYWORD) ||
                     callableDeclaration.hasModifier(KtTokens.OPERATOR_KEYWORD) ||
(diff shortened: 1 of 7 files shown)


goodwinnk:
I have accepted commit with reformatting already.

t-kameyama:
Fixed

goodwinnk:
I'm afraid, that the original request wasn't correct. It was a wise for inspection to check usages before reporting. Please take a look at this example:

```
class Test(val str: String) {
    private fun Test.print() {
        println(str)
        println(this.str)
        println(this@Test.str)
        println(this@print.str)
    }

    fun test(test: Test) {
        test.print()
    }
}

    Test("one").test(Test("two"))
}
```

It was unfortunate that there were no tests for this situation, and I think only they should be commited as result.

goodwinnk:
The original request might be still valid, but change signature should not be used as a fix, as for this case it's an error to remove the receiver at call place. 

The inspection should still be inapplicable if there's an access to surrounding this via label `this@ClassName`.

t-kameyama:
Fixed. Please take a look.

goodwinnk:
Wow, it was very fast fixes! 🚀 

I have squashed your commits and pushed them to master as e7a9614fb5eab5f1ed45ee8760566a0c53f8a954. 
As a follow-up commit, I have removed runtime usage for tests in fdc6372776b187ec4a08cb73f2b0b2d1630b2765, generally it better to avoid it, if it's not relevant for a test.

PR #1650, opened 2018-05-01
[KT-10456] [JS] Common Int.toString(radix: Int) method

FSteitz:
JS implementation of [KT-10456](https://youtrack.jetbrains.com/issue/KT-10456).


ilya-g:
Nice. 
Please move these tests https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/test/text/StringNumberConversionJVMTest.kt#L47-L82 to common `StringNumberConversionTest` class, to ensure these functions behave the same in JVM and JS.

PR #1651, opened 2018-05-01
KT-23756 Bogus "Might be const" warning in object expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23756


mglukhikh:
Thank you, merged.

PR #1652, opened 2018-05-01
KT-23788 Can't convert long char literal to string if it starts with backslash

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23788


idea/src/org/jetbrains/kotlin/idea/quickfix/QuickFixRegistrar.kt
```diff
@@ -461,6 +461,7 @@ class QuickFixRegistrar : QuickFixContributor {
         CANNOT_CHECK_FOR_ERASED.registerFactory(MakeTypeParameterReifiedAndFunctionInlineFix)
 
         TOO_MANY_CHARACTERS_IN_CHARACTER_LITERAL.registerFactory(TooLongCharLiteralToStringFix)
+        ILLEGAL_ESCAPE.registerFactory(TooLongCharLiteralToStringFix)
```
mglukhikh: What if ILLEGAL_ESCAPE is reported inside string literal?

mglukhikh: I will consider this case separately

mglukhikh:
Fixed and merged, thanks.

PR #1653, opened 2018-05-03
KT-23452 "Remove unnecessary parentheses" reports parens of returned function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23452


idea/src/org/jetbrains/kotlin/idea/intentions/RemoveEmptyParenthesesFromLambdaCallIntention.kt
```diff
@@ -42,6 +44,9 @@ class RemoveEmptyParenthesesFromLambdaCallIntention : SelfTargetingRangeIntentio
         val parent = element.parent as? KtCallExpression ?: return null
         val singleLambdaArgument = parent.lambdaArguments.singleOrNull() ?: return null
         if (element.getLineNumber(start = false) != singleLambdaArgument.getLineNumber(start = true)) return null
+        val prevCall = element.getPrevSiblingIgnoringWhitespaceAndComments() as? KtCallExpression
+        if (prevCall?.lambdaArguments?.let { it.size > 0 } == true
```
mglukhikh: It seems to me these checks are not relevant here already (we can return null without them). Will recheck it and fix if necessary.

mglukhikh: Another example you've probably missed:

```
fun returnFun2(i: Int): (() -> Unit) -> Unit = {}

fun test22() {
    returnFun2(1)() {}
}
```

PR #1654, opened 2018-05-03
[KT-10456] [JS] Common Int.toString(radix: Int) method

FSteitz:
JS implementation of [KT-10456](https://youtrack.jetbrains.com/issue/KT-10456). Was #1650 before I broke my branch.


ilya-g:
FYI, you can push or force-push to the same branch to update pull request. There's no need to recreate it.

ilya-g:
Thanks, merged!

PR #681, opened 2015-05-29
a bunch of fixes to the grammar

yole:

PR #1655, opened 2018-05-04
New annotation @kotlin.Generated. Annotating toString, hashCode, equals, copy and componentN functions in data classes.

andreyfomenkov:
Code coverage libraries such as [JaCoCo](https://github.com/jacoco/jacoco) can use this feature to skip generated code and thus give cleaner coverage reports (after merge [#681](https://github.com/jacoco/jacoco/pull/681)).
The idea is described in this post: [Kotlin + JaCoCo: Tuning Compiler to Skip Generated Code.](https://medium.com/@andrey.fomenkov/kotlin-jacoco-tuning-compiler-to-skip-generated-code-935fcaeaa391)

PR contains:

- [x] New annotation `@kotlin.Generated` for methods generated by the Kotlin compiler;
- [x] Annotating `toString`, `hashCode`, `equals`, `copy` and `componentN` functions for Data classes;
- [x] Using existing test `CompilerLightClassTestGenerated.java` with slightly modified test data ([source Kotlin file](https://github.com/andreyfomenkov/kotlin/blob/fomenkov/generated-annotation-data-class/compiler/testData/asJava/lightClasses/DataClassWithCustomImplementedMembers.kt) and [expected Java code](https://github.com/andreyfomenkov/kotlin/blob/fomenkov/generated-annotation-data-class/compiler/testData/asJava/lightClasses/DataClassWithCustomImplementedMembers.java)). Exactly what I need to compare between overridden and compiler generated methods.

By applying patched Kotlin compiler locally to our Android [MVI](http://hannesdorfmann.com/android/model-view-intent) project we managed to reduce total uncovered lines from **44K down to 28K** without adding any new tests! Obviously because there are tons of data classes.


compiler/tests/org/jetbrains/kotlin/codegen/DataClassGeneratedTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    @Test
+    fun testOverrideHashCode() {
+        loadText(
+                "data class User(val name: String, var age: Int) {" +
+                        "override fun hashCode(): Int = 123" +
+                        "}")
+        val clazz = generateClass("User")
+        val methods = hashSetOf("toString", "equals", "component1", "component2", "copy")
+
+        hasGeneratedAnnotationForMethods(clazz, methods)
```
JLLeitschuh: I would also suggest that you have a check that makes sure that the method `hashCode` does not have the annotation on it.

andreyfomenkov: It already has. The test iterates over all methods (even setters/getters) and if method name contains inside Set<> we assume that this method was annotated, otherwise - asserting that method doesn't have `kotlin.Generated` annotation

andreyfomenkov:
@sashache just added link to this issue

JLLeitschuh:
What happened to the Unit test you had before?

andreyfomenkov:
@JLLeitschuh I found the existing test that perfectly fits in to my requirement (only need to add expected annotation). That's why the old one was removed.

core/runtime.jvm/src/kotlin/Generated.kt
```diff
@@ -0,0 +1,8 @@
+package kotlin
+
+/**
+ * This annotation is present for methods generated by the Kotlin compiler
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class Generated
```
gildor: Why is this annotation is a part of `runtime.jvm`? And if it supposed to be Jvm only, according to existing naming convention it should be `JvmGenerated` instead

andreyfomenkov: This new annotation is consistent with `@kotlin.Metadata`. You may find it together in the same package.

gildor: Huh, you right, interesting.

JLLeitschuh:
@andreyfomenkov Don't coroutines generate methods as well. Do you want to add the `@kotlin.Generated` annotations to those as well?

andreyfomenkov:
@JLLeitschuh Yes, it would be great to add this annotation for other places like coroutines and generated get/set methods for `lateinit` properties, but in a separate PR. Have plans for that.

JLLeitschuh:
Is this PR considered "complete".
Really hoping to see this in a release of Kotlin soon.

core/descriptors/src/org/jetbrains/kotlin/descriptors/FunctionDescriptor.java
```diff
@@ -79,6 +79,8 @@
 
     boolean isSuspend();
 
+    boolean isGenerated();
```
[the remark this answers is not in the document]
JLLeitschuh: > so descriptor retrieved from compiled code are equivalent to descriptors before compilation.

Can you explain that in a bit more detail?

Also, @goodwinnk, wouldn't this information be good to have on the `FunctionDescriptor` interface to prevent future developers from making the same mistake?

core/runtime.jvm/src/kotlin/Generated.kt
JLLeitschuh: @goodwinnk I agree with your comment, but I'm not sure what the implications are with respect to what you expect @andreyfomenkov to change in this PR.

andreyfomenkov:
@goodwinnk I made some changes:
- method `isGenerated()` removed from `FunctionDescriptor`. Instead added boolean flag argument for `FunctionCodegen.generateMethod(...)`;
- we should keep `kotlin.Generated` class because without it annotation processor gives `NoClassDefFoundError` when running tests on Android;
- new annotation descriptor added to `JvmAnnotationNames` in the same way like `kotlin.Metadata`;
- added util method `WriteAnnotationUtil.writeGeneratedAnnotation(MethodVisitor)` to use it here and for similar PRs.

Regarding disadvantage with bytecode increase - does compiler have a flag for producing debug information so we can disable / enable adding `kotlin.Generated`? In that case we will not get bytecode increase for release builds when feature is disabled.

compiler/backend/src/org/jetbrains/kotlin/codegen/writeAnnotationUtil.kt
```diff
@@ -51,3 +52,7 @@ fun writeSyntheticClassMetadata(cb: ClassBuilder, state: GenerationState) {
         // Do nothing
     }
 }
+
+fun markMethodAsGenerated(mv: MethodVisitor) {
```
JLLeitschuh: Kotlin doc?

JLLeitschuh:
> does compiler have a flag for producing debug information so we can disable / enable adding kotlin.Generated?

Personally, I think this is a bad idea, but I'm not a Jetbrains dev, so feel free to ignore me.

Am I correct in reading that you change adds an annotation that is purely a bytecode annotation and doesn't have  a strongly typed representation backing it?

compiler/backend/src/org/jetbrains/kotlin/codegen/writeAnnotationUtil.kt
andreyfomenkov: Good point :+1: Added small comment

andreyfomenkov:
@JLLeitschuh yes, now just adding bytecode descriptor. As @goodwinnk mentioned it's enough. But at the same time we're keeping `kotlin.Generated` class for library to work correctly with annotation processors.

goodwinnk:
@andreyfomenkov Thank you for the fixes! I think at this point code would be ready for the push.

> we should keep `kotlin.Generated` class because without it annotation processor gives NoClassDefFoundError when running tests on Android;

I'm sorry, I was wrong about it. I didn't think about annotations processors at the moment of commenting. An absence of the annotation class file will probably also cause a warning from the proguard tool. So you're absolutely right, the annotation must be kept.

But there's a problem. After an internal discussion, we have decided that such annotation should not be added to Kotlin runtime till other ways of resolving coverage problem are tried (I added some reasons behind this decision in this comment: https://youtrack.jetbrains.com/issue/KT-18383#focus=streamItem-27-2890396-0-0).

I'm still hoping that this pull request can be partially reused (the commit and authorship will be preserved in the log in that case) for writing a compiler plugin if it would be necessary.

I'm very grateful for your effort as it practically closed one possible solution and I believe gave a great boost in resolving the coverage problem issue.

yaitskov:
so how to get Generated annotation?

PR #1656, opened 2018-05-07
KT-23775 IntelliJ plugin: improve "accessor call that can be replaced with property"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23775


mglukhikh:
Merged, thank you!

PR #1657, opened 2018-05-07
KT-21878 arrayOf() call can be replaced by array litteral [...]

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21878


asedunov:
Merged manually. Thank you for your contribution!

PR #1658, opened 2018-05-08
fix gradle license path

dschaller:
The location of this gradle license file is incorrect according to the path listed under:
https://github.com/JetBrains/kotlin/tree/master/license#kotlin-tools-and-libraries-tests


dschaller:
PTAL @4u7

dschaller:
PTAL @goodwinnk

PR #1660, opened 2018-05-08
KT-23752 False positive "Remove variable" quick fix on property has lambda or anonymous function initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23752


mglukhikh:
Merged, thank you!

PR #1661, opened 2018-05-08
KT-24263 Add `Make variable immutable` quickfix for const

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24263


mglukhikh:
Thank you, merged

PR #1662, opened 2018-05-08
KT-21998 Inspection to detect incompatible @Target & @Retention

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21998

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+class AnnotationTargetExpressionInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return classVisitor(fun(element: KtClass) {
+            if (!element.isAnnotation()) return
+            val target = element.targetAnnotationArgument()?.asElement() ?: return
+            if (element.retentionAnnotation() == null) return
+
+            val start = element.startOffset
+            val highlightRange = TextRange(target.startOffset - start, target.endOffset - start)
+            holder.registerProblem(
+                holder.manager.createProblemDescriptor(
+                    element,
+                    highlightRange,
+                    "annotation will not be stored",
+                    ProblemHighlightType.GENERIC_ERROR_OR_WARNING,
+                    isOnTheFly,
+                    ChangeAnnotationRetentionToSourceFix()
+                )
+            )
+        })
+    }
+}
+
+private class ChangeAnnotationRetentionToSourceFix : LocalQuickFix {
+
+    private val source = AnnotationRetention.SOURCE.name
+
+    override fun getName() = "Change @Retention to $source"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val element = descriptor.psiElement as? KtClass ?: return
+        val retention = element.retentionAnnotation() ?: return
+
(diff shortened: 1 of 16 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+class AnnotationTargetExpressionInspection : AbstractKotlinInspection() {
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return classVisitor(fun(element: KtClass) {
```
yanex: Please replace an anonymous function with a lambda argument.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+            if (!element.isAnnotation()) return
+            val target = element.targetAnnotationArgument()?.asElement() ?: return
```
yanex: Don't trust the names in PSI, this won't work with type/import aliases.
It's good idea to check them firstly (as it's very cheap), but please check the calls/descriptors as well.
If would be really nice if you could reuse a single `BindingContext` from the `element.analyze()`.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    override fun getName() = "Change @Retention to $source"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val element = descriptor.psiElement as? KtClass ?: return
+        val retention = element.retentionAnnotation() ?: return
+
+        val annotationTarget = "${KotlinBuiltIns.FQ_NAMES.annotationRetention.shortName().asString()}.$source"
```
yanex: Please move `KotlinBuiltIns.FQ_NAMES.annotationRetention.shortName().asString()` to a constant.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+                    ChangeAnnotationRetentionToSourceFix()
+                )
+            )
+        })
+    }
+}
+
+private class ChangeAnnotationRetentionToSourceFix : LocalQuickFix {
+
+    private val source = AnnotationRetention.SOURCE.name
```
yanex: Please move this to a companion object.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+            val target = element.targetAnnotationArgument()?.asElement() ?: return
+            if (element.retentionAnnotation() == null) return
+
+            val start = element.startOffset
+            val highlightRange = TextRange(target.startOffset - start, target.endOffset - start)
+            holder.registerProblem(
+                holder.manager.createProblemDescriptor(
+                    element,
+                    highlightRange,
+                    "annotation will not be stored",
```
yanex: Message is not capitalized:

screenshot 2018-11-05 at 19 45 11

yanex: By the way, in Kotlin 1.3. we have compiler warnings/errors for this (`RESTRICTED_RETENTION_FOR_EXPRESSION_ANNOTATION`, `RESTRICTED_RETENTION_FOR_EXPRESSION_ANNOTATION_WARNING`). I think it will make the perfect sense to remove `AnnotationTargetExpressionInspection` and bind `ChangeAnnotationRetentionToSourceFix` directly to them.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val element = descriptor.psiElement as? KtClass ?: return
+        val retention = element.retentionAnnotation() ?: return
+
+        val annotationTarget = "${KotlinBuiltIns.FQ_NAMES.annotationRetention.shortName().asString()}.$source"
+        val psiFactory = KtPsiFactory(element)
+        if (retention.valueArgumentList == null) {
+            retention.add(psiFactory.createCallArguments("($annotationTarget)"))
```
yanex: Please insert the qualified name and then call ShortenReferences.
You code won't work with something like this:

```kotlin
class AnnotationRetention // unrelated class in the same package

@Retention()
@Target(AnnotationTarget.EXPRESSION)
annotation class Ann
```

idea/testData/inspectionsLocal/annotationTargetExpression/noRetention.kt
```diff
@@ -0,0 +1,4 @@
+// PROBLEM: none
+// DISABLE-ERRORS
```
yanex: Looks like switching to `RESTRICTED_RETENTION_FOR_EXPRESSION_ANNOTATION_WARNING` should fix this.

idea/src/org/jetbrains/kotlin/idea/inspections/AnnotationTargetExpressionInspection.kt
t-kameyama: Please review https://github.com/JetBrains/kotlin/pull/1784. I'll close this PR.

PR #1663, opened 2018-05-09
Fix completion for back-ticked name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19863
https://youtrack.jetbrains.com/issue/KT-19864


goodwinnk:
Merged manually as https://github.com/JetBrains/kotlin/commit/18ed031e7e559d4de79dda045375f32c0593df1c. I've added additional check for index  is zero as `CodeFragmentCompletionHandlerTestGenerated.testInsertImport`  fails without it.

Thank you!

PR #1664, opened 2018-05-10
KT-11850 Add `nested lambdas with implicit parameters` warning

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11850

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/NestedLambdaShadowedImplicitParameterInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/NestedLambdaShadowedImplicitParameterInspection.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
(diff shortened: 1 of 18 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/NestedLambdaShadowedImplicitParameterInspection.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return lambdaExpressionVisitor(fun(lambda: KtLambdaExpression) {
+            val context: BindingContext by lazy { lambda.analyze(BodyResolveMode.PARTIAL) }
+            if (lambda.valueParameters.isNotEmpty() || lambda.functionDescriptor(context)?.valueParameters?.size != 1) return
+            if (lambda.getParentImplicitParameterLambda(context) == null) return
+            val callee = lambda.getStrictParentOfType<KtCallExpression>()?.calleeExpression ?: return
+            holder.registerProblem(
+                callee,
+                "implicit parameter 'it' is shadowed",
```
4u7: Maybe it is worth to make this message more specific, something like "Implicit parameter 'it' of enclosing lambda is shadowed"

4u7:
Thanks for the contribution!

jt-gilkeson:
I think I may have found a bug in this new lint warning: https://youtrack.jetbrains.com/issue/KT-26027

PR #1665, opened 2018-05-10
KT-24353 add support for junit 5 in kotlin.test

evgzakharov:


build.gradle.kts
```diff
@@ -128,6 +128,8 @@ extra["versions.jsr305"] = "1.3.9"
 extra["versions.jansi"] = "1.16"
 extra["versions.jline"] = "3.3.1"
 extra["versions.junit"] = "4.12"
+extra["versions.junit5"] = "5.2.0"
+extra["versions.junit-platform"] = "1.2.0"
```
ilya-g: Here we specify the versions that will be used throughout the entire project,  e.g. in case if Kotlin project would be using JUnit 5 to run its own tests. That version should not necessary be the same as the version against which kotlin-test-junit5 is being built.

evgzakharov: So, better to move concrete version to library project? 
I have just take the latest version of junit, and forgot to remove `versions.junit-platform`(it's not longer needed because gradle 4.6 now has native support of junit5)

ilya-g:
Given that `kotlin-test-junit5` depends only on `junit-jupiter-api` and doesn't bring the implementation as a transitive dependency, so that that dependency has to be declared in a consuming project, it makes sense to depend on the _minimal compatible_  junit api version instead of the latest.

evgzakharov:
Ok.  I've downgraded version of `junit` to 5.0.0.

PR #1667, opened 2018-05-11
KT-23306 Add intention of putting remaining when-values even in end, and even if there is "else"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23306


mglukhikh:
Merged, thank you!

PR #1668, opened 2018-05-11
KT-22371 "Create secondary constructor" quick fix is not suggested for supertype constructor reference

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22371


goodwinnk:
I have merged the fix as https://github.com/JetBrains/kotlin/commit/5564102a3e4c64213c2431776edeb783186f57a6. Thank you!

PR #1671, opened 2018-05-14
KT-21556 "Call chain on collection type may be simplified" generates uncompiled code on IntArray

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21556


mglukhikh:
Merged, thank you!

PR #1672, opened 2018-05-14
KT-23513 "Remove parameter" quick fix makes caret jump to the top of the editor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23513


t-kameyama:
@asedunov 
Closed because this issue has already been fixed in #1620

mglukhikh:
@t-kameyama Thank you!

PR #1673, opened 2018-05-14
 KT-21763 Provide Char MIN_VALUE/MAX_VALUE constants

Hiosdra:
https://youtrack.jetbrains.com/issue/KT-21763


libraries/stdlib/src/kotlin/text/Char.kt
```diff
@@ -37,3 +37,17 @@ public fun Char.equals(other: Char, ignoreCase: Boolean = false): Boolean {
  * Returns `true` if this character is a Unicode surrogate code unit.
  */
 public fun Char.isSurrogate(): Boolean = this in Char.MIN_SURROGATE..Char.MAX_SURROGATE
+
+/**
+ * Minimum value for character
+ * @see <a href="https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html">
```
ilya-g: I don't think this way of linking is supported in `@see` tag.  I'm not sure what is the purpose of these links anyway.

Hiosdra: Is it good now?

core/builtins/native/kotlin/Char.kt
```diff
@@ -89,18 +89,6 @@ public class Char private constructor() : Comparable<Char> {
...
-         * The minimum value of a character code unit
-         */
-        public val Char.Companion.MIN_VALUE: Char
-            get() = '\u0000'
-
-        /**
-         * The maximum value of a character code unit
-         */
-        public val Char.Companion.MAX_VALUE: Char
-            get() = '\uffff'
```
ilya-g: The extensions are removed, but instead of them there should be const properties in this companion.

ilya-g:
Ok, I've added these constants. Thanks for helping with actual companion object implementations.

See this commit d85c21ba14600a739219547e0813f494b3d691cf in `1.3-M1` branch.

PR #1674, opened 2018-05-14
KT-22552 SimplifiableCallChain should keep formatting and comments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22552


idea/testData/inspectionsLocal/collections/simplifiableCallChain/listOfNotNullWIthComment.kt
```diff
@@ -0,0 +1,8 @@
+// WITH_RUNTIME
+
+fun test() {
+    <caret>listOf(
```
mglukhikh: I think one example may be not enough here. At least, it would be good to test comment preserving for all three `when` branches above.

t-kameyama:
Added test

mglukhikh:
Finally I decided that I refactor this a little bit and merge. Mostly I do not like this piece of code:

```
val argumentsText = when {
                secondCallArgumentList?.arguments?.isNotEmpty() == true && firstCallArgumentList?.arguments?.isNotEmpty() == true -> {
                    "${secondCallArgumentList.text.removeSuffix(")")}, ${firstCallArgumentList.text.removePrefix("(")}"
                }
                secondCallArgumentList?.arguments?.isNotEmpty() == true -> secondCallArgumentList.text
                firstCallArgumentList?.arguments?.isNotEmpty() == true -> firstCallArgumentList.text
                else -> ""
            }
```

Because it's very non-regular and relies on a fact that argument list starts with "(" and ends with ")" which can be true but not very stable :(. Your fixes are not necessary here, I will finish everything myself.

mglukhikh:
Finally merged after some changes.

PR #1675, opened 2018-05-16
KT-18106 Intention "to raw/ordinary string literal" should leave the cursor at the same offset in the string

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18106


4u7:
Thanks for the contribution!

PR #1676, opened 2018-05-16
KT-22298 Improve docs for Array.copyOf(newSize: Int)

neonailol:
Issue: https://youtrack.jetbrains.com/issue/KT-22298 
Finished work started in #1477

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/arrays.kt
+++ b/libraries/stdlib/samples/test/samples/collections/arrays.kt
@@ -76,4 +76,28 @@ class Arrays {
         }
     }
 
+    class CopyOfOperations {
+
+        @Sample
+        fun copyOf() {
+            val array = arrayOf("apples, oranges, lime")
+            val arrayCopy = array.copyOf()
+            assertPrints(arrayCopy.contentToString(), "[apples, oranges, lime]")
+        }
+
+        @Sample
+        fun resizingCopyOf() {
+            val array = arrayOf("apples, oranges, lime")
+            val arrayCopyResized = array.copyOf(5)
+            assertPrints(arrayCopyResized.contentToString(), "[apples, oranges, lime, null, null]")
+        }
+
+        @Sample
+        fun resizedPrimitiveCopyOf() {
+            val array = intArrayOf(1, 2, 3)
+            val arrayCopyResized = array.copyOf(5)
+            assertPrints(arrayCopyResized.contentToString(), "[1, 2, 3, 0, 0]")
+        }
+    }
+
 }
--- a/libraries/stdlib/samples/test/samples/collections/arrays.kt
+++ b/libraries/stdlib/samples/test/samples/collections/arrays.kt
@@ -17,7 +17,6 @@
 package samples.collections
 
 import samples.*
-import kotlin.test.*
 
 
 @RunWith(Enclosed::class)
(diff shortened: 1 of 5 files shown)


libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
```diff
@@ -443,7 +444,17 @@ object ArrayOps : TemplateGroupBase() {
...
+            """
+            Returns new array which is a copy of the original array, resized to the given [newSize], truncating or padding
+            with nulls if necessary. If the object is a primitive type, it will be padded with its default value.
+            For all indices that are valid in both the original array and the copy, the two arrays contents will be identical.
+            For all indices that are valid in the copy, but not in the original, the copy will contain either null or the default
+            value.
+            """
+        }
+        sample("samples.collections.Arrays.CopyOfOperations.resizingCopyOf")
+        sample("samples.collections.Arrays.CopyOfOperations.resizedPrimitiveCopyOf")
```
ilya-g: Better to provide each sample only to the related overload of `copyOf`

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
```diff
@@ -443,7 +444,17 @@ object ArrayOps : TemplateGroupBase() {
         include(ArraysOfPrimitives, PrimitiveType.defaultPrimitives)
         include(InvariantArraysOfObjects)
     } builder {
-        doc { "Returns new array which is a copy of the original array, resized to the given [newSize]." }
+        doc {
+            """
+            Returns new array which is a copy of the original array, resized to the given [newSize], truncating or padding
+            with nulls if necessary. If the object is a primitive type, it will be padded with its default value.
```
ilya-g: Please add a blank line to separate summary and description.
The text can be specialized for primitives based on the current value of overloads `family`.

libraries/stdlib/samples/test/samples/collections/arrays.kt
```diff
@@ -76,4 +76,28 @@ class Arrays {
...
+            val array = arrayOf("apples, oranges, lime")
+            val arrayCopy = array.copyOf()
+            assertPrints(arrayCopy.contentToString(), "[apples, oranges, lime]")
+        }
+
+        @Sample
+        fun resizingCopyOf() {
+            val array = arrayOf("apples, oranges, lime")
+            val arrayCopyResized = array.copyOf(5)
+            assertPrints(arrayCopyResized.contentToString(), "[apples, oranges, lime, null, null]")
```
ilya-g: Worth to add truncating sample too.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
neonailol: About text specialization, I am not sure how to approach this without coping description, the only thing I found is `textWhen` method, but with it description looks messy.

```
            """
            Returns new array which is a copy of the original array, resized to the given [newSize],
            truncating or padding with ${textWhen(family == ArraysOfPrimitives) { "primitive default value" }}${textWhen(family == InvariantArraysOfObjects) { "null" }} if necessary.

            For all indices that are valid in both the original array and the copy, the two arrays contents will be identical.
            For all indices that are valid in the copy, but not in the original, the copy will contain ${textWhen(family == ArraysOfPrimitives) { "primitive default value" }}${textWhen(family == InvariantArraysOfObjects) { "null" }}.
            """
```

ilya-g: If `textWhen` is messy, you can use plain `if` or `when` expression inside `${...}`

neonailol:
@ilya-g Implemented all requested changes

ilya-g:
Thanks, I've merged new docs with some edits.

PR #1677, opened 2018-05-17
KT-24425 wrong hint remove redundant Companion

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24425


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
```diff
@@ -29,6 +32,21 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
             if (!objectDeclaration.isCompanion()) return
             if (expression.text != objectDeclaration.name) return
 
+            val property = parent.getStrictParentOfType<KtProperty>()
+            if (property != null && property.name == parent.selectorExpression?.text) return
```
mglukhikh: What about case like below? Here we still cannot delete `Companion`, but your check does not work here.

```
class My {
    val name = "My name"
    val greeting = "Hello " + Companion.name
    companion object {
        val name = "Your name"
    }
}
```

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
```diff
@@ -29,6 +32,21 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
...
+
+            val function = parent.getStrictParentOfType<KtNamedFunction>()
+            if (function != null) {
+                val callee = (parent.selectorExpression as? KtCallExpression)?.calleeExpression
+                if (function.name == callee?.text) {
```
mglukhikh: Here we can build same example as for properties above

t-kameyama:
Fixed

mglukhikh:
Merged

PR #1680, opened 2018-05-21
KT-23356 Cross-platform function to convert CharArray slice to String

neonailol:
https://youtrack.jetbrains.com/issue/KT-23356


libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
```diff
@@ -267,7 +267,7 @@ public inline fun String(chars: CharArray): String =
  * Converts the characters from a portion of the specified array to a string.
  */
 @kotlin.internal.InlineOnly
-public inline fun String(chars: CharArray, offset: Int, length: Int): String =
+public actual inline fun String(chars: CharArray, offset: Int, length: Int): String =
```
ilya-g: It's worth to provide common `String(chars: CharArray)` too.

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -46,3 +46,11 @@ public inline val CharSequence.size: Int get() = length
 
 @kotlin.internal.InlineOnly
 internal inline fun String.nativeReplace(pattern: RegExp, replacement: String): String = asDynamic().replace(pattern, replacement)
+
+/**
+ * Converts the characters from a portion of the specified array to a string.
+ */
+@kotlin.internal.InlineOnly
+public actual inline fun String(chars: CharArray, offset: Int, length: Int): String {
+    return chars.joinToString(separator = "").substring(offset, offset + length)
```
ilya-g: Not a very efficient way to build a string from a `CharArray` especially if the `length` is way more smaller than the length of the array.

libraries/stdlib/js/test/core/StringsJsTest.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package test.js
+
+import kotlin.test.*
+
+class StringsJsTest {
```
ilya-g: Better to move common tests into `libraries/stdlib/test/text/StringTest.kt` instead of duplicating them for each platform.

libraries/stdlib/js/src/kotlin/string.kt
neonailol: How about this?
```kotlin
return chars.sliceArray(offset until offset + length).joinToString(separator = "")
```

libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
neonailol: Are simple call to `joinToString` will be fine in js version?

libraries/stdlib/js/src/kotlin/string.kt
ilya-g: I mean something like that: https://developers.google.com/web/updates/2012/06/How-to-convert-ArrayBuffer-to-and-from-String

neonailol:
@ilya-g moved tests to a common module and took implementation from the provided article. Although the article is old, and API for text manipulation no longer in draft state.

libraries/stdlib/common/src/kotlin/TextH.kt
```diff
@@ -270,3 +270,13 @@ expect fun Long.toString(radix: Int): String
...
+
+/**
+ * Converts the characters in the specified array to a string.
+ */
+expect inline fun String(chars: CharArray): String
+
+/**
+ * Converts the characters from a portion of the specified array to a string.
+ */
+expect inline fun String(chars: CharArray, offset: Int, length: Int): String
```
ilya-g: `inline` is an implementation detail of these functions, which may be or may not be suitable on a particular platform.
There's no reason to require it to be inline in the expectation.

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -46,3 +46,19 @@ public inline val CharSequence.size: Int get() = length
...
+@kotlin.internal.InlineOnly
+public actual inline fun String(chars: CharArray): String {
+    return js("String.fromCharCode").apply(null, chars)
+}
+
+/**
+ * Converts the characters from a portion of the specified array to a string.
+ */
+@kotlin.internal.InlineOnly
+public actual inline fun String(chars: CharArray, offset: Int, length: Int): String {
```
ilya-g: This overload doesn't benefit from being inline

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -46,3 +46,19 @@ public inline val CharSequence.size: Int get() = length
...
+    return String(chars.sliceArray(offset until offset + length))
```
ilya-g: Consider using `copyOfRange` instead of `sliceArray` as the latter requires a range to be instantiated.

neonailol:
@ilya-g made requested changes, removed inline from common expectations, and used copyOfRange. 
Not sure why `String(chars: CharArray, offset: Int, length: Int)` does not benefit from inlining, but removed this part from declaration

libraries/stdlib/test/text/StringTest.kt
```diff
@@ -1314,4 +1314,19 @@ ${"    "}
         assertEquals("  ABC\n   \n  123", "ABC\n   \n123".prependIndent("  "))
         assertEquals("  ", "".prependIndent("  "))
     }
+
+    @Test fun charArrayToStringFullSlice() {
+        val chars: CharArray = charArrayOf().plus("Kotlin".toList())
```
sandwwraith: I wonder if it was tested on strings with non-ASCII characters and surrogate pairs.

neonailol: @sandwwraith well String.fromCharCode does not support surrogate pairs, what can be done then?

sandwwraith: From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode#Getting_it_to_work_with_higher_values : 

> Although most common Unicode values can be represented with one 16-bit number (as expected early on during JavaScript standardization) and fromCharCode() can be used to return a single character for the most common values (i.e., UCS-2 values which are the subset of UTF-16 with the most common characters), in order to deal with ALL legal Unicode values (up to 21 bits), fromCharCode() alone is inadequate. Since the higher code point characters use two (lower value) "surrogate" numbers to form a single character, String.fromCodePoint() (part of the ES2015 standard) can be used to return such a pair and thus adequately represent these higher valued characters.

So I suppose you should use `fromCodePoint` instead. In Java, as I remember, char code represents also a whole complete char, while code point can be split to pair

sandwwraith: However, I just looked up `java.lang.String` documentation and found that it has [separate special constructor](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-int:A-int-int-) for working with code points array. To preserve this semantic in multiplatform code, it's better to leave things without surrogate pairs support. However, we probably need special function for handling code points. @ilya-g what do you think?

ilya-g: Given that both `String`  and `CharArray` are sequence of UTF-16 code units, I don't think any special support for surrogate pairs is required in these two functions.
The function that converts an array of Unicode code points to a `String` is out of scope of this PR.

sandwwraith: OK, than it's better to revert last commit and stick with previously tested version. Sorry for inconvenience.

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -52,7 +52,18 @@ internal inline fun String.nativeReplace(pattern: RegExp, replacement: String):
...
+    var result = ""
+    var position = 0
+    while (position < chars.size) {
+        if (chars.size > position + 1 && chars[position].isHighSurrogate() && chars[position + 1].isLowSurrogate()) {
+            result += js("String.fromCharCode(chars[position], chars[position + 1])")
+            position += 2
+        } else {
+            result += js("String.fromCharCode(chars[position])")
+            position += 1
+        }
```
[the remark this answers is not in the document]
neonailol: @ilya-g Yes, just got lost with these surrogate pairs, reverted last changes, and updated tests

ilya-g:
Good work!
I've merged it.

PR #1681, opened 2018-05-21
KT-24336 Kotlin/JS missing SVGMaskElement interface

neonailol:
https://youtrack.jetbrains.com/issue/KT-24336

### DIFF
--- a/libraries/stdlib/js/src/org.w3c/org.w3c.dom.svg.kt
+++ b/libraries/stdlib/js/src/org.w3c/org.w3c.dom.svg.kt
@@ -857,3 +857,14 @@ public external abstract class SVGViewElement : SVGElement, SVGFitToViewBox, SVG
     }
 }
 
+/**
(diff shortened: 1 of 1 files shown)


ilya-g:
Note that this file is generated. Have you obtained it by running the generator?

neonailol:
@ilya-g I have troubles with the generator, so I did this by hand.

`mvn exec:java` says `'mainClass'` is not valid
And code generated by antler missing class `WebIDLBaseVisitor`

Schahen:
@neonailol  as it've been told, this file is generated from and there are some reasons why it's not autocommited. From here we can do one of two things:

1. You create *.idl definition and add it to this PR, I'll  explain what would be the next step if we move in this direction.
2. In case you don't have time or interest in supporting this PR further, I'll just create new  one for supporting `SVGMaskElement`.

neonailol:
@Schahen I'll try to figure out how to make an idl definitions

Schahen:
[This commit](https://github.com/JetBrains/kotlin/commit/c1f75d2324fb0321cbcec1c1e3753f0c49a4e280) made this PR obsolete

PR #1682, opened 2018-05-21
KT-23932 add "PointerEvent" for kotlin-stdlib-js

neonailol:
https://youtrack.jetbrains.com/issue/KT-23932

### DIFF
--- a/libraries/stdlib/js/src/org.w3c/org.w3c.dom.events.kt
+++ b/libraries/stdlib/js/src/org.w3c/org.w3c.dom.events.kt
@@ -9,9 +9,19 @@
 @file:Suppress("NESTED_CLASS_IN_EXTERNAL_INTERFACE")
 package org.w3c.dom.events
 
-import org.w3c.dom.EventInit
-import org.w3c.dom.UnionElementOrMouseEvent
-import org.w3c.dom.Window
+import kotlin.js.*
+import org.khronos.webgl.*
+import org.w3c.dom.*
+import org.w3c.dom.css.*
+import org.w3c.dom.parsing.*
+import org.w3c.dom.svg.*
+import org.w3c.dom.url.*
+import org.w3c.fetch.*
+import org.w3c.files.*
+import org.w3c.notifications.*
+import org.w3c.performance.*
+import org.w3c.workers.*
+import org.w3c.xhr.*
 
 /**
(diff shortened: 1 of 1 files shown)


libraries/stdlib/js/src/org.w3c/org.w3c.dom.events.kt
```diff
@@ -483,3 +483,133 @@ public external interface EventListener {
...
+    open val height: Int
+    open val pressure: Double
+    open val tangentialPressure: Double
+    open val tiltX: Int
+    open val tiltY: Int
+    open val twist: Int
+    open val pointerType: String
+    open val isPrimary: Boolean
+    fun getCoalescedEvents(): Array<PointerEvent>
+}
```
Schahen: Here are some alternative definitions used in [Chromium](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/events/pointer_event.idl)  and [Gecko](https://github.com/mozilla/gecko-dev/blob/75ec5b5525839e8e8acc11b449ced08a4ce2ef5e/dom/webidl/PointerEvent.webidl) compared to this PR

Attribute | Gecko  |Chromium | This PR
------------ | ------------- | ------------- | -------------
pointerId | long | long | Int
width | long | double | Int
height | long | double | Int
pressure | float | float | Double
tangentialPressure | float | float | Double
tiltX | long | long | Int 
tiltY | long | long | Int
twist | long | long | Int
pointerType | DOMString | DOMString | String (that's fine)
isPrimary | boolean | boolean | Boolean
 getCoalescedEvents() |  sequence<PointerEvent> |  sequence<PointerEvent> | Array<PointerEvent>

Regarding this table, I think it worth to discuss following issues:

1.  Where is *.idl definition. Here's how it works - a corresponding idl-definition is created somewhere in `libraries/stdlib/js/idl`, then you can run gradle idl2k in `libraries/tools/idl2k` and commit generated files - there's some reasoning behind not having autcommited them. We can discuss step 1b more specifically later, however, the first step is still having idl definitions. 

2. I see no reason why we should be more demanding than browser engines, so I suggest to keep float for both `pressure` and `tangentialPressure ` (as of `width` and `height` I think it's fair thing to keep them Int)

Schahen:
Following commit made this PR irrelevant - https://github.com/JetBrains/kotlin/commit/44da3c89b7047fe65824f91933b05397c3c6670d

PR #1683, opened 2018-05-22
KT-23082 Add quick-fix for type variance conflict

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23082


mglukhikh:
Thank you, merged!

PR #1684, opened 2018-05-22
Fix for Kotlin/kotlinx.serialization/136

pdvrieze:
Create a version that actually passes a type parameter in write$Self

### DIFF
--- a/plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
+++ b/plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
@@ -366,6 +366,7 @@ object KSerializerDescriptorResolver {
 
         val args = mutableListOf<ValueParameterDescriptor>()
         var i = 0
+
         args.add(ValueParameterDescriptorImpl(
                 f,
                 null,
@@ -394,9 +395,13 @@ object KSerializerDescriptorResolver {
                 f.source)
         )
 
-        val kSerialClass = thisClass.getClassFromSerializationPackage("KSerializer").toSimpleType(false)
+        val kSerialClassDesc = thisClass.getClassFromSerializationPackage("KSerializer")
 
         thisClass.declaredTypeParameters.forEach {
+            val typeArgument = TypeProjectionImpl(it.defaultType)
+            val kSerialClass = KotlinTypeFactory.simpleNotNullType(Annotations.EMPTY, kSerialClassDesc, listOf(typeArgument))
+
+
             args.add(ValueParameterDescriptorImpl(
                     f,
                     null,
--- a/plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
+++ b/plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
@@ -366,7 +366,6 @@ object KSerializerDescriptorResolver {
 
         val args = mutableListOf<ValueParameterDescriptor>()
         var i = 0
-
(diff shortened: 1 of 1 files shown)


plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
```diff
@@ -366,13 +366,20 @@ object KSerializerDescriptorResolver {
 
         val args = mutableListOf<ValueParameterDescriptor>()
         var i = 0
+
+        val targ = TypeParameterDescriptorImpl.createWithDefaultBound(f, Annotations.EMPTY, false, Variance.INVARIANT, Name.identifier("T"), 0)
+        val typeArguments = listOf(TypeProjectionImpl(targ.defaultType))
+
+        val koutputClass = KotlinTypeFactory.simpleNotNullType(Annotations.EMPTY,
```
sandwwraith: `KOutput` does not have type parameters, so I have no idea why it helps here

plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
```diff
@@ -394,7 +401,7 @@ object KSerializerDescriptorResolver {
                 f.source)
         )
 
-        val kSerialClass = thisClass.getClassFromSerializationPackage("KSerializer").toSimpleType(false)
+        val kSerialClass = KotlinTypeFactory.simpleNotNullType(Annotations.EMPTY, thisClass.getClassFromSerializationPackage("KSerializer"), typeArguments)
```
sandwwraith: In general, this seems to be the right approach to fix issue, because `KSerializer` indeed has type arguments. However, you've created here one typearg `T`, and if your class will have more than one typearg, e.g. `Data<T0, T1>`, signature should look like `write$Self(KOutput, KSerialClassDesc, KSerializer<T0>, KSerializer<T1>`, not `...KSerializer<T>, KSerializer<T>)`. You can take a look at how `createSerializerGetterDescriptor` implemented, since it should has almost same signature.

plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
```diff
@@ -366,13 +366,20 @@ object KSerializerDescriptorResolver {
 
         val args = mutableListOf<ValueParameterDescriptor>()
         var i = 0
+
+        val targ = TypeParameterDescriptorImpl.createWithDefaultBound(f, Annotations.EMPTY, false, Variance.INVARIANT, Name.identifier("T"), 0)
```
sandwwraith: It would be perfect to take type arguments from `thisClass` instead of creating them in-place, because they should match – to operate on `Data<E,V>` we need `KSerializer<E>` and `KSerializer<V>`. 
However, this might be tricky and requires knowledge of compiler API; to bypass annotation processors/bytecode validators, AFAIK, it is sufficient just to declare them in-place.

pdvrieze:
You are right. I guess it "works" because of the way generics works and because the parameters to the serializers are unconstrained.

plugins/kotlin-serialization/kotlin-serialization-compiler/src/org/jetbrains/kotlinx/serialization/compiler/resolve/KSerializerDescriptorResolver.kt
pdvrieze: Probably by accident. I didn't look carefully at koutput

pdvrieze:
I have changed the code to take into account your suggestions. It now takes the type parameters of the class and passes them as type parameters for the serializer references for each of the class type parameters. The generated code now actually is conceptually correct, not just "working"

PR #1685, opened 2018-05-23
KT-22089 Explict This inspection false negative with synthetic Java property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22089


mglukhikh:
Merged, thank you!

PR #1686, opened 2018-05-23
KT-23511 "Remove parameter" quick fix: Remove unused type parameter used as type projection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23511#comment=27-2882256

Fix for this case:
```kt
interface TypeHolder<T, U>

fun <T, U> baz(<caret>x: TypeHolder<T, U>) {}

fun test(holder: TypeHolder<String, Int>) {
    baz(holder)
}
```


asedunov:
Merged! Thank you

PR #1687, opened 2018-05-24
KT-12298 Fix override signature doesn't remove bogus reciever

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12298


mglukhikh:
Merged, thank you!

PR #1688, opened 2018-05-25
KT-24557 False warning "Remove redundant call" for nullable.toString

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24557


mglukhikh:
Merged, thank you!

PR #1689, opened 2018-05-25
KT-22041 "Convert lambda to reference" suggested incorrectly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22041


mglukhikh:
Merged

PR #1690, opened 2018-05-25
KT-22333 Extract function: make name editor focused

maxmedvedev:


asedunov:
Merged manually. Thank you!

PR #1692, opened 2018-05-28
KT-21503 J2K: Intention to replace getOrDefault

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21503

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMapGetOrDefaultIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMapGetOrDefaultIntention.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.intentions
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.idea.inspections.collections.isCalling
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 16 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMapGetOrDefaultIntention.kt
```diff
@@ -0,0 +1,51 @@
...
+    KtDotQualifiedExpression::class.java, "Replace with indexing and elvis operator"
+) {
+    companion object {
+        private val getOrDefaultFqName = FqName("kotlin.collections.Map.getOrDefault")
+    }
+
+    override fun applicabilityRange(element: KtDotQualifiedExpression): TextRange? {
+        val callExpression = element.callExpression ?: return null
+        val calleeExpression = callExpression.calleeExpression ?: return null
+        if (!callExpression.isCalling(getOrDefaultFqName)) return null
```
darthorimar: I think it is better to check that callee's name is `getOrDefault` before resolving. A new J2K has a lot of post-processings and resolving every call expression on every pos-processing iteration may be rather expensive

idea/src/org/jetbrains/kotlin/idea/intentions/ReplaceMapGetOrDefaultIntention.kt
```diff
@@ -0,0 +1,51 @@
...
+        val (firstArg, secondArg) = callExpression.arguments() ?: return null
+        text = "Replace with ${element.receiverExpression.text}[${firstArg.text}] ?: ${secondArg.text}"
+        return calleeExpression.textRange
```
darthorimar: We should also check that `Map`'s value type is not nullable, otherwise we will get a wrong result like 
```kotlin
fun test(map: Map<Int, String?>) =
  map[1] ?: "default" // Overwriting returned null value :(
```

t-kameyama: Fixed

darthorimar: Thank you! will merge as remote run will be green ^_^

darthorimar:
Thank you for your contribution :)

PR #1693, opened 2018-05-29
KT-24632 Quick fix to add getter and setter shouldn't use `field` when it is not allowed

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24632


mglukhikh:
Merged, thank you!

PR #1695, opened 2018-05-30
Add nullability annotations to data class equals and toString.

JakeWharton:
[KT-22201](https://youtrack.jetbrains.com/issue/KT-22201)


udalov:
Sorry for the delay! We've merged your change and it'll be released in 1.2.70.

The delay was mainly caused by us trying to find the correct set of changes to adapt our optimized representation of Kotlin classes in the IDE, so that newly added nullability annotations would be seen correctly from Java sources: see eaa58c3ec226b94a8613439f21efd6e199bfba4c, be611cd15464f31c9d9a755f8f6c9491e0c28a53.

I've also changed the way the annotations are generated to make them appear in `RuntimeInvisibleAnnotations`, not `RuntimeVisibleAnnotations`, since `NotNull`/`Nullable` have retention `CLASS`, not `RUNTIME`: 628833af65c4cb4ac8be6376bf5971950e70ef2e.

PR #1696, opened 2018-06-01
KT-24509 Inspection "JUnit tests should return Unit"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24509

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/MainFunctionReturnUnitInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/MainFunctionReturnUnitInspection.kt
@@ -44,7 +44,7 @@ class MainFunctionReturnUnitInspection : AbstractKotlinInspection() {
 
             holder.registerProblem(
                 function.nameIdentifier ?: function,
-                "${if (isMain) "main" else "test"} should return Unit",
+                "${if (isMain) "main" else "JUnit test"} should return Unit",
                 ProblemHighlightType.GENERIC_ERROR_OR_WARNING,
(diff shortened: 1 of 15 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/MainFunctionReturnUnitInspection.kt
```diff
@@ -10,39 +10,54 @@ import com.intellij.codeInspection.ProblemDescriptor
...
+            } else {
+                val junitTestFqNames = listOf(FqName("org.junit.Test"), FqName("org.junit.jupiter.api.Test"))
+                if (testAnnotations.none { it.fqName() in junitTestFqNames }) return
+            }
+            if (descriptor.returnType?.let { KotlinBuiltIns.isUnit(it) } == true) return
 
             holder.registerProblem(
                 function.nameIdentifier ?: function,
-                "main should return Unit",
+                "${if (isMain) "main" else "test"} should return Unit",
```
mglukhikh: I'd suggest to report "JUnit test should return Unit" or "Test function should return Unit" or something like. "test should return Unit" looks not very pretty

idea/testData/inspectionsLocal/mainFunctionReturnUnit/junit4Test.kt
```diff
@@ -0,0 +1,9 @@
+// FIX: Add explicit Unit return type
```
mglukhikh: Please add `// PROBLEM: <problem text>` to one test with JUnit function and one test with main function

t-kameyama:
Fixed

mglukhikh:
Merged

PR #1698, opened 2018-06-04
KT-22871 CONST_VAL_NOT_TOP_LEVEL_OR_OBJECT should have quickfix to move property into companion object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22871


idea/src/org/jetbrains/kotlin/idea/quickfix/MoveConstValToCompanionObjectFix.kt
```diff
@@ -0,0 +1,35 @@
...
+
+class MoveConstValToCompanionObjectFix(property: KtProperty) : KotlinQuickFixAction<KtProperty>(property) {
+
+    override fun getText() = "Move to companion object"
+
+    override fun getFamilyName() = text
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val property = element ?: return
+        MoveMemberToCompanionObjectIntention().applyTo(property, editor)
```
mglukhikh: Please look at the following line in QuickFixRegistrar:

```
PROPERTY_INITIALIZER_IN_INTERFACE.registerFactory(RemovePartsFromPropertyFix, ConvertPropertyInitializerToGetterIntention)
```

As I can see we are able to use intention as quick fix, and should not provide a wrapper. Try to do the same here

t-kameyama:
Fixed

mglukhikh:
Merged

PR #1699, opened 2018-06-04
Separated idl2k frontend and backend invocations.

alexander-gorshenev:
This PR just separates the idl2k front end and back end invocations.
So that Kotlin/Native could provide their own back end.

### DIFF
--- a/libraries/tools/idl2k/src/main/kotlin/build.kt
+++ b/libraries/tools/idl2k/src/main/kotlin/build.kt
@@ -1,0 +1,129 @@
+package org.jetbrains.idl2k
+
+import org.antlr.v4.runtime.ANTLRFileStream
+import kotlin.collections.HashSet
+import java.io.File
+import java.io.IOException
+import java.net.URL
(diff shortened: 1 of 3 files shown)


libraries/tools/idl2k/src/main/kotlin/build.kt
```diff
@@ -0,0 +1,129 @@
...
+    }
+    val unions = generateUnions(definitions, repository.typeDefs.values)
+
+    init {
+        implementInterfaces(definitions)
+    }
+
+    val allPackages = (definitions.asSequence().map { it.namespace } + repository.enums.values.map { it.namespace }).distinct().sorted()
+
+    val mdnCache = updateMdnCache()
```
cy6erGn0m: Populating MDN cache over internet in the constructor... well,  I'd extract it to an additional function `prepare()`

alexander-gorshenev: I made `mdnCache` `by lazy`.

PR #1700, opened 2018-06-05
KT-22030 Invalid Function can be private inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22030


mglukhikh:
Generally it's good, but (probably) can be simplified. Will refactor and merge.

mglukhikh:
Simplified and merged.

PR #1702, opened 2018-06-05
KT-13782 Intention (and may be inspection) to convert toString() call to string template

tommykw:
https://youtrack.jetbrains.com/issue/KT-13782


mglukhikh:
Looks good. I will test it on our code base and then merge.

mglukhikh:
I optimized inspection a bit and merged it, see commit: https://github.com/jetbrains/kotlin/commit/e1d223937b54e8abb3dbc83917eef6de65c316d6. Usually it's good for performance to check first something on PSI and only then perform more precise check via resolved call or descriptor. Anyway, thank you.

PR #1704, opened 2018-06-05
KT-22162 Add indices to loop fails on destructing declarator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22162


mglukhikh:
Merged

PR #1706, opened 2018-06-07
KT-24792 Convert lambda to anonymous function should not be suggested inside inline function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24792


mglukhikh:
Merged

PR #1707, opened 2018-06-07
KT-13688 Add 'Change to val' quickfix for delegates without setValue

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13688


mglukhikh:
Merged

PR #1711, opened 2018-06-07
Implement GooglePluginUpdateVerifier for Android Studio 3.2.

juan-chen:
The verifier reads metadata from a static xml file, looks for a release entry in the file matching the current build, and checks if the entry contains the plugin version to be advertised.

Only two files are changed: GooglePluginUpdateVerifier.kt and KotlinLanguageConfiguration.kt. Other file changes are due to "bunch switch".


idea/src/org/jetbrains/kotlin/idea/configuration/KotlinLanguageConfiguration.kt
```diff
@@ -159,7 +159,7 @@ class KotlinUpdatesSettingsConfigurable : SearchableConfigurable, Configurable.N
 
                 is PluginUpdateStatus.Unverified -> {
                     val version = pluginUpdateStatus.updateStatus.pluginDescriptor.version
-                    form.setUpdateStatus("A new version $version is found. ${pluginUpdateStatus.message}", false)
+                    form.setUpdateStatus("A new version $version is found, but not yet verified by Android Studio: ${pluginUpdateStatus.message}", false)
```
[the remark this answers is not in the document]
juan-chen: Sounds good. I'll make the change.

idea/src/org/jetbrains/kotlin/idea/update/GooglePluginUpdateVerifier.kt
```diff
@@ -6,75 +6,66 @@
...
+
+            val release = getRelease(pluginCompatibility) ?: return PluginVerifyResult.decline("no verified versions for this build")
+            val version = pluginDescriptor.version
+            if (release.plugins().any { version == it.version }) {
+                return PluginVerifyResult.accept()
+            } else {
+                return PluginVerifyResult.decline("version to be verified")
+            }
+        } catch (e : Exception) {
+            val reason = String.format("Failed to verify plugin %s version %s due to exception %s",
```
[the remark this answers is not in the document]
juan-chen: Sure, I'll handle different kinds of exceptions separately. Thanks!

idea/src/org/jetbrains/kotlin/idea/configuration/KotlinLanguageConfiguration.kt
```diff
@@ -159,7 +159,8 @@ class KotlinUpdatesSettingsConfigurable : SearchableConfigurable, Configurable.N
 
                 is PluginUpdateStatus.Unverified -> {
                     val version = pluginUpdateStatus.updateStatus.pluginDescriptor.version
-                    form.setUpdateStatus("A new version $version is found, but not yet verified by Android Studio: ${pluginUpdateStatus.message}", false)
+                    val message = "<html>A new version $version is found, but unverified:<br/> ${pluginUpdateStatus.message}</html>"
```
goodwinnk: I this message should have a common and reasoning part as a separate sentence. It might not flexible enough to have extensions contributing to the message, as you must change full message thinking about all possible contributors, and you have to write extensions with the full message in mind.

Also, it's now unclear who is responsible for the unverified result.

I have prepared c31a6af98c4062723b7a61eaa3e94d5b623012d2 commit with adressesing these concerns. Can you please check if such solution suitable for you?

idea/src/org/jetbrains/kotlin/idea/update/GooglePluginUpdateVerifier.kt
```diff
@@ -32,16 +36,16 @@ class GooglePluginUpdateVerifier : PluginUpdateVerifier() {
...
-            }
+            return if (release.plugins().any { KOTLIN_PLUGIN_ID == it.id && version == it.version }) PluginVerifyResult.accept()
+                   else PluginVerifyResult.decline("version to be verified")
         } catch (e : Exception) {
-            val reason = String.format("Failed to verify plugin %s version %s due to exception %s",
-                    pluginDescriptor.name, pluginDescriptor.version, e.message)
-            return PluginVerifyResult.decline(reason)
+            LOG.error("Exception when verifying plugin ${pluginDescriptor.pluginId.idString} version $version", e)
+            return when (e) {
+                is MalformedURLException, is IOException ->
```
goodwinnk: I think some minor fixes are still possible for exception handling.

For example, separate processing for `MalformedURLException` looks to be unneeded, because only a static URL is processed here.

What's more important, is that `IOException` probably shouldn't have `LOG.error()` call, because it will be possible to send such messages through exception reporter. These errors can be caused by user local network settings and nothing can be done about them by authors of the code.

goodwinnk:
Hello, we may push an intermediate result to Kotlin repository if we agree on the general `PluginUpdateVerifier` extension structure. I've added some notes after reviewing dce30351ad14ffb5560cfaf87a6b2c9b5334f880 commit.

idea/src/org/jetbrains/kotlin/idea/configuration/KotlinLanguageConfiguration.kt
juan-chen: The separation of reasoning makes sense. Thanks!

juan-chen:
Thanks Nikolay! I've added a commit to fix the minor exception handling issues. Please take a look.

goodwinnk:
Merged manually, thanks!

PR #1712, opened 2018-06-08
KT-24728 Add quickfix to remove single explicit & unused lambda parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24728


asedunov:
Merged! Thank you for the contribution

PR #1714, opened 2018-06-08
Translate Readme.md files to Korean

tsfo1489:
We(BJH9750, EherSenaw, tsfo1489, zyoa) translate some Readme files to Korean
(It have merge commit, but we think it does not matter because we just add the non-executable files)


Alefas:
It's cool, but very hard to maintain (for example, I did yesterday commit to readme.md file). How we can solve it?

.idea/misc.xml
```diff
@@ -44,6 +44,31 @@
       </list>
     </option>
   </component>
+  <component name="NullableNotNullManager">
```
Alefas: Please remove changes in non-md files.

.idea/codeStyles/Project.xml
```diff
@@ -68,6 +68,31 @@
     <MarkdownNavigatorCodeStyleSettings>
       <option name="RIGHT_MARGIN" value="72" />
     </MarkdownNavigatorCodeStyleSettings>
+    <Objective-C-extensions>
```
Alefas: Please remove changes in non-md files.

Alefas:
Another thing is that for us it's hard to verify change (of course, I can check with google translate, but is it enough?).

.idea/codeStyles/Project.xml
EherSenaw: Thanks for your review, I will remove changes of non-md files. Thanks.

.idea/misc.xml
EherSenaw: Thanks, removed non-md changes.

EherSenaw:
Thanks for your kind review.
I and my team had some concerns about that problem.
We are also aware of that: your team might not be able to check whether our translation is appropriate or not.
So we are going to ask for help in our Korean programmer community group, "생활코딩", etc., to get advanced contribute for translating Kotlin readme.md files.
In that way, Korean developers' interest would get much higher, and therefore we are visualizing Kotlin's development will go much faster.

Alefas:
@EherSenaw, unfortunately, in any case, we will reject this PR.
This work is very valuable and it would be great to find a solution where to put it. I can suggest creating a separate repository, where you will be able to have a synchronous translation with some possible automation with alerts, when some md file is changed in Kotlin repo. Then we can put into our Readme file link to "Unofficial translation to the Korean language". 

Please contact me directly by e-mail (Alexander.Podkhalyuzin@jetbrains.com) to discuss possible solutions. I only want to warn, I'll be on vacation until 2 July.

Alefas:
@EherSenaw I still think it would be good to advertise somehow this work. It can't be official as I said already, however, it can be "unofficial" with the possibility for you and others to maintain this work.

JackyAndroid:
@Alefas Try to think about non-native English developers, if the corresponding documents are translated, it will greatly improve the speed of learning kotlin source.

Of course, understanding document translation is not well controlled. However, automated robots can handle such problems very well.

This is very helpful for promoting kotlin, I very much hope that you can rethink this question.

thank you very much.

These words were originally in Chinese, translated by Google.

PR #1717, opened 2018-06-11
KT-24815 Add Quick fix to remove illegal "const" modifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24815


mglukhikh:
Merged

PR #1718, opened 2018-06-12
KT-20583 Report "redundant let" even for `it` in argument position

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20583


mglukhikh:
Looks good, but I'd like to test it on our code base first. I remember well enough that "redundant let" has a tendency of giving false positives :)

mglukhikh:
BTW this case is not "false positive". Inspection is not reported when we expect, so it's "false negative".

t-kameyama:
Sorry my mistake. This is "false negative".

mglukhikh:
Merged, thank you. @t-kameyama I sent you an e-mail yesterday, have you received it?

t-kameyama:
@mglukhikh I replied. Thank you.

PR #1719, opened 2018-06-13
KT-24385 Convert lambda to reference refactor produces red code with companion object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24385


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
```diff
@@ -226,10 +227,11 @@ open class ConvertLambdaToReferenceIntention(text: String) :
                     val calleeReferenceExpression = singleStatement.calleeExpression as? KtNameReferenceExpression ?: return null
                     val resolvedCall = calleeReferenceExpression.resolveToCall() ?: return null
                     val receiver = resolvedCall.dispatchReceiver ?: resolvedCall.extensionReceiver
+                    val descriptor by lazy { receiver?.type?.constructor?.declarationDescriptor }
                     val receiverText = when {
-                        receiver == null -> ""
+                        receiver == null || descriptor?.isCompanionObject() == true -> ""
```
mglukhikh: What if this companion object belongs to some other class, not to the class we are inside? Please add test for this case and fix it if necessary.

t-kameyama: Added test

PR #1720, opened 2018-06-13
KT-11253 sumByLong

neonailol:
PR For [KT-11253 missing sumByLong function](https://youtrack.jetbrains.com/issue/KT-11253)


ilya-g:
Thank you for the PR, but we've taken a different route and introduced `sumOf { }` function overloaded by different numeric types of result.

PR #1723, opened 2018-06-15
Android Extensions: make synthetic import resolution more resilient

rsauciuc:
In commit ec0abb0 (Fix EA-79206: Process only valid layout.xml...) we started to drop layouts with invalid PsiFiles during resolution of (synthetic) package fragments.

In Android Studio we run into invalid files quite often, unfortunately non-deterministically, and it seems to be caused by a race with some invokeLater'ed code from android-ndk; this has the side effect of now showing all the corresponding symbols as unresolved, and the imports themselves are suggested for removal.

As a temporary fix I propose we try again to get a valid PsiFile.

Bug: https://issuetracker.google.com/78547457

Sample invalidation trace for a layout PsiFile:
[invalidation_trace.txt](https://github.com/JetBrains/kotlin/files/2104401/invalidation_trace.txt)


yanex:
Merged manually. Thanks!

PR #1724, opened 2018-06-15
KT-23458 ClassCastException when compiling when statements to JS

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23458


anton-bannykh:
@t-kameyama Thanks!

PR #1725, opened 2018-06-15
KT-22376 JS: TranslationRuntimeException on 'for (x in ("a"))'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22376


anton-bannykh:
@t-kameyama Thanks!

PR #1727, opened 2018-06-19
KT-16552 Add Sequence.orEmpty

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16552


ilya-g:
I'm ok to merge the first commit into the 1.3 branch.

Regarding the inspection, I wonder if it takes into account that the extension is introduced in Kotlin 1.3, and thus it should not trigger when apiVersion 1.2 and lower is used in a project?

mglukhikh:
@ilya-g I don't think it's really important. As you can see, this inspection works for a wide range of functions. `Sequence.orEmpty()` does not exist in 1.2, but this only means that in 1.2 inspection has no chances to highlight this call. So I'd recommend to merge both commits into 1.3 branch.

ilya-g:
@mglukhikh Thanks for the clarification.
@t-kameyama Thanks for the PR, I've merged it into 1.3-M1 branch.

PR #1730, opened 2018-06-20
as33: Android dependency reversal

rsauciuc:
A simpler diff view for this is https://github.com/rsauciuc/kotlin/commit/24fee7733d0be57d76628b6c992956de95d8f95b. For now everything is left mostly in place, to be removed only in the final build steps.


yanex:
Merged manually. Thanks!

PR #1735, opened 2018-06-21
Add companion object to Boolean

pakoito:
This PR adds a companion object to Boolean, as it's the only primitive type without one.

The rationale behind is being able to add discoverable APIs via extensions to the `Boolean` companion, the same way it's currently possible for every other primitive type, like [`Byte`](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/number.kt#L28-L41), [`String`](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/string.kt#L16-L31) or [`Short`](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/number.kt#L211-L224).

YouTrack: https://youtrack.jetbrains.com/issue/KT-7922

### DIFF
--- a/core/builtins/native/kotlin/Boolean.kt
+++ b/core/builtins/native/kotlin/Boolean.kt
@@ -21,6 +21,8 @@ package kotlin
  * represented as values of the primitive type `boolean`.
  */
 public class Boolean private constructor() : Comparable<Boolean> {
+    companion object {}
+
     /**
      * Returns the inverse of this boolean.
      */


ilya-g:
The issue is here: https://youtrack.jetbrains.com/issue/KT-7922. Could you please add there the examples of Boolean companion extensions as use cases?

westito:
Nice. This one-lined PR took 3 years to achieve :D I didn't remember why I need boolean companion and created that bug report.

core/builtins/native/kotlin/Boolean.kt
```diff
@@ -21,6 +21,8 @@ package kotlin
  * represented as values of the primitive type `boolean`.
  */
 public class Boolean private constructor() : Comparable<Boolean> {
+    companion object {}
```
ilya-g: This is only sufficient to _declare_ that `Boolean` has a companion. Boolean class is not compiled neither in JVM nor in JS, thus in order to provide the actual companion object some special support in the compiler is required.

pakoito: Will I need to write that support myself, or will you take care of it? If it's the former, can you please at least provide some pointers of where to look.

udalov: @pakoito Please see [primitiveCompanionObjects.kt](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/js/runtime/primitiveCompanionObjects.kt) (JS) and [PrimitiveCompanionObjects.kt](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/runtime/kotlin/jvm/internal/PrimitiveCompanionObjects.kt) (JVM) and look for how primitive companions are substituted with these internal classes in the compiler code by grepping for example "IntCompanionObject" on the whole project.

pakoito: Perfect! And how can I test if the changes have been applied correctly afterwards?

Kerooker: Maybe create a test that uses Boolean companion object?

ilya-g: I've refactored companion object tests, so a test for Boolean companion can be added there.
https://github.com/JetBrains/kotlin/blob/cbc3480d9e7c62e5ded6c6c339fcf17ac1adf3b2/libraries/stdlib/test/numbers/BuiltinCompanionTest.kt

Kerooker: Hey, @ilya-g . I'm trying to make the changes, but some tests are failing due to java JDK not being present in System Environment Variables. But they are. Could you give me some help in making the tests run?

```
Class org.jetbrains.kotlin.jvm.compiler.Java9ModulesIntegrationTest

java.lang.AssertionError: Environment variable JDK_9 is not set!
	at org.jetbrains.kotlin.test.KotlinTestUtils.getJdk9Home(KotlinTestUtils.java:604)

```

Kerooker: Hey, @ilya-g ! I'm trying to make the changes suggested, but I'm having troubles getting Kotlin to compile. Some tests are failing, and I'm unsure on how to test them. Would you mind running tests on my branch? https://github.com/Kerooker/kotlin/tree/companion-object

ilya-g: The missing part is some compiler code that glues together these declared companions and their runtime implementations. I don't know where it is exactly, but I've found some entry points in CompanionObjectMapping.kt and ClassMapperLite.kt.

After that you will have to bootstrap compiler before compiling stdlib tests, so that it would know how to compile this new companion object reference. This can be done with commands:
```
./gradlew publish
./gradlew coreLibsTest -Pbootstrap.local
```
Or you may postpone that part with stdlib tests for now.

pakoito: @Kerooker are you working in it? That'd be great and it'll free me up to focus on adding parallelism to `arrow-effects` :D

Kerooker:
@pakoito im trying to. :P

core/builtins/native/kotlin/Boolean.kt
Kerooker: @ilya-g I'm digging for the implementatioons, and I found this commit:
https://github.com/JetBrains/kotlin/commit/1f79e0905f3a5bffa6100ac4079f08fc3f1a1098

I'm going to try to replicate some of that behaviour. Tell me if I'm on the wrong path

pakoito:
Move the party to #1761

PR #1739, opened 2018-06-26
KT-24694 Move lambda out of parentheses should not be applied for multiple functional parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24694


mglukhikh:
Thank you, merged.

PR #1740, opened 2018-06-26
Added cycle functions, tests and documentation

NKb03:
[https://youtrack.jetbrains.com/issue/KT-25110](https://youtrack.jetbrains.com/issue/KT-25110)

### DIFF
--- a/libraries/stdlib/src/kotlin/collections/Sequences.kt
+++ b/libraries/stdlib/src/kotlin/collections/Sequences.kt
@@ -607,3 +607,84 @@ public fun <T : Any> generateSequence(seed: T?, nextFunction: (T) -> T?): Sequen
 public fun <T : Any> generateSequence(seedFunction: () -> T?, nextFunction: (T) -> T?): Sequence<T> =
     GeneratorSequence(seedFunction, nextFunction)
 
+private class CycleSequence<T>(private val iterable: Iterable<T>, private val cycleCount: Int): Sequence<T> {
+    override fun iterator(): Iterator<T> {
+        return object : Iterator<T> {
+            private var currentCycleCount = 1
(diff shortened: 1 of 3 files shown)


libraries/stdlib/src/kotlin/collections/Sequences.kt
```diff
@@ -607,3 +607,84 @@ public fun <T : Any> generateSequence(seed: T?, nextFunction: (T) -> T?): Sequen
...
+                    throw NoSuchElementException("Maximum cycle count is reached")
+                }
+            }
+        }
+    }
+}
+
+/**
+ * Returns a sequence cycling through the iterable [cycleCount] times
+ * * If the cycle count is exceeded a [NoSuchElementException] is thrown
```
ilya-g: The idea of throwing exception in the end of the resulting sequence is very questionable. Can you provide a justification for it?

ilya-g: According to your tests the resulting sequence completes normally. If you're documenting the `next` function behavior, than it's definitely wrong place to do it, and it's actually not needed because that's the standard iterator contract.

NKb03: When I call cycle on a three-element-list with cycleCount = 3 the resulting sequence should have 9 elements. When somebody calls calls iterator.next() 10 times, the last time ignoring that iterator.hasNext() returned false, it's completely normal that the last call fails. It's just an assurance that no-one iterates through the sequence to often. If only the for-loop is used then it's not necessary to throw the exception, but who can assure that. I think you're right that the documentation should not mention that and maybe the NoSuchElementException should have the standard message.

PR #1741, opened 2018-06-27
KT-15476 Inspection to convert non-lazy chains of collection functions into sequences

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15476


idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertCallChainIntoSequenceInspection.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+    collect(this)
+
+    if (calls.size < 2) return emptyList()
+
+    val transformationCalls = calls
+        .asSequence()
+        .dropWhile { it !in transformations }
+        .takeWhile { call ->
+            call in transformations && call.valueArguments.none { arg ->
```
mglukhikh: It's definitely not the fastest way to find the compatible transformation, if any. You have O(n) in your solution, but you can easily get O(1) instead.

idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertCallChainIntoSequenceInspection.kt
```diff
@@ -0,0 +1,164 @@
...
+                arg.anyDescendantOfType<KtReturnExpression> { it.labelQualifier == null }
+            }
+        }
+        .toList()
+    if (transformationCalls.size < 2) return emptyList()
+
+    return transformationCalls
+}
+
+private operator fun List<FqName>.contains(call: KtCallExpression): Boolean = any {
```
mglukhikh: Looks nice but it's a bit unusual to see such things in source code. My first reaction on `call in transformations` was "it's just a mistake, because types are incompatible". Well, it's not a mistake, but it takes time to understand. See also previous comment.

t-kameyama:
Fixed

mglukhikh:
Additional question: what about termination-like calls (`first()` etc.)? Should they block conversion into sequence in situations like `someList.map { it.toSomething() }.firstOrNull()`?. What do you think about this, @t-kameyama ?

mglukhikh:
And I found another point which is better to be fixed: highlighting range is too wide, it looks ugly. You can compare it, for example, with `SimplifiableCallChainInspection`.

t-kameyama:
No, it should be converted. Because it can be reduced unuseless intermediate list creation and looping by `filter`. I will fix.

t-kameyama:
Fixed

mglukhikh:
Final testing in progress...

mglukhikh:
Note: probably `forEach` & `forEachIndexed` should not be count as termination function to prevent possible concurrent modifications.

mglukhikh:
Finally merged it. Thank you.

PR #1743, opened 2018-06-28
KT-13515 Add intention to replace '?.let' with null check

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13515

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/SafeAccessToIfThenIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/SafeAccessToIfThenIntention.kt
@@ -20,7 +20,6 @@ import com.intellij.codeInsight.intention.LowPriorityAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
-import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
 import org.jetbrains.kotlin.idea.core.replaced
 import org.jetbrains.kotlin.idea.intentions.ReplaceSingleLineLetIntention
 import org.jetbrains.kotlin.idea.intentions.SelfTargetingRangeIntention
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/SafeAccessToIfThenIntention.kt
```diff
@@ -57,11 +64,24 @@ class SafeAccessToIfThenIntention : SelfTargetingRangeIntention<KtSafeQualifiedE
...
+                replaceSingleLineLetIntention.applyTo(it, editor)
+                val descriptor = (ifExpression.condition as? KtBinaryExpression)?.left?.resolveToCall()?.resultingDescriptor
+                ifExpression.then?.collectDescendantsOfType<KtNameReferenceExpression>()?.firstOrNull {
+                    it.resolveToCall()?.resultingDescriptor == descriptor
+                }
+            } else {
+                null
+            }
+        }
+
```
mglukhikh: I'm not quite sure about this code. First of all, the `replaceReference` is required to be calculated only in one branch, see the only usage. Second, I'd rename `let` parameter here, otherwise it's difficult to understand the code. And at last, this search among `KtNameReferenceExpression`s looks suspicious. Too many of them can exist in general case and I suspect (may be I'm wrong) this can give false positives.

idea/testData/intentions/branched/safeAccessToIfThen/letTopLevelVar.kt.after
```diff
@@ -0,0 +1,7 @@
+// WITH_RUNTIME
+
+var a: String? = "A"
+fun main(args: Array<String>) {
+    val a1 = a
+    if (a1 != null) a1.length + 1
+}
```
mglukhikh: I'd add at least one test with regular (non-redundant) `let` and more tests with redundant `let`, e.g. `a?.let { foo(it) }` and may be something else.

t-kameyama:
Fixed

mglukhikh:
Merged with some style fixes, thank you.

PR #1744, opened 2018-06-29
KT-25146 Add quick-fix for default parameter value removal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25146


idea/src/org/jetbrains/kotlin/idea/quickfix/RemoveDefaultParameterValueFix.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    override fun getText() = "Remove default parameter value"
+
+    override fun getFamilyName() = text
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val parameter = element ?: return
+        val typeReference = parameter.typeReference ?: return
+        val defaultValue = parameter.defaultValue ?: return
+        parameter.deleteChildRange(typeReference.nextSibling, defaultValue)
```
mglukhikh: What if we have a comment between type reference and `= defaultValue`, like this: `a: Int /* very important */ = 42`?

t-kameyama:
Fixed

mglukhikh:
Thank you, merged.

PR #1746, opened 2018-06-29
KT-7675 Create inspection to replace if with let

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-7675

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/branchedTransformations/IfThenToSafeAccessInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/branchedTransformations/IfThenToSafeAccessInspection.kt
@@ -83,6 +83,9 @@ class IfThenToSafeAccessInspection : AbstractApplicabilityBasedInspection<KtIfEx
         negatedClause == null && baseClause.isUsedAsExpression(context) -> false
         negatedClause != null && !negatedClause.isNullExpression() -> false
         else -> baseClause.evaluatesTo(receiverExpression) || baseClause.hasFirstReceiverOf(receiverExpression) ||
-                receiverExpression is KtThisExpression && getImplicitReceiver()?.let { it.type == receiverExpression.getType(context) } == true
+                receiverExpression is KtThisExpression && getImplicitReceiver()?.let { it.type == receiverExpression.getType(context) } == true ||
+                (baseClause as? KtCallExpression)?.valueArguments?.map { it.getArgumentExpression() }?.let {
+                    it.any { it?.evaluatesTo(receiverExpression) == true } && it.all { it is KtNameReferenceExpression }
+                } == true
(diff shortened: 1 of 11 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/IfThenUtils.kt
```diff
@@ -223,6 +225,24 @@ data class IfThenToSelectData(
...
+        return factory.buildExpression {
+            appendExpression(receiver)
+            appendFixedText("?.let {")
+            appendExpression(baseClause.calleeExpression)
+            appendFixedText("(")
+            baseClause.valueArguments.forEachIndexed { index, arg ->
+                if (index != 0) appendFixedText(", ")
+                val argExpression = arg.getArgumentExpression()
+                if (argExpression?.evaluatesTo(receiverExpression) == true)
+                    appendFixedText("it")
```
mglukhikh: What if we have already `it` in name scope. Is it possible to get a name conflict and/or semantic change?

idea/src/org/jetbrains/kotlin/idea/inspections/branchedTransformations/IfThenToSafeAccessInspection.kt
```diff
@@ -83,6 +83,9 @@ class IfThenToSafeAccessInspection : AbstractApplicabilityBasedInspection<KtIfEx
         negatedClause == null && baseClause.isUsedAsExpression(context) -> false
         negatedClause != null && !negatedClause.isNullExpression() -> false
         else -> baseClause.evaluatesTo(receiverExpression) || baseClause.hasFirstReceiverOf(receiverExpression) ||
-                receiverExpression is KtThisExpression && getImplicitReceiver()?.let { it.type == receiverExpression.getType(context) } == true
+                receiverExpression is KtThisExpression && getImplicitReceiver()?.let { it.type == receiverExpression.getType(context) } == true ||
+                (baseClause as? KtCallExpression)?.valueArguments?.map { it.getArgumentExpression() }?.let {
+                    it.any { it?.evaluatesTo(receiverExpression) == true } && it.all { it is KtNameReferenceExpression }
```
mglukhikh: Nested `it`s are very hard to read here.

mglukhikh: And I'd extract the whole condition (beginning from `baseClause as?`) into a function and simplify it a bit, for example to avoid safe call chains and `let`

mglukhikh:
I will test it a little bit and then merge.

mglukhikh:
Merged, thank you. NB: I downgraded this case to INFORMATION level (so just an intention).

PR #1747, opened 2018-06-29
Feature - KT-25180 Stdlib: Extension function to convert object to maps

navi25:
[YouTrack Link](https://youtrack.jetbrains.com/issue/KT-25180)
This code converts Objects to Maps.
I'm relatively new to this so I'm little lost about testing this module and finding the right place to define the to Method.
Any help will be appreciated.


ilya-g:
Closing for now as per the decision here https://youtrack.jetbrains.com/issue/KT-25180

PR #1749, opened 2018-07-01
Add diagnostic spec tests to "When expression" section (proof of concept)

PetukhovVictor:
This PR adds 134 [Kotlin language specification](https://github.com/JetBrains/kotlin-spec/tree/spec-rework) diagnostic tests to "When expression" section.

These tests verify that statements from the Kotlin specification correspond to the reality (or not if negative tests) at the level of compiler front-end using diagnostic representation of the source code.

Each test relates to a specific section, paragraph, and sentence of the Kotlin language specification, and is either positive or negative.

This PR adds tests (by paragraphs):
- paragraph 2: **4 tests** (negative: **2**, positive: **2**)
- paragraph 3: **36 tests** (negative: **6**, positive: **30**)
- paragraph 4: **69 tests** (negative: **12**, positive: **57**)
- paragraph 5: **8 tests** (negative: **4**, positive: **4**)
- paragraph 6: **17 tests** (negative: **8**, positive: **9**)

More about the diagnostic spec tests you can read in [ReadMe](https://github.com/PetukhovVictor/kotlin/blob/rr/petukhov/spec-tests/compiler/testData/diagnostics/testsSpec/ReadMe.md).

Also in this PR added gradle task `printSpecTestsStatistic` in `compiler:tests-common` that displays statistics on the spec tests with grouping.

`specSentenceUrl` on line 181 of `org.jetbrains.kotlin.test.SpecTestUtil` is supposed to be used when the handling of links to specific sentence appears in the [kotlin-spec](https://github.com/JetBrains/kotlin-spec) html version.
Before the appearance of this feature, you can see an example here: https://petukhovvictor.github.io/kotlin-spec/#when-expression:4:9
When following on such a link, a specified sentence will be highlighted in yellow.

Also in this version of the kotlin-spec html you can see the visual tests coverage: https://petukhovvictor.github.io/kotlin-spec/?mode=view_tests_coverage#when-expression

You can run the tests from `org.jetbrains.kotlin.checkers.DiagnosticsTestSpecGenerated` in IDE or using gradle: `./gradlew compiler:test --tests "org.jetbrains.kotlin.checkers.DiagnosticsTestSpecGenerated"`


dsavvinov:
A couple of suggestions:
* Successors of `BaseDiagnosticTest` generate `txt` file with descriptors dump. It is useful if we're interested in *declarations*: types, overrides, fake overrides from inheritance, etc. If there's no need in checking declarations (as in tests on `when`), generation of `.txt`-files can be omitted by adding `// SKIP_TXT` to the beginning of `.kt`-file.

* Probably it's worth to implement some library "for spec tests", with template functions and classes, like `getInt`/`getString`/etc., and compile spec tests against that "library" instead of copy-pasting such functions in each test. 

* Wild idea: allow spec tests to reference not only corresponding sentence, but other related sentences. E.g. if sentence says something like: "A when expression is called exhaustive if ... he bound expression is of type kotlin.Boolean and the conditions contain ... a constant expression evaluating to value true", then it would be great to reference part of spec about what is "constant expression" (because tests, most likely, will replicate tests on constant expression in context of `when`). 
  I don't know how it should be implemented (and if it should be at all) so it is "a wild idea" :)

* As have been discussed, we need to reduce volume of testdata. I propose to start with simple fix: do not try to extensively check all possible types in position of expression (like [here](https://github.com/PetukhovVictor/kotlin/blob/0e319789e110a9e67544802fbcf12064a6cb7144/compiler/testData/diagnostics/testsSpec/s-16.30_when-expression/p-3/pos/1.1.kt)). There is no type inference involved, and in most cases, `when` doesn't add any context, so it's quite unlikely that `when` will work with `Byte` but fail with `Int` or `Char` :) 
  Note that it still makes sense to check how compiler works with types when inference is involved, i.e. in [tests on `when-expression` type inference](https://github.com/PetukhovVictor/kotlin/blob/0e319789e110a9e67544802fbcf12064a6cb7144/compiler/testData/diagnostics/testsSpec/s-16.30_when-expression/p-5/pos/1.1.kt#L50)

* Instead of trying various calls with various types, it may be worthwhile to give more attention to language constructs: if-expression, nested `when`, elvis, nullability assertions, `as`-casts, etc. -- in other words, everything which is declared by language rather than user. 
  Example: it may be surprising, but it's actually OK (currently, at least) to write `return` as when-entry condition:
```
when {
    return -> println("WAT")
}
```

compiler/frontend/src/org/jetbrains/kotlin/checkers/CheckerTestUtil.java
```diff
@@ -583,11 +583,11 @@ public SyntaxErrorDiagnostic(@NotNull PsiErrorElement errorElement) {
     }
 
     public static class DebugInfoDiagnosticFactory extends DiagnosticFactory<DebugInfoDiagnostic> {
-        public static final DebugInfoDiagnosticFactory SMARTCAST = new DebugInfoDiagnosticFactory("SMARTCAST");
+        public static final DebugInfoDiagnosticFactory SMARTCAST = new DebugInfoDiagnosticFactory("SMARTCAST", Severity.INFO);
```
dsavvinov: We've discussed internally this change: it's OK, but could you please also change `Severity` to `INFO` for: `IMPLICIT_RECEIVER_SMARTCAST`,  `CONSTANT` and `DYNAMIC`

compiler/tests-spec/testData/diagnostics/s-16.30_when-expression/p-2/neg/3.1.kt
```diff
@@ -0,0 +1,15 @@
+// SKIP_TXT
```
udalov: Maybe we could skip txt generation for all these tests automatically by adding an abstract method into `AbstractDiagnosticsTest` and overriding it in this test, instead of adding `SKIP_TXT` to each file?

PetukhovVictor: Good idea.
In `AbstractDiagnosticsTest` already exist similar abstract method — `skipDescriptorsValidation`. I just used it in `AbstractDiagnosticsTestSpec`.
See: 8dcbd8074d6814218e1e16c373cb41e552a5942a

udalov: Thanks. There's an empty line at the beginning of each file now though 😀

PetukhovVictor: Exactly! Thanks. aa16d88693d212e37782c138601e3c78e1ad23a2

PetukhovVictor:
@dsavvinov, answers to [this](https://github.com/JetBrains/kotlin/pull/1749#issuecomment-402187737):

1. 👍, in this portion of the tests descriptors are not needed. Fixed. See [this comment](https://github.com/JetBrains/kotlin/pull/1749#discussion_r207180615).
2. Yes, it's a good idea. I wrote it down, a little later I implement a library with these and other functions and classes for spec tests.
3. Yes, we thought about it. Later I'll try to add the ability to write related sentences (or paragraphs, or maybe subsections) in the tests meta-information parser. This will allow us to collect test lists even with an indirect mention of language feature.

The remaining suggestions I will comment a little later.

Thanks for the valuable suggestions!

PetukhovVictor:
@dsavvinov, I implement helper files with template functions and classes for diagnostic spec tests. The needed file is included using the corresponding directive (as well as `CHECK_TYPE`, `EXPLICIT_FLEXIBLE_TYPES`, etc).

I implement following helpers:
- `basicTypes.kt`
- `classes.kt`
- `enumClasses.kt`
- `sealedClasses.kt`
- `funs.kt`
- `objects.kt`
- `typeAliases.kt`

Corresponding directive: `!WITH_BASIC_TYPES`, `!WITH_SEALED_CLASSES`, etc.

See b4049c6f4a9c4c6e29e4ba03f551c1fdcd788bc6 + 4091f4746e387e51c24fad533a4d0b3e0b266cc8.

PR #1751, opened 2018-07-05
KT-25177 Report asDynamic on dynamic type as a warning

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25177


mglukhikh:
Asked somebody from JS team to review.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantAsDynamicInspection.kt
```diff
@@ -0,0 +1,45 @@
...
+
+class RedundantAsDynamicInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        callExpressionVisitor(fun(call) {
+            if (call.platform != JsPlatform) return
+            if (call.calleeExpression?.text != "asDynamic") return
+            if (call.getQualifiedExpressionForSelector()?.receiverExpression?.getCallableDescriptor()?.returnType !is DynamicType) return
+            holder.registerProblem(
+                call,
+                "Redundant 'asDynamic' call",
```
anton-bannykh: It seems to me that the message is a bit misleading. The important part here is that the code might fail at run time, because the programmer doesn't realize the `asDynamic` is a member, not the extension function from stdlib s/he might be thinking of.

So maybe it should be something along the lines of "Suspicious 'asDynamic' member invocation" ? "Possibly eroneous"? Something else?

In any case I don't think the word "redundant" conveys the actual problem properly.

anton-bannykh:
Thanks! Will merge manually

anton-bannykh:
Merged as https://github.com/JetBrains/kotlin/commit/f2fec6b07815f9474241fc2588fa08e5d77bc0a1

PR #1752, opened 2018-07-06
KT-22072 "Convert MutableMap.put to assignment" should not be applicable when put is used as expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22072


mglukhikh:
Ok but I'm not sure `isUsedAsExpression` works correctly here. Asked Dmitry Savvinov to review this part.

mglukhikh:
Thank you, merged.

PR #1753, opened 2018-07-09
KT-25089 False-positive "Call chain on collection type can be simplified" for `map` and `joinToString` on a `HashMap`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25089

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
@@ -12,6 +12,7 @@ import org.jetbrains.kotlin.js.resolve.JsPlatform
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 import org.jetbrains.kotlin.psi.qualifiedExpressionVisitor
 import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.typeUtil.builtIns
 import org.jetbrains.kotlin.types.typeUtil.isSubtypeOf
 
@@ -25,11 +26,7 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
                 if (firstReceiverType != null) {
                     if (conversion.replacement == "mapNotNull" && KotlinBuiltIns.isPrimitiveArray(firstReceiverType)) return@check false
                     val firstReceiverRawType = firstReceiverType.constructor.declarationDescriptor?.defaultType
-                    if (firstReceiverRawType != null) {
-                        if (firstReceiverRawType.isSubtypeOf(builtIns.map.defaultType) ||
-                            firstReceiverRawType.isSubtypeOf(builtIns.mutableMap.defaultType)
-                        ) return@check false
-                    }
+                    if (firstReceiverRawType.isMap(builtIns)) return@check false
                 }
                 if (conversion.replacement.startsWith("joinTo")) {
                     // Function parameter in map must have String result type
@@ -83,4 +80,19 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
             Conversion("kotlin.collections.listOf", "kotlin.collections.filterNotNull", "listOfNotNull")
         )
     }
+}
+
+private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
+    if (this == null) return false
+
+    val map = builtIns.map.defaultType
+    val mutableMap = builtIns.mutableMap.defaultType
+
+    fun isMap(type: KotlinType?): Boolean {
+        if (type == null) return false
+        return type.isSubtypeOf(map) || type.isSubtypeOf(mutableMap) || type.constructor.supertypes.reversed().any {
+            isMap(it.constructor.declarationDescriptor?.defaultType)
+        }
+    }
+    return isMap(this)
 }
--- a/idea/testData/inspectionsLocal/collections/simplifiableCallChain/joinToStringOnHashMap.kt
+++ b/idea/testData/inspectionsLocal/collections/simplifiableCallChain/joinToStringOnHashMap.kt
@@ -1,0 +1,6 @@
+// PROBLEM: none
+// WITH_RUNTIME
+
+fun test(data: HashMap<String, String>) {
+    val result = data.<caret>map { "${it.key}: ${it.value}" }.joinToString("\n")
+}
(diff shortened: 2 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -83,4 +80,19 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
...
+private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
+    if (this == null) return false
+
+    val map = builtIns.map.defaultType
+    val mutableMap = builtIns.mutableMap.defaultType
+
+    fun isMap(type: KotlinType?): Boolean {
+        if (type == null) return false
+        return type.isSubtypeOf(map) || type.isSubtypeOf(mutableMap) || type.constructor.supertypes.reversed().any {
+            isMap(it.constructor.declarationDescriptor?.defaultType)
```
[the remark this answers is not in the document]
udalov: @mglukhikh In fact, `TypeConstructor.getSupertypes` returns immediate supertypes only (as a proof, see how `TypeUtils.getImmediateSupertypes` and `TypeUtils.getAllSupertypes` are implemented).

Nevertheless, I agree that this function is not implemented optimally. In addition to the name heuristic, I'd propose to get rid of recursion here completely, and also get rid of checking if the type is a subtype of `MutableMap` (`MutableMap` is a subtype of `Map` anyway) and implement it like this:
```
private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
    return this != null && constructor.declarationDescriptor?.name?.asString()?.endsWith("Map") == true && isSubtypeOf(map)
}
```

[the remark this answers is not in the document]
mglukhikh: @udalov Thank you for clarification!

t-kameyama:
Fixed

t-kameyama:
Fixed

idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -82,17 +79,10 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
     }
 }
 
-private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
-    if (this == null) return false
-
-    val map = builtIns.map.defaultType
-    val mutableMap = builtIns.mutableMap.defaultType
+private fun KotlinType?.isMapOrSubType(): Boolean {
+    return this != null && (isMap() || constructor.supertypes.any { it.isMap() })
```
udalov: It's not the implementation that @mglukhikh or me have suggested. Please re-read the discussion. This code would return true on a Java class named `SomethingMap` which does not inherit from `Map`, which is incorrect. What we're looking for is classes whose names both end with `"Map"` _and_ who inherit from `Map`.

idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -79,10 +82,14 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
...
 }
 
-private fun KotlinType?.isMapOrSubType(): Boolean {
-    return this != null && (isMap() || constructor.supertypes.any { it.isMap() })
-}
-
-private fun KotlinType.isMap(): Boolean {
-    return constructor.declarationDescriptor?.name?.asString()?.endsWith("Map") == true
-}
+private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
```
udalov: @t-kameyama Why don't you just use the implementation I've suggested in my [first comment](https://github.com/JetBrains/kotlin/pull/1753#discussion_r204414873) in this PR? Is there anything wrong with it?
```
private fun KotlinType?.isMap(builtIns: KotlinBuiltIns): Boolean {
    return this != null && constructor.declarationDescriptor?.name?.asString()?.endsWith("Map") == true && isSubtypeOf(map)
}
```
(note that this is copied from https://github.com/JetBrains/kotlin/pull/1753#discussion_r204414873)

t-kameyama: If `type` is` HashMap`, `type.isSubtype (map)` returns false, so I think that recursive checks is necessary.

t-kameyama: It also returns false if `type` is` MutableMap`.

udalov: The reason for this behavior is that you're taking the _default type_ of `kotlin.Map` (`builtIns.map.defaultType`). Default type of that class is `Map<K, V>` where `K` and `V` -- type parameters of `Map`. This type only makes sense inside the `Map` interface, as that is the only place where these type parameters can be used. Using this type for subtype checks is incorrect. Taking the example of `MutableMap`: `MutableMap<K, V>` is not a subtype of `Map<K, V>` because although the class `MutableMap` is a subclass of `Map`, the type arguments `K` and `V` in the left type are not the subtypes of the type arguments `K` and `V` in the right type, because they're different `K` and `V`: the ones on the left are from `MutableMap` and the ones on the right are from `Map`. Of course, they aren't compatible in any way because these are type parameters declared in different declarations, and there's no explicit relation between them. You can see how using default type makes little sense in this type of code.

I suggest you use `isSubclassOf` instead, which checks that out of any types in the hierarchy of the given type, _some_ type has the given class as the type constructor.

idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -83,13 +85,6 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
...
-        val constructor = type.constructor
-        if (constructor.declarationDescriptor?.name?.asString()?.endsWith("Map") != true) return false
-        return type.isSubtypeOf(map) || constructor.supertypes.any { isMap(it.constructor.declarationDescriptor?.defaultType) }
-    }
-    return isMap(this)
-}
\ No newline at end of file
+    val classDescriptor = this?.constructor?.declarationDescriptor as? ClassDescriptor ?: return false
+    return classDescriptor.name.asString().endsWith("Map") && classDescriptor.isSubclassOf(builtIns.map)
+}
```
udalov: 👍

PR #1759, opened 2018-07-10
KT-13854 Need intention actions: to convert property with getter to initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13854


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPropertyGetterToInitializerIntention.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.intellij.openapi.editor.Editor
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+
+class ConvertPropertyGetterToInitializerIntention : SelfTargetingIntention<KtPropertyAccessor>(
+    KtPropertyAccessor::class.java, "Convert property getter to initializer"
+) {
+
+    override fun isApplicableTo(element: KtPropertyAccessor, caretOffset: Int): Boolean {
+        return element.isGetter && (element.parent as? KtProperty)?.hasInitializer() != true && element.singleExpression() != null
```
mglukhikh: There are properties which can have getter but cannot have initializer, e.g. interface properties. Please add extra check and tests for these cases.

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPropertyGetterToInitializerIntention.kt
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    override fun applyTo(element: KtPropertyAccessor, editor: Editor?) {
+        val property = element.parent as? KtProperty ?: return
+        property.initializer = element.singleExpression()
+        property.deleteChildRange(property.initializer?.nextSibling, element)
+        editor?.caretModel?.moveToOffset(property.endOffset)
```
mglukhikh: Please check how comments are handled here.

idea/testData/intentions/convertPropertyGetterToInitializer/expectedProperty.kt
```diff
@@ -0,0 +1,11 @@
+// ENABLE_MULTIPLATFORM
+// ERROR: Expected declaration must not have a body
+// IS_APPLICABLE: false
+expect class C {
+    val p: Int
+        <caret>get() = 1
```
mglukhikh: For multi-platform testing, we have a specific group of tests. In your case, `multiModuleQuickFix` group should work properly (it should work for intentions too). This should also solve your problem with target platform detection.

mglukhikh:
Thank you, merged with some corrections.

PR #1760, opened 2018-07-11
KT-18807 Complete Current Statement doesn't work in function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18807

### DIFF
--- a/idea/tests/org/jetbrains/kotlin/idea/codeInsight/smartEnter/SmartEnterTest.kt
+++ b/idea/tests/org/jetbrains/kotlin/idea/codeInsight/smartEnter/SmartEnterTest.kt
@@ -1460,7 +1460,7 @@ class SmartEnterTest : KotlinLightCodeInsightFixtureTestCase() {
         """
         fun foo(i: Int) = 1
         fun test4() {
-            foo(1,<caret> )
+            foo(1, <caret>)
         }
         """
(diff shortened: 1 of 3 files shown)


idea/tests/org/jetbrains/kotlin/idea/codeInsight/smartEnter/SmartEnterTest.kt
```diff
@@ -1405,6 +1405,81 @@ class SmartEnterTest : KotlinLightCodeInsightFixtureTestCase() {
...
+        """
+        fun foo(i: Int) = 1
+        fun test4() {
+            foo(1,<caret>
+        }
+        """,
+        """
+        fun foo(i: Int) = 1
+        fun test4() {
+            foo(1,<caret> )
```
goodwinnk: This test looks strange. Please try this example as well, the caret is shifted after smart enter, and it doesn't look as expected behavior. 

```
fun foo(i: Int) = 1
fun test4() {
    foo(1, <caret>
}
```

t-kameyama:
Fixed

goodwinnk:
I have merged the commits manually as https://github.com/JetBrains/kotlin/commit/2547612d54b606858e0232796d7884cd0a3515d2. Thank you!

PR #1761, opened 2018-07-11
KT-7922 Added Boolean Companion Object

Kerooker:
https://youtrack.jetbrains.com/issue/KT-7922

### DIFF
--- a/core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/ClassMapperLite.kt
+++ b/core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/ClassMapperLite.kt
@@ -63,7 +63,7 @@ object ClassMapperLite {
             add("reflect/KFunction$i", "kotlin/reflect/KFunction")
         }
 
-        for (klass in listOf("Char", "Byte", "Short", "Int", "Float", "Long", "Double", "String", "Enum")) {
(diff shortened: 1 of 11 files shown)


Kerooker:
@ilya-g

core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/ClassMapperLite.kt
```diff
@@ -63,7 +63,7 @@ object ClassMapperLite {
             add("reflect/KFunction$i", "kotlin/reflect/KFunction")
         }
 
-        for (klass in listOf("Char", "Byte", "Short", "Int", "Float", "Long", "Double", "String", "Enum")) {
+        for (klass in listOf("Char", "Byte", "Short", "Int", "Float", "Long", "Double", "String", "Enum", "Boolean")) {
```
udalov: This change will probably confuse the compiler and reflection of older versions when they encounter a declaration with `kotlin.Boolean.Companion` in the signature. The generated bytecode will be correct, but annotations (such as `Deprecated`, for example) will not be loaded, and reflective call will throw an exception. For more information, see the comment on this class.

Please revert this change and add a short comment noting that we explicitly omit `Boolean` here. Thanks!

Kerooker: Done

ilya-g:
I've fixed remaining tests and merged it into 1.3-M1 branch

PR #1762, opened 2018-07-14
Create sample to String.toLowerCase

bloderxd:


bloderxd:
@ilya-g Could you review again, please?

libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
```diff
@@ -94,12 +94,16 @@ public actual fun String.replaceFirst(oldValue: String, newValue: String, ignore
 
 /**
  * Returns a copy of this string converted to upper case using the rules of the default locale.
+ *
+ * @sample samples.text.Strings.toUpperCase
```
ilya-g: There's `expect` counterpart and another `actual` implementation for JS, it's worth to place a sample reference there too.

bloderxd:
@ilya-g What do you think now?

ilya-g:
Squashed and merged, thank you for the PR

PR #1763, opened 2018-07-16
KT-22068 Force usage of “it” in .forEach{} calls

cornwe19:
https://youtrack.jetbrains.com/issue/KT-22068

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
@@ -18,7 +18,7 @@ import org.jetbrains.kotlin.psi.callExpressionVisitor
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 
-class ItNotUsedInLoopInspection : AbstractKotlinInspection() {
+class ForEachParameterNotUsed : AbstractKotlinInspection() {
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
         return callExpressionVisitor {
             if (it.calleeExpression?.text == "forEach") {
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
@@ -10,28 +10,38 @@ import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.psi.PsiElementVisitor
 import org.jetbrains.kotlin.descriptors.VariableDescriptor
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.intentions.getCallableDescriptor
 import org.jetbrains.kotlin.idea.refactoring.getThisLabelName
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.KtVisitorVoid
 import org.jetbrains.kotlin.psi.callExpressionVisitor
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 
 class ForEachParameterNotUsed : AbstractKotlinInspection() {
+    companion object {
+        private val COLLECTIONS_FOREACH_FQNAME = FqName("kotlin.collections.forEach")
+        private val SEQUENCES_FOREACH_FQNAME = FqName("kotlin.sequences.forEach")
+    }
+
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
         return callExpressionVisitor {
-            if (it.calleeExpression?.text == "forEach") {
-                val lambda = it.lambdaArguments.singleOrNull()?.getLambdaExpression() ?: return@callExpressionVisitor
-                val descriptor = lambda.analyze()[BindingContext.FUNCTION, lambda.functionLiteral] ?: return@callExpressionVisitor
-                val iterableParameter = descriptor.valueParameters.singleOrNull() ?: return@callExpressionVisitor
+            when(it.getCallableDescriptor()?.fqNameOrNull()) {
+                COLLECTIONS_FOREACH_FQNAME, SEQUENCES_FOREACH_FQNAME -> {
+                    val lambda = it.lambdaArguments.singleOrNull()?.getLambdaExpression() ?: return@callExpressionVisitor
+                    val descriptor = lambda.analyze()[BindingContext.FUNCTION, lambda.functionLiteral] ?: return@callExpressionVisitor
+                    val iterableParameter = descriptor.valueParameters.singleOrNull() ?: return@callExpressionVisitor
 
-                if (lambda.bodyExpression?.usesDescriptor(iterableParameter) != true) {
-                    holder.registerProblem(
-                        it.calleeExpression!!,
-                        "Loop parameter '${iterableParameter.getThisLabelName()}' is unused",
-                        ProblemHighlightType.GENERIC_ERROR_OR_WARNING
-                    )
+                    if (lambda.bodyExpression?.usesDescriptor(iterableParameter) != true && it.calleeExpression != null) {
+                        holder.registerProblem(
+                            it.calleeExpression!!,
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return callExpressionVisitor {
+            if (it.calleeExpression?.text == "forEach") {
+                val lambda = it.lambdaArguments.singleOrNull()?.getLambdaExpression() ?: return@callExpressionVisitor
+                val descriptor = lambda.analyze()[BindingContext.FUNCTION, lambda.functionLiteral] ?: return@callExpressionVisitor
+                val iterableParameter = descriptor.valueParameters.singleOrNull() ?: return@callExpressionVisitor
+
+                if (lambda.bodyExpression?.usesDescriptor(iterableParameter) != true) {
+                    holder.registerProblem(
+                        lambda,
```
mglukhikh: It's better to highlight just `forEach` name and not the whole lambda.

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.jetbrains.kotlin.psi.KtVisitorVoid
+import org.jetbrains.kotlin.psi.callExpressionVisitor
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+
+class ForEachParameterNotUsed : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return callExpressionVisitor {
+            if (it.calleeExpression?.text == "forEach") {
+                val lambda = it.lambdaArguments.singleOrNull()?.getLambdaExpression() ?: return@callExpressionVisitor
```
mglukhikh: Before it, I'd obtain a descriptor for `forEach` call and make sure that it's `forEach` from standard library, defined either in `kotlin.collections` or `kotlin.sequences`. Otherwise we have a risk of reporting this inspection on wrong call (some user-defined `forEach`).

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
mglukhikh: It's better to highlight just `forEach` name and not the whole lambda.

idea/testData/inspections/forEachParameterNotUsed/test.kt
```diff
@@ -0,0 +1,9 @@
+fun test() {
+    val items = listOf<Any>()
+    items.forEach { }
+    items.forEach { item -> }
+    items.forEach { doSomething(it) }
+    items.forEach { item -> doSomething(item) }
+}
```
mglukhikh: I'd add some tests with nested `forEach`, to make sure we handle correctly situations with multiple `it`.

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
cornwe19: Validated this with the declaration's FqName

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
cornwe19: Highlighting only the calleeExpression now.

idea/testData/inspections/forEachParameterNotUsed/test.kt
cornwe19: Validated nested calls. No logic needed to change to support this, it looks like the scope was already being handled properly.

cornwe19:
@mglukhikh Should be all set for another look. Thanks for the feedback.

mglukhikh:
Thank you, everything looks good. I will perform the final testing and accept it, either with some fixes or not.

cornwe19:
Thanks!

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtBlockExpression
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtVisitorVoid
+import org.jetbrains.kotlin.psi.callExpressionVisitor
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+
+class ForEachParameterNotUsed : AbstractKotlinInspection() {
```
mglukhikh: Most inspection classes have `Inspection` suffix. Please add it.

idea/src/org/jetbrains/kotlin/idea/inspections/ForEachParameterNotUsed.kt
```diff
@@ -0,0 +1,68 @@
...
+    companion object {
+        private val COLLECTIONS_FOREACH_FQNAME = FqName("kotlin.collections.forEach")
+        private val SEQUENCES_FOREACH_FQNAME = FqName("kotlin.sequences.forEach")
+    }
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return callExpressionVisitor {
+            when(it.getCallableDescriptor()?.fqNameOrNull()) {
```
mglukhikh: For performance reasons, it's a bad idea to get descriptor at the very first line. Here I recommend to check name first via PSI, and only after it check name via descriptors.

mglukhikh: Also, space is required after `when` keyword (code style requirements)

idea/testData/inspections/forEachParameterNotUsed/inspectionData/inspections.test
```diff
@@ -0,0 +1 @@
+// INSPECTION_CLASS: org.jetbrains.kotlin.idea.inspections.ForEachParameterNotUsed
```
mglukhikh: Here you have to add `// WITH_RUNTIME` line, otherwise all your tests fail. In the future, please re-run your own tests before re-submitting PR.

mglukhikh: (comment: this line adds full Kotlin SDK to test project, without it `forEach` function is not visible).

mglukhikh:
Final testing revealed important problems, sorry (I can fix it myself if you wish).

cornwe19:
@mglukhikh no problem, I can get those issues patched up a little later today. Thanks for the detailed feedback and good call about the performance issue.

mglukhikh:
Also, I have recently found that this inspection is reported when destructuring declaration is used inside `forEach` lambda, no matter whether its components are used or not. Example:

```
fun foo(map: Map<String, String>) {
    map.forEach { (t, u) -> 
        println("$t: $u")
    }
}
```

cornwe19:
Good catch. I'll include that with the other fixes you mentioned yesterday. Thanks for the continued testing.

mglukhikh:
And yet another revealed false positive:

```
class My {
    operator fun invoke() /* can include parameters, no matter */ {}
}

fun bar(my: List<My>) {
    my.forEach { it() }
}
```

mglukhikh:
Meanwhile, I found one true positive in Kotlin project (ConvertSealedClassToObjectFix):

```
    private fun KtClass.changeToObject(factory: KtPsiFactory) {
        getClassOrInterfaceKeyword()?.replace(factory.createExpression(KtTokens.OBJECT_KEYWORD.value))
        secondaryConstructors.forEach { delete() } // should be it.delete()
        primaryConstructor?.delete()
    }
```

So thank you for your work!

cornwe19:
Awesome! Credit really goes to Christoffer for requesting the inspection. Seems to be a fairly common source of bugs.

I'll have to do a deeper dive on some of the false positives that are cropping up. There are a few scenarios here I don't use in my day to day development that I didn't think of.

mglukhikh:
I hope now it's the final state :) Thank you.

cornwe19:
Almost! I'm still working on detecting the invokable reference properly. Everything else should be all set through.

cornwe19:
@mglukhikh I think this is all set now. Thanks for all the feedback.

mglukhikh:
Finally merged it. Thank you for your efforts!

cornwe19:
Awesome! Happy to help

PR #1765, opened 2018-07-17
KT-17687 Quickfix for "Interface doesn't have constructors" to convert to anonymous object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17687


asedunov:
Thanks! We'll merge it manually

PR #1767, opened 2018-07-19
KT-20344 Unused private setter created for property

fabianishere:
This pull request will prevent the compiler for generating Java bytecode for
private property setters that are trivial.

Since Kotlin uses direct field access for private properties, it will result
in the private setter never been used and since it cannot be accessed by any
other class without reflection, the setter cannot reasonably be covered by code
coverage tools.

See https://youtrack.jetbrains.com/issue/KT-20344 for the related YouTrack issue.


udalov:
Thank you for the contribution and sorry for the long delay! I've merged your commit in 05f6ed40f1, and used a new test instead of the obsolete `PropertyGenTest` in 5fb1bbe3f3. The fix will be available since Kotlin 1.3.20.

PR #1768, opened 2018-07-19
KT-6633 Inspection to detect unnecessary "with" calls

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-6633


t-kameyama:
I will add tests.

NataliaUkhorskaya:
Merged manually. Thank you!

PR #1769, opened 2018-07-22
KT-1440 added bitwise operators

Redrield:
This PR adds several new overloadable operators to Kotlin. `|`, `&`, `^`, `<<`, `>>`, `>>>`, and assignment operators.

The infix functions that currently serve this functionality have been marked as deprecated in favour of these operators. Their functionality remains.

https://youtrack.jetbrains.com/issue/KT-1440


abreslav:
Hello,

We appreciate the amount of work you invested into this PR.

Unfortunately, it introduces changes to the language, and such things can not be done without a thorough discussion and coordination with the language designers. 
The main reason why the features you implemented are not in the language yet is that there are many questions about possible better uses of the syntactic elements they employ. There are very few special characters on our keyboards, so we have to be very cautious here. My personal hunch is that `|` and `&` have a decent chance of becoming overloadable operators in Kotlin at some point, but it’s less likely for `^`, `<<`, `>>`, `>>>`. In any case, to make such decisions we have to finalize a few other bigger features with high syntactic uncertainty: collection literals, some DSL-related features, etc.

Due to these reasons, this PR cannot be accepted at this time, so we will close it. Thanks for your time and effort, we regret that we can’t accept your work.

P.S. As a side note, I haven’t really run the tests, but from a glance at the code I suspect that this PR may break a lot of functionality around generics because things like `A<B<C>>` are now lexed in a different way than before.

Redrield:
@abreslav Would it be possible to reopen this PR if I were to open an equivalent KEEP for the language changes that are being introduced, and that is accepted by the language designers?

P.S. lexing for generics was fixed through commits fc49a45 and d83b045

abreslav:
When a KEEP gets accepted, of course, we can reopen this PR if it will still be relevant.

PR #1770, opened 2018-07-23
KT-25608 Confusing "Redundant override" inspection message

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25608


mglukhikh:
Thank you, merged.

mglukhikh:
I noticed a wording problem here, because in Kotlin we prefer using the term "function" or "member function" instead of the term "method" (http://kotlinlang.org/docs/reference/functions.html). Will think a little bit and change wording myself if decide to do so.

PR #1771, opened 2018-07-23
KT-20357 Add sample code for repeat

ymnder:


libraries/stdlib/samples/test/samples/misc/repeat.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun repeatAnyTimes() {
+        var sum10Times = 0
+        repeat(10) {
+            sum10Times += it
+        }
+        assertPrints(sum10Times, "45")
+
+        var sumZeroTimes = 0
+        repeat(0) {
+            sumZeroTimes += it
```
ilya-g: You can use some action here that normally fails, to emphasize that the lambda is not executed.

ilya-g:
Thanks. I've simplified the sample and merged it.

PR #1772, opened 2018-07-23
[Kapt] Pass current ClassLoader as parent class loader by default (KT…

mmdango:
…-24540)
[Youtrack: KT-24540](https://youtrack.jetbrains.com/issue/KT-24540)

There is a bug with how Kapt creates the `ClassLoader` used to instantiate annotation processors, particularly the ones that depend on `tools.jar`. The `SystemClassLoader` is used by default as the parent `ClassLoader`, but it may or may not include `tools.jar`, which causes Kapt runtime errors. I've personally seen this manifest in a couple of different ways:

In this example: https://github.com/felipecsl/kapt-classloader-bug. This project is built with Gradle and has an annotation processor that depends on tools.jar. During Kapt, the `Trees` class is loaded twice (once in Kapt itself and once in the annotation processor) by two separate `ClassLoaders`, which causes a `ClassCastException` since a `Class` cannot be loaded by two different `ClassLoaders`.

In this example: https://github.com/mmdango/okbuck_kapt_mwe. The project is built with Buck and uses Butterknife which depends on `tools.jar`. In this case, the `ClassLoader` that Kapt creates doesn't have `tools.jar` at all, causing a `ClassNotFoundException`.

I'm proposing to pass in by default the current `ClassLoader` as the parent `ClassLoader` for annotation processors. This fixes the errors seen in the above projects.

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
@@ -21,7 +21,7 @@ class ProcessorLoader(
 ) : Closeable {
     private var annotationProcessingClassLoader: URLClassLoader? = null
 
-    fun loadProcessors(parentClassLoader: ClassLoader = ClassLoader.getSystemClassLoader()): List<Processor> {
+    fun loadProcessors(parentClassLoader: ClassLoader = javaClass.classLoader): List<Processor> {
         clearJarURLCache()
 
         val classpath = (paths.annotationProcessingClasspath + paths.compileClasspath).distinct()


mmdango:
@yanex Any chance you could take a look at this soon?

plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
```diff
@@ -21,7 +21,7 @@ class ProcessorLoader(
 ) : Closeable {
     private var annotationProcessingClassLoader: URLClassLoader? = null
 
-    fun loadProcessors(parentClassLoader: ClassLoader = ClassLoader.getSystemClassLoader()): List<Processor> {
+    fun loadProcessors(parentClassLoader: ClassLoader = javaClass.classLoader): List<Processor> {
```
artem-zinnatullin: Can you introspect that please? I'm afraid that'd pass classpath of whole kapt plugin here

You probably want something more isolated that only have `tools.jar` classes

hsyed: To ensure a pristine classpath. I think the parent class loader should be `null` on jdk8 (boot classpath) and use `Classloader.getPlatformClassloader` on jdk9. 

I don't know if the boot classpath has `tools.jar` loaded though, if it doesn't I think a compat classloader should be created for jdk8.

hsyed: @yanex could we prioritise a fix for this -- it's quite problematic. What about using this to select the default ? 

```
import java.net.URLClassLoader
import java.nio.file.Paths

/**
 * Select a clean parent classloader containing the java compiler. The classloader must not descend from the system classpath. [candidate]
 * should generally be set to some class from the compiler distribution (for example kotlin-compiler.jar or kotlin-annotation-processing.jar)
 * to enable efficient use of a user prepared classloader.
 *
 * @param candidate an optional context classloader. If this is provided then it is first checked to see if it contains the compiler
 *  classes.
 * @throws IllegalStateException if [candidate] descends from the system classloader.
 */
@Throws(IllegalStateException::class)
fun selectJavacClassloader(candidate: ClassLoader? = null): ClassLoader {
    fun ClassLoader.descendsFromSystemClassloader(): Boolean =
        when (this.parent) {
            null -> false
            ClassLoader.getSystemClassLoader() -> true
            else -> this.parent.descendsFromSystemClassloader()
        }


    if(candidate != null) {
        try {
            candidate.loadClass("com.sun.tools.javac.Main")
            check(!candidate.descendsFromSystemClassloader()) {
                "compiler context classloader descends from the system classloader"
            }
            return candidate
        } catch (ex: ClassNotFoundException) {
        }
    }

    // If jdk9 the platform classloader should provide access to the compiler classes.
    try {
        return ClassLoader::class.java.getMethod("getPlatformClassLoader").invoke(null) as ClassLoader
    } catch (t: NoSuchElementException) {
    }

    // Not on JDK9 so create a pristine classloader descending from the boot classloader.
    return URLClassLoader(
        arrayOf(
            Paths.get(System.getProperty("java.home")).let { path ->
                path.takeIf { !it.endsWith(Paths.get("jre")) } ?: path.parent
            }.resolve(Paths.get("lib", "tools.jar"))
                .toFile().also { f -> check(f.exists()) }.toURI().toURL()
        ),
        null
    )
}
```

The bazel worker currently loads kotlin-compiler.jar and kotlin-annotation-processing.jar into a Preloaded classloader along with tools.jar (if not jdk9).

mmdango: @hsyed Does this fix the scenario described in https://github.com/felipecsl/kapt-classloader-bug?

The cause of that bug is that classes from `tools.jar` get loaded from two independent classloaders, wouldn't creating a new classloader with no parent cause the same error?

hsyed: I haven't tested the code -- I don't generally develop with gradle or build the kotlin compiler jar (yet). The solution I presented is to prevent the system classloader being a descendant of the kapt classloader -- which is needed to prevent collisions with the kotlin bazel builder classpath.

I've done a once over for your problem statement. I'll commnet on this part now:

> The project is built with Buck and uses Butterknife which depends on tools.jar. In this case, the ClassLoader that Kapt creates doesn't have tools.jar at all, causing a ClassNotFoundException.

The kotlin builder:

* loads `kotlin-compiler.jar` and `tools.jar` into a disjoint classloader on jdk8.
* loads `kotlin-compiler.jar` and inherits from `platformClassloader` on jdk9.

The buck kotlin integration might need to be updated similarly ?

mmdango:
@yanex @hsyed any movement on this on your ends?

yanex:
I don't like the `javaClass.classLoader` approach as it includes kotlinc+kapt classes, and I don't see any good reason for allowing annotation processors to have access to kotlinc.
Looks like we need to change our ClassLoader hierarchy inside the compiler daemon to fix this. I hope we will manage to do it in the scope of 1.3.20 (1.3.10 is reserved for the critical issue fixes).

mmdango:
@yanex Would you be able to comment if a fix for this was made in 1.3.20?

nskpaun:
@yanex I'd be happy to help out with this if you can give some code pointers and an approach

nskpaun:
Found a workaround: https://youtrack.jetbrains.com/issue/KT-24540#focus=streamItem-27-3518069.0-0

PR #1773, opened 2018-07-24
KT-25579 Redundant semicolon erroneously reported during local var modifier ambiguity

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25579


mglukhikh:
Thank you, merged.

PR #1774, opened 2018-07-26
K/N: Properly add modules for KLIBs generated by cinterop

ddolovov:
1. Don't need to append "/linkdata" to the end of KLIB path
2. Add modules for all KLIBs (both directory- and file-based) found on Gradle dependencies list

### DIFF
--- a/buildSrc/src/main/kotlin/intellijDependencies.kt
+++ b/buildSrc/src/main/kotlin/intellijDependencies.kt
@@ -130,7 +130,7 @@ fun Project.runIdeTask(name: String, ideaPluginDir: File, ideaSandboxDir: File, 
         workingDir = File(intellijRootDir(), "bin")
 
         jvmArgs(
-            "-Xmx1250m",
+            "-Xmx2048m",
             "-XX:ReservedCodeCacheSize=240m",
             "-XX:+HeapDumpOnOutOfMemoryError",
             "-ea",
@@ -142,7 +142,8 @@ fun Project.runIdeTask(name: String, ideaPluginDir: File, ideaSandboxDir: File, 
             "-Dapple.awt.graphics.UseQuartz=true",
             "-Dsun.io.useCanonCaches=false",
             "-Dplugin.path=${ideaPluginDir.absolutePath}",
-            "-Didea.additional.classpath=../idea-kotlin-runtime/kotlin-runtime.jar,../idea-kotlin-runtime/kotlin-reflect.jar"
+            "-Didea.additional.classpath=../idea-kotlin-runtime/kotlin-runtime.jar,../idea-kotlin-runtime/kotlin-reflect.jar",
+            "-Didea.max.intellisense.filesize=5000"
         )
 
         if (rootProject.findProperty("versions.androidStudioRelease") != null) {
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/intellijDependencies.kt
```diff
@@ -142,7 +142,8 @@ fun Project.runIdeTask(name: String, ideaPluginDir: File, ideaSandboxDir: File,
             "-Dapple.awt.graphics.UseQuartz=true",
             "-Dsun.io.useCanonCaches=false",
             "-Dplugin.path=${ideaPluginDir.absolutePath}",
-            "-Didea.additional.classpath=../idea-kotlin-runtime/kotlin-runtime.jar,../idea-kotlin-runtime/kotlin-reflect.jar"
+            "-Didea.additional.classpath=../idea-kotlin-runtime/kotlin-runtime.jar,../idea-kotlin-runtime/kotlin-reflect.jar",
+            "-Didea.max.intellisense.filesize=5000"
```
stigger: I don't think these changes should be committed.

idea/idea-native/src/org/jetbrains/konan/analyser/KonanAnalyzerFacade.kt
```diff
@@ -61,66 +63,76 @@ class KonanAnalyzerFacade : ResolverForModuleFactory() {
...
     val fragmentProviders = mutableListOf(packageFragmentProvider)
-    descriptors.mapTo(fragmentProviders) { it.packageFragmentProvider }
+    libraryDescriptors.mapTo(fragmentProviders) { it.packageFragmentProvider }
     fragmentProviders.add(forwardDeclarationsModule.packageFragmentProvider)
-
     return ResolverForModule(CompositePackageFragmentProvider(fragmentProviders), container)
   }
 
-  private fun findLibrary(module: Module, path: Path): Library? {
+  private fun findLibrary(module: Module, libraryPath: File): Library? {
```
stigger: Is this method still needed?

idea/idea-native/src/org/jetbrains/konan/analyser/KonanAnalyzerFacade.kt
```diff
@@ -61,66 +63,76 @@ class KonanAnalyzerFacade : ResolverForModuleFactory() {
...
+        libraryPaths.forEach { createDescriptor(it)?.also { libraryDescriptors.add(it) } }
+
+        true // continue the loop while there are more libs
+      }
+    }
 
-    val dependencies = descriptors.toMutableList()
-    if (stdlibDescriptor != null)
-      dependencies.add(stdlibDescriptor)
+    val stdlibDescriptor = konanPaths.konanStdlib()?.let { createDescriptor(it.toFile()) }
```
stigger: At this point `libraryDescriptors` should already contain a descriptor for stdlib.

buildSrc/src/main/kotlin/intellijDependencies.kt
ddolovov: OK. reverted back

idea/idea-native/src/org/jetbrains/konan/analyser/KonanAnalyzerFacade.kt
ddolovov: dropped

idea/idea-native/src/org/jetbrains/konan/analyser/KonanAnalyzerFacade.kt
ddolovov: rewritten in another way

PR #1776, opened 2018-07-29
Added an extension function on Char objects to get numeric value.

AdamMc331:
Ticket: https://youtrack.jetbrains.net/issue/KT-21569


AdamMc331:
@ilya-g Following up on this to see if you have any questions. Let me know if there are changes that should be made. :)

AdamMc331:
Thanks for the update! Will follow that conversation.

AdamMc331:
@ilya-g popping in and rebased this branch. It looks like there's still no decision on the attached ticket. 

Figured I would nudge this and say I'm still happy to contribute this change and make any necessary adjustments. Let me know what I can do to help.

ilya-g:
Thanks for PR, but we've decided to implement the functions to get the digit value of char a bit different in JVM: basically they will use https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#digit-char-int- under the hood.

PR #1777, opened 2018-07-30
KT-25228 Create function '*' from a protected inline method should not produce a private method

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25228


asedunov:
Merged manually. Thank you!

PR #1778, opened 2018-07-30
KT-25739 "Convert to run" / "Convert to with" intentions incorrectly process references to Java static members

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25739


mglukhikh:
Thank you, merged

PR #1780, opened 2018-07-31
KT-20626 Inspection for '+= creates a new list under the hood'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20626


idea/src/META-INF/plugin.xml
```diff
@@ -2985,6 +2985,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.SuspiciousCollectionReassignmentInspection"
+                     displayName="Suspicious collection reassignment by augmented assignment"
```
mglukhikh: I don't like this description. It says nothing to normal user. I think we should say something about list under the hood, at least.

idea/resources/inspectionDescriptions/SuspiciousCollectionReassignment.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection reports suspicious collection reassignment by augmented assignment.
```
mglukhikh: I think we need more detailed explanation here.

idea/src/org/jetbrains/kotlin/idea/inspections/SuspiciousCollectionReassignmentInspection.kt
```diff
@@ -0,0 +1,70 @@
...
+            if ((left.mainReference?.resolve() as? KtProperty)?.isVar != true) return
+
+            val context = binaryExpression.analyze(BodyResolveMode.PARTIAL)
+            val type = left.getType(context)?.constructor?.declarationDescriptor?.defaultType ?: return
+            val builtIns = binaryExpression.builtIns
+            if (type != builtIns.list.defaultType && type != builtIns.set.defaultType && type != builtIns.map.defaultType) return
+
+            holder.registerProblem(
+                binaryExpression,
+                "'${left.text}' is reassigned by augmented assignment",
```
mglukhikh: This description is bad for the same reasons as above.

idea/src/org/jetbrains/kotlin/idea/inspections/SuspiciousCollectionReassignmentInspection.kt
```diff
@@ -0,0 +1,70 @@
...
+                ProblemHighlightType.LIKE_UNUSED_SYMBOL,
+                IntroduceLocalVariableFix()
```
mglukhikh: I'd consider other fixes, like "change type to mutable list"

idea/testData/inspectionsLocal/suspiciousCollectionReassignment/mutableList.kt
```diff
@@ -0,0 +1,7 @@
+// PROBLEM: none
+// ERROR: Assignment operators ambiguity: <br>public operator fun <T> Collection<Int>.plus(element: Int): List<Int> defined in kotlin.collections<br>@InlineOnly public inline operator fun <T> MutableCollection<in Int>.plusAssign(element: Int): Unit defined in kotlin.collections
+// WITH_RUNTIME
+fun test() {
+    var list = mutableListOf(1)
+    <caret>list += 2
```
mglukhikh: Just as an idea: why don't fix it by changing `var` to `val`?

t-kameyama: Created: https://youtrack.jetbrains.com/issue/KT-26236

mglukhikh:
Thank you. I will take this inspection on remote run and complex testing.

mglukhikh:
Ok, I approved this. I consider this inspection also a bit questionable (that's why I'm holding this PR for so long time), but yet I think we can accept it but should consider all suggested quick-fixes carefully (and probably do not highlight anything in situations we have nothing to suggest). At this moment fix to `list = list + ...` is considered as most reliable and understandable (it shows explicitly that the list is changed, and we should not suggest "replace with +=" for this particular case). Will look a bit how it works in practice and either fix myself or return back to you.

mglukhikh:
Finally merged this (manually).

PR #1781, opened 2018-07-31
Add documentation for @JvmSynthetic

artem-zinnatullin:
Docs on this annotation were missing since its addition to the stdlib & compiler support.

This PR adds missing documentation while highlighting that the annotation is not intended for mass use, but rather for *rare cases*. 

I've tried to keep the style of the text consistent with other documentation within the file.

Use case: https://github.com/gojuno/koptional/pull/30


ilya-g:
Please delimit summary from the detail description with a blank line. You can also use blank lines to separate paragraphs in the description.

artem-zinnatullin:
@ilya-g I've used blank lines as you suggested 👍

ilya-g:
Thanks, I've merged it.

PR #1782, opened 2018-08-01
KT-16422 Replace lambda with method reference inspections fails

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16422


mglukhikh:
Thank you, merged

PR #1785, opened 2018-08-02
KT-21999 Convert lambda to reference adds this with incorrect label

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21999


mglukhikh:
Thank you, merged.

PR #1787, opened 2018-08-03
KT-18067 KotlinJS - String.compareTo(other: String, ignoreCase: Boolean = false): Int

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18067

### DIFF
--- a/libraries/stdlib/js/src/kotlin/string.kt
+++ b/libraries/stdlib/js/src/kotlin/string.kt
@@ -65,8 +65,7 @@ public inline val CharSequence.size: Int get() = length
 @kotlin.internal.InlineOnly
 internal inline fun String.nativeReplace(pattern: RegExp, replacement: String): String = asDynamic().replace(pattern, replacement)
 
-@kotlin.internal.InlineOnly
-public actual inline fun String.compareTo(other: String, ignoreCase: Boolean): Int {
+public actual fun String.compareTo(other: String, ignoreCase: Boolean): Int {
     if (ignoreCase) {
         val n1 = this.length
         val n2 = other.length
--- a/libraries/stdlib/js/src/kotlin/string.kt
+++ b/libraries/stdlib/js/src/kotlin/string.kt
@@ -6,7 +6,6 @@
 package kotlin.text
 
 import kotlin.js.RegExp
-import kotlin.math.min
 
 /**
  * Converts the characters in the specified array to a string.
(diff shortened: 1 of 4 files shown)


ilya-g:
Could you elaborate why it is upper case comparison and not the lower case or both of them?

Could you ensure that the order of `"A"`, `"a"` and `"["` is the same according to the proposed case-insensitive comparison on both platforms?

Look for the hints in the contract of `compareToIgnoreCase`, `regionMatches` functions https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#compareToIgnoreCase-java.lang.String-

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -64,3 +65,30 @@ public inline val CharSequence.size: Int get() = length
...
 @kotlin.internal.InlineOnly
 internal inline fun String.nativeReplace(pattern: RegExp, replacement: String): String = asDynamic().replace(pattern, replacement)
+
+@kotlin.internal.InlineOnly
+public actual inline fun String.compareTo(other: String, ignoreCase: Boolean): Int {
+    if (ignoreCase) {
+        val n1 = this.length
+        val n2 = other.length
+        val min = min(n1, n2) - 1
+        for (i in 0..min) {
```
ilya-g: You can use `until` instead of subtracting 1 from the end.

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -6,6 +6,7 @@
 package kotlin.text
 
 import kotlin.js.RegExp
+import kotlin.math.min
```
ilya-g: `minOf` is available without import

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -64,3 +65,30 @@ public inline val CharSequence.size: Int get() = length
...
+                if (c1 != c2) {
+                    c1 = c1.toLowerCase()
+                    c2 = c2.toLowerCase()
+                    if (c1 != c2) {
+                        return c1 - c2
+                    }
+                }
+            }
+        }
+        return n1 - n2
```
ilya-g: I doubt that this method of comparison is efficient, given how `Char` type and its `toUpper/LowerCase` extensions are implemented in JS. Here we make char code -> 1 char string -> upper/lower case -> back to char code conversion two times for each char. At least we can take a substring of length 1 instead of a char and avoid char code <-> string conversions.

Have you considered converting the entire strings `toUpperCase().toLowerCase()` and then comparing them? That however will probably be less efficient for long strings that differ in first several chars. 
 
Maybe we can apply some compromise approach so that the strings are split to batches each of several chars long and than each batch is transformed (toUpper -> toLower) and compared?

libraries/stdlib/js/src/kotlin/string.kt
```diff
@@ -64,3 +64,35 @@ public inline val CharSequence.size: Int get() = length
 
 @kotlin.internal.InlineOnly
 internal inline fun String.nativeReplace(pattern: RegExp, replacement: String): String = asDynamic().replace(pattern, replacement)
+
+@kotlin.internal.InlineOnly
+public actual inline fun String.compareTo(other: String, ignoreCase: Boolean): Int {
```
ilya-g: I don't think we should inline it, given its current size.

ilya-g:
Ok, I've merged it and implemented `String.CASE_INSENSITIVE_ORDER` comparator on top of that.
Thanks for the implementation.

PR #1788, opened 2018-08-03
Extension function similar to T.apply and T.run that returns Unit

bradynpoulsen:
Supporting implementation of [KT-25881](https://youtrack.jetbrains.com/issue/KT-25881)

### DIFF
--- a/libraries/stdlib/src/kotlin/util/Standard.kt
+++ b/libraries/stdlib/src/kotlin/util/Standard.kt
@@ -77,6 +77,17 @@ public inline fun <T> T.apply(block: T.() -> Unit): T {
 }
 
 /**
+ * Calls the specified function [block] with `this` value as its receiver and returns Unit.
+ */
+@kotlin.internal.InlineOnly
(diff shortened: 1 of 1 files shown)


libraries/stdlib/src/kotlin/util/Standard.kt
```diff
@@ -76,6 +76,17 @@ public inline fun <T> T.apply(block: T.() -> Unit): T {
     return this
 }
 
+/**
+ * Calls the specified function [block] with `this` value as its receiver and returns Unit.
+ */
+@kotlin.internal.InlineOnly
+public inline fun <T> T.unit(block: T.() -> Unit) {
```
bradynpoulsen: May a more appropriate name would be `scope`?

```kotlin
fun foo() {
    val list = listOf("one", "two", "three")
    list.scope {
        forEach {
            println("item: $it")
        }

        map(String::capitalize).forEach(::println)

        ...
    }
}
```

PR #1789, opened 2018-08-04
KT-23467 False positive `suspicious callable reference` on scoping function called on another lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23467


mglukhikh:
Merged, thank you

PR #1791, opened 2018-08-06
KT-20357 Add sample code for toList

takattata:

### DIFF
--- a/libraries/stdlib/samples/test/samples/misc/tuples.kt
+++ b/libraries/stdlib/samples/test/samples/misc/tuples.kt
@@ -22,21 +22,25 @@ class Tuples {
 
     @Sample
     fun pairToList() {
-        val list = Pair(1, "a").toList()
-        assertPrints(list, "[1, a]")
-        assertTrue(list[0] is Int)
-        assertTrue(list[1] is String)
-        assertFalse(list[1] is Int)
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/misc/tuples.kt
```diff
@@ -20,4 +20,23 @@ class Tuples {
         assertPrints(c, "[null]")
     }
 
+    @Sample
+    fun pairToList() {
+        val list = Pair(1, "a").toList()
```
ilya-g: It would be informative to show what is the static type of `list` by using explicit type specification
```
val list: List<...> = ...
```
It's worth to add another example, where both components have the same type.

libraries/stdlib/samples/test/samples/misc/tuples.kt
```diff
@@ -20,4 +20,23 @@ class Tuples {
...
+        assertPrints(list, "[1, a]")
+        assertTrue(list[0] is Int)
+        assertTrue(list[1] is String)
+        assertFalse(list[1] is Int)
```
ilya-g: I'm not sure this assertion adds something, after you've asserted that this element is `String`

ilya-g:
Thanks, I've merged it manually.

PR #1792, opened 2018-08-06
KT-25270 "return@foo" outside of lambda should have quickfix to remove "@foo" label

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25270


idea/testData/inspectionsLocal/redundantReturnLabel/inLambda.kt
```diff
@@ -0,0 +1,9 @@
+// PROBLEM: none
+fun foo(f: (String?) -> Int) {}
+
+fun test() {
+    foo {
+        if (it != null) return@foo<caret> 1
```
udalov: Please make sure that the inspection is not suggested when using the function expression syntax:
```
foo(fun(it: String?): Int {
    if (it != null) return@foo 1
    return 0
})
```

t-kameyama: Why we should not suggest?

udalov: Because this is valid code, unlike in the issue fixed in this pull request

mglukhikh:
Generally looks good, I will test the inspection a bit and then fix / merge it.

mglukhikh:
Finally merged, thank you

PR #1793, opened 2018-08-06
KT-20357 Add samples for padStart() and padEnd()

gen0083:

### DIFF
--- a/libraries/stdlib/samples/test/samples/text/strings.kt
+++ b/libraries/stdlib/samples/test/samples/text/strings.kt
@@ -104,8 +104,8 @@ class Strings {
 
     @Sample
     fun padStart() {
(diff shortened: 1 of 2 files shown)


gen0083:
Thank you for reviewing.
I updated to wrap single quotes before printing, is this ok?

ilya-g:
Now it's better, thanks. I'd rather rename `padWithoutChar` to `padWithSpace`, because it still is padded with char, just that char is space by default.

gen0083:
ok, done it 👍

gen0083:
hi, this pr seems to be already merged. so what should I do? should i close this?

ilya-g:
Thanks for the samples, we've merged them manually.
Now it's ok to close this.

PR #1796, opened 2018-08-08
KT-19668 Inspection "Redundant else in if"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19668


idea/src/META-INF/plugin.xml
```diff
@@ -2985,6 +2985,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.RedundantElseInIfInspection"
+                     displayName="Redundant 'else' in 'if'"
+                     groupPath="Kotlin"
+                     groupName="Redundant constructs"
+                     enabledByDefault="true"
+                     level="WEAK WARNING"
```
mglukhikh: I suggest it should be `INFORMATION` by default

idea/src/META-INF/plugin.xml
```diff
@@ -2985,6 +2985,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.RedundantElseInIfInspection"
+                     displayName="Redundant 'else' in 'if'"
+                     groupPath="Kotlin"
+                     groupName="Redundant constructs"
```
mglukhikh: Here I'd prefer just "Style issues"

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantElseInIfInspection.kt
```diff
@@ -0,0 +1,106 @@
...
+        ifExpressionVisitor(fun(ifExpression) {
+            if (ifExpression.isElseIf()) return
+            val elseKeyword = ifExpression.singleElseKeyword(true) ?: return
+            val rangeInElement = elseKeyword.textRange?.shiftRight(-ifExpression.startOffset) ?: return
+            holder.registerProblem(
+                holder.manager.createProblemDescriptor(
+                    ifExpression,
+                    rangeInElement,
+                    "Redundant 'else'",
+                    ProblemHighlightType.LIKE_UNUSED_SYMBOL,
```
mglukhikh: Don't report it as unused symbol, otherwise it looks like the whole `else` block is redundant which is not true.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantElseInIfInspection.kt
```diff
@@ -0,0 +1,106 @@
...
+        if (copy is KtBlockExpression) {
+            copy.lBrace?.delete()
+            copy.rBrace?.delete()
+        }
+        val added = ifExpression.parent.addAfter(copy, ifExpression)
+        elseExpression.delete()
+        elseKeyword.delete()
+        val nextSibling = added.nextSibling
+
+        if (nextSibling != null) {
```
mglukhikh: I don't like this piece of code. Are you sure it's required here? Please explain why.

t-kameyama: I would like to indent after deleting `else` block. Without this code, it looks like this:

before
```
fun test(x: Boolean) {
    if (x) {
        return
    } else<caret> {
        foo()
        bar()
    }
}
```

after
```
fun test(x: Boolean) {
    if (x) {
        return
    }
    foo()
        bar()
}
```

mglukhikh:
Thanks, taken on complex testing.

mglukhikh:
During complex testing, I noticed the following:

- fix works incorrectly for code like `if (...) return ... else return ...`, converting it to `if (...) return ... return ...`, I expect here a new line before second return
- sometimes (yet I don't understand when and why) it works incorrectly with long block of statements in else, shifting one tab back only first of them and keeping indent of all other statements
- in tailrec functions, it may be dangerous to use the fix (not quite sure, I should think about it)

@t-kameyama , you may try to investigate and/or fix some of these issues yourself, or I will do it when have some time.

mglukhikh:
Also we may have some performance problems here because when launched on Kotlin code base, the inspection eats all memory, CPU time and hangs IDEA (at least it happened in my only experiment). This can mean nothing, or can point on some important problem. To be investigated.

mglukhikh:
Ok, I've taken it on final testing again.

mglukhikh:
This problem is still intact: "it works incorrectly with long block of statements in else, shifting one tab back only first of them and keeping indent of all other statements". I cannot accept it until it's fixed.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantElseInIfInspection.kt
```diff
@@ -0,0 +1,115 @@
...
+            )
+        })
+}
+
+private class RemoveRedundantElseFix : LocalQuickFix {
+    override fun getName() = "Remove redundant 'else'"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
```
mglukhikh: This fix has a problem: it works incorrectly with long block of statements in else, shifting one tab back only first of them and keeping indent of all other statements. If you like, you can download custom plugin from here: https://teamcity.jetbrains.com/viewLog.html?buildId=1749508&buildTypeId=Kotlin_dev_CompilerAllPlugins&tab=artifacts (beware, after some days artifacts will perish) and test the fix just on Kotlin project (I ran the inspection on just idea module and applied all fixes automatically).

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantElseInIfInspection.kt
mglukhikh: Looks like it does not work as you want, see my title comment.

mglukhikh:
Ok, I'm taking this for testing third time :)

mglukhikh:
Finally merged. Thank you.

PR #1801, opened 2018-08-10
KT-25928 "Let extend" quick fix is suggested in case of nullable/non-null TYPE_MISMATCH collision

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25928


mglukhikh:
Thank you, merged.

PR #1802, opened 2018-08-11
KT-20273 Inspection to report a setter of a property with a backing field that doesn't update the backing field

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20273


idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+
+class SetterBackingFieldAssignmentInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return propertyAccessorVisitor(fun(accessor) {
+            if (!accessor.isSetter) return
+            if (accessor.hasLowerVisibilityThanProperty()) return
```
mglukhikh: Not quite sure is it really important. Could you explain why did you insert this check?

idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
```diff
@@ -0,0 +1,49 @@
...
+            val bodyExpression = accessor.bodyExpression as? KtBlockExpression ?: return
+            if (bodyExpression.findDescendantOfType<KtBinaryExpression> {
+                    it.operationToken == KtTokens.EQ && it.left?.text == "field"
```
mglukhikh: What do you think about other assignment operators, like `PLUSEQ`? I'm not sure exactly but think that with them we should report this.

idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.findDescendantOfType
+
+class SetterBackingFieldAssignmentInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return propertyAccessorVisitor(fun(accessor) {
+            if (!accessor.isSetter) return
+            if (accessor.hasLowerVisibilityThanProperty()) return
+            val bodyExpression = accessor.bodyExpression as? KtBlockExpression ?: return
+            if (bodyExpression.findDescendantOfType<KtBinaryExpression> {
```
mglukhikh: There is `anyDescendantOfType`, use it here please.

idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
t-kameyama: It is not necessary, I will remove it.

mglukhikh:
Thanks, taken on remote testing

idea/src/META-INF/plugin.xml
```diff
@@ -2988,6 +2988,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.SetterBackingFieldAssignmentInspection"
+                     displayName="Setter backing field should be assigned"
```
mglukhikh: I'd change name to something like "Existing backing field is not assigned by the setter" or "Setter for a property with backing field should assign backing field". Choose what you think is better, or suggest your own name (current name is grammatically incorrect).

idea/src/META-INF/plugin.xml
```diff
@@ -2988,6 +2988,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
...
+                     groupPath="Kotlin"
+                     groupName="Redundant constructs"
```
mglukhikh: It's not a redundant construct, it's much more likely to be in "Probable bugs".

idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+
+class SetterBackingFieldAssignmentInspection : AbstractKotlinInspection(), CleanupLocalInspectionTool {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
+        return propertyAccessorVisitor(fun(accessor) {
+            if (!accessor.isSetter) return
+            val bodyExpression = accessor.bodyExpression as? KtBlockExpression ?: return
+            val parameter = accessor.valueParameters.singleOrNull()
+            val parameterDescriptor = parameter?.descriptor
+            val context = accessor.analyze(BodyResolveMode.PARTIAL)
```
mglukhikh: You forget the most important check here. The property should have backing field for this inspection to be reported. Property may have no backing field at all, and in this case it's ok when its setter does not access backing field. Please add also relevant tests.

idea/src/org/jetbrains/kotlin/idea/inspections/SetterBackingFieldAssignmentInspection.kt
```diff
@@ -35,12 +36,18 @@ class SetterBackingFieldAssignmentInspection : AbstractKotlinInspection(), Clean
                                 arg.text == parameter?.text
                                         && arg.getArgumentExpression().getResolvedCall(context)?.resultingDescriptor == parameterDescriptor
                             }
+                        is KtNameReferenceExpression -> {
+                            hasBackingField = it.text == KtTokens.FIELD_KEYWORD.value
```
mglukhikh: There is `BindingContext.BACKING_FIELD_REQUIRED`, please use it and not a raw text check

mglukhikh:
Thank you, merged

PR #1803, opened 2018-08-13
Add sync after kotlin configured in Android Studio projects.

juan-chen:
Configuring Kotlin in Android Studio projects changes build scripts  and thus needs a gradle sync afterwards.


4u7:
Merged manually

PR #1804, opened 2018-08-14
KT-26042 False positive "Remove redundant '.let' call" for lambda with destructured arguments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26042


mglukhikh:
Thank you, merged with minor corrections.

PR #1805, opened 2018-08-14
Move some parts of Kotlin/Native into the main repo

ilmat192:


settings.gradle
```diff
@@ -230,6 +231,7 @@ project(':compiler:ir.psi2ir').projectDir = "$rootDir/compiler/ir/ir.psi2ir" as
 project(':compiler:ir.ir2cfg').projectDir = "$rootDir/compiler/ir/ir.ir2cfg" as File
 project(':compiler:ir.backend.common').projectDir = "$rootDir/compiler/ir/backend.common" as File
 project(':compiler:backend.js').projectDir = "$rootDir/compiler/ir/backend.js" as File
+project(':native:kotlin-native-utils').projectDir = "$rootDir/native/utils" as File
```
ddolovov: Lets double check if we are going to keep K/N stuff in "native" or "konan" folder.

olonho: Package names in compiler to be `konan`.

ilmat192: Package names are still `konan` and I'll rename a project directory (`native` -> `konan`). But what about the output jar? It will be published, and an artifctId `kotlin-native-utils` seems less confusing than `konan-utils`. What do you think?

ddolovov: I would suggest to keep `konan` for uniformity. @olonho what's your opinion?

ilmat192: Renamed in https://github.com/JetBrains/kotlin/pull/1811

ilmat192:
Closed since the same changes are included in https://github.com/JetBrains/kotlin/pull/1811

PR #1806, opened 2018-08-15
KT-25995 "Simplify comparison" should try to apply "Simplify if expression" when necessary

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25995

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
@@ -26,56 +26,76 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 class ConstantConditionIfInspection : AbstractKotlinInspection() {
 
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
-        return ifExpressionVisitor(fun(expression) {
-            val condition = expression.condition ?: return
+        return ifExpressionVisitor { expression -> registerProblem(expression, holder) }
+    }
 
-            val context = condition.analyze(BodyResolveMode.PARTIAL_WITH_CFA)
-            val constantValue = condition.constantBooleanValue(context) ?: return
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -26,56 +26,76 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
...
 
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
-        return ifExpressionVisitor(fun(expression) {
-            val condition = expression.condition ?: return
+        return ifExpressionVisitor { expression -> registerProblem(expression, holder) }
+    }
 
-            val context = condition.analyze(BodyResolveMode.PARTIAL_WITH_CFA)
-            val constantValue = condition.constantBooleanValue(context) ?: return
+    private fun registerProblem(expression: KtIfExpression, holder: ProblemsHolder?): List<ConstantConditionIfFix> {
```
mglukhikh: I don't like the name here (because it associates directly with IDEA `registerProblem`). Something like `getFixesAndRegisterProblem` or `collectFixesAndRegisterProblem` would be better here.

idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -26,56 +26,76 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
...
+        holder?.registerProblem(
+            condition,
+            "Condition is always '$constantValue'",
+            *fixes.toTypedArray()
+        )
+
+        return fixes
+    }
+
+    fun applyFixIfNeed(ifExpression: KtIfExpression) {
```
mglukhikh: I'd extract this and previous function to companion object, if possible. Also I don't like this name too. Something like `applySingleFixIfAny` or `applyFixIfSingle` would be better.

mglukhikh:
It looks now generally good. I will polish it a little bit and merge.

PR #1808, opened 2018-08-16
KT-26009 "Convert to 'also'" intention adds an extra `it` expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26009


asedunov:
Merged! Thank you

PR #1809, opened 2018-08-16
KT-26051 False positive "Redundant visibility modifier" for overridden protected property setter made public

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26051

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantVisibilityModifierInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantVisibilityModifierInspection.kt
@@ -7,9 +7,14 @@ package org.jetbrains.kotlin.idea.inspections
 
 import com.intellij.codeInspection.*
 import com.intellij.psi.PsiElementVisitor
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantVisibilityModifierInspection.kt
```diff
@@ -18,7 +23,18 @@ class RedundantVisibilityModifierInspection : AbstractKotlinInspection(), Cleanu
     override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean, session: LocalInspectionToolSession): PsiElementVisitor {
         return declarationVisitor { declaration ->
             val visibilityModifier = declaration.visibilityModifier() ?: return@declarationVisitor
-            val implicitVisibility = declaration.implicitVisibility()
+            val implicitVisibility = when {
```
mglukhikh: I'd say here we should fix `KtDeclaration.implicitVisibility()` itself and not insert some work-arounds just here.

mglukhikh:
Ok, I will polish it a little bit and merge

PR #1812, opened 2018-08-17
Fix bugs causing early task evaluation

SUPERCILEX:
See this chart: https://docs.gradle.org/4.9/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview

This PR removes calls to APIs that force all tasks to be evaluated.

Starts work on [KT-26205](https://youtrack.jetbrains.com/issue/KT-26205).


SUPERCILEX:
Of note, any APIs that somehow uses the `TaskCollection` iterable forces everything to be evaluated.

h0tk3y:
@SUPERCILEX Since the Kotlin Gradle plugins targets a range of Gradle versions starting from as early as 3.5, the calls to the new API need compatibility wrappers that fall back to the old behavior with older Gradle versions. Also, the new API is not going to be available in the plugin sources at compile time until we migrate our build to at least Gradle 4.9. Until then, the new API should be called with reflection.

SUPERCILEX:
@abelkov @h0tk3y Merged, PTAL

h0tk3y:
Thanks! Sorry for taking so long.

SUPERCILEX:
No problem, glad to see it go through!

pun-ky:
waiting for release. looks good 👍 

for me it is a showstopper. without it I cannot apply task configuration avoidance in my builds having Kotlin code being compiled.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
```diff
@@ -367,25 +367,21 @@ internal class GradleCompilerRunner(private val project: Project) : KotlinCompil
...
-
-                            if (task is Kotlin2JsCompile) {
-                                jarForSourceSet(project, task.sourceSetName)?.let {
-                                    jarToModule[it] = module
-                                }
-                            }
-                        }
-                        is InspectClassesForMultiModuleIC -> {
-                            jarToClassListFile[File(task.archivePath)] = task.classesListFile
+                project.tasks.withType(AbstractKotlinCompile::class.java).forEach { task ->
```
JLLeitschuh: If you want to use the new lazy configuration API, this needs to be `configureEach`.
Also, `forEach` won't configure new `AbstractKotlinCompile` tasks that are added later in the build. Is this intentional?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
```diff
@@ -367,25 +367,21 @@ internal class GradleCompilerRunner(private val project: Project) : KotlinCompil
...
+                    task.javaOutputDir?.let { dirToModule[it] = module }
+                    nameToModules.getOrPut(module.name) { HashSet() }.add(module)
+
+                    if (task is Kotlin2JsCompile) {
+                        jarForSourceSet(project, task.sourceSetName)?.let {
+                            jarToModule[it] = module
                         }
                     }
                 }
+                project.tasks.withType(InspectClassesForMultiModuleIC::class.java).forEach { task ->
```
JLLeitschuh: Same problem here as above.

PR #1813, opened 2018-08-17
KT-25933 ReplaceCallWithBinaryOperator should not suggest to replace 'equals' involving floating-point types

tommykw:
https://youtrack.jetbrains.com/issue/KT-25933

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
@@ -22,13 +22,13 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.tree.IElementType
 import org.jetbrains.kotlin.idea.inspections.AbstractApplicabilityBasedInspection
-import org.jetbrains.kotlin.idea.intentions.callExpression
+import org.jetbrains.kotlin.idea.intentions.*
 import org.jetbrains.kotlin.idea.intentions.conventionNameCalls.isAnyEquals
-import org.jetbrains.kotlin.idea.intentions.isReceiverExpressionWithValue
-import org.jetbrains.kotlin.idea.intentions.toResolvedCall
 import org.jetbrains.kotlin.lexer.KtSingleValueToken
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getChildOfType
+import org.jetbrains.kotlin.psi.psiUtil.getChildrenOfType
 import org.jetbrains.kotlin.psi.psiUtil.getLastParentOfTypeInRow
 import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
 import org.jetbrains.kotlin.resolve.calls.model.ArgumentMatch
@@ -123,6 +123,7 @@ class ReplaceCallWithBinaryOperatorInspection : AbstractApplicabilityBasedInspec
         return when (identifier) {
             OperatorNameConventions.EQUALS -> {
                 if (!dotQualified.isAnyEquals()) return null
+                if (dotQualified.isFloatingPointNumberEquals()) return null
 
                 val prefixExpression = dotQualified.getWrappingPrefixExpressionIfAny()
                 if (prefixExpression != null && prefixExpression.operationToken == KtTokens.EXCL) KtTokens.EXCLEQ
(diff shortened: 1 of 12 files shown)


idea/testData/inspectionsLocal/conventionNameCalls/replaceCallWithBinaryOperator/equalsFloat.kt
```diff
@@ -0,0 +1,5 @@
+// PROBLEM: none
+
+fun test(a: Any, b: Any) =
```
mglukhikh: Please add also test with explicitly given `Float` or `Double` types, like this: `fun test(a: Double, b: Double) = ...`, and another test with `Double` or `Float` literals, like `fun checkPi(x: Double) = x.equals(3.14) // or 3.14.equals(x)`

idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
```diff
@@ -148,4 +149,19 @@ class ReplaceCallWithBinaryOperatorInspection : AbstractApplicabilityBasedInspec
             else -> OperatorConventions.BINARY_OPERATION_NAMES.inverse()[identifier]
         }
     }
+
+    private fun KtDotQualifiedExpression.isFloatingPointNumberEquals(): Boolean {
```
mglukhikh: This implementation is very strange. You focus here on a given example, but there are some simpler cases which probably will not be detected by this implementation correctly. I'd suggest you should find (via `resolveToCall()`) type of `equals` receiver and type of `equals` argument, and in case both are floating-point types, the function should return `true`, otherwise `false`.

idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
```diff
@@ -123,6 +123,7 @@ class ReplaceCallWithBinaryOperatorInspection : AbstractApplicabilityBasedInspec
         return when (identifier) {
             OperatorNameConventions.EQUALS -> {
                 if (!dotQualified.isAnyEquals()) return null
+                if (dotQualified.isFloatingPointNumberEquals()) return null
```
mglukhikh: Here we should not just drop a suggestion. It should be kept, but its highlight type should be changed to `INFORMATION`, and its fix text changed to `Replace total order equality with IEEE 754 equality`.

mglukhikh: (See related functions above)

mglukhikh:
@tommykw Are you going to fix it yourself? This issue has some importance, so please do it or leave it to me, as you wish.

tommykw:
@mglukhikh Sorry for the late reply. I'm going to fix it.

idea/testData/inspectionsLocal/conventionNameCalls/replaceCallWithBinaryOperator/equalsFloat.kt
tommykw: I added tests.

idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
tommykw: I fixed it.

idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
tommykw: Thanks for your comments and suggestions. It's fixed now. How can I handle the following cases?

```kotlin
fun test(a: Any, b: Any) =
        a is Double && b is Double &&
                a.equals(b) // Here a and b are statically known to be Double.
```

mglukhikh:
Thanks for fixes, I will now consider everything more attentively.

idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceCallWithBinaryOperatorInspection.kt
mglukhikh: Looks like it's a hard question and a hard task. I will implement this piece myself

mglukhikh:
Merged manually, the part related to smart-casts is implemented aside. Thank you.

PR #1817, opened 2018-08-20
[JS] Improve isInheritanceFromInterface performance

soywiz:
isInheritanceFromInterface is pretty slow, this patch lazily precomputes all the possible interfaces of a type, so it can be checked with a single `indexOf`, that some browsers optimize by converting the array into a set internally.

### DIFF
--- a/libraries/stdlib/js/src/js/rtti.js
+++ b/libraries/stdlib/js/src/js/rtti.js
@@ -39,22 +39,35 @@ Kotlin.callSetter = function (thisObject, klass, propertyName, value) {
     Kotlin.callSetter(thisObject, Object.getPrototypeOf(klass), propertyName, value);
 };
 
-function isInheritanceFromInterface(ctor, iface) {
-    if (ctor === iface) return true;
-
+function getAllInterfaces(ctor) {
+    if (ctor.$metadata$ == null) ctor.$metadata$ = {};
     var metadata = ctor.$metadata$;
-    if (metadata != null) {
+    if (metadata._allInterfaces === undefined) {
+        var allInterfaces = metadata._allInterfaces = [];
+
+        allInterfaces.push(ctor);
+
         var interfaces = metadata.interfaces;
-        for (var i = 0; i < interfaces.length; i++) {
-            if (isInheritanceFromInterface(interfaces[i], iface)) {
-                return true;
+        if (interfaces !== undefined) {
+            for (var i = 0; i < interfaces.length; i++) {
+                allInterfaces.push.apply(allInterfaces, getAllInterfaces(interfaces[i]));
             }
         }
-    }
 
-    var superPrototype = ctor.prototype != null ? Object.getPrototypeOf(ctor.prototype) : null;
-    var superConstructor = superPrototype != null ? superPrototype.constructor : null;
-    return superConstructor != null && isInheritanceFromInterface(superConstructor, iface);
+        var superPrototype = ctor.prototype != null ? Object.getPrototypeOf(ctor.prototype) : null;
+        var superConstructor = superPrototype != null ? superPrototype.constructor : null;
+        if (superConstructor != null) {
+            allInterfaces.push.apply(allInterfaces, getAllInterfaces(superConstructor));
+        }
+        // @TODO: faster duplicate removal
+        metadata._allInterfaces = metadata._allInterfaces.filter(function(item, pos, self) { return self.indexOf(item) === pos; });
+    }
+    return metadata._allInterfaces;
+}
+
+function isInheritanceFromInterface(ctor, iface) {
+    if (ctor === iface) return true;
+    return getAllInterfaces(ctor).indexOf(iface) >= 0;
 }
 
 /**


anton-bannykh:
@soywiz Could you share the use case where the instance check performance was significant (just curious)? Also did you stumble upon any other performance issues?

soywiz:
Sure. I did this optimization after profiling using chrome a proof of concept of a 2d game that has a render tree of nodes. Not remember exactly where I had instance checking or pattern matching, but happens per frame. Also not sure if I was later able to remove those checks with inheritance. But in fact instance checking is costly on JS where on JVM is pretty cheap, so this patch amends it a bit by flattening the structure and memoizing the results but might require a bit more of refining.

I was dirty applying that patch here when combining and serving the client: https://github.com/mmo-poc/mmo-poc/blob/eedbb437c96d84cb5358b49113e963da2a064715/mmo/jvm/src/mmo/server/Backend.kt#L309-L346

Without the patch `isInheritanceFromInterface` takes 4.4% of the time:
screenshot 2018-11-26 at 04 25 28

With the patch `isInheritanceFromInterface` cost doesn't appear at all (`Kotlin.isType` still appears):
screenshot 2018-11-26 at 04 26 22

I guess this optimization would help in other use cases too.

anton-bannykh:
@soywiz Thanks for the info!

libraries/stdlib/js/src/js/rtti.js
```diff
@@ -39,22 +39,35 @@ Kotlin.callSetter = function (thisObject, klass, propertyName, value) {
...
 };
 
-function isInheritanceFromInterface(ctor, iface) {
-    if (ctor === iface) return true;
-
+function getAllInterfaces(ctor) {
+    if (ctor.$metadata$ == null) ctor.$metadata$ = {};
     var metadata = ctor.$metadata$;
-    if (metadata != null) {
+    if (metadata._allInterfaces === undefined) {
```
bashor: would be better to use `void 0` instead of `undefined`

libraries/stdlib/js/src/js/rtti.js
```diff
@@ -39,22 +39,35 @@ Kotlin.callSetter = function (thisObject, klass, propertyName, value) {
     Kotlin.callSetter(thisObject, Object.getPrototypeOf(klass), propertyName, value);
 };
 
-function isInheritanceFromInterface(ctor, iface) {
-    if (ctor === iface) return true;
-
+function getAllInterfaces(ctor) {
+    if (ctor.$metadata$ == null) ctor.$metadata$ = {};
```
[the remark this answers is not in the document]
bashor: would be better to avoid changing `ctor`, and I think we can just return any of `[]`, `null`, `void 0`

libraries/stdlib/js/src/js/rtti.js
```diff
@@ -39,22 +39,35 @@ Kotlin.callSetter = function (thisObject, klass, propertyName, value) {
...
+                allInterfaces.push.apply(allInterfaces, getAllInterfaces(interfaces[i]));
             }
         }
+
+        var superPrototype = ctor.prototype != null ? Object.getPrototypeOf(ctor.prototype) : null;
+        var superConstructor = superPrototype != null ? superPrototype.constructor : null;
+        if (superConstructor != null) {
+            allInterfaces.push.apply(allInterfaces, getAllInterfaces(superConstructor));
+        }
+        // @TODO: faster duplicate removal
```
bashor: TODO: use Set when it's possible?

libraries/stdlib/js/src/js/rtti.js
```diff
@@ -39,22 +39,35 @@ Kotlin.callSetter = function (thisObject, klass, propertyName, value) {
...
-    if (ctor === iface) return true;
-
+function getAllInterfaces(ctor) {
+    if (ctor.$metadata$ == null) ctor.$metadata$ = {};
     var metadata = ctor.$metadata$;
-    if (metadata != null) {
+    if (metadata._allInterfaces === undefined) {
+        var allInterfaces = metadata._allInterfaces = [];
+
+        allInterfaces.push(ctor);
```
bashor: actually, it's not all interfaces, but all parents + class itself

bashor: I think we can push `superConstructor` at 60 line instead of pushing `ctor` here.

libraries/stdlib/js/src/js/rtti.js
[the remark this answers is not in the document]
soywiz: I don't remember the details. But maybe to not fail on non-kotlin types? But maybe that's handled in the functions calling this one so if that's the case, that line could be removed.

---

To prevent changing the shape of the objects, maybe in the type declaration at the code generation phase, `_allInterfaces` could be set to null so the property at least exists even if computed lazily, or even to an empty array having a boolean elsewhere to determine if it was computed if changing a null still changes the shape.

In a full AOT would be possible to assign numbers to each type beforehand and make a much faster lookup, but having separate modules and being able to depend on super types defined in other module, the lazy option is the best option I thought at that time.

libraries/stdlib/js/src/js/rtti.js
soywiz: Makes sense.

soywiz:
This should be pretty outdated already, going to close it.

PR #1818, opened 2018-08-20
Creating IrLineNumberTestGenerated, adding line numbers for few expressions

denisvnukov:
This PR enables LineNumberTestGenerated test on IR backend. The testing of
hardcoded sequence of line numbers is replaced with mere checks for set-like
checks for expected line numbers.


compiler/testData/lineNumber/custom/functionCallWithDefault.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: Don't add new line!  Use empty  one before '// 2 3 ....'

compiler/testData/lineNumber/custom/inTheEndOfLambdaArgumentOfInlineCall.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: Same, don't change number count

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
```diff
@@ -58,7 +58,7 @@ private KotlinCoreEnvironment createEnvironment() {
     public void setUp() throws Exception {
         super.setUp();
 
-        KotlinCoreEnvironment environment = createEnvironment();
+        KotlinCoreEnvironment environment = createEnvironment(true);
```
max-kammerer: Why not compile this with ir backend?

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
```diff
@@ -166,24 +170,29 @@ protected void doTestCustom(String path) {
...
                         if (LINE_NUMBER_FUN.equals(name)) {
                             assert lastLabel != null : "A function call with no preceding label";
                             labels.add(lastLabel);
                         }
-                        lastLabel = null;
                     }
 
                     @Override
                     public void visitLabel(@NotNull Label label) {
+                        if (!labels2LineNumbers.containsKey(lastLabel) && lastLine >= 0) {
```
max-kammerer: What is purpose for this tricky code?

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
```diff
@@ -44,11 +44,11 @@
     private static final Pattern TEST_LINE_NUMBER_PATTERN = Pattern.compile("^.*test." + LINE_NUMBER_FUN + "\\(\\).*$");
 
     @NotNull
-    private KotlinCoreEnvironment createEnvironment() {
+    protected KotlinCoreEnvironment createEnvironment(boolean setup) {
         return KotlinCoreEnvironment.createForTests(
                 myTestRootDisposable,
                 KotlinTestUtils.newConfiguration(
-                        ConfigurationKind.JDK_ONLY, TestJdkKind.MOCK_JDK, KotlinTestUtils.getAnnotationsJar(), tmpdir
+                        ConfigurationKind.ALL, TestJdkKind.MOCK_JDK, KotlinTestUtils.getAnnotationsJar(), tmpdir
```
max-kammerer: Use updateConfiguration/extractConfigurationKind to configure ir stuff (rebase is required)

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractIrLineNumberTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.test.testFramework.KtUsefulTestCase
+import org.jetbrains.org.objectweb.asm.ClassReader
+
+abstract class AbstractIrLineNumberTest : AbstractLineNumberTest() {
+    override fun createEnvironment(setup: Boolean): KotlinCoreEnvironment {
+        val environment = super.createEnvironment(setup)
+        if (!setup) {
+            // Compile setup class always with original non-IR backend
+            environment.configuration.put(JVMConfigurationKeys.IR, true)
```
max-kammerer: Use updateConfiguration/extractConfigurationKind to configure ir stuff (rebase is required)

compiler/tests/org/jetbrains/kotlin/generators/tests/GenerateCompilerTests.kt
```diff
@@ -360,6 +357,11 @@ fun main(args: Array<String>) {
             model("checkLocalVariablesTable", targetBackend = TargetBackend.JVM_IR)
         }
 
+        testClass<AbstractIrLineNumberTest> {
+            model("lineNumber", recursive = false, targetBackend = TargetBackend.JVM_IR)
```
max-kammerer: A bit obsolete: see updated origin

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
denisvnukov: The errors I am getting during compilation look like:

```
UNRESOLVED_REFERENCE: Unresolved reference: lineNumber (2,21) in /class.kt
java.lang.IllegalStateException: UNRESOLVED_REFERENCE: Unresolved reference: lineNumber (2,21) in /class.kt
```

I read it as: currently ir backend does not generate kotlin.Metadata annotation and the generated class is not properly loaded.

compiler/testData/lineNumber/custom/inTheEndOfLambdaArgumentOfInlineCall.kt
denisvnukov: Done

compiler/testData/lineNumber/custom/functionCallWithDefault.kt
denisvnukov: Right, I thought the common approach is to use automatic disabling tests. 

If the test gets enabled automatically, It's going to remove this line anyways, right?

Done.

max-kammerer: In some cases manual changes are required, especially in test for linenumbers and smap. 
If not removed automatically it will fail again.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
max-kammerer: I should be redundant after rebase, it's only one compilation

compiler/testData/lineNumber/custom/functionCallWithDefault.kt
denisvnukov: Noted

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.java
denisvnukov: Right, it is gone after rebase

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.kt
```diff
@@ -57,102 +53,114 @@ abstract class AbstractLineNumberTest : CodegenTestCase() {
...
+                            labels.add(lastLabel ?: error("A function call with no preceding label"))
+                        }
+                    }
 
-        private fun extractSelectedLineNumbersFromSource(file: KtFile): List<String> {
-            val fileContent = file.text
-            val lineNumbers = arrayListOf<String>()
-            val lines = StringUtil.convertLineSeparators(fileContent).split("\n".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()
+                    override fun visitLabel(label: Label) {
+                        if (lastLabel != null && !labels2LineNumbers.containsKey(lastLabel) && lastLine >= 0) {
```
max-kammerer: Is it ir specific? Than please move condition (maybe: fun shouldKeepInheritedLines(...)) to ir part

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -141,6 +139,28 @@ class ExpressionCodegen(
         mv.visitEnd()
     }
 
+    private fun markFunctionLineNumber() {
+        if (irFunction.origin == JvmLoweredDeclarationOrigin.CLASS_STATIC_INITIALIZER) {
+            return
+        }
+        if (irFunction is IrConstructor && irFunction.isPrimary) {
```
max-kammerer: This method called after function body generation (irFunction.body!!.accept(this, info)). Should we use  markLineNumber(startOffset = FALSE)?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -175,7 +195,9 @@ class ExpressionCodegen(
 
     override fun visitBlockBody(body: IrBlockBody, data: BlockInfo): StackValue {
         return body.statements.fold(none()) { _, exp ->
-            exp.accept(this, data)
+            val result = exp.accept(this, data)
+            (exp as? IrExpression)?.markEndOfStatementIfNeeded()
```
max-kammerer:
```
exp.accept(this, data).also {
       (exp as? IrExpression)?.markEndOfStatementIfNeeded()
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -919,6 +975,9 @@ class ExpressionCodegen(
         }
 
         if (tryCatchBlockEnd != null) {
+            if (tryInfo != null) {
+                tryInfo.tryBlock.finallyExpression!!.markLineNumber(startOffset = false)
```
max-kammerer: This line is not clear, do we really need it?

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractLineNumberTest.kt
denisvnukov: Moved

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
denisvnukov: I tried to match the current behavior, but was not able to do so precisely since some code is not exactly relevant to IR (see for example markLineNumberForConstructor). As a result I decided to make sure we match what is reflected in tests and them fix the rest of the cases when we find them.

In line #146, for example, primary constructor end offset location location is taken from the class at least in some cases, which leads to weird breakpoints.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
denisvnukov: Sure, done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
denisvnukov: Added this to match genFinallyBlockOrGoto(), (see old ExpressionCodegen.java#1537.). I am not sure this is actually hit by any tests. Feel free to remove.

max-kammerer:
Thanks! Merged manually

PR #1819, opened 2018-08-21
KT-26236 QuickFix for ASSIGN_OPERATOR_AMBIGUITY on mutable collection '+=', '-='

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26236


idea/src/org/jetbrains/kotlin/idea/quickfix/AssignOperatorAmbiguityFactory.kt
```diff
@@ -0,0 +1,91 @@
...
+        val property = element ?: return
+        property.valOrVarKeyword.replace(KtPsiFactory(property).createValKeyword())
+    }
+}
+
+private class ChangeToAssignFunctionCallFix(
+    element: KtBinaryExpression,
+    private val operationText: String
+) : KotlinQuickFixAction<KtBinaryExpression>(element) {
+    override fun getText() = "Change to '${operationText}Assign()' call"
```
NataliaUkhorskaya: I'd suggest to change the name to 'replace with ...' as in ReplaceCallFix-quickfixes

idea/src/org/jetbrains/kotlin/idea/quickfix/AssignOperatorAmbiguityFactory.kt
```diff
@@ -0,0 +1,91 @@
...
+                    if (property != null && propertyName != null && property.isLocal) fixes.add(ChangeToValFix(property, propertyName))
+                    fixes.add(ChangeToAssignFunctionCallFix(element, operationText))
+                }
+            }
+        }
+        return fixes
+    }
+}
+
+private fun KotlinType?.isMutableCollection(): Boolean {
```
NataliaUkhorskaya: There is `JavaToKotlinClassMap.isMutable(type: KotlinType)`, so you can check the type it self and it's supertypes: `type.constructor.supertypes.any {..}`

t-kameyama: I don't understand how to use `JavaToKotlinClassMap.isMutable(type: KotlinType)`. `JavaToKotlinClassMap.isMutable(type: KotlinType)` returns `false` when type is ArrayList, HashMap, HashSet.

NataliaUkhorskaya: You should check the type it self (for MutableList and etc)  and also its supertypes (for ArrayList and etc)

t-kameyama:
There is b0c3461 in master, so I closed this PR.

NataliaUkhorskaya:
@t-kameyama Yes, I've merged it manually, thank you for contribution!

PR #1821, opened 2018-08-22
KT-26268 Inspection "Nested lambda has shadowed implicit parameter" should only warn if parameter is used

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26268

### DIFF
--- a/idea/src/META-INF/plugin.xml
+++ b/idea/src/META-INF/plugin.xml
@@ -2944,7 +2944,7 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      groupPath="Kotlin"
                      groupName="Style issues"
                      enabledByDefault="true"
-                     level="WARNING"
+                     level="WEAK WARNING"
                      language="kotlin"
     />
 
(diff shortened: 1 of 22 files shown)


4u7:
Thank you!

- Looks like now it's better to show warning on the shadowing `it` reference. 
- It's a bit odd that quickfix tries to rename only `it` corresponding to enclosing lambda. May be it is 
better to rename shadowing it, or even have both quickfixes, what do you think?
- It would be great to reduce severity of the inspection to weak warning.

t-kameyama:
> Looks like now it's better to show warning on the shadowing it reference.

OK, I'll fix it.

>It's a bit odd that quickfix tries to rename only it corresponding to enclosing lambda. May be it is
better to rename shadowing it, or even have both quickfixes, what do you think?

I think it's better to have both fixes.

>It would be great to reduce severity of the inspection to weak warning.

OK.

4u7:
Thank you, just a couple of minor notes:
- If there is more than one `it` we should highlight all of them, not just the first one
- It is better to reuse `ReplaceItWithExplicitFunctionLiteralParamIntention` with `IntentionWrapper`, it will prevent the second intention which is already there from showing. You may just pass it to `registerProblem` call like `IntentionWrapper(ReplaceItWithExplicitFunctionLiteralParamIntention(), it.containingFile)`

And a side note, you may test your inspections on Kotlin project itself with `Analyze -> Inspect Code`, there are lots of interesting cases to check if everything looks good.

PR #1824, opened 2018-08-22
KT-25570 - Add random extension functions to generate unsigned randoms

Kerooker:
Solves https://youtrack.jetbrains.com/issue/KT-25570

### DIFF
--- a/libraries/stdlib/src/kotlin/random/Random.kt
+++ b/libraries/stdlib/src/kotlin/random/Random.kt
@@ -438,7 +438,7 @@ public fun Random.nextUBytes(array: UByteArray) = nextUBytes(array, 0, array.siz
  * Creates a byte array of the specified [size], filled with random bytes.
  */
 @SinceKotlin("1.3")
-public fun Random.nextUBytes(size: Int): UByteArray = nextUBytes(UByteArray(size))
+public fun Random.nextUBytes(size: Int): UByteArray = UByteArray(nextBytes(size))
 
 /**
  * Fills a subrange of the specified UByte [array] starting from [fromIndex] inclusive and ending [toIndex] exclusive with random UBytes
--- a/libraries/stdlib/src/kotlin/random/Random.kt
+++ b/libraries/stdlib/src/kotlin/random/Random.kt
@@ -404,7 +404,7 @@ public fun Random.nextULong(origin: ULong, bound: ULong): ULong {
     checkULongRangeBounds(origin, bound)
 
     val originTransformedToLong = origin.toLong() xor Long.MIN_VALUE
-    val boundTransformedToLong = origin.toLong() xor Long.MIN_VALUE
+    val boundTransformedToLong = bound.toLong() xor Long.MIN_VALUE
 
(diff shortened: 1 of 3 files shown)


libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -330,7 +459,9 @@ internal fun Int.takeUpperBits(bitCount: Int): Int =
     this.ushr(32 - bitCount) and (-bitCount).shr(31)
 
 internal fun checkRangeBounds(origin: Int, bound: Int) = require(bound > origin) { boundsErrorMessage(origin, bound) }
+internal fun checkUIntRangeBounds(origin: UInt, bound: UInt) = require(bound > origin) { boundsErrorMessage(origin, bound) }
 internal fun checkRangeBounds(origin: Long, bound: Long) = require(bound > origin) { boundsErrorMessage(origin, bound) }
+internal fun checkULongRangeBounds(origin: ULong, bound: ULong) = require(bound > origin) { boundsErrorMessage(origin, bound) }
```
hho: Since the other variants are all called `checkRangeBounds`, I would prefer that name for the unsigned variants as well.

Or, because there are so many of them now, just replace all of them with the single

    internal fun <T : Comparable<T>> checkRangeBounds(origin: T, bound: T) = require(bound > origin) { "Random range is empty: [$origin, $bound)." }

Kerooker: I wanted to do that, but it's not possible.

In the JVM, both checkRangeBounds(long, long) and checkRangeBounds(ULong, ULong) have the same signature. They'll both have the same name in JVM, and the compiler didn't allow me to do that

Kerooker: I liked your approach more, I'll see if it works

ilya-g: We're going to mangle top-level function names with unsigned types in their signatures to avoid this type of clashes. It would be possible to rename function after that mangling is turned on.

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -321,6 +320,136 @@ public fun Random(seed: Int): Random = XorWowRandom(seed, seed.shr(31))
...
+ * @return [array] filled with random bytes.
+ */
+@SinceKotlin("1.3")
+public fun Random.nextUBytes(array: UByteArray) = nextUBytes(array, 0, array.size)
+
+/**
+ * Creates a byte array of the specified [size], filled with random bytes.
+ */
+@SinceKotlin("1.3")
+public fun Random.nextUBytes(size: Int): UByteArray = nextUBytes(UByteArray(size))
```
ilya-g: Better to delegate each `nextUBytes` to the corresponding `nextBytes` in order to benefit from the better implementations of more specialized overloads.

Kerooker: Done.

ilya-g:
Thanks I've reviewed this PR and merged it manually into 1.3-M2 branch.

PR #1825, opened 2018-08-24
KT-26230 Inspection: replace safe cast (as?) with `if` (instance check + early return)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26230

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/SafeCastWithReturnInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/SafeCastWithReturnInspection.kt
@@ -13,6 +13,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElementVisitor
 import org.jetbrains.kotlin.diagnostics.Errors
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.BindingContext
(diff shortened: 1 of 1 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/SafeCastWithReturnInspection.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+
+class SafeCastWithReturnInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor =
+        binaryWithTypeRHSExpressionVisitor(fun(expression) {
+            if (expression.right == null) return
+            if (expression.operationReference.getReferencedName() != "as?") return
```
mglukhikh: There is `KtTokens.AS_SAFE`, please use it instead of raw text

idea/src/org/jetbrains/kotlin/idea/inspections/SafeCastWithReturnInspection.kt
```diff
@@ -0,0 +1,64 @@
...
+
+            val parent = expression.getStrictParentOfType<KtBinaryExpression>() ?: return
+            if (KtPsiUtil.deparenthesize(parent.left) != expression) return
+            if (parent.operationReference.getReferencedName() != "?:") return
```
mglukhikh: Please use `KtTokens.ELVIS`

mglukhikh:
Merged manually, thank you.

PR #1827, opened 2018-08-24
Add String multiplication

hamzamuric:
Create a new string by concatenating given string n times.
 It behaves like multiplication in some other languages.
 Returns the built string.


ilya-g:
We believe that this is a case of operator overloading misuse. 
The standard library provides the `repeat(n)` extension for CharSequences, that should be used instead: http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/repeat.html.

PR #1830, opened 2018-08-27
KT-26015 Intention to move property to constructor adds @field: qualifier to annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26015

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/MovePropertyToConstructorIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/MovePropertyToConstructorIntention.kt
@@ -76,12 +76,9 @@ class MovePropertyToConstructorIntention :
 
         val commentSaver = CommentSaver(element)
 
+        val context = element.analyze(BodyResolveMode.PARTIAL)
         val propertyAnnotationsText = element.modifierList?.annotationEntries?.joinToString(separator = " ") {
-            if (it.isApplicableToConstructorParameter()) {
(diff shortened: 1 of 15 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/MovePropertyToConstructorIntention.kt
```diff
@@ -129,19 +126,22 @@ class MovePropertyToConstructorIntention :
         return parameterDescriptor.source.getPsi() as? KtParameter
     }
 
-    private fun KtAnnotationEntry.isApplicableToConstructorParameter(): Boolean {
-        val context = analyze(BodyResolveMode.PARTIAL)
-        val descriptor = context[BindingContext.ANNOTATION, this] ?: return false
+    private fun KtAnnotationEntry.getTextWithUseSite(context: BindingContext): String {
```
mglukhikh: The priority of annotation without use-site application is value parameter (the highest), property, field (the lowest) for constructor parameter property. For regular property it is property (the highest), field (the lowest). So, if some annotation is not applicable to value parameter (VALUE_PARAMETER target is not included), then we should not add use-site while moving to constructor because it will be redundant (priority queue will be the same). And only in case VALUE_PARAMETER target is included we should do all this checks you have done here.

mglukhikh:
Merged manually (with small refactoring), thank you.

Montesdodgers13:
Closed

Montesdodgers13:
Closed

PR #1833, opened 2018-08-29
KT-25454 Extract function: make default visibility private

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25454


asedunov:
Merged! Thank you

PR #1835, opened 2018-08-29
KT-24978 Do not highlight foldable if-then for is checks

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24978

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/IfThenUtils.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/IfThenUtils.kt
@@ -299,7 +299,7 @@ internal fun KtIfExpression.buildSelectTransformationData(): IfThenToSelectData?
 internal fun KtExpression?.isClauseTransformableToLetOnly() =
     this is KtCallExpression && resolveToCall()?.getImplicitReceiverValue() == null
 
-internal fun KtIfExpression.shouldBeTransformed(): Boolean {
+internal fun KtIfExpression.shouldBeTransformed(transformToSafeAccess: Boolean = false): Boolean {
     val condition = condition
     return when (condition) {
         is KtBinaryExpression -> {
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/IfThenUtils.kt
```diff
@@ -299,19 +299,23 @@ internal fun KtIfExpression.buildSelectTransformationData(): IfThenToSelectData?
 internal fun KtExpression?.isClauseTransformableToLetOnly() =
     this is KtCallExpression && resolveToCall()?.getImplicitReceiverValue() == null
 
-internal fun KtIfExpression.shouldBeTransformed(): Boolean {
+internal fun KtIfExpression.shouldBeTransformed(transformToSafeAccess: Boolean = false): Boolean {
```
mglukhikh: I think this parameter should be always true (yes, even for `IfThenToElvisInspection`), so it's not required.

idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/IfThenUtils.kt
```diff
@@ -299,19 +299,23 @@ internal fun KtIfExpression.buildSelectTransformationData(): IfThenToSelectData?
...
         }
         is KtIsExpression -> {
-            val baseClause = (if (condition.isNegated) `else` else then)?.unwrapBlockOrParenthesis()
-            when {
-                baseClause.isClauseTransformableToLetOnly() -> false
-                !isMultiLine() -> true
-                else -> baseClause !is KtDotQualifiedExpression
+            if (transformToSafeAccess) {
+                false
+            } else {
```
mglukhikh: This piece of code will be dropped (see comment above).

mglukhikh:
Merged manually, thank you.

PR #1836, opened 2018-08-30
KT-20357: Add sample for mutableSet.

kitwtnb:

Add samples for `mutableSet()`.


ilya-g:
Thanks, I've merged the sample with minor edits.

PR #1839, opened 2018-08-31
KT-23691 Warn about `var` properties with default setter and getter that doesn't reference backing field

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23691


idea/src/META-INF/plugin.xml
```diff
@@ -3039,6 +3039,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.PropertyVarInspection"
+                     displayName="Property 'var' has no meaning"
```
mglukhikh: I think something like "Property setter does not influence its getter result"

idea/src/META-INF/plugin.xml
```diff
@@ -3039,6 +3039,15 @@ The Kotlin plugin provides language support in IntelliJ IDEA and Android Studio.
...
+                     groupPath="Kotlin"
+                     groupName="Style issues"
```
mglukhikh: "Probable bugs"

idea/src/org/jetbrains/kotlin/idea/inspections/PropertyVarInspection.kt
```diff
@@ -0,0 +1,37 @@
...
+class PropertyVarInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) =
+        propertyVisitor(fun(property: KtProperty) {
+            if (property.isLocal || !property.isVar || property.setter != null) return
+            val initializer = property.initializer ?: return
+            val getter = property.getter ?: return
+            val getterExpression = getter.initializer?.unwrapBlockOrParenthesis()
+                ?: (getter.bodyExpression?.unwrapBlockOrParenthesis() as? KtReturnExpression)?.returnedExpression
+                ?: return
+            if (initializer.text != getterExpression.text) return
```
mglukhikh: It's only one case when they are the same. The important thing here that getter should not reference backing field.

idea/src/org/jetbrains/kotlin/idea/inspections/PropertyVarInspection.kt
```diff
@@ -0,0 +1,37 @@
...
+            holder.registerProblem(
+                property.valOrVarKeyword,
+                "This 'var' has no meaning",
```
mglukhikh: Take from inspection description

idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeVariableMutabilityFix.kt
```diff
@@ -33,10 +33,12 @@ import org.jetbrains.kotlin.util.OperatorNameConventions
...
     element: KtValVarKeywordOwner,
     private val makeVar: Boolean,
-    private val actionText: String? = null
+    private val actionText: String? = null,
+    private val deleteInitializer: Boolean = false
 ) : KotlinQuickFixAction<KtValVarKeywordOwner>(element) {
 
-    override fun getText() = actionText ?: if (makeVar) "Change to var" else "Change to val"
+    override fun getText() = actionText
+        ?: (if (makeVar) "Change to var" else "Change to val")+(if (deleteInitializer) " and delete initializer" else "")
```
mglukhikh: `...) + (...`

idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeVariableMutabilityFix.kt
```diff
@@ -51,6 +53,7 @@ class ChangeVariableMutabilityFix(
         val factory = KtPsiFactory(project)
         val newKeyword = if (makeVar) factory.createVarKeyword() else factory.createValKeyword()
         element.valOrVarKeyword!!.replace(newKeyword)
+        if (deleteInitializer) (element as? KtProperty)?.initializer = null
```
mglukhikh: Add braces

idea/src/org/jetbrains/kotlin/idea/inspections/PropertyVarInspection.kt
mglukhikh: Please add tests for getter which references backing field (no warning) and does not reference it (warning).

idea/src/org/jetbrains/kotlin/idea/inspections/PropertyVarInspection.kt
```diff
@@ -0,0 +1,43 @@
...
+                    ChangeVariableMutabilityFix(property, makeVar = false, deleteInitializer = true),
+                    property.containingFile
+                )
+            )
+        })
+
+    private fun KtPropertyAccessor.hasBackingFieldReference(): Boolean {
+        val bodyExpression = this.bodyExpression ?: return false
+        val p = this.property
+        return bodyExpression.isBackingField(p) || bodyExpression.anyDescendantOfType<KtNameReferenceExpression> { it.isBackingField(p) }
```
mglukhikh: There is `BindingContext.BACKING_FIELD_REQUIRED`, please use it and not a raw text check

t-kameyama: `context[BindingContext.BACKING_FIELD_REQUIRED, propertyDescriptor]` returns `true` in this case:
```kotlin
class Test {
    var foo = 1
        get() = 1
}
```

t-kameyama: I want to report in following case:
```kotlin
    var foo = 1
        get() = 1
```

I don't want to report in following case:
```kotlin
    var foo: Int = 1
        get() = field
```

`BindingContext.BACKING_FIELD_REQUIRED` returns `true` in both cases, so I do not know how to check property that has backing field reference using `BindingContext.BACKING_FIELD_REQUIRED`.

mglukhikh: Yes, understood. So you should  both check BACKING_FIELD_REQUIRED and check we do not reference it in getter. Ok.

mglukhikh:
Thank you, merged

PR #1842, opened 2018-09-02
efficient last and lastOrNull for Iterables that are Lists

isopov:


ilya-g:
This way the `predicate` function is inlined twice which is undesirable.

isopov:
Really `lastIndexOf` that I used for reference is not inline function. I started playing with Kotlin 2 days ago (maybe PR-ing stdlib is not the right game) and don't know how to deal with that problem and if it serious enough to not make such a change.

isopov:
@ilya-g so how about making these functions noinline (for Iterables only)? Updated PR with it

Montesdodgers13:
Test email

ilya-g:
Making a function non-inline is a source-breaking change in case if that function takes a functional parameter, because it prohibits non-local control flows (`return` from an outer function) that was possible before in a lambda passed to that parameter:
```
fun doSearch(searchValue: String) {
    val result = items.last { 
         if (it == null) return  // returns from 'doSearch', only possible when 'last' is inline
         it.property == searchValue
    }
    ...
}
```

isopov:
Reverted to not inline versions that should not break things.

So should we close this PR or inlining predicate two time is acceptable price?

ilya-g:
Inlining lambda body several times may lead to an exponential growth of the generated bytecode in  cases like:
```
elements.last { it.properties.last { p -> ... some predicate ... } }
```
Here the code of the inner lambda becomes inlined 4 times. Or even twice more if this code is in a `finally` block.

Therefore I think we should not merge this PR.

PR #1844, opened 2018-09-04
KT-26540: Support generated superclasses in KAPT

mpetrov:
https://youtrack.jetbrains.net/issue/KT-26540

Included an integration test that relies on a generated superclass.


yanex:
Thanks for the contribution! I've merged the commit manually.
 Note that I've also added this commit to support non-existent super interface types: https://github.com/JetBrains/kotlin/commit/c517b85de1ec42c57fc30b6e9b4652b6e508c86c.

PR #1845, opened 2018-09-04
KT-22388 Join lines: variable declaration + if (decl == null) {return smth}

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22388


goodwinnk:
It was a smart inspection reuse. Thank you!

Merged as e409007749114edc2cec8c06e400acf827305cdc.

PR #1847, opened 2018-09-04
KT-26586: Sequence<T>.batch(n) operator implementation and documentation

bradynpoulsen:
https://youtrack.jetbrains.com/issue/KT-26586

The extension function was manually added to the generated _Sequences.kt file.
Does this need to be done elsewhere so it is properly generated?

### DIFF
--- a/libraries/stdlib/common/src/generated/_Sequences.kt
+++ b/libraries/stdlib/common/src/generated/_Sequences.kt
@@ -961,6 +961,20 @@ public fun <T, K> Sequence<T>.distinctBy(selector: (T) -> K): Sequence<T> {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ilya-g:
How does this relate to [KT-10900 (Lazy Sequence<T> concatenation)](https://youtrack.jetbrains.com/issue/KT-10900)?

bradynpoulsen:
@ilya-g good catch! I didn't realize I had copied another issue number before starting my work. I updated the commit, let me know if you'd like me to re-open this with a corrected branch name as well

bradynpoulsen:
Provided by [`Sequence<T>.chunked(Int)`](http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/chunked.html)

PR #1849, opened 2018-09-06
Simplify detailed receiver

cuihtlauac:
Use let in order to define detailedReceiver in a single statement


dsavvinov:
Again, can you please elaborate, why do you think such a change is necessary? Does it contradict some explicit guidelines? If so, it'd be nice to mention them somewhere in the commit message

dsavvinov:
Sorry, I don't think that this cosmetic change improves readability, as it essentially repeats the old code, but also adds one extra call and unnamed parameter `it`. 

Thanks for the PR, nevertheless!

PR #1850, opened 2018-09-06
Remove useless parameter

cuihtlauac:
LanguageVersionSettings is not used in convertToOverloadResults. Let's remove it.


cuihtlauac:
You're absolutely right, updated commit message as requested.

dsavvinov:
Thanks for your contribution!

PR #1851, opened 2018-09-06
Turn ImplicitScopeTowerImpl into a function

cuihtlauac:
Implementation of the ImplicitScopeTower interface can be replaced
by method returning an anonymous instance-implementation.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/tower/NewResolutionOldInference.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/tower/NewResolutionOldInference.kt
@@ -170,8 +170,7 @@ class NewResolutionOldInference(
             context.transformToReceiverWithSmartCastInfo(explicitReceiver as ReceiverValue)
         }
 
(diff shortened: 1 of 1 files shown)


cuihtlauac:
I initially wanted to avoid passing useless parameters at the instantiation call site: dynamicScope, syntheticScope and location do not depend from runResolution(). This makes it explicit that scopeTower functionally depends on each context of resolution.

Once done, turning the implementation class into a nameless one seemed a way to stress ImplicitScopeTower is the important thing and actually just a stateless bundle.

Arguably, this is cosmetic change.

dsavvinov:
Well, I get where you're coming from, but I don't think that it's the undeniably best way to convey that meaning. Which is an issue -- Kotlin is a big public project, and we'd like to change things only if they are completely non-controversial (otherwise it may be impossible to maintain stable code base, as there are a lot of people with different mindsets/tastes/coding habits)

So, unfortunately, I have to reject your pull request -- hope that's understandable. Thanks for your willingness to contribute, though!

cuihtlauac:
No problem, this is actually very understandable, and helps a lot understanding your approach towards your code base. I will refrain from making such kind of commits in the future. Thanks for taking the time to review my code and provide this kind explanation.

PR #1853, opened 2018-09-07
Minor improvements for the native part of kotlin-multiplatform

ilmat192:


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinTargetConfigurator.kt
```diff
@@ -411,6 +411,10 @@ open class KotlinNativeTargetConfigurator(
             // TODO: Provide a normal test path!
             outputDir = project.layout.buildDirectory.dir("test-results").get().asFile
 
+            if (project.hasProperty("teamcity")) {
```
h0tk3y: Check the value that TC sets, possibly handle `-Pteamcity=false`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/KotlinNativeTasks.kt
```diff
@@ -42,6 +42,10 @@ open class KotlinNativeCompile : DefaultTask() {
     val libraries: FileCollection
         @InputFiles get() = compilation.compileDependencyFiles
 
+    val friendModule: FileCollection?
```
h0tk3y: Consider restricting access

PR #1854, opened 2018-09-08
KT-26653 Intention to replace if-else with `x?.let { ... } ?: ...`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26653


mglukhikh:
Merged manually with minor fixes, thank you

PR #1857, opened 2018-09-09
KT-8673: forEach analog which takes function with receiver for Iterable 

bloderxd:
This is related with: https://youtrack.jetbrains.com/issue/KT-8673

### DIFF
--- a/libraries/stdlib/test/collections/IterableTests.kt
+++ b/libraries/stdlib/test/collections/IterableTests.kt
@@ -488,6 +488,17 @@ abstract class IterableTests<T : Iterable<String>>(val createFrom: (Array<out St
         assertEquals(listOf("bar"), result as? List)
     }
 
+    @Test
+    fun withEach() {
+        val substringList = mutableListOf<String>()
+        data.withEach {
+          substringList.add(substring(0, 1))
+        }
+        assertEquals(2, substringList.size)
+        assertEquals("f", substringList[0])
+        assertEquals("b", substringList[1])
+    }
+
 }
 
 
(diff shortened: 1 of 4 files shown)


ilya-g:
Thank you for your pull request. We haven't yet decided on the naming of the proposed function, but should we settle on `withEach`, we might use this pull request as a base.

libraries/tools/kotlin-stdlib-gen/src/templates/Mapping.kt
```diff
@@ -441,4 +441,17 @@ object Mapping : TemplateGroupBase() {
         }
     }
 
+    val f_withEach = fn("withEach(operation: T.() -> Unit)") {
+      include(Iterables)
```
ilya-g: Consider providing it for the same receivers as `forEach`.
Also it makes sense to make it inline same as `forEach` (note that different overloads have different inline modifiers)

libraries/stdlib/test/collections/IterableTests.kt
```diff
@@ -488,6 +488,17 @@ abstract class IterableTests<T : Iterable<String>>(val createFrom: (Array<out St
         assertEquals(listOf("bar"), result as? List)
     }
 
+    @Test
+    fun withEach() {
+        val substringList = mutableListOf<String>()
+        data.withEach {
+          substringList.add(substring(0, 1))
+        }
+        assertEquals(2, substringList.size)
```
ilya-g: Could be just `assertEquals(listOf("f", "b"), substrinList)`

bloderxd:
@ilya-g Thank you about feed back, now I tried to implement like `forEach` was implemented.

PR #1858, opened 2018-09-10
KT-26662 Corner cases around 'this' inside "replace if with safe access"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26662


mglukhikh:
Thank you!

PR #1861, opened 2018-09-10
WIP: Split the Android extensions runtime by feature

JakeWharton:
The original artifact is retained to aggregate the feature artifacts in case a library contains a dependency on it. You can either use an explicit dependency on the aggregate artifact, a resolution strategy (with Gradle), or an exclude to handle this case.


yanex:
That's ok :) Please ping me when the branch becomes ready.

PR #1866, opened 2018-09-12
Remove CandidateWithBoundDispatchReceiverImpl

cuihtlauac:
Interface CandidateWithBoundDispatchReceiver had a unique
implementation, which was private.

Turned interface CandidateWithBoundDispatchReceiver into a
class. Removed CandidateWithBoundDispatchReceiverImpl. Removed
string "Impl" suffix at constructor call sites.


dsavvinov:
I don't have an objections, but maybe @zarechenskiy knows if there are some arguments for leaving this as is?

zarechenskiy:
@dsavvinov I don't have objections too

PR #1868, opened 2018-09-14
Expand the fix from commit 6c274cecff3 to cover AndroidX

rsauciuc:
Android Studio users who migrated to AndroidX are running into the
databinding deadlock again: https://issuetracker.google.com/111788726


rsauciuc:
PS. Alternatively, I could explicitly match androidx.databinding... in the class name. Let me know which one you prefer.

mglukhikh:
@rsauciuc Thank you. I will merge it into master (after remote run), and then into 1.2.71 and 1.3 branch too.

PR #1869, opened 2018-09-14
Provide a DSL for publication settings

ilmat192:
#KT-26771 fixed.


libraries/tools/kotlin-gradle-plugin-api/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinTarget.kt
```diff
@@ -39,5 +42,8 @@ interface KotlinTarget: Named, HasAttributes {
 
     fun createUsageContexts(): Set<UsageContext>
 
+    fun publication(action: Closure<Unit>)
```
h0tk3y: Maybe a better naming scheme would provide more concrete understanding that this publication comes from the `maven-publish` plugin, like be `mavenPublication { ... }` (probably not the best option, since there's also the older `maven` plugin).

PR #1870, opened 2018-09-16
KT-20357 Add sample for orEmpty method in collections/Maps.kt

mii-chan:


ilya-g:
Thanks for the sample!

PR #1873, opened 2018-09-17
Support OptionalExpectation in native

ilmat192:


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/KotlinNativeTasks.kt
```diff
@@ -221,11 +221,11 @@ open class KotlinNativeCompile : AbstractCompile() {
...
             }
 
-            addListArg("-linkerOpts", linkerOpts)
+            addListArg("-linker-options", linkerOpts)
 
             // Sources.
-            // TODO: Filter only kt files?
-            addAll(sources.files.map { it.absolutePath })
+            addAll(getSource().map { it.absolutePath })
+            commonSources.mapTo(this) { "-Xcommon-sources=${it.absolutePath}" }
```
h0tk3y: Can these be passed in a single `-Xcommon-sources=<path>` argument?

PR #1874, opened 2018-09-18
Bad quickfix name for "Call on non-null type may be reduced"

tommykw:
https://youtrack.jetbrains.com/issue/KT-26902


mglukhikh:
Thank you!

PR #1879, opened 2018-09-21
KT-21520 'Assignment can be lifted out of if' in cases where argument are of different types

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21520


mglukhikh:
Generally, it looks good, thank you.

PR #1884, opened 2018-09-21
Use new API from AGP to get JavaCompile task

gavra0:
New API for getting the JavaCompile task has been
added to AGP, and the old one is deprecated. It is
now returning TaskProvider<JavaCompile>, which allows
API users to avoid configuring tasks unnecessarily.

This commit does not change anything functionally, it
is only to avoid the warning message, as reporter in
https://issuetracker.google.com/116198439.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/agp25/kotlin/org/jetbrains/kotlin/gradle/plugin/android/Android25ProjectHandler.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/agp25/kotlin/org/jetbrains/kotlin/gradle/plugin/android/Android25ProjectHandler.kt
@@ -88,9 +88,15 @@ class Android25ProjectHandler(kotlinConfigurationTools: KotlinConfigurationTools
         else -> null
     }
 
-    override fun getJavaTask(variantData: BaseVariant): AbstractCompile? =
-            @Suppress("DEPRECATION") // There is always a Java compile task -- the deprecation was for Jack
-            variantData.javaCompile
+    override fun getJavaTask(variantData: BaseVariant): AbstractCompile? {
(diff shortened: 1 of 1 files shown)


libraries/tools/kotlin-gradle-plugin/src/agp25/kotlin/org/jetbrains/kotlin/gradle/plugin/android/Android25ProjectHandler.kt
```diff
@@ -88,9 +88,15 @@ class Android25ProjectHandler(kotlinConfigurationTools: KotlinConfigurationTools
...
     }
 
-    override fun getJavaTask(variantData: BaseVariant): AbstractCompile? =
-            @Suppress("DEPRECATION") // There is always a Java compile task -- the deprecation was for Jack
-            variantData.javaCompile
+    override fun getJavaTask(variantData: BaseVariant): AbstractCompile? {
+        @Suppress("DEPRECATION") // There is always a Java compile task -- the deprecation was for Jack
+        return variantData::class.java.methods.firstOrNull { it.name == "getJavaCompileProvider" }
+            ?.invoke(variantData)
+            .let { taskProvider ->
```
h0tk3y: The control flow based on nullability seems to be wrong here, as it says: regardless of whether the `<...>?.invoke(...)` result is null or not, run the `let`
s `{ taskProvider -> ... }` block, possibly with `null` as `taskProvider`. Should it be `?.let { taskProvider -> ... }`?

gavra0: Thanks for spotting this. It's fixed now.

PR #1885, opened 2018-09-23
KT-19333 Character literal used as a value of type 'int' in Java is missing explicit type conversion in Kotlin

Monchi:


Monchi:
https://youtrack.jetbrains.com/issue/KT-19333

TWiStErRob:
Originating issue closed as Obsolete.

goodwinnk:
@Monchi Thank you for this pull request. Sometime ago we started a full Java to Kotlin converter redesign and the original issue (KT-19333) was accidentally fixed during this major rewrite. As this pull request was meant to make a fix in the previous version of Java to Kotlin converter that is now unsupported, the only thing I can do now is to close this pull request without a merge. 

At the moment we started the major redesign of converter we didn't know the final list of issues that are going to be fixed at the end.

I know it's a sad feeling when something is done for nothing and I'm really sorry for this situation. Thank you again for your time and effort.

Monchi:
@goodwinnk 
hank you for explanation.

First, new J2K is really pleasing, it solves many problems. However, the issues for old J2K were marked as "Up for Grabs" and we didn't know it was going to be replaced. This is the problem i think. I understand that nobody knew which issues would be fixed at the moment rewriting was started. So when rewriting something at the next time, please announce that and mark issues as "On Hold" or something like that. This will solve those problems.

Once again, congrats for publishing new J2K!

goodwinnk:
> So when rewriting something at the next time, please announce that and mark issues as "On Hold" or something like that. This will solve those problems.

Of course we will try to avoid miscommunication of this scale in the future. We removed "Up for Grabs" from the issues in the subsystem as soon we understood it's going to affect the whole converter but several pull requests were already there at the moment. 

> Once again, congrats for publishing new J2K!
Thank you!

PR #1886, opened 2018-09-23
KT-17004 There is no suggestion to add var or val to supertype

Monchi:


idea/src/org/jetbrains/kotlin/idea/quickfix/AddPropertyToSupertypeFix.kt
```diff
@@ -0,0 +1,191 @@
...
+            return supertypes.mapNotNull { it.constructor.declarationDescriptor as? ClassDescriptor }
+        }
+
+        private fun MutableList<KotlinType>.sortSubtypesFirst(): List<KotlinType> {
+            val typeChecker = KotlinTypeChecker.DEFAULT
+            for (i in 1 until size) {
+                val currentType = this[i]
+                for (j in 0 until i) {
+                    if (typeChecker.isSubtypeOf(currentType, this[j])) {
+                        this.removeAt(i)
```
mglukhikh: Your manual sorting method has estimated performance of `O(N^3)` because of this `removeAt`, and this could slow everything down.

mglukhikh:
In future, please don't forget to run "Generate All Tests" and then run at least all new tests locally. At this moment two of your tests fail.

Monchi:
@mglukhikh I'm sorry i tested without "Generate All Tests". Should I fix tests and commit that?

mglukhikh:
Fixed and merged manually. Thank you.

PR #1894, opened 2018-09-25
[K/N] Stand-alone Native Application wizard in IDEA

ddolovov:
Issue #KT-27076 Fixed


idea/idea-gradle-native/src/org/jetbrains/kotlin/ide/konan/gradle/KotlinGradleNativeMultiplatformModuleBuilder.kt
```diff
@@ -0,0 +1,34 @@
...
+    override val notImportedCommonSourceSets = true
+
+    override fun buildMultiPlatformPart(): String {
+        return """
+            kotlin {
+                targets {
+                    // For ARM, preset should be changed to presets.iosArm32 or presets.iosArm64
+                    // For Linux, preset should be changed to e.g. presets.linuxX64
+                    // For MacOS, preset should be changed to e.g. presets.macosX64
+                    fromPreset(presets.iosX64, '$nativeTargetName')
```
mglukhikh: Suggestion from Lilia: (applicable also to Multiplatform Library sample): it would be nice to detect which Native target may be actually built on the given host and uncomment the corresponding target in the build.gradle file instead of leaving iOS by default. If you know how to implement such logic, please do it here, otherwise I will think about it myself (and fix later both this wizard and library wizard in case of success)

idea/src/META-INF/gradle-java.xml.as33
```diff
@@ -78,6 +78,7 @@
 
         <scriptDefinitionContributor implementation="org.jetbrains.kotlin.idea.core.script.GradleScriptDefinitionsContributor" order="first"/>
 
+        <moduleBuilder implementation="org.jetbrains.kotlin.ide.konan.gradle.KotlinGradleNativeMultiplatformModuleBuilder"/>
```
mglukhikh: You should (probably) change also `gradle.xml.181`, otherwise your wizard will not bee seen in IDEA 181 or earlier versions.

idea/idea-gradle-native/src/org/jetbrains/kotlin/ide/konan/gradle/KotlinGradleNativeMultiplatformModuleBuilder.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    private var nativeTargetName: String = "ios"
+
+    override fun getBuilderId() = "kotlin.gradle.multiplatform.native"
+
+    override fun getPresentableName() = "Kotlin/Native"
+
+    override fun getDescription() = "Kotlin module for iOS target"
+
+    override val notImportedCommonSourceSets = true
```
mglukhikh: I'm not sure how valid is the hack related to this property. Hope @asedunov will tell something about it.

idea/idea-gradle-native/src/org/jetbrains/kotlin/ide/konan/gradle/KotlinGradleNativeMultiplatformModuleBuilder.kt
ddolovov: I know how to do it. Will try to.

idea/src/META-INF/gradle-java.xml.as33
ddolovov: OK.

idea/idea-gradle-native/src/org/jetbrains/kotlin/ide/konan/gradle/KotlinGradleNativeMultiplatformModuleBuilder.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override fun buildMultiPlatformPart(): String {
+        return """
+            kotlin {
+                targets {
+                    // For ARM, preset should be changed to presets.iosArm32 or presets.iosArm64
+                    // For Linux, preset should be changed to e.g. presets.linuxX64
+                    // For MacOS, preset should be changed to e.g. presets.macosX64
```
mglukhikh: May be it's a good idea to add also a line about `iosX64`

PR #1897, opened 2018-09-26
KT-27138 Change visibility intentions are suggested on properties marked with @JvmField

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27138

### DIFF
--- a/idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
+++ b/idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
@@ -28,8 +28,10 @@ import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
 import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.idea.util.IdeDescriptorRenderers
+import org.jetbrains.kotlin.idea.util.findAnnotation
 import org.jetbrains.kotlin.lexer.KtModifierKeywordToken
 import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.addRemoveModifier.MODIFIERS_ORDER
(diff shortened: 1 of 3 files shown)


idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
```diff
@@ -269,7 +271,10 @@ fun KtDeclaration.implicitVisibility(): KtModifierKeywordToken? =
         }
     }
 
-fun KtModifierListOwner.canBePrivate() = modifierList?.hasModifier(KtTokens.ABSTRACT_KEYWORD) != true
+fun KtModifierListOwner.canBePrivate(): Boolean {
```
mglukhikh: What about `internal` visibility?

idea/idea-core/src/org/jetbrains/kotlin/idea/core/psiModificationUtils.kt
```diff
@@ -271,10 +269,7 @@ fun KtDeclaration.implicitVisibility(): KtModifierKeywordToken? =
         }
     }
 
-fun KtModifierListOwner.canBePrivate(): Boolean {
-    if (this is KtProperty && this.findAnnotation(FqName("kotlin.jvm.JvmField")) != null) return false
-    return modifierList?.hasModifier(KtTokens.ABSTRACT_KEYWORD) != true
-}
+fun KtModifierListOwner.canBePrivate() = modifierList?.hasModifier(KtTokens.ABSTRACT_KEYWORD) != true
```
mglukhikh: I'd say it's not the best way to fix it. It's more correct to have all necessary checks inside `canBePrivate` and `canBeInternal` (which does not yet exist). Take into account the fact that both `ChangeVisibilityFix` and `ChangeVisibilityModifierIntention` use `canBePrivate` (and should probably share the same logic). Also I'd recommend to consider content of `canBePrivate` inside `MemberVisibilityCanBePrivateInspection` (private fun). I suspect that some checks from there (not all) could be used inside global `canBePrivate` (that can also fix a bunch of minor bugs, not reported yet).

idea/src/org/jetbrains/kotlin/idea/intentions/ChangeVisibilityModifierIntention.kt
```diff
@@ -113,6 +112,7 @@ open class ChangeVisibilityModifierIntention protected constructor(
     class Private : ChangeVisibilityModifierIntention(KtTokens.PRIVATE_KEYWORD), HighPriorityAction {
         override fun applicabilityRange(element: KtDeclaration): TextRange? {
             if (isAnnotationClassPrimaryConstructor(element)) return null
+            if (element is KtProperty && element.hasJvmFieldAnnotation()) return null
```
mglukhikh: I'd say it's not the best way to fix it. It's more correct to have all necessary checks inside `canBePrivate` and `canBeInternal` (which does not yet exist). Take into account the fact that both `ChangeVisibilityFix` and `ChangeVisibilityModifierIntention` use `canBePrivate` (and should probably share the same logic). Also I'd recommend to consider content of `canBePrivate` inside `MemberVisibilityCanBePrivateInspection` (private fun). I suspect that some checks from there (not all) could be used inside global `canBePrivate` (that can also fix a bunch of minor bugs, not reported yet).

PR #1898, opened 2018-09-27
KT-27071 "Add non-null asserted (!!) call" places `!!` at wrong position with operator `get` (array indexing)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27071


NataliaUkhorskaya:
Hello @t-kameyama! First of all I want to apologize for such a long delay in my response. Thank you for your pull-request.
The initial problem of this issue is that the quickfix is created in two different situations: when there is a type mismatch error (then the quickfix should be applied for the whole array expression) and when there is an unsafe call (then the quickfix should be applied for the receiver of array get call).
So I've implemented my own solution, feel free to see it here: https://github.com/JetBrains/kotlin/commit/1b715ab25371b00225c9ee80299dd640c6545a6e

PR #1899, opened 2018-09-27
KT-27016 Replace 'if' with elvis operator w/ interface generates invalid code (breaks type inference)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27016


mglukhikh:
I added one more example (see last comment in the issue). Also, I don't like your check itself, I'd say it's easier to get variable type from its `VariableDescriptor`. Will try to rework myself.

mglukhikh:
Merged and extended by the following commit: https://github.com/jetbrains/kotlin/commit/9b4bcabbd3104a5a8bdef36d4a690d171d13bd06. Thank you.

PR #1901, opened 2018-09-28
KT-27104 False positive "Convert call chain into Sequence" with groupingBy

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27104


idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertCallChainIntoSequenceInspection.kt
```diff
@@ -130,6 +130,7 @@ private fun KtQualifiedExpression.collectCallExpression(context: BindingContext)
         .dropWhile { !it.isTransformationOrTermination(context) }
         .takeWhile { it.isTransformationOrTermination(context) && !it.hasReturn() }
         .toList()
+        .dropLastWhile { it.calleeExpression?.text == "groupingBy" }
```
mglukhikh: Why you did not want just to remove `groupingBy` from termination list, is there some reason I do not see?

mglukhikh: Ok, I will add this as an additional test and probably create `listOf("groupingBy")` as a list of some "lazy terminators".

mglukhikh:
Merged manually

PR #1902, opened 2018-09-28
KT-26571 Too much highlighting from "convert call chain into sequence"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26571


udalov:
Minor, but performance is usually said to be _improved_, not increased. Maybe "increase" is also fine in some contexts, but here performance is not a number, so it feels incorrect

mglukhikh:
Merged manually

PR #1906, opened 2018-10-01
KT-5088 "Add else branch" quickfix on when should not add braces

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-5088


idea/src/org/jetbrains/kotlin/idea/quickfix/AddWhenElseBranchFix.kt
```diff
@@ -36,10 +39,11 @@ class AddWhenElseBranchFix(element: KtWhenExpression) : KotlinQuickFixAction<KtW
     override fun invoke(project: Project, editor: Editor?, file: KtFile) {
         val element = element ?: return
         val psiFactory = KtPsiFactory(file)
-        val entry = psiFactory.createWhenEntry("else -> {}")
+        val entry = psiFactory.createWhenEntry("else -> TODO()")
```
mglukhikh: Not quite sure adding `TODO()` is a good idea here (because removing it can be also not easy enough). I'd agree with the suggestion from an issue comment: "generate `"else -> "` and put the caret after it (do not forget space after the arrow: `"-> "`!)"

mglukhikh:
Merged manually, thank you

PR #1573, opened 2018-03-19
Simplification of 'getOrElseNullable' and 'getOrPut'

elifarley:
This PR removes some statements and also decreases nesting level, which I contend that makes the code easier to read and understand.

PR #1908, opened 2018-10-02
Simplify MutableMap.getOrPut

cypressious:


ilya-g:
There was already an attempt to simplify this function (see #1573), but unfortunately it didn't bring any advantages to its callers. 
On the contrary it leads to a small increase of the inlined bytecode size, and that doesn't justify a slight readability improvement this change tries to achieve.

cypressious:
Alright, closing this then.

PR #1914, opened 2018-10-06
[KT-27068] Add Community Discord and Slack

MattIPv4:
Relates to https://youtrack.jetbrains.com/issue/KT-27068


orangy:
JetBrains doesn't maintain Discord and no official support for it is planned. Please don't submit any more PRs to our public pages. The badge you use cannot be accepted since it is not our official resource.

PR #1916, opened 2018-10-10
KT-27408 "Add braces to 'if' statement" moves end-of-line comment inside an `if` branch

DeFuex:
https://youtrack.jetbrains.com/issue/KT-27408

### DIFF
--- a/idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
+++ b/idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
@@ -33,6 +33,8 @@ class CommentSaver(originalElements: PsiChildRange, private val saveLineBreaks: 
     private val SAVED_TREE_KEY = Key<TreeElement>("SAVED_TREE")
     private val psiFactory = KtPsiFactory(originalElements.first!!)
 
+    private var testCount = 0
+
     private abstract class TreeElement {
         companion object {
             fun create(element: PsiElement): TreeElement? {
@@ -339,7 +341,7 @@ class CommentSaver(originalElements: PsiChildRange, private val saveLineBreaks: 
             }
 
             // shift (possible contained) comment in expression underneath braces
-            if (isSingleExpressionWithCommentBeneath) {
+            if (isSingleExpressionWithCommentBeneath && resultElements.count() == 1) {
                 val element = resultElements.first
                 element?.add(psiFactory.createWhiteSpace("\n"))
                 element?.add(restored)
--- a/idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
+++ b/idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
@@ -140,7 +140,6 @@ class CommentSaver(originalElements: PsiChildRange, private val saveLineBreaks: 
     private val lineBreaksToRestore = ArrayList<LineBreakTreeElement>()
     private var toNewPsiElementMap by Delegates.notNull<MutableMap<TreeElement, MutableCollection<PsiElement>>>()
     private var needAdjustIndentAfterRestore = false
-    private var isSingleExpressionWithCommentBeneath = false
 
     init {
         if (saveLineBreaks || originalElements.any { it.anyDescendantOfType<PsiComment>() }) {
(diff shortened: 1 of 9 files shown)


mglukhikh:
Yes, it's something close to the thing we need here. Will review it more attentively. Generally, I don't like the fact we have to create comment first, then create it again at another place, and then delete it at the old place. You may think about it if your want, however, I'm not exactly sure that better solution is possible here.

DeFuex:
i agree, it's a little messy, but it's the first possible solution i came up with.

idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
```diff
@@ -76,6 +77,20 @@ class AddBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class.ja
 
         val result = expression.replace(psiFactory.createSingleStatementBlock(expression))
 
+        // Ensure that expression is in single line, due to possible multiple comments
```
mglukhikh: And yes, you should add tests to this pull request. You need to create some files in `idea\testData\intentions\<YourIntention>` (look at existing examples please), then run "Generate All Tests" and then run `IntentionTestGenerated` to check your tests are passing. Please add tests to this pull request.

DeFuex:
I tried adding a check to be sure that there's a comment underneath a single expression. Doesn't seem to work quite right. Could be that i didn't quite fully understand the PSI system. Maybe you'll have a look at it.

idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
```diff
@@ -33,6 +33,8 @@ class CommentSaver(originalElements: PsiChildRange, private val saveLineBreaks:
     private val SAVED_TREE_KEY = Key<TreeElement>("SAVED_TREE")
     private val psiFactory = KtPsiFactory(originalElements.first!!)
 
+    private var testCount = 0
```
mglukhikh: Looks like not used

idea/idea-analysis/src/org/jetbrains/kotlin/idea/util/CommentSaver.kt
```diff
@@ -140,6 +142,7 @@ class CommentSaver(originalElements: PsiChildRange, private val saveLineBreaks:
     private val lineBreaksToRestore = ArrayList<LineBreakTreeElement>()
     private var toNewPsiElementMap by Delegates.notNull<MutableMap<TreeElement, MutableCollection<PsiElement>>>()
     private var needAdjustIndentAfterRestore = false
+    private var isSingleExpressionWithCommentBeneath = false
```
mglukhikh: Please convert it to a parameter of `restore`, with default value. Don't introduce additional class state when it's not required

idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
```diff
@@ -83,7 +99,9 @@ class AddBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class.ja
             is KtIfExpression ->
                 (result?.parent?.nextSibling as? PsiWhiteSpace)?.delete()
         }
-        saver.restore(result)
+
+        // Check for single line expression with comment beneath.
+        saver.restore(result, isCommentBeneath, false)
```
mglukhikh: Please use named argument syntax for boolean arguments

idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
```diff
@@ -63,6 +64,21 @@ class AddBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class.ja
                 semicolon.delete()
         }
 
+        // Check for single line if expression
+        if (element is KtIfExpression && expression.isInSingleLine() && element.`else` == null) {
+            // Check if a comment is actually underneath (\n) the expression
+            val allElements = element.siblings(withItself = false).filterIsInstance<PsiElement>()
```
mglukhikh: I think separate `allElements` aren't required, because anyway you perform additional filtering for `sibling`.

DeFuex: So, what is your suggestion? that i chain every filtering and avoid the declaration of `allElements`? Don't i need Sequence of PsiElements?

mglukhikh:
Ok, I will take it on remote run and merge if successful.

mglukhikh:
Merged, congratulations!

PR #1920, opened 2018-10-11
Do not configure module SDK after the Gradle import if it was set explicitly with the Gradle Idea plugin (fixes KT-27530)

ingokegel:
If `kotlinOptions.jdkHome` is set, this will still override the setting from the Gradle Idea plugin.


asedunov:
Merged! Thank you

PR #1925, opened 2018-10-15
KT-27539 False positive `Redundant Companion reference` when val in companion is effectively shadowed by inherited val

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27539

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
@@ -18,12 +18,13 @@ import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.idea.search.usagesSearch.descriptor
 import org.jetbrains.kotlin.idea.util.getResolutionScope
 import org.jetbrains.kotlin.incremental.components.NoLookupLocation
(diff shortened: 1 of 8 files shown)


mglukhikh:
Extended by https://github.com/JetBrains/kotlin/commit/c5c0cbccde1e53a651b6a2815129f4075e77eb96 (see commit message).

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
```diff
@@ -47,24 +48,25 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
...
-                        if (functionParams.size == calleeParams.size &&
-                            functionParams.zip(calleeParams).all { it.first.type == it.second.type }
-                        ) return
-                    }
+                    )
+                    if (functions.any {
+                            val functionParams = it.valueParameters
+                            val calleeParams = (selectorExpression as? KtCallExpression)?.calleeExpression?.getCallableDescriptor()
+                                ?.valueParameters.orEmpty()
+                            functionParams.size == calleeParams.size && functionParams.zip(calleeParams).all { param ->
```
mglukhikh: Here we can have problems with default parameters. I think it's better to simplify this part and check just by name.

mglukhikh:
Merged 👍 manually

PR #1928, opened 2018-10-17
KT-22704 - Allow expect annotations with actual typealias to Java to have default argument values both in expected and in actual 

ricmf:
This pull request allows removes the conflicting values error for Literals, but not arrays or annotations, because i didn't figure out how to extract the compile time value from a PsiCall. I could traverse the call arguments until they're a literal expression but I don't know if that's the right approach.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/checkers/ExpectedActualDeclarationChecker.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/checkers/ExpectedActualDeclarationChecker.kt
@@ -17,7 +17,10 @@
 package org.jetbrains.kotlin.resolve.checkers
 
 import com.intellij.openapi.vfs.VfsUtilCore
+import com.intellij.psi.PsiAnnotationMethod
+import com.intellij.psi.PsiCall
 import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiLiteral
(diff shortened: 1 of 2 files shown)


udalov:
Thank you for the contribution! I've taken a look and figured out a way to avoid dealing with the PSI directly and support most of the annotation argument kinds, but that involved using code from another module (`frontend.java`), so I've moved this logic there: https://github.com/JetBrains/kotlin/commit/5fa6f7db5a0e7eea5af00df5ac3fdf7ff1cc449f

I'm now finishing work on my other branch which would help in supporting class literal (`.class`) arguments here too, after which I'll merge this branch to master. Sadly, I have to postpone the support of annotations as annotation arguments because it requires a major refactoring around `AnnotationValue` and related classes. But otherwise, all argument kinds are supported and I've added them all to the [test](https://github.com/JetBrains/kotlin/commit/5fa6f7db5a0e7eea5af00df5ac3fdf7ff1cc449f#diff-8c2edf86e67b9e4870c6d14647335866R49).

udalov:
I've merged your change as c08540175b, and my follow-ups as 49d6a7a7cb and 3bfe138dbe. These changes will be available since Kotlin 1.3.20. I've reported [KT-28077](https://youtrack.jetbrains.com/issue/KT-28077) for the remaining case with the annotation values in annotation arguments.

Once again, thanks for the contribution!

PR #1929, opened 2018-10-17
Restore `buildSequence` and `buildIterator` samples

Mishkun:
It seems that `buildSequence` and `buildIterator` samples were accidentally
replaced by `sequence` and `iterator` ones.
All samples in [SequenceBuilder docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental/-sequence-builder/index.html) are irrelevant to experimental builders

In this pull-request I duplicate the [SequenceScope docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/-sequence-scope/index.html) samples, replace 1.3 builders with 1.2 experimental ones and update paths to samples in experimental package

### DIFF
--- a/libraries/stdlib/coroutines-experimental/src/kotlin/coroutines/experimental/SequenceBuilder.kt
+++ b/libraries/stdlib/coroutines-experimental/src/kotlin/coroutines/experimental/SequenceBuilder.kt
@@ -17,8 +17,8 @@ import kotlin.experimental.ExperimentalTypeInference
  *
  * @see kotlin.sequences.generateSequence
  *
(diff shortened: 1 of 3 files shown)


ilya-g:
Thank you for the effort.

We do not encourage using experimental coroutines anymore, when there are stable ones in 1.3.

If you find that reusing samples from the stable sequence/iterator builders is very confusing, we could consider removing them entirely.

Mishkun:
@ilya-g I removed the SequenceBuilder samples and provided notes about it being deprecated in favor of SequenceScope in v1.3.

PR #1930, opened 2018-10-17
KT-27445: Add Quick Fix for DEPRECATED_JAVA_ANNOTATION compiler warning

inktomi:
This adds a Quick Fix that allows users to automatically switch deprecated Java annotations with their Kotlin counterparts. The fix does not attempt to migrate arguments for the annotation as that would introduce significant complexity and be frail for changes to the detection in the future. I can, however, add this logic if requested though I may need some pointers towards wrangling the needed PSI statements. 

I included tests for both annotations without arguments like `@Documented`  and with arguments like `@Retention`.  I also tested each annotation supported by the `JavaAnnotationCallChecker` manually using the _runIde_ build target. 

This fixes https://youtrack.jetbrains.com/issue/KT-27445


mglukhikh:
Yes, it really would be great to migrate arguments. However, PR can be accepted as is and enhanced later. Thank you!

PR #1933, opened 2018-10-19
WIP: KT-19762 Add inspection for unclosed `(Auto)Closeable` with quickfix to wrap usage in `use`

DeFuex:


idea/src/org/jetbrains/kotlin/idea/intentions/WrapResourceWithUseCallIntention.kt
```diff
@@ -0,0 +1,182 @@
...
+import org.jetbrains.kotlin.idea.intentions.loopToCallChain.previousStatement
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.addRemoveModifier.addModifier
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.referenceExpression
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.types.Variance
+
+class WrapResourceWithUseCallInspection : AbstractKotlinInspection() {
```
mglukhikh: No tests are inside. Please add them. You could use `LocalInspectionTestGenerated` test group, see `idea/testData/inspectionsLocal`

cypressious:
Do you check anywhere that the expression is actually of type `Closeable`?

DeFuex:
@cypressious I actually added a lambda expression check, which should be comparable to a closeable type in that regard? Maybe i didn't understand the context of your question, but i hope that this check solves it.

@mglukhikh i added tests, but i'm not sure what i'm doing wrong. the two tests fail atm, due to some range error. Looking closely i think there are problems with the TextRange class used for highlighting maybe? i tried several different approaches though.

cypressious:
@DeFuex What I mean is, nowhere in your code is there a check that the expression is of type `Closable`.

How do you distinguish between 

```kotlin
val stream = FileOutputStream("")
stream.write("")
```

and 

```kotlin
val string = String()
string.toUpperCase()
```

?

DeFuex:
@cypressious @mglukhikh Just added a check for expressions containing Closeable as superclass. What's not that nice about this solution is the direct descriptor name check `!(descriptor?.name.toString() == "Closeable")` ...i tried using `?as Closeable` to cast and return if it's not containing superclass Closeable, but it works. Any suggestions?

mglukhikh:
@DeFuex You should generate tests using "Generate All Tests" and then run generated tests to be sure that your code works correctly.

mglukhikh:
@DeFuex Excuse me please, I have to close this PR without accepting because of just too many problems inside. If you wish to continue working under it, please squash all your commits and try to fix all problems I will try to enumerate here (and then create a new PR).

- don't forget to generate and run tests (use "Generate All Tests" run configuration)
- pay attention to IDE warning (condition is always true, potential memory leak in quick-fix, etc), you should not have them remaining in your code
- don't leave TODO in your code. This task is not so complex to leave some parts for future
- large if / else in your visitor contains code which is mostly duplicating
- look at ConvertTryFinallyToUseCallIntention which solves similar task, may be some code could be reused
- your comparison of descriptor name with Closeable don't probably work, you should use qualified name here
- all your tests are very strange. Where do you see any resources inside them? Why in this situations we should use this syntax with `use`? Please use real resources, like `BufferedWriter` for example, in tests
- use highlight type GENERIC_ERROR_OR_WARNING, not UNUSED_SYMBOL
- think carefully about inspection & fix message, they should be clear for user
- and so...

DeFuex:
First of all, thanks for the review 👍 , but

1) some of the answers given i asked early on, like which highlighting (wasn't really clear to me)
2) ConvertTryFinallyToUseCallIntention isn't written with `AbstractKotlinInspection` and only the actual quickfix work was reuseable.
3) My comparison of descriptor name with Closable works, since it's practical testable without writing tests.
4) My tests are strange because i didn't come around making them work properly yet.
5) Due to the latest kotlin version and the merge (i know, i didn't use rebase) involved i had to refactor some of the code to work again. hence, duplication, but it works.

6) i'll gladly squash all commits and create a new PR 😃

PR #1936, opened 2018-10-22
KT-5071 Properly surround a function invocation in string template by curly braces

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-5071

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToFunctionInvocationFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToFunctionInvocationFix.kt
@@ -28,11 +28,21 @@ class ChangeToFunctionInvocationFix(element: KtExpression) : KotlinQuickFixActio
 
     public override fun invoke(project: Project, editor: Editor?, file: KtFile) {
         val element = element ?: return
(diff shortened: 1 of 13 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeToFunctionInvocationFix.kt
```diff
@@ -31,7 +28,11 @@ class ChangeToFunctionInvocationFix(element: KtExpression) : KotlinQuickFixActio
 
     public override fun invoke(project: Project, editor: Editor?, file: KtFile) {
         val element = element ?: return
-        element.replace(KtPsiFactory(file).createExpressionByPattern("$0()", element))
+        val nextLiteralString = element.parent.nextSibling as? KtLiteralStringTemplateEntry
+        val parentheses = nextLiteralString?.text
+            ?.takeIf { it.startsWith("(") && it.endsWith(")") }?.also { nextLiteralString.delete() }
```
mglukhikh: I would consider also the case like `"$bar(1, 2) sometext"`

mglukhikh:
Merged, thank you!

PR #1937, opened 2018-10-23
Resolve DIR properly in compiler CLI BASH scripts when they are symlinked

svitalsky:
In case I have symlinks such as
`/usr/local/bin/kotlinc-jvm → /opt/kotlin/1.2.71/bin/kotlinc-jvm`
and try to run Kotlin REPL issuing `$ kotlinc-jvm` I get error like this:
`/usr/local/bin/kotlinc: line 75: /usr/local/bin/kotlin-compiler: No such file or directory`

The reason is that shell variable DIR is not resolved properly in case of such symlinks in these scripts. This patch fixes that.

### DIFF
--- a/compiler/cli/bin/kotlin
+++ b/compiler/cli/bin/kotlin
@@ -16,7 +16,8 @@
 
 export KOTLIN_RUNNER=1
 
-DIR="${BASH_SOURCE[0]%/*}"
+REAL_SOURCE="$( readlink -f "${BASH_SOURCE[0]}" )"
+DIR="${REAL_SOURCE%/*}"
 : ${DIR:="."}
 
 "${DIR}"/kotlinc "$@"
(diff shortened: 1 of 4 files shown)


svitalsky:
I've just created an issue regarding this: https://youtrack.jetbrains.com/issue/KT-27880

udalov:
Thanks for your contribution!

I'm afraid this approach won't work on macOS because `readlink` on Darwin is derived from BSD and has a [slightly different syntax](https://stackoverflow.com/q/1055671/288456), for example `-f` there means the format, not the option to canonicalize the path. Also, I'm having some doubts in general about the problem you're seeing, let's head to the [issue](https://youtrack.jetbrains.com/issue/KT-27880) you've created and discuss it there, after which we can update the pull request as necessary.

udalov:
As discussed above and in [KT-27880](https://youtrack.jetbrains.com/issue/KT-27880), we can't accept this pull request because it won't work on macOS. However, even though KT-27880 turned out to be a different problem, it still fixes a problem reported as [KT-28217](https://youtrack.jetbrains.com/issue/KT-28217). I've left a reference to this pull request there, and we'll get back to it when fixing KT-28217. Thank you again!

PR #1938, opened 2018-10-24
KT-2029 Add inspection for boolean literals passed without using named parameters feature

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-2029


idea/src/org/jetbrains/kotlin/idea/inspections/BooleanLiteralArgumentInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+
+            holder.registerProblem(
+                argument,
+                "Boolean literal arguments",
+                highlightType,
+                fix
+            )
+        })
+
+    private class AddNameFix : LocalQuickFix {
```
mglukhikh: There is already `AddNameToArgumentFix`. Why didn't use it?

idea/src/META-INF/plugin-common.xml
```diff
@@ -3078,6 +3078,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.BooleanLiteralArgumentInspection"
+                     displayName="Boolean literal arguments"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="INFO"
```
mglukhikh: I confess I'm not sure about default level it should have. Will discuss it with my colleagues.

idea/src/org/jetbrains/kotlin/idea/inspections/BooleanLiteralArgumentInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    private class AddNamesToCallArgumentsFix : LocalQuickFix {
+        override fun getName() = "Add names to call arguments"
+
+        override fun getFamilyName() = name
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            val argument = descriptor.psiElement as? KtValueArgument ?: return
+            val call = argument.getStrictParentOfType<KtCallElement>() ?: return
+            AddNamesToCallArgumentsIntention.addNamesToCallArguments(call)
```
mglukhikh: Why you do not want to use `IntentionWrapper` here?

idea/src/org/jetbrains/kotlin/idea/inspections/BooleanLiteralArgumentInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+            if (argument.getArgumentName() != null) return
+            val argumentExpression = argument.getArgumentExpression() as? KtConstantExpression ?: return
+            if (argumentExpression.node.elementType != KtNodeTypes.BOOLEAN_CONSTANT) return
+            if (argumentExpression.analyze().diagnostics.forElement(argumentExpression).any { it.severity == Severity.ERROR }) return
+            val call = argument.getStrictParentOfType<KtCallElement>() ?: return
+            if (call.resolveToCall()?.resultingDescriptor?.hasStableParameterNames() != true) return
+
+            val (highlightType, fix) = when {
+                call.valueArguments.takeLastWhile { it != argument }.none { !it.isNamed() } -> GENERIC_ERROR_OR_WARNING to AddNameFix()
+                AddNamesToCallArgumentsIntention.canAddNamesToCallArguments(call) -> INFORMATION to AddNamesToCallArgumentsFix()
```
mglukhikh: This logic looks as good idea. May be, in this case we can really left `INFO` level for the inspection itself.

mglukhikh:
I approved changed however I still consider the inspection itself as very questionable. For example, I'm almost sure that we will have tons of warnings from this inspection in Kotlin project itself. Also: in case when inspection is explicitly downgraded to INFORMATION level, in IDE we have *two* actions "add name to call argument", and to remove this probably `AddNameToCallArgument` should be just converted to an inspection (with INFORMATION level or not depending on exact situation).

mglukhikh:
Finally merged this (manually).

PR #1939, opened 2018-10-25
KT-27173 "Lift return out of `...`" should work on any of targeted `return` keywords

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27173

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/LiftReturnOrAssignmentInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/LiftReturnOrAssignmentInspection.kt
@@ -17,6 +17,8 @@
 package org.jetbrains.kotlin.idea.inspections
 
 import com.intellij.codeInspection.*
+import com.intellij.codeInspection.ProblemHighlightType.GENERIC_ERROR_OR_WARNING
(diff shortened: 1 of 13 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/LiftReturnOrAssignmentInspection.kt
```diff
@@ -24,80 +24,94 @@ import org.jetbrains.kotlin.idea.intentions.branchedTransformations.isElseIf
...
-                    if (assignmentNumber > 0) {
-                        val verb = if (assignmentNumber > 1) "should" else "can"
-                        holder.registerProblemWithoutOfflineInformation(
-                                keyword,
-                                "Assignment $verb be lifted out of '${keyword.text}'",
-                                isOnTheFly,
-                                if (assignmentNumber > 1) ProblemHighlightType.GENERIC_ERROR_OR_WARNING
-                                else ProblemHighlightType.INFORMATION,
-                                LiftAssignmentOutFix(keyword.text)
+                            expression,
```
mglukhikh: May be we should extract this fragment to some local function, to avoid code duplication? What do you think, @t-kameyama ?

mglukhikh:
Thank you, merged.

mglukhikh:
Good job!

PR #1940, opened 2018-10-26
KT-25171 Inspection: Change indexed access operator on maps to `Map.getValue`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25171


mglukhikh:
Looks good, I will merge it, thank you.

PR #1943, opened 2018-10-29
Improve initialization performance of SynchronizedLazyImpl

negator:
The current version of SynchronizedLazyImpl forces all waiting threads to block on a mutex, which can hamper performance when there are many waiting threads and if the initialization block is a time consuming operation. 

This change introduces an [AQS](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/AbstractQueuedSynchronizer.html) implementation which mimics a BooleanLatch (as described in the AQS javadoc), to allow all waiting reader threads to acquire a shared lock immediately after the value has been initialized, rather then forcing them to acquire a mutex. The value is guaranteed to be initialized by a single writer thread.

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt
+++ b/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt
@@ -52,11 +52,18 @@ public actual fun <T> lazy(lock: Any?, initializer: () -> T): Lazy<T> = Synchron
 
 
 
-private class SynchronizedLazyImpl<out T>(initializer: () -> T, lock: Any? = null) : Lazy<T>, Serializable {
+private class SynchronizedLazyImpl<out T>(initializer: () -> T, lock: Any? = null) : Lazy<T>, java.util.concurrent.locks.AbstractQueuedSynchronizer(), Serializable {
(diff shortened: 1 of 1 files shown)


ilya-g:
Interesting approach to the initialization synchronization, however what concerns me is the increased memory footprint due to using `AbstractQueuedSynchronizer`. Did you estimate how much more memory it would require for:
- uninitialized lazy
- uncontended lazy initialization
- contentded lazy initialization
- initialized lazy?

negator:
This is a great point. AQS uses an internal queue of waiting threads, which will impact the memory footprint as the number of threads grows. However, AQS is the basis for most Java synchronization primitives so I would assume it is implemented optimally for the common use cases. Although, if the goal is to minimize the footprint to greatest extent, this is probably not appropriate.

I does occur to me now that we can simplify this greatly, using a global AtomicReferenceUpdater to simulate an exclusive write lock, and a volatile boolean to simulate the latch. This would impose a small constant amount of memory of an object and a volatile boolean for every lazy val.

ilya-g:
I'm closing this PR since there's no recent activity here.
Thanks for an interesting implementation idea anyway.

PR #1944, opened 2018-10-30
KT-17502 Do not disable "Generate equals and hashCode" actions for data classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17502


asedunov:
Merged to master. Thank you!

PR #1945, opened 2018-10-30
Update for antlr dependency and inroduce gradle to build in idl2k

Schahen:
1. Gradle was introduced because (at least it looks like so pretty much)
maven exec target were already broken and it's easier to run compiled
code from gradle.
1.1 So far gradle project does not belong to any parent project (that
is, it's have a separate settings.gradle file)
1.2. I removed from maven exec task but left it -  for a while it feels safer
to have an opportunity to build in maven as well.

2. Antlr was upgraded from 4.5.3 to 4.7.1 because some fixes (including
in java-generated code) has been introduced and I decided that this PR
is a nice opportunity to have such update.
2.1 I'd have to slightly modify WebIDL.g4 definition though -
and updated copyright accordingly - my updates were based on
https://raw.githubusercontent.com/antlr/grammars-v4/master/webidl/WebIDL.g4
2.2. I've checked all generated output to make sure it is identical to
what we hade before dependency update.

3. Packaging was introduced because it's best practices - if you have
package declaration to keep files in package I mean. Also, IDEA treats
such code better by default.


ilya-g:
In pure Kotlin project there's no need to create all these subdirectories for common package prefix: `.../org/jetbrains/idl2k/...`

libraries/tools/idl2k/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+
+generateGrammarSource {
+    arguments += ["-visitor", "-long-messages", "-package", "org.antlr.webidl"]
+}
+
+
+compileKotlin.dependsOn generateGrammarSource
+compileTestKotlin.dependsOn generateTestGrammarSource
+
+task downloadAllIdls(type:JavaExec) {
```
ilya-g: add space
```suggestion
task downloadAllIdls(type: JavaExec) {
```

ilya-g:
I don't think it's worth to leave this project in the state of dual maven/gradle build, so it makes sense to include it in the main Kotlin project.

Currently this project is deployed when `idl2k.deploy.skip` system property is set to `false` and then that artifact is used in `kotlin-native` to generate declarations from IDL for WebAssembly target (@alexander-gorshenev, correct me if I'm wrong).

Schahen:
I actually agree, so if  @alexander-gorshenev won't have objections - I'll remove pom.xml completely. As of now the only thing that is done there and can not be done in gradle - is exactly invoking
maven-deploy-plugin

ilya-g:
We can setup deployment as in other libraries projects, e.g. https://github.com/JetBrains/kotlin/blob/769a0d7701a7eae9ee7b49e4a26e62961eca8392/libraries/tools/kotlin-annotations-jvm/build.gradle#L8
though it should be done conditionally, based on the value of that `idl2k.deploy.skip` property.

Schahen:
Removed maven build, made idl2k part of default build lifecycle

libraries/tools/idl2k/build.gradle
```diff
@@ -0,0 +1,41 @@
+buildscript {
+    ext.antlr4_version = '4.7.1'
+}
+
+apply plugin: "antlr"
+apply plugin: "kotlin"
+
+configurePublishing(project)
```
ilya-g: Please preserve the old behavior: the artifact was published only when `idl2k.deploy.skip` property is `false`

ilya-g: Note that the artifact name is changed to `kotlin-idl2k` (by default it's the name of the project). It may be ok or may be not, please ask @alexander-gorshenev, are they ready for that?

alexander-gorshenev: We're fine with that, I suppose.

ilya-g: I've checked that the property was set only for the maven build, so we should setup something like that to preserve the same for gradle build — I can do that. 
But we do not longer need to preserve the inverted logic, i.e. publishing when the property is false, so I propose to introduce new property `idl2k.deploy` and publish only when it is set to `true`.

Schahen: @ilya-g  I agree, inverted boolean logic can be confusing

ilya-g: No, by default it's `false`

libraries/tools/idl2k/build.gradle
```diff
@@ -5,7 +5,9 @@ buildscript {
 apply plugin: "antlr"
 apply plugin: "kotlin"
 
-configurePublishing(project)
+if (System.getProperty("idl2k.deploy", "false").toBoolean()) {
```
ilya-g: `system.idl2k.deploy` teamcity paramater translates to a project property `idl2k.deploy` in Gradle, rather than a JVM system property. So we need to use gradle's `Project.findProperty/getProperty/hasProperty`

ilya-g:
Merged manually

PR #1946, opened 2018-10-30
Renamed ContractBuilder.callsInPlace 'lambda' parameter to 'funciton'

Miha-x64:
Purpose: to make clear that it not must only be lambda, but can also be a function reference, anonymous function, or a property reference coerced to getter or setter. Trying to fix a popular misunderstanding that 'lambda' means 'technically an anonymous class'.
Binary compatibility: looks like contracts take place only in Metadata, but no real invocations generated.
Source compatibility: created a deprecated extension-function which should be visible when one writes `callsInPlace(lambda = ...`.
Usages: honestly, I don't know whether this function used inside Kotlin project with named parameters.

### DIFF
--- a/libraries/stdlib/src/kotlin/contracts/ContractBuilder.kt
+++ b/libraries/stdlib/src/kotlin/contracts/ContractBuilder.kt
@@ -81,8 +81,13 @@ public interface ContractBuilder {
     * @sample samples.contracts.callsInPlaceExactlyOnceContract
     * @sample samples.contracts.callsInPlaceUnknownContract
     */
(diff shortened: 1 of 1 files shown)


ilya-g:
Doesn't the contract only have effect if the value passed to `lambda` is a lambda expression?

Miha-x64:
Interesting point...
Unfortunately, you're right, this doesn't work as I expect:
```kt
inline fun eval(func: () -> Unit) {
    contract {
        callsInPlace(func, InvocationKind.EXACTLY_ONCE)
    }
}

fun x() {
    eval(fun() { error("") })
    println("yay!")
}
```

dsavvinov:
So, I think we can leave it as is, at least until other function constructions won't be supported in contracts. Thanks for your willingness to contribute, nevertheless! :)

PR #1947, opened 2018-10-31
KT-26669 "Remove unnecessary parentheses" reports parens of function returned from extension function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26669


mglukhikh:
Looks good, thank you :)

PR #1949, opened 2018-11-01
KT-27906 SafeCastAndReturn is not reported on code block with unqualified return

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27906


mglukhikh:
Looks good, thank you

PR #1950, opened 2018-11-01
Add mininal test suite to idl2k, remove deprecated ANTLRFileStream

Schahen:
- There's only one test so far that covers the very basic functionality.
- For this particular PR the goal was not to change the code tested at all, that's the reason there's that two new lines in the generated code.
- parameter .. is never used warning fixed
- ANTLRFileStream is deprecated and removed

### DIFF
--- a/libraries/tools/idl2k/src/main/kotlin/build.kt
+++ b/libraries/tools/idl2k/src/main/kotlin/build.kt
@@ -6,21 +6,6 @@ import java.io.IOException
 import java.net.URL
 import java.util.ArrayList
 import java.util.LinkedHashMap
-import kotlin.collections.HashSet
-import kotlin.collections.List
-import kotlin.collections.Map
-import kotlin.collections.MutableList
-import kotlin.collections.asSequence
-import kotlin.collections.contains
-import kotlin.collections.emptyMap
-import kotlin.collections.fold
-import kotlin.collections.forEach
-import kotlin.collections.map
-import kotlin.collections.mapValues
-import kotlin.collections.plus
-import kotlin.collections.reduce
-import kotlin.collections.set
-import kotlin.collections.sortedBy
 
 class BuildWebIdl(val mdnCacheFile: File, val srcDir: File) {
     val repositoryPre = loadPreliminaryRepository()
--- a/libraries/tools/idl2k/src/main/kotlin/build.kt
+++ b/libraries/tools/idl2k/src/main/kotlin/build.kt
@@ -4,11 +4,23 @@ import org.antlr.v4.runtime.CharStreams
 import java.io.File
 import java.io.IOException
 import java.net.URL
-import java.nio.charset.StandardCharsets
 import java.util.ArrayList
 import java.util.LinkedHashMap
 import kotlin.collections.HashSet
-
+import kotlin.collections.List
+import kotlin.collections.Map
+import kotlin.collections.MutableList
+import kotlin.collections.asSequence
+import kotlin.collections.contains
+import kotlin.collections.emptyMap
+import kotlin.collections.fold
+import kotlin.collections.forEach
+import kotlin.collections.map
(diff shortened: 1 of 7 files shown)


libraries/tools/idl2k/src/main/kotlin/build.kt
```diff
@@ -18,16 +19,17 @@ class BuildWebIdl(val mdnCacheFile: File, val srcDir: File) {
             System.exit(1)
         }
 
-        return srcDir.walkTopDown().filter { it.isDirectory || it.extension == "idl" }.asSequence().filter { it.isFile }.toList().sortedBy { it.absolutePath }.fold(Repository(emptyMap(), emptyMap(), emptyMap(), emptyMap())) { acc, e ->
+        return srcDir.walkTopDown().filter { it.isDirectory || it.extension == "idl" }.asSequence().filter { it.isFile }.toList()
+            .sortedBy { it.absolutePath }.fold(Repository(emptyMap(), emptyMap(), emptyMap(), emptyMap())) { acc, e ->
             System.err.flush()
             System.err.println("Parsing ${e.absolutePath}")
-            val fileRepository = parseIDL(ANTLRFileStream(e.absolutePath, "UTF-8"))
+            val fileRepository = parseIDL(CharStreams.fromFileName(e.absolutePath, StandardCharsets.UTF_8))
```
ilya-g:
```suggestion
            val fileRepository = parseIDL(CharStreams.fromFileName(e.absolutePath, Charsets.UTF_8))
```
`Charsets` object available in `kotlin.text`

libraries/tools/idl2k/src/main/kotlin/build.kt
```diff
@@ -1,12 +1,13 @@
...
-import org.antlr.v4.runtime.ANTLRFileStream
-import kotlin.collections.HashSet
+import org.antlr.v4.runtime.CharStreams
 import java.io.File
 import java.io.IOException
 import java.net.URL
+import java.nio.charset.StandardCharsets
 import java.util.ArrayList
 import java.util.LinkedHashMap
+import kotlin.collections.HashSet
```
ilya-g: import looks redundant

libraries/tools/idl2k/src/main/kotlin/build.kt
```diff
@@ -1,13 +1,26 @@
...
+import kotlin.collections.contains
+import kotlin.collections.emptyMap
+import kotlin.collections.fold
+import kotlin.collections.forEach
+import kotlin.collections.map
+import kotlin.collections.mapValues
+import kotlin.collections.plus
+import kotlin.collections.reduce
+import kotlin.collections.set
+import kotlin.collections.sortedBy
```
ilya-g: `kotlin.collections` package is imported by default. Why these imports are required? 😕

libraries/tools/idl2k/src/test/kotlin/Idl2kTest.kt
```diff
@@ -0,0 +1,58 @@
...
+            "src/test/resources/SomethingNotInCache.idl", """public external open class SomethingNotInCache {
+    open val someReadOnlyParam: dynamic
+    var someWriteableParam: dynamic
+    fun someEmptyMethod(): String
+    fun someMethod(root: dynamic): String
+    fun optionalUsvStringFetcher(name: String): String?
+}
+
+
+"""
```
ilya-g: Perhaps it would be better to read expected output from a file too

Schahen: I suggest to have this step after the code will be refactored so  that there will be no two new lines. The thing is in file it's less obvious that there are two empty lines.

ilya-g (approved):

Schahen:
Merged as squashed and cherry-picked commit

PR #1954, opened 2018-11-02
Do not resolve configuration for scripting Gradle plugin

gavra0:
Avoid resolving kotlinScriptDef configurations during
configuration phase. Also, make task creation lazy.


alexey-tsvetkov:
@gavra0 we cannot use configuration avoidance API to be compatible with older versions of Gradle. At least without checking Gradle version first. I have a non-pushed branch containing the fix for the same issue (delayed pushing it because I wanted to resolve KT-25749 blocked by KT-26065).
Besides the usage of Lazy Task API, the fix looks good to me.

alexey-tsvetkov:
@gavra0 if making the task lazy is not critical for you now, I've reverted that part of your commit and pushed the branch for remote run rr/gradle/avoid-script-conf-resolution
Will merge after RR.

gavra0:
That sound good. Thanks for the review, and for fixing the backwards compatibility. We can ignore the lazy tasks for now.

alexey-tsvetkov:
Merged manually.
@gavra0 thanks for your contribution!

PR #1955, opened 2018-11-02
Register KotlinModelBuilder for kotlin-android plugin

Slvin:
This change ensures that the KotlinModelBuilder is registered for the kotlin-android plugin. This is required since the Android IDEA plugin (and Android Studio) lacks the information needed to configure the Kotlin facet. Information we are unable to get from the Android specific models include, compiler arguments and co-routine support.

This change also ensures that the KotlinProject that is built for Android projects has its project type set to PLATFORM_JVM and that the above information is set correctly on the model. Note: for some fields that we obtain from the Android Gradle Plugin we leave empty or return placeholder values.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/model/builder/KotlinModelBuilder.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/model/builder/KotlinModelBuilder.kt
@@ -26,7 +26,7 @@ import org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile
 
 /**
  * [ToolingModelBuilder] for [KotlinProject] models.
- * This model builder is registered for base Kotlin JVM, Kotlin JS and Kotlin Common plugins.
+ * This model builder is registered for base Kotlin JVM (including Android), Kotlin JS and Kotlin Common plugins.
  */
 class KotlinModelBuilder(private val kotlinPluginVersion: String) : ToolingModelBuilder {
 
@@ -42,7 +42,9 @@ class KotlinModelBuilder(private val kotlinPluginVersion: String) : ToolingModel
                 project.name,
                 kotlinPluginVersion,
                 projectType,
-                kotlinCompileTasks.mapNotNull { it.createSourceSet(project, projectType) },
+                kotlinCompileTasks.mapNotNull {
+                    if (project.isAndroid()) it.createAndroidSourceSet() else it.createSourceSet(project, projectType)
+                },
                 getExpectedByDependencies(project),
                 kotlinCompileTasks.first()!!.createExperimentalFeatures()
             )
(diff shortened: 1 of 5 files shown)


Slvin:
I'm also currently working on a way to correctly fill the missing fields while constructing the SourceSetImpl so that in Android Studio we can correctly handle Kotlin srcDirs.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/model/builder/KotlinModelBuilder.kt
```diff
@@ -91,6 +101,33 @@ class KotlinModelBuilder(private val kotlinPluginVersion: String) : ToolingModel
...
+        /**
+         * The [SourceSet] returned by this method is not intended to be complete, most of the information here is exposed by the
+         * Android Gradle plugin and as such is not populated here. The important information here that is required by
+         * Android Studio are the [CompilerArguments].
+         */
+        private fun AbstractKotlinCompile<*>.createAndroidSourceSet(androidTarget: KotlinAndroidTarget): SourceSet {
+            val variantName = sourceSetName
+            val compilation = androidTarget.compilations.getByName(variantName)
+            // Merge all sources and resource dirs from the different Source Sets that make up this variant.
+            val sources = compilation.kotlinSourceSets.flatMap {
```
h0tk3y: Use [`allKotlinSourceSets`](https://github.com/JetBrains/kotlin/blob/a44d70e79d06a5133b86470c749cc2d0bf870053/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/kotlinCompilations.kt#L166) instead. The difference between them is that `kotlinSourceSets` only contains the source sets added directly to the compilation, while `allKotlinSourceSets` also includes source sets that are transitively added to the compilation through the `dependsOn` relation.

h0tk3y:
Pushed for remote run https://teamcity.jetbrains.com/viewQueued.html?itemId=1746965. If the tests pass, I'll fix the issue I mentioned above and merge the PR.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/model/builder/KotlinModelBuilder.kt
Slvin: Changed to allKotlinSourceSets

Slvin:
> Pushed for remote run https://teamcity.jetbrains.com/viewQueued.html?itemId=1746965. If the tests pass, I'll fix the issue I mentioned above and merge the PR.

Thanks! I've made the change you mentioned above as I noticed that the method comment was out of date, I hope this doesn't mess up anything.

h0tk3y:
@Slvin Thanks for your contribution!

Slvin:
@h0tk3y Do you know what will be the earliest version to include these changes?

h0tk3y:
@Slvin 1.3.20, which is the first 1.3.x tooling update that is not a bugfix release.

Slvin:
@h0tk3y Great, thanks again :)

PR #1957, opened 2018-11-05
KT-27861 RedundantCompanionReference false positive for nested class with name "Companion"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27861

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
@@ -40,6 +40,7 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
             if (expression.text != objectDeclaration.name) return
 
             val containingClass = objectDeclaration.containingClass() ?: return
+            if (parent.getStrictParentOfType<KtEnumEntry>()?.containingClass() == containingClass) return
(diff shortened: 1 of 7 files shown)


idea/testData/inspectionsLocal/redundantCompanionReference/inEnumEntry.kt
```diff
@@ -0,0 +1,10 @@
+// PROBLEM: none
+// DISABLE-ERRORS
+
+enum class E(val value: String) {
+    E1(<caret>Companion.foo);
```
TWiStErRob: This is a valid warning, `E.Companion.foo` is being referenced as `Companion.foo`, which is redundant since the `companion object` is in scope. This should be `E1(foo)` in `.after.kt`.

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
```diff
@@ -40,6 +40,7 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
             if (expression.text != objectDeclaration.name) return
 
             val containingClass = objectDeclaration.containingClass() ?: return
+            if (parent.getStrictParentOfType<KtEnumEntry>()?.containingClass() == containingClass) return
```
TWiStErRob: Self-referencing our own `companion object` is always redundant, similar to how we never qualify static methods when inside the class that declares it.

PR #1958, opened 2018-11-06
KT-27184 "Replace with safe call" is not suggested for nullable var property that is impossible to smart cast

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27184


mglukhikh:
I think (not exactly sure) that existing local quick fix and intention can be just re-used here, without modifying their source code. You can try to do this yourself, or I will think about it more attentively later.

t-kameyama:
I'll try to do.

mglukhikh:
I think it should be accepted now.

mglukhikh:
Merged 👍 and then modified a bit

PR #1959, opened 2018-11-06
Make context available to suppressors

jimgoog:
We've found that the BindingContext is necessary for any meaningful suppressor.
This change adds the ability for suppressors to use the BindingContext when available.

The change was intentionally formulated to minimize churn/changes to the codebase, but if you would prefer that we have only a single `isSuppressed` method on the interface (specifically, the isSuppressed that includes the BindingContext), I'm happy to update the PR.  Or if you have any other feedback, just let me know.

cc @erokhins

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/diagnostics/KotlinSuppressCache.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/diagnostics/KotlinSuppressCache.kt
@@ -35,6 +35,7 @@ import org.jetbrains.kotlin.util.ExtensionProvider
 
 interface DiagnosticSuppressor {
     fun isSuppressed(diagnostic: Diagnostic): Boolean
+    fun isSuppressed(diagnostic: Diagnostic, bindingContext: BindingContext?): Boolean = isSuppressed(diagnostic)
 
     companion object {
         val EP_NAME: ExtensionPointName<DiagnosticSuppressor> =
@@ -68,7 +69,7 @@ abstract class KotlinSuppressCache {
 
         if (request is DiagnosticSuppressRequest) {
             for (suppressor in diagnosticSuppressors.get()) {
-                if (suppressor.isSuppressed(request.diagnostic)) return true
+                if (isSuppressedByExtension(suppressor, request.diagnostic)) return true
             }
         }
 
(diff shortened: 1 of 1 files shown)


jimgoog:
@erokhins Ping.

compiler/frontend/src/org/jetbrains/kotlin/resolve/diagnostics/KotlinSuppressCache.kt
```diff
@@ -35,6 +35,7 @@ import org.jetbrains.kotlin.util.ExtensionProvider
 
 interface DiagnosticSuppressor {
     fun isSuppressed(diagnostic: Diagnostic): Boolean
+    fun isSuppressed(diagnostic: Diagnostic, bindingContext: BindingContext?): Boolean = isSuppressed(diagnostic)
```
zarechenskiy: Why `bindingContext` is nullable here?
Note that we have `BindingContext.EMPTY` that can be used when there is no binding context

jimgoog: In my mind, there was/is a difference between a bindingContext that doesn't have the slices resolved, and being in a situation where a bindingContext is not supported.  In our case, if the bindingContext is null, we follow a different best-effort code path.

I actually don't feel strongly here.  Happy to change if BindingContext.EMPTY is more desirable.  For our best-effort codepath, we can just check if the bindingContext is equal to BindingContext.EMPTY.  Would you like me to change it to use BindingContext.EMPTY?

Longer term, I think a better solution might be to get rid of the `fun isSuppressed(diagnostic: Diagnostic)` in favor of `fun isSuppressed(diagnostic: Diagnostic, bindingContext: BindingContext)` and remove nullability.  Such that you can't have a DiagnosticSuppressor without being able to get a context for the given diagnostic.  That was a potentially bigger / moredisruptive change, which is why we opted for the less disruptive variant.  Thoughts?

jimgoog:
Ping @zarechenskiy

compiler/frontend/src/org/jetbrains/kotlin/resolve/diagnostics/KotlinSuppressCache.kt
zarechenskiy: The current implementation is totally fine, I just wanted to figure out why it's done this way. 
In addition, we have freedom here to change the method's signature (or add a new method) if necessary, so I think we can now take the current approach

PR #1962, opened 2018-11-07
Multiple templates with concatenation generator should add spaces after commas (KT-27954)

technoir42:
https://youtrack.jetbrains.com/issue/KT-27954


asedunov:
Merged! Thank you

PR #1964, opened 2018-11-07
Used System.nanoTime() for measuring elapsed time

Miclebrick:
System.currentTimeMillis() is less accurate for measuring elapsed time, as System.nanoTime() is guaranteed to have at least as good accuracy.

According to the [Oracle docs](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#currentTimeMillis()):
> while the unit of time of the return value is a millisecond, the granularity of the value depends on the underlying operating system and may be larger. For example, many operating systems measure time in units of tens of milliseconds. 

Whereas, nanotime:
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative).

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/system/Timing.kt
+++ b/libraries/stdlib/jvm/src/kotlin/system/Timing.kt
@@ -9,11 +9,7 @@ package kotlin.system
 /**
  * Executes the given [block] and returns elapsed time in milliseconds.
  */
-public inline fun measureTimeMillis(block: () -> Unit): Long {
-    val start = System.currentTimeMillis()
-    block()
-    return System.currentTimeMillis() - start
-}
+public inline fun measureTimeMillis(block: () -> Unit): Long = measureNanoTime(block) / 1000000L
 
 /**
  * Executes the given [block] and returns elapsed time in nanoseconds.


libraries/stdlib/jvm/src/kotlin/system/Timing.kt
```diff
@@ -9,11 +9,7 @@ package kotlin.system
 /**
  * Executes the given [block] and returns elapsed time in milliseconds.
  */
-public inline fun measureTimeMillis(block: () -> Unit): Long {
-    val start = System.currentTimeMillis()
-    block()
-    return System.currentTimeMillis() - start
-}
+public inline fun measureTimeMillis(block: () -> Unit): Long = measureNanoTime(block) / 1_000_000L
```
HaasJona: Rounding down always may add a systematic error here, that isn't present when simply using millisecond resolution.

HaasJona: No, I mean if you measure something that takes on 10.5 millis, with `currentTimeMillis()` you'd probably measure 11ms about half the time and 10ms the other time. If you add the times of multiple measurements together, you would get about the right result.

If you use `nanoTime()`, that will (ideally) return 10,500,000 every measurement, which will always be rounded down to 10ms, which will return a to low value, especially when adding multiple measurements together.

Miclebrick: I see what you mean, but I'm not sure if using currentTimeMillis is a better solution than just rounding it properly, since on some machines that's actually a 1/64 of a second instead of 1/1000

HaasJona: The alternative solution would be to divide both nano values by 1000000 (normal long division, so rounding down) and then computing the difference between that. That wouldn't have a bias, but it would be slower (double the divisions) and you'd need to deal with the long value potentially overflowing.

I'm not sure what the best solution is, using nanoTime is certainly the right idea, especially as it's unrelated to changes of the system clock. Just showing potential problems.

Davio: A very simple solution could be to add 500,000 before dividing by 1,000,000.

If we take the range 9,500,000 to 10,500,000 this results in:
[9,500,000 - 10,4999,999] gets rounded down to 10,000,000 (10,000,000 - 10,999,999 after adding 500,000, then dividing by 1,000,000)
10,500,000 gets rounded down to 11,000,000

PR #1965, opened 2018-11-07
KT-4645 Unexpected behevior of "Replace 'if' with 'when'" intention when called on second or third 'if'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-4645


mglukhikh:
Thank you, good work :)

PR #1967, opened 2018-11-08
KT-13311 IDE marks fun finalize() as unused and says that its effective visibility is private

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-13311


mglukhikh:
@t-kameyama I accepted it but you handled only the second part of the issue (about effective visibility). The first part is still intact (`finalize` is marked as unused).

PR #1970, opened 2018-11-08
KT-19762 Add inspection for unclosed `(Auto)Closeable` with quickfix to wrap usage in `use` 

DeFuex:

Old PR -> #1933


mglukhikh:
I still do not see generated tests. Please generate tests and run them.

oy-paddy:
雷猴

mglukhikh:
Also, there is already `ConvertTryFinallyToUseCallIntention` which performs similar transformation but assumes there is `try...finally` and `resource.close()` is performed inside `finally`. May be some parts of code could be reused.

DeFuex:
i created a new pr, since i had some issues with rebasing -> #2082

PR #1975, opened 2018-11-12
KT-27556 Add intention for collections, !collection.isEmpty() -> collection.isNotEmpty()

tommykw:
https://youtrack.jetbrains.com/issue/KT-27556

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.idea.inspections.collections.isCalling
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getLastParentOfTypeInRow
+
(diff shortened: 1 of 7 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
```diff
@@ -0,0 +1,82 @@
...
+
+class ReplaceNegatedIsEmptyWithIsNotEmptyInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+
+        return simpleNameExpressionVisitor { simpleNameExpression ->
+            if (simpleNameExpression.isEmptyNegation()) {
+                holder.registerProblem(
+                    simpleNameExpression,
+                    "Replace negated 'isEmpty' with 'isNotEmpty'",
+                    ProblemHighlightType.INFORMATION,
```
mglukhikh: Please change it to GENERIC_ERROR_OR_WARNING, otherwise users will not be able to change inspection level manually.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getLastParentOfTypeInRow
+
+class ReplaceNegatedIsEmptyWithIsNotEmptyInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+
+        return simpleNameExpressionVisitor { simpleNameExpression ->
+            if (simpleNameExpression.isEmptyNegation()) {
+                holder.registerProblem(
+                    simpleNameExpression,
```
mglukhikh: Please check highlighting range here when inspection's level is increased manually. Too wide highlighting range is bad. I think here we should highlight just `isEmpty` reference, not the whole `!someVeryImportantList.isEmpty()`

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
tommykw: Fixed.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNegatedIsEmptyWithIsNotEmptyInspection.kt
tommykw: Thank you.
I have confirmed that only isEmpty reference is highlighted.

tommykw:
Thank you for the review. Could you review this PR, again?

mglukhikh:
I think it's Ok, thank you. Will merge now.

PR #1976, opened 2018-11-12
[idea-gradle] improved transitive resolution

fkorotkov:
Fixed BFS. We need to either mark dependency nodes as visited after adding them to the queue or double check if a node has been already processed.

Fixes KT-27832

@asedunov PTAL

### DIFF
--- a/idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleProjectResolverExtension.kt
+++ b/idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleProjectResolverExtension.kt
@@ -139,6 +139,8 @@ class KotlinGradleProjectResolverExtension : AbstractProjectResolverExtension() 
 
                 while (toProcess.isNotEmpty()) {
                     val moduleNode = toProcess.pollFirst()
+                    if (processed.contains(moduleNode)) continue
+
                     processed.add(moduleNode)
 
                     val moduleNodeForGradleModel = if (useModulePerSourceSet()) {


ilya-g:
As far as I can see, transitive dependencies are already being filtered to exclude processed ones in the line https://github.com/JetBrains/kotlin/pull/1976/files#diff-43944464c59a60c7a1fcd00f7b969ed5R175.

What cases the proposed condition is supposed to cover?

fkorotkov:
@ilya-g yes, but one module can be added multiple times sine it can be already in `toProcess` but not yet marked as `processed`. Imagine a complete dependency graph `{1, 2, 3, ..., n}` where each node connected to each other:

1. We pick the node `1` and add it to `processed`.
2. Add all children to `toProcess` which are equal to `{2, ..., n}`
3. Let say we'll pick node `2` from `toProcess`.
4. We add `2` to processed which now consists of `{1, 2}`.
5. We will add **again** all nodes but `{1, 2}` to `toProcess`. Now there are duplicates in `toProcess`! `{3, ..., n, 3, ..., n}`
6. continue...

It will mean for a case of a complete dependency graph that on each step `i` we will add `n-i` nodes to `toProcess` which means in the end we'll add `O(n!)` nodes to it. **Exponential!**

ilya-g:
That sounds reasonable. Perhaps it could be better to track the set of discovered nodes instead of processed ones.

fkorotkov:
@ausatiy sounds good! I initially though in [KT-27832](https://youtrack.jetbrains.com/issue/KT-27832) that the slowness is because of `O(n^2)` but it appeared it was because of `O(n!)` 😅 Will be great to get it even faster but I had [issues with setting up Kotlin project in IntelliJ](https://youtrack.jetbrains.com/issue/KT-28082).

ausatiy:
@fkorotkov , I forgot to write a summary. Your fix improves performance of dependency resolution up to 2 times (here and later the test case I've mentioned in the issue). Adding caches gives additional speedup approx. 5 times (already done too) giving total speedup approx 10 times!
As result the whole import process becomes up to 2 times faster.

fkorotkov:
Wow! Big win! Thanks!

PR #1977, opened 2018-11-13
KT-23134 "Remove single lambda parameter" quick fix applied to a lambda parameter with explicit type breaks ::invoke reference on lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23134


mglukhikh:
Merged, thank you 👍

PR #1978, opened 2018-11-14
KT-15835 "Leaking 'this' in constructor" for enum class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15835


mglukhikh:
Great, thank you!

PR #1981, opened 2018-11-15
KT-28224 "Add braces to 'else' statement" moves comment outside braces when 'if-else' is inside 'if / when' branch 

DeFuex:
https://youtrack.jetbrains.com/issue/KT-28224


mglukhikh:
Merged manually. Thank you.

PR #1982, opened 2018-11-16
KT-8875 "Remove explicit type" produce red code for extension lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-8875

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveExplicitTypeIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveExplicitTypeIntention.kt
@@ -18,13 +18,17 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.builtins.isExtensionFunctionType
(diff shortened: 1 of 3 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/RemoveExplicitTypeIntention.kt
```diff
@@ -76,7 +80,12 @@ class RemoveExplicitTypeIntention : SelfTargetingRangeIntention<KtCallableDeclar
         fun redundantTypeSpecification(element: KtCallableDeclaration, initializer: KtExpression?): Boolean {
             if (initializer == null) return true
             if (initializer !is KtLambdaExpression && initializer !is KtNamedFunction) return true
-            val functionType = element.typeReference?.typeElement as? KtFunctionType ?: return true
+            val typeReference = element.typeReference
+            val functionType = typeReference?.typeElement as? KtFunctionType ?: return true
+
+            val type = element.analyze(BodyResolveMode.PARTIAL)[BindingContext.TYPE, typeReference]
```
mglukhikh: Why we cannot just check `functionType.receiver != null`?

mglukhikh:
Merged 👍

PR #1983, opened 2018-11-16
KT-27699 "Remove redundant spread operator" produces incorrect code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27699


mglukhikh:
Thank you :)

PR #1984, opened 2018-11-16
KT-12479 IDEA doesn't propose to replace all usages of deprecated annotation when it declared w/o parentheses

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12479


mglukhikh:
Great work, thanks!

PR #1988, opened 2018-11-16
Fix kotlinx-serialization plugin failed to generate serializer with unsigned properties

Prototik:
... also it ads basic tests for jvm version of that plugin.

I'm unable to find issue on YouTrack, so here is a short summary of the problem:

Using serialization plugin at properties with `inline class` types (especially unsigned types) isn't working right now, it throws backend exception:
```
java.lang.IllegalStateException: Backend Internal error: Exception during code generation
Cause: Back-end (JVM) Internal error: wrong code generated
org.jetbrains.kotlin.codegen.CompilationException Back-end (JVM) Internal error: Couldn't transform method node:
<init> (ILkotlin/UByte;Lkotlin/UShort;Lkotlin/UInt;Lkotlin/ULong;Lkotlinx/serialization/SerializationConstructorMarker;)V:
...
```

Seems bytecode is correct (iload vs aload /etc), but constructor signature uses reference types instead of primitives. That PR fixed this issue (but I'm not sure that it's correct way) and adds simple tests for that case.


Prototik:
Closing as I'm working on full reworking on serialization plugin.
Anyway this patch just dumb hack for special case, that's not I'm want to.

PR #1989, opened 2018-11-17
KT-28268 Don't suggest "make abstract" quick fix for inline classes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28268


mglukhikh:
Merged 👍

PR #1991, opened 2018-11-20
KT-28286 "Unused symbol" inspection: Interface is reported as "class"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28286


mglukhikh:
Merged 👍

PR #1992, opened 2018-11-20
KT-23639 False positive "Unused symbol" for sealed class type parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23639


mglukhikh:
Merged 👍

PR #1994, opened 2018-11-21
KT-28341 False positive "Introduce backing property" intention for `const` values

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28341


mglukhikh:
Merged 👍

PR #1995, opened 2018-11-22
KT-28289 "Extend selection" doesn't select blank line after method in class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28289


Alefas:
Merged manually

PR #1996, opened 2018-11-22
KT-28209 Live Template Variable "functionParameters()" inserts square brackets around parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28209


Alefas:
This PR is rejected as people have to use groovyScript to join to String as the actually need. Similar to methodParameters() in java.

PR #1997, opened 2018-11-23
Fix several obvious typos/misspels

Quasilyte:
Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


goodwinnk:
Have you used some tool for spotting those typos? I think it would be more useful to clean all such typos at once, than fixing them one by one.

Quasilyte:
@goodwinnk, OK, you can fix them all if you wish.
I'm not motivated to perform more work on this, since it requires some manual work and it can't be done "perfect" since there is always a chance that the tool you're using is not good enough.

But keep in mind that you have to check the diff anyway, since some things look like a typo for a tool but they're not.

I've used very simple misspell tool that finds few typos, but the false positive rate is low.

PR #1998, opened 2018-11-26
Fix common misspellings

MarcinMoskala:
Corrections for common misspellings that are still present in the Kotlin project


goodwinnk:
Thank you, I'm going to accept this PR after remote run.

PR #2001, opened 2018-11-26
KT-28443 "Move out of companion object" intention is suggested for @JvmField property inside companion object of interface

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28443


mglukhikh:
Merged 👍

PR #2002, opened 2018-11-26
KT-27138 Change visibility intentions are suggested on properties marked with @JvmField

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27138#focus=streamItem-27-3179248-0-0


mglukhikh:
Merged 👍

PR #2005, opened 2018-11-27
KT-11253 Implemented sumByLong

Davio:
Implemented sumByLong the same way as sumByDouble, verified and edited all related files, such as tests and IDE inspections, see https://youtrack.jetbrains.com/issue/KT-11253


idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertCallChainIntoSequenceInspection.kt
```diff
@@ -296,6 +296,7 @@ private val terminations = listOf(
     "singleOrNull",
     "sum",
     "sumBy",
+    "sumByLong",
```
0legg: I guess all these call are intentionally alphabetically ordered, thus `sumByLong` should go below `sumByDouble`

Davio:
I think we can go ahead and merge this, right?

ilya-g:
@Davio, Thank you for you effort. Unfortunately, we haven't decided yet on whether introducing this function is the right way to implement KT-11253 (see [this comment](https://youtrack.jetbrains.com/issue/KT-11253#focus=streamItem-27-3512826.0-0) for details). Until then we can't merge this PR.

ilya-g:
Since Kotlin 1.4, we finally have [`sumOf`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/sum-of.html) function which is overloaded for a multiple of numeric accumulator types. Thus, `sumByLong` is no longer needed, so I'm closing this PR.

Although I see that you have supported `sumByLong` in various inspections. AFAIK, this has not been done for the new `sumOf`. If you're interested, you could redo the support for `sumOf` there in a separate PR.

PR #2007, opened 2018-11-29
Consider non-existent parent `AttributeContainer`s empty

calexHG:
Given the prior faulty logic, no tree of `HierarchyAttributeContainer`s would ever report itself as empty (not even a 'tree' consisting of a single, attribute-less/empty HierarchyAttributeContainer), since every tree will have a root, which will lack a parent, and it considered the lack of existence of a parent to mean that that root container is not empty.

// good work on this project, btw :+1:


h0tk3y:
Thanks!

PR #2010, opened 2018-12-02
#KT-27670 Add quick fix: wrap expression in a lambda if compatible functional type is required

goodsauce:
Hi, first open source contribution! Hope I did everything right.

https://youtrack.jetbrains.com/issue/KT-27670

Does NOT check for nullability
```kotlin
fun foo(str: () -> String) {}
...
val str: String? = null
foo(str) // shows quickfix
```


goodsauce:
Requested changes were made,
"Reformat Code" didn't change anything

Confirmed it was the right button by applying "Reformat Code" to:
fun thisStyle()
{
}

First commit did actually check for nullability, "makeNotNullable" now allows hint to be shown regardless of nullability

4u7:
Thank you!

PR #2012, opened 2018-12-03
KT-20743: Introduce MessageEventHandler to idl definitions

Schahen:
This particular PR affects only onmessage handling


Schahen:
This will be solved more generally and slightly differently so I'm closing this one.

PR #2015, opened 2018-12-05
KT-28394 Improve code completion for top level class/interface to incorporate filename

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28394


goodwinnk:
Thank you for this contribution!

Your commit is now in master: https://github.com/JetBrains/kotlin/commit/827e04ae3e1dd436c5ba7f409e49918f027afe7c. The only addition I have done is adding a weigher test to test the order in completion list.

PR #2016, opened 2018-12-05
[JVM_IR]: Generate throwing mutating methods on immutable collections.

madsager:
This reuses the existing JVM backend code for computing and generating the methods. It only generates the direct methods and not bridges at this point.


madsager:
This adds a call from compiler/ir/backend.jvm to code in compiler/backend. Not sure if such dependencies should exist?

dnpetrov:
I think it should be ok in short-term. 
Ideally this should be done as an IR lowering pass, somewhat similar to `BridgeLowering` in `org.jetbrains.kotlin.backend.jvm.lower`, with `CollectionStubMethodGenerator#computeTasksToGenerate` logic reimplemented using IR declarations instead of descriptors.

madsager:
Thanks Dmitry, I'll be happy to look into doing this as a lowering pass. That should be a good exercise. Let me know if you think we should land this as a step in the right direction or if we should wait until we have it reimplemented as a lowering pass.

udalov:
Closing this because it has been superseded by #2296, which is now merged.

PR #2019, opened 2018-12-06
BlockingMethodInNonBlockingContextInspection: Dispatchers IO is now considered non-blocking context

NikitaKatkov:


NikitaKatkov:
@mglukhikh , it would be very nice to merge commit so that it appears in Kotlin 1.3.20

PR #2021, opened 2018-12-07
Recreated PR due to migration to idea 183

NikitaKatkov:
@mglukhikh , updated PR #2019


mglukhikh:
@NikitaKatkov Why do we need similar `DispatchersTypeCheck.kt` and `DispatchersTypeCheck.kt.182`? What did you have in mind?

mglukhikh:
@NikitaKatkov I made `DispatchersTypeCheck.kt.182` empty and then merged it.

PR #2022, opened 2018-12-10
KT-28694 "Assign backing field" quick fix adds empty line before created assignment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28694


mglukhikh:
Merged 👍

PR #2023, opened 2018-12-10
Add documentation on how merging works for toMap() for Iterables, Arrays and Sequences

BernhardPosselt:
I hope I got all occurrences where this behavior is the case. I was wondering how it would behave especially because Java's **Stream.collect(Collectors.toMap())** throws when duplicate keys are encountered


ilya-g:
Thanks for the proposal.

I'd stick with the wording used in `associate*` function docs, because `associate` is essentially `map { e -> Pair<K,V> }.toMap()` operation:

> If any of two pairs would have the same key the last one gets added to the map.

http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate.html
http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate-by.html
http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate-with.html

BernhardPosselt:
Updated

ilya-g:
Merged, thanks

PR #2025, opened 2018-12-12
KT-12273 "Replace with operator" intention is suggested for some non-operator functions and produces invalid code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12273


mglukhikh:
Merged 👍

PR #2026, opened 2018-12-12
Complete Java usage tracker if analysis extension finishes (KT-27487)

gavra0:
In cases when kapt compiler plugin runs in stubs generation mode,
Java usage tracker was not completed. This caused issues with
the incremental compilation if Java files changed.

E.g. if a constant defined in Java source file changed, stubs
would not be recompiled because the Java usage tracker would
not report changed type. This commit fixes that issue.

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/TopDownAnalyzerFacadeForJVM.kt
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/TopDownAnalyzerFacadeForJVM.kt
@@ -91,6 +91,7 @@ object TopDownAnalyzerFacadeForJVM {
         val analysisHandlerExtensions = AnalysisHandlerExtension.getInstances(project)
 
         fun invokeExtensionsOnAnalysisComplete(): AnalysisResult? {
+            container.get<JavaClassesTracker>().onCompletedAnalysis(module)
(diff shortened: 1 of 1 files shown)


yanex:
👍

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/TopDownAnalyzerFacadeForJVM.kt
```diff
@@ -99,20 +99,21 @@ object TopDownAnalyzerFacadeForJVM {
             return null
         }
 
+        fun getExtensionsResult(): AnalysisResult? {
```
dzharkov: I'd prefer moving the line `container.get<JavaClassesTracker>().onCompletedAnalysis(module)` to the beginning of `invokeExtensionsOnAnalysisComplete` since `JavaClassesTracker` is also a kind of extension and `invokeExtensionsOnAnalysisComplete` has the single call-site now

gavra0: Thanks for the comment, it does make it more readable.

dzharkov:
I've merged your commit manually.
It's expected to be released with the next eap of 1.3.20.

PR #2027, opened 2018-12-13
Update JavaSdkUtil.java

denisgaebler:
Added path for IBM JDK on platform z/OS
I tried Kotlin on z/OS and the hello world worked, but a more complex sample initially showed errors that java.lang.Object could not be resolved. So I added the -jdk-home parameter and got the error: no class roots are found in the JDK path.
So I looked into the Kotlin sources and found hardcoded paths in the JavaSdkUtil.java source. I changed this at my workstation, added "lib/s390x/default" and added it to kotlin-compiler.jar and now the compile works.
I hope you guys can add it soon.

Thanks, Denis.

### DIFF
--- a/compiler/util/src/org/jetbrains/kotlin/utils/JavaSdkUtil.java
+++ b/compiler/util/src/org/jetbrains/kotlin/utils/JavaSdkUtil.java
@@ -91,7 +91,7 @@ public class JavaSdkUtil {
       }
     }
 
-    String[] ibmJdkVmJarDirs = {"bin/default", "lib/i386/default", "lib/amd64/default"};
+    String[] ibmJdkVmJarDirs = {"bin/default", "lib/i386/default", "lib/amd64/default", "lib/s390x/default"};
     for (String relativePath : ibmJdkVmJarDirs) {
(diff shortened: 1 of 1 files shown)


denisgaebler:
One more comment on IBM JDKs, usually the path for vm.jar is lib/<hardware_architecture>/default. So maybe instead of hard coding the directorys for the IBM JDK, you might want do search all directories in the Java home. However, I don't know which IBM JDK uses bin/default.

Thanks, Denis.

udalov:
This file is actually copied [from intellij-community](https://github.com/JetBrains/intellij-community/blob/master/jps/model-impl/src/org/jetbrains/jps/model/java/impl/JavaSdkUtil.java). @Alefas Do you remember the reasons why it was copied to our repository? Maybe it'd make more sense to get rid of the copy-paste, and improve the logic in the original class instead, if that's possible.

Alefas:
Merged for now, thanks for the PR.
Actual fix will be added a bit later. I hope we would better to properly fix it by removing copy/paste rather than copypasting fix from the platform. The reason was in cutting dependencies for CLion. We need to refactor things in the platform to be able to reach this code.

PR #2029, opened 2018-12-14
JVM_IR: Do not materialize negated boolean for branches.

madsager:
Instead, flip the branch targets. This generates java byte code
such as:

   L2
    IFNE L3
    ALOAD 0
    INVOKEVIRTUAL A.getX ()F
    GOTO L4
   L3

instead of:

   L2
    IFNE L3
    ICONST_1
    GOTO L4
   L3
    ICONST_0
   L4
    IFEQ L5
    ALOAD 0
    INVOKEVIRTUAL A.getX ()F
    GOTO L6
   L5

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -643,8 +643,17 @@ class ExpressionCodegen(
         //TODO don't generate condition for else branch - java verifier fails with empty stack
         val elseBranch = branch is IrElseBranch
         if (!elseBranch) {
-            gen(condition, data)
-            BranchedValue.condJump(StackValue.onStack(condition.asmType), elseLabel, true, mv)
+            var jumpIfFalse = true
+            if (condition is IrCall && condition.symbol == classCodegen.context.irBuiltIns.booleanNotSymbol) {
+                // Instead of materializing a negated value when used for control flow, flip the branch
+                // targets instead. This significantly cuts down the amount of branches and loads of
+                // const_0 and const_1 in the generated java bytecode.
+                gen(condition.getValueArgument(0)!!, data)
+                jumpIfFalse = false
+            } else {
+                gen(condition, data)
+            }
+            BranchedValue.condJump(StackValue.onStack(condition.asmType), elseLabel, jumpIfFalse, mv)
         }
 
         val end = Label()


max-kammerer:
Suggested solution wouldn't work now as is, cause there is some ambiguites in 'not' operator representation:
1. From one side it's defined in IR symbols:
```
fun defineOperator(name: String, returnType: KotlinType, valueParameterTypes: List<KotlinType>): IrSimpleFunction {
        val operatorDescriptor = IrSimpleBuiltinOperatorDescriptorImpl(packageFragment, Name.identifier(name), returnType)
        for ((i, valueParameterType) in valueParameterTypes.withIndex()) {
            operatorDescriptor.addValueParameter(
                IrBuiltinValueParameterDescriptorImpl(operatorDescriptor, Name.identifier("arg$i"), i, valueParameterType)
            )
        }
        return addStubToPackageFragment(operatorDescriptor)
    }
```
2. but from another side, there is bug in psi2ir that creates new 'not'  symbol from deserialized descriptor regards on symbol in 1: 
```
            val functionSymbol = context.symbolTable.referenceFunction(functionDescriptor.original)
           val superQualifierSymbol = call.superQualifier?.let { context.symbolTable.referenceClass(it) }
           val irCall = IrCallImpl(
               startOffset, endOffset,
               returnType.toIrType(),
               functionSymbol,
               functionDescriptor,
               origin,
               superQualifierSymbol
           ).apply {
               ....
           }
```
3. 2 symbol expects dispatch receiver but 1 expects value parameters (also bug)

So working solution should looks like:
```
if (condition is IrCall && (
                        condition.symbol == classCodegen.context.irBuiltIns.booleanNotSymbol ||
                                classCodegen.state.intrinsics.getIntrinsic(condition.symbol.descriptor) is Not
                        )
            ) {
                gen(condition.dispatchReceiver ?: condition.getValueArgument(0)!!, data)
                jumpIfFalse = false
            }
```

NB: New behavior is untested cause  'bytecodeText' test is not yet added to IR-test framework

madsager:
Quick update: there is one test failure with the new code. I'll dig into that and report back when I know what is going on.

max-kammerer:
Yes, i meant bytecode tests. We have such kind test for old backend in 'bytecodeText', to support IR we need to make AbstractIrBytecodeTextTest extending AbstractBytecodeTextTest by AbstractIrBlackBoxInlineCodegenTest analogue. 

For first execution switch KotlinTestUtils.AUTOMATICALLY_MUTE_FAILED_TESTS for test muting

madsager:
Thanks! Removing the intrinsic method call removed the conversion of the argument to the right asmType which is done as part of the invocation of the intrinsic. That was causing the errors. That has been fixed and the current code passes all tests. I'll get started on bytecodeText tests next. I will leave it up to you whether to land this change as is or wait for the testing changes.

max-kammerer:
Let's wait for tests)

madsager:
Rebased on top of master with the bytecode text tests enabled. There is one very local test that tests for exactly this patter which now passes. Next steps: make more of these branching tests generate reasonable java bytecode so that more of these tests pass.

Mike, can you take another look?

Thanks!

PR #2030, opened 2018-12-14
Debugger improvements

yanex:
- Support new local variable convention for receivers (`$receiver` -> `$this$<label>` or `$receiver` if the label is unknown);
- Support Kotlin/JVM variable modes in Watches tool window;
- Async stack trace support for coroutines.


goodwinnk:
The was an internal review, so this PR doesn't show the actual status anymore.

PR #2031, opened 2018-12-15
Error on calling suspend functions inside Lock.withLock body

NikitaKatkov:
See SuspensionPointInsideCriticalSectionChecker

### DIFF
--- a/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/diagnostics/DefaultErrorMessagesJvm.java
+++ b/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/diagnostics/DefaultErrorMessagesJvm.java
@@ -137,7 +137,7 @@ public class DefaultErrorMessagesJvm implements DefaultErrorMessages.Extension {
         MAP.put(NON_JVM_DEFAULT_OVERRIDES_JAVA_DEFAULT, "Non-@JvmDefault interface method cannot override default Java method. Please annotate this method with @JvmDefault");
         MAP.put(EXPLICIT_METADATA_IS_DISALLOWED, "Explicit @Metadata is disallowed");
         MAP.put(SUSPENSION_POINT_INSIDE_MONITOR, "A suspension point at {0} is inside a critical section", STRING);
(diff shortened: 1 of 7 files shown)


compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/diagnostics/DefaultErrorMessagesJvm.java
```diff
@@ -137,7 +137,7 @@
         MAP.put(NON_JVM_DEFAULT_OVERRIDES_JAVA_DEFAULT, "Non-@JvmDefault interface method cannot override default Java method. Please annotate this method with @JvmDefault");
         MAP.put(EXPLICIT_METADATA_IS_DISALLOWED, "Explicit @Metadata is disallowed");
         MAP.put(SUSPENSION_POINT_INSIDE_MONITOR, "A suspension point at {0} is inside a critical section", STRING);
-        MAP.put(SUSPENSION_POINT_INSIDE_SYNCHRONIZED, "The ''{0}'' suspension point is inside a synchronized block", NAME);
+        MAP.put(SUSPENSION_POINT_INSIDE_MUTEX_LOCK, "The ''{0}'' suspension point inside a mutex lock can lead to runtime exceptions", NAME);
```
ilmirus: The wording here suggest that mutex is being held in `synchronized` block, which is not the case.
It should be either more general `... suspension point is inside a critical section` or just two different error messages.

Additionally, `can lead to runtime exceptions` part shall be omitted, since this is an error, not a warning. The code does not compile and thus cannot lead to runtime exceptions.

The rest is OK.

NikitaKatkov: @ilmirus , so I'll change the error message, but leave map key name as it is now, right?
Naming is one of the worst things in programming -_-

ilmirus: Well, ideally, name should be kind of similar to the message.

NikitaKatkov: replaced error name and description with more general wording

PR #2032, opened 2018-12-16
KT-17119 Inspection for (Scala-like) `= { ... }` syntax without expected type in function definition

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17119


mglukhikh:
I have a vague feeling that somebody already implemented something like this in the past... Will try to find.

mglukhikh:
Got it, it was `MoveIntoParenthesesIntention`

mglukhikh:
Merged 👍

PR #2037, opened 2018-12-19
Add ConvertToAlsoIntention to plugin-common.xml

t-kameyama:
It was deleted by 0439536.
https://youtrack.jetbrains.com/issue/KT-28699


mglukhikh:
@t-kameyama Thank you, you have noticed an important mistake.

PR #2038, opened 2018-12-20
KT-28803 Inspection: result of enum entries comparison is always false / true

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28803

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.idea.intentions.branchedTransformations.unwrapBlockO
 import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.containingClass
 import org.jetbrains.kotlin.psi.psiUtil.getChildrenOfType
 import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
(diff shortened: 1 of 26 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -116,12 +119,47 @@ class ConstantConditionIfInspection : AbstractKotlinInspection() {
...
+    }
     val type = getType(context) ?: return null
-
     val constantValue = ConstantExpressionEvaluator.getConstant(this, context)?.toConstantValue(type)
     return constantValue?.value as? Boolean
 }
 
+private fun KtExpression.enumEntriesComparison(): Boolean? {
+    if (this !is KtBinaryExpression) return null
+    val left = this.left as? KtDotQualifiedExpression ?: return null
```
mglukhikh: Could we also handle the case when enum entry is imported? E.g.
```
import test.A.* // enum A { X, Y }

fun foo() {
    if (X == Y) { ... }
}
```

idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -116,12 +119,47 @@ class ConstantConditionIfInspection : AbstractKotlinInspection() {
...
+    val right = this.right as? KtDotQualifiedExpression ?: return null
+
+    val leftEnum = left.enumClass() ?: return null
+    val rightEnum = right.enumClass() ?: return null
+    if (leftEnum != rightEnum) return null
```
mglukhikh: It's correct, because for entries of different enum classes we have the compiler warning `INCOMPATIBLE_ENUM_COMPARISON`

mglukhikh:
Merged manually, thank you

PR #2039, opened 2018-12-21
JVM_IR: Enable bytecode text tests

madsager:


compiler/testData/codegen/bytecodeText/lineNumbers/when.kt
```diff
@@ -5,4 +5,6 @@ fun foo() {
         else -> System.out?.println()
     }
 }
-// 1 LINENUMBER 3
\ No newline at end of file
+// 1 LINENUMBER 3
+// Adding ignore flags below the test since the test relies on line numbers.
+// IGNORE_BACKEND: JVM_IR
```
madsager: This is the only special handling. As this test relies on line numbers the suppression line cannot be above the code.

PR #2040, opened 2018-12-25
KT-28592 False positive "Remove redundant backticks" for underscore variable name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28592


mglukhikh:
Merged 👍

PR #2041, opened 2018-12-25
Add an integration test around generated source sets

akhaku:
Opening this PR up sort of as a heads-up and hopefully a bit of discussion - I wrote a gradle plugin for my employer that started failing with the kotlin gradle plugin and Gradle 5+, and I was curious so started digging through the kotlin-gradle-plugin source. A little curious as to why it is set up the way it is, i.e. create kotlin tasks for all the java sourcesets and then running them before the java compile step for all the sourcesets. This PR has an example of the sorts of issues that can cause - generally speaking the problem seems to be that we're inserting ourselves before java compiles without caring what the java compile itself depends on (eg generate sources steps). Gradle<5 masked the issue by not failing hard, a recent change just surfaced it.


StefMa:
It's only a test, but I would recommend to use the new lazy methods 'tasks.register' and 'tasks.named' instead of the "old" 'tasks.create' and 'tasks.getByName' methods...

akhaku:
Absolutely, thanks for the tip!

PR #2042, opened 2018-12-26
KT-28698 "Convert to apply" intention: include function calls with `this` passed as an argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28698


mglukhikh:
Merged manually (with minor corrections). Thank you 👍

PR #2043, opened 2018-12-26
Minor: cleanup grammar for callSuffix and typeReference rules

vlsi:
This makes http://kotlinlang.org/docs/reference/grammar.html closer to the actual grammar.


h0tk3y:
@vlsi Sorry, we have updated the Grammar page independently and it is now close to the actual Kotlin parser. This PR now has a conflict, so I'm closing it.

PR #2044, opened 2018-12-26
Implements `KAnnotatedElement.hasAnnotation()`

Kerooker:
This commit fixes KT-29041, by adding a convenience method to verify if a certain `KAnnotatedElement` has an annotation or not.

I'm not sure how to add tests to this, as there are no tests for KAnnotatedElement nor there are tests for KAnnotatedElements file

### DIFF
--- a/compiler/testData/codegen/box/reflection/annotations/hasAnnotation.kt
+++ b/compiler/testData/codegen/box/reflection/annotations/hasAnnotation.kt
@@ -1,0 +1,25 @@
+// IGNORE_BACKEND: JS_IR
+// IGNORE_BACKEND: JS, NATIVE
+// WITH_REFLECT
+
+import kotlin.reflect.full.hasAnnotation
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+annotation class Baz
+annotation class Far
+
+@Baz
+@Far
+class Foo
+
+class Bar
+
+fun box(): String {
+    assertFalse(Bar::class.hasAnnotation<Baz>)
+    assertFalse(Bar::class.hasAnnotation<Far>)
+
+    assertTrue(Foo::class.hasAnnotation<Baz>)
+    assertTrue(Foo::class.hasAnnotation<Far>)
+    return "OK"
+}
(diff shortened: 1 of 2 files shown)


udalov:
Hello and thanks for your pull request! Since Kotlin 1.3 is already released and we cannot add new API to kotlin-reflect in an incremental release (see [Stability of Different Components](https://kotlinlang.org/docs/reference/evolution/components-stability.html)), this can only be released starting from 1.4. Therefore please adjust the `@SinceKotlin` annotation value to `"1.4"`.

> I'm not sure how to add tests to this, as there are no tests for KAnnotatedElement nor there are tests for KAnnotatedElements file

Please add a test on this to a new file near [`compiler/testData/codegen/box/reflection/annotations/findAnnotation.kt`](https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/reflection/annotations/findAnnotation.kt), regenerate tests by `./gradlew generateTests` (or "Generate Compiler Tests" IntelliJ Run configuration) and check that all newly added tests pass. Thank you!

Kerooker:
Should be OK now. For some reason this pull request isn't updating with my commit https://github.com/Kerooker/kotlin/commit/07dc077bb5d73e797c27e7bab1feea83f77d744b.

Kerooker:
Should be OK now

udalov:
@Kerooker Sorry for the delay! No, the change looks OK now (except for the compilation error in the test, which I'll fix). The problem is that the 1.4 release is so far away that we don't even have the necessary branch in place, and we can't merge this change to master since it's a new API (even though it's `@SinceKotlin("1.4")`, it's still a change in the ABI and can be observed if called from Java for example).

I'll try creating a separate branch soon where stable language/API versions are increased to 1.4, and at that point it'll be possible to merge your changes. Thanks once again!

udalov:
Instead of creating a separate branch for 1.4 as we did in previous releases, we've decided to release most of the new library API as [experimental](https://github.com/Kotlin/KEEP/issues/95) in master, under [`@ExperimentalStdlibApi`](https://github.com/JetBrains/kotlin/blob/3a33f68fc9e05bbd520a6bc9eb63ccbef0558fd6/libraries/stdlib/src/kotlin/annotations/ExperimentalStdlibApi.kt). I've annotated `hasAnnotation` with this marker and merged it to master. Therefore `hasAnnotation` will be available since 1.3.40, but to use it, one must either annotate the call site with `@UseExperimental(ExperimentalStdlibApi::class)` or provide the compiler flag `-Xuse-experimental=kotlin.ExperimentalStdlibApi`.

PR #2045, opened 2018-12-28
KT-28773 Kotlin/JS: Wrong inspection to replace .equals() with == on dynamic values

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28773


mglukhikh:
Merged 👍

PR #2046, opened 2018-12-30
KT-28641 "Remove useless cast" produces a dangling lambda ("Too many arguments" error)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28641


mglukhikh:
Merged 👍

PR #2047, opened 2019-01-02
KT-27143 Intention "Replace camel-case name with spaces" is suggested for snake_case names in test functions and renames them incorrectly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27143


mglukhikh:
Merged 👍

PR #2049, opened 2019-01-03
JVM IR: simplify chains of negations in if conditions.

madsager:
Introduce lowering phase that turns !!exp -> exp for the boolean 'not' builtin. This makes sure that code such as

```
if (!!!!!booleanValue) {
    doStuff()
}
```

generates only one branch.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLoweringPhases.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLoweringPhases.kt
@@ -211,11 +211,16 @@ private val TailrecPhase = makeJvmPhase(
     description = "Handle tailrec calls"
 )
 
-
 private val ToArrayPhase = makeJvmPhase(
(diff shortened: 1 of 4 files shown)


max-kammerer:
Discussed to make separate lowering

madsager:
Thanks Mike, I updated the pull request to do this as a simple lowering. Is something like this what you had in mind?

max-kammerer:
@madsager Yeah, i meant exactly this. Will look more precisly on monday

max-kammerer:
Seems 'isNegation' checking could be shared with ExpressionCodegen: you have done similar logic in previous pr

madsager:
Thanks! Shared 'isNegation' and 'negationArgument' between ExpressionCodegen and this new lowering as top-level functions in ExpressionCodegen.

max-kammerer:
@madsager It's best to move them to new NegatedExpressionLowering companions, as at some point it would be common.

madsager:
@max-kammerer agreed and done! :)

max-kammerer:
Thanks!

PR #2050, opened 2019-01-05
KT-28969 TYPE_MISMATCH in array vs non-array case: two quick fixes exist for annotation and none of them adds array literal

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28969


mglukhikh:
Merged 👍

PR #2051, opened 2019-01-06
KT-20357: Add a sample for `isNotBlank`

burakeregar:


ilya-g:
Thanks for the effort, but now we already have samples covering the functions `isEmpty`, `isBlank`, etc:  6bc7c060386

PR #2052, opened 2019-01-06
KT-27869: Convert Java to Kotlin dialog ask yes/no question but shows OK/Cancel buttons

burakeregar:
https://youtrack.jetbrains.com/issue/KT-27869

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/actions/JavaToKotlinAction.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/actions/JavaToKotlinAction.kt
@@ -131,7 +131,7 @@ class JavaToKotlinAction : AnAction() {
 
             if (enableExternalCodeProcessing && converterResult!!.externalCodeProcessing != null) {
                 val question = "Some code in the rest of your project may require corrections after performing this conversion. Do you want to find such code and correct it too?"
-                if (!askExternalCodeProcessing || (Messages.showOkCancelDialog(project, question, title, Messages.getQuestionIcon()) == Messages.OK)) {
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/actions/JavaToKotlinAction.kt
```diff
@@ -131,7 +131,7 @@ class JavaToKotlinAction : AnAction() {
 
             if (enableExternalCodeProcessing && converterResult!!.externalCodeProcessing != null) {
                 val question = "Some code in the rest of your project may require corrections after performing this conversion. Do you want to find such code and correct it too?"
-                if (!askExternalCodeProcessing || (Messages.showOkCancelDialog(project, question, title, Messages.getQuestionIcon()) == Messages.OK)) {
+                if (!askExternalCodeProcessing || (Messages.showYesNoDialog(project, question, title, Messages.getQuestionIcon()) == Messages.OK)) {
```
mglukhikh: Here and below, I suppose you should compare the result of `showYesNoDialog` with `Messages.YES`, not with `Messages.OK`. According to `showYesNoDialog` contract, it can return `YES`, `NO` or `CANCEL`, and never returns `OK`.

burakeregar: You are right, I missed that. Fixed and pushed. Thanks @mglukhikh

mglukhikh:
@semoro IMO now it can be accepted. The conflict arose because we've recently deleted AS31 branch.

semoro:
Thank you for your contribution! 🙂

PR #2054, opened 2019-01-08
Allow specifying isolation mode for KAPT workers

ting-yuan:
There are a few isolation modes for Gradle workers. KAPT uses 2 of them.

IsolationMode.NONE is the default. It runs workers in Gradle daemon and
is more memory efficient.

IsolationMode.PROCESS forks workers into individual processes. This
keeps some resource leaking annotation processors from affecting the
rest of compilation.

Users can specify kapt.workers.isolation=process, if they need to use
some annotation processor that are known to be resource leaking.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptWithoutKotlincTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptWithoutKotlincTask.kt
@@ -82,7 +82,12 @@ open class KaptWithoutKotlincTask @Inject constructor(private val workerExecutor
         val kaptClasspath = kaptJars + findKotlinStdlibClasspath(project)
 
         workerExecutor.submit(KaptExecution::class.java) { config ->
-            config.isolationMode = IsolationMode.PROCESS
+            val isolationModeStr = project.findProperty("kapt.workers.isolation") as String? ?: "none"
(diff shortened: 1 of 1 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptWithoutKotlincTask.kt
```diff
@@ -82,7 +82,11 @@ open class KaptWithoutKotlincTask @Inject constructor(private val workerExecutor
         val kaptClasspath = kaptJars + findKotlinStdlibClasspath(project)
 
         workerExecutor.submit(KaptExecution::class.java) { config ->
-            config.isolationMode = IsolationMode.PROCESS
```
yanex: Not critical, but:
Maybe use case-insensitive comparison instead of hard-coding several possible combinations?

ting-yuan: Done. Thanks for the review!

yanex:
Merged manually.
1.3.20: https://github.com/JetBrains/kotlin/commit/12e03c34c66baf0712c196bbde35485c1e8e2dd6
master: https://github.com/JetBrains/kotlin/commit/72e351a0e3610051fe4222dca4e1eeedf7ae45da

Thanks for the contribution!

PR #2055, opened 2019-01-09
"Add parameter to constructor": fix for enum entry

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28995
https://youtrack.jetbrains.com/issue/KT-29051
https://youtrack.jetbrains.com/issue/KT-29052

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/refactoring/changeSignature/usages/KotlinFunctionCallUsage.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/refactoring/changeSignature/usages/KotlinFunctionCallUsage.kt
@@ -34,14 +34,12 @@ import org.jetbrains.kotlin.idea.refactoring.introduce.extractionEngine.createNa
 import org.jetbrains.kotlin.idea.refactoring.introduce.introduceVariable.KotlinIntroduceVariableHandler
 import org.jetbrains.kotlin.idea.refactoring.replaceListPsiAndKeepDelimiters
 import org.jetbrains.kotlin.idea.core.ShortenReferences
+import org.jetbrains.kotlin.idea.search.usagesSearch.constructor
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/refactoring/changeSignature/usages/KotlinFunctionCallUsage.kt
```diff
@@ -101,7 +99,10 @@ class KotlinFunctionCallUsage(
         if (resolvedCall == null || resolvedCall.isReallySuccess()) return false
 
         // TODO: investigate why arguments are not recorded for enum constructor call
-        if (element is KtSuperTypeCallEntry && element.parent.parent is KtEnumEntry) return false
+        val enumEntry = element.parent.parent as? KtEnumEntry
+        if (element is KtSuperTypeCallEntry && enumEntry != null) {
+            if (element.valueArguments.isEmpty() && enumEntry.containingClass()?.constructor?.valueParameters?.size == 1) return false
```
mglukhikh: I suppose this check covers only one particular case. See also test comment below.

idea/testData/quickfix/changeSignature/addEnumConstructorParameter3.kt
```diff
@@ -0,0 +1,8 @@
+// "Add parameter to constructor 'Foo'" "true"
+// DISABLE-ERRORS
+enum class Foo(n: Int) {
+    A(1, 2<caret>),
+    B(3),
+    C(),
```
mglukhikh: What if we write `C(3, 4)` here? Suppose this `C(3, 4)` will be changed to `C(3, 2)`, which is not correct.

mglukhikh:
I will test it a bit and merge.

mglukhikh:
Merged manually 👍

PR #2056, opened 2019-01-10
KT-28851 'Convert parameter to receiver' adds `Array<out T>` wrapper to `vararg` parameter and drops `override` modifier in implementations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28851


mglukhikh:
Merged 👍

PR #2057, opened 2019-01-11
Make script engine work out of the box

rjolly:
How to run:
```
jrunscript -classpath kotlin-compiler-1.3.11.jar;kotlin-daemon-client-1.3.11.jar;kotlin-reflect-1.3.11.jar;kotlin-script-runtime-1.3.11.jar;kotlin-script-util-1.3.11.jar;kotlin-scripting-common-1.3.11.jar;kotlin-scripting-jvm-1.3.11.jar;kotlin-stdlib-1.3.11.jar;kotlin-stdlib-common-1.3.11.jar -l kotlin
kotlin> "hello world"
hello world
kotlin>
```

### DIFF
--- a/libraries/tools/kotlin-script-util/build.gradle.kts
+++ b/libraries/tools/kotlin-script-util/build.gradle.kts
@@ -15,7 +15,7 @@ dependencies {
     compile(project(":kotlin-script-runtime"))
     compile(project(":kotlin-scripting-jvm"))
     compile(commonDep("org.jetbrains.intellij.deps", "trove4j"))
-    compileOnly(project(":compiler:cli"))
+    compile(project(":compiler:cli"))
     compileOnly(project(":compiler:daemon-common"))
     compile(projectRuntimeJar(":kotlin-daemon-client"))
     compileOnly("com.jcabi:jcabi-aether:1.0-SNAPSHOT")
--- a/libraries/tools/kotlin-script-util/src/main/resources/META-INF/services/javax.script.ScriptEngineFactory
+++ b/libraries/tools/kotlin-script-util/src/main/resources/META-INF/services/javax.script.ScriptEngineFactory
@@ -1,0 +1,1 @@
+org.jetbrains.kotlin.script.jsr223.KotlinJsr223JvmLocalScriptEngineFactory


ligee:
Very sorry for the delay, I forgot about this one.
The runtime dependencies to the compiler subprojects are not allowed for the published jars. 
And I need more info to understand the reasons behind this PR.
And this will be probably obsolete with 1.4.

rjolly:
Reasons behind this PR : to be able to use Kotlin as a JSR-223 scripting engine. For instance in this software : https://meditorworld.appspot.com/ 
Runtime dependencies to subprojects not allowed : in that case, would it be possible to integrate the script engine in the main jar ? That would be very useful indeed.

ligee:
What is wrong with the existing JSR-223 support? There are at least two engines that are now officially distributed with Kotlin, `kotlin-scripting-jsr223` for the plain scripts, and `kotlin-main-kts` for the more advanced `.main.kts` ones.
Here are the tests/examples of using the first one - https://github.com/JetBrains/kotlin/blob/1.3.70/libraries/scripting/jsr223-test/test/kotlin/script/experimental/jsr223/test/KotlinJsr223ScriptEngineIT.kt#L60
and here is with `main-kts` - https://github.com/JetBrains/kotlin/blob/1.3.70/libraries/tools/kotlin-main-kts-test/test/org/jetbrains/kotlin/mainKts/test/mainKtsJsr223Test.kt#L39

Initially `script-util` was meant to be a library for constructing e.g. your own JSR-223 engines, not a ready-to-use JSR-223 engine, therefore things like the ScriptEngineFactory service configuration was omitted on purpose. Since then the "official" engines were introduced, and `script-util` itself will be obsolete soon.

rjolly:
Ilya, kotlin-scripting-jsr223 works for me, thanks ! One little itch though : it (transitively) depends on org.jetbrains.intellij.deps#trove4j;1.0.20181211 , which conflicts with net.sf.trove4j#trove4j;3.0.3 , a dependency of cc.redberry#rings;2.5.7 , which is a computer algebra library that I am considering developing a Kotlin DSL for. Would it be a problem to upgrade ? In any case, it is great that a JSR-223 engine exists. Again, thank you.

PR #2058, opened 2019-01-14
JVM_IR: Optimize null checks.

madsager:
Introduce lowering to remove null checks for primitive types and
replace them when constant true or false.

Generate ifnull/ifnonnull instructions for null checks instead
of materializing a null literal for an equality check.


compiler/testData/codegen/bytecodeText/nullCheckOptimization/localLateinit/initialized.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
madsager: This test is now failing because it contain an IFNULL instruction. However, it only passed before because the IFNULL was an IFEQ comparing to the null constant.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -653,8 +652,23 @@ class ExpressionCodegen(
                 condition = negationArgument(condition as IrCall)
                 jumpIfFalse = false
             }
-            gen(condition, data).put(condition.asmType, mv)
-            BranchedValue.condJump(StackValue.onStack(condition.asmType), elseLabel, jumpIfFalse, mv)
+            if (isNullCheck(condition)) {
```
max-kammerer: Maybe extract whole if (!elseBranch)  branch body to separate method: 'genConditionWithOptimizationsIfPossible()'

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinCallLowering.kt
```diff
@@ -11,36 +11,64 @@ import org.jetbrains.kotlin.codegen.intrinsics.Not
...
-        fun negationArgument(call: IrCall) : IrExpression {
+        fun negationArgument(call: IrCall): IrExpression {
             // TODO: there should be only one representation of the 'not' operator.
             // Once there is only the IR definition the negation argument will
             // always be a value argument and this method can be removed.
             return call.dispatchReceiver ?: call.getValueArgument(0)!!
         }
     }
 
+    private fun isNullCheckOfPrimitiveTypeValue(call: IrCall, context: JvmBackendContext): Boolean {
```
max-kammerer: Checked expression could have side effects:
calcPrimitiveWithSideEffects() != null

madsager: Thanks for catching that Mike! I have updated the code to only allow the elimination of null checks for primitive type constants and for primitive type values loaded from a local variable. It is OK to eliminate the local variable load of a simple type value - even if it is changed during debugging it still cannot be updated to a null value. What do you think?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
madsager: Done, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinCallLowering.kt
madsager: As we discussed offline, in order to get rid of all null comparisons with primitive type arguments, we now transform the potentially side-effecting cases:

if (side-effecting-expr-with-primitive-type == null) ...

--->

side-effecting-expr-with-primitive-type
pop
if (false) ...

compiler/testData/codegen/bytecodeText/conditions/nullCompareSideEffectingExpression.kt
```diff
@@ -0,0 +1,16 @@
+fun f() : Int {
```
max-kammerer: pls, don't use println in test
it would be best to rewrite it via 
```var result = "fail"

fun f() : Int {
    result = "OK"
    return 42
}

fun box(): String {
    if (f() == null) {
        return "fail"
    }
    result = "fail"

    if (f() != null) {
        return result
    }

    return "fail"
}
```
And add same test to box ones.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinCallLowering.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.jetbrains.kotlin.ir.expressions.impl.IrBlockImpl
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.types.isPrimitiveType
+import org.jetbrains.kotlin.ir.types.toKotlinType
+import org.jetbrains.kotlin.ir.util.coerceToUnitIfNeeded
+import org.jetbrains.kotlin.ir.util.isNullConst
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+
+class JvmBuiltinCallLowering(val context: JvmBackendContext) : FileLoweringPass {
```
max-kammerer: pls, rename class to JvmBuiltinOptimizationLowering

max-kammerer:
Thanks Mads! Merged manually

PR #2061, opened 2019-01-14
expose AbstractMutableList.modCount in common

tvierling:
K/Native didn't have this before, but it does now (looking at the current source), so it is available everywhere. Access to this var is needed for implementing more complex subclasses in a modification-safe manner.


tvierling:
Never mind, going to have to submit separate project pull requests with "actual" declarations, closing this for now.

PR #2062, opened 2019-01-16
KT-18715: don't swallow comments for if->elvis

fg123:
https://youtrack.jetbrains.com/issue/KT-18715

Adds a CommentSaver to the IfThenToElvis intention that will save and
restore comments when converting an if expression to elvis expression.

As a side note, when I'm editing `IfThenToElvisIntention.kt`, I am unable to Ctrl-Click into other classes like, like if I ctrl-click `CommentSaver` IDEA doesn't do anything. I also have no auto-imports and it doesn't detect errors (like spelling or type checking) until I compile. Is there a way to fix this? This is on a fresh IDEA installation.


mglukhikh:
@fg123 I think it's good enough, thank you. If we have a comment saver, why don't use it here? So I'll merge it.

mglukhikh:
@fg123 Answering your side note: you should perform Gradle import to fix your problems.

mglukhikh:
@fg123 For your future PRs, could you please run tests at least for the intention you are trying to fix? Your commit broke 17 tests, 15 of them are directly for `IfThenToElvisIntention`.

fg123:
@mglukhikh Very sorry, I think I made the line-break change after I ran the tests. I will take another look into fixing it; I see you already fixed some of the tests? How do I view the build on TeamCity that ran from this PR?

fg123:
@mglukhikh Very sorry, I will be more careful with that next time. I see you already pushed a fix for the newline issue; why is it that running the intention in my `runIde` editor produces different result from the test?

mglukhikh:
I think style settings in your IDE differs from default.

PR #2063, opened 2019-01-16
KT-18715-2: fix broken test

fg123:
- use default value of `false` for `CommentSaver`
- fixed whitespace issues on added test


mglukhikh:
@fg123 At this moment it's easier for me to fix test myself (because I already fixed part of the problem). However, thank you for the PR.

PR #2065, opened 2019-01-16
KT-29143: add OmitConstructorKeyword intention

fg123:
https://youtrack.jetbrains.com/issue/KT-29143

Add intention to remove redundant `constructor` keyword from a primary constructor, with associated tests.

This can only be applied if the constructor doesn't have any visibility modifiers or annotations:
```kotlin
annotation class Ann
class AnnotatedParam @Ann constructor(x: Int) // won't work
class Foo private constructor(x: Int) // won't work
```


idea/resources/intentionDescriptions/OmitConstructorKeywordIntention/after.kt.template
```diff
@@ -0,0 +1 @@
+class Foo (x: Int, y: Int)
```
goodwinnk: Can you please try to remove the space between class name and `(`, otherwise result is not properly formatted.

goodwinnk:
Also I recommend to replace `Omit` -> `Replace` as it's going to be more consistent with other intentions.

idea/resources/intentionDescriptions/OmitConstructorKeywordIntention/after.kt.template
goodwinnk: >  if you do `Make public`, it becomes `class Foo (x: Int)`
Oh...

You may choose to fix the issue only for removing constructor intention, as other intentions lay out of scope for this PR. You may report found bug with `Make public` to tracker and decide if you want to fix it too later.

On the other hand if you see some general solution for both intentions I don't mind if it would be a separate commit fixing them at once in this PR.

fg123: Looks like the other intention's issue is in `KtPrimaryConstructor.addModifier` and `KtPrimaryConstructor.removeModifier` where they remove the constructor keyword. I added a function instead that will remove keyword and space (and I call that in my intention), but I have not changed the behavior in the other intention because that would require updating a lot of other tests. I will file an bug to the tracker in the meantime.

goodwinnk:
I have merged commits after squashing and renaming. This is the result commit: https://github.com/JetBrains/kotlin/commit/185d0c61659c8c2855c27e102237e8cbb67cd7a9.

PR #2066, opened 2019-01-17
Skip annotation processing when there's no AP specified

ting-yuan:
This also fixes KT-29302 where stub generation is skipped while the
subsequent annotation processing phase anticipates the generated stubs.

This patch addresses cases where kapt.use.worker.api=true. The cases
where kapt.use.worker.api=false are handled in the same place of stub
generation.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kapt3IT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kapt3IT.kt
@@ -56,6 +56,17 @@ class Kapt3WorkersIT : Kapt3IT() {
             assertSubstringCount("Loaded com.sun.tools.javac.util.Context from", 1)
         }
     }
+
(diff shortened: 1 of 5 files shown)


yanex:
I suppose this solution may be incomplete, as kapt doesn't require `annotationProcessorFqNames` to be declared explicitly. If the user provides no AP names, kapt uses a `ServiceLoader` to load processors. So it's better to check both `annotationProcessorFqNames` and `kaptClasspath`.
If I am right, some of the Gradle integration tests should fail after your change. Did you run them (`./gradlew :kotlin-gradle-plugin-integration-tests:test`)? Also, I would add a new test for this (or reflect the change in the existing ones).

ting-yuan:
Thanks for the pointers! kaptClasspath is added in the check, as well as a new test which would fail if annotation processing isn't skipped correctly.

The updated PR is tested locally with the command below in Linux. There are 6 tests from 2 classes failed but none of them seem related. They also failed without this patch.

$ ./gradlew -Pteamcity=true :kotlin-gradle-plugin-integration-tests:test

ting-yuan:
Looks like this was already merged. Closing manually.

PR #2067, opened 2019-01-17
KT-29238 Non-canonical modifiers order inspection incorrectly includes annotations into range

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29238


idea/src/org/jetbrains/kotlin/idea/inspections/SortModifiersInspection.kt
```diff
@@ -47,7 +47,7 @@ class SortModifiersInspection : AbstractKotlinInspection(), CleanupLocalInspecti
 
                 val descriptor = holder.manager.createProblemDescriptor(
                     startElement,
-                    list,
+                    modifierElements.last(),
```
mglukhikh: I think it would be better to highlight all modifiers, not only the last one (but annotations should not be included)

mglukhikh:
Merged 👍

PR #2069, opened 2019-01-21
JVM_IR: Generate better code for null checks.

madsager:
Simplify ifs when branches have condition true/false.

Simplify blocks containing only a variable declaration
and a variable get of the same variable. Simplify to
just the condition.

Do not introduce temporary variables for constants for
null checks. Constants have no side-effects and can be
reloaded freely instead of going through a local.

This simplifies code such as "42.toLong()!!" so that the
resulting code has no branches and uses no locals. The
simplifications happen as follows:

```
block
  temp = 42.toLong()
  when
    (temp == null) throw NPE
    (true) load temp

---> null test simplification

block
  temp = 42.toLong()
  when
    (false) throw NPE
    (true) load temp

---> when simplification

block
  temp = 42.toLong()
  load temp

---> block simplification

42.toLong()
```

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
@@ -8,17 +8,19 @@ package org.jetbrains.kotlin.backend.jvm.lower
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.codegen.intrinsics.Not
+import org.jetbrains.kotlin.ir.IrStatement
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.declarations.IrFile
-import org.jetbrains.kotlin.ir.expressions.IrCall
-import org.jetbrains.kotlin.ir.expressions.IrConst
-import org.jetbrains.kotlin.ir.expressions.IrExpression
-import org.jetbrains.kotlin.ir.expressions.IrGetValue
+import org.jetbrains.kotlin.ir.declarations.IrVariable
+import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrBlockImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
 import org.jetbrains.kotlin.ir.types.isPrimitiveType
 import org.jetbrains.kotlin.ir.types.toKotlinType
 import org.jetbrains.kotlin.ir.util.coerceToUnitIfNeeded
+import org.jetbrains.kotlin.ir.util.isFalseConst
 import org.jetbrains.kotlin.ir.util.isNullConst
+import org.jetbrains.kotlin.ir.util.isTrueConst
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 
@@ -57,11 +59,12 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
         return false
     }
 
-    private fun isNullCheckOfNullConstant(call: IrCall, context: JvmBackendContext): Boolean {
+    private fun isNullCheckOfConstant(call: IrCall, context: JvmBackendContext): Boolean {
         if (call.symbol == context.irBuiltIns.eqeqSymbol) {
             val left = call.getValueArgument(0)!!
             val right = call.getValueArgument(1)!!
-            return right.isNullConst() && left.isNullConst()
+            return (right.isNullConst() && left is IrConst<*>)
+                    || (left.isNullConst() && right is IrConst<*>)
         }
         return false
     }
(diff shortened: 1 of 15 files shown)


compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/OperatorExpressionGenerator.kt
```diff
@@ -147,20 +148,33 @@ class OperatorExpressionGenerator(statementGenerator: StatementGenerator) : Stat
         }
     }
 
+    private fun IrStatementsBuilder<*>.generateIrIfNull(
```
madsager: I'm unsure if this has consequences for non-jvm backends? On the JVM, introducing a temporary for constants makes little sense as the constant can just be reloaded. Is that a reasonable assumption for the other backends as well?

madsager:
As discussed offline, I updated this to avoid the change to Psi2Ir translation. Instead I have integrated the removal of immutable IrTemporary variables in the lowering phase. What do you think?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -93,6 +92,69 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+            override fun visitWhen(expression: IrWhen): IrExpression {
+                expression.transformChildrenVoid(this)
+                // Remove all branches with constant false condition.
+                expression.branches.removeIf() {
+                    it.condition.isFalseConst()
+                }
+                // If the only condition that is left has a constant true condition remove the
+                // when in favor of the result. If there are no conditions left, remove the when
+                // entirely and replace it with an empty block.
+                if (expression.branches.size == 1) {
```
max-kammerer: Maybe best:
```                return if (expression.branches.isEmpty())
                    IrBlockImpl(expression.startOffset, expression.endOffset, context.irBuiltIns.unitType)
                else
                    expression.branches.first().takeIf { it.condition.isTrueConst() }?.result ?: expression
```

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/OperatorExpressionGenerator.kt
```diff
@@ -45,7 +45,6 @@ import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.typeUtil.isPrimitiveNumberType
 import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
 import org.jetbrains.kotlin.types.typeUtil.makeNullable
-import java.lang.AssertionError
```
max-kammerer: Pls remove this file from pr

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -93,6 +92,69 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+                }
+                return if (expression.branches.size == 0)
+                    IrBlockImpl(expression.startOffset, expression.endOffset, context.irBuiltIns.unitType)
+                else
+                    expression
+            }
+
+            private fun isImmutableTemporaryVariableWithConstantValue(statement: IrStatement): Boolean {
+                return statement is IrVariable &&
+                        statement.descriptor is IrTemporaryVariableDescriptor &&
```
max-kammerer: Please avoid descriptors in ir lowerings when possible, we are still in process of removing them from ir transformations.
Seems next code would be equivalent: statement.origin == IrDeclarationOrigin.IR_TEMPORARY_VARIABLE

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/OperatorExpressionGenerator.kt
madsager: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
madsager: Done, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
madsager: Neat! Done. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -93,6 +95,66 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+            }
+
+            private fun isImmutableTemporaryVariableWithConstantValue(statement: IrStatement): Boolean {
+                return statement is IrVariable &&
+                        statement.origin == IrDeclarationOrigin.IR_TEMPORARY_VARIABLE &&
+                        !statement.isVar &&
+                        statement.initializer is IrConst<*>
+            }
+
+            override fun visitBlock(expression: IrBlock): IrExpression {
```
max-kammerer: If I understanding correctly visitBlock/visitGetValue transformations are required to clean redundant bytecode after other transformation in this lower? If true please mention it in code with some examples

madsager: That is correct. I have added fairly large comments to explain what is going on. Let me know what you think.

max-kammerer:
Many Thanks!

PR #2070, opened 2019-01-22
JS: Add support for Closure library modules.

mpetrov:
This makes it much easier to integrate with existing codebases using
closure libraries and the closure compiler.
See https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide
for more info.

Note that this doesn't currently work well with DCE, but the intent is to
compile the output with closure anyway.


mpetrov:
Just synced to handle some merge conflicts that arose while this PR was out. Please take a look.

anton-bannykh:
@mpetrov Sorry, missed the original notification.  =(
Thank you for the PR, it looks interesting. =) Adding a new module kind requires consideration and might yield some amount of debate, so it might take some time still.

mpetrov:
Thanks @anton-bannykh,

For some prior art, here are some other relevant tools that emit closure modules: 
TypeScript: https://github.com/angular/tsickle
Java: https://github.com/google/j2cl

A big benefit here is one could do something like have a mixed TS and Kotlin codebase, have both emit closurized output, and have a single tool responsible for linking and optimizing the mixed language codebase.

Since the Kotlin output isn't type annotated, the number of optimizations that can be reliably performed is still limited. This is more of a first step to allow easy integration into existing closure projects and build systems (more specifically bazel in Google's case). Type annotated output could let you do interesting things like leverage clutz (https://github.com/angular/clutz) to emit .d.ts files (see KT-16604).

Schahen:
Hi Martin! Thank you for your PR. Just like it goes with every promising proposal, some things needs to be clarified so that we can move further and see what step makes sense to take next.

When talking of integrating "with existing codebases using closure libraries and the closure compiler" can you come up with some example of such integration. It would also be helpful if we can discuss how such integration is achievable with the current state of tools (if it is achievable at all).

mpetrov:
Sorry, going to close this out for now since I haven't had cycles to work on the JS side.

PR #2071, opened 2019-01-23
KT-28081 "Convert to lambda" changes expression type for interface with multiple supertypes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28081


mglukhikh:
Merged 👍

PR #2072, opened 2019-01-24
KT-29469 False positive in "Boolean literal argument without parameter name" inspection for varargs parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29469


mglukhikh:
Merged 👍

PR #2073, opened 2019-01-24
KT-28131 False positive "Redundant lambda arrow" with a functional type argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28131


idea/testData/inspectionsLocal/redundantLambdaArrow/typeParameter.kt
```diff
@@ -0,0 +1,6 @@
+// PROBLEM: none
```
mglukhikh: Just to be sure, could you please also add original example from the issue?

mglukhikh:
Merged 👍

PR #2077, opened 2019-01-25
Stop using System.currentTimeMillis for durations

ansman:
`measureTimeMillis` currently uses `Sytem.currentTimeMillis()` to measure elapsed time which is unsafe seeing as it is based on the system's clock which can be changed by the system at any time (for example by NTP or the user).

The new implementation would use `measureNanoTime` internally and calculate the number of milliseconds from that.


JakeWharton:
See also #1964

ansman:
Ah, nice catch!

PR #2081, opened 2019-01-25
KT-28485 "Incorrect parameter name after running "Add parameter to function" intention when argument variable is upper case const"

Tiltorito:
https://youtrack.jetbrains.com/issue/KT-28485

### DIFF
--- a/core/util.runtime/src/org/jetbrains/kotlin/utils/capitalizeDecapitalize.kt
+++ b/core/util.runtime/src/org/jetbrains/kotlin/utils/capitalizeDecapitalize.kt
@@ -54,7 +54,7 @@ fun String.decapitalizeSmart(asciiOnly: Boolean = false): String {
             validWords++
         }
 
-        return if(validWords > 1) builder.toString() else null
+        return if (validWords > 1) builder.toString() else null
     }
 
     val constant = decapitalizeConstant()
--- a/core/util.runtime/src/org/jetbrains/kotlin/utils/capitalizeDecapitalize.kt
+++ b/core/util.runtime/src/org/jetbrains/kotlin/utils/capitalizeDecapitalize.kt
@@ -36,28 +36,26 @@ fun String.decapitalizeSmart(asciiOnly: Boolean = false): String {
      * "__F_BAR" -> "fBar"
      * "F_BAR" -> "fBar"
      */
-    fun decapitalizeConstant(): String? {
+    fun decapitalizeConst(str: String): String? {
+        val words = str.split("_").filter { it.isNotEmpty() }
+
+        if (words.size <= 1) return null
+
         val builder = StringBuilder()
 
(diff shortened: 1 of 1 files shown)


goodwinnk:

I did only a quick overview, but as `decapitalizeSmart` is used from compiler, it means this might be a breaking change. Can KT-28485 be resolved for IDE only?

Also some tests are expected. Do you need help with writing them?

core/util.runtime/src/org/jetbrains/kotlin/utils/capitalizeDecapitalize.kt
```diff
@@ -20,13 +20,47 @@ package org.jetbrains.kotlin.util.capitalizeDecapitalize
...
     }
 
+    /**
+     * FOOBAR -> null
+     * FOO_BAR -> "fooBar"
+     * FOO_BAR_BAZ -> "fooBarBaz"
+     * "__F_BAR" -> "fBar"
+     * "F_BAR" -> "fBar"
+     */
+    fun decapitalizeConstant(): String? {
```
igorwojda: I think this implementation can be simplified.
We can also postpone `StringBuilder` creation when the input does not meet "const criteria"

```
private fun decapitalizeConst(str: String): String? {
    val words = str.split("_").filter { it.isNotEmpty() }
    
    if(words.size == 1) return null
    
    val builder = StringBuilder()
    
    words.forEachIndexed { index, word ->
        if(index == 0) {
            builder.append(word.toLowerCase())
        } else {
            builder.append(word.first().toUpperCase())
            builder.append(word.drop(1).toLowerCase())
        }
    }

    return builder.toString()
}
```

Tiltorito: Yeah, your code seems better :)

Tiltorito:
> Thank you for the contribution!
> 
> I did only a quick overview, but as `decapitalizeSmart` is used from compiler, it means this might be a breaking change. Can KT-28485 be resolved for IDE only?

It can be moved on KotlinNameSuggester in org.jetbrains.kotlin.idea.core package.

> Also some tests are expected. Do you need help with writing them?

Yes, please. Sorry about that, it is my first time contributing.

igorwojda:
@goodwinnk what do you think?

goodwinnk:
@Tiltorito @igorwojda Thank you for the contribution! I have pushed the commit to master as https://github.com/JetBrains/kotlin/commit/6ba134b1be3239920d264e0bac7e86c1c4dd6016 after squashing and humanizing commit message a bit. 

Also please take a look at https://github.com/JetBrains/kotlin/commit/213b98fefe86b822063450c8432770cd1f97dfe9 commit with reverting compiler behaviour and adding tests that might be useful for future contributions.

PR #2082, opened 2019-01-27
KT-19762 Add inspection for unclosed `(Auto)Closeable` with quickfix to wrap usage in `use`

DeFuex:

Old PR -> #1970

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -2301,6 +2301,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ConvertResourceToUseCallInspection"
+                     displayName="Convert resource to use() call"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="WEAK WARNING"
+                     language="kotlin"
+    />
+
     <localInspection implementationClass="org.jetbrains.kotlin.idea.intentions.ConvertTryFinallyToUseCallInspection"
                      displayName="Convert try / finally to use() call"
                      groupPath="Kotlin"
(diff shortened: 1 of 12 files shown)


DeFuex:
> 
> 
> Thank you for this amazing contribution. I see that a lot of work has been done but unfortunately a lot of corner cases are left. This is a hard inspection to implement and probably it should not have been marked with `Up For Grabs` tag.
> 
> I have commented the original request at https://youtrack.jetbrains.com/issue/KT-19762 and think that we need to discuss the corner cases there before putting additional efforts to the implementation.
> 
> If you're going to continue working on this inspection I suggest to wait for discussion end and divide the task to sub tasks. For example you could start from inspection only for some particular scenario. After that's is done and approved we could gradually add quick fixes. Also it probably should be two different inspections for the code that has an explicit `close()` and for code that doesn't.

I understand that it needs to be more defined and wasn't expecting it to be in the previous Kotlin release :smile: ...i think it just would be a waste of effort if i would stop now. Just let me know when you guys think that things are ready. In the meanwhile i'll try to improve the code and add more test cases when i got time. 

In any case, currently it should be testable from a perspective that it inspects every function call from an instance, assigned to variables (also non-assigned) and when it is derived from a Closeable. And of course the convertion works too...a lot of rough testing went into this... but i see what you mean with instances that shouldn't be leaked and close(), referring to your youtrack comment.

goodwinnk:
We now have more strict set of rules for adding new inspections and intentions and trying to avoid adding new checks when they have a big chance of false positives. 

Nobody is driving this inspection right now, so I'm going to close this pull-request. There is a link left in the ticket to this PR so your work could be reused when working on the `(Auto)Closeable` inspection is continued. 

Thank you for work and time and I'm sorry that we didn't recognize potential problems with this inspection beforehand.

DeFuexx:
No problem, i was just super busy with work lately. If i get the chance i may tackle it again or do another ticket :)

PR #2085, opened 2019-01-29
KT-20357 - add sample for Regex.findAll()

kevpeek:

### DIFF
--- a/libraries/stdlib/samples/test/samples/text/regex.kt
+++ b/libraries/stdlib/samples/test/samples/text/regex.kt
@@ -51,7 +51,7 @@ class Regexps {
         val text = "Hello Alice. Hello Bob. Hello Eve."
         val regex = Regex("Hello (.*?)[.]")
         val matches = regex.findAll(text)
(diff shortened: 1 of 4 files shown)


libraries/stdlib/samples/test/samples/text/regex.kt
```diff
@@ -45,4 +45,13 @@ class Regexps {
         // Because the search starts from the index 2, it finds the last "to be".
         assertPrints(regex3.find(inputString, 2)!!.range, "13..17")
     }
+
+    @Sample
+    fun findAll() {
+        val text = "Hello Alice. Hello Bob. Hello Eve."
+        val regex = Regex("Hello (.*?)[.]")
+        val matches = regex.findAll(text)
+        val names = matches.map(MatchResult::groupValues).map {it[1]} .joinToString(",")
```
ilya-g: Two subsequent `map` calls can be joined together.
The default delimiter of `joinToString` would produce more pretty output.

kevpeek:
@ilya-g I combined the two map operations and changed to the default separator value in joinToString.

ilya-g:
I've merged it, thanks.

PR #2086, opened 2019-01-29
KT-11629 Inspection: creating Throwable without throwing it

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11629

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.search.searches.ReferencesSearch
+import com.siyeh.ig.psiutils.TestUtils
+import org.jetbrains.kotlin.builtins.DefaultBuiltIns
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.ConstructorDescriptor
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getParentOfTypes
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
+import org.jetbrains.kotlin.resolve.descriptorUtil.isSubclassOf
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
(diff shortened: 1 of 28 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+        val description = if (resultingDescriptor is ConstructorDescriptor) {
+            "Throwable instance '${calleeExpression.text}' is not thrown"
+        } else {
+            "Result of '${calleeExpression.text}' call is not thrown"
+        }
+        holder.registerProblem(calleeExpression, description, ProblemHighlightType.GENERIC_ERROR_OR_WARNING)
+    })
+
+    private fun KtExpression.isUsed(checkDeep: Boolean = true): Boolean {
+        val context = analyze(BodyResolveMode.PARTIAL_WITH_DIAGNOSTICS)
```
mglukhikh: `PARTIAL_WITH_CFA` should be enough here

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+        if (context[BindingContext.USED_AS_EXPRESSION, this] != true) return false
```
mglukhikh: Why don't you want to use `isUsedAsExpression` extension here?

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        val property = getParentOfTypes(
+            true,
+            KtThrowExpression::class.java,
+            KtReturnExpression::class.java,
+            KtProperty::class.java
+        ) as? KtProperty ?: return true
+        if (!property.isLocal || !checkDeep) return true
+        return ReferencesSearch.search(property).any {
+            (it.element as? KtExpression)?.isUsed(checkDeep = false) == true
```
mglukhikh: Not sure it's necessary here. I think it's enough just to check that property is used at all.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinThrowableNotThrownInspection.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.jetbrains.kotlin.resolve.descriptorUtil.isSubclassOf
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.isNullable
+import org.jetbrains.kotlin.types.typeUtil.isNothing
+
+class KotlinThrowableNotThrownInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = callExpressionVisitor(fun(callExpression) {
+        val calleeExpression = callExpression.calleeExpression ?: return
+        if (TestUtils.isInTestSourceContent(callExpression)) return
+        val resultingDescriptor = callExpression.resolveToCall()?.resultingDescriptor ?: return
```
mglukhikh: Consider using a heuristics to prevent resolve here: we could first check whether `Exception` or `Error` is a substring in `calleeExpression`, and if not, just return (I'm not quite sure is it a good idea or not, decide it yourself)

mglukhikh:
Merged, thank you.

PR #2087, opened 2019-01-30
KT-29567 "Remove empty class body" is a poor name for inspection text

tommykw:
https://youtrack.jetbrains.com/issue/KT-29567


mglukhikh:
Merged 👍

PR #2088, opened 2019-01-31
KT-26269 Inspection to replace associate with associateWith or associateBy

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26269

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceAssociateFunctionInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceAssociateFunctionInspection.kt
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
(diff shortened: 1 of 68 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceAssociateFunctionInspection.kt
```diff
@@ -0,0 +1,189 @@
...
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+
+class ReplaceAssociateFunctionInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = dotQualifiedExpressionVisitor(fun(dotQualifiedExpression) {
+        if (dotQualifiedExpression.languageVersionSettings.languageVersion < LanguageVersion.KOTLIN_1_3) return
+        val callExpression = dotQualifiedExpression.callExpression ?: return
+        val calleeExpression = callExpression.calleeExpression ?: return
+
+        val context = dotQualifiedExpression.analyze(BodyResolveMode.PARTIAL)
+        val isAssociate = callExpression.isCalling(FqName("kotlin.collections.associate"), context)
```
mglukhikh: I have two concerns here: 

- I would extract all four FqNames into companion object properties. 
- We call `analyze` on each dot qualified expression, which is slow (yes, it's slow in practice, you can try to launch this inspection on Kotlin project). `isCalling` could solve this problem, but unfortunately you call `analyze` before. It's also bad solution to remove `context` here. I think we should first check that callee name starts with `associate` and then check that function is resolved to one of four (better not via `isCalling` because it would repeat the same text check).

mglukhikh:
Merged, thank you!

PR #2089, opened 2019-01-31
KT-29606 Do not propose to remove unused parameter of property setter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29606


mglukhikh:
Merged 👍

PR #2090, opened 2019-01-31
JVM_IR: Do not add redundant field initializers.

punzki:
Initializers are "set field" expressions and are considered redundant when they are:
1. In the primary constructor; and
2. Set the field to `0`, `false`, or `null`; and
3. Have a `null` origin. I.e., not in an initializer block or constructor body, and therefore the field could not have been set by a prior expression.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -33,6 +33,8 @@ import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.symbols.IrSymbol
+import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.types.isMarkedNullable
 import org.jetbrains.kotlin.ir.types.isNothing
 import org.jetbrains.kotlin.ir.types.toKotlinType
 import org.jetbrains.kotlin.ir.util.dump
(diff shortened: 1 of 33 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -455,12 +457,49 @@ class ExpressionCodegen(
...
 
+    private fun skipDefaultValue(fieldType: IrType, constExpression: IrConst<*>): Boolean {
+        val value = constExpression.value
+        val type = constExpression.asmType
+        if (isPrimitive(type)) {
+            if (!fieldType.isMarkedNullable() && value is Number) {
+                if (type in setOf(Type.INT_TYPE, Type.BYTE_TYPE, Type.LONG_TYPE, Type.SHORT_TYPE) && value.toLong() == 0L) {
+                    return true
+                }
+                if (type === Type.DOUBLE_TYPE && value == 0.0) {
```
max-kammerer: This is not correct value  == 0.0 check performed via IEEE754,
so 0.0 == - 0.0 (true),  
but 0.0.equals (-0.0 ) (false)

max-kammerer: Please also add test for such cases

compiler/testData/codegen/bytecodeText/redundantValInitializer.kt
```diff
@@ -0,0 +1,30 @@
+package a
+
+class A {
+    companion object {
+        const val constInt: Int = 0
```
max-kammerer: I'm not sure but maybe we need also tests non const vals?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
punzki: Actually this works out fine because `value` is of type `Number` and therefore `value == 0.0` invokes `Number.equals`. But to be explicit, changed to `value.toDouble().equals(0.0)` and added tests.

Tangentially related: When I copy pasted code from https://github.com/JetBrains/kotlin/blob/8eeb9d4faab68b4821bbad62ae38a59b03f16930/compiler/backend/src/org/jetbrains/kotlin/codegen/MemberCodegen.java#L569, I had the Kotlin plugin convert the code as a starting point, and it used `==` instead of `.equals`. Bug in the converter (at least for doubles and floats)? So the simplest repro of this bug is something like this:

```
        Double posZero = 0.0;
        Double negZero = -0.0;
        if (posZero.equals(negZero)) System.out.println("EQUAL");
        else System.out.println("NOT EQUAL");
```

gets converted to
```
        val posZero = 0.0
        val negZero = -0.0
        if (posZero == negZero)
            println("EQUAL")
        else
            println("NOT EQUAL")
```

...which aren't the same. And the editor gives a warning that `if (posZero == negZero)` is always false, which isn't correct either.

compiler/testData/codegen/bytecodeText/redundantValInitializer.kt
punzki: I figured that I covered both by having the "my*" fields pointing to the constants in the companion object. Anyway to be explicit, I added fields directly initialized to literal constants.

I also added tests for fields.

punzki:
@max-kammerer I somehow messed up this PR and couldn't fix it. Created a new one: https://github.com/JetBrains/kotlin/pull/2094

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: @punzki Thanks, I've filed KT-29691, KT-29692

PR #2091, opened 2019-02-01
Navigation to implementation / overriding method: do not show method of inline class twice

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28661
https://youtrack.jetbrains.com/issue/KT-26924


goodwinnk:
Great commit, thank you! 

Resulting commit is https://github.com/JetBrains/kotlin/commit/80e1fc2acefff3ee9ad345d7b0ee4cb48ff593b9.

PR #2092, opened 2019-02-01
KT-17278 Inspection to replace Java 8 Map.forEach with Kotlin's forEach

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17278


mglukhikh:
Merged manually, thank you.

PR #2093, opened 2019-02-01
JVM_IR: Fix stack underflow for unit coercion of when expressions.

madsager:
Code such as

```
val b = getBoolean()
if (b) 4
else if (b) 5
```

didn't generate a value on the stack always and therefore would
have control-flow paths leading to a pop instruction with nothing
on the stack.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLoweringPhases.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLoweringPhases.kt
@@ -183,7 +183,6 @@ private val BridgePhase = makeJvmPhase(
     description = "Generate bridges"
 )
 
-
 private val JvmOverloadsAnnotationPhase = makeJvmPhase(
     { context, file -> JvmOverloadsAnnotationLowering(context).lower(file) },
     name = "JvmOverloadsAnnotation",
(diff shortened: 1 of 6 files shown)


jimgoog:
cc @erokhins @max-kammerer @romanart

jimgoog:
Not sure if this is related (ie. should be in this PR or a followup), but this still fails AFAIK:

```
fun Foo(children: () -> Unit) {}
fun Bar(children: () -> Unit) {
    Foo {
        val c = true
        if (c) {
            throw RuntimeException()
        } else {
            children()
        }
    }
}
```

max-kammerer:
@madsager  your approach is one of possible solution.
But maybe best to move COERCION_TO_UNIT inside branches
But I'm not sure would it be much more tricky:
It's also intresting to mix long and int types in branches, e.g.:
```
fun test(arg: Int) {
    if (arg == 1) {
        result = "firstResult"
        longValue
    } else if (arg == 2) {
        result = "secondResult"
        arg
    }
}
```

madsager:
Thanks for the comments Mike. I had a closer look and it seems like my current patch is working properly and that it will not immediately work to push the coercions into the branch results.

Here is what is happening in my current patch for the double/int mix: the type of the when expression becomes Object and therefore, there will be boxing of the int and the double and therefore one pop instruction is enough to get rid of that object. In the low level bytecode optimizer this case is then recognized and the boxing to an object that is then immediately popped is replace with a pop of the right type. Therefore, the correct code which uses the right pop instruction pop2 for the double and pop for the int, is generated.

If we push the unit coercion into the branches we still get unbalanced stacks unless we also add an else branch as done in the current patch. The reason is that the unit coercion puts a unit instance on the stack in both branches. However, when none of the branches are taking we don't get a value on the stack and therefore we get flow to a pop instruction with nothing on the stack.

Therefore, it seems to me that the current patch is the simplest to make this work. Also, it doesn't lead to pop instructions in all the branches when one pop at the end is enough.

What do you think?

max-kammerer:
@madsager I will merge pr as is, seems we need to perform more investigations to avoid Unit instance generation in branches with further poping them outside of when.
Many thanks!

PR #2094, opened 2019-02-01
JVM_IR: Do not add redundant field initializers.

punzki:
Initializers are "set field" expressions and are considered redundant
when they are:
1. In the primary constructor; and
2. Set the field to `0`, `false`, or `null`; and
3. Have a `null` origin. I.e., not in an initializer block or
constructor body, and therefore the field could not have been set by a
prior expression.


punzki:
@max-kammerer Messed up the previous PR, but I've addressed your comments from https://github.com/JetBrains/kotlin/pull/2090 here.

max-kammerer:
@punzki Many thanks!

PR #2095, opened 2019-02-02
KT-29128 False positive 'Explicitly given type is redundant here' when typealias is used

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29128


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantExplicitTypeInspection.kt
```diff
@@ -20,7 +21,7 @@ class RedundantExplicitTypeInspection : AbstractKotlinInspection() {
             if (!property.isLocal) return
             val typeReference = property.typeReference ?: return
             val initializer = property.initializer ?: return
-
+            if ((typeReference.typeElement as? KtUserType)?.referenceExpression?.mainReference?.resolve() is KtTypeAlias) return
```
mglukhikh: I'm almost sure this can be done easier, without additional reference resolve. Probably `type` should be `WrappedType` for type alias (not sure exactly).

PR #2097, opened 2019-02-02
KT-29085 False positive "Class member can have 'private' visibility" for a `const val` used in a public inline function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29085


mglukhikh:
Merged 👍

PR #2098, opened 2019-02-02
KT-29248 "Convert member to extension" doesn't preserve visibility

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29248


mglukhikh:
Merged 👍

PR #2099, opened 2019-02-02
KT-29556 "Remove redundant 'let' call" doesn't rename parameter with convention `invoke` call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29556


mglukhikh:
Merged 👍

PR #2101, opened 2019-02-04
KT-29321 "Remove empty primary constructor": apply for enum entries

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29321


mglukhikh:
Merged 👍

PR #2102, opened 2019-02-04
KT-29113 Warn about  redundant requireNotNull and checkNotNull usages

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29113

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantRequireNotNullCallInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantRequireNotNullCallInspection.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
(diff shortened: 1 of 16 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantRequireNotNullCallInspection.kt
```diff
@@ -0,0 +1,75 @@
...
+
+class RedundantRequireNotNullCallInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = callExpressionVisitor(fun(callExpression) {
+        val callee = callExpression.calleeExpression ?: return
+        val functionName = callee.text
+        if (functionName !in listOf("requireNotNull", "checkNotNull")) return
+
+        val context = callExpression.analyze(BodyResolveMode.PARTIAL)
+        val fqName = callee.getResolvedCall(context)?.resultingDescriptor?.fqNameSafe ?: return
+        if (fqName !in listOf(FqName("kotlin.requireNotNull"), FqName("kotlin.checkNotNull"))) return
```
mglukhikh: Probably we should use here `isCalling` extension we have added some time ago

mglukhikh:
Merged 👍

PR #2103, opened 2019-02-05
KT-29677 "Specify type explicitly" intention produces invalid output for type escaped with backticks

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29677


mglukhikh:
Merged 👍

PR #2104, opened 2019-02-07
KT-25262 Intention: Rename class to containing file name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25262


mglukhikh:
Merged 👍

PR #2105, opened 2019-02-07
KT-29193 Quick fix "Create extension function" `List<Int>.set` should not be suggested for read-only collections

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29193


mglukhikh:
Merged 👍

PR #2106, opened 2019-02-07
JVM_IR: Block with nothing type do not generate values on the stack.

madsager:


madsager:
@max-kammerer Minor follow-up to previous change dealing with blocks with Nothing type.

max-kammerer:
Thanks!

PR #2107, opened 2019-02-07
Add Locale-accepting overloads for (de)capitalize in JDK stdlib

JakeWharton:
#KT-28933 fixed

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
+++ b/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
@@ -568,6 +568,31 @@ public actual fun String.capitalize(): String {
 }
 
 /**
+ * Returns a copy of this string having its first letter uppercased preferring [Character.toTitleCase] (if different from
+ * [Character.toUpperCase]) or by [String.toUpperCase] using [locale], or the original string, if it's empty or already
+ * starts with an upper case letter.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun String.capitalize(locale: Locale): String {
+    if (isNotEmpty()) {
+        val firstChar = this[0]
+        if (firstChar.isLowerCase()) {
+            return buildString {
+                val titleChar = firstChar.toTitleCase()
+                if (titleChar != firstChar.toUpperCase()) {
+                    append(titleChar)
+                } else {
+                    append(this@capitalize.substring(0, 1).toUpperCase(locale))
+                }
+                append(this@capitalize.substring(1))
+            }
+        }
+    }
+    return this
+}
+
+/**
  * Returns a copy of this string having its first letter lowercased, or the original string,
  * if it's empty or already starts with a lower case letter.
  *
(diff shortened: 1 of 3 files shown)


JakeWharton:
Anything I can do here?

libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
```diff
@@ -424,8 +424,14 @@ public inline fun String.toPattern(flags: Int = 0): java.util.regex.Pattern {
  *
  * @sample samples.text.Strings.capitalize
  */
-public actual fun String.capitalize(): String {
-    return if (isNotEmpty() && this[0].isLowerCase()) substring(0, 1).toUpperCase() + substring(1) else this
+public actual fun String.capitalize() = capitalize(Locale.getDefault())
```
ilya-g: Better to leave the previous implementation to avoid accidentally changing behavior in 1.3.x release.

libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
```diff
@@ -424,8 +424,14 @@ public inline fun String.toPattern(flags: Int = 0): java.util.regex.Pattern {
...
+
+/**
+ * Returns a copy of this string having its first letter uppercased using [locale], or the original string,
+ * if it's empty or already starts with an upper case letter.
+ */
+public fun String.capitalize(locale: Locale): String {
```
ilya-g: The added functions should be annotated with `@SinceKotlin("1.3")` and the new experimental annotation `@ExperimentalStdlibApi`, so we could publish them in 1.3.x release.

ilya-g:
When capitalizing a string if the first character is not in the upper case, we'd like to honor the title case variant of that character. Unfortunately JDK doesn't provided access to Unicode [special casing rules](https://unicode.org/Public/UNIDATA/SpecialCasing.txt) for titlecasing: `Char.toTitleCase` only handles 1-to-1 title case mappings, and `String` doesn't have `toTitleCase` function which could cover other 1-to-n mappings and locale or context sensitive mappings.

At least we can try to capitalize characters to title case for which `Char.toTitleCase()` returns different result comparing to `Char.toUpperCase()` (there are not so many of them and they are not locale dependent). Otherwise, we can resort to uppercasing the first char with the given locale.

JakeWharton:
Wouldn't using title case be a behavior change? Or are you saying you only want the locale-accepting versions of this function to have that behavior?

ilya-g:
For the new experimental functions that won't be a behavior change. For the existing functions this change could be discussed in the language committee, and based on that either the new or old function's behavior could be changed, or left to be different.

JakeWharton:
Updated. Let me know what you think.

JakeWharton:
Rebased to fix conflict in API mapping file

libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
```diff
@@ -577,6 +602,16 @@ public actual fun String.decapitalize(): String {
...
 }
 
+/**
+ * Returns a copy of this string having its first letter lowercased using [locale], or the original string,
+ * if it's empty or already starts with a lower case letter.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun String.decapitalize(locale: Locale): String {
+    return if (isNotEmpty() && this[0].isUpperCase()) substring(0, 1).toLowerCase(locale) + substring(1) else this
```
ilya-g: I've also changed `isUpperCase()` to `!isLowerCase()` here to cover cases, when the first char is in title case. See 8badfca4592dbac6faf0260159c73aed72285d69

JakeWharton: Ah, yep. Thanks!

JakeWharton:
Merged as af31794f605426cb014161d79e6201a0a309f09d

PR #2108, opened 2019-02-08
KT-28596 "Can be replaced with binary operator" shouldn't be suggested when receiver or argument is floating point type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28596

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/smartcasts/DataFlowUtils.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/smartcasts/DataFlowUtils.kt
@@ -25,7 +25,8 @@ fun KtExpression?.getKotlinTypeWithPossibleSmartCastToFP(
     bindingContext: BindingContext,
     descriptor: DeclarationDescriptor?,
     languageVersionSettings: LanguageVersionSettings,
-    dataFlowValueFactory: DataFlowValueFactory
+    dataFlowValueFactory: DataFlowValueFactory,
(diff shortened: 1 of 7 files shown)


t-kameyama:
I will fix some mistakes.

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/smartcasts/DataFlowUtils.kt
```diff
@@ -40,13 +40,13 @@ fun KtExpression?.getKotlinTypeWithPossibleSmartCastToFP(
...
         val stableTypes = bindingContext.getDataFlowInfoBefore(this).getStableTypes(dataFlow, languageVersionSettings)
-        stableTypes.firstNotNullResult {
+        return stableTypes.firstNotNullResult {
             when {
                 KotlinBuiltIns.isDoubleOrNullableDouble(it) -> it
                 KotlinBuiltIns.isFloatOrNullableFloat(it) -> it
                 else -> null
             }
-        }?.let { return it }
+        } ?: stableTypes.firstOrNull() ?: givenType
```
mglukhikh: I'm afraid we can break something in back-end with this change, because this function has the second use-site in back-end. Also, this change contradicts with the function name. We should keep back-end behavior here. I'd consider adding a lambda as the last function parameter, which takes `givenType` and `stableTypes` as arguments and returns something like `defaultType`. By default this lambda should return `givenType`, but in inspection we can change its value to `stableTypes.firstOrNull() ?: givenType`.

mglukhikh:
Merged 👍

PR #2109, opened 2019-02-10
KT-29056 KNPE in ConvertPrimaryConstructorToSecondary with missing property identifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29056


mglukhikh:
Merged 👍

PR #2111, opened 2019-02-12
Implement constant folding in the IR backend for JVM

ting-yuan:
The newly added pass folds the same set of constant functions of the
current backend.

Change-Id: I739bcb3e0fc549a87123422d7ca46e44d61f24dd

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
@@ -674,20 +674,9 @@ private class ConstantExpressionEvaluatorVisitor(
         ConstantExpressionEvaluator.getConstant(expression, trace.bindingContext)?.isPure ?: false
 
     private fun evaluateUnaryAndCheck(receiver: OperationArgument, name: String, callExpression: KtExpression): Any? {
-        val functions = unaryOperations[UnaryOperationKey(receiver.ctcType, name)] ?: return null
-
-        val (function, check) = functions
-        val result = function(receiver.value)
-        if (check == emptyUnaryFun) {
-            return result
-        }
-        assert(isIntegerType(receiver.value)) { "Only integer constants should be checked for overflow" }
-        assert(name == "minus" || name == "unaryMinus") { "Only negation should be checked for overflow" }
-
-        if (receiver.value == result && !isZero(receiver.value)) {
+        return evaluateUnaryAndCheck(name, receiver.ctcType, receiver.value) {
             trace.report(Errors.INTEGER_OVERFLOW.on(callExpression.getStrictParentOfType<KtExpression>() ?: callExpression))
         }
-        return result
     }
 
     private fun evaluateBinaryAndCheck(
@@ -696,48 +685,13 @@ private class ConstantExpressionEvaluatorVisitor(
         name: String,
         callExpression: KtExpression
     ): Any? {
-        val functions = getBinaryOperation(receiver, parameter, name) ?: return null
-
-        val (function, checker) = functions
-        val actualResult = try {
-            function(receiver.value, parameter.value)
-        } catch (e: Exception) {
-            null
-        }
-        if (checker == emptyBinaryFun) {
-            return actualResult
-        }
-        assert(isIntegerType(receiver.value) && isIntegerType(parameter.value)) { "Only integer constants should be checked for overflow" }
-
-        fun toBigInteger(value: Any?) = BigInteger.valueOf((value as Number).toLong())
-
-        val refinedChecker = if (name == OperatorNameConventions.MOD.asString()) {
-            getBinaryOperation(receiver, parameter, OperatorNameConventions.REM.asString())?.second ?: return null
-        } else {
-            checker
-        }
-
-        val resultInBigIntegers = refinedChecker(toBigInteger(receiver.value), toBigInteger(parameter.value))
-
-        if (toBigInteger(actualResult) != resultInBigIntegers) {
+        return evaluateBinaryAndCheck(name, receiver.ctcType, receiver.value, parameter.ctcType, parameter.value) {
             trace.report(Errors.INTEGER_OVERFLOW.on(callExpression.getStrictParentOfType<KtExpression>() ?: callExpression))
         }
-        return actualResult
     }
-
-    private fun getBinaryOperation(receiver: OperationArgument, parameter: OperationArgument, name: String) =
-        binaryOperations[BinaryOperationKey(receiver.ctcType, parameter.ctcType, name)]
 
     private fun isDivisionByZero(name: String, parameter: Any?): Boolean {
         return name in DIVISION_OPERATION_NAMES && isZero(parameter)
-    }
-
-    private fun isZero(value: Any?): Boolean {
-        return when {
-            isIntegerType(value) -> (value as Number).toLong() == 0L
-            value is Float || value is Double -> (value as Number).toDouble() == 0.0
-            else -> false
-        }
     }
 
     override fun visitUnaryExpression(expression: KtUnaryExpression, expectedType: KotlinType?): CompileTimeConstant<*>? {
@@ -1167,4 +1121,97 @@ fun CompileTimeConstant<*>.isStandaloneOnlyConstant(): Boolean {
         is TypedCompileTimeConstant -> this.constantValue.isStandaloneOnlyConstant()
         else -> return false
     }
+}
+
+private fun isZero(value: Any?): Boolean {
+    return when {
+        isIntegerType(value) -> (value as Number).toLong() == 0L
+        value is Float || value is Double -> (value as Number).toDouble() == 0.0
+        else -> false
+    }
+}
+
+private fun typeStrToCompileTimeType(str: String) = when (str) {
+    BYTE.name -> BYTE
+    SHORT.name -> SHORT
+    INT.name -> INT
+    LONG.name -> LONG
+    DOUBLE.name -> DOUBLE
+    FLOAT.name -> FLOAT
+    CHAR.name -> CHAR
+    BOOLEAN.name -> BOOLEAN
+    STRING.name -> STRING
+    ANY.name -> ANY
+    else -> throw IllegalArgumentException("Unsupported type: $str")
+}
+
+fun evaluateUnary(name: String, typeStr: String, value: Any, tracer: () -> Unit = {}): Any? {
+    return evaluateUnaryAndCheck(name, typeStrToCompileTimeType(typeStr), value)
+}
+
+private fun evaluateUnaryAndCheck(name: String, type: CompileTimeType<*>, value: Any, tracer: () -> Unit = {}): Any? {
+    val functions = unaryOperations[UnaryOperationKey(type, name)] ?: return null
+
+    val (function, check) = functions
+    val result = function(value)
+    if (check == emptyUnaryFun) {
+        return result
+    }
+    assert(isIntegerType(value)) { "Only integer constants should be checked for overflow" }
+    assert(name == "minus" || name == "unaryMinus") { "Only negation should be checked for overflow" }
+
+    if (value == result && !isZero(value)) {
+        tracer()
+    }
+    return result
+}
+
+fun evaluateBinary(
+    name: String,
+    receiverTypeStr: String,
+    receiverValue: Any,
+    parameterTypeStr: String,
+    parameterValue: Any
+): Any? {
+    val receiverType = typeStrToCompileTimeType(receiverTypeStr)
+    val parameterType = typeStrToCompileTimeType(parameterTypeStr)
+
+    return evaluateBinaryAndCheck(name, receiverType, receiverValue, parameterType, parameterValue)
+}
+
+private fun evaluateBinaryAndCheck(
+    name: String,
+    receiverType: CompileTimeType<*>,
+    receiverValue: Any,
+    parameterType: CompileTimeType<*>,
+    parameterValue: Any,
+    tracer: () -> Unit = {}
+): Any? {
+    val functions = binaryOperations[BinaryOperationKey(receiverType, parameterType, name)] ?: return null
+
+    val (function, checker) = functions
+    val actualResult = try {
+        function(receiverValue, parameterValue)
+    } catch (e: Exception) {
+        null
+    }
+    if (checker == emptyBinaryFun) {
+        return actualResult
+    }
+    assert(isIntegerType(receiverValue) && isIntegerType(parameterValue)) { "Only integer constants should be checked for overflow" }
+
+    fun toBigInteger(value: Any?) = BigInteger.valueOf((value as Number).toLong())
+
+    val refinedChecker = if (name == OperatorNameConventions.MOD.asString()) {
+        binaryOperations[BinaryOperationKey(receiverType, parameterType, OperatorNameConventions.REM.asString())]?.second ?: return null
+    } else {
+        checker
+    }
+
+    val resultInBigIntegers = refinedChecker(toBigInteger(receiverValue), toBigInteger(parameterValue))
+
+    if (toBigInteger(actualResult) != resultInBigIntegers) {
+        tracer()
+    }
+    return actualResult
 }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.jvm.lower
+
+import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
+import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.descriptors.IrBuiltIns
+import org.jetbrains.kotlin.ir.expressions.IrCall
+import org.jetbrains.kotlin.ir.expressions.IrConst
+import org.jetbrains.kotlin.ir.expressions.IrExpression
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+import org.jetbrains.kotlin.resolve.constants.evaluate.evaluateBinary
+import org.jetbrains.kotlin.resolve.constants.evaluate.evaluateUnary
+
+internal val foldConstantLoweringPhase = makeIrFilePhase(
+    ::FoldConstantLowering,
+    name = "FoldConstantLowering",
+    description = "Constant Folding"
+)
+
+/**
+ * A pass to fold constant expressions of most common types.
+ *
+ * For example, the expression "O" + 'K' + (1.toLong() + 2.0) will be folded to "OK3.0" at compile time.
+ *
+ * TODO: constant fields (e.g. Double.NaN)
+ */
+class FoldConstantLowering(private val context: JvmBackendContext) : IrElementTransformerVoid(), FileLoweringPass {
+
+    /**
+     * ID of an unary operator / method.
+     *
+     * An unary operator / method can be identified by its operand type (in full qualified name) and its name.
+     */
+    private data class UnaryOp(
+        val operandType: String,
+        val operatorName: String
+    )
+
+    /**
+     * ID of an binary operator / method.
+     *
+     * An binary operator / method can be identified by its operand types (in full qualified names) and its name.
+     */
+    private data class BinaryOp(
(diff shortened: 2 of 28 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -0,0 +1,348 @@
...
+                registerBinaryOp<Long, Any>(LONG, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Float, Any>(FLOAT, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Double, Any>(DOUBLE, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<String, Any>(STRING, type, "equals") { a, b -> a.equals(b) }
+            }
+        }
+    }
+
+    private fun buildIrConstant(call: IrCall, v: Any): IrExpression {
+        return if (call.type.isInt()) {
```
dnpetrov: Minor (code style): replace if-else if-... with `when`.

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
```diff
@@ -1,13 +1,16 @@
 // IGNORE_BACKEND: JVM_IR
 fun box() : String {
-    230?.toByte()?.hashCode()
-    9.hashCode()
+    // Just hard enough that the test won't get optimized away at compile time.
```
dnpetrov: Were there any issues with this tests involving constant folding?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
max-kammerer: Pls, replace with 'when'

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -0,0 +1,348 @@
...
+    name = "FoldConstantLowering",
+    description = "Constant Folding"
+)
+
+/**
+ * A pass to fold constant expressions of most common types.
+ *
+ * For example, the expression "O" + 'K' + (1.toLong() + 2.0) will be folded to "OK3.0" at compile time.
+ */
+class FoldConstantLowering(private val context: JvmBackendContext) : IrElementTransformerVoid(), FileLoweringPass {
```
dnpetrov: This lowering pass adds quite some functionality to the JVM_IR back-end. I'd like to see some test for that.
Ideally, this should include two kinds of tests:
1. Box tests - to make sure everything works as expected;
2. Bytecode tests - to make sure that constant expressions are, indeed, folded.
Look into compiler/testData/codegen/box and compiler/testData/codegen/bytecodeText.

Pay attention to tests that can cause exceptions - e.g., division by zero.
```
fun test1() = 1 / 0 // ArithmeticException at run-time
fun test2() = 1.0 / 0.0 // Infinity
```

Also make sure that floating-point number comparisons work fine with respect to IEEE754 vs total order (various "special" numbers such as `-0.0` and `java.lang.Double.NaN`).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        return buildIrConstant(call, evaluated)
+    }
+
+    override fun lower(irFile: IrFile) {
+        irFile.transformChildrenVoid(object : IrElementTransformerVoid() {
+            override fun visitCall(expression: IrCall): IrExpression {
+                expression.transformChildrenVoid(this)
+                return when (expression.valueArgumentsCount) {
```
max-kammerer: We have 2 representations for compare and equals operation in IR:
1 < 1 is binary operation with 2 argument 
1.compareTo(1) - call with dispatch receiver and 1 argument 
Maybe call.receiverAndArgs()  would be worth
Seems this cases are not tested.

Please also note that:
-0.0 < 0.0 calculation is performed in IEEE754 arithmetics and
-0.0.compareTo(0.0) in non-IEEE754 
This affects  registerSymmetricBinaryOp<Double, Double>(DOUBLE, opType, "compareTo") { a, b -> a.compareTo(b) } same applies for == and equals operation. (Similar with Floats)

Seems other operations should be equals to each other (+ and plus and so on), but maybe I miss smth @dnpetrov

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
dnpetrov: BlackBoxCodegenTestGenerated$PrimitiveTypes.testKt2269 fails: 
Unresolved reference: toInt (4,27) in /kt2269.kt

max-kammerer: seems // WITH_RUNTIME directive is required

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: It looks like that most of the box tests are already there in the code base? Although only either constant folding or generated bytecode will be tested, but not both; If an expression is folded, the bytecode generated won't be tested, and vice versa.

Since this is constant folding, the result should be a constant stored in the text. I'll start by adding bytecode tests and check whether there are box tests which are folded/optimized away.

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
ting-yuan: With this pass and some branch optimizations by Mads, the test simply returns OK :) The conditions are optimized away:
```
...
LDC "OK"
ARETURN
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: Done

ting-yuan: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: Please take a look at the new commit. They are added but not enabled yet; Because the current backed doesn't fold those comparisons, the new tests would fail on it. This [PR](https://github.com/JetBrains/kotlin/pull/2137) tries to fix that.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: Thanks for the pointers! I've been wondering where those IrBuiltins are used :) Please take a look at the new commit, where the support of folding some IrBuiltins, mainly for those comparisons, are added.

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
ting-yuan: Updated with another approach so that  runtime isn't required. Also verified with :compiler:tests

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -0,0 +1,385 @@
...
+            registerUnaryOp<T>(operandType, "toByte") { it.toByte() }
+            registerUnaryOp<T>(operandType, "toShort") { it.toShort() }
+            registerUnaryOp<T>(operandType, "toInt") { it.toInt() }
+            registerUnaryOp<T>(operandType, "toLong") { it.toLong() }
+            registerUnaryOp<T>(operandType, "toFloat") { it.toFloat() }
+            registerUnaryOp<T>(operandType, "toDouble") { it.toDouble() }
+            registerUnaryOp<T>(operandType, "toString") { it.toString() }
+        }
+
+        init {
```
udalov: Is it possible to somehow use operations in `OperationsMapGenerated.kt` instead of manually enumerating them here? That would also make it future-proof in case we add more built-in operations in the future (like we did with `rem` in 1.1 for example)

compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
```diff
@@ -2,16 +2,16 @@ class A
 
 fun foo(x: Any?) {}
 
-fun box() {
+fun box(u: Int) {
```
max-kammerer: Why you touch this test?

compiler/testData/codegen/bytecodeText/deadCodeElimination/simpleConstructorNotRedundant.kt
```diff
@@ -1,8 +1,8 @@
 class A
-fun box() {
+fun box(u: Int) {
```
max-kammerer: and this?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -0,0 +1,385 @@
...
+                registerBinaryOp<Byte, Any>(BYTE, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Short, Any>(SHORT, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Int, Any>(INT, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Long, Any>(LONG, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Float, Any>(FLOAT, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<Double, Any>(DOUBLE, type, "equals") { a, b -> a.equals(b) }
+                registerBinaryOp<String, Any>(STRING, type, "equals") { a, b -> a.equals(b) }
+            }
+
+            // IrBuiltins
```
udalov: It would be awesome to extract names used here into constants in IrBuiltIns and use them where the corresponding operators are created, to be able to quickly see that both refer to the same concept

compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
ting-yuan: Similar to changes in concat.kt and primitiveToString.kt, this prevents the relevant part of the test being optimized away. Without this change, it would be:

expected:
<0 IFNULL
[1 IFNONNULL
1] throwNpe
0 ATHROW
> but was:
<0 IFNULL
[0 IFNONNULL
0] throwNpe
0 ATHROW

Namely, the check and potential throw for z!! would be optimized away.

compiler/testData/codegen/bytecodeText/deadCodeElimination/simpleConstructorNotRedundant.kt
ting-yuan: Similarly, the check and throw for y!! would be optimized away without this.

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
max-kammerer: @madsager seems it's a bug in your optimizer:
```
fun box() {
    val y: String?
    if (1 == 0) { //no nop and linenumber in bytecode        
        y = "123"
    }
    else {
        y = null
    }    
}
```

madsager: Thanks Mike, I agree. Ting-Yuan, we should work together on finding a way to fix that.

compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
max-kammerer: @ting-yuan I would prefer to add new tests with suggested changes (for this one and below). And for kt2269.kt

max-kammerer: @ting-yuan change in concat.kt is still totatly unclear for me, maybe this change is redundant after pr2132.
For primitiveToString.kt it would be best to add new test with suggested change

ting-yuan: @max-kammerer The test in concat.kt has a flew IMHO. Because the constants get folded into one, there is only one concatenation after constant folding (in both backends): "12345.06.07" + A(). Because the test looks for exactly 1 StringBuilder, it does not test whether the unnecessary StringBuilders are eliminated or not.

ting-yuan: @max-kammerer I touched kt2269.kt because :compiler:test fails and complains that kt2269.kt is a passing test disabled. So I have either to:
1. remove // IGNORE_BACKEND: JVM_IR so that it can pass falsely. I'll add kt2269-jvm-ir.kt
2. change the test as in PR to make it test what it meant to test.

Should I just go for 1, or other recommendations?

compiler/testData/codegen/box/primitiveTypes/kt2269.kt
ting-yuan: I've disabled these 2 tests and let's address it in another PR. This one is probably too large already.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: Thanks for the pointer! I updated the PR quite a bit to reuse OperationsMapGenerated.kt. Please take a look again.

compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
ting-yuan: @max-kammerer I've forked those tests with proposed changes and disabled the original tests for JVM_IR, if they fail. Basically testABC.kt is forked into testABCNotOptimizable.kt.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
ting-yuan: Done.

ting-yuan:
Hi @max-kammerer @udalov @dnpetrov,

Besides the changes to
1. tests.
2. factoring out operator names 
3. reuse OperationMapGenerated,

I also tried to get rid of descriptors. Please take a look again.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrBuiltIns.kt
```diff
@@ -215,5 +215,19 @@ class IrBuiltIns(
 
     companion object {
         val KOTLIN_INTERNAL_IR_FQN = FqName("kotlin.internal.ir")
+
+        const val OPERATOR_NAME_LESS = "less"
+        const val OPERATOR_NAME_LESS_OR_EQUAL = "lessOrEqual"
```
max-kammerer: Maybe it's best to extract const's to separate object:

```
class IrBuiltIns {
   .....
    object OperatorNames {
        const val LESS = "less"
       ....
     }
}
```
cc @udalov

compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
```diff
@@ -1168,3 +1122,96 @@ fun CompileTimeConstant<*>.isStandaloneOnlyConstant(): Boolean {
...
+    parameterTypeStr: String,
+    parameterValue: Any
+): Any? {
+    val receiverType = typeStrToCompileTimeType(receiverTypeStr)
+    val parameterType = typeStrToCompileTimeType(parameterTypeStr)
+
+    return evaluateBinaryAndCheck(name, receiverType, receiverValue, parameterType, parameterValue)
+}
+
+internal fun evaluateBinaryAndCheck(
```
max-kammerer: private

compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
```diff
@@ -1168,3 +1122,96 @@ fun CompileTimeConstant<*>.isStandaloneOnlyConstant(): Boolean {
...
+    STRING.name -> STRING
+    ANY.name -> ANY
+    else -> throw IllegalArgumentException("Unsupported type: $str")
+}
+
+fun evaluateUnary(name: String, typeStr: String, value: Any, tracer: () -> Unit = {}): Any? {
+    return evaluateUnaryAndCheck(name, typeStrToCompileTimeType(typeStr), value)
+}
+
+internal fun evaluateUnaryAndCheck(name: String, type: CompileTimeType<*>, value: Any, tracer: () -> Unit = {}): Any? {
```
max-kammerer: private

compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
```diff
@@ -1168,3 +1122,96 @@ fun CompileTimeConstant<*>.isStandaloneOnlyConstant(): Boolean {
...
+
+internal fun isZero(value: Any?): Boolean {
+    return when {
+        isIntegerType(value) -> (value as Number).toLong() == 0L
+        value is Float || value is Double -> (value as Number).toDouble() == 0.0
+        else -> false
+    }
+}
+
+internal fun typeStrToCompileTimeType(str: String) = when (str) {
```
max-kammerer: private

compiler/frontend/src/org/jetbrains/kotlin/resolve/constants/evaluate/ConstantExpressionEvaluator.kt
```diff
@@ -1168,3 +1122,96 @@ fun CompileTimeConstant<*>.isStandaloneOnlyConstant(): Boolean {
         else -> return false
     }
 }
+
+internal fun isZero(value: Any?): Boolean {
```
max-kammerer: private

ting-yuan:
Thanks for the review! @max-kammerer. Please take a look at the updated PR again.

I'm going to resolve the comments about the tests and will remove // IGNORE_BACKEND: JVM_IR for the 2 lineNumber tests once https://github.com/JetBrains/kotlin/pull/2153 is merged.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrBuiltIns.kt
```diff
@@ -251,4 +251,20 @@ class IrBuiltIns(
     companion object {
         val KOTLIN_INTERNAL_IR_FQN = FqName("kotlin.internal.ir")
     }
+
+    object OperatorNames {
+        const val OPERATOR_NAME_LESS = "less"
+        const val OPERATOR_NAME_LESS_OR_EQUAL = "lessOrEqual"
+        const val OPERATOR_NAME_GREATER = "greater"
```
max-kammerer: @ting-yuan Please remove also OPERATOR_NAME prefixes (it wasn't clearly mentioned but presented in suggested example in https://github.com/JetBrains/kotlin/pull/2111/files#r264669713)

max-kammerer:
@ting-yuan Please rebase: #2153 is merged

ting-yuan:
@max-kammerer Thanks, I've rebased and removed `// IGNORE_BACKEND: JVM_IR` from the 2 line number tests, which are fixed in #2153. Also verified again with `compiler:test`.

max-kammerer:
@ting-yuan https://github.com/JetBrains/kotlin/pull/2111/files#r265006614 is still missed

ting-yuan:
@max-kammerer Sorry, I missed that. Please take a look again.

compiler/testData/codegen/bytecodeText/constants/comparisonFalse.kt
```diff
@@ -0,0 +1,22 @@
+// TARGET_BACKEND: JVM_IR
```
max-kammerer: @ting-yuan Please add box tests for  comparisonFalse/comparisonTrue

ting-yuan: Done!

max-kammerer:
@ting-yuan What is 'Change-Id ' in  commit message?

max-kammerer:
Please also mute JS_IR backend for newly added tests:
https://teamcity.jetbrains.com/viewLog.html?buildId=2074413&tab=buildResultsDiv&buildTypeId=Kotlin_dev_Compiler

ting-yuan:
@max-kammerer Thanks for catching this. I've removed `Change-Id`s and muted the 2 box tests for JR_IR.

To ensure the quality of pull request and make the review quicker, we use [Gerrit](https://www.gerritcodereview.com/) internally to have a first-round review before sending PRs. Gerrit relies on `Change-Id` to keep track of the patch history.

max-kammerer:
@ting-yuan Many thanks!

PR #2113, opened 2019-02-13
KT-23259 False positive unchecked cast warning/quickfix result in good code turning red

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23259


t-kameyama:
Fixed tests

dimonchik0036:
Thanks!

PR #2114, opened 2019-02-14
KT-29763 False negative "Object literal can be converted to lambda" for block body function with explicit return

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29763


mglukhikh:
Merged 👍

PR #2116, opened 2019-02-15
KT-29796 Label rename refactoring does not work on label usage

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29796


Alefas:
Merged manually.

PR #2118, opened 2019-02-15
KT-29939 - Add fancy infinite loop.

hamzamuric:
https://youtrack.jetbrains.com/issue/KT-29939


ilya-g:
At this point this pull request is too hurried, please follow https://youtrack.jetbrains.com/issue/KT-27970 for updates on this feature.

PR #2119, opened 2019-02-16
KT-29099 "Convert to apply" intention is not available for a single function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29099


mglukhikh:
Merged 👍

PR #2120, opened 2019-02-16
Keyword completion: do not suggest `const` except inside top level or object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23026
https://youtrack.jetbrains.com/issue/KT-29840


t-kameyama:
https://github.com/JetBrains/kotlin/commit/8cdf9015387100003d1920bfaf86a40f8a333e8c

PR #2121, opened 2019-02-16
Make IrDeclaration.origin mutable

jimgoog:
When plugins perform various IR transforms, it sometimes becomes necessary to modify the origin of a declaration.  When the origin is a `val`, this requires creating a new declaration and copying the old values into the new declaration.  Unfortunately, this also requires applying a transform over the entire module to rewrite references from the old declaration to the new declaration, which ends up becoming a rather complex codepath for what would otherwise be a straightforward lowering transformation.

By making the `origin` field be a `var` instead of a `val`, these lowering transformations can be dramatically simplified.

cc @udalov @erokhins


dnpetrov:
We are going to rework `origin` to make it more suitable for actual use-cases.
Other than that, it's Ok to make the property mutable for now.

PR #2122, opened 2019-02-18
KT-26979 "Lambda argument inside parentheses" inspection is not reported, if function type is actual type argument, but not formal parameter type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26979


mglukhikh:
Merged 👍

PR #2123, opened 2019-02-18
KT-28838 Group by file structure doesn't work for text search in Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28838


goodwinnk:
Thank you! Your commit in master https://github.com/JetBrains/kotlin/commit/7149d4cd1ce8e2dc423ee5b70db4979d6fec6b63.

PR #2125, opened 2019-02-19
KT-29416 False positive "Redundant property getter" for `external` getter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29416


mglukhikh:
Merged 👍

PR #2126, opened 2019-02-19
Generate better code for branches based on comparisons.

madsager:
For comparison intrinsics and for instanceof checks, make
it possible to get the the stack value produced and branch
on that directly instead of materializing a boolean to
branch on from it.

That reduces code such as

```
    IF_CMPEQ L1
    CONST_0
    GOTO L2
L1: CONST_1
L2: IFEQ L3
```

to just one IF_CMP instruction.


madsager:
@max-kammerer

max-kammerer:
Thanks!

PR #2128, opened 2019-02-19
Generate synthetic functions for local functions with default values,…

punzki:
… and invoke them with the correct function descriptor.

This is accomplished by having DefaultArgumentStubGenerator transform StatementContainers (local function declarations are statements) in addition to DeclarationContainers.

Without the changes to AsmUtil, the generated functions are not invoked with the correct descriptor; the enclosing class is included in the descriptor as the dispatch receiver parameter.

This change still does not fix local functions with default values from closures (i.e., KT-5589).

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/AsmUtil.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/AsmUtil.java
@@ -302,7 +302,8 @@ public class AsmUtil {
     }
 
     public static boolean isStaticMethod(OwnerKind kind, CallableMemberDescriptor functionDescriptor) {
-        return isStaticKind(kind) ||
+        return isStaticDeclaration(functionDescriptor) ||
+               isStaticKind(kind) ||
                KotlinTypeMapper.isStaticAccessor(functionDescriptor) ||
                CodegenUtilKt.isJvmStaticInObjectOrClassOrInterface(functionDescriptor);
     }
--- a/core/descriptors/src/org/jetbrains/kotlin/name/NameUtils.kt
+++ b/core/descriptors/src/org/jetbrains/kotlin/name/NameUtils.kt
@@ -19,7 +19,7 @@ package org.jetbrains.kotlin.name
 import java.util.*
 
 object NameUtils {
-    private val SANITIZE_AS_JAVA_INVALID_CHARACTERS = "[^\\p{L}\\p{Digit}]".toRegex()
+    private val SANITIZE_AS_JAVA_INVALID_CHARACTERS = "[^\\p{L}\\p{Digit}\\$]".toRegex()
 
     @JvmStatic
     fun sanitizeAsJavaIdentifier(name: String): String {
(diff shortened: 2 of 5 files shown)


core/descriptors/src/org/jetbrains/kotlin/name/NameUtils.kt
```diff
@@ -19,7 +19,7 @@ package org.jetbrains.kotlin.name
 import java.util.*
 
 object NameUtils {
-    private val SANITIZE_AS_JAVA_INVALID_CHARACTERS = "[^\\p{L}\\p{Digit}]".toRegex()
+    private val SANITIZE_AS_JAVA_INVALID_CHARACTERS = "[^\\p{L}\\p{Digit}\\$]".toRegex()
```
punzki: Purpose of this change: LocalDeclarationsLowering uses sanitizeAsJavaIdentifier when naming the "lifted" local functions. The generated synthetic function for local function `bar` inside function `foo` is named `bar$default`. When lifted, the name should be `foo$bar$default` but without this change, `bar$default` is incorrectly sanitized and the lifted function name becomes `foo$bar_default`. This change ensures that `$` is not sanitized into `_`. ExpressionCodegen assumes that the names of the generated function end with `$default`.

compiler/backend/src/org/jetbrains/kotlin/codegen/AsmUtil.java
```diff
@@ -302,7 +302,8 @@ public static boolean isAbstractMethod(FunctionDescriptor functionDescriptor, Ow
     }
 
     public static boolean isStaticMethod(OwnerKind kind, CallableMemberDescriptor functionDescriptor) {
-        return isStaticKind(kind) ||
+        return isStaticDeclaration(functionDescriptor) ||
```
punzki: Purpose of this change: Without this, the generated functions are not invoked with the correct descriptor; the enclosing class is included in the descriptor as the dispatch receiver parameter. This causes a stack underflow error during verification.

punzki: I'm not sure if this change is safe to make since this affects the current non-IR backend, but all the tests are passing.

core/descriptors/src/org/jetbrains/kotlin/name/NameUtils.kt
punzki: I'm not sure if this change is safe to make since this affects the current non-IR backend, but all the tests are passing. There doesn't seem to be any code within the compiler relying on `$` being sanitized into `_`, but it's possible there's some code out there that does.

max-kammerer: @punzki LocalNameProvider in LocalDeclarationsLowering is totally wrong concept. It was appeared only for fast bootstrapping. In some future it would be replaced with proper logic similar to one to CodegenAnnotatingVisitor in old backend. 
Seems name related problems will be solved with gettting rid of descriptors (https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2697) and totally rewrited KotlinTypeMapper for IR.

compiler/backend/src/org/jetbrains/kotlin/codegen/AsmUtil.java
max-kammerer: It could be a bug in LocalDeclarationLowering, but i'm not sure.
In any way maybe it's best to postpone this pr for some time until new typeMapper will be appeared and KOTLIN-CR-2697 merged in master

core/descriptors/src/org/jetbrains/kotlin/name/NameUtils.kt
punzki: @max-kammerer Got it! Will hold off on this PR for now, and revisit it after that change is landed, and new type mapper is available. Is the type mapper in a different change under review, or still not in progress yet?

max-kammerer: I'm just starting it

punzki:
Abandoning in favor of #2412 which supercedes this PR

PR #2129, opened 2019-02-21
KT-30038 'Remove redundant Unit" false positive when return type is nullable Unit

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30038


mglukhikh:
Merged 👍

PR #2130, opened 2019-02-22
when: emit lookupswitch/tableswitch if possible

ting-yuan:
A lookupswitch or tableswitch can be used if all conditions are equality
checks to constants. To be more specific, it can be done if:
  1. All conditions are CALL 'EQEQ(Any?, Any?)': Boolean
  2. All types of variables involved in comparison are in the same group
     of Char/Byte/Short/Int, String or enum.
  3. All arg0 refer to the same value.
  4. All arg1 are IrConst<*>.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -688,7 +688,8 @@ class ExpressionCodegen(
 
     override fun visitWhen(expression: IrWhen, data: BlockInfo): StackValue {
         expression.markLineNumber(startOffset = true)
-        return genIfWithBranches(expression.branches[0], data, expression.type.toKotlinType(), expression.branches.drop(1))
+        val switch = SwitchGenerator(expression, data, this).generate()
+        return switch ?: genIfWithBranches(expression.branches[0], data, expression.type.toKotlinType(), expression.branches.drop(1))
     }
 
     private fun genIfWithBranches(branch: IrBranch, data: BlockInfo, type: KotlinType, otherBranches: List<IrBranch>): StackValue {
@@ -1204,7 +1205,7 @@ class ExpressionCodegen(
 
     }
 
-    private fun coerceNotToUnit(fromType: Type, fromKotlinType: KotlinType?, toKotlinType: KotlinType): StackValue {
+    internal fun coerceNotToUnit(fromType: Type, fromKotlinType: KotlinType?, toKotlinType: KotlinType): StackValue {
         val asmToType = toKotlinType.asmType
         if (asmToType != AsmTypes.UNIT_TYPE || TypeUtils.isNullableType(toKotlinType)) {
             coerce(fromType, fromKotlinType, asmToType, toKotlinType, mv)
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.jvm.codegen
+
+import org.jetbrains.kotlin.codegen.*
+import org.jetbrains.kotlin.codegen.StackValue.*
+import org.jetbrains.kotlin.codegen.`when`.SwitchCodegen.Companion.preferLookupOverSwitch
+import org.jetbrains.kotlin.ir.IrElement
+import org.jetbrains.kotlin.ir.expressions.*
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.util.dump
+import org.jetbrains.kotlin.ir.util.isTrueConst
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.org.objectweb.asm.Label
+import org.jetbrains.org.objectweb.asm.Type
+import java.util.*
+
+// TODO: eliminate the temporary variable
+class SwitchGenerator(private val expression: IrWhen, private val data: BlockInfo, private val codegen: ExpressionCodegen) {
+    private val mv = codegen.mv
+
+    // @return null if the IrWhen cannot be emitted as lookupswitch or tableswitch.
+    fun generate(): StackValue? {
+        val endLabel = Label()
+        var defaultLabel = endLabel
+        val thenExpressions = ArrayList<Pair<IrExpression, Label>>()
+        var elseExpression: IrExpression? = null
+        val allConditions = ArrayList<Pair<IrCall, Label>>()
+
+        // Parse the when structure. Note that the condition can be nested. See matchConditions() for details.
+        for (branch in expression.branches) {
+            if (branch is IrElseBranch) {
+                elseExpression = branch.result
+                defaultLabel = Label()
+            } else {
+                val conditions = matchConditions(branch.condition) ?: return null
+                val thenLabel = Label()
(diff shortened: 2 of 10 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -684,9 +682,171 @@ class ExpressionCodegen(
...
     override fun visitWhen(expression: IrWhen, data: BlockInfo): StackValue {
         expression.markLineNumber(startOffset = true)
+        val switchGen = SwitchGenerator(expression, data, intrinsics)
+        if (switchGen.match())
+            return switchGen.gen()
         return genIfWithBranches(expression.branches[0], data, expression.type.toKotlinType(), expression.branches.drop(1))
     }
 
+    // TODO: Char, Byte, Short, String, enum, eliminate the temporary variable
+    inner class SwitchGenerator(val expression: IrWhen, val data: BlockInfo, val intrinsics: IrIntrinsicMethods) {
```
udalov: Can we extract this to a separate file? (Would be nice to avoid making this ExpressionCodegen as enormous as the one in the old backend :) )

ting-yuan: Done. Please take a look again :)

max-kammerer:
@ting-yuan Do you plan to support String/Enums in separate commit?

ting-yuan:
Yes. I'm trying to see how enum is implemented in the current backend. String looks easier, although slightly different than the integers (jump v.s. hash -> jump -> compare).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -0,0 +1,181 @@
...
+        gen(subject!!, data)
+
+        // Emit either tableswitch or lookupswitch, depending on the code size.
+        //
+        // lookupswitch is 2X as large as tableswitch with the same entries. However, lookupswitch is sparse while tableswitch must
+        // enumerate all the entries in the range.
+        val caseMin = cases.first().first
+        val caseMax = cases.last().first
+        val rangeLength = caseMax - caseMin + 1L
+        if (rangeLength > 2L * cases.size) {
```
max-kammerer: @punzki Old codegen also checks range inerval:
```
if (rangeLength > 2L * labelsNumber || rangeLength > Int.MAX_VALUE) 
```
Please move this code to old codegen companion and reuse here

punzki: I think you mean @ting-yuan :-)

max-kammerer: @punzki Sorry!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -684,6 +684,9 @@ class ExpressionCodegen(
 
     override fun visitWhen(expression: IrWhen, data: BlockInfo): StackValue {
         expression.markLineNumber(startOffset = true)
+        val switchGen = SwitchGenerator(expression, data, this)
+        if (switchGen.match())
```
udalov: I'd somewhat prefer if the public API would be something like this:
```
val switch = SwitchGenerator(expression, data, this).generate()
if (switch != null) return switch
```

In this case, you wouldn't need mutable fields in `SwitchGenerator` to retain information between `match` and `gen`; the logic in `match` could be inlined into `gen` with local variables holding things like `thenExpressions ` and `elseExpression`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -0,0 +1,181 @@
...
+            // 2. All types of variables involved in comparison are in the same group of Char/Byte/Short/Int, String or enum.
+            // 3. All arg0 refer to the same value.
+            // 4. All arg1 are IrConst<*>.
+
+            if (condition.symbol != codegen.classCodegen.context.irBuiltIns.eqeqSymbol)
+                return false
+
+            // TODO: The check relies on the implementation that subject is always lowered to a local variable.
+            val candidate = condition.getValueArgument(0) as? IrGetValue ?: return false
+            subject = subject ?: candidate
```
udalov: Please consider rewriting this to avoid modification of mutable fields, as it isn't very easy to debug code operating with mutable state. For example, instead of modifying `subject` and `cases` along the way of computing whether this is an optimizable when-condition, consider making `matchConditions` return a nullable list of case + 
subject, where null would mean that this is not an optimizable when-condition. Something like this:
```
private data class ConditionInfo(val case: Pair<Int, Label>, val subject: IrGetValue)

...

private fun matchConditions(condition: IrExpression, thenLabel: Label): List<ConditionInfo>? { ... }
```

At the call site, we can then return if any `matchConditions` returned null, ensure that all `subject`s of these conditions are the same, and flatten the results to pass to `gen`. An example advantage here is that if I'm debugging this code I can easily go backwards and forwards, re-evaluate `matchConditions` on anything I like without side effects, etc.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -0,0 +1,181 @@
...
+            val case = condition.getValueArgument(1) as? IrConst<*> ?: return false
+            if (case.kind != IrConstKind.Int)
+                return false
+
+            cases.add(Pair(case.value as Int, thenLabel))
+
+            return true
+        } else if (condition is IrWhen && condition.origin == IrStatementOrigin.WHEN_COMMA) {
+            // WHEN_COMMA should always be a Boolean.
+            assert(condition.type.isBoolean())
```
udalov: Please move the comment to the assertion message and add context-specific info, e.g. something like
```
assert(condition.type.isBoolean() { "WHEN_COMMA should always be a Boolean: ${condition.dump()}" }
```

(We try to avoid asserts without messages in our project as much as is reasonable.)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
ting-yuan: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Done. matchConditions now only get a list of IrCall without having side effects. The result can also be reused, e.g., for String and Enum.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
udalov: Thanks!

max-kammerer:
@ting-yuan Many thanks!

PR #2131, opened 2019-02-22
KT-29666: Updated Deprecated Java Quickfix to migrate RetentionPolicy arguments

inktomi:
I did this as straight forward as I know how, trying to keep it generalized to cover all the potential arguments. I feel like there **might be a better way** to do the checks, comparing PSI elements and their types rather than strings. 

If anyone can point me to a helpful bit of documentation, I'd be more than happy to update this PR to remove the string manipulation. 

I only attempted to cover `@Retention` here because the mapping for `@Target` is much more convoluted and it really seems like automatic mapping of those types can introduce a lot of bugs for people.


idea/src/org/jetbrains/kotlin/idea/quickfix/DeprecatedJavaAnnotationFix.kt
```diff
@@ -40,6 +42,35 @@ internal class DeprecatedJavaAnnotationFix(element: KtAnnotationEntry,
         file.importList?.add(psiFactory.createImportDirective(ImportPath(annotationFqName, false, null)))
     }
 
+    private fun updateAnnotationArgument(file: KtFile, args: List<KtValueArgument>): String {
+        return args.joinToString(",") {
+            val type = it.text.split('.')
```
inktomi: I was trying to find a better way to do this using the PSI elements, but nothing seemed to work. I am relatively new at PSI though, so I'm very much aware that I could be missing something obvious. Please point it out if you know!

I need to handle things like `@Retention(RetentionPolicy.SOURCE)` but also `@Target(CLASS, PACKAGE, METHOD)` in a way that lets me know the type of the arguments, so that I can map the old Java arguments to their Kotlin versions.

mglukhikh: You could use `ktAnnotationEntry.analyze()` to get `BindingContext`, and then use `context[BindingContext.annotation, ktAnnotationEntry]` to get `AnnotationDescriptor` which contains all necessary information. Of course it's much better than parse PSI text manually. Note: to do all of this, you should pass the wholy `ktAnnotationEntry` to quick-fix constructor, not just arguments and annotation FQ name.

mglukhikh: Sorry for late review.

inktomi: I will look at this, sorry! I totally missed the reply here. Apologies.

inktomi:
I'm working on this, sorry for the slow progress. I had a vacation in the middle of this.

mglukhikh:
Thank you for your contribution. There are some minor problems (for example, you often state a local variable type explicitly when it's not required). Also, I advice you to avoid using `!!` without major reason. I think it will be easier for me to finish this change myself.

mglukhikh:
Fixed & merged manually. NB: tests in the raw PR did not work.

inktomi:
Interesting - they passed for me locally using gradle to run them. 

Sorry for the local variable types and such - this is very much an artifact of my learning PSI while working on this. It was helping me to keep things straight in my mind to be super explicit about what was going on.

PR #2133, opened 2019-02-22
Make parcelize a non-experimental feature

JakeWharton:
It was already being added to the feature list by default, but was only enabled when isExperimental was true.


JakeWharton:
Merged (with some much-appreciated help) separately.

PR #2137, opened 2019-02-24
Set target backend of BytecodeTextTestGenerated to JVM

ting-yuan:
as BytecodeTextTestGenerated is only available to JVM.


max-kammerer:
@ting-yuan Why you need this change?

ting-yuan:
There seems to be no way to disable a single test for the JVM backend in BytecodeTextTestGenerated. I tried IGNORE_BACKEND and TARGET_BACKEND and neither of them work. After tracing the code a little bit, it looks like that the default target backend [ANY keeps those 2 flags from working](https://github.com/JetBrains/kotlin/blob/master/compiler/tests-common/tests/org/jetbrains/kotlin/test/InTextDirectivesUtils.java#L233).

max-kammerer:
I've merged pr.
Js tests are generated via another generator - GenerateJsTestsKt,
so all tests (except JVM_IR) could be annotated with JVM

max-kammerer:
Thanks

PR #2138, opened 2019-02-25
KT-29927 Missing "Import members from" intention with type check operator in `when` branch

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29927


dimonchik0036:
Thanks!

PR #2139, opened 2019-02-25
KT-30010 Introduce alternative quick-fixes for `map[key]!!`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30010


mglukhikh:
Merged 👍

PR #2140, opened 2019-02-25
Fix quick fix tests for "Convert to scope function"

t-kameyama:
Fix for https://github.com/JetBrains/kotlin/pull/2119


mglukhikh:
I already fixed these tests myself.

mglukhikh:
BTW there are also five tests in "QuickFixMultiFile" group broken in the same way.

PR #2143, opened 2019-02-26
KT-30128 Change Signature should move lambda outside of parentheses if the arguments are reordered so that the lambda goes last

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30128


klunnii:
Thank you! Merged,

PR #2144, opened 2019-02-27
KT-30123 Add intention to replace isEmpty/isNotEmpty method negation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30123


dimonchik0036:
Should we change also [ReplaceNegatedIsEmptyWithIsNotEmpty.html](https://github.com/JetBrains/kotlin/blob/master/idea/resources/inspectionDescriptions/ReplaceNegatedIsEmptyWithIsNotEmpty.html)?

dimonchik0036:
Thanks!

PR #2146, opened 2019-02-27
KT-30082 False positive "redundant `.let` call" for lambda functions stored in nullable references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30082


mglukhikh:
Merged 👍

PR #2147, opened 2019-02-28
KT-29038 Autocomplete "suspend" into "suspend fun" at top level and class level (except in kts top level)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29038


goodwinnk:
Thank you! Your commit was cherry-picked as https://github.com/JetBrains/kotlin/commit/dfd39478564459406c639d9095f18e1b0c2ab704 with a small fix in scripts check https://github.com/JetBrains/kotlin/commit/dfd39478564459406c639d9095f18e1b0c2ab704#diff-55ae7e03e0a32466e3603dba94628d58R104. Please use `isScript()` check instead of direct check of file extension. Also it's worth to check if some string literal is already declared as constant (`STD_SCRIPT_EXT` could be used).

PR #2149, opened 2019-02-28
Introduce support optional j2v8-backed interop for tests

Schahen:
I've checked `quickTest --tests=*.Ir*Long*` command and all tests are passing. For a bigger set some tests are failing both for nashorn and j2v8 but i think the same happens in master ;)

Also - j2v8 are not enabled so far.
I've just wanted to have a logically separate branch where we introduce interop after all.


js/js.tests/test/org/jetbrains/kotlin/js/test/interop/InteropEngine.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.js.test.interop
+
+interface InteropEngine {
```
romanart: JDK uses `ScriptEngine` for such things. I'd rather keep this name

romanart: Maybe one of `KotlinScriptEngine` or `TestScriptEngine`

js/js.tests/test/org/jetbrains/kotlin/js/test/InteropJsTestChecker.kt
```diff
@@ -5,65 +5,61 @@
...
+import org.jetbrains.kotlin.js.test.interop.InteropNashorn
+import org.jetbrains.kotlin.js.test.interop.InteropV8
 import org.junit.Assert
-import javax.script.Invocable
-import javax.script.ScriptEngine
 
-fun createScriptEngine(): ScriptEngine =
-    // TODO use "-strict"
-    NashornScriptEngineFactory().getScriptEngine("--language=es5", "--no-java", "--no-syntax-extensions")
+private val USE_J2V8_INTEROP_FOR_JS_TESTS = java.lang.Boolean.getBoolean("org.jetbrains.kotlin.use.j2v8.interop.for.js.tests")
```
romanart: Could you please move this flag into `KotlinTestUtil.java`? I think keeping such flags in one specific place is good.

js/js.tests/test/org/jetbrains/kotlin/js/test/InteropJsTestChecker.kt
```diff
@@ -5,65 +5,61 @@
...
-            testPackageName.contains(".") ->
-                testPackageName.split(".").fold(testModule) { p, part -> p[part] as ScriptObjectMirror }
-            else ->
-                testModule[testPackageName]!!
-        }
+): String? {
+    var script = when {
+        withModuleSystem -> BasicBoxTest.KOTLIN_TEST_INTERNAL + ".require('" + testModuleName!! + "')"
+        testModuleName === null -> "this"
+        else -> testModuleName
```
romanart: Shouldn't it be `get(testModuleName)`?

js/js.tests/test/org/jetbrains/kotlin/js/test/InteropJsTestChecker.kt
Schahen: @romanart in that case I'll  have to make it (unlike other flags declared there) public or come up with some contrived  way to pass it to InteropJsTestChecker (who we'll call it btw - I mean ScriptEngineTestChecker sounds slightly off).

If we are ok with this - sure - let me do it.

romanart: It is also possible to pass engine type as parameter from outside

romanart: My point is to keep all flags in one place

js/js.tests/test/org/jetbrains/kotlin/js/test/InteropJsTestChecker.kt
```diff
@@ -5,65 +5,61 @@
...
-    f: ScriptEngine.() -> Any?
+    f: InteropEngine.() -> Any?
 ): Any? {
     return try {
         this.f()
     } finally {
         for (key in globalObject.keys) {
-            globalObject[key] = originalState[key] ?: ScriptRuntime.UNDEFINED
+            if (originalState[key] == null) {
+                globalObject.remove(key)
```
anton-bannykh: If I recall correctly removing globals doesn't really work with Nashorn. At least that was the case at some point and that's why it was being set to undefined instead of being removed.

Also does removing globals work correctly in v8? (Just wondering, it probably does, but that's worth checking =))

js/js.tests/test/org/jetbrains/kotlin/js/test/InteropJsTestChecker.kt
```diff
@@ -5,65 +5,54 @@
...
+    var script = when {
+        withModuleSystem -> BasicBoxTest.KOTLIN_TEST_INTERNAL + ".require('" + testModuleName!! + "')"
+        testModuleName === null -> "this"
+        else -> testModuleName
+    }
 
-    return (this as Invocable).invokeMethod(testPackage, testFunctionName)
-}
+    if (testPackageName !== null) {
+        script += "[$testPackageName]"
```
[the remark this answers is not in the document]
anton-bannykh: Also shouldn't it be `"[\"$testPackageName\"]"` for a simple package (e.g. `foo`)?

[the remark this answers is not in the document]
Schahen: now it's:
```
 if (testPackageName !== null) {
        script += ".$testPackageName"
    }
```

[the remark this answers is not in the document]
anton-bannykh: :+1:

bashor:
@Schahen @romanart could we close this?

shabunc:
This one can be closed I guess

PR #2151, opened 2019-03-01
Implement property references in JVM_IR

pyos:
This change fixes 192 tests. Not implemented:
  * references to plain Java fields;
  * getDelegate.

Also, a few of the property reference tests seem to be failing due to unrelated bugs.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+        val kProperties = mutableMapOf<CallableDescriptor, Pair<IrExpression, Int>>()
+        // TODO: annotate as shared-immutable? The old backend does not do so.
+        val kPropertiesField = WrappedFieldDescriptor(Annotations.create(listOf())).let { descriptor ->
+            IrFieldImpl(
+                UNDEFINED_OFFSET, UNDEFINED_OFFSET,
+                DECLARATION_ORIGIN_KPROPERTIES_FOR_DELEGATION,
+                IrFieldSymbolImpl(descriptor),
+                "\$delegatedProperties".synthesizedName,
+                kPropertiesFieldType,
+                Visibilities.PRIVATE,
```
max-kammerer: In old backend properties are 'final static synthetic'
Syntheticness could be specified via IrDeclarationOriginImpl("KPROPERTIES_FOR_DELEGATION", true)
`TODO` in such case would be redundant or I miss smth?

pyos: Haven't noticed that IrDeclarationOriginImpl has an additional argument for syntheticness, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
```diff
@@ -0,0 +1,236 @@
...
+    override fun lower(irClass: IrClass) {
+        // Maintain an array of `KProperty` instances, one per delegated property, as the delegate target is allowed to use identity
+        // comparison of these references for differentiating properties.
+        val kProperties = mutableMapOf<CallableDescriptor, Pair<IrExpression, Int>>()
+        val kPropertiesField = WrappedFieldDescriptor(Annotations.create(listOf())).let { descriptor ->
+            IrFieldImpl(
+                UNDEFINED_OFFSET, UNDEFINED_OFFSET,
+                DECLARATION_ORIGIN_KPROPERTIES_FOR_DELEGATION,
+                IrFieldSymbolImpl(descriptor),
+                "\$delegatedProperties".synthesizedName,
```
max-kammerer: JvmAbi.DELEGATED_PROPERTIES_ARRAY_NAME

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
max-kammerer: Please add TODO that visibility should be package local,
or maybe best switch it temporary to INTERNAL (some inliner test should be passed)  with same TODO

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
```diff
@@ -0,0 +1,236 @@
...
+
+internal class PropertyDelegationLowering(val context: JvmBackendContext) : ClassLoweringPass {
+    private object DECLARATION_ORIGIN_KPROPERTIES_FOR_DELEGATION :
+        IrDeclarationOriginImpl("KPROPERTIES_FOR_DELEGATION", true)
+
+    private val plainJavaClass =
+        context.getIrClass(FqName("java.lang.Class")).owner
+
+    private val reflectionClass =
+        context.getIrClass(FqName("kotlin.jvm.internal.Reflection")).owner
```
max-kammerer: AsmType.REFLECTION

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    private val getOrCreateKotlinClass =
+        reflectionClass.declarations.find { it.descriptor.name == Name.identifier("getOrCreateKotlinClass") }
+
+    private val getOrCreateKotlinPackage =
+        reflectionClass.declarations.find { it.descriptor.name == Name.identifier("getOrCreateKotlinPackage") }
+
+    private fun propertyTypeAndFactory(prefix: String, i: Int) =
```
max-kammerer: Would it simpler to use AsmTypes.PROPERTY_REFERENCE_IMPL/MUTABLE_PROPERTY_REFERENCE_IMPL ?

pyos: Probably not, as these are not IR objects, I see no obvious way to used them to obtain an IrFunction referencing the constructor, and there's also the factory method anyway.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
pyos: Can you clarify, please? Do you mean I should replace `"kotlin.jvm.internal.Reflection"` with `AsmTypes.REFLECTION.replace("/", ".")`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
```diff
@@ -0,0 +1,236 @@
...
+                Visibilities.PRIVATE,
+                isFinal = true,
+                isExternal = false,
+                isStatic = true
+            ).apply { descriptor.bind(this) }
+        }
+
+        irClass.transformChildrenVoid(object : IrElementTransformerVoidWithContext() {
+            override fun visitPropertyReference(expression: IrPropertyReference): IrExpression =
+                // This actually generates KProperties for all property references, not just delegated ones.
```
max-kammerer: Property references should not be processed as delegated ones: anonymous class for them should be generated in CallableReferenceLowering.kt to support non-reflecsive get/set/invoke

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
max-kammerer: It's more tricky than I though:
JvmClassName.byInternalName(AsmTypes.REFLECTION).fqNameForClassNameWithoutDollars

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt
pyos: That makes sense, but I don't understand why not do that for delegated references as well?

max-kammerer: Do you mean this case:
```
val z by lazy  { 1 } // propertyMetadata

fun main() {
    ::z.get()     //new anonymous class  should be generated
}
``` ?

Comment (pyos) [reply compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt]: As I understand it, a delegated property

```
class C {
    val x by Delegate()
}
```

is simply transformed into something that resembles this:

```
class C {
    val x$delegate = Delegate()
    val x
        get() = x$delegate.getValue(this, C::x)
        set(val: T) = x$delegate.setValue(this, C::x, val)
}
```

and the code in this lowering is required to generate an appropriate KProperty in place of `C::x`. So my question is, why handle `C::x` here and a reference to a non-delegated property elsewhere differently? Seems to me like both can use a subclass of KProperty in which overrides `get` and `set` call `C::getX` and `C::setX` directly instead of going through reflection. (Of course, calling `get`/`set` on a KProperty passed to `getValue`/`setValue` is of rather limited usefulness, so the performance of that might not matter, but still.)

Comment (max-kammerer) [reply compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt]: > Of course, calling get/set on a KProperty passed to getValue/setValue is of rather limited usefulness, so the performance of that might not matter, but still.

Yes, it's useless (and in most case will produce SOE)
Reflection here is used to avoid a lot of anonymous class generation.

Please also note that there are two strategy for local delegation properties: 
```
object O {
    operator fun getValue(a: Any?, b: Any?): String = "OK"
}

fun noInline() {
    val z by O //same as z$delegate.getValue(this, ::z)
}

inline fun InlineFoo() {
    val b by O // after inline access to $$delegatedProperties in most cases are lost so anonymous object as for general CallableReference is generated
}
```

File: compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/DelegationLowering.kt [hunk 6]
```diff
@@ -0,0 +1,236 @@
...
+                while (container is IrFunction)
+                    container = container.parent
+                if (container !is IrClass)
+                    // Package-level properties are in an implicit class corresponding to the entire file.
+                    throw AssertionError("reference to a property of what?")
+
+                val boundReceivers = listOf(expression.dispatchReceiver, expression.extensionReceiver).count { it != null }
+                val needReceivers =
+                    listOf(getter.owner.dispatchReceiverParameter, getter.owner.extensionReceiverParameter).count { it != null }
+                val (propertyReference, factoryMethod) = if (mutable)
```
max-kammerer: Maybe: 
```
(if (mutable) mutableProperties else properties)[needReceivers - boundReceivers]
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+        expression.getter?.owner?.let {
+            val boundReceivers = listOf(expression.dispatchReceiver, expression.extensionReceiver).count { it != null }
+            val needReceivers = listOf(it.dispatchReceiverParameter, it.extensionReceiverParameter).count { it != null }
+            propertyReferenceKind(expression.setter?.owner?.visibleFrom(scope) ?: false, needReceivers - boundReceivers)
+        } ?: expression.field?.owner?.let {
+            propertyReferenceKind(!it.isFinal, if (it.isStatic || expression.dispatchReceiver != null) 0 else 1)
+        } ?: throw AssertionError("property has no getter and no field")
+
+    override fun lower(irClass: IrClass) {
+        val kProperties = mutableMapOf<Pair<CallableDescriptor, Boolean /* reflected implementation */>, Pair<IrExpression, Int>>()
```
max-kammerer: Please use data classes here ...

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+        val kPropertyClasses = mutableMapOf<Triple<CallableDescriptor, Boolean, Boolean /* bound receivers */>, IrClass>()
```
max-kammerer: and here

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+
+                    body = context.createIrBuilder(symbol).irBlockBody(startOffset, endOffset) {
+                        val dispatchReceiverPair = dispatchReceiverField?.let { it to buildReceiverParameter(it) }
+                        val extensionReceiverPair = extensionReceiverField?.let { it to buildReceiverParameter(it) }
+                        val superArgs = if (dispatchReceiverPair == null) 0 else 1
+                        +irDelegatingConstructorCall(superClass.constructors.single { it.valueParameters.size == superArgs }).apply {
+                            // TODO: what about the extension receiver?
+                            dispatchReceiverPair?.let { putValueArgument(0, irGet(it.second)) }
+                        }
+                        dispatchReceiverPair?.let { +irSetField(irGet(referenceThis), it.first, irGet(it.second)) }
```
max-kammerer: Reuse 'receiver' from base class

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrDelegateDescriptor.kt
```diff
@@ -95,6 +96,9 @@ class IrPropertyDelegateDescriptorImpl(
 ) :
     IrDelegateDescriptorBase(
         correspondingProperty.containingDeclaration,
+        // Inherit the source node so that the parent class of package-level
```
max-kammerer: I would prefer to patch KotliTypeMapper.internalNameForPackageMemberOwner
```
if (isIrBackend && directMember is IrPropertyDelegateDescriptor) {
                return internalNameForPackageMemberOwner(directMember.correspondingProperty, publicFacade, isIrBackend)
            }
```

With some time there would be IR-specific KotliTypeMapper without any hacks

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt
```diff
@@ -42,6 +42,7 @@ interface IrDeclarationOrigin {
     object IR_BUILTINS_STUB : IrDeclarationOriginImpl("IR_BUILTINS_STUB")
     object BRIDGE : IrDeclarationOriginImpl("BRIDGE", isSynthetic = true)
     object BRIDGE_SPECIAL: IrDeclarationOriginImpl("BRIDGE_SPECIAL")
+    object GENERATED_PROPERTY_REFERENCE : IrDeclarationOriginImpl("GENERATED_PROPERTY_REFERECE", isSynthetic = true)
```
udalov: Please move to `JvmLoweredDeclarationOrigin`

udalov:
@pyos Please configure `git config user.name` to show your first/last name

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
+/*
+ * Copyright 2019 Google LLC. Use of this source code is governed by the Apache 2.0 license
```
udalov: According to our [license](https://github.com/JetBrains/kotlin/tree/master/license), all code in this repository is copyright by JetBrains. Therefore we need our default copyright on all sources. Is there any particular reason why you've changed this to Google LLC?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+        visibility == Visibilities.LOCAL || visibility.isVisible(Visibilities.ALWAYS_SUITABLE_RECEIVER, descriptor, scope.scopeOwner)
+
+    private val plainJavaClass =
+        context.getIrClass(FqName("java.lang.Class")).owner
+
+    private val reflectionClass =
+        context.getIrClass(FqName("kotlin.jvm.internal.Reflection")).owner
+
+    private val getOrCreateKotlinClass =
+        reflectionClass.declarations.find { it.descriptor.name == Name.identifier("getOrCreateKotlinClass") }
```
udalov: Please use `reflectionClass.functions` here and below, and remove casts in `buildReflectedContainerReference`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+    private fun propertyReferenceKind(mutable: Boolean, i: Int) = PropertyReferenceKind(
+        context.getIrClass(FqName("kotlin.jvm.internal.${if (mutable) "Mutable" else ""}PropertyReference${i}")),
+        context.getIrClass(FqName("kotlin.jvm.internal.${if (mutable) "Mutable" else ""}PropertyReference${i}Impl")),
+        reflectionClass.declarations.find { it.descriptor.name == Name.identifier(if (mutable) "mutableProperty$i" else "property$i") }
+                as IrFunction
+    )
+
+    private fun propertyReferenceKindFor(expression: IrMemberAccessExpression, scope: Scope): PropertyReferenceKind =
+        expression.getter?.owner?.let {
+            val boundReceivers = listOf(expression.dispatchReceiver, expression.extensionReceiver).count { it != null }
```
udalov: `listOfNotNull(...).size` here and below (will get rid of the IDE inspection about `it`)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,347 @@
...
+    private class PropertyReferenceKind(
+        val interfaceSymbol: IrClassSymbol,
+        val reflectedSymbol: IrClassSymbol,
+        val wrapper: IrFunction
+    )
+
+    private fun propertyReferenceKind(mutable: Boolean, i: Int) = PropertyReferenceKind(
+        context.getIrClass(FqName("kotlin.jvm.internal.${if (mutable) "Mutable" else ""}PropertyReference${i}")),
+        context.getIrClass(FqName("kotlin.jvm.internal.${if (mutable) "Mutable" else ""}PropertyReference${i}Impl")),
+        reflectionClass.declarations.find { it.descriptor.name == Name.identifier(if (mutable) "mutableProperty$i" else "property$i") }
```
udalov: `reflectionClass.functions` -> no `as IrFunction` needed

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,336 @@
...
+
+    private val IrMemberAccessExpression.signature: String
+        // Plain Java fields do not have a getter, but can be referenced nonetheless.
+        get() = getter?.owner?.let { context.state.typeMapper.mapSignatureSkipGeneric(it.descriptor).toString() }
+            ?: TODO("plain Java field signature")
+
+    private val IrMemberAccessExpression.symbol: IrSymbol
+        get() = getter?.owner?.symbol ?: field!!.owner.symbol
+
+    private fun IrFunction.visibleFrom(scope: Scope): Boolean =
```
udalov: I think we should avoid doing complicated logic like this in the compiler backend. It's the frontend's job to resolve everything down to the specific types that the backend should use in code generation, including whether to use `KMutableProperty` or `KProperty` as a property reference superclass.

Currently the frontend already infers this information in `DoubleColonExpressionResolver.createKCallableTypeForReference` and stores it into `BindingContext.VARIABLE` binding context slice (at `DoubleColonExpressionResolver.bindPropertyReference`). It seems that `psi2ir` loses this information when transforming `KtCallableReferenceExpression` to `IrCallableReference` in `ReflectionReferencesGenerator.generateCallableReference`. Could you please look into obtaining that information there and changing the created `IrPropertyReference` to have no setter if it's not visible? I.e. get the "synthetic" local variable via `getOrFail(BindingContext.VARIABLE, ktCallableReference)` and only generate a setter in the property reference if that variable is mutable.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,336 @@
...
+                    }
+                    irCall(arrayItemGetter).apply {
+                        dispatchReceiver = irGetField(null, kPropertiesField)
+                        putValueArgument(0, irInt(index))
+                    }
+                }
+            }
+
+            // Create an instance of KProperty that uses Java reflection to locate the getter and the setter. This kind of reference
+            // does not support local variables or bound receivers (e.g. `Class().field`) and is slower, but takes up less space.
```
udalov: Do you mean `Class()::field` in the comment?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
```diff
@@ -33,6 +33,7 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object JVM_OVERLOADS_WRAPPER : IrDeclarationOriginImpl("JVM_OVERLOADS_WRAPPER")
     object SYNTHETIC_METHOD_FOR_PROPERTY_ANNOTATIONS :
         IrDeclarationOriginImpl("SYNTHETIC_METHOD_FOR_PROPERTY_ANNOTATIONS", isSynthetic = true)
+    object GENERATED_PROPERTY_REFERENCE : IrDeclarationOriginImpl("GENERATED_PROPERTY_REFERECE", isSynthetic = true)
```
udalov: There's a typo in the string... :) REFERECE -> REFERENCE

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
pyos: Seems like the type is the *only* place where the frontend records mutability, which is not exactly convenient as the information about subtyping is not available afterwards, and checking against specific types seems like a hack anyway. Would it be ok to modify the frontend to also set `isVar` of the synthetic local variable accordingly?

udalov: I think it should be fine. That local variable is only needed for backends anyway, so I hope no other code relies on it being a `val` right now.

compiler/frontend/src/org/jetbrains/kotlin/types/expressions/DoubleColonExpressionResolver.kt
```diff
@@ -777,17 +779,26 @@ class DoubleColonExpressionResolver(
     }
 
     companion object {
+        private fun receiverTypeFor(descriptor: CallableDescriptor, lhs: DoubleColonLHS?): KotlinType? =
+            (descriptor.extensionReceiverParameter ?: descriptor.dispatchReceiverParameter)?.let { (lhs as? DoubleColonLHS.Type)?.type }
```
udalov: IMHO an `if` would be simpler here but it's minor
```
if (descriptor.extensionReceiverParameter != null && descriptor.dispatchReceiverParameter != null && lhs is DoubleColonLHS.Type)
    lhs.type
else null
```

udalov: IMHO an `if` was simpler (in the original version) but it's minor

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,329 @@
...
+            val boundReceivers = listOfNotNull(expression.dispatchReceiver, expression.extensionReceiver).size
+            val needReceivers = listOfNotNull(it.dispatchReceiverParameter, it.extensionReceiverParameter).size
+            // PropertyReference1 will swap the receivers if bound with the extension one, and PropertyReference0
+            // has no way to bind two receivers at once.
+            if (boundReceivers == 2 || (expression.extensionReceiver != null && needReceivers == 2))
+                TODO("property reference with 2 receivers")
+            propertyReferenceKind(expression.setter != null, needReceivers - boundReceivers)
+        } ?: expression.field?.owner?.let {
+            propertyReferenceKind(!it.isFinal, if (it.isStatic || expression.dispatchReceiver != null) 0 else 1)
+        } ?: throw AssertionError("property has no getter and no field")
```
udalov: Please add `expression.dump()` to the message

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,329 @@
...
+                    }
+                    call.extensionReceiver = call.symbol.owner.extensionReceiverParameter?.let {
+                        if (expression.extensionReceiver != null)
+                            irGetField(irGet(referenceThis), receiverField)
+                        else
+                            irImplicitCast(irGet(valueParameters[index++]), it.type)
+                    }
+                }
+
+                buildOverride(superClass.properties.single { it.name == Name.identifier("name") }.getter!!) {
```
udalov: Please replace `it.name == Name.identifier("xxx")` with `it.name.asString() == "xxx"`

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/ReflectionReferencesGenerator.kt
```diff
@@ -84,27 +82,27 @@ class ReflectionReferencesGenerator(statementGenerator: StatementGenerator) : St
...
+    ): IrCallableReference {
+        val startOffset = ktElement.startOffsetSkippingComments
+        val endOffset = ktElement.endOffset
+        return when (callableDescriptor) {
+            is FunctionDescriptor -> {
+                val symbol = context.symbolTable.referenceFunction(callableDescriptor.original)
+                generateFunctionReference(startOffset, endOffset, type, symbol, callableDescriptor, typeArguments, origin)
+            }
+            is PropertyDescriptor -> {
+                val mutable = get(BindingContext.VARIABLE, ktElement)?.isVar ?: true
```
udalov: I suppose we could use `getOrFail` here

pyos: The frontend doesn't actually generate fake variables for references used in delegated properties' accessors, so `getOrFail` fails on those.

udalov: Oh, you're right

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,350 @@
...
+            return current
+        }
+
+    // TODO: remove code duplication with CallableReferenceLowering
+    private val IrMemberAccessExpression.parentJavaClassReference
+        get() = IrClassReferenceImpl(
+            startOffset, endOffset,
+            plainJavaClass.defaultType,
+            plainJavaClass.symbol,
+            CrIrType(context.state.typeMapper.mapOwner(propertyContainerChild!!.descriptor))
```
max-kammerer: mapImplementationOwner: $$delegatedProperties are always located in DefaultImpls, not in interface

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -0,0 +1,350 @@
...
+    private val arrayItemGetter =
+        context.ir.symbols.array.owner.functions.single { it.name.asString() == "get" }
+
+    private val kPropertyType =
+        context.getIrClass(FqName("kotlin.reflect.KProperty")).owner.defaultType
+
+    private val kPropertiesFieldType =
+        context.ir.symbols.array.createType(false, listOf(makeTypeProjection(kPropertyType, Variance.OUT_VARIANCE)))
+
+    // Return some declaration the parent of which is the containing class/package for the referenced property.
```
max-kammerer: Maybe best to move this lowering after LocalDeclarationsLowering?

pyos: Right now I've placed it before PropertiesToFieldsLowering because otherwise support for Java fields would also be a requirement for some Kotlin properties. Once that's implemented, maybe there's a more optimal place in the pipeline.

pyos: Added a TODO about that.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
max-kammerer: @pyos Even with mapImplementationOwner $$delegatedProperties are located and initialized in interface not interface DefaultImpl. Moving this phase after 'interfacePhase' should help, but could you add some clarification about java field problem?

pyos: The problem is that for reflection to work correctly, getSignature should return the signature of a property's getter, even if it does not actually exist (because the property is a Java field or the getter has been removed by PropertiesToFieldsLowering). I don't see an obvious way to do that using the current type mapper.

max-kammerer: @pyos Oh, so could you add several TODOs: about location of $$delegatedProperties and it's visibility and I will merge pr

pyos: @max-kammerer done, added the second TODO and a bunch of other comments.

max-kammerer:
@pyos Great work!

PR #2152, opened 2019-03-01
KT-30215 No "surround with null" check is suggested for an assignment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30215


mglukhikh:
Merged 👍

PR #2153, opened 2019-03-04
 JVM_IR: defer some branch optimizations to codegen.

ting-yuan:
Specifically, defer the removal of hand-written "if (true)" from
JvmBuiltinOptimizationLowering into codegen so that appropriate debug
info (and a NOP) can be inserted.

Some optimization in visitBlock in JvmBuiltinOptimizationLowering depends on the removal of the "if (true)" in visitWhen. Therefore, it cannot be deferred (unless we move the logic into visiBlock to make it even more complicated). Fortunately, the optimization in visitBlock only cares about compiler generated constructs, which is transparent to users and need no debug info. To defer the optimization for non-compiler generated constructs, cl:1260 marks "hand-written" ifs by setting origin to IF so that
* compiler generate stuffs have origin == null and can be removed in IR.
* hand-written stuffs have origin == IF and can only be removed, with correct debug info generated, in ExpressionCodegen.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -39,9 +39,7 @@ import org.jetbrains.kotlin.ir.types.IrType
 import org.jetbrains.kotlin.ir.types.isMarkedNullable
 import org.jetbrains.kotlin.ir.types.isNothing
 import org.jetbrains.kotlin.ir.types.toKotlinType
-import org.jetbrains.kotlin.ir.util.dump
-import org.jetbrains.kotlin.ir.util.isNullConst
-import org.jetbrains.kotlin.ir.util.render
+import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.IrElementVisitor
 import org.jetbrains.kotlin.resolve.DescriptorUtils
 import org.jetbrains.kotlin.resolve.DescriptorUtils.isEnumClass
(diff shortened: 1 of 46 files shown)


ting-yuan:
In the first commit, only BranchingExpressionGenerator.kt and ExpressionHelpers.kt contain "real" changes. The remainings are just corresponding changes in the expected outputs.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -695,9 +696,16 @@ class ExpressionCodegen(
         val elseLabel = Label()
         val thenBranch = branch.result
         //TODO don't generate condition for else branch - java verifier fails with empty stack
-        val elseBranch = branch is IrElseBranch
-        if (!elseBranch) {
+        // Else branch comes from "if (ALWAYS_TRUE)", where ALWAYS_TRUE can be handwritten, inferred by compiler,
+        // or a construct of IrElseBranch.
+        val isElseBranch = branch.condition.isTrueConst()
```
max-kammerer: @ting-yuan Please name variable properly: 'true' condition could also be in non-else

compiler/testData/codegen/bytecodeText/lineNumbers/ifElseConstTrue.kt
```diff
@@ -0,0 +1,8 @@
+fun foo() {
+    if (true) {
+        System.out?.println()
```
max-kammerer: Don't use print in tests

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/BranchingExpressionGenerator.kt
```diff
@@ -77,7 +77,7 @@ class BranchingExpressionGenerator(statementGenerator: StatementGenerator) : Sta
         if (irBranches.size == 1) {
```
max-kammerer: @dnpetrov Shoudn't we use IF origin  for irBranches.size > 1 below:
```
val irWhen = IrWhenImpl(ktIf.startOffsetSkippingComments, ktIf.endOffset, resultType, IrStatementOrigin.WHEN)
```

@ting-yuan Please also add next test:
```
fun cond() = false

fun bar() {}

fun foo() {
    if (cond()) {
        bar()
    } else if (true) {
        bar()
    } else {
        bar()
    }
}
// 1 LINENUMBER 6
// 1 LINENUMBER 8
```

max-kammerer: Similiar tests with 'false' would be also appreciated

ting-yuan:
Hi @max-kammerer , please take a look again. Changes are:
1. handled `if (false)` as well and added test cases.
2. refactored `genIfWithBranches` a little bit.
3. renamed the tests for true.
4. removed `println` from the tests. In fact, I copied them from existing tests: `ifElse.kt` and `singleThen.kt`. I'm not sure why they use `println` so I didn't touch them. Let me know if you'd like them be replaced, too.

max-kammerer:
@ting-yuan Many Thanks!

PR #2154, opened 2019-03-05
KT-29812 False positive for HasPlatformType with member extension on 'dynamic'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29812


mglukhikh:
Merged 👍

PR #2155, opened 2019-03-05
Avoid true condition generation for do while loops

sfs:
...since they confuse the Java verifier.

The fix already exists for while loops, this is the same thing for do...while loops.


max-kammerer:
@sfs Many thanks!

PR #2156, opened 2019-03-05
KT-25272: Unused expression as last expression of normal function should have quickfix to add "return"

devbridie:
Related YouTrack issue: https://youtrack.jetbrains.com/issue/KT-25272

This PR adds two quickfixes and tests for two types of diagnostic message: `NO_RETURN_IN_FUNCTION_WITH_BLOCK_BODY` and `UNUSED_EXPRESSION`, both of which are shown in this example program:

```kotlin
fun test(): Boolean {
    true // UNUSED_EXPRESSION reported here
} // NO_RETURN_IN_FUNCTION_WITH_BLOCK_BODY reported here
```
Both quick fixes will add `return` before `true` in this case.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/AddReturnToUnusedLastExpressionInFunctionFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/AddReturnToUnusedLastExpressionInFunctionFix.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.quickfix
+
+import com.intellij.codeInsight.intention.IntentionAction
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import org.jetbrains.kotlin.diagnostics.Diagnostic
+import org.jetbrains.kotlin.diagnostics.Errors
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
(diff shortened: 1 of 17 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/AddReturnToUnusedLastExpressionInFunctionFix.kt
```diff
@@ -0,0 +1,57 @@
...
+        val exprType = expr.getType(context) ?: return false
+        val function = expr.parent.parent as? KtNamedFunction ?: return false
+        val functionReturnType = function.resolveToDescriptorIfAny()?.returnType ?: return false
+        if (!exprType.isSubtypeOf(functionReturnType)) return false
+
+        return true
+    }
+
+    private fun KtExpression.isLastStatementInFunctionBody(context: BindingContext): Boolean {
+        if (!this.isUsedAsStatement(context)) return false
```
dimonchik0036: Can you give an example where this check is needed? Looks redundant

idea/src/org/jetbrains/kotlin/idea/quickfix/AddReturnToUnusedLastExpressionInFunctionFix.kt
```diff
@@ -0,0 +1,57 @@
...
+        val last = body.statements.lastOrNull() ?: return false
+        return last === this
+    }
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val element = element ?: return
+        element.replace(KtPsiFactory(project).createExpression("return ${element.text}"))
+    }
+
+    companion object Factory : KotlinIntentionActionsFactory() {
```
dimonchik0036: Same as above

devbridie:
Updated.

dimonchik0036:
Thanks!

PR #2158, opened 2019-03-07
KT-30322 Memory leak in CompilationSourceSetUtil

morj:


libraries/tools/kotlin-gradle-plugin-api/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinSourceSet.kt
```diff
@@ -10,6 +10,7 @@ import org.gradle.api.Named
 import org.gradle.api.file.SourceDirectorySet
 
 interface KotlinSourceSet : Named, HasKotlinDependencies {
+    val id: Long
```
h0tk3y: In a single project, source set names are unique. Using the `name` as a key would be sufficient. Why a separate `id` is needed?

morj: Well, I am not a domain expert, but I can easily change this to maps of String.

morj: Fixed

h0tk3y:
Thanks!

PR #2159, opened 2019-03-07
Add intrinsics for compile-time computing trimMargin/trimIndent

JakeWharton:
These only apply when the receiver is a non-interpolated constant and the margin prefix, if specified, is also a non-interpolated constant.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/IntrinsicMethods.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/IntrinsicMethods.java
@@ -64,10 +64,10 @@ public class IntrinsicMethods {
     private final IntrinsicsMap intrinsicsMap = new IntrinsicsMap();
 
     public IntrinsicMethods(JvmTarget jvmTarget) {
-        this(jvmTarget, true);
+        this(jvmTarget, false, true);
     }
 
-    public IntrinsicMethods(JvmTarget jvmTarget, boolean shouldThrowNpeOnExplicitEqualsForBoxedNull) {
+    public IntrinsicMethods(JvmTarget jvmTarget, boolean canReplaceStdlibRuntimeApiBehavior, boolean shouldThrowNpeOnExplicitEqualsForBoxedNull) {
         intrinsicsMap.registerIntrinsic(KOTLIN_JVM, RECEIVER_PARAMETER_FQ_NAME, "javaClass", -1, JavaClassProperty.INSTANCE);
         intrinsicsMap.registerIntrinsic(KOTLIN_JVM, KotlinBuiltIns.FQ_NAMES.kClass, "java", -1, new KClassJavaProperty());
         intrinsicsMap.registerIntrinsic(KOTLIN_JVM, KotlinBuiltIns.FQ_NAMES.kClass, "javaObjectType", -1, new KClassJavaObjectTypeProperty());
@@ -142,6 +142,11 @@ public class IntrinsicMethods {
 
         declareIntrinsicFunction(FQ_NAMES.string, "plus", 1, new Concat());
         declareIntrinsicFunction(FQ_NAMES.string, "get", 1, new StringGetChar());
+
+        if (canReplaceStdlibRuntimeApiBehavior) {
+            intrinsicsMap.registerIntrinsic(TEXT_PACKAGE_FQ_NAME, FQ_NAMES.string, "trimMargin", 1, new TrimMargin());
+            intrinsicsMap.registerIntrinsic(TEXT_PACKAGE_FQ_NAME, FQ_NAMES.string, "trimIndent", 0, new TrimIndent());
+        }
 
         declareIntrinsicFunction(FQ_NAMES.cloneable, "clone", 0, CLONE);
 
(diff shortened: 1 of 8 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/Trim.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+private val TYPE_STRING = Type.getType(String::class.java)
+private class StringConstant(private val text: String) : IntrinsicCallable(TYPE_STRING, emptyList(), null, null), IntrinsicWithSpecialReceiver {
+    override fun invokeMethodWithArguments(resolvedCall: ResolvedCall<*>, receiver: StackValue, codegen: ExpressionCodegen) =
+        StackValue.constant(text, TYPE_STRING)
+}
+
+private val KtStringTemplateExpression.literalText: String
```
udalov: I think this value is actually already computed by the compiler frontend and you can get it via `(codegen.getCompileTimeConstant(resolvedCall.getReceiverExpression()) as? StringValue)?.value`. To obtain the `codegen` instance, intrinsics can override the other `toCallable` method (with 4 parameters). You can also use the fact that `null` is returned if the string is not constant to get rid of checks on interpolation

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/Trim.kt
```diff
@@ -0,0 +1,74 @@
...
+        if (receiverExpression !is KtStringTemplateExpression || receiverExpression.hasInterpolation()) {
+            return method
+        }
+
+        val text = receiverExpression.literalText.trimIndent()
+        return StringConstant(text)
+    }
+}
+
+private val TYPE_STRING = Type.getType(String::class.java)
```
udalov: `AsmTypes.JAVA_STRING_TYPE`

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/IntrinsicMethods.java
```diff
@@ -143,6 +143,9 @@ public IntrinsicMethods(JvmTarget jvmTarget, boolean shouldThrowNpeOnExplicitEqu
         declareIntrinsicFunction(FQ_NAMES.string, "plus", 1, new Concat());
         declareIntrinsicFunction(FQ_NAMES.string, "get", 1, new StringGetChar());
 
+        intrinsicsMap.registerIntrinsic(TEXT_PACKAGE_FQ_NAME, FQ_NAMES.string, "trimMargin", 1, new TrimMargin());
```
udalov: We've discussed internally the idea to apply the intrinsic unconditionally and figured that it might lead to compatibility problems in case we decide to change the behavior of `trimMargin`/`trimIndent` on some corner case in a future release. It seems that these kinds of problems will only arise if the compiler of version X is compiling code against standard library of version Y where Y > X: the user expects behavior of version Y at runtime, but will actually observe behavior of version X because that's what was executed in the compiler at compilation time.

Therefore we'd like to restrict these intrinsics to be applicable only if the version of the standard library in dependencies is _not greater_ than the version of the compiler itself. This can be achieved by checking if the API version (can be obtained via `languageVersionSettings.apiVersion` at the call site) is <= than the version of the standard library in the runtime of the compiler (can be obtained via `KotlinVersion.CURRENT`). (Sorry if this sounds confusing, I can do it myself if it's unclear, just wanted to notify you that we'd like it to work that way -- maybe you have some other related concerns.)

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/Trim.kt
JakeWharton: Fixed. And yes, it does solve those TODOs!

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/IntrinsicMethods.java
JakeWharton: Makes sense. Added this check. Had to use `ApiVersion.parse(KotlinVersion.CURRENT.toString())` since those were of different types. Let me know if there's some other way to compare these two.

compiler/backend/src/org/jetbrains/kotlin/codegen/intrinsics/IntrinsicMethods.java
```diff
@@ -64,10 +64,10 @@
     private final IntrinsicsMap intrinsicsMap = new IntrinsicsMap();
 
     public IntrinsicMethods(JvmTarget jvmTarget) {
-        this(jvmTarget, true);
+        this(jvmTarget, false, true);
     }
 
-    public IntrinsicMethods(JvmTarget jvmTarget, boolean shouldThrowNpeOnExplicitEqualsForBoxedNull) {
+    public IntrinsicMethods(JvmTarget jvmTarget, boolean canReplaceStdlibRuntimeApiBehavior, boolean shouldThrowNpeOnExplicitEqualsForBoxedNull) {
```
JakeWharton: Open to different/better names for this.

udalov:
Merged manually at 6ee987fa2e.

@JakeWharton Thank you!

PR #2160, opened 2019-03-08
KT-30228 'Convert to also/apply/run/with' intention behaves differently depending on the position of infix function call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30228


mglukhikh:
Merged 👍

PR #2161, opened 2019-03-08
KT-29799 Inspection: class with non-null self-reference as a parameter in its primary constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29799


goodwinnk:
Thank you! Pushed to master as https://github.com/JetBrains/kotlin/commit/bcfab83d1c775371e6d070661aa1649dc9c4bb00.

PR #2162, opened 2019-03-08
KT-28619 "Add braces to 'if' statement" moves end-of-line comment inside an `if` branch if statement inside `if` is block

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28619

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
@@ -66,7 +66,7 @@ class AddBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class.ja
         }
 
         // Check for single line if expression
-        if (element is KtIfExpression && expression.isInSingleLine() && element.`else` == null) {
+        if (element is KtIfExpression && element.`else` == null) {
             // Check if a comment is actually underneath (\n) the expression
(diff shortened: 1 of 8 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/AddBracesIntention.kt
```diff
@@ -68,7 +66,7 @@ class AddBracesIntention : SelfTargetingIntention<KtElement>(KtElement::class.ja
         }
 
         // Check for single line if expression
-        if (element is KtIfExpression && expression.isInSingleLine() && element.`else` == null) {
+        if (element is KtIfExpression && element.`else` == null) {
```
mglukhikh: Please change (or delete) the comment above, because after your change it becomes incorrect.

idea/testData/intentions/addBraces/addBracesForSingleStatementIfWithCommentBeneath.kt
```diff
@@ -0,0 +1,10 @@
+fun test() {
+    if (false)<caret>
+        while (true) {
+            foo(1)
+        }
+    //comment about foo(2)
```
mglukhikh: Could you please add similar test with `else` block, like this
```
if (...) doSmth()
else 
    while (true) {
        foo(1)
    }
//  comment about call below
```

and check how it works now? The code `element.else == null` is a bit suspicious.

mglukhikh:
Merged manually, thank you.

PR #2164, opened 2019-03-09
Optimize JVM bytecode for for-loops.

punzki:
Diffs resulting from the port:
https://github.com/punzki/kotlin-native/compare/master...punzki:for-loop-diff

When ForLoopsLowering is added to the JVM lowering phases, this causes some of the forLoop bytecode tests to pass, and many still fail due to differences in behavior compared to non-IR backend:
- Generated conditions for lowered loop
- Generated temporary variables
- Supported iterables (e.g., withIndex(), CharSequences)
- Means of incrementing induction variable

The phase will be added once more TODOs are resolved and it is more functionally complete, to prevent breaking for loops.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
@@ -8,17 +8,17 @@ package org.jetbrains.kotlin.backend.common.lower.loops
 import org.jetbrains.kotlin.backend.common.CommonBackendContext
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
-import org.jetbrains.kotlin.backend.common.lower.DeclarationIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irIfThen
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.ir.IrStatement
-import org.jetbrains.kotlin.ir.builders.*
+import org.jetbrains.kotlin.ir.builders.irCallOp
+import org.jetbrains.kotlin.ir.builders.irGet
+import org.jetbrains.kotlin.ir.builders.irSetVar
 import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrCompositeImpl
-import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
 import org.jetbrains.kotlin.ir.symbols.IrVariableSymbol
 import org.jetbrains.kotlin.ir.types.toKotlinType
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
@@ -178,9 +178,7 @@ private class RangeLoopTransformer(
             forLoopInfo.inductionVariable.type.toKotlinType(),
             forLoopInfo.step.type.toKotlinType()
         )
-        if (forLoopInfo is ProgressionLoopHeader) {
-            forLoopInfo.loopVariable = variable
-        }
+        forLoopInfo.loopVariable = variable
         return with(context.createIrBuilder(getScopeOwnerSymbol(), initializer.startOffset, initializer.endOffset)) {
             variable.initializer = forLoopInfo.initializeLoopVariable(symbols, this)
             val increment = irSetVar(
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
@@ -1,0 +1,257 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.common.lower.loops
+
+import org.jetbrains.kotlin.backend.common.CommonBackendContext
+import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
+import org.jetbrains.kotlin.backend.common.lower.DeclarationIrBuilder
+import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.common.lower.irIfThen
+import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.ir.IrStatement
+import org.jetbrains.kotlin.ir.builders.*
+import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.declarations.IrVariable
+import org.jetbrains.kotlin.ir.expressions.*
+import org.jetbrains.kotlin.ir.expressions.impl.IrCompositeImpl
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.symbols.IrVariableSymbol
+import org.jetbrains.kotlin.ir.types.toKotlinType
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+import org.jetbrains.kotlin.util.OperatorNameConventions
+
+val forLoopsPhase = makeIrFilePhase(
+    ::ForLoopsLowering,
+    name = "ForLoopsLowering",
+    description = "For loops lowering"
+)
+
+/**
+ * This lowering pass optimizes for-loops.
+ *
+ * Replace iteration over ranges (X.indices, a..b, etc.) and arrays with
+ * simple while loop over primitive induction variable.
+ */
+internal class ForLoopsLowering(val context: CommonBackendContext) : FileLoweringPass {
+
+    private val headerInfoBuilder = HeaderInfoBuilder(context)
+
+    override fun lower(irFile: IrFile) {
+        val oldLoopToNewLoop = mutableMapOf<IrLoop, IrLoop>()
+        val transformer = RangeLoopTransformer(context, oldLoopToNewLoop, headerInfoBuilder)
+        irFile.transformChildrenVoid(transformer)
+        // Update references in break/continue.
+        irFile.transformChildrenVoid(object : IrElementTransformerVoid() {
+            override fun visitBreakContinue(jump: IrBreakContinue): IrExpression {
+                oldLoopToNewLoop[jump.loop]?.let { jump.loop = it }
+                return jump
+            }
+        })
+    }
+}
+
+private class RangeLoopTransformer(
+    val context: CommonBackendContext,
+    val oldLoopToNewLoop: MutableMap<IrLoop, IrLoop>,
+    headerInfoBuilder: HeaderInfoBuilder
+) : IrElementTransformerVoidWithContext() {
+
+    private val symbols = context.ir.symbols
+    private val iteratorToLoopHeader = mutableMapOf<IrVariableSymbol, ForLoopHeader>()
+    private val headerProcessor = HeaderProcessor(context, headerInfoBuilder, this::getScopeOwnerSymbol)
+
+    fun getScopeOwnerSymbol() = currentScope!!.scope.scopeOwnerSymbol
+
+    override fun visitVariable(declaration: IrVariable): IrStatement {
+        val initializer = declaration.initializer
+        if (initializer == null || initializer !is IrCall) {
+            return super.visitVariable(declaration)
+        }
+        return when (initializer.origin) {
+            IrStatementOrigin.FOR_LOOP_ITERATOR ->
+                processHeader(declaration)
+            IrStatementOrigin.FOR_LOOP_NEXT ->
+                processNext(declaration)
+            else -> null
+        } ?: super.visitVariable(declaration)
+    }
+
+    private fun processHeader(variable: IrVariable): IrStatement? {
+        assert(variable.symbol !in iteratorToLoopHeader)
+        val forLoopInfo = headerProcessor.processHeader(variable)
+            ?: return null
+        iteratorToLoopHeader[variable.symbol] = forLoopInfo
+        return IrCompositeImpl(
+            variable.startOffset,
+            variable.endOffset,
+            context.irBuiltIns.unitType,
+            null,
+            forLoopInfo.declarations
+        )
+    }
+
+    /**
+     * This loop
+     *
+     * for (i in first..last step foo) { ... }
+     *
+     * is represented in IR in such a manner:
+     *
+     * val it = (first..last step foo).iterator()
+     * while (it.hasNext()) {
+     *     val i = it.next()
+     *     ...
+     * }
+     *
+     * We transform it into the following loop:
+     *
+     * var it = first
+     * if (it <= last) {  // (it >= last if the progression is decreasing)
+     *     do {
+     *         val i = it++
+     *         ...
+     *     } while (i != last)
+     * }
+     *
+     * In case of iteration over array we transform it into following:
+     * while (i <= array.size - 1) {
+     *     val element = array[i]
+     *     i++
+     *     ...
+     * }
+     */
+    // TODO:  Lower `for (i in a until b)` to loop with precondition: for (i = a; i < b; a++);
+    override fun visitWhileLoop(loop: IrWhileLoop): IrExpression {
+        if (loop.origin != IrStatementOrigin.FOR_LOOP_INNER_WHILE) {
+            return super.visitWhileLoop(loop)
+        }
+        return with(context.createIrBuilder(getScopeOwnerSymbol(), loop.startOffset, loop.endOffset)) {
+            val newBody = loop.body?.transform(this@RangeLoopTransformer, null)?.let {
+                if (it is IrContainerExpression && !it.isTransparentScope) {
+                    IrCompositeImpl(startOffset, endOffset, it.type, it.origin, it.statements)
+                } else {
+                    it
+                }
+            }
+            val loopHeader = getLoopHeader(loop.condition)
+                ?: return super.visitWhileLoop(loop)
+            val newLoop = loopHeader.buildBody(this, loop, newBody)
+            oldLoopToNewLoop[loop] = newLoop
+            // Build a check for an empty progression before the loop.
+            if (loopHeader is ProgressionLoopHeader) {
+                buildEmptinessCheck(newLoop, loopHeader)
+            } else {
+                newLoop
+            }
+        }
+    }
+
+    private fun DeclarationIrBuilder.buildMinValueCondition(forLoopHeader: ForLoopHeader): IrExpression {
+        // Condition for a corner case: for (i in a until Int.MIN_VALUE) {}.
+        // Check if forLoopHeader.bound > MIN_VALUE.
+        val progressionType = forLoopHeader.progressionType
+        val irBuiltIns = context.irBuiltIns
+        val minConst = when (progressionType) {
+            ProgressionType.INT_PROGRESSION -> IrConstImpl
+                .int(startOffset, endOffset, irBuiltIns.intType, Int.MIN_VALUE)
+            ProgressionType.CHAR_PROGRESSION -> IrConstImpl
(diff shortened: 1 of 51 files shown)


punzki:
The most notable difference in the port is the use of `SimpleType` instead of `IrClassSymbol` when matching iterables to support (e.g., checking for supported array type).

I couldn't figure out why, but currently the symbol table contains duplicate symbols for the same class, which seems to be a difference in behavior from K/N. For example, I've seen multiple symbols in the table for `kotlin.Int`, therefore the array element types wouldn't match.

max-kammerer:
@punzki You have several TODOs in codes, it would be best to implement them before this pr would be merged in master (so K/N could migrate to common lower). 
It also worth to await JVM integration to check correctness at least in one of backend

max-kammerer:
>
I couldn't figure out why, but currently the symbol table contains duplicate symbols for the same class, which seems to be a difference in behavior from K/N. For example, I've seen multiple symbols in the table for kotlin.Int, therefore the array element types wouldn't match.


Yes, we have such problem and @udalov working on it.
Migration to irTypes looks good, but seems you will get same instance duplication even with irTypes, but AbstractTypeChecker.equalTypes should help (https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2795)

punzki:
> @punzki You have several TODOs in codes, it would be best to implement them before this pr would be merged in master (so K/N could migrate to common lower).
> It also worth to await JVM integration to check correctness at least in one of backend

@max-kammerer The 2 major TODOs are to use `getProgressionLast` and `checkProgressionStep`. (Both functions are only present in some internal package in K/N and I would have to port it.)  My intention is to diverge a little from how the lowering is implemented in K/N, and do it more similar to how it's done in the current, non-IR JVM backend.

My understanding of the code is this: ForLoopsLowering from K/N elides the instantiation of the Progression objects with step > 0 (e.g., if it sees something like `for (i in 1..10 step 2)`). This is why it has to insert calls to assert the provided step is > 0 (`checkProgressionStep`) and compute the "actual" last value (`getProgressionLast`, the actual last in the above example would be 9).

The current JVM backend does not do this. It simply gets the `first/last/step` properties when it sees a non-standard step. It can do this because the `step()` extension function on Progressions asserts the provided step, and the Progression constructor computes the actual last value (by calling `getProgressionLastElement` which is internal only).

My intention is to make ForLoopsLowering use the properties for non-trivial Progressions. This would greatly simplify the lowering logic at the cost of creating the Progression object. What are your thoughts on this?

Another difference is that the current JVM backend uses while loops instead of do-while where possible and it is safe to do so (for Progressions), but the ForLoopsLowering always uses do-while for Progressions. I do _not_ plan to implement this decision logic in ForLoopsLowering, also to simplify the lowering logic (I might have to fork the bytecode text tests).

P.S. I'll do better in the future in communicating my intentions, when submitting the PR. 

> Yes, we have such problem and @udalov working on it.
> Migration to irTypes looks good, but seems you will get same instance duplication even with irTypes, but AbstractTypeChecker.equalTypes should help (https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2795)

Sounds good. Note that I'm using `KotlinType`/`SimpleType` as well to match the progression types. Should we avoid using `KotlinType` in the future?

max-kammerer:
@punzki Yes, it best to avoid KotlinTypes and Descriptors in new code

> My intention is to make ForLoopsLowering use the properties for non-trivial Progressions. This would greatly simplify the lowering logic at the cost of creating the Progression object. What are your thoughts on this?

Yes, it would as same as old backend works (seems you already made investigation in ForLoopGenerator/AbstractForInRangeLoopGenerator/AbstractForInProgressionLoopGenerator hierarchy in old backed).

punzki:
> > My intention is to make ForLoopsLowering use the properties for non-trivial Progressions. This would greatly simplify the lowering logic at the cost of creating the Progression object. What are your thoughts on this?
> 
> Yes, it would as same as old backend works (seems you already made investigation in ForLoopGenerator/AbstractForInRangeLoopGenerator/AbstractForInProgressionLoopGenerator hierarchy in old backed).

@max-kammerer Do you have any objection to doing the lowering that way? I.e. simplify the logic by using the Progression properties.

max-kammerer:
@punzki Just to clarify your original point
> ... use the properties for non-trivial Progressions

Did you mean that trivial progression wouldn't be constructed at all?

punzki:
> @punzki Just to clarify your original point
> 
> > ... use the properties for non-trivial Progressions
> 
> Did you mean that trivial progression wouldn't be constructed at all?

@max-kammerer Correct, but to clarify what are "trivial" vs "non-trivial": Trivial would be any Progression expression that are any of:
1. From a call to `rangeTo / ..`
2. From a call to `until`
3. From a call to `downTo`
4. From a call to `step` with `1` or `-1` on a trivial Progression, so something like `1..10 step 2 step 1` or `someProgression step 1` is _not_ trivial.

Currently the old JVM backend removes the construction of those Progressions and we can do the same for the IR backend.

"Non-trivial" would be Progression expressions that are any from a call to `step` with a value that's _not_ `1` or `-1`, e.g., `1..10 step 2`.

The current JVM backend allows the construction of the non-trivial Progressions (as described above), and the ForLoopsLowering from K/N attempts to remove its construction. To be clear, my proposal is to port ForLoopsLowering but instead allow the construction, and read the Progression's `first/last/step` properties in the same way we would for a Progression that's already constructed.

max-kammerer:
@punzki If I undertand you correctly, new lower will  simulate logic of old JVM backend. OK!

punzki:
@max-kammerer I've completed additional changes on top of the ported ForLoopsLowering. These are significant changes from the kotlin-native version, with the goal of simplifying the lowering logic:
- (As we discussed previously) The kotlin-native version elides the instantiation of progressions built using the `step` extension function. It does this by inserting calls to assert the provided step (`checkProgressionStep`) and compute the actual last value (`getProgressionLast`).
- This version does not do special handling for `step` progressions and instead allows them to be created. Basically I deleted `StepHandler`.
- Added a `DefaultProgressionHandler` that handles for-loops over non-specialized progressions. This also handles `step` progressions. `DefaultProgressionHandler` uses the `first/last/step` properties of the progression when building the loop header.
- The `HeaderInfo` for the kotlin-native version has boolean properties for "increasing" and "closed". The emptiness check and "last" expression depend on the value of "closed". This version uses an enum for "increasing" and assumes all progressions are closed. This means that the `HeaderInfo` for arrays and
`until` progressions already include a decrement for the "last" expression. This simplifies the building of the loop and its conditions, moving the necessary work to specific handlers. This will also simplify the handling of `reversed()` in the future since we don't need to consider the case where the lower bound is open.
- `HeaderInfo.step` must be explicitly specified by the handlers; no "default" step creation based on the given direction. In fact, the `direction` property is lazily computed based on the step.

There are also minor changes:
- Tried to keep consistency across in code style. Added more comments and documentation to clarify the code.
- AFAICT, avoided the use of descriptors altogether.
- In `HeaderInfo`, renamed `lowerBound` to `first` and `upperBound` to `last` to reduce confusion when `step` is decreasing.
- Refactored `ProgressionType` extension functions as member functions.
- Moved the emptiness condition building from `ForLoopsLowering` into the `ForLoopHeader` class.
- Moved the special emptiness condition (for the `for (i in a until MIN_VALUE)` corner case) from `ForLoopsLowering` into `UntilHandler` when building the `HeaderInfo`.

During development, I split up the changes over smaller commits which you can review, if it's easier to review in smaller chunks.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
```diff
@@ -0,0 +1,239 @@
...
+ *         val i = it++
+ *         ...
+ *     } while (i != last)
+ * }
+ * ```
+ *
+ * In case of iteration over array we transform it into following:
+ *
+ * ```
+ * while (it <= array.size - 1) {
```
max-kammerer: Please add bytecode tests

max-kammerer: Seems jvm backend should have different behaviour for such optimizaion on array reassignment inside loops depending of compiler version, @dnpetrov correct me if not. Maybe it shouldn't be supported in IR

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
```diff
@@ -0,0 +1,239 @@
...
+ * For example, this loop:
+ *
+ * ```
+ * for (i in first..last step foo) { ... }
+ * ```
+ *
+ * is represented in IR in such a manner:
+ *
+ * ```
+ * val it = (first..last step foo).iterator()
```
max-kammerer: I would prefer to supports several different templates here (and seems they could vary on backends) with less tempory variables:
1. For totally safe cases with no overflow and non-empty interval, e.g. fir (i in 1..20), generate:
```
var i = first
while (i <= first) {
...
 i++
}
```
2. for other cases (this template looks like current jvm-backend work)
```
var i = first
  if (i <= last) { 
      while(true) {
 
          ...
 
           overflow check if required (closed bound) and break
           i++
      }
  }
```

max-kammerer: @punzki It'm suggesting very conservative approach (maybe we need some investigations with benchmarking) cause I don't know how good different JVM will optimize your template

max-kammerer: @dzharkov @udalov @dnpetrov Any thoughts?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
udalov: @max-kammerer Reassignment of array loop variables has changed semantics in 1.3 ([KT-21354](https://youtrack.jetbrains.com/issue/KT-21354)) and we don't intend (yet, at least) to support `-language-version 1.2` in the IR backend, so I think we should only support the new semantics.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
udalov: I agree, I would also prefer even more optimized version of the loop in case the right bound is a constant expression, similarly to how the current JVM backend manages to optimize it. But I'm not opposed to the current @punzki 's implementation either; it's OK to postpone this since it's not the _most_ practical use case of for loops. Maybe we need to ensure that there are bytecode text tests on this optimization though (that check that there's only one `if_icmp..` for a constant-int for loop)

punzki: > I would prefer to supports several different templates here (and seems they could vary on backends) with less tempory variables:

@max-kammerer I noted this in a [TODO in `HeaderProcessor.kt`](https://github.com/JetBrains/kotlin/pull/2164/files#diff-8070e1cd08057790e0480b78ebc62042R78):
```
// TODO: Build while loop (instead of do-while) where possible, e.g., in "until" ranges, or when "last" is known to be < MAX_VALUE
```

I chose to leave it as a TODO since the original `ForLoopsLowering` didn't specifically handle the "no possible overflow" case. I do have a change in-progress to handle this. Would you prefer I include this change in this PR? Or in a future PR? I would personally prefer to push this to a different PR (makes reviews and my life simpler). :smile:

> Maybe we need to ensure that there are bytecode text tests on this optimization though (that check that there's only one if_icmp.. for a constant-int for loop)

@udalov There are in fact assertions to that effect in `forInRangeTo(Char|Long)Const.kt` and the tests are still failing (and disabled) for that very reason.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
punzki: > Please add bytecode tests

@max-kammerer Done!

> Reassignment of array loop variables has changed semantics in 1.3 (KT-21354) ... we should only support the new semantics.

@udalov @max-kammerer **Please confirm:** The optimization in this PR _only_ supports the new semantics (since a temporary variable for the array is created). This is my understanding if I'm reading KT-21354 (and what you said) correctly.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
max-kammerer: @punzki 
Just note: in old backend we are awraing overflow only on increment operation, 
with always close bounds additional operation is introduced `last = rightOpenBound -1` and it would also can cause overflow, e.g. `for (i in 1..MIN_INT)`, so always closed bounds could add some complexity in generated bytecode.

Further optimizations could be added as separate PR. But we need to clarify situation about number of overflow checks and reduce its count.

punzki: @max-kammerer Right, that's already considered in this PR. For `until` loops (see `UntilHandler`), there is an additional condition in the "emptiness check": `inductionVariable <= last && bound > MIN_VALUE`.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
udalov: Yes, this is correct. Since the tests on the new semantics pass in the IR backend in master and are not changed in this optimization, the new semantics is supported.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
max-kammerer: @punzki Yes, but I'm awaring about nested loops:
```
for (ext in smth) {
   for (i in 1 until ext) {
  }
}
```
In such case we would check additional condition on every external loop iteration. But this can be avoided if we would keep information about interval bound - closed/opened.

punzki: @max-kammerer I'm having trouble getting your point. Maybe you can explain in the form of the lowered for-loop?

The original `ForLoopsLowering` implementation distinguished between closed/open bounds and [always added the additional `bound > MIN_VALUE` condition](https://github.com/JetBrains/kotlin/pull/2164/commits/a1e4ab6a46d8f486a12f53cba6c17bcf859e55a7#diff-f2ebc3530ff622def330bd11ef340f7eL208). In your example, since we don't know what the value of `ext` could be, then we have to always add the extra condition. I don't see how it could be avoided by keeping the closed/open information.

punzki: Alternatively, instead of computing `last = rightOpenBound - 1`, we can change the comparison operators and structure of the loop depending on the closed/open information. Is this what you were describing? Something like:

```
// 1. If right bound is OPEN (i.e., "until" loop, array, ".indices" loop), we can use a "while" loop:
var inductionVar = leftBound
while (inductionVar < rightBound) {
  val loopVar = inductionVar
  inductionVar += step
  // Loop body
}

// 2. If induction variable can NOT overflow, and right bound is CLOSED (I can include this in this PR since we are doing "while" loops above anyway)
var inductionVar = leftBound
while (inductionVar <= rightBound) {
  val loopVar = inductionVar
  inductionVar += step
  // Loop body
}

// 3. Default loop structure:
var inductionVar = leftBound
if (inductionVar <= rightBound) {
  do {
    val loopVar = inductionVar
    inductionVar += step
    // Loop body
  } while (inductionVar != rightBound)
}
```

punzki: @max-kammerer I have a change that keeps the open/closed bound info and modifying the loop condition operator based on that (`<=` for closed, `<` for open). With this, the additional `bound > MIN_VALUE` condition is no longer necessary. If you think this is the way to go, let me know and I'll add it to the PR.

Personally, I think this is the right approach and it is much simpler. I should have done this in the first place but I got anchored on how the original `ForLoopsLowering` handled `until` loops. The old JVM backend just changes the operator based on open/closed.

max-kammerer: @punzki Yes, you catch my thoughts! (I meant additional condition about 'bound > MIN_VALUE'.)
And sorry for incomprehensibility from my side (last days I'm a bit sickly).

punzki: @max-kammerer No trouble at all, hope you are better now! Updated the PR with changes to use while loops for possible, and use appropriate operator based on open/closed.

compiler/testData/codegen/bytecodeText/forLoop/forInObjectArray.kt
```diff
@@ -0,0 +1,11 @@
+fun array() = arrayOfNulls<Any>(4)
+
+fun f() {
+    for (i in array()) {
+    }
```
max-kammerer: Please add some foo() call inside to avoid clashing with another optimizations in future:
```
 for (i in array()) {
    foo()
    }
```

compiler/testData/codegen/bytecodeText/forLoop/forInPrimitiveArray.kt
```diff
@@ -0,0 +1,16 @@
+fun intArray() = intArrayOf(0, 0, 0, 0)
+fun longArray() = longArrayOf(0, 0, 0, 0)
+
+fun f() {
+    for (i in intArray()) {
```
max-kammerer: ...and here too

compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInPrimitiveArrayIndices.kt
```diff
@@ -12,6 +11,4 @@ fun test() {
 // 0 getFirst
 // 0 getLast
 
-// 0 IF_ICMPGT
-// 0 IF_ICMPEQ
-// 1 IF_ICMPGE
+// 1 IF_ICMP
```
max-kammerer:
```
// 1 IF_ICMPGE
```
Could be reverted, please also add :
```
// 1 IF
``

File: compiler/testData/codegen/bytecodeText/forLoop/forInDownToCharMinValue.kt [hunk 6]
```diff
@@ -0,0 +1,14 @@
+const val M = Char.MIN_VALUE
+
+fun f(a: Char) {
+    for (i in a downTo M) {
```
max-kammerer: ....Please add some call inside loops here and below

compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInObjectArrayIndices.kt
```diff
@@ -12,6 +11,4 @@ fun test() {
 // 0 getFirst
 // 0 getLast
 
-// 0 IF_ICMPGT
-// 0 IF_ICMPEQ
-// 1 IF_ICMPGE
+// 1 IF_ICMP
```
max-kammerer: Please change to
```
// 1 IF_ICMPGE
// 1 IF
```

compiler/testData/codegen/bytecodeText/forLoop/forInRangeToCharConst.kt
```diff
@@ -9,5 +8,10 @@ fun test(): Int {
...
-// 0 IF_ICMPEQ
-// 1 IF_ICMPGT
\ No newline at end of file
+// 0 iterator
+// 0 getStart
+// 0 getEnd
+// 0 getFirst
+// 0 getLast
+// 0 getStep
+// 1 IF_ICMP
```
max-kammerer: // 1 IF_ICMPGE
// 1 IF

max-kammerer:
@punzki  FYI: We have test generator for comparison logic - GenerateRangesCodegenTestData.
And similar generator could be implemented for loops (but this is not necessary for this pr)

compiler/testData/codegen/bytecodeText/forLoop/forInRangeToCharConst.kt
punzki: I've tried to add it where it makes sense, however in some cases I couldn't be very specific with the instructions because of differences between IR and non-IR. One example is in `downTo` loops, where the condition for IR is something like `while (last <= inductionVar)` whereas for non-IR it is `while (inductionVar >= last)`, resulting in different `IF_ICMPXX` instructions.

max-kammerer:
@punzki Thank You! Great Work!

PR #2165, opened 2019-03-09
KT-29572 Smart completing anonymous object uses incorrect code style

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29572


goodwinnk:
Such spaces should be controlled by formater. An ideal solution should check state of "Before colon in new type definition" option or just reformat prepared template. The proposed solution does the right step and now we have a right behaviour with all default settings. 

So this is now in master: https://github.com/JetBrains/kotlin/commit/fd262bcd8dd2cbfc8ba5c09577e645fb3588eb71. Thank you!

Please consider another PR witch will work for any code style.

t-kameyama:
I will send another PR.

PR #2167, opened 2019-03-11
Run ExpectDeclarationsRemoving in the JVM_IR backend.

sfs:
The expect/actual mechanism is already implemented in the JS backend and seems to work fine in the JVM backend as well.


max-kammerer:
@sfs Thanks!

max-kammerer:
@sfs Could you look at failed test after pr: https://teamcity.jetbrains.com/viewLog.html?buildId=2051455&buildTypeId=Kotlin_dev_Compiler ?

sfs:
On it!

Looks like this test was actually broken before (it didn't use the default arguments). The test doesn't actually check for this though and thus didn't fail.

Expect a fix soon.

PR #2168, opened 2019-03-12
when: emit switch for String if possible

ting-yuan:
Effectively, the following when structure:
```
  when (s) {
    s1, s2 -> e1,
    s3 -> e2,
    s4 -> e3,
    ...
    else -> e
  }
```
is implemented as:
```
  when (s.hashCode()) {
    h1 -> {
      if (s == s1)
        e1
      else if (s == s2)
        e1
      else if (s == s3)
        e2
      else
        e
    }
    h2 -> if (s == s3) e2 else e,
    ...
    else -> e
  }
```
where `s1.hashCode() == s2.hashCode() == s3.hashCode() == h1,
      s4.hashCode() == h2`.

A `tableswitch` or `lookupswitch` is used for the hash code lookup.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
@@ -22,13 +22,17 @@ import java.util.*
 class SwitchGenerator(private val expression: IrWhen, private val data: BlockInfo, private val codegen: ExpressionCodegen) {
     private val mv = codegen.mv
 
+    data class ExpressionToLabel(val expression: IrExpression, val label: Label)
+    data class CallToLabel(val call: IrCall, val label: Label)
+    data class ValueToLabel(val value: Any?, val label: Label)
+
     // @return null if the IrWhen cannot be emitted as lookupswitch or tableswitch.
     fun generate(): StackValue? {
         val endLabel = Label()
         var defaultLabel = endLabel
-        val thenExpressions = ArrayList<Pair<IrExpression, Label>>()
+        val expressionToLabels = ArrayList<ExpressionToLabel>()
         var elseExpression: IrExpression? = null
-        val allConditions = ArrayList<Pair<IrCall, Label>>()
+        val callToLabels = ArrayList<CallToLabel>()
 
         // Parse the when structure. Note that the condition can be nested. See matchConditions() for details.
         for (branch in expression.branches) {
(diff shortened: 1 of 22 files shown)


compiler/testData/codegen/bytecodeText/whenStringOptimization/duplicatingItemsSameHashCode3.kt
```diff
@@ -0,0 +1,19 @@
...
+
+fun foo(x : String) : String {
+    assert("abz]".hashCode() == "aby|".hashCode())
+
+    when (x) {
+        "abz]" -> return "abz"
+        "ghi"  -> return "ghi"
+        "aby|" -> return "aby"
+        "abz]" -> return "fail"
+        "xyz" -> return "xyz"
```
max-kammerer: Please add box test for this case similar to original one in compiler/testData/codegen/box/when/stringOptimization/duplicatingItemsSameHashCode.kt

compiler/testData/codegen/bytecodeText/whenStringOptimization/duplicatingItemsSameHashCode2.kt
```diff
@@ -0,0 +1,17 @@
...
+        "abz]" -> return "abz"
+        "ghi"  -> return "ghi"
+        "aby|" -> return "aby"
+        "abz]" -> return "fail"
+    }
+
+    return "other"
+}
+
+// 0 LOOKUPSWITCH
```
max-kammerer: Is LOOKUPSWITCH is absent cause of hashAndSwitchLabels.size <= 2 condition?

ting-yuan: Yes. I've updated the comments in it.

max-kammerer: Please also add box test for such case

ting-yuan: Didn't add it because it is exactly the same to the existing test. Will update after the tests finished.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -137,58 +173,190 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+
+        protected abstract fun genSubject()
+
+        protected abstract fun genSwitch()
+
+        protected fun genIntSwitch(unsortedIntCases: List<Pair<Int, Label>>) {
+            val intCases = unsortedIntCases.sortedBy { it.first }
+            val caseMin = intCases.first().first
+            val caseMax = intCases.last().first
+            val rangeLength = caseMax - caseMin + 1L
```
max-kammerer: val rangeLength = caseMax.toLong() - caseMin.toLong() + 1L
It would be best to also add some test for overflow case

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -137,58 +173,190 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
-    private fun gen(
-        cases: ArrayList<Pair<Int, Label>>,
+    abstract inner class Switch(
+        val subject: IrGetValue,
+        val defaultLabel: Label,
+        val endLabel: Label,
+        val elseExpression: IrExpression?,
+        val thenExpressions: ArrayList<Pair<IrExpression, Label>>
+    ) {
+        open fun gen(): StackValue? {
```
max-kammerer: I would prefer to explicitly clarify here that gen works only if there are enough cases
gen -> genOptimizedIfEnoughCases (not sure that new name couldn't be better)?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,36 +43,72 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+        // switch isn't applicable if there's no case at all, e.g., when() { else -> ... }
+        if (allConditions.size == 0)
             return null
 
-        if (areConstIntComparisons(allConditions.map { it.first })) {
+        // Remove labels and "then expressions" that are not reachable.
+        fun removeUnreachableExpressions(reachables: HashSet<Label>) = thenExpressions.removeIf { it.second !in reachables }
+
+        val calls = allConditions.map { it.first }
+        if (areConstIntComparisons(calls)) {
```
max-kammerer: Branches seem identical except return value: maybe merge then with explicit cast for cases in each constructor

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,36 +43,72 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
             // if all conditions are CALL EQEQ(tmp_variable, some_int_constant)
-            val cases = allConditions.mapTo(ArrayList()) { Pair((it.first.getValueArgument(1) as IrConst<*>).value as Int, it.second) }
             val subject = allConditions[0].first.getValueArgument(0)!! as IrGetValue
-            return gen(cases, subject, defaultLabel, endLabel, elseExpression, thenExpressions)
+
+            // Don't generate repeated cases, which are unreachable but allowed in Kotlin.
+            // Only keep the first encountered case:
+            val cases =
+                allConditions.map { Pair((it.first.getValueArgument(1) as IrConst<*>).value as Int, it.second) }.distinctBy { it.first }
```
max-kammerer: Please use data class instead of Pair

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,36 +43,72 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
 
-        // IF is more compact when there are only 1 or fewer branches, in addition to else.
-        if (allConditions.size <= 1)
+        // switch isn't applicable if there's no case at all, e.g., when() { else -> ... }
+        if (allConditions.size == 0)
             return null
 
-        if (areConstIntComparisons(allConditions.map { it.first })) {
+        // Remove labels and "then expressions" that are not reachable.
+        fun removeUnreachableExpressions(reachables: HashSet<Label>) = thenExpressions.removeIf { it.second !in reachables }
```
max-kammerer: Please add box test for integers (if it's not present)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -137,58 +173,190 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+        subject: IrGetValue,
+        defaultLabel: Label,
+        endLabel: Label,
+        elseExpression: IrExpression?,
+        thenExpressions: ArrayList<Pair<IrExpression, Label>>,
+        private val cases: List<Pair<String?, Label>>
+    ) : Switch(subject, defaultLabel, endLabel, elseExpression, thenExpressions) {
+
+        private val hashToStringAndExprLabels = HashMap<Int, ArrayList<Pair<String?, Label>>>()
+        private val hashAndSwitchLabels = ArrayList<Pair<Int, Label>>()
```
max-kammerer: Avoid Pair usage (it's not trivial to read code with pairs below)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -137,58 +173,190 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
-            mv.lookupswitch(defaultLabel, cases.map { it.first }.toIntArray(), cases.map { it.second }.toTypedArray())
-        } else {
-            val labels = Array(rangeLength.toInt()) { defaultLabel }
+        thenExpressions: ArrayList<Pair<IrExpression, Label>>,
+        private val cases: List<Pair<Int, Label>>
+    ) : Switch(subject, defaultLabel, endLabel, elseExpression, thenExpressions) {
+
+        override fun gen(): StackValue? {
+            // IF is more compact when there are only 1 or fewer branches, in addition to else.
+            if (cases.size <= 1)
```
max-kammerer: Could  be extracted to precondition check and called within base class

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Done. `areConstComparisons` are hoisted out of `areConst[type]Comparisons` to guard the shared operations, which are moved before `areConst[type]Comparisons`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -38,41 +42,82 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+        // Don't generate repeated cases, which are unreachable but allowed in Kotlin.
+        // Only keep the first encountered case:
+        val cases =
+            callToLabels.map { ValueToLabel((it.call.getValueArgument(1) as IrConst<*>).value, it.label) }.distinctBy { it.value }
+
+        // Remove labels and "then expressions" that are not reachable.
+        val reachableLabels = HashSet(cases.map { it.label })
+        expressionToLabels.removeIf { it.label !in reachableLabels }
+
+        if (areConstIntComparisons(calls)) {
```
max-kammerer: Maybe best
```
return when {
            areConstIntComparisons(calls) ->
                IntSwitch(
                    subject,
                    defaultLabel,
                    endLabel,
                    elseExpression,
                    expressionToLabels,
                    cases
                )
            areConstStringComparisons(calls) ->
                StringSwitch(
                    subject,
                    defaultLabel,
                    endLabel,
                    elseExpression,
                    expressionToLabels,
                    cases
                )
            else -> null // TODO: Enum, etc.
        }?.genOptimizedIfEnoughCases()
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -137,58 +182,184 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
-        val result = if (elseExpression == null) {
-            // There's no else part. No stack value will be generated.
-            StackValue.putUnitInstance(mv)
-            onStack(Type.VOID_TYPE)
-        } else {
-            // Generate the else part.
-            mv.visitLabel(defaultLabel)
-            val stackValue = elseExpression.run { gen(this, data) }
-            coerceNotToUnit(stackValue.type, stackValue.kotlinType, expression.type.toKotlinType())
+        override fun shouldOptimize() = hashAndSwitchLabels.size > 2
```
max-kammerer: Please add comment to this heuristic (seems it was lost)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Looks better, thanks!

max-kammerer:
@ting-yuan Many Thanks!

PR #2169, opened 2019-03-12

burakeregar:


mglukhikh:
@burakeregar Sorry, this code is very ad-hoc, I cannot accept this. In the general case we have no guarantees that generated code is correct. The problem arises here because `safeDeparenthesize` in the beginning drops label, you can think about a version of `safeDeparenthesize` that keeps labels.

burakeregar:
@mglukhikh Thanks for your comment. Yeah, that makes sense. I have created another PR for this: https://github.com/JetBrains/kotlin/pull/2218

PR #2170, opened 2019-03-12
KT-29869 Disable `WhenWithOnlyElse` by default.

burakeregar:
https://youtrack.jetbrains.com/issue/KT-29869


mglukhikh:
Merged 👍

PR #2172, opened 2019-03-13
JVM_IR: Perform asm type conversion for string concat arguments.

madsager:
This ensures that the Unit instance is loaded when string
concatenation contains a sub part that is a call of a Unit
return-type method.


max-kammerer:
@madsager Thanks!

PR #2173, opened 2019-03-13
Resolve type aliases when looking for default arguments to actual method

sfs:
This fixes `annotationsViaActualTypeAliasFromBinary.kt`.

### DIFF
--- a/compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
+++ b/compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
@@ -91,7 +91,7 @@ object ExpectedActualResolver {
                             Substitutor(expectedClass.declaredTypeParameters, container.declaredTypeParameters)
                         }
                         else null
-                    areCompatibleCallables(declaration, actual, commonModule, parentSubstitutor = substitutor)
+                    areCompatibleCallables(declaration, actual, parentSubstitutor = substitutor)
                 }
             }
             is ClassifierDescriptorWithTypeParameters -> {
@@ -228,7 +228,7 @@ object ExpectedActualResolver {
     private fun areCompatibleCallables(
         a: CallableMemberDescriptor,
         b: CallableMemberDescriptor,
-        platformModule: ModuleDescriptor,
+        platformModule: ModuleDescriptor = b.module,
         parentSubstitutor: Substitutor? = null
     ): Compatibility {
         assert(a.name == b.name) { "This function should be invoked only for declarations with the same name: $a, $b" }
--- a/compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
+++ b/compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
@@ -50,7 +50,7 @@ object ExpectedActualResolver {
                     // TODO: support non-source definitions (e.g. from Java)
                     actual.source.containingFile != SourceFile.NO_SOURCE_FILE
                 }.groupBy { actual ->
-                    areCompatibleCallables(expected, actual)
+                    areCompatibleCallables(expected, actual, platformModule)
                 }
             }
             is ClassDescriptor -> {
@@ -91,7 +91,7 @@ object ExpectedActualResolver {
                             Substitutor(expectedClass.declaredTypeParameters, container.declaredTypeParameters)
                         }
                         else null
-                    areCompatibleCallables(declaration, actual, parentSubstitutor = substitutor)
+                    areCompatibleCallables(declaration, actual, commonModule, parentSubstitutor = substitutor)
                 }
             }
             is ClassifierDescriptorWithTypeParameters -> {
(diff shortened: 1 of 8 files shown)


compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
```diff
@@ -91,7 +91,7 @@ object ExpectedActualResolver {
                             Substitutor(expectedClass.declaredTypeParameters, container.declaredTypeParameters)
                         }
                         else null
-                    areCompatibleCallables(declaration, actual, parentSubstitutor = substitutor)
+                    areCompatibleCallables(declaration, actual, commonModule, parentSubstitutor = substitutor)
```
max-kammerer: @sfs Seems this is redundant

max-kammerer:
@sfs Please add next test to compiler/testData/codegen/box/multiplatform/defaultArguments
```
// !LANGUAGE: +MultiPlatformProjects
// WITH_RUNTIME
// FILE: common.kt

expect class Foo() {
    fun member(a: String, b: Int = 0, c: Double? = null): String
}

// FILE: jvm.kt

import kotlin.test.assertEquals

actual typealias Foo = Foo2

class Foo2 constructor() {
    fun member(a: String, b: Int = 0, c: Double? = null): String = a + "," + b + "," + c
}

fun box(): String {
    val foo = Foo()
    assertEquals("OK,0,null", foo.member("OK"))
    assertEquals("OK,42,null", foo.member("OK", 42))
    assertEquals("OK,42,3.14", foo.member("OK", 42, 3.14))

    return "OK"
}
```

compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
sfs: I got rid of the default argument to `areCompatibleCallables`, since it uses the wrong `ModuleDescriptor` when called on a resolved class descriptor for a type alias.

To avoid more bugs in the future it's probably better to be explicit here.

sfs:
@max-kammerer I added the tests and confirmed that they succeed.

compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
max-kammerer: @sfs areCompatibleCallables asumes platformModule in parameter list, but here commonModule is passed.
Could you explain in mode details why default parameter is wrong for typeAliases?

sfs: @max-kammerer the platformModule parameter is used in `areCompatibleTypes` to compare two types up to expansion of type aliases for expected and actual type constructors. The `commonModule` contains both declarations, since otherwise you wouldn't have found the expected declaration in the first place.

However, if you expand a typealias, the module for the expanded declaration contains only its own dependencies and in particular might not contain the expected/actual declarations at all. This happens in the `annotationsViaActualTypeAliasFromBinary.kt` test case, where the module for `Jnno` doesn't contain either declaration of `Anno`.

max-kammerer: @sfs Seems here some misunderstooding: platformModule always contains expect declarations, common only common/expect ones. 
There is only one platform module (common one is merged in platform //@udalov) in 'annotationsViaActualTypeAliasFromBinary' test

sfs: @max-kammerer If I understand you correctly I could simply change `commonModule` to `actual.module`, right?

I'm not sure I understand what you mean though - is there supposed to be only a single module which is the same for all declarations? This is not what I'm seeing; the declaration for `Jnno` points to a different module ("dependencies for test[...]").

max-kammerer: @sfs 

> commonModule to actual.module, right?
yes, and it's same as default  parameter

Test is a bit tricky: there is module for java annotation which is used as dependency for kotlin module. And this kotlin module is actually a platform module not common!

sfs: @max-kammerer I switched it back to using a default argument. Should be fine now.

compiler/resolution/src/org/jetbrains/kotlin/resolve/multiplatform/ExpectedActualResolver.kt
```diff
@@ -50,7 +50,7 @@ object ExpectedActualResolver {
                     // TODO: support non-source definitions (e.g. from Java)
                     actual.source.containingFile != SourceFile.NO_SOURCE_FILE
                 }.groupBy { actual ->
-                    areCompatibleCallables(expected, actual)
+                    areCompatibleCallables(expected, actual, platformModule)
```
max-kammerer: Missed

sfs: @max-kammerer 
No, this is the one that's required. In the `annotationsViaActualTypeAliasFromBinary` test case the function `findNamesakesFromModule` resolves the constructor of Anno to the constructor of Jnno.  This constructor is assigned to `actual` here.

Calling `actual.module` yields the module for the Java annotation which doesn't contain Anno and causes the test to fail.

max-kammerer:
@sfs Thanks! Merged manually!

PR #2174, opened 2019-03-13
KT-15537: Add inspection + intention to replace IntRange.start/endInclusive with first/last

devbridie:
This inspection reports usages of IntRange start and endInclusive. These properties can be replaced with equivalent `kotlin.collections` `first` and `last`.

Example: range.start` can be replaced with range.first.

Issue: https://youtrack.jetbrains.com/issue/KT-15537

The issue explicitly mentions `IntRange`, so this has been done, but I would imagine that it should work on all ranges `ClosedRange<T>` which also implement `Iterable<T>` e.g. `LongRange`, `UIntRange`, `CharRange`, etc. Thoughts?

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -3306,6 +3306,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceRangeStartEndInclusiveWithFirstLastInspection"
+                     displayName="Replace Range 'start' or 'endInclusive' with 'first' or 'last'"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="WEAK WARNING"
+                     language="kotlin"
+    />
+
     <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.RedundantRequireNotNullCallInspection"
                      displayName="Redundant 'requireNotNull' or 'checkNotNull' call"
                      groupPath="Kotlin"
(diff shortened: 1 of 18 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceIntRangeStartEndInclusiveWithFirstLastInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+                    ReplaceIntRangeEndInclusiveWithLastQuickFix()
+                )
+            }
+        }
+    }
+}
+
+private fun KtDotQualifiedExpression.getFunctionName(): FqName? {
+    val context = this.analyze()
+    return this.getResolvedCall(context)?.resultingDescriptor?.fqNameOrNull()
```
mglukhikh: Here you can use `resolveToCall` which does not requires explicit context. It's a bit shorter.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceIntRangeStartEndInclusiveWithFirstLastInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+private fun FqName.isStart(): Boolean {
+    return this == FqName("kotlin.ranges.IntRange.start")
```
mglukhikh: I think it's a good idea to check here different range types, not only `IntRange`, and change inspection description accordingly.

mglukhikh:
Sorry for late review.

devbridie:
Rebased to current master and added support for other `Range` types.

mglukhikh:
Merged manually, thank you.

ilya-g:
Does the inspection explain _why_ these usages should be replaced?

PR #2179, opened 2019-03-13
Incremental annotation processing with KAPT

gavra0:
Add support for incremental annotation processors in KAPT. These
processors conform to https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing
specification.

Support is provided by using javac compiler APIs and
recording the source file structure. At runtime, processors
are instrumented with custom Filer that is used to keep track of generated
files. In order to support classpath changes, stub generation task is
used to generated a list of changed FQCNs, and this is simply used by KAPT.
Both worker and non-worker mode are supported.

Issue is https://youtrack.jetbrains.com/issue/KT-23880

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
@@ -7,6 +7,9 @@ package org.jetbrains.kotlin.kapt3.base
 
 import org.jetbrains.kotlin.base.kapt3.KaptFlag
 import org.jetbrains.kotlin.base.kapt3.KaptOptions
+import org.jetbrains.kotlin.kapt3.base.incremental.DeclaredProcType
+import org.jetbrains.kotlin.kapt3.base.incremental.IncrementalProcessor
+import org.jetbrains.kotlin.kapt3.base.incremental.getIncrementalProcessorsFromClasspath
 import org.jetbrains.kotlin.kapt3.base.util.KaptLogger
 import org.jetbrains.kotlin.kapt3.base.util.info
 import java.io.Closeable
@@ -17,7 +20,7 @@ import java.net.URLClassLoader
 import java.util.*
 import javax.annotation.processing.Processor
 
-class LoadedProcessors(val processors: List<Processor>, val classLoader: ClassLoader)
+class LoadedProcessors(val processors: List<IncrementalProcessor>, val classLoader: ClassLoader)
 
 open class ProcessorLoader(private val options: KaptOptions, private val logger: KaptLogger) : Closeable {
     private var annotationProcessingClassLoader: URLClassLoader? = null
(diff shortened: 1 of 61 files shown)


alexey-tsvetkov:
LGTM, merged manually. Thanks for the contribution!

plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
```diff
@@ -48,7 +51,21 @@ open class ProcessorLoader(private val options: KaptOptions, private val logger:
             logger.info { "Annotation processors: " + processors.joinToString { it::class.java.canonicalName } }
         }
 
-        return LoadedProcessors(processors, classLoader)
+        return LoadedProcessors(wrapInIncrementalProcessor(processors, classpath), classLoader)
```
brianduff: qq: should this be checking options.incrementalKapt before wrapping the processor? I'm seeing new behavior related to reading zip entries from the classpath from the getIncrementalProcessorsFromClasspath() method despite the incrementalKapt option being in its default state (off). Unsure if this is intentional or not.

gavra0: Thanks for catching this, I've uploaded https://github.com/JetBrains/kotlin/pull/2334 to fix it.

plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/incrementalProcessors.kt
```diff
@@ -0,0 +1,159 @@
...
+}
+
+private fun getSrcFiles(elements: Array<out Element?>): List<File> {
+    return elements.filterNotNull().mapNotNull { elem ->
+        var origin = elem
+        while (origin.enclosingElement != null && origin.enclosingElement !is PackageElement) {
+            origin = origin.enclosingElement
+        }
+        val uri = (origin as? Symbol.ClassSymbol)?.sourcefile?.toUri()?.takeIf { it.isAbsolute }
+        uri?.let { File(it) }
```
technoir42: @gavra0 I think we should deduplicate file paths here in case multiple elements originate from the same source file. This could happen, for example, if annotation processor generates multiple Kotlin top-level functions per element.

The place where Gradle handles this, for reference:
https://github.com/gradle/gradle/blob/e7dd3ba/subprojects/language-java/src/main/java/org/gradle/api/internal/tasks/compile/processing/IsolatingProcessingStrategy.java#L67-L71

gavra0: I agree. That is a case I did not consider, I'll upload a PR to address this.

gavra0: I've uploaded https://github.com/JetBrains/kotlin/pull/2361 to fix this.

PR #2181, opened 2019-03-14
Commonized IR serializer infrastructure between native and js

alexander-gorshenev:


compiler/ir/backend.common/src/KotlinIr.proto
```diff
@@ -74,6 +74,11 @@ message IrDeclarationOrigin {
   }
 }
 
+message Name {
```
udalov: Why isn't it enough to encode names as strings with `Name.guessByFirstCharacter` as a way to distinguish special names from identifiers?

alexander-gorshenev: Should be enough, I suppose. Thanks.

alexander-gorshenev:
For reference, the corresponding native PR is here: https://github.com/JetBrains/kotlin-native/pull/2769

compiler/ir/backend.common/src/KotlinIr.proto
alexander-gorshenev: Moved to use Name.guessByFirstCharacter().

udalov:
Has this been reviewed by someone else?

romanart:
@udalov Yes, in [upsource](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2822)

PR #2182, opened 2019-03-15
KT-5412 "Replace non-null assertion with `if` expression" should replace parent expression

burakeregar:
https://youtrack.jetbrains.com/issue/KT-5412


mglukhikh:
Mostly it's Ok (however, I must confess this intention is used very-very rarely). @burakeregar Could you please squash all your commits into one?

mglukhikh:
Never mind, I squashed everything myself. Thank you.

PR #2184, opened 2019-03-15
Refactor InnerClassesLowering & fix `this` remapping

pyos:
Inner class constructors should use the argument instead of reading outer `this` from a field because if such an access happens before a delegating constructor call, e.g. when evaluating an argument, a JVM
bytecode validation error will be thrown. (The only operation on `this` allowed before a delegating constructor call is SETFIELD, and only if the field in question is declared in the same class.)


max-kammerer:
@pyos there are several failed JS tests (seems mostly unmuted): https://teamcity.jetbrains.com/viewLog.html?buildId=2067446&tab=buildResultsDiv&buildTypeId=Kotlin_dev_Compiler

pyos:
Yeah, I've already noticed that in the local build as well. All of them could indeed be unmuted, although I'm not sure why this change would affect them at all, as none of them use inner classes...

max-kammerer:
@pyos Many thanks!

PR #2185, opened 2019-03-15
KT 13962: Intention to replace Java collection constructor calls with  function calls from stdlib (ArrayList() → arrayListOf())

devbridie:
YouTrack: https://youtrack.jetbrains.com/issue/KT-13962

Example of this intention:

```
fun foo() {
    var list: ArrayList<Int> = ArrayList()
}
```

to
```
fun foo() {
    var list: ArrayList<Int> = arrayListOf()
}
```

### DIFF
--- a/idea/resources/intentionDescriptions/ConvertCollectionConstructorToFunction/description.html
+++ b/idea/resources/intentionDescriptions/ConvertCollectionConstructorToFunction/description.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This intention reports explicit calls of <b>Collection constructor</b> calls which can be replaced by function calls from the stdlib.
+</body>
+</html>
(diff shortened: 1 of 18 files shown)


idea/resources/intentionDescriptions/ConvertCollectionConstructorToFunction/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection reports explicit calls of <b>Collection constructor</b> calls which can be replaced by function calls from the stdlib.
```
mglukhikh: It's an intention, not an inspection.

devbridie:
Updated and rebased to current master.

mglukhikh:
Merged 👍

PR #2186, opened 2019-03-15
IR: introduce builtin removal lowering phase.

madsager:
The idea is to make sure that the Ir NOT builtin is not used at
all after this lowering and therefore there is only one Boolean
not definition to deal with and not two.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/BuiltinRemovalLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/BuiltinRemovalLowering.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.common.lower
+
+import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.backend.common.CommonBackendContext
+import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.expressions.IrCall
+import org.jetbrains.kotlin.ir.expressions.IrExpression
+import org.jetbrains.kotlin.ir.util.functions
+import org.jetbrains.kotlin.ir.util.irCall
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+
+val builtinRemovalLoweringPhase = makeIrFilePhase(
+    ::BuiltinRemovalLowering,
+    name = "BuiltinRemovalLowering",
+    description = "Replace usages of builtins with actual library method calls"
+)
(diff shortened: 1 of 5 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/BuiltinRemovalLowering.kt
```diff
@@ -0,0 +1,41 @@
...
+)
+
+class BuiltinRemovalLowering(val context: CommonBackendContext) : FileLoweringPass {
+
+    override fun lower(irFile: IrFile) {
+        irFile.transformChildrenVoid(object : IrElementTransformerVoid() {
+            override fun visitCall(expression: IrCall): IrExpression {
+                expression.transformChildrenVoid(this)
+                return if (expression.symbol == context.irBuiltIns.booleanNotSymbol) {
+                    val booleanNotFunction = context.irBuiltIns.booleanClass.functions.find { it.descriptor.name.asString() == "not" }
```
max-kammerer: Please avoid descriptor usage:
```
it.owner.name.asString() == "not" 
```

madsager: Done, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -35,17 +35,11 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
 
     companion object {
         fun isNegation(expression: IrExpression, context: JvmBackendContext): Boolean {
```
udalov: I think both these functions can now be inlined

madsager: They can. Done.

udalov:
Hi Mads,

We've looked at this problem and your solution in more detail and @anton-bannykh convinced me that the current approach of dealing with `Boolean.not` needs to be refactored in another way. Namely, we should avoid using the fake `NOT` operator completely, even in psi2ir, and always get the real declaration `not` found in `kotlin.Boolean`. Anton has implemented this solution and it's being reviewed at https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-2839. It has allowed to [get rid](https://github.com/JetBrains/kotlin/commit/dd1c96ba83f6583b4c518dfd24ad90f4a01d1939) of the complex check in `JvmBuiltinOptimizationLowering` in exactly the same way as in your pull request.

Also, by @max-kammerer's request I un-inlined `isNegation` since the code looks more concise this way.

I apologize for the way this has been handled. Since this code affects all backends, I think I needed to raise this issue to JS and Native teams earlier, to make sure that they're fine with the proposed solution, before implementing it. Will do that for such issues from now on.

madsager:
Thanks Alex! No worries at all, I'm happy that the discussion led to a solution that works for everyone. Thanks for the pointers to the changes! :-)

HanyLuv:
멋져
2019년 3월 20일 (수) 오후 3:54, Mads Sig Ager <notifications@github.com>님이 작성:

PR #2189, opened 2019-03-16
KT-30414 "Replace return with 'if' expression" drops return label

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30414


mglukhikh:
Merged 👍

PR #2190, opened 2019-03-18
KT-30381 Inspection + quickfix to replace non-null assertion with return

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30381

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -3355,6 +3355,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceNotNullAssertionWithElvisReturnInspection"
+                     displayName="Replace '!!' with '?: return'"
+                     groupPath="Kotlin"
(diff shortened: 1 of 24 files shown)


idea/resources/META-INF/plugin-common.xml
```diff
@@ -3322,6 +3322,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceNotNullAssertionWithElvisReturnInspection"
+                     displayName="Replace '!!' with '?: return'"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="false"
+                     level="INFORMATION"
```
dimonchik0036: `WEAK WARNING`

dimonchik0036:
``` Kotlin
fun test(a: Int?): Int? {
    return check(a!!) // false negative
}

fun check(i: Int): Int = TODO()
```

dimonchik0036:
Thanks! dffcfdc02f3ab62dc971f41272339a60dd0617f2

PR #2192, opened 2019-03-18
KT-30341 False positive 'Use withIndex() instead of manual index increment' inspection with destructive declaration in 'for' loop

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30341


mglukhikh:
Merged 👍

PR #2194, opened 2019-03-19
KT-30173 "Nested lambda has shadowed implicit parameter" is suggested when both parameters are logically the same

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30173


mglukhikh:
Merged 👍

PR #2195, opened 2019-03-20
KT-30426 Don't preserve extra line when adding remaining branches for when

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30426


mglukhikh:
Merged 👍

PR #2196, opened 2019-03-20
Allow tests to be run in parallel.

sfs:
Parallel tests are currently broken for two reasons.

- Tests sometimes use the same root for temporary directories (e.g., /tmp/unitTest_simple), which is deleted once one of the tests finishes.
- There are several places in the code which delete temporary directories. The root temporary directory for a test is always deleted in the `shutDown` method. Other temporary directories are deleted by the `com.intellij.openapi.util.io.FileUtil` class when the test runner shuts down. Some of these temporary directories are also redundantly handled by the `com.intellij.openapi.util.ShutDownTracker` class. The latter two cause problems when running tests in parallel since they may attempt to delete files which are still in use.

This PR fixes these problems by consolidating all temporary files for a test under the root temporary directory for this test and deleting this directory once in the `shutDown` method.

### DIFF
--- a/compiler/build.gradle.kts
+++ b/compiler/build.gradle.kts
@@ -3,6 +3,10 @@ import org.gradle.api.tasks.bundling.Jar
 import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmCompile
 
+tasks.withType<Test> {
+    maxParallelForks = Math.max(Runtime.getRuntime().availableProcessors() / 2, 1)
+}
+
 plugins {
     kotlin("jvm")
     id("jps-compatible")
--- a/js/js.tests/build.gradle.kts
+++ b/js/js.tests/build.gradle.kts
@@ -2,6 +2,10 @@ import com.moowork.gradle.node.NodeExtension
 import com.moowork.gradle.node.npm.NpmTask
 import org.gradle.internal.os.OperatingSystem
 
+tasks.withType<Test> {
+    maxParallelForks = Math.max(Runtime.getRuntime().availableProcessors() / 2, 1)
+}
+
 plugins {
     kotlin("jvm")
     id("jps-compatible")
--- a/compiler/tests-common/tests/org/jetbrains/kotlin/checkers/AbstractForeignAnnotationsNoAnnotationInClasspathTest.kt
+++ b/compiler/tests-common/tests/org/jetbrains/kotlin/checkers/AbstractForeignAnnotationsNoAnnotationInClasspathTest.kt
@@ -22,7 +22,11 @@ import org.jetbrains.kotlin.test.KotlinTestUtils
 import java.io.File
 
 abstract class AbstractForeignAnnotationsNoAnnotationInClasspathTest : AbstractForeignAnnotationsTest() {
-    private val compiledJavaPath = KotlinTestUtils.tmpDir("java-compiled-files")
+    // This should be executed after setUp runs, since setUp changes the root folder
+    // for temporary files.
+    private val compiledJavaPath by lazy {
+        KotlinTestUtils.tmpDir("java-compiled-files")
+    }
 
     override fun getExtraClasspath(): List<File> {
         val foreignAnnotations = createJarWithForeignAnnotations()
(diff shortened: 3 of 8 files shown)


max-kammerer:
@sfs Many thanks for your pr!
But in current state it breaks at least js ir tests on windows: https://teamcity.jetbrains.com/viewLog.html?buildId=2082599&buildTypeId=Kotlin_dev_Compiler
```
J2V8 native library not loaded
```
It could be related to parallezation of j2v8

max-kammerer:
@sfs We are now investigating on our side this fails, maybe they are not related to pr

bashor:
@max-kammerer I think we can merge it w/o changes in `js/js.tests/build.gradle.kts`if everything else ok. WDYT?

js/js.tests/build.gradle.kts
```diff
@@ -2,6 +2,10 @@ import com.moowork.gradle.node.NodeExtension
 import com.moowork.gradle.node.npm.NpmTask
 import org.gradle.internal.os.OperatingSystem
 
+tasks.withType<Test> {
```
romanart: Could we turn it for JS off for now?

compiler/build.gradle.kts
```diff
@@ -3,6 +3,10 @@ import org.gradle.api.tasks.bundling.Jar
 import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmCompile
 
+tasks.withType<Test> {
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
```
max-kammerer: @sfs Please use Math.max

sfs:
I don't have a Windows machine to test this on, but this seems to be a problem in J2V8.

As far as I can tell, Gradle will launch multiple JUnit runner processes which will race to extract the j2v8 library and delete any library which is already present (see the extract method in https://github.com/eclipsesource/J2V8/blob/25440aa6b02b928da3a32d03df843df50dfae1ac/src/main/java/com/eclipsesource/v8/LibraryLoader.java#L75).

If that's the problem then there's an easy way to fix it, by passing a process specific tempDirectory to  com.eclipsesource.v8.V8.createV8Runtime in org.jetbrains.kotlin.js.test.interop.ScriptEngineV8. I'll submit an updated patch later, but I cannot test this myself.

sfs:
Seems to be a known issue in J2V8: https://github.com/eclipsesource/J2V8/issues/39

compiler/build.gradle.kts
sfs: Sure. The code is from the Gradle documentation by the way (https://guides.gradle.org/performance/#parallel_test_execution).

js/js.tests/build.gradle.kts
sfs: I would rather add a workaround if possible. If it's still broken after the latest changes I'll get rid of the parallelization for the JS build.

sfs:
Added a potential workaround, which still needs testing on Windows.

@max-kammerer could you please test this? Thanks.

max-kammerer:
@sfs Thank! Will run remote run on Teamcity!

max-kammerer:
@sfs Many Thanks! Great Work!

max-kammerer:
@sfs after merging pr we got a lot of randomly failed test in IDE related configurations (https://teamcity.jetbrains.com/viewType.html?buildTypeId=Kotlin_dev_Aggregate&branch_Kotlin_dev=%3Cdefault%3E&tab=buildTypeStatusDiv). So pr is temporary rollbacked during investigation

max-kammerer:
Some of problems relates to 'public static File tmpDir(String name) '  changes, cause idea tests sometimes try to reuse library from neighbour test. And with new changes such libraries are deleted after previous test execution

max-kammerer:
I started teamcity build with https://github.com/JetBrains/kotlin/commit/a35f2564d70f8a035a12bf136152e58a502c9125

max-kammerer:
PR is restored in master

sfs:
Thank you, @max-kammerer for fixing the IDE bugs! Great to have the PR merged. :)

PR #2197, opened 2019-03-20
JVM_IR: handle Nothing and Unit more consistently.

pyos:
* In blocks, discard the result of any statement that has a return type other than void. This was previously done by wrapping each statement into an "implicit Unit conversion" that was actually compiled down to a stack pop instead. If an expression happened to already have type Unit, however, such a conversion was not inserted, resulting in a stray reference on the stack. These conversions are now redundant and should probably be removed.

 * In assignments and non-exhaustive conditionals, materialize a Unit on the stack to avoid depth mismatches that trip up the bytecode validator. Because such expressions are generally used at block level (and, indeed, the frontend will reject a non-exhaustive conditional used as an expression), combined with the above change this results in no additional GETSTATIC opcodes, as they are immediately removed by the peephole optimizer.


udalov:
@pyos Please add your first/last name in `git config user.name`

pyos:
@udalov I've updated the name & rebased the commit onto master

max-kammerer:
@pyos Many Thanks!

PR #2198, opened 2019-03-21
Implement support for KClass fields in annotation classes in the JVM IR backend

sfs:
KClass and Array<KClass> fields in annotation classes have to be translated to use java.lang.Class on the JVM. In the old backend this translation happens in several different places during code generation. This patch consolidates the same logic in a lowering pass on the IR instead.

Note that the code does not fix the types of the descriptors, since descriptors seem to be deprecated and code generation still expects KClass descriptors.

The PR also does not remove the annotation class logic from code generation, since some of it is shared with the old backend.


compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -85,4 +88,9 @@ fun IrType.isPrimitiveArray() = isTypeFromKotlinPackage { it in FQ_NAMES.primiti
 
 fun IrType.getPrimitiveArrayElementType() = (this as? IrSimpleType)?.let {
     (it.classifier.owner as? IrClass)?.fqName?.toUnsafe()?.let { fqn -> FQ_NAMES.arrayClassFqNameToPrimitiveType[fqn] }
-}
\ No newline at end of file
+}
+
+fun IrType.isNonPrimitiveArray(): Boolean {
+    val descriptor = classifierOrNull?.descriptor as? ClassDescriptor ?: return false
+    return KotlinBuiltIns.isNonPrimitiveArray(descriptor)
```
max-kammerer: Please avoid descriptor usage,
here should be smth like
```
(this.isArray() || this.isNullableArray()) && !this.isPrimitiveArray()
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
```diff
@@ -5,26 +5,181 @@
...
+        for (declaration in irClass.declarations)
+            if (declaration is IrSimpleFunction)
+                lowerAnnotationField(declaration)
+
+        return irClass
+    }
+
+    // Lower the types on annotation class fields (KClass -> Class, Array<KClass> -> Array<Class>)
+    private fun lowerAnnotationField(declaration: IrSimpleFunction) {
+        val property = declaration.correspondingProperty ?: return
```
max-kammerer: val property = declaration.correspondingPropertySymbol?.owner ?: return

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
sfs: In my defense, I wrote the code before it was deprecated. :)
Anyways, changed it to use correspondingPropertySymbol.

max-kammerer:
@sfs Could you look at failed tests in https://teamcity.jetbrains.com/viewLog.html?buildId=2088071&tab=buildResultsDiv&buildTypeId=Kotlin_dev_Compiler

Seems this failings could be related to new commits from @gogabr  around AnnotationCodegen

gogabr:
This may be my fault -- since KClass arguments were broken while I wrote my version of annotation codegen, I could not properly test them.
I will have a look.

gogabr:
I don't see default values for KClass-typed properties in annotation classes. You can look at `FunctionCodegen.getAnnotationDefaultValueExpression()` to see where they should go.
(Yes, the data structure configuration is convoluted, but that is just a consequence of treating annotation classes mostly as normal classes.)

sfs:
It was a problem in my code. I forgot to carry over the initializers and didn't notice since they used to be accessed through the descriptors.

Everything should work now.

max-kammerer:
@sfs Many Thanks!

PR #2199, opened 2019-03-21
Remove unused SimpleDirtyData from incremental compilation

gavra0:
This is not used any more and all information about
the changed symbols and fqNames is passed alongside artifacts.


alexey-tsvetkov:

PR #2201, opened 2019-03-21
KT 14344: Suggest to replace manual range with explicit `indices` call

devbridie:
YouTrack: https://youtrack.jetbrains.com/issue/KT-14344

Adds inspection that suggests `0 .. collection.size-1` and `0 until collection.size` → `collection.indices`.

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -1817,6 +1817,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceManualRangeWithIndicesCallsInspection"
+                     displayName="Convert manual ranges to Collection.indices calls"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="INFO"
+                     language="kotlin"
+    />
+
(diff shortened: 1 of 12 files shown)


mglukhikh:
I have general point here, unbound to some concrete source code. There are cases when 
```
for (i in 0 until collection.size) { ... }
```

is better to be replaced with
```
for (element in collection) { ... }
```

without any indices at all.

devbridie:
@mglukhikh Should your suggestion be included in this PR?

mglukhikh:
Yes, I'd say it's better to do it, just to replace "bad" code with a "good" one, not with a kind of "not so bad but still not good".

mglukhikh:
Thank you, I think it's a good contribution now. I will test the inspection on the Kotlin codebase, fix bugs if any will be detected, and then merge into master.

devbridie:
I posted a bug I'm encountering on #kontributors, where the AST becomes invalid after executing this replacement. See this video that shows this behavior.

https://kotlinlang.slack.com/files/U18RKJMV2/FJ853K665/identifier.mp4

mglukhikh:
@devbridie So finally I merged it, thanks. However, I had to change a lot of minor details. And you are right about invalid AST, it occurs because you should, for example, replace parameter with parameter, not with identifier, and expression with expression, not with identifier again. See final version here: https://github.com/JetBrains/kotlin/commit/3451c60f9383756df33b6985f5c5a7a2435c1d90.

PR #2203, opened 2019-03-21
CoroutineNonBlockingContextChecker: prefer to check module language version instead of project one (KT-30320)

NikitaKatkov:
https://youtrack.jetbrains.com/issue/KT-30320


mglukhikh:
@NikitaKatkov Here we have the same situation as in https://github.com/JetBrains/kotlin/pull/2223. `CoroutineNonBlockingContextChecker.kt.182` is not changed which looks strange.

idea/src/org/jetbrains/kotlin/idea/inspections/blockingCallsDetection/CoroutineNonBlockingContextChecker.kt
```diff
@@ -30,7 +32,7 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 class CoroutineNonBlockingContextChecker : NonBlockingContextChecker {
 
     override fun isApplicable(file: PsiFile): Boolean {
-        val languageVersionSettings = file.project.getLanguageVersionSettings()
+        val languageVersionSettings = file.module?.languageVersionSettings ?: file.project.getLanguageVersionSettings()
```
mglukhikh: I think here you should write just `file.languageVersionSettings` without these hacks. If you look inside, you see it handles your case, as well as some other cases.

PR #2204, opened 2019-03-22
KT-30078 Add "Add getter/setter" quick fix for uninitialized property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30078

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/AddAccessorIntentions.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/AddAccessorIntentions.kt
@@ -16,18 +16,21 @@
 
 package org.jetbrains.kotlin.idea.intentions
 
+import com.intellij.codeInsight.intention.IntentionAction
 import com.intellij.codeInsight.intention.LowPriorityAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
 import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
+import org.jetbrains.kotlin.diagnostics.Diagnostic
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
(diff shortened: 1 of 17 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/AddAccessorIntentions.kt
```diff
@@ -45,25 +48,56 @@ abstract class AbstractAddAccessorsIntention(
...
             val getter = psiFactory.createPropertyGetter(expression)
-            if (element.setter != null)
+            val added = if (element.setter != null) {
                 element.addBefore(getter, element.setter)
-            else
+            } else {
                 element.add(getter)
+            }
+            if (!hasInitializer) {
+                (added as? KtPropertyAccessor)?.bodyBlockExpression?.statements?.firstOrNull()?.also {
```
dimonchik0036: `let` looks better for this case.

idea/src/org/jetbrains/kotlin/idea/intentions/AddAccessorIntentions.kt
```diff
@@ -45,25 +48,56 @@ abstract class AbstractAddAccessorsIntention(
...
+            }
         }
         if (addSetter) {
-            val expression = psiFactory.createBlock("field = value")
+            val expression = if (hasInitializer) psiFactory.createBlock("field = value") else psiFactory.createEmptyBody()
             val setter = psiFactory.createPropertySetter(expression)
-            element.add(setter)
+            val added = element.add(setter)
+            if (!hasInitializer && !addGetter) {
+                (added as? KtPropertyAccessor)?.bodyBlockExpression?.lBrace?.also {
```
dimonchik0036: Same as above.

dimonchik0036:
Thanks! https://github.com/JetBrains/kotlin/commit/f861b107988a9f228cb7dac5edf40f0e048724a2

PR #2205, opened 2019-03-22
only load inspections when groovy plugin is present KT-30579

nskvortsov:
otherwise, a user who disables Groovy support gets NoClassDefFound errors

Fixes https://youtrack.jetbrains.com/issue/KT-30579


mglukhikh:
Merged 👍

PR #2206, opened 2019-03-22
Rework cinterop linkerOpts/compilerOpts (KT-29970)

LepilkinaElena:


LepilkinaElena:
Fix for https://youtrack.jetbrains.com/issue/KT-29970

Other parts of change in other repos:

- https://github.com/JetBrains/kotlin-native/pull/2805
- https://github.com/JetBrains/kotlin-native-shared/pull/5

ilmat192:
If I understand correctly, there is no such option as `linkerOpt` in the current K/N build the big Kotlin depends on. So we also need to publish a new K/N build with https://github.com/JetBrains/kotlin-native/pull/2805 merged and update the dependency in [Kotlin Gradle plugin](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/NativeCompilerDownloader.kt#L28).

PR #2207, opened 2019-03-22
Label improper IEEE 754 tests.

ting-yuan:
Those tests don't expect ProperIeee754Comparisons.
See KT-22723 for details.


dnpetrov:
That's correct, thanks!

PR #2209, opened 2019-03-25
KT-30559 Redundant Getter, Redundant Setter: reduce range to getter/setter header

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30559


mglukhikh:
Merged 👍

PR #2211, opened 2019-03-25
Forward stdout and stderr logger of out of process though gradle logger

breskeby:
Instead of forwarding the external process output directly to stdout / stderr we forward the output through the gradle logging system which allows associate output to the producing gradle tasks. 

This fix in combination with https://github.com/gradle/gradle/pull/8807 will result in better user experience when using kotlin plugin with gradle and gradle build scans.

This is a fix for https://youtrack.jetbrains.com/issue/KT-30596


breskeby:
any update on this?

alexey-tsvetkov:
Merged manually,
@breskeby thanks for the contribution!

PR #2212, opened 2019-03-25
KT-26629 Inspection to replace `==` operator on Double.NaN with `equals` call

devbridie:
YouTrack: https://youtrack.jetbrains.com/issue/KT-26629

Replaces fragments like `5.0 == Double.NaN` with `5.0.equals(Double.NaN)`.

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -2365,6 +2365,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ConvertNaNEqualityInspection"
+                     displayName="Convert equality check with 'NaN' to 'isNaN' call"
+                     groupPath="Kotlin"
+                     groupName="Probable bugs"
+                     enabledByDefault="true"
+                     level="WARNING"
+                     language="kotlin"
+    />
+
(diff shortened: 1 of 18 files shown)


ilya-g:
- Have you considered using `number.isNaN()` function as a replacement?
- Does the inspection work for `!=` operator too? I believe it would be inconsistent to implement it only for `==`.

devbridie:
Indeed, using `isNaN` seems better. I have used this as a replacement and have added support for `!=`.

idea/resources/META-INF/plugin-common.xml
```diff
@@ -2350,6 +2350,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ConvertEqualsDoubleNaNInspection"
+                     displayName="Convert '== Double.NaN' to 'equals' call"
```
mglukhikh: This name contradicts with description and inspection logic.

idea/resources/META-INF/plugin-common.xml
```diff
@@ -2350,6 +2350,15 @@
...
+                     groupPath="Kotlin"
+                     groupName="Probable bugs"
```
mglukhikh: I think it would be good to explain in description why it's a probable bug.

devbridie:
Small ping, as I didn't comment when I hade made the changes!

mglukhikh:
Thank you, I think now it can be accepted.

PR #2213, opened 2019-03-25
KT-30389 Fix to convert argument to Int: suggest roundToInt()

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30389

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/RoundNumberFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/RoundNumberFix.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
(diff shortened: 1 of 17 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/RoundNumberFix.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.jetbrains.kotlin.types.typeUtil.isInt
+import org.jetbrains.kotlin.types.typeUtil.isLong
+
+class RoundNumberFix(
+    element: KtExpression,
+    type: KotlinType,
+    private val disableIfAvailable: IntentionAction? = null
+) : KotlinQuickFixAction<KtExpression>(element), LowPriorityAction {
+
+    private val isTarget = type.isLongOrInt() && element.analyze(BodyResolveMode.PARTIAL).getType(element)?.isDoubleOrFloat() == true
```
mglukhikh: Could you please add a pair of false-positive tests  (checking that quick-fix is NOT here when types do not match your condition)?

mglukhikh:
Merged manually, thank you.

PR #2214, opened 2019-03-25
Don't add implicit casts in return from constructor.

sfs:


compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt
```diff
@@ -107,7 +108,9 @@ open class InsertImplicitCasts(
 
     override fun visitReturn(expression: IrReturn): IrExpression =
         expression.transformPostfix {
-            value = value.cast(expression.returnTarget.returnType)
+            // Don't insert an implicit cast when returning from a (secondary) constructor.
+            if (expression.returnTargetSymbol !is IrConstructorSymbol)
+                value = value.cast(expression.returnTarget.returnType)
```
dnpetrov:
```
class C() {
    constructor(x: Any?) : this() {
        if (x is Unit) return x
    }
}
```
The artificial example above is considered well-formed, and contains smart cast to Unit.
However. this smart cast will not be represented in IR.
Actually should cast constructor return value to `kotlin.Unit`.

dnpetrov:
Please, fix the issue mentioned above (with smart cast to unit in constructor return value) and add a corresponding test. It doesn't have much practical value, but such corner case behavior should be represented in tests.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt
sfs: You're right, I've changed it to coerce the result to Unit instead.

sfs:
@dnpetrov I've updated the PR with the implicit cast to unit and an additional test.

dnpetrov:
Looks good to me. I'll run tests and merge it manually.

PR #2216, opened 2019-03-25
KT-30565 False positive "Suspicious 'var' property" inspection with annotated default property getter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30565


mglukhikh:
Merged 👍

PR #2217, opened 2019-03-25
TotalOrderEquals: Handle nulls

ting-yuan:


dnpetrov:
Looks good to me. I'll run tests and merge manually.

max-kammerer:
@dnpetrov Could we use here optimized variant similar to 'genIEEE754EqualForNullableTypesCall'?

dnpetrov:
@max-kammerer 
Ideally we should, especially since we know argument types.
Do we have bytecodeText tests for those specialized intrinsics?

max-kammerer:
@dnpetrov seems not,
even equalsNull_withExplicitFlag.kt test in non-ir backend generates unoptimized variant

dnpetrov:
Merged manually.

PR #2218, opened 2019-03-25

burakeregar:


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToStringTemplateIntention.kt
```diff
@@ -80,7 +80,11 @@ open class ConvertToStringTemplateIntention : SelfTargetingOffsetIndependentInte
         fun buildText(expr: KtExpression?, forceBraces: Boolean): String {
             if (expr == null) return ""
             val expression = KtPsiUtil.safeDeparenthesize(expr).let {
-                if ((it as? KtDotQualifiedExpression)?.isToString() == true) it.receiverExpression else it
+                when {
+                    (it as? KtDotQualifiedExpression)?.isToString() == true -> it.receiverExpression
+                    it is KtLambdaExpression -> expr
```
mglukhikh: I think it's good to check here that it's a lambda AND it has a label. For example, with lambda in brackets everything should work fine.

PR #2219, opened 2019-03-26
Fix parser for linkerOpts/compilerOpts (KT-29970)

LepilkinaElena:
This is same fix https://github.com/JetBrains/kotlin/pull/2206 but based on version of kotlin which works fine for kotlin-native.


ilmat192:
Just a reminder: we still need to build K/N first and add a dependency on it here: https://github.com/JetBrains/kotlin/pull/2206#issuecomment-476141745

PR #2221, opened 2019-03-26
JVM_IR: implement single-abstract-method conversions

pyos:


dnpetrov:
Since there're changes in psi2ir, I'd like to see psi2ir tests that cover them.

pyos:
@dnpetrov I've added a basic test. (The difference from the current behavior is the kind of the type conversion in arguments to array and augmented assignment operators: IMPLICIT_CAST -> SAM_CONVERSION).

dnpetrov:
@pyos Ok, looks good to me.

Other operator expressions which are desugared to method calls (`+` and so on) should already work.
I'd add corresponding psi2ir tests later.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+
+class SingleAbstractMethodLowering(val context: CommonBackendContext) : ClassLoweringPass {
+    override fun lower(irClass: IrClass) {
+        // TODO use `IrType`s; unfortunately, they are currently not comparable.
+        //      (Can't use class symbols here, we want type parameters as well.)
+        val cachedImplementations = mutableMapOf<Pair<KotlinType, KotlinType>, IrClass>()
+        val localImplementations = mutableListOf<IrClass>()
+        irClass.transformChildrenVoid(object : IrElementTransformerVoidWithContext() {
+            override fun visitCall(expression: IrCall): IrExpression {
+                if (expression.symbol.owner.descriptor !is SamConstructorDescriptor)
```
udalov: @dnpetrov Do we have, or can we add a descriptor-less way to detect that `IrFunction` is a SAM constructor? Maybe a new origin or something?

compiler/testData/codegen/boxAgainstJava/sam/differentFqNames.kt
```diff
@@ -1,5 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
-// WITH_RUNTIME
```
udalov: Please return `WITH_RUNTIME` in these tests:
* `boxAgainstJava/sam/differentFqNames.kt`
* `boxAgainstJava/sam/samConstructorGenericSignature.kt`
* `boxAgainstJava/sam/adapters/inheritedOverriddenAdapter.kt`

(see cb7727d51a)

udalov:
Please investigate failures in
* `IrInlineSuspendTestsGenerated.CallableReference.testOrdinaryParameter`
* `IrInlineSuspendTestsGenerated.CallableReference.testSuspendOfOrdinary`

```
TYPE_MISMATCH: Type mismatch: inferred type is KSuspendFunction0<String> but () -> String was expected (35,18) in Z:/buildAgent/temp/buildTmp/unitTest_ordinaryParameter/js-tests/JS_TESTS/box.kt
```

compiler/testData/codegen/boxAgainstJava/sam/differentFqNames.kt
pyos: Whoops, should've paid more attention when rebasing, sorry!

pyos:
@udalov must be flaky tests, since I've not touched any js-related code, and both pass locally anyway.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+                buildOverride: IrSimpleFunction.(List<IrField>) -> IrBody
+            ): IrClass {
+                val superClass = superType.classifierOrFail.owner as IrClass
+                // The language documentation prohibits casting lambdas to classes, but if it was allowed,
+                // the `irDelegatingConstructorCall` in the constructor below would need to be modified.
+                assert(superClass.kind == ClassKind.INTERFACE) { "SAM conversion to an abstract class not allowed" }
+
+                val superClassName = superClass.fqNameSafe.pathSegments().joinToString("_") { it.toString() }
+                val subclass = buildClass {
+                    // TODO this is not the name some tests (e.g. kt11519) expect
```
udalov: I suppose we could extract a bytecode text test out of that test, to prevent looking up a class by a non-stable name. Or maybe try and look up both names to make that test pass in JVM IR

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+                        type = fieldType
+                        origin = subclass.origin
+                        visibility = Visibilities.PRIVATE
+                    }.apply {
+                        parent = subclass
+                    }
+                }
+
+                subclass.declarations += fields
+                subclass.declarations += buildConstructor {
```
udalov: Maybe consider using (and improving if needed) the new declarations `addConstructor`, `addFunction`, etc which at least remove the need to specify parent every time

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+                return subclass
+            }
+
+            // Construct a class that wraps an invokable object into an implementation of an interface:
+            //     class sam$n(private val invokable: F) : Interface { override fun method(...) = invokable(...) }
+            private fun createObjectProxy(superType: IrType, invokableType: IrType): IrClass =
+                implement(superType, listOf(invokableType)) { fields ->
+                    context.createIrBuilder(symbol).irBlockBody(startOffset, endOffset) {
+                        val invokableClass = invokableType.classifierOrFail.owner as IrClass
+                        +irReturn(irCall(invokableClass.functions.single { it.name.toString() == "invoke" }).apply {
```
udalov: `it.name == OperatorNameConventions.INVOKE`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+                            var boundOffset = 0
+                            var offset = 0
+                            val arguments = reference.arguments.map {
+                                if (it != null)
+                                    irGetField(irGet(dispatchReceiverParameter!!), fields[boundOffset++])
+                                else
+                                    irGet(valueParameters[offset++])
+                            }.toMutableList()
+
+                            dispatchReceiver = reference.symbol.owner.dispatchReceiverParameter?.let { arguments.removeAt(0) }
```
udalov: This is minor, but please don't abuse `let`
```
if (reference.symbol.owner.dispatchReceiverParameter != null) {
    dispatchReceiver = arguments.removeAt(0)
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,237 @@
...
+
+internal val singleAbstractMethodPhase = makeIrFilePhase(
+    ::SingleAbstractMethodLowering,
+    name = "SingleAbstractMethod",
+    description = "Replace SAM conversions with instances of interface-implementing classes"
+)
+
+class SingleAbstractMethodLowering(val context: CommonBackendContext) : ClassLoweringPass {
+    override fun lower(irClass: IrClass) {
+        // TODO use `IrType`s; unfortunately, they are currently not comparable.
```
udalov: FYI we're looking into implementing `IrType.equals/hashCode` right now

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,231 @@
...
+
+            // Replace a SAM-conversion of a callable object with an instance of an appropriate implementation.
+            private fun createInstance(superType: IrType, invokable: IrExpression): IrExpression {
+                // Running in the class context, so empty scope means this is a field/anonymous initializer.
+                val scopeOwnerSymbol = currentScope?.scope?.scopeOwnerSymbol ?: irClass.symbol
+                return context.createIrBuilder(scopeOwnerSymbol).irBlock(invokable, null, superType) {
+                    // Not an exhaustive check, but enough to cover the most common cases.
+                    if (invokable is IrFunctionReference) {
+                        // Runnable(::function)
+                        createFunctionProxyInstance(superType, invokable)
```
max-kammerer: Please add bytecode test with number of generatedf class check

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -0,0 +1,231 @@
...
+                    } else if (invokable is IrBlock && invokable.statements.last() is IrFunctionReference) {
+                        // Runnable { lambda }
+                        for (statement in invokable.statements.dropLast(1))
+                            +statement
+                        createFunctionProxyInstance(superType, invokable.statements.last() as IrFunctionReference)
```
max-kammerer: And here

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
dnpetrov: Ideally, SamConstructorDescriptor should not be called in IR generated by psi2ir: SAM conversion and explicit SAM construction becomes IrTypeOperatorCall for IrTypeOperator.SAM_CONVERSION. 
If SamConstructorDescriptor calls "leak" to IR somehow, it's probably a psi2ir bug or underdesign on my side.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
pyos: Turns out the non-IR backend doesn't actually do that for references, only for lambdas. Should I remove this branch, or mark the test as IGNORE_BACKEND: JVM?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
pyos: Must be a bug; this seems to happen when the interface has type parameters, in which case the unbound descriptor is a SamConstructorDescriptor, but the one with substituted parameters is a SimpleFunctionDescriptor, so the call slips by the check in psi2ir.

pyos: Added a fix and a test. The additional IMPLICIT_CAST to `kotlin.Function2<T of java.util.Comparator?, T of java.util.Comparator?, kotlin.Int>` looks suspicious to me, though.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
max-kammerer: @pyos Yes, mark as ignored for current backend

compiler/testData/codegen/bytecodeText/sam/samWrapperForNullInitialization.kt
```diff
@@ -17,10 +16,8 @@ fun test() {
 }
 
 // @TestKt.class:
-// 1 NEW TestKt\$test\$1
```
max-kammerer: It's best keep smth like
 //1 NEW TestKt\$ 
and add // 2 NEW

compiler/testData/codegen/bytecodeText/sam/samWrapperForNullInitialization.kt
```diff
@@ -17,10 +16,8 @@ fun test() {
...
 // 1 NEW kotlin/jvm/internal/Ref\$IntRef
 // 2 NEW
-// 1 INVOKESPECIAL TestKt\$test\$1.<init>
```
max-kammerer: @pyos Same here 
// 1 INVOKESPECIAL TestKt\$[a-z$]*.<init>

pyos: This line seems redundant, given that calls to <init> are paired with NEW.

max-kammerer:
@pyos Thanks You!

PR #2222, opened 2019-03-26
KT-30451 Redundant call of selector in maxBy&minBy

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30451

### DIFF
--- a/libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
+++ b/libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
@@ -343,14 +343,15 @@ object Aggregates : TemplateGroupBase() {
         body {
             """
             val iterator = iterator()
+            if (!iterator.hasNext()) return null
 
-            var minElem: T? = null
-            var minValue: R? = null
+            var minElem = iterator.next()
+            if (!iterator.hasNext()) return minElem
+            var minValue = selector(minElem)
             while (iterator.hasNext()) {
                 val e = iterator.next()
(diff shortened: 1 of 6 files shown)


ilya-g:
On the second consideration, I now think it might be not so good idea to refactor these functions this way due to the risk of losing boxing elimination for primitive returning selector functions.

We need to ensure that making `maxValue`/`minValue` nullable doesn't prevent boxing elimination after `minBy`/`maxBy` is inlined when the selector returns a primitive value, e.g. by comparing the bytecode of `strings.minBy { it.length }` before and after this change.

ilya-g:
I've compared the old and the new implementation and unfortunately the new one indeed causes boxing of a primitive value returned by the `selector` function, because it has to store it in a nullable variable.  This would cause a significantly worse performance impact compared to inlining the `selector` function twice.

libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
```diff
@@ -346,6 +346,7 @@ object Aggregates : TemplateGroupBase() {
             if (!iterator.hasNext()) return null
 
             var minElem = iterator.next()
+            if (!iterator.hasNext()) return minElem
             var minValue = selector(minElem)
             while (iterator.hasNext()) {
```
ilya-g: Condition check now can be moved to the end of the loop by replacing `while(...) { }` loop with `do { } while(...)` to avoid checking `iterator.hasNext()` twice on the first iteration.

PR #2223, opened 2019-03-26
CoroutineNonBlockingContextChecker:

NikitaKatkov:
added check inside lambdas assigned to suspend properties and functions (KT-30540)

removed false positives when lambda inside suspend isn't actually called (KT-30321)

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/blockingCallsDetection/CoroutineNonBlockingContextChecker.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/blockingCallsDetection/CoroutineNonBlockingContextChecker.kt
@@ -8,7 +8,7 @@ package org.jetbrains.kotlin.idea.inspections.blockingCallsDetection
 import com.intellij.codeInspection.blockingCallsDetection.NonBlockingContextChecker
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
-import com.intellij.psi.util.PsiTreeUtil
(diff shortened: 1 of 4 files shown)


mglukhikh:
@NikitaKatkov You did not change `CoroutineNonBlockingContextChecker.kt.182` (same file for IDEA 182), is it intentional?

idea/src/org/jetbrains/kotlin/idea/inspections/blockingCallsDetection/CoroutineNonBlockingContextChecker.kt
```diff
@@ -40,9 +42,9 @@ class CoroutineNonBlockingContextChecker : NonBlockingContextChecker {
 
         val containingLambda = element.parents
             .firstOrNull { it is KtLambdaExpression && it.analyze().get(BindingContext.LAMBDA_INVOCATIONS, it) == null }
-        val containingArgument = PsiTreeUtil.getParentOfType(containingLambda, KtValueArgument::class.java)
+        val containingArgument = containingLambda?.getParentOfType<KtValueArgument>(true, KtCallableDeclaration::class.java)
         if (containingArgument != null) {
-            val callExpression = PsiTreeUtil.getParentOfType(containingArgument, KtCallExpression::class.java) ?: return false
+            val callExpression = containingArgument.getParentOfType<KtCallExpression>(true) ?: return false
```
mglukhikh: May be `getStrictParentOfType` or `getNonStrictParentOfType` is better here?

PR #2224, opened 2019-03-26
KT-30614: String templates suggest removing curly braces for backtick escaped identifiers

burakeregar:
https://youtrack.jetbrains.com/issue/KT-30614


goodwinnk:
Thank you for the contribution! I have merged it manually as https://github.com/JetBrains/kotlin/commit/9ab54d0e44181680b6255d7332dffddefbf09e82.

PR #2226, opened 2019-03-27
KT-12721: Add inspection for Integer.toString(i) to i.toString()

devbridie:
YouTrack: https://youtrack.jetbrains.com/issue/KT-12721

Converts `Integer.toString(5)` to `5.toString()`, taking radixes into account.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaIntegerToStringWithMemberInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaIntegerToStringWithMemberInspection.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameUnsafe
+
+class ReplaceJavaIntegerToStringWithMemberInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return dotQualifiedExpressionVisitor { dotQualifiedExpression ->
+            if (dotQualifiedExpression.isToString()) {
+                holder.registerProblem(
+                    dotQualifiedExpression,
+                    "Could be replaced with `int.toString()`",
+                    ConvertIntegerToStringQuickFix()
+                )
+            }
+        }
+    }
+}
+
+private class ConvertIntegerToStringQuickFix : LocalQuickFix {
+    override fun getName() = "Replace with 'int.toString()'"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val dotQualifiedExpression = descriptor.psiElement as? KtDotQualifiedExpression ?: return
+        val element = dotQualifiedExpression.selectorExpression as? KtCallExpression ?: return
+        val arguments = element.valueArguments
+        val integerArg = arguments.getOrNull(0) ?: return
+        val radixArg = arguments.getOrNull(1)
+
+        dotQualifiedExpression.replace(
+            KtPsiFactory(element).createExpressionByPattern(
+                "$0.toString($1)",
+                integerArg.getArgumentExpression() ?: return,
+                radixArg?.getArgumentExpression() ?: ""
+            )
+        )
+    }
+}
+
+private fun KtDotQualifiedExpression.isToString(): Boolean {
+    val calleeExpression = (selectorExpression as? KtCallExpression)?.calleeExpression ?: return false
+    val nameReference = calleeExpression as? KtNameReferenceExpression ?: return false
+    if (nameReference.getReferencedName() != "toString") return false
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaIntegerToStringWithMemberInspection.kt
```diff
@@ -0,0 +1,56 @@
...
+                "$0.toString($1)",
+                integerArg.text,
+                radixArg?.text ?: ""
+            )
+        )
+    }
+}
+
+private fun KtDotQualifiedExpression.isToString(): Boolean {
+    val fqName = this.resolveToCall()?.resultingDescriptor?.fqNameUnsafe?.asString()
```
goodwinnk: I suggest to check presence of "toString" in the call before calling `resolveToCall()`, this way it would be possible to avoid calling resolve for most irrelevant references.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaIntegerToStringWithMemberInspection.kt
```diff
@@ -0,0 +1,56 @@
...
+private class ConvertIntegerToStringQuickFix : LocalQuickFix {
+    override fun getName() = "Replace with 'int.toString()'"
+
+    override fun getFamilyName() = name
+
+    override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+        val dotQualifiedExpression = descriptor.psiElement as? KtDotQualifiedExpression ?: return
+        val element = dotQualifiedExpression.selectorExpression as? KtCallExpression ?: return
+        val arguments = element.valueArguments
+        val integerArg = arguments[0]
```
goodwinnk: It's possible to get an `ArrayIndexOutOfBoundsException` here (try to do quick "Del" and "Alt + Enter" on `Integer.toString(<caret>1)` and apply the fix from popup).

Any checks that were done in intention can be outdated at the moment of quick fix execution.

idea/testData/inspectionsLocal/replaceJavaIntegerToStringWithMemberInspection/replaceIntToString.kt
```diff
@@ -0,0 +1,5 @@
+// WITH_RUNTIME
```
goodwinnk: Some test when inspection isn't applicable might be useful, I suggest to do one with invalid code `Integer.toString(<caret>)`, as `isToString()` implicitly suggests that this case won't be resolved to a needed qualified name.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaIntegerToStringWithMemberInspection.kt
```diff
@@ -0,0 +1,56 @@
...
+        val radixArg = arguments.getOrNull(1)
+
+        dotQualifiedExpression.replace(
+            KtPsiFactory(element).createExpressionByPattern(
+                "$0.toString($1)",
+                integerArg.text,
```
goodwinnk: A quick fix does a wrong thing when compound expression is used in arguments:

`Integer.toString(1 + 3)`

This can be fixed by giving actual expressions to `createExpressionByPattern`:

```
KtPsiFactory(element).createExpressionByPattern(
                "$0.toString($1)",
                integerArg.getArgumentExpression() /* check for null and exit */,
                radixArg?.getArgumentExpression() ?: ""
            )
```

`createExpressionByPattern` has logic for wrapping expressions to `()` if needed.

devbridie:
One question: this inspection will incorrectly flag the following call coinciding with the name `toString`:

```
fun Integer.toString(b: String) = ""

fun main() {
    val test = <caret>Integer.toString("")
}
```

How can this be prevented?

idea/testData/inspectionsLocal/replaceJavaIntegerToStringWithMemberInspection/replaceIntToString.kt
devbridie: I have added this test, adding the directive:
```
// ERROR: None of the following functions can be called with the arguments supplied: <br>public open fun toString(i: Int): String! defined in java.lang.Integer<br>public open fun toString(i: Int, radix: Int): String! defined in java.lang.Integer
```
Is this correct? It looks fishy to me.

goodwinnk: I agree it doesn't look very nice to have a full error message in the test data. Maybe only message code should be there (no need to implement this in this PR).

It was decided to list errors in the test data file to prevent a situation when an intention is not applicable not because of program logic, but because of compile errors in the test data file.

goodwinnk:
> How can this be prevented? I have added this test in the commit.

A nice catch! Let's activate the inspection only for calls that have `resolveToCall.status.isSuccess` equals to `true`.

devbridie:
@goodwinnk, thanks for your patience. Adding this check does not solve this test case, as the resolved call is valid, but does not point to a method descriptor that makes sense for this inspection, but instead to the defined extension method.

devbridie:
Rebased to current master.

goodwinnk:
> @goodwinnk, thanks for your patience. Adding this check does not solve this test case, as the resolved call is valid, but does not point to a method descriptor that makes sense for this inspection, but instead to the defined extension method.

It looks I wasn't patient enough and didn't notice an extension method. My comment was a solution that helps preventing reporting for badly resolved method when there's a type mismatch error on a parameter.

```
fun main() {
    java.lang.Integer.toString("")
}
```

Considering your example with the custom extension. I can't reproduce it right now (in my case resolve still goes to original `toString(int)` and reports type mismatch). Checking `resolveToCall.status.isSuccess` should also help.

devbridie:
I see, my counter example doesn't make sense; I was under the impression that you could overload parameter types using extension functions, but this is not the case.

I have updated the PR once again to add `resolveToCall.status.isSuccess`.

goodwinnk:
Thank you for the contribution. I have relayed your commit manually: https://github.com/JetBrains/kotlin/commit/efcc6f096781b68053e8d2d6186db4274db67b11.

ilya-g:
Is it possible to extend this inspection for other numeric types?

devbridie:
I have done this in #2239.

PR #2227, opened 2019-03-27
KT-25439: Add inspection "Map replaceable with EnumMap"

devbridie:
YouTrack: https://youtrack.jetbrains.com/issue/KT-25439

Offers to change
```
class MyClass {
	enum class MyEnum { A, B, C }
	var map: Map<MyEnum, *> = HashMap<MyEnum, Any>()
}
```

For `EnumMap`, where valid.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithEnumMapInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithEnumMapInspection.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElementVisitor
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.caches.resolve.resolveImportReference
+import org.jetbrains.kotlin.idea.project.platform
+import org.jetbrains.kotlin.idea.util.ImportInsertHelper
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.callExpressionVisitor
+import org.jetbrains.kotlin.psi.createExpressionByPattern
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameUnsafe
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithEnumMapInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+private val hashMapCreationFqNames = setOf(
+    "java.util.HashMap.<init>",
+    "kotlin.collections.HashMap.<init>",
+    "kotlin.collections.hashMapOf"
+)
+
+class ReplaceWithEnumMapInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return callExpressionVisitor { element ->
+            if (element.platform !is JvmPlatform) return@callExpressionVisitor
```
mglukhikh: This code with a lot of `return@...` will become nicer if you convert lambda to anonymous function

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithEnumMapInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+            val context = element.analyze()
+            val fqName = element.getResolvedCall(context)?.resultingDescriptor?.fqNameUnsafe?.asString() ?: return@callExpressionVisitor
+            if (!hashMapCreationFqNames.contains(fqName)) return@callExpressionVisitor
+            if (element.valueArguments.size != 0) return@callExpressionVisitor
```
mglukhikh: Please use `isNotEmpty()`

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceWithEnumMapInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameUnsafe
+import org.jetbrains.kotlin.resolve.jvm.platform.JvmPlatform
+import org.jetbrains.kotlin.types.typeUtil.isEnum
+
+private val hashMapCreationFqNames = setOf(
+    "java.util.HashMap.<init>",
+    "kotlin.collections.HashMap.<init>",
+    "kotlin.collections.hashMapOf"
```
mglukhikh: What about `mapOf` and `linkedMapOf`, as well as `LinkedHashMap.<init>`?

devbridie: These cases are more difficult, as `mapOf(Enum.A to 5, Enum.B to 7)` does not generate nice code with `EnumMap`, it would probably become something like `EnumMap().let { it[Enum.A] = 5; it[Enum.B] = 7;` Furthermore, `mapOf` implies immutability, where `EnumMap` is mutable.

With `LinkedHashMap`, behavior is different: `LinkedHashMap` guarantees the maintence of insertion order, where EnumMap has natural enum order. 

These are the reasons for which I excluded these method calls. If it is not a problem, then I will at them anyways.

mglukhikh: Clear, I think we should leave it as is.

mglukhikh:
Merged manually, thank you. BTW you forgot an inspection description.

PR #2229, opened 2019-03-28
JVM_IR: remove most uses of StackValue from ExpressionCodegen

pyos:
(Probably easier to look at each commit separately.)

Visiting a node with ExpressionCodegen now produces a PromisedValue, a stripped down version of StackValue. It is not guaranteed to be fully materialized, which helps a lot with some optimizations, in particular those of chained boolean operations. Currently, however, a *partially* materialized value can be returned, so when calling `accept` twice in a row, the first result must be materialized explicitly to avoid interleaving the instruction streams. This might be a barrier to optimizing (n > 1)-ary intrinsics, so perhaps it'd be better to switch to always generating lazy values later.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -879,14 +879,7 @@ class ExpressionCodegen(
         val tryBlockEnd = markNewLabel()
         val tryBlockGaps = tryInfo?.gaps?.toList() ?: listOf()
         val tryCatchBlockEnd = Label()
-        if (tryInfo != null) {
-            data.handleBlock { tryInfo.onExit.accept(this, data).discard() }
-            tryInfo.onExit.markLineNumber(startOffset = false)
-            mv.goTo(tryCatchBlockEnd)
-            tryInfo.gaps.add(tryBlockEnd to markNewLabel())
-        } else {
-            mv.goTo(tryCatchBlockEnd)
-        }
+        mv.goTo(tryCatchBlockEnd)
 
         val catches = aTry.catches
         for (clause in catches) {
@@ -909,12 +902,7 @@ class ExpressionCodegen(
                 index
             )
 
-            if (tryInfo != null) {
-                data.handleBlock { tryInfo.onExit.accept(this, data).discard() }
-                tryInfo.onExit.markLineNumber(startOffset = false)
-                mv.goTo(tryCatchBlockEnd)
-                tryInfo.gaps.add(clauseEnd to markNewLabel())
-            } else if (clause != catches.last()) {
+            if (tryInfo != null || clause !== catches.last()) {
                 mv.goTo(tryCatchBlockEnd)
             }
 
@@ -925,6 +913,8 @@ class ExpressionCodegen(
             // Generate `try { ... } catch (e: Any?) { <finally>; throw e }` around every part of
             // the try-catch that is not a copy-pasted `finally` block.
             val defaultCatchStart = markNewLabel()
+            // TODO remove this hack; it is currently needed to avoid confusing the inliner.
+            tryInfo.gaps.add(defaultCatchStart to defaultCatchStart)
             // While keeping this value on the stack should be enough, the bytecode validator will
             // complain if a catch block does not start with ASTORE.
             val savedException = frame.enterTemp(AsmTypes.JAVA_THROWABLE_TYPE)
@@ -937,12 +927,17 @@ class ExpressionCodegen(
             frame.leaveTemp(AsmTypes.JAVA_THROWABLE_TYPE)
             mv.athrow()
 
-            // Include the ASTORE into the covered region. That's what javac does as well.
+            // Covering the ASTORE is not necessary, it is part of the above hack.
             genTryCatchCover(defaultCatchStart, tryBlockStart, finallyStart, tryInfo.gaps, null)
         }
 
         mv.mark(tryCatchBlockEnd)
-        // TODO: generate a common `finally` for try & catch blocks here? Right now this breaks the inliner.
+        if (tryInfo != null) {
+            data.handleBlock { tryInfo.onExit.accept(this, data).discard() }
+            // TODO should probably also add this breakpoint to other copies of `finally`?
+            tryInfo.onExit.markLineNumber(startOffset = false)
+            mv.nop()
+        }
         return tryResult
     }
 
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -67,30 +67,34 @@ class TryInfo(val tryBlock: IrTry) : ExpressionInfo(tryBlock) {
 }
 
 class BlockInfo private constructor(val parent: BlockInfo?) {
-    val variables: MutableList<VariableInfo> = mutableListOf()
-
-    private val infos = Stack<ExpressionInfo>()
+    val variables = mutableListOf<VariableInfo>()
+    val infos = Stack<ExpressionInfo>()
 
     fun create() = BlockInfo(this).apply {
         this@apply.infos.addAll(this@BlockInfo.infos)
     }
 
-    fun addInfo(loop: ExpressionInfo) {
-        infos.add(loop)
-    }
-
-    fun removeInfo(info: ExpressionInfo) {
-        assert(peek() == info)
-        pop()
-    }
-
-    fun pop(): ExpressionInfo = infos.pop()
-
-    fun peek(): ExpressionInfo = infos.peek()
-
     fun isEmpty(): Boolean = infos.isEmpty()
 
     fun hasFinallyBlocks(): Boolean = infos.firstIsInstanceOrNull<TryInfo>() != null
+
+    inline fun <R> withBlock(info: ExpressionInfo, f: (ExpressionInfo) -> R): R {
+        infos.add(info)
+        try {
+            return f(info)
+        } finally {
+            infos.pop()
+        }
+    }
+
+    inline fun <R> handleBlock(f: (ExpressionInfo) -> R): R {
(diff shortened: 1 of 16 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -867,150 +805,98 @@ class ExpressionCodegen(
...
-        gen(aTry.tryResult, aTry.asmType, data)
+        val tryResult = aTry.tryResult.accept(this, data).coerce(aTry.asmType).materialized
         val tryBlockEnd = markNewLabel()
-
-        val tryRegions = getCurrentTryIntervals(tryInfo, tryBlockStart, tryBlockEnd)
-
+        val tryBlockGaps = tryInfo?.gaps?.toList() ?: listOf()
         val tryCatchBlockEnd = Label()
-        genFinallyBlockOrGoto(tryInfo, tryCatchBlockEnd, null, data)
+        mv.goTo(tryCatchBlockEnd)
```
max-kammerer: @pyos It's best to avoid any changes in finally block scheme generation. Old scheme is necessary for Inliner to  process non-locals returns (they are not supported yet in IR backend, testData/codegen/boxInline/nonLocalReturns).
Similar problem could be with new logic of gap interval processing. 
So I would prefer to posptone these to changes until non-local return support in Inliner.

pyos: Yes, I've noticed that, which is why there's an additional insertion of an empty gap (i.e. a TRYCATCHBLOCK that covers nothing and jumps to itself) at line 934. This is enough for the inliner to detect the finally block, although it seems to not work anyway because of a missing constant load of what I assume is the length of the finally right after ASTORE.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -60,110 +52,211 @@ import org.jetbrains.org.objectweb.asm.Type
...
+        assert(lineNumber > 0)
+        if (lastLineNumber != lineNumber) {
+            lastLineNumber = lineNumber
+            mv.visitLineNumber(lineNumber, markNewLabel())
+        }
+    }
+
+    // TODO remove
+    fun gen(expression: IrElement, type: Type, data: BlockInfo): StackValue {
+        expression.accept(this, data).coerce(type).materialize()
```
udalov: It does look like this pattern (`accept` + `coerce` + `materialize`) is needed in a lot of places though. So I'd consider not removing this function but renaming it (`generateMaterialized` or something shorter) and using everywhere instead

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -60,110 +52,211 @@ import org.jetbrains.org.objectweb.asm.Type
...
+fun PromisedValue.discard(): MaterialValue {
+    materialize()
+    if (type !== Type.VOID_TYPE)
+        AsmUtil.pop(codegen.mv, type)
+    return MaterialValue(codegen, Type.VOID_TYPE)
+}
+
+// On materialization, cast the value to a different type.
+fun PromisedValue.coerce(target: Type) = when (target) {
+    Type.VOID_TYPE -> discard()
```
udalov: Why are we materializing the value if the expected type is void, and don't materialize it otherwise? This doesn't look very obvious

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -60,110 +52,211 @@ import org.jetbrains.org.objectweb.asm.Type
...
+    }
+}
 
-    fun peek(): ExpressionInfo = infos.peek()
+class VariableInfo(val declaration: IrVariable, val index: Int, val type: Type, val startLabel: Label)
 
-    fun isEmpty(): Boolean = infos.isEmpty()
+// A value that may not have been fully constructed yet. The ability to "roll back" code generation
+// is useful for certain optimizations.
+abstract class PromisedValue(val codegen: ExpressionCodegen, val type: Type) {
```
udalov: Let's please move this class and subclasses to a separate file

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -60,110 +52,211 @@ import org.jetbrains.org.objectweb.asm.Type
...
+        get() = type.asmType
+
+    // Assume this expression's result has already been materialized on the stack
+    // with the correct type.
+    val IrExpression.onStack: MaterialValue
+        get() = MaterialValue(this@ExpressionCodegen, asmType)
 
-    private val sourceManager = classCodegen.context.psiSourceManager
+    val voidValue: MaterialValue
+        get() = MaterialValue(this@ExpressionCodegen, Type.VOID_TYPE)
```
udalov: `this@ExpressionCodegen` -> `this`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: This branch was probably necessary to avoid breaking something while I was iterating on the code, but doesn't seem to affect anything now, so I'll remove it.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -0,0 +1,80 @@
...
+package org.jetbrains.kotlin.backend.jvm.codegen
+
+import org.jetbrains.kotlin.codegen.AsmUtil
+import org.jetbrains.kotlin.codegen.StackValue
+import org.jetbrains.org.objectweb.asm.Label
+import org.jetbrains.org.objectweb.asm.Type
+
+// A value that may not have been fully constructed yet. The ability to "roll back" code generation
+// is useful for certain optimizations.
+abstract class PromisedValue(val codegen: ExpressionCodegen, val type: Type) {
```
udalov: It looks like this class and its subclasses depend only on the `mv` field of `ExpressionCodegen`? Can we change this to `MethodVisitor` then, to allow it to be possibly used in more contexts?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ArrayConstructor.kt
```diff
@@ -34,10 +34,14 @@ object ArrayConstructor : IntrinsicMethod() {
     ): IrIntrinsicFunction {
         return object : IrIntrinsicFunction(expression, signature, context, expression.argTypes(context)) {
 
-            override fun invoke(v: InstructionAdapter, codegen: ExpressionCodegen, data: BlockInfo): StackValue =
-                codegen.generateCall(expression, this, data).let {
-                    return StackValue.onStack(Type.getObjectType("[" + AsmTypes.OBJECT_TYPE.internalName))
+            override fun invoke(v: InstructionAdapter, codegen: ExpressionCodegen, data: BlockInfo): StackValue {
+                // TODO fix this hack
```
udalov: 1) Why `with` is necessary here instead of just `codegen.generateCall(...).materialize()`?
2) I think `ExpressionCodegen.generateCall` should always materialize the generated value, at least until we have use cases to optimize away unneeded calls. Moreover, maybe we could have some sort of a convention that methods named `generateX` always return an already materialized value

pyos: It used to be an extension method StackValue.materialize defined inside ExpressionCodegen. `with` is redundant now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: @pyos I'm now in process of fixing basic non-local returns in inliner. It would be new fails with changed logic around finally generations and gaps

max-kammerer:
@pyos I've just merged  commits from 1 to 7 and 9 to partly avoid merge  problems with my branch.
It's not finished (rr/mb/nlr2, 4 failed test for Inliner), but I expect that refactorings around 'finally' block cause problems with Inliner.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -233,41 +247,34 @@ class ExpressionCodegen(
...
 
-    override fun visitContainerExpression(expression: IrContainerExpression, data: BlockInfo): StackValue {
-        val result = visitStatementContainer(expression, data)
-        return coerceNotToUnit(result.type, result.kotlinType, expression.type.toKotlinType())
-    }
+    override fun visitContainerExpression(expression: IrContainerExpression, data: BlockInfo) =
+        visitStatementContainer(expression, data).coerce(expression.asmType)
 
-    override fun visitMemberAccess(expression: IrMemberAccessExpression, data: BlockInfo): StackValue {
+    override fun visitFunctionAccess(expression: IrFunctionAccessExpression, data: BlockInfo): PromisedValue {
```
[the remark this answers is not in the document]
pyos: Looks like this is only invoked via a super call in visitDelegatingConstructorCall, so the two functions can be merged.

pyos:
@max-kammerer cool, I'll create a new PR with the rest of the commits then. I don't think 7 could affect the inliner as it produces the same bytecode; 20 might, so I'm going to remove it for now.

PR #2230, opened 2019-03-29
Fix a memory leak in PerModulePackageCacheService.onTooComplexChange

sprigogin:
Dispose all disposable objects created by the packageExists method before calling cache.clear() in the onTooComplexChange method.

https://youtrack.jetbrains.com/issue/KT-30688


semoro:
There is an alternative fix, scheduled for 1.3.40
It basically does the same, but without explicit dispose, see `https://github.com/JetBrains/kotlin/commit/6173feb9b63112e0fb8f0e8294fcd780275d3334`

But anyway thank you for the PR, as it was base to this fix 
Closing PR now

PR #2231, opened 2019-03-29
kapt: fix kt-26203

ting-yuan:
Gradle worker throws NPE when the parameter to the runnable is null.
This patch works that around by passing URL spec as String instead of
File?

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kapt3IT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kapt3IT.kt
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.gradle
 import org.jetbrains.kotlin.gradle.tasks.USING_JVM_INCREMENTAL_COMPILATION_MESSAGE
 import org.jetbrains.kotlin.gradle.util.*
 import org.junit.Assert
+import org.junit.Assume
 import org.junit.Test
 import java.io.File
 import java.util.zip.ZipFile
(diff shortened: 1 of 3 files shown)


ting-yuan:
@yanex @alexey-tsvetkov The bug is about JDK 10/11 where tools.jar isn't available. I'm trying to write a test but I don't even know if it is possible to specify a JDK version in integration tests. Locally, I can use JAVA_HOME and jvmTarget, though. Any thoughts?

alexey-tsvetkov:
@ting-yuan there are `JDK_10`, `JDK_11` extra properties set in Kotlin Project build process (see https://github.com/JetBrains/kotlin/blob/master/build.gradle.kts#L113).

I think you can pass them to `Test` tasks in `kotlin-gradle-plugin-integration-tests` via [systemProperties](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:systemProperties). Then you should be able to pass a custom JDK path from a test process to a test build process via `-Porg.gradle.java.home`.

ting-yuan:
@alexey-tsvetkov Thanks for the hints! Please take a look.

In case JDK_10 or JDK_11 are not available or not specified, the tests will pass silently. Otherwise all developers would need JDK_10 and JDK_11 if they don't want to see 2 failures sticking around. Does that make sense?

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kapt3IT.kt
```diff
@@ -67,6 +67,32 @@ class Kapt3WorkersIT : Kapt3IT() {
             assertSuccessful()
         }
     }
+
+    private fun testSimpleWithCustomJdk(gradleVersion: String, javaHome: File) {
+        val gradleVersionRequired = GradleVersionRequired.AtLeast(gradleVersion)
+
+        // Skip test if the specified jdk isn't available in the test environment.
+        if (!javaHome.isDirectory)
+            return
```
alexey-tsvetkov: I recommend to replace `return` with `org.junit.Assume#assumeTrue`.
When assumption is violated, a test is also skipped (it is not considered failed),
but it is shown as ignored in IJ and in TeamCity test reports.

alexey-tsvetkov:
@ting-yuan `org.junit.Assume#assumeTrue` can be used instead of `return`. Otherwise the test looks good to me.

ting-yuan:
Thanks, I didn't know that it can be done in that way, which is way better. Please take a look again.

ting-yuan:
@alexey-tsvetkov Just in case this is buried in other PRs, could you please take a look? :)

yanex:
Thanks for the PR! Merged manually.

PR #2232, opened 2019-03-29
KT-18629 Add BFS File walking.

aivinog1:


libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -12,6 +12,20 @@ import java.io.File
...
 
+/**
+ * An enumeration to describe possible algorithms to traverse.
+ * There are two of them: search in breadth (same directory depth)
+ */
+public enum class WalkAlgorithm {
+    /**
+     * Breadth-first search
+     */
+    BFS,
```
sschuberth: I actually believe we should write these out as `BREADTH_FIRST` and `DEPTH_FIRST`, just like we have `TOP_DOWN` instead of `TD` below.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -12,6 +12,20 @@ import java.io.File
...
+    /**
+     *  Depth-first search
+     */
+    DFS
+}
```
sschuberth: Nit: Missing blank line after this one.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -21,29 +35,29 @@ public enum class FileWalkDirection {
     /** Depth-first search, directory is visited BEFORE its files */
     TOP_DOWN,
     /** Depth-first search, directory is visited AFTER its files */
-    BOTTOM_UP
-    // Do we want also breadth-first search?
+    BOTTOM_UP;
```
sschuberth: Nit: Unnecessary change.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -21,29 +35,29 @@ public enum class FileWalkDirection {
...
 }
 
 /**
  * This class is intended to implement different file traversal methods.
  * It allows to iterate through all files inside a given directory.
  *
- * Use [File.walk], [File.walkTopDown] or [File.walkBottomUp] extension functions to instantiate a `FileTreeWalk` instance.
+ * Use [File.walk], [File.walkTopDown],[File.walkBottomUp],[File.walkBottomUpWithAlgorithm] or [File.walkTopDownWithAlgorithm] extension functions to instantiate a `FileTreeWalk` instance.
```
sschuberth: Nit: Use a space after `,` and hard-wrap line at 120 chars.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -77,7 +92,13 @@ public class FileTreeWalk private constructor(
         }
 
         override fun computeNext() {
-            val nextFile = gotoNext()
+            val nextFile: File?
+            if (algorithm.equals(WalkAlgorithm.BFS)) {
```
sschuberth: It's probably nicer to use `if` as an expression here:

            val nextFile = if (algorithm.equals(WalkAlgorithm.BFS)) {
                ++currentDepth
                gotoNextBfs()
            } else {
                gotoNextDfs()
            }

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -92,14 +113,14 @@ public class FileTreeWalk private constructor(
             }
         }
 
-        private tailrec fun gotoNext(): File? {
+        private tailrec fun gotoNextDfs(): File? {
```
sschuberth: Again, I'd prefer to spell out `Dfs`, and likewise for `Bfs` below.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -108,11 +129,31 @@ public class FileTreeWalk private constructor(
...
+            if (listFiles != null) {
+                for (fileItem in listFiles) {
+                    if (fileItem.isDirectory) {
+                        state.add(directoryState(fileItem))
+                    } else {
+                        state.add(SingleFileState(fileItem))
+                    }
+                }
+            }
+            return file
```
sschuberth: How about increment `currentDepth` already as part of this function? Then we wouldn't need to do it in `computeNext()` above.

sschuberth:
Also, it would be nice if the merge commits would be removed from this PR for more convenient commit-wise review.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree. But do I need to remain this comment? I think I should remove it.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Agree.

ilya-g:
I wonder what are the use cases for bottom-up breadth first search?

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > I actually believe we should write these out as `BREADTH_FIRST` and `DEPTH_FIRST`, just like we have `TOP_DOWN` instead of `TD` below.

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > Nit: Missing blank line after this one.

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > Nit: Use a space after `,` and hard-wrap line at 120 chars.

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > It's probably nicer to use `if` as an expression here:
> 
> ```
>         val nextFile = if (algorithm.equals(WalkAlgorithm.BFS)) {
>             ++currentDepth
>             gotoNextBfs()
>         } else {
>             gotoNextDfs()
>         }
> ```

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > Again, I'd prefer to spell out `Dfs`, and likewise for `Bfs` below.

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: > How about increment `currentDepth` already as part of this function? Then we wouldn't need to do it in `computeNext()` above.

Done.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -108,11 +137,40 @@ public class FileTreeWalk private constructor(
...
+         */
+        private fun gotoNextBreadthFirst(): File? {
+            val walkState = state.poll() ?: return null
+            val file = walkState.root
+            if (currentDepth >= maxDepth) {
+                return file
+            }
+
+            // To process maximum depth
+            val nameCount = file.toComponents().size
```
ilya-g: I believe, the next depth level could be easily determined without splitting each file to components: instead another queue for the next level should be introduced. When "this level" queue becomes empty, it is swapped with "the next level" queue and then processing continues with that next level queue.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
```diff
@@ -108,11 +137,40 @@ public class FileTreeWalk private constructor(
...
+                previousDepth = nameCount
+                currentDepth++
+            }
+            val listFiles = file.listFiles()
+            if (listFiles != null) {
+                for (fileItem in listFiles) {
+                    if (fileItem.isDirectory) {
+                        state.add(directoryState(fileItem))
+                    } else {
+                        state.add(SingleFileState(fileItem))
```
ilya-g: How listing files from the root of this `SingleFileState` is prevented later?

libraries/stdlib/jvm/test/io/FileTreeWalks.kt
```diff
@@ -459,4 +474,110 @@ class FileTreeWalkTest {
...
+                namesTopDownWithBfs.add(name)
+            }
+            assertEquals(referenceNames, namesTopDownWithBfs)
+            referenceNames.equalsOneToOne(namesTopDownWithBfs)
+        } finally {
+            basedir.deleteRecursively()
+        }
+    }
+
+    private fun Set<String>.equalsOneToOne(set: Set<String>) {
```
ilya-g: Why this comparison that treats sets as lists is required? Could these be lists from the start?

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
ilya-g: Or perhaps some other technique could be used to determine where the next level of depth is reached even in the single queue. For example by incrementing a next-level counter when enqueueing items and decrementing this-level counter when dequeueing items.

ilya-g:
I propose making BFS top-down visiting order just another entry in `FileWalkDirection` enum. This way it will require much less changes in the public API surface.

As to bottom-up BFS visitor — it could be postponed until its use cases are clear.

ihdadi3:
قناة شليلة[قناةشليلة ](قناة شليلة `)`

libraries/stdlib/jvm/test/io/FileTreeWalks.kt
ihdadi3:
```suggestion
     متعة خاصة تعيين <سلسلة>. equalsOneToOne ( مجموعة :  تعيين < سلسلة >) {
```
```suggestion
     متعة خاصة تعيين <سلسلة>. equalsOneToOne ( مجموعة :  تعيين < سلسلة >) {
```
```suggestion
     متعة خاصة تعيين <سلسلة>. equalsOneToOne ( مجموعة :  تعيين < سلسلة >) {قناةشليلة [+](url)
```

ilya-g:
@ihdadi3 please use English.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Ok, I need to think about it and I will choose better technique.

libraries/stdlib/jvm/test/io/FileTreeWalks.kt
aivinog1: @ilya-g Sure, this sets could be lists I will fix that.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: Hmm, I need to think about it.

aivinog1:
> I propose making BFS top-down visiting order just another entry in `FileWalkDirection` enum. This way it will require much less changes in the public API surface.
> 
> As to bottom-up BFS visitor — it could be postponed until its use cases are clear.

I'm agree with you.

libraries/stdlib/jvm/src/kotlin/io/files/FileTreeWalk.kt
aivinog1: @ilya-g What do you think about store current depth of node in `kotlin.io.FileTreeWalk.WalkState`? This class is private (and all his childrens), so public API will not be affected.

ilya-g: @aivinog1 That's a viable solution too, though it could slightly increase memory consumption.

Btw, it might be easier to not try to shoehorn BFS into the existing WalkState classes, but rather write BFS algorithm separately with the coroutine-based sequence and iterator builders.

E.g. in `FileTreeWalk.iterator` function:
```
fun iterator() = if (direction == BFS) bfsIterator() else FileTreeWalkIterator()

private fun bfsIterator(): Iterator<File> = iterator {
   // write BFS iteration algorithm imperatively here, 
   // use yield and yieldAll functions to returns files to sequence consumer
}
```
If you find this easier, you can go this way. But if not, it's not a problem: you can continue with the current approach and we'll investigate this refactoring separately.

sschuberth:
> I wonder what are the use cases for bottom-up breadth first search?

See e.g. the discussion at https://github.com/heremaps/oss-review-toolkit/pull/1903#discussion_r341245108 for a use-case.

ilya-g:
@sschuberth I've skimmed through the discussion, but haven't got an idea how the described use case is different from a regular BFS. And I wondered originally about the _bottom-up_ BFS.
In DFS, the bottom-up and top-down walk orders differ only in a way how directory entries are visited compared to their nested file entries. In the referenced case the author filters the visited directory entries out with `.filter { it.isFile }`, so it doesn't matter in which order they were visited.

sschuberth:
Sorry, you're right @ilya-g. I didn't get that the emphasis was on *bottom-up*, not on BFS in general. The discussion I referred to only explains the use-case for BFS vs. DFS.

PR #2233, opened 2019-03-29
KT-30208 Convert to anonymous object: lambda generic type argument is lost

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30208


dimonchik0036:
Thanks!

PR #2234, opened 2019-03-31
KT-22807: allow asynchronous testing in Kotlin/JS with QUnit

gcx11:
https://youtrack.jetbrains.com/issue/KT-22807

### DIFF
--- a/libraries/kotlin.test/js/it/js/qunit-reporter.js
+++ b/libraries/kotlin.test/js/it/js/qunit-reporter.js
@@ -1,12 +1,6 @@
 var Tester = require('./test-result-checker');
-
-var full = require('./expected-outcomes');
-var allAsyncPass = {};
-for (var name in full) {
-    allAsyncPass[name] = name.startsWith('AsyncTest ') ? 'pass' : full[name];
-}
-
-var tester = new Tester(allAsyncPass, 'qunit');
+var expectedOutcomes = require('./expected-outcomes');
+var tester = new Tester(expectedOutcomes, 'qunit');
 
 QUnit.testDone(function (details) {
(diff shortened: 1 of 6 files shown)


anton-bannykh:
Don't see anything wrong with it at glance. Let me run the tests and get back to you.

anton-bannykh:
Apart from `Any` -> `Any?` (which is nitpicking) I see nothing wrong. The code seems to be working and reporting failing/passing tests correctly.
cc @ilya-g

ilya-g:
Why should it be `Any?` though? 
Both `Unit` and `Promise` have `Any` common supertype, right?

anton-bannykh:
@ilya-g A `() -> Unit` lambda is translated to a function, which returns `undefined` in the current backend. Which means that an `external` function gets `undefined` as the result of the invocation.

A `() -> Any` lambda always returns something, e.g. `Unit` instance. Which means that a test framework will be receiving a `Unit` instance if it gets a Kotlin lambda with type `() -> Any`.

This is probably fine, because:
- These lambdas are generated as JS code (for now) and kind of bend the type system a little bit here
- Most test frameworks ignore the return value they don't understand

While all that is true, changing the lambda type to `() -> Any?` allows us to create lambdas which explicitely return `undefined` if needed (e.g. a test framework which asserts nothing is returned for some reason).

But as I said, this is basically nitpicking, is based on the current backend gotchas, and basically doesn't matter much.

gcx11:
@anton-bannykh @ilya-g I have made changes to the return type, can this be merged?

anton-bannykh:
@gcx11 done. Thanks!

PR #2235, opened 2019-04-01
KT-27411 Inspection and Quickfix to replace System.exit() with exitProcess()

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27411

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaStaticMethodWithTopLevelFunctionInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaStaticMethodWithTopLevelFunctionInspection.kt
@@ -22,7 +22,7 @@ import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtPsiFactory
 import org.jetbrains.kotlin.psi.callExpressionVisitor
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
-import org.jetbrains.kotlin.psi.psiUtil.getTopmostParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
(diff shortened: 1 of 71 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaStaticMethodWithTopLevelFunctionInspection.kt
```diff
@@ -0,0 +1,121 @@
...
+            val dotQualified = descriptor.psiElement as? KtDotQualifiedExpression ?: return
+            val call = dotQualified.callExpression ?: return
+            val file = dotQualified.containingKtFile
+            val psiFactory = KtPsiFactory(call)
+            val valueArguments = call.valueArguments
+            if (replacement.toExtensionFunction) {
+                val receiverText = valueArguments.first().text
+                val argumentsText = valueArguments.drop(1).joinToString(separator = ", ") { it.text }
+                dotQualified.replaced(psiFactory.createExpression("$receiverText.${replacement.kotlinFunctionShortName}($argumentsText)"))
+                file.resolveImportReference(FqName(replacement.kotlinFunctionFqName)).firstOrNull()?.also {
```
dimonchik0036: `let` looks better for this case.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaStaticMethodWithTopLevelFunctionInspection.kt
```diff
@@ -0,0 +1,121 @@
...
+    ) {
+        private fun String.shortName() = takeLastWhile { it != '.' }
+
+        val javaMethodShortName = javaMethodFqName.shortName()
+
+        val kotlinFunctionShortName = kotlinFunctionFqName.shortName()
+    }
+
+    companion object {
+        private val replacements = listOf(
```
dimonchik0036: It is better to use `linkedSetOf`. You should override `equals` and `hashCode` for `Replacement` with `javaMethodFqName` as the key.

dimonchik0036: Use `linkedMapOf` with `javaMethodFqName` as the key and `Replacement` as the value, also you can remove `javaMethodFqName` from `Replacement`.

dimonchik0036: Or you can convert this list to a map.

dimonchik0036:
Thanks! faa6eacb256e36b0c7f345e47c68d4dcbae2e719

PR #2236, opened 2019-04-02
IrBuiltins: refactoring for equality checks

ting-yuan:
Previously,
* Equals performs IEEE 754 equality check for floating points and
  byte-to-byte checks for other types, including references.
* Ieee754Equals performs IEEE 754 for primitive types
* TotalOrderEquals performs total order equals to all types, including
  floating points.

Now it is simplified,
* Equals performs total order checks for all types.
* Ieee754Equals performs IEEE 754 for primitive types.
* (TotalOrderEquals is removed.)


dnpetrov:
Looks good to me. 

NB old back-end has some specialized implementations of primitive number equality that avoids boxing when it's not required. See org.jetbrains.kotlin.codegen/BoxedVsPrimitiveBranchedValues.kt and related tests.

max-kammerer:
@ting-yuan Thanks!

PR #2239, opened 2019-04-04
Add other types for static Java Number.toString(i) to i.toString() (KT-12721)

devbridie:
https://youtrack.jetbrains.com/issue/KT-12721
See #2226.

Adds support for Double, Float, Long, Byte, Short.

Note the difference in test files: where `Integer` will always resolve to `java.lang.Integer`, `Double` needs to be qualified because it will refer to [`Kotlin.Double`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-double/index.html), which does not have `toString(d: double)`.


idea/resources/META-INF/plugin-common.xml
```diff
@@ -3346,8 +3346,8 @@
                      language="kotlin"
     />
 
-    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceJavaIntegerToStringWithMemberInspection"
-                     displayName="Replace 'Integer.toString(int)' with 'int.toString()'"
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceJavaNumberToStringWithMemberInspection"
+                     displayName="Replace 'Number.toString(int)' with 'number.toString()'"
```
ilya-g: I find `Number.toString(int)` here slightly confusing, because `Number` class doesn't have such method. Also `int` parameter would be confusing for other numeric types.

Perhaps `Replace static 'toString(number)' with 'number.toString()'` would read better?

devbridie:
I think this branch is now invalidated by `ReplaceJavaStaticMethodWithTopLevelFunctionInspection`, which now contains this PR. https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceJavaStaticMethodWithKotlinAnalogInspection.kt#L98

PR #2240, opened 2019-04-04
[IR] introduced additional field 'inlineFunctionSymbol' to IrReturnableBlock

vvlevchenko:
This change let backend calculate and bind right function to the code block on generation debug information
e.g. for Kotlin/Native backend this let codegnerator to produce DILocation(..., inlinedAt:..) with correct value
for inlinedAt parameter, thus signifiacntly improve behaviour of debugger.

https://github.com/JetBrains/kotlin-native/pull/2849

### DIFF
--- a/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/IrBlock.kt
+++ b/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/IrBlock.kt
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.ir.expressions
 import org.jetbrains.kotlin.ir.declarations.IrReturnTarget
 import org.jetbrains.kotlin.ir.declarations.IrSymbolOwner
 import org.jetbrains.kotlin.ir.symbols.IrFileSymbol
+import org.jetbrains.kotlin.ir.symbols.IrFunctionSymbol
 import org.jetbrains.kotlin.ir.symbols.IrReturnableBlockSymbol
 
 interface IrContainerExpression : IrExpression, IrStatementContainer {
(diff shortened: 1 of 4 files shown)


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/impl/IrBlockImpl.kt
```diff
@@ -65,13 +65,14 @@ fun IrBlockImpl.inlineStatement(statement: IrStatement) {
...
 }
 
-class IrReturnableBlockImpl(
-    startOffset: Int,
-    endOffset: Int,
-    type: IrType,
-    override val symbol: IrReturnableBlockSymbol,
-    origin: IrStatementOrigin? = null,
-    override val sourceFileSymbol: IrFileSymbol? = null
+open class IrReturnableBlockImpl(
```
romanart: Why is it `open`?

vvlevchenko: Ooops.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/IrBlock.kt
```diff
@@ -38,6 +41,13 @@ interface IrComposite : IrContainerExpression {
...
-}
\ No newline at end of file
+    val inlineFunctionSymbol: IrFunctionSymbol?
+}
+
+val IrReturnableBlock.sourceFileSymbol: IrFileSymbol?
+    get() = inlineFunctionSymbol?.owner?.file?.symbol
+
+
+val IrReturnableBlock.sourceFileName: String
```
bashor: how about deprecating it?

vvlevchenko: ok

PR #2241, opened 2019-04-05
JVM_IR: remove most uses of StackValue from ExpressionCodegen

pyos:
Part 2 (cont. of #2229)

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -102,7 +102,7 @@ class VariableInfo(val declaration: IrVariable, val index: Int, val type: Type, 
 
 class ExpressionCodegen(
     val irFunction: IrFunction,
-    val frame: IrFrameMap,
+    override val frameMap: IrFrameMap,
     val mv: InstructionAdapter,
     val classCodegen: ClassCodegen,
     val isInlineLambda: Boolean = false
@@ -117,9 +117,6 @@ class ExpressionCodegen(
     private val state = classCodegen.state
 
     private val fileEntry = classCodegen.context.psiSourceManager.getFileEntry(irFunction.fileParent)
-
-    override val frameMap: IrFrameMap
-        get() = frame
 
     override val visitor: InstructionAdapter
         get() = mv
@@ -246,7 +243,7 @@ class ExpressionCodegen(
         }
 
         info.variables.reversed().forEach {
-            frame.leave(it.declaration.symbol)
+            frameMap.leave(it.declaration.symbol)
         }
     }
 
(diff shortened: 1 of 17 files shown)


pyos:
@max-kammerer I've made a couple changes to make this easier to merge with the inlining: removed a bytecode-changing commit, left `irFunction` and `frame` public (although the latter is also available as `frameMap`...) and did not remove `genFinallyBlock` so you have a place to put the marker method calls. Merging should now boil down to combining changes to visitReturn and using unwindBlockStack instead of generateFinallyBlocksIfNeeded. I've checked the tests and it seems like there are no new failing ones now (commit 7 of the previous review only affected one bytecode text test because it did not use genFinallyBlock for one of the copies; the inliner doesn't actually need that one, but I've left it in anyway).

max-kammerer:
@pyos please rebase on master there is conflict with #2257

pyos:
@max-kammerer done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -104,26 +100,33 @@ class BlockInfo private constructor(val parent: BlockInfo?) {
 
 class VariableInfo(val declaration: IrVariable, val index: Int, val type: Type, val startLabel: Label)
 
-@Suppress("IMPLICIT_CAST_TO_ANY")
 class ExpressionCodegen(
     val irFunction: IrFunction,
     val frame: IrFrameMap,
```
max-kammerer: val frame ->  override val frameMap,
and next frameMap declaration would be redundant

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -183,36 +176,23 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
                 return null
 
-            genSubject()
             genSwitch()
-            genThenExpressions()
-            val result = genElseExpression()
-
-            mv.mark(endLabel)
-            return result
+            return genBranchTargets()
```
max-kammerer: @ting-yuan Please review changes in SwitchGenerator.kt, as I understand there should be several more specific generators and method merging could break your plans/design
```
            genSubject()
            genSwitch()
            genThenExpressions()
```

max-kammerer:
@pyos I would prefer to keep 'gaps' logic in one place, so please rollback "Partially inline genFinallyBlock pyos"

pyos:
@max-kammerer done. I guess there's not much point in that if a redundant copy of `finally` cannot be removed because of the inliner, yeah.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
ting-yuan: Although I prefer smaller functions as they are easier to read, this looks fine to me.

max-kammerer:
@pyos Please rebase last time and I will merge pr: github warns about conflict

max-kammerer:
@pyos Sorry, I missed that you updated last commit. I would prefer to distinguish NewArray and ArrayConstructors: maybe with more proper naming. At least last commit could be postponed to another PR

pyos:
@max-kammerer ok, I'll submit that separately.

>Please rebase last time and I will merge pr: github warns about conflict

Weird, it said there were no conflicts, and git pull --rebase completed automatically.

max-kammerer:
@pyos Thank you! Nice work!

PR #2242, opened 2019-04-05
Fix IR declaration parents.

sfs:
This PR fixes various places where the lateinit parent field of an IrDeclaration was either unset or not updated properly.

- Reparent declarations in FileClassLowering, InterfaceLowering,
  StaticDefaultFunctionsLowering, EnumClassLowering, and
  RenameFieldsLowering
- Set correspondingPropertySymbol in MoveCompanionObjectFieldsLowering
- Reparent field initializers in MoveCompanionObjectFieldsLowering and
  EnumClassLowering
- Ensure that parents are unique in PropertiesLowering
- Set declaration parent in SharedVariableManager
- Set field declaration parents in CallableReferenceLowering
- Set declaration parents in FunctionNVarargInvokeLowering
- Set declaration parents for external declaration fields

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
@@ -106,7 +106,7 @@ class SharedVariablesLowering(val context: BackendContext) : FunctionLoweringPas
                     if (declaration !in sharedVariables) return declaration
 
                     val newDeclaration = context.sharedVariablesManager.declareSharedVariable(declaration)
-                    newDeclaration.parent = irFunction
+                    newDeclaration.parent = declaration.parent
                     transformedSymbols[declaration.symbol] = newDeclaration.symbol
 
                     return context.sharedVariablesManager.defineSharedValue(declaration, newDeclaration)
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
@@ -153,7 +153,7 @@ private class EnumClassLowering(val context: JvmBackendContext) : ClassLoweringP
                     }
                 })
 
-                body = enumConstructor.body // will be transformed later
+                body = enumConstructor.body?.patchDeclarationParents(this)
 
                 loweredEnumConstructors[enumConstructor.symbol] = this
                 metadata = enumConstructor.metadata
@@ -219,7 +219,9 @@ private class EnumClassLowering(val context: JvmBackendContext) : ClassLoweringP
             context.declarationFactory.getFieldForEnumEntry(
                 enumEntry, (enumEntry.correspondingClass ?: enumEntry.parentAsClass).defaultType
             ).also {
-                it.initializer = IrExpressionBodyImpl(enumEntry.initializerExpression!!)
+                it.initializer = IrExpressionBodyImpl(
+                    enumEntry.initializerExpression!!.patchDeclarationParents(it)
+                )
                 it.annotations.addAll(enumEntry.annotations)
                 enumEntryFields.add(it)
                 enumEntriesByField[it] = enumEntry
--- a/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
+++ b/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
@@ -54,10 +54,15 @@ class PatchDeclarationParentsVisitor() : IrElementVisitorVoid {
     }
 
     override fun visitProperty(declaration: IrProperty) {
-        declaration.getter?.let { it.correspondingProperty = declaration }
-        declaration.setter?.let { it.correspondingProperty = declaration }
-        declaration.backingField?.let { it.correspondingProperty = declaration }
-
+        declaration.getter?.let {
+            it.correspondingPropertySymbol = declaration.symbol
+        }
+        declaration.setter?.let {
+            it.correspondingPropertySymbol = declaration.symbol
+        }
+        declaration.backingField?.let {
+            it.correspondingPropertySymbol = declaration.symbol
+        }
         super.visitProperty(declaration)
     }
 
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/PropertiesLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/PropertiesLowering.kt
@@ -8,6 +8,7 @@ package org.jetbrains.kotlin.backend.common.lower
 import org.jetbrains.kotlin.backend.common.BackendContext
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.descriptors.WrappedSimpleFunctionDescriptor
+import org.jetbrains.kotlin.backend.common.ir.copyTo
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.Visibilities
@@ -80,7 +81,7 @@ class PropertiesLowering(
         ).apply {
             descriptor.bind(this)
 
-            extensionReceiverParameter = declaration.getter?.extensionReceiverParameter
+            extensionReceiverParameter = declaration.getter?.extensionReceiverParameter?.copyTo(this)
 
             body = IrBlockBodyImpl(-1, -1)
 
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/SharedVariablesManager.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/SharedVariablesManager.kt
@@ -193,6 +193,7 @@ class JvmSharedVariablesManager(
         ).apply {
             (descriptor as WrappedVariableDescriptor).bind(this)
             initializer = refConstructorCall
+            parent = originalDeclaration.parent
         }
     }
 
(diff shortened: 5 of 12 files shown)


max-kammerer:
Should 'makePatchParentsPhase' be removed from JvmLower?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
```diff
@@ -54,9 +54,15 @@ class PatchDeclarationParentsVisitor() : IrElementVisitorVoid {
     }
 
     override fun visitProperty(declaration: IrProperty) {
-        declaration.getter?.let { it.correspondingProperty = declaration }
-        declaration.setter?.let { it.correspondingProperty = declaration }
-        declaration.backingField?.let { it.correspondingProperty = declaration }
+        declaration.getter?.let {
```
max-kammerer: Please keep correspondingProperty  initialization

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
```diff
@@ -106,7 +106,6 @@ class SharedVariablesLowering(val context: BackendContext) : FunctionLoweringPas
                     if (declaration !in sharedVariables) return declaration
 
                     val newDeclaration = context.sharedVariablesManager.declareSharedVariable(declaration)
-                    newDeclaration.parent = irFunction
```
max-kammerer: This could affect JS and Native

sfs: I am not sure about native, but the JS sharedVariablesManager sets this field correctly. The problem with this line is that it doesn't set the correct parent.

I'll check what kotlin-native does.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
sfs: I changed it because it's currently deprecated, but I can remove it from this PR.

max-kammerer: No, please support both correspondingPropertySymbol and correspondingProperty, otherwise we can break smth in Native

sfs: Alright, I got rid of this part.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
sfs: Nice catch! This does indeed have to be fixed in kotlin-native. I've created a separate PR for this: https://github.com/JetBrains/kotlin-native/pull/2855

sfs:
@max-kammerer I would leave makePatchParentsPhase (or at least the final one) for now. These are all the wrong declaration parents that were in the code as of 4 days ago, but it's easy enough to introduce more bugs of this kind.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
max-kammerer: @sfs 
Oh, you made right change, but I would prefer to also keep old assignment:
```
declaration.getter?.let { 
    it.correspondingProperty = declaration 
   it.correspondingProperty = declaration.symbol 
}
...
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
```diff
@@ -153,7 +153,7 @@ private class EnumClassLowering(val context: JvmBackendContext) : ClassLoweringP
                     }
                 })
 
-                body = enumConstructor.body // will be transformed later
+                body = enumConstructor.body?.patchDeclarationParents(this)
```
romanart: Is this really needed? If so I think there is another issue with lost parent

sfs: I thought the comment referred to the patch parents phase which runs a bit later in the pipeline. There is nothing else in EnumClassLowering which would fix the declarations here.

Which other issue did you have in mind?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
```diff
@@ -120,6 +120,7 @@ class DeclarationStubGenerator(
                     constantValueGenerator.generateConstantValueAsExpression(UNDEFINED_OFFSET, UNDEFINED_OFFSET, it)
                 )
             }
+            parent = generateParentStub(descriptor)
```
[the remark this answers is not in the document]
sfs: Right, that was my intent. Previously deserialized IrFields ended up with unset parents. Is there a reason why this would be the desired behavior?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
sfs: @max-kammerer Done

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
sfs: @max-kammerer I changed it to keep the assignment here. At worst it's redundant but at least it won't break anything.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
romanart: I mean what the exact reason that there is still uninitialized `parent` by this time. The best solution might be fixing parent there.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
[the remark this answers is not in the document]
romanart: `parent` should be generated only for synthetic declarations because everything comes from deserializer is properly initialized. If you have an example when it is not true could you please provide it?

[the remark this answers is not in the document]
max-kammerer: Seems this piece of original code is totally wrong, there should be IrLazyFieldImpl with lazy parent and initializer calculation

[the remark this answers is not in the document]
max-kammerer: Similar to another 'declares', e.g.:
```
return symbolTable.declareClass(UNDEFINED_OFFSET, UNDEFINED_OFFSET, origin, descriptor) {
            deserializer?.findDeserializedDeclaration(referenceClass) as? IrClass
                ?: IrLazyClass(UNDEFINED_OFFSET, UNDEFINED_OFFSET, origin, it, this, typeTranslator)
        }
```

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
max-kammerer: @sfs Sorry, but I made a big mistake: you original solution was totally right, I've rechecked implementation of corrrespondingProperty and found that it's delegates to correspondingPropertySymbol. So could you revert this change to your original solution

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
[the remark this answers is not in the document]
max-kammerer: @sfs This part is fixed in my branch: 
https://github.com/JetBrains/kotlin/commits/rr/mb/lazyfield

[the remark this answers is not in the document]
max-kammerer: And pushed to master

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
sfs: What is happening is that the original body of the constructor is copied to a new protected constructor with a slightly different signature. Later on, `EnumClassBodyTransformer` runs, which updates the parameter references and constructor calls. In principle, this pass could also update the parent fields. However, to do this we'd have to duplicate the implementation from `PatchDeclarationParentsVisitor`, since the visitor in question runs over the whole enum class instead of just over the constructor.

It is much more expedient to fix the declaration parents when we copy the constructor body.

I think the comment in the original source code refers to the `makePatchParentsPhase` which runs sometime after `EnumClassLowering`.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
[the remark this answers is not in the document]
sfs: Great, thank you! I got rid of this part in the PR.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/PatchDeclarationParents.kt
sfs: @max-kammerer Done!

sfs:
Updated the PR to the current master, removed the IR declaration stub changes.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
romanart: Thank you for explanation

max-kammerer:
@sfs Thank you!

PR #2243, opened 2019-04-05
Remove broken KotlinType.toIrType function

sfs:
The KotlinType.toIrType function sometimes produces unbound symbols and was used in IrIntrinsicFunction and SharedVariablesManager.

This change removes the broken function and changes SharedVariableManager to use IrType consistently and IrIntrinsicFunction to use KotlinType consistently.

Note that I haven't checked whether this function is used by kotline-native (build problems), so you may want to hold off on merging the changes to `compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt`.


romanart:
I am not sure that it is save to delete those function since they are likely be used in Kotlin/Native. 
cc: @SvyatoslavScherbina

udalov:
> Note that I haven't checked whether this function is used by kotlin-native (build problems)

It's usually pretty easy to get an idea whether something is used in kotlin-native by a simple text search, or even search on GitHub. In this case, `git grep toIrType` returns nothing, so I assume it's safe to remove it

max-kammerer:
@sfs Thanks!

PR #2244, opened 2019-04-05
Remove Concat intrinsic in favor of IrStringConcatenation.

sfs:
This PR makes FlattenStringConcatenation apply to nullable strings, which means that all occurrences of String.plus will be represented by IrStringConcatenation. This in turn reduces code duplication in the JVM_IR backend.


sfs:
Updated the PR to remove the StringPlus intrinsic, which was also redundant.

sfs:
Rebased to current master.

@max-kammerer Is there anything preventing this PR from being merged?

max-kammerer:
@sfs Sorry for delay

compiler/testData/codegen/bytecodeText/stringOperations/nullableStringPlus.kt
```diff
@@ -1,3 +1,6 @@
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: @sfs Please add same bytecode test for IR backend

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
```diff
@@ -73,11 +73,11 @@ class FlattenStringConcatenationLowering(val context: CommonBackendContext) : Fi
             return when (expression) {
                 is IrStringConcatenation -> true
                 is IrCall -> {
-                    val dispatchReceiver = expression.dispatchReceiver
-                    dispatchReceiver != null &&
-                            dispatchReceiver.type.isString() &&
+                    val receiver = expression.dispatchReceiver ?: expression.extensionReceiver
```
max-kammerer: @sfs Please also check package for irCall, this code would not work correctly for next cases (please also add test for it):
```
operator fun String?.plus(p: String) {
    this
}

fun test(a: String?, b: String): String {
return a + b
}
```

udalov:
Please ensure there's a test that checks that bytecode is optimized in simple call case such as
```
fun f(s: String?, t: String): String {
    return s.plus(t)
}
```

max-kammerer:
@sfs Please rebase on master there could be some affects of #2274

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
sfs: Thanks, I completely missed this!

sfs:
@max-kammerer @udalov 
Rebased and added more tests.

I changed ExpressionCodegen to generate a call to Intrinsics.stringPlus whenever possibly instead of always using StringBuilder. This way it's compatible with the current tests and produces slightly smaller code.

compiler/testData/codegen/box/strings/simpleStringPlus.kt
```diff
@@ -0,0 +1,18 @@
...
+fun g(s: String, t: Any?): String {
+    return "$s$t"
+}
+
+fun h(s: String, t: Any?): String {
+    return s + t
+}
+
+fun box(): String {
+    if (f("O", "K") != "OK") return "Fail"
```
max-kammerer: Fail 1

compiler/testData/codegen/box/strings/simpleStringPlus.kt
```diff
@@ -0,0 +1,18 @@
...
+    if (g("O", "K") != "OK") return "Fail"
```
max-kammerer: 'Fail 2' and so on

compiler/testData/codegen/bytecodeText/stringOperations/multipleNestedConcat.kt
```diff
@@ -4,7 +4,7 @@
 // - In argument to 2nd call to foo() inside string literal
 fun test(s1: String, s2: String, s3: String): String {
     fun foo(s: String) = s
-    return "foo: " + foo(s1 + s2 + " ${foo("\${s3.length} = ${s3.length}")}")
+    return "foo: " + foo(s1 + s2 + " ${foo("\${s3.length} = ${s3.length} ")}") + " "
```
max-kammerer: @sfs Why test update is required?

max-kammerer: @sfs Oh..seems 'Intrinsics.stringPlus'. It's intresting that old backend uses 'Intrinsics.stringPlus' only if receiver is nullable, with receiver is non-nullable than StringBuilder is used directly. It's not clear which way is better. Is JVM differ these two cases on any string optimizations? //cc @udalov

dnpetrov: HotSpot optimizes string concatenation (`"a" + "b" + "c"`, which is, effectively, `new StringBuilder().append("a").append("b").append("c")`).

max-kammerer: @dnpetrov Yes pr does same, 
the question is 
```
new StringBuilder
dup
init
aload var1
append
aload var2 or LDC constant
append 
toString
```
would be same as efficient on Hotspot  and other JVMs as
```
aload var1
aload var2 or LDC constant
invoke Intrinsics.stringPlus

fun Intrinsics.stringPlus(var1, var2)
new StringBuilder
dup
init
aload var1
append
aload var2
append 
toString
```

sfs: Having a single stringPlus call is slightly less code than creating a new StringBuilder.

If there are no plans to remove the stringPlus intrinsic then we might as well use it. I don't think there would be any performance difference in practice, since Intrinsic.stringPlus is a small leaf function which should be inlined (when hot) and uses StringBuilder.append internally.

compiler/testData/codegen/box/strings/simpleStringPlus.kt
sfs: Done

compiler/testData/codegen/box/strings/simpleStringPlus.kt
sfs: Done

compiler/testData/codegen/bytecodeText/stringOperations/multipleNestedConcat.kt
max-kammerer: @sfs Yes it should be inlined, but after inlining it could be a bit different bytecode, so if some pattern based string optimization is used further...

sfs: @max-kammerer I can't give a definite answer to this, but as far as I can tell the StringBuilder related optimizations (StringOpts) run [as part of inlining](https://github.com/sourcemirror/jdk-8-hotspot/blob/af290d0fb302ee6a355fc55bc11602fddbac08bb/src/share/vm/opto/compile.cpp#L2013).

max-kammerer:
@sfs Thank you!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
```diff
@@ -73,11 +75,13 @@ class FlattenStringConcatenationLowering(val context: CommonBackendContext) : Fi
...
-                    val dispatchReceiver = expression.dispatchReceiver
-                    dispatchReceiver != null &&
-                            dispatchReceiver.type.isString() &&
-                            expression.symbol.owner.name == PLUS_NAME &&
-                            expression.type.isString() &&
+                    val function = expression.symbol.owner
+                    val receiver = expression.dispatchReceiver ?: expression.extensionReceiver
+                    receiver != null &&
+                            receiver.type.isStringClassType() &&
+                            function.getPackageFragment()?.fqName == KotlinBuiltIns.BUILT_INS_PACKAGE_FQ_NAME &&
```
udalov: @sfs Please use `parent` directly and check that it's a package fragment instead of using `getPackageFragment` which walks up until it finds a package fragment (which is not precise enough and _could_ result in incorrect behavior here for a member function, but it seems impossible due to other conditions).

sfs: @udalov That was actually on purpose, since there are two functions we check for, `kotlin.String.plus` for non-nullable strings and `kotlin.plus` for nullable strings. If you prefer I could instead check for these two cases explicitly.

udalov: Thanks for the clarification. Yes, I'd prefer that since it's more explicit and avoids confusion with other functions named `plus` in the standard library (e.g. if `receiver`'s initializer is refactored to `expression.extensionReceiver ?: expression.dispatchReceiver`, we could incorrectly identify some member extension on `String` named `plus` in an irrelevant class here)

PR #2245, opened 2019-04-05
Handle reversals of Progressions in ForLoopsLowering.

punzki:
This includes calls to `reversed()` on Progressions, and `reversed()` and `reversedArray()` on Arrays. Note that the non-IR backend does _not_ optimize for-loop over reversed arrays, hence the new bytecode tests for reversed arrays are disabled.

Extracting `HeaderInfo` from iterables is now done in a unified manner, so that we can handle chained calls better, i.e., `ReversedHandler` can reverse `HeaderInfo` from both Progressions and Arrays.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
@@ -13,15 +13,20 @@ import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.descriptors.IrBuiltIns
 import org.jetbrains.kotlin.ir.expressions.IrCall
 import org.jetbrains.kotlin.ir.expressions.IrConst
+import org.jetbrains.kotlin.ir.expressions.IrConstKind
 import org.jetbrains.kotlin.ir.expressions.IrExpression
+import org.jetbrains.kotlin.ir.expressions.impl.IrCallImpl
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.symbols.IrSymbol
 import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.types.getClass
 import org.jetbrains.kotlin.ir.types.isSubtypeOfClass
+import org.jetbrains.kotlin.ir.util.functions
 import org.jetbrains.kotlin.ir.visitors.IrElementVisitor
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
 
 // TODO: Handle withIndex()
-// TODO: Handle reversed()
 // TODO: Handle Strings/CharSequences
 // TODO: Handle UIntProgression, ULongProgression
 
@@ -62,7 +67,9 @@ internal sealed class HeaderInfo(
     val first: IrExpression,
     val last: IrExpression,
     val step: IrExpression,
-    val isLastInclusive: Boolean
+    val isFirstInclusive: Boolean,
+    val isLastInclusive: Boolean,
+    val isReversed: Boolean
 ) {
     val direction: ProgressionDirection by lazy {
         // If step is a constant (either Int or Long), then we can determine the direction.
@@ -75,6 +82,13 @@ internal sealed class HeaderInfo(
             else -> ProgressionDirection.UNKNOWN
         }
     }
+
+    /**
+     * Returns a copy of this [HeaderInfo] with the values reversed.
+     * I.e., first and last (and their inclusiveness) are swapped, step is negated.
+     * Returns null if the iterable cannot be iterated in reverse.
+     */
+    abstract fun asReversed(): HeaderInfo?
 }
 
 /** Information about a for-loop over a progression. */
@@ -83,14 +97,15 @@ internal class ProgressionHeaderInfo(
     first: IrExpression,
     last: IrExpression,
     step: IrExpression,
+    isFirstInclusive: Boolean = true,
     isLastInclusive: Boolean = true,
-    canOverflow: Boolean? = null,
+    isReversed: Boolean = false,
     val additionalVariables: List<IrVariable> = listOf()
-) : HeaderInfo(progressionType, first, last, step, isLastInclusive) {
+) : HeaderInfo(progressionType, first, last, step, isFirstInclusive, isLastInclusive, isReversed) {
 
-    private val _canOverflow: Boolean? = canOverflow
     val canOverflow: Boolean by lazy {
-        if (_canOverflow != null) return@lazy _canOverflow
+        // Last-exclusive progressions can never overflow.
+        if (!isLastInclusive) return@lazy false
 
         // Induction variable can overflow if it is not a const, or is MAX/MIN_VALUE (depending on direction).
         val lastValue = (last as? IrConst<*>)?.value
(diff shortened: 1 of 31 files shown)


compiler/testData/codegen/bytecodeText/forLoop/forInReversed/forInReversedRangeLiteral.kt
```diff
@@ -24,12 +23,12 @@ fun box(): String {
...
 // 0 getStart
 // 0 getEnd
 // 0 getFirst
 // 0 getLast
 // 0 getStep
-// 0 IFEQ
-// 0 IF_ICMPEQ
-// 2 IF_ICMPLT
-// 1 LCMP
+// 2 IF_ICMP
```
max-kammerer: // 2 IF_ICMP[LG]T
// 1 IF[LG]T
// 3 IF

compiler/testData/codegen/box/ranges/forInReversed/forInReversedArray.kt
```diff
@@ -0,0 +1,14 @@
+// KJS_WITH_FULL_RUNTIME
+// WITH_RUNTIME
+import kotlin.test.*
+
+fun box(): String {
+    val arr = intArrayOf(1, 2, 3, 4)
+    var sum = 0
+    for (i in arr.reversed()) {
```
max-kammerer: @punzki  reversed/reversedArray() is totatlly unsafe optimization:
```
fun box(): String {
    val arr = intArrayOf(1, 2, 3, 4)
    var sum = 0
    var index = 0
    //for (i in arr.reversedArray()) {
    for (i in arr.reversed()) {
        arr[index++] = 0
        sum = sum * 10 + i
    }
    assertEquals(4321, sum)

    return "OK"
}
```

max-kammerer: Please add such test cases for reversed and reversedArray and double reversed/reversedArray to don't miss such cases in future

punzki: @max-kammerer Thank you for pointing this out, I hadn't considered this!
1. Removed optimization for `Array.reversed()/reversedArray()`
2. Added new box tests for what you described
3. Removed bytecode tests for optimized `Array.reversed()/reversedArray()`

compiler/testData/codegen/bytecodeText/forLoop/forInReversed/forInReversedRangeLiteral.kt
punzki: Good suggestion, didn't consider that!

max-kammerer:
@punzki 
1. Test without array modification are redundant (they are tested via stdlib tests) please remove them
2. Please move array box tests under codegen/box/arrays/forInReversed 
3. please squash commits into one

compiler/testData/codegen/box/ranges/forInReversed/forInReversedArrayUsingReversedArrayOriginalUpdatedInLoopBody.kt
```diff
@@ -0,0 +1,18 @@
+// KJS_WITH_FULL_RUNTIME
+// WITH_RUNTIME
+import kotlin.test.*
+
+fun box(): String {
+    val arr = intArrayOf(1, 2, 3, 4)
+    var sum = 0
+    var index = 0
+    for (i in arr.reversed()) {
```
max-kammerer: reversed -> reversedArray?

punzki: Oops, fixed.

punzki:
> @punzki
> 
> 1. Tests without array modification are redundant (they are tested via stdlib tests) please remove them
> 2. Please move array box tests under codegen/box/arrays/forInReversed
> 3. please squash commits into one

@max-kammerer Done. Also, @sfs pointed out an issue with the declaration parents, so I fixed that as well.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
```diff
@@ -118,6 +134,18 @@ internal class ProgressionHeaderInfo(
         }
         constLimitAsLong == lastValueAsLong
     }
+
+    override fun asReversed() = ProgressionHeaderInfo(
```
max-kammerer: Seems next case (copy of forInReversedUntil.kt with double reverse) woudn't optimized as simple until:
```
fun box(): String {
    var sum = 0
    for (i in (1 until 5).reversed().reversed()) {
        sum = sum * 10 + i
    }

    var sumL = 0L
    for (i in (1L until 5L).reversed().reversed()) {
        sumL = sumL * 10 + i
    }

    var sumC = 0
    for (i in ('1' until '5').reversed().reversed()) {
        sumC = sumC * 10 + i.toInt() - '0'.toInt()
    }

    return "OK"
}

```

max-kammerer: Same with arr.indices.reversed().reversed() wouldn't equal to arr.indices optimization

max-kammerer: Please add bytecode test with double reverse for array.indices and constant until ranges, e.g.:
```
fun box(): String {
    var sum = 0
    for (i in (1 until 5).reversed().reversed()) {
        sum = sum * 10 + i
    }

    var sumL = 0L
    for (i in (1L until 5L).reversed().reversed()) {
        sumL = sumL * 10 + i
    }

    var sumC = 0
    for (i in ('1' until '5').reversed().reversed()) {
        sumC = sumC * 10 + i.toInt() - '0'.toInt()
    }

    return "OK"
}
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -121,58 +128,67 @@ internal class ProgressionLoopHeader(
...
+                    body = newBody
+                }
+                val notEmptyCheck = irIfThen(buildLoopCondition(builder), newLoop)
+                LoopReplacement(newLoop, notEmptyCheck)
+            } else {
+                // If the induction variable can NOT overflow, use a simple while loop. Loop is lowered into something like:
+                //
+                //   while (inductionVar <= last) {
+                //       val loopVar = inductionVar
+                //       inductionVar += step
```
max-kammerer: Why not 
```
                //   while (loopVar <= last) {
                //       loopVar = inductionVar
                //       // Loop body
                //       loopVar += step                
                //   }
```

max-kammerer:
There are a bunch of affected bytecode tests in pr, I think  it would be worth to add also 'if' checks for them, especially for double 'reversed' to be sure that they would be equal to unreversed variant.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
punzki: That would mean moving `loopVar` declaration outside of the while loop, and removing `inductionVar` since it's not needed anymore. It's possible, but it would only work just for this one case.

punzki:
> There are a bunch of affected bytecode tests in pr, I think it would be worth to add also 'if' checks for them, especially for double 'reversed' to be sure that they would be equal to unreversed variant. But it could be postponed for next prs.

I went ahead and added `IF` checks where they make sense.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
punzki: Done!

max-kammerer:
@punzki Thank you!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
max-kammerer: As I see that could affect reversing

PR #2248, opened 2019-04-08
KT-30613 "Convert to anonymous function" should not insert named argument when interoping with Java functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30613


mglukhikh:
Merged 👍

PR #2249, opened 2019-04-08
KT-30725 Inspection which replaces `.sorted().first()` with `.min()`

t-kameyama:
Replace `.sorted().first()` with `.min()!!`.
https://youtrack.jetbrains.com/issue/KT-30725#focus=streamItem-27-3391048.0-0


mglukhikh:
Merged 👍

PR #2251, opened 2019-04-08
Add IR equivalent of the WriteFlags test

pyos:


max-kammerer:
@pyos Thanks! Merged manually with small changes: IGNORE_EXPECTED_FAILURES mechanism is required for failed test output elumunation on CI, but ideally it should be moved to base test class. For writeFlag tests it's not necessery

PR #2253, opened 2019-04-09
KT-30481 Do not report ImplicitNullableNothingType on a function/property that overrides a function/property of type 'Nothing?'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30481


idea/src/org/jetbrains/kotlin/idea/inspections/ImplicitNullableNothingTypeInspection.kt
```diff
@@ -23,13 +23,14 @@ import org.jetbrains.kotlin.psi.*
...
-        additionalChecker = { declaration ->
-            SpecifyTypeExplicitlyIntention.getTypeForDeclaration(declaration).isNullableNothing() &&
-            (declaration.getModalityFromDescriptor() == KtTokens.OPEN_KEYWORD ||
-             declaration is KtProperty && declaration.isVar)
-        },
-        problemText = "Implicit `Nothing?` type"
+    intention = SpecifyTypeExplicitlyIntention::class,
+    additionalChecker = { declaration ->
+        SpecifyTypeExplicitlyIntention.getTypeForDeclaration(declaration).isNullableNothing() &&
+                !declaration.hasModifier(KtTokens.OVERRIDE_KEYWORD) &&
```
mglukhikh: It's not enough. You have to check that base declaration also has `Nothing?` type. Otherwise you will break the common case like this:
```
interface Base {
    fun getMessage(): String?
}
abstract class Abstract : Base {
    override fun getMessage() = null
}
```

Here nobody can override `getMessage()` from the abstract class, because its type is now `Nothing?` and any other implementation will also have to return null.

PR #2254, opened 2019-04-09
KT-30761 Replace assert boolean with assert equality produces uncompilable code when compared arguments type are different

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30761


mglukhikh:
Merged 👍

PR #2255, opened 2019-04-09
KT-20357 Add sample code for hashMapOf

keijumt:


ilya-g:
Thanks for the sample, I've merged it.

PR #2256, opened 2019-04-09
Flatten phase compositions

pyos:
This makes stack traces and flame graphs significantly more readable, which is important when trying to improve performance.


gogabr:
Thank you!

PR #2257, opened 2019-04-09
Create a single instance of IrIntrinsicMethods per context

pyos:
This improves JVM_IR compilation times a bit.


max-kammerer:
@pyos Thanks!

PR #2259, opened 2019-04-09
JVM_IR: Lower IrGetEnumValue to `this` whenever possible.

ting-yuan:
The reference can be lowered to `this` if it is captured in the lexical
scope of the corresponding enum entry, and not used by the enum entry's
super constructor. Otherwise, it is lowered to
`GETFIELD SomeEnum.SomeEntry`.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingletonReferencesLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingletonReferencesLowering.kt
@@ -5,14 +5,17 @@
 
 package org.jetbrains.kotlin.backend.jvm.lower
 
-import org.jetbrains.kotlin.backend.common.BodyLoweringPass
+import org.jetbrains.kotlin.backend.common.ClassLoweringPass
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.backend.common.pop
+import org.jetbrains.kotlin.backend.common.push
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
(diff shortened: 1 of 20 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingletonReferencesLowering.kt
```diff
@@ -22,18 +23,77 @@ internal val singletonReferencesPhase = makeIrFilePhase(
...
+    private lateinit var containingClass: IrClass
+    private val constructingEnums = arrayListOf()
+
+    override fun lower(irClass: IrClass) {
+        containingClass = irClass
+        irClass.transformChildrenVoid(this)
+    }
+
+    override fun visitEnumConstructorCall(expression: IrEnumConstructorCall): IrExpression {
+        constructingEnums.add(expression.symbol.owner.parent)
```
max-kammerer:
```
import org.jetbrains.kotlin.backend.common.pop
import org.jetbrains.kotlin.backend.common.push

constructingEnums.push
constructingEnums.pop 
```

max-kammerer:
@ting-yuan Is it necessery to move singletonReferencesPhase before jvmLocalDeclarationsPhase? Or it can be moved just after?

ting-yuan:
@max-kammerer It is necessary to move `singletonReferencesPhase` before `jvmLocalDeclarationsPhase`. The former relies on the latter to pass `this` properly. Is there any concern / potential issue?

max-kammerer:
@ting-yuan It can be problem in future with common prefix extraction, but I think we should support all tests before. Then we can more precisely define dependencies between lowers and perform required changes having global view.

max-kammerer:
@ting-yuan Thanks!

PR #2262, opened 2019-04-10
JVM_IR: Output EnclosingMethod attribute.

madsager:
This works in many cases, however, it is incomplete since there
are cases where classes are extracted to top-level and therefore
reparented. Therefore, we lose the information about the function
class are nested inside.


madsager:
@max-kammerer Simplest change that I could come up with to output these.

The reparenting issue seems similar to what @udalov had to deal with in connection with the output of metadata where he ended up holding on to the original descriptor if the descriptor changed. However, I wanted to not use descriptors as you are working on getting rid of them. If I use descriptors for this one more test pass (one where the class is reparented and lifted out from a method to top level).

max-kammerer:
@madsager Thanks!

PR #2263, opened 2019-04-11
KT-30769 Override quickfix creates "sealed fun"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30769


dimonchik0036:
Thanks!

PR #2264, opened 2019-04-11
KT-30883 Inspection + quickfix to replace non-null assertion with mapNotNull

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30883

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -3384,6 +3384,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.collections.ConvertMapWithNotNullAssertionToMapNotNullInspection"
+                     displayName="'map' with not-null assertion should be simplified to 'mapNotNull'"
+                     groupPath="Kotlin"
+                     groupName="Style issues"
+                     enabledByDefault="true"
+                     level="WEAK WARNING"
+                     language="kotlin"
+    />
+
     <referenceImporter implementation="org.jetbrains.kotlin.idea.quickfix.KotlinReferenceImporter"/>
 
     <fileType.fileViewProviderFactory filetype="KJSM" implementationClass="com.intellij.psi.ClassFileViewProviderFactory"/>
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertMapWithNotNullAssertionToMapNotNullInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertMapWithNotNullAssertionToMapNotNullInspection.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license 
+ * that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.inspections.collections
+
+import com.intellij.codeInsight.intention.LowPriorityAction
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.codeInspection.ProblemHighlightType
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.openapi.project.Project
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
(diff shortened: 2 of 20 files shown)


idea/resources/META-INF/plugin-common.xml
```diff
@@ -3404,7 +3404,16 @@
                      level="WEAK WARNING"
                      language="kotlin"
     />
-    
+
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ReplaceNotNullAssertionWithMapNotNullInspection"
+                     displayName="Replace '!!' with 'mapNotNull'"
```
dimonchik0036: `'map' with not-null assertion should be simplified to 'mapNotNull'` looks better.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNotNullAssertionWithMapNotNullInspection.kt
```diff
@@ -0,0 +1,91 @@
...
+            if (returnedExpression != postfix) return
+        } else {
+            if (blockLastStatement != postfix) return
+        }
+
+        val call = blockParent.getStrictParentOfType<KtCallExpression>() ?: return
+        if (!call.isCalling(mapNotNullFqName)) return
+
+        holder.registerProblem(
+            operationReference,
```
dimonchik0036: Should highlight `map`.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNotNullAssertionWithMapNotNullInspection.kt
```diff
@@ -0,0 +1,91 @@
...
+            "Replace '!!' with 'mapNotNull'",
```
dimonchik0036: Same as above.

idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceNotNullAssertionWithMapNotNullInspection.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.jetbrains.kotlin.idea.quickfix.createFromUsage.callableBuilder.getReturnTypeReference
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.typeUtil.makeNullable
+
+class ReplaceNotNullAssertionWithMapNotNullInspection : AbstractKotlinInspection() {
```
dimonchik0036: `ConvertMapWithNotNullAssertionToMapNotNullInspection`?

idea/src/org/jetbrains/kotlin/idea/inspections/collections/ConvertMapWithNotNullAssertionToMapNotNullInspection.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.typeUtil.makeNullable
+
+class ConvertMapWithNotNullAssertionToMapNotNullInspection : AbstractKotlinInspection() {
+    companion object {
+        private val MAP_NOT_NULL = FqName("kotlin.collections.map")
+    }
+
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = qualifiedExpressionVisitor(fun(qualifiedExpression) {
```
dimonchik0036: `callExpressionVisitor` for the case without a receiver
``` Kotlin
fun baz(list: List<Int?>) {
    with(list) {
        map {
            it!!
        }
    }
}
```

goodwinnk:
@t-kameyama Thank you for the amazing work on this issue, but as Niko Knop noticed in the issue, this inspection doesn't take into account a major semantic behavior change. We are now trying to minimize amount of false positive in new inspections and considering this and changed semantic we've decided to reject the original request in the tracker.

PR #2265, opened 2019-04-11
JVM_IR: set ConstantValue on static final primitive fields

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -420,6 +420,15 @@ class ExpressionCodegen(
     }
 
     override fun visitFieldAccess(expression: IrFieldAccessExpression, data: BlockInfo): PromisedValue {
+        classCodegen.fieldConstantValue(expression.symbol.owner)?.let {
+            if (expression is IrSetField) {
```
max-kammerer: @pyos Old backend also checks const modifier: StackValue.inlineConstantIfNeeded

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -326,6 +328,17 @@ open class ClassCodegen protected constructor(
         }
     }
 
+    // JVMS 4.7.2, The ConstantValue Attribute: https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.2
+    // Constant-initialized primitive static final fields act like language constants: they are inlined and can be used
+    // in annotations and Java switch cases.
+    fun fieldConstantValue(field: IrField): IrConst<*>? {
+        // TODO return null if NoConstantValueAttributeForNonConstVals is set and the property is not explicitly const
+        val asmType = typeMapper.mapType(field.descriptor)
+        if (!field.isStatic || !field.isFinal || (!AsmUtil.isPrimitive(asmType) && asmType != AsmTypes.JAVA_STRING_TYPE))
```
dnpetrov: Since Kotlin 1.4 (language feature: NoConstantValueAttributeForNonConstVals), we generate ConstantValue attributes for `const val`s only. That's because such fields can be used in constant expressions (e.g., in annotations) in Java code, and if someone changes property definition, it can implicitly break binary compatibility.

pyos: Yeah, that's what the TODO is about. I'm trying to figure out how to correctly determine constness without using descriptors right now.

max-kammerer: irField.corresponginPropertySymbol.owner.isConst?

pyos: That's unfortunately not always correct for Java fields.

pyos: I've added a condition that almost does the right thing; the only exception is an inherited Java field, in which case the field has origin FAKE_OVERRIDE with empty overriddenSymbols, so I see no way to check that it actually came from Java.

compiler/testData/codegen/bytecodeText/companion/kt14258_1.kt
```diff
@@ -3,7 +3,7 @@
 
 class My {
     companion object {
-        private val my: String = "OK"
+        val my: String = "OK"
```
max-kammerer: @pyos Please keep old test

pyos: Then I'll have to mute it for JVM_IR, since a private const field's read is inlined. Is there any reason not to inline such fields that I've missed, by the way?

max-kammerer: Seems it's a bug in psi2ir or in some Lower, correspondingProperty/correspondingPropertySymbol should be always present //cc @dnpetrov

max-kammerer: @pyos This field could be leaked by @PublishedApi inline function, in another case it could be safely inlined

pyos: @max-kammerer I found a bug in MoveCompanionObjectFieldsLowering: it does not set correspondingPropertySymbol on the new static backing fields.

max-kammerer: @pyos Good catch!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -420,6 +420,15 @@ class ExpressionCodegen(
     }
 
     override fun visitFieldAccess(expression: IrFieldAccessExpression, data: BlockInfo): PromisedValue {
+        classCodegen.fieldConstantValue(expression.symbol.owner)?.let {
```
max-kammerer: @pyos Is it possible to move this part to ConstLowering? In such case additional optimization would be applied during lowers phase, e.g. foldConstantLoweringPhase

pyos: Yes, I'm actually looking at this right now.

pyos: @max-kammerer I've moved inlining of Java constants into ConstLowering and added an assertion to codegen to ensure that nothing has been missed, could you have a look, please?

max-kammerer:
@pyos Many Thanks!

PR #2267, opened 2019-04-12
Add the IR version of CompileJavaAgainstKotlin tests

pyos:
Only the WithoutJavac version for now, because the other one ignores
javac errors.


max-kammerer:
@pyos It's OK to support just one for now cause of different failing tests

max-kammerer:
@pyos Thanks!

PR #2268, opened 2019-04-12
KT-30900 Invert 'if' condition respects neither code formatting nor inline comments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30900


t-kameyama:
Fixed

dimonchik0036:
Thanks!

PR #2269, opened 2019-04-12
KT-30910 "Use property access syntax" is not suitable as text for inspection problem text

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30910


dimonchik0036:
Thanks!

PR #2270, opened 2019-04-14
KT-30640 Add inspection for check/require/checkNotNull/requireNotNull

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30640
https://youtrack.jetbrains.com/issue/KT-22412

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceGuardClauseWithFunctionCallInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceGuardClauseWithFunctionCallInspection.kt
@@ -7,7 +7,8 @@ package org.jetbrains.kotlin.idea.inspections
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
-import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.core.ShortenReferences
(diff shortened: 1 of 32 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ReplaceGuardClauseWithFunctionCallInspection.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    override fun fixText(element: KtIfExpression) =
+        element.getKotlinFunction()?.let { "Replace with '${it.functionName}()' call" } ?: defaultFixText
+
+    override fun isApplicable(element: KtIfExpression): Boolean {
+        if (element.condition == null) return false
+        val call = element.getCallExpression() ?: return false
+        val calleeText = call.calleeExpression?.text ?: return false
+        val valueArguments = call.valueArguments
+        if (valueArguments.size > 1) return false
```
mglukhikh: What if an argument is not a string, but, for example, a cause exception? E,g,

```
try {
...
} catch (e: Exception) {
    if (e is RuntimeException) throw IllegalStateException(e)
}
```

mglukhikh:
I have to mention there is a serious problem with this inspection: yet it breaks smart casts, and we cannot enable it until contract-based smart casts are supported...

mglukhikh:
We will duscuss how to handle this situation better. Probably we can commit this with default OFF state, or guard the inspection with some language feature check.

mglukhikh:
After investigation I've found that contracts is ON by default in 1.3, so in 1.3 the inspection can be enabled, but availability of `LanguageFeature.UseReturnsEffect` should be checked.

mglukhikh:
Merged manually, thank you.

PR #2271, opened 2019-04-14
KT-31008 Fix fat dSYM Info.plist copying

nikolaykasyanov:
https://youtrack.jetbrains.com/issue/KT-31008

I tested it on my project and it does fix the issue.
The fix is consistent with [another use](https://github.com/JetBrains/kotlin/blob/977d3ef840604959c90d950f94d2fc6069eb556c/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/FatFrameworkTask.kt#L268) of `project.copy` in the same file.

Would it be possible to include it into a 1.3.30 bugfix release?


ilmat192:
Included in the bugfix branch: https://github.com/JetBrains/kotlin/commit/01b1c77c436f2e79f5bd31b221396fba768f63c2.

nikolaykasyanov:
@ilmat192 thanks!

PR #2273, opened 2019-04-15
KT-25006 Add inspection "'equals()' between objects of inconvertible primitive / enum / string types"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25006

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinEqualsBetweenInconvertibleTypesInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinEqualsBetweenInconvertibleTypesInspection.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license 
+ * that can be found in the license/LICENSE.txt file.
+ */
+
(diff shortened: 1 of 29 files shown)


mglukhikh:
Looks good, I will test it a bit and accept.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinEqualsBetweenInconvertibleTypesInspection.kt
```diff
@@ -0,0 +1,45 @@
...
+            holder.registerProblem(callee, "'equals()' between objects of inconvertible types")
+        }
+    })
+
+    companion object {
+        fun BindingContext.isInconvertibleTypes(expr1: KtExpression?, expr2: KtExpression?): Boolean {
+            if (expr1 == null || expr2 == null) return false
+            val type1 = getType(expr1)?.takeIf { it.isTargetType() } ?: return false
+            val type2 = getType(expr2)?.takeIf { it.isTargetType() } ?: return false
+            return type1 != type2
```
mglukhikh: After testing I found a bug here:
```
System.getProperty("some")?.equals("true") // highlighted, probably due to nullability reason
```

mglukhikh:
Merged manually

PR #2274, opened 2019-04-15
JVM_IR: fold constant string concatenations

pyos:


pyos:
@max-kammerer could you have a look, please?

max-kammerer:
@pyos Yes, sorry there was long holidays in German

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FoldConstantLowering.kt
```diff
@@ -176,6 +174,18 @@ class FoldConstantLowering(private val context: JvmBackendContext) : IrElementTr
                     else -> expression
                 }
             }
+
+            override fun visitStringConcatenation(expression: IrStringConcatenation): IrExpression {
+                expression.transformChildrenVoid(this)
+                return if (expression.arguments.all { it is IrConst<*> && it.value is String }) {
```
max-kammerer: @pyos Seems it's possible to support some additional cases when begining of IrStringConcatenation is constant concatenation:
expression.arguments.split {!(it is IrConst<*> && it.value is String)

max-kammerer: @pyos Please also take into account #2244 that also could unmute several tests

JakeWharton: Would also like to see `Char` handled.

And really so long as the stdlib version is equal to or less than the compiler version all constants should be able to be supported.

max-kammerer: @pyos Really why not support all types,e.g.:
val p = "123${1}"?

pyos: >Seems it's possible to support some additional cases when begining of IrStringConcatenation is constant concatenation

Maybe fold all constant subsequences then?

>Please also take into account #2244 that also could unmute several tests

I've checked, doesn't look like there is a test for a combination of these changes. Probably because there's no way to produce a constant of type `String?`, so a test would have to do something like `fun foo(x: String?) = x + "a" + "b"` and the non-IR backend does not optimize partially constant concatenations.

pyos: Yeah, good point, fixed that.

max-kammerer: > Maybe fold all constant subsequences then?
Yes

compiler/testData/codegen/bytecodeText/stringOperations/constConcat.kt
```diff
@@ -0,0 +1,5 @@
+val s = "1" + "2" + 3 + 4L + 5.0 + 6F + '7'
```
max-kammerer: @pyos Please add similar test with const vals instead of const literals

val c = "${strConst}...${intConst}....

max-kammerer:
@pyos Thank you!

PR #2275, opened 2019-04-16
KT-8958 ReplaceWith intention message could be more helpful in case of generic substitution

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-8958


idea/src/org/jetbrains/kotlin/idea/quickfix/replaceWith/DeprecatedSymbolUsageFixBase.kt
```diff
@@ -100,15 +103,41 @@ abstract class DeprecatedSymbolUsageFixBase(
...
+            val classLiteral = when (expressionFromPattern) {
+                is KtClassLiteralExpression -> expressionFromPattern
+                is KtDotQualifiedExpression -> expressionFromPattern.receiverExpression as? KtClassLiteralExpression
+                else -> null
+            }
+            if (classLiteral != null) {
+                val receiver = classLiteral.receiverExpression ?: return this
+                val typeParameterText = (descriptor as? CallableDescriptor)?.typeParameters?.firstOrNull()?.name?.asString() ?: return this
+                if (receiver.text != typeParameterText) return this
+                val typeReference = (element.parent as KtCallExpression).typeArguments.firstOrNull()?.typeReference ?: return this
```
dimonchik0036: I'd prefer a safe cast.

PR #2277, opened 2019-04-16
JVM_IR: Generate java 8 parameter names when -java-parameters is passed.

madsager:
The type mapper does not map enum parameters and outer this parameters
to the right parameter signature kinds so around half the tests
are still failing. Since a new type mapper is being worked
on I will not investigate that further right now.


max-kammerer:
@madsager Thanks!

PR #2278, opened 2019-04-16
JVM_IR: a couple of small fixes to codegen

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -446,10 +447,10 @@ class ExpressionCodegen(
     override fun visitSetField(expression: IrSetField, data: BlockInfo): PromisedValue {
         val expressionValue = expression.value
         // Do not add redundant field initializers that initialize to default values.
-        // "expression.origin == null" means that the field is initialized when it is declared,
-        // i.e., not in an initializer block or constructor body.
-        val skip = irFunction is IrConstructor && irFunction.isPrimary &&
-                expression.origin == null && expressionValue is IrConst<*> &&
+        val inConstructor = irFunction is IrConstructor && irFunction.isPrimary
```
max-kammerer: inConstructor -> inPrimaryConstructor

max-kammerer: Please also keep  comment
//"expression.origin == null" means that the field is initialized when it is declared...

pyos: I felt that the name of the variable made it sufficiently clear that this is an initializer rather than an assignment, but if not, then I'll leave the comment.

max-kammerer:
@pyos Thank  you!

PR #2279, opened 2019-04-16
Add a common JVM/JS lowering for Array(size, function)

pyos:
and remove the hack from JVM_IR codegen that replaces this call with hardcoded inline function bytecode.

Defining inline secondary constructors in stdlib would be even better, of course, but neither the JVM inliner nor the JS one appear to support that. For now, this at least makes JVM codegen a bit cleaner (and produces slightly smaller bytecode).


anton-bannykh:
@pyos I believe that both Kotlin/JS-IR and Kotlin/Native do support inline Array constructors. Having said that, utilizing a lowering common for all three backends might be preferrable.

pyos:
Kotlin/Native does (via an internal annotation), true, and that's how these constructors are currently implemented there. Not sure about JS_IR - there is no annotation and the `inline` keyword does not seem to work, although that might be a simple psi2ir bug because I think I noticed that the constructors have `isInline == false` regardless.

By the way, this PR contains some ad-hoc IR lambda inlining logic to work around the fact that JVM_IR codegen does not support IrReturnableBlock; I'm thinking of implementing that instead and then using Native/JS FunctionInlining.Inliner class.

anton-bannykh:
JS_IR [rewrites](https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/lower/inline/FunctionInlining.kt#L52) the inline constructor calls into internal stdlib function invocations with proper bodies. As far as I know that works, including support for non-local returns, etc.

Regarding the ad-hoc IR lambda inlining: that probably doesn't make much sense for JS_IR and Native, since is just duplicated existing logic. So I would agree that delegating to the FunctionInlining seems to a cleaner approach.

anton-bannykh:
LGTM

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -62,13 +60,15 @@ class TryInfo(val onExit: IrExpression) : ExpressionInfo() {
...
+    val returnSymbol: IrSymbol,
+    val returnTemporary: Int? = null
+) : ExpressionInfo()
 
-    fun create() = BlockInfo(this).apply {
-        this@apply.infos.addAll(this@BlockInfo.infos)
-    }
+class BlockInfo(val parent: BlockInfo? = null) {
+    val variables = mutableListOf<VariableInfo>()
+    val infos: Stack<ExpressionInfo> = parent?.infos ?: Stack()
```
max-kammerer: Shouldn't 'parent?.infos' be copied to another stack?

pyos: Doesn't matter, wirhBlock/handleBlock guarantee that scopes are exited correctly before destruction of BlockInfo.

max-kammerer:
@pyos Please add inline tests with *Arrays to boxInline/nonLocalReturns and boxInline/reified

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: @pyos Than it's best to hide this field: `private` (that will require to make some of inline function 'internal')

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/ArrayConstructorLowering.kt
```diff
@@ -0,0 +1,122 @@
...
+        //     return result as Array<T>
+        // }
+        // (and similar for primitive arrays)
+        val size = expression.getValueArgument(0)!!.transform(this, null)
+        val invokable = expression.getValueArgument(1)!!.transform(this, null)
+        return context.createIrBuilder(currentScope!!.scope.scopeOwnerSymbol).irBlock(expression.startOffset, expression.endOffset) {
+            val index = irTemporaryVar(irInt(0))
+            val sizeVar = irTemporary(size)
+            val result = irTemporary(irCall(sizeConstructor, expression.type).apply {
+                if (symbol.owner.typeParameters.isNotEmpty()) {
```
max-kammerer: copyTypeArgumentsFrom(expression)?

pyos:
@max-kammerer could you clarify what kind of tests you mean, please? E.g. for nonLocalReturns, should it be something like
```
fun box(): String {
    Array<Int>(1) { return "OK" }
    return "FAIL"
}
```
? There's already a test like that in box/arrays.

max-kammerer:
@pyos than it's redundant (could you point a test file name)?

For reification:
```
// FILE: 1.kt
package test

inline fun <reified T> f(x : () -> T): Array<T> {
    return Array(1) {
        x()
    }
}

// FILE: 2.kt

import test.*

fun box(): String {
    return f { "OK" }[0]    
}
```

pyos:
>(could you point a test file name)?

box/arrays/nonLocalReturnArrayConstructor.kt.

libraries/stdlib/js-ir/builtins/Arrays.kt
```diff
@@ -53,17 +59,6 @@ public class CharArray(size: Int) {
     public operator fun iterator(): CharIterator
 }
 
-@kotlin.internal.InlineOnly
```
max-kammerer: @pyos Seems this removal causes next fail: https://teamcity.jetbrains.com/viewLog.html?buildId=2236246&tab=buildResultsDiv&buildTypeId=Kotlin_dev_Compiler
//cc @anton-bannykh

anton-bannykh: @max-kammerer Those tests are built with bootstrap compiler. Builds successfully with bootstrap.local=true. Should be fixed by advancing the version.

max-kammerer:
@pyos Please rebase on master

pyos:
@max-kammerer done

max-kammerer:
@pyos Thanks!

PR #2280, opened 2019-04-16
Generate accessors for super calls if necessary

ting-yuan:
Current implementation of calls with super qualifier relies on
invokespecial, which has some more constraints than regular virtual
invocations. When those constraints aren't met, accessors are needed.


ting-yuan:
Hi @gogabr and @max-kammerer, would you mind to take a look when you have time? Thanks!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -212,20 +212,31 @@ private fun IrTypeParameter.copySuperTypesFrom(source: IrTypeParameter) {
...
 }
 
+// Copy value parameters, dispatch receiver, and extension receiver from source to value parameters of this function.
+// Type of dispatch receiver defaults to source's dispatch receiver. It is overridable in case the new function and the old one are used in
+// different contexts and expect different type of dispatch receivers. The overriding type should be assign compatible to the old type.
 fun IrFunction.copyValueParametersToStatic(
     source: IrFunction,
-    origin: IrDeclarationOrigin
+    origin: IrDeclarationOrigin,
+    dispatchReceiverType: IrType? = null
```
gogabr: Why not `source.dispatchReceiverParameter?.type`?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -212,20 +212,31 @@ private fun IrTypeParameter.copySuperTypesFrom(source: IrTypeParameter) {
...
 ) {
     val target = this
     assert(target.valueParameters.isEmpty())
 
     var shift = 0
     source.dispatchReceiverParameter?.let { p ->
```
max-kammerer: p -> originalDispatchReceiver

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -130,34 +146,49 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
             origin = JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR
             name = source.accessorName()
             visibility = Visibilities.PUBLIC
             isSuspend = source.isSuspend
         }.also { accessor ->
-            accessor.parent = source.parent
+            accessor.parent = if ((source.parent as? IrClass)?.isInterface == true) {
+                // Before Java 8, interfaces can't contain implementations.
+                // Place accessor in DefaultImpls for interfaces.
+                // TODO: remove this after targeting Java 8 or newer.
```
max-kammerer: NB: Accessors in interfaces could be only for private interface methods, they should be placed always in DefaulImpls, except cases when private methods annotated with @JvmDefault (only in such cases JVM default method is generated)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
ting-yuan: Assuming `C.foo()` is a fake override of `P.foo()`. `source.dispatchReceiverParameter?.type` is `P`. In `C.access@foo()`, passing it as type `P` to `C.foo()` would upset the bytecode verifier, because `P` isn't assign compatible to `C`.

Other approaches considered are:
1. Always use source.parent.type. Although all the tests passed, I'm not sure whether source.parent is always the desired class.
2. Insert a runtime cast. There are overheads, though.
3. In the accessor, change the call of the fake override to the real function. This is simplest, but it makes a few other errors. Some of them seem to require usage of `KotlinType`, so I didn't investigate further.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -387,13 +417,23 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
             return false
 
         val contextDeclarationContainer = allScopes.lastOrNull { it.irElement is IrDeclarationContainer }?.irElement
+        // context for functions in an interface is DefaultImpls
+        val adjustedContext = contextDeclarationContainer.let {
```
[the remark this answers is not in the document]
ting-yuan: > Any super call from an interface should be a static call to the super-interface's DefaultImpls by this point.

May I know how to determine if a call is static or not? I tried to check whether it has a dispatch receiver as well as IrFunction.isStatic (almost the same), but with [simpleCallWithHierarchy](https://github.com/JetBrains/kotlin/blob/master/compiler/testData/compileKotlinAgainstKotlin/jvm8/jvm8against6/simpleCallWithHierarchy.kt), the callee `Test.test` doesn't look like static but is generated with invokestatic. In that example, without adjusting the context, it would generate an unnecessary accessor.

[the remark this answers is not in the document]
gogabr: 1. A static function is indeed one that is `isStatic` (it should not have a `dispatchReceiverParameter`), and a static call is one without a `dispatchReceiver`. What exceptions do you have?
2. The example in question does not have an `INVOKESPECIAL`. In fact, `INVOKESTATIC` was generated there only because the current descriptor-based type mapper is clever enough to recognize interface super calls. A better way is to lower them at an earlier stage. Here is a branch that does exactly that: https://github.com/JetBrains/kotlin/tree/rr/gogabr/interface-super-calls

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
gogabr: My concern was not so much the program logic, but clarity of the code. Here, passing `null` invokes a special behaviour, and it is not clear from the function header what that behaviour is.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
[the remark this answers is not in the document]
ting-yuan: @gogabr Thanks, that's what I'm looking for. I've been trying to adjust the logic according to typemapper's behavior. Apparently I was in a wrong direction.

I'll undo the context adjustment and wait for your patch to land. Please correct me if I misunderstand anything.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
ting-yuan: Thanks for the clarification. I've updated it accordingly. Please take a look again.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
ting-yuan: Updated the comments accordingly. Please take a look again.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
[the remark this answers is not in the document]
ting-yuan: @gogabr Thanks for the branch! I've tested the updated PR with your branch and it works well.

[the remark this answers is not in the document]
gogabr: That PR is now merged into master.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -394,13 +423,20 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
             return false
 
         val contextDeclarationContainer = allScopes.lastOrNull { it.irElement is IrDeclarationContainer }?.irElement
+//        val contextDeclarationContainer = allScopes.lastOrNull { it.irElement is IrDeclarationContainer }?.irElement.let {if (it is IrClass && it.origin == JvmLoweredDeclarationOrigin.DEFAULT_IMPLS) it.parent else it} as IrElement
```
max-kammerer: @ting-yuan Obsolete?

ting-yuan: @max-kammerer Oops. Thanks for catching this.

max-kammerer:
@ting-yuan Thanks!

PR #2283, opened 2019-04-19
Optimize JVM byte code generation for conditional conjunction

neetopia:
Implement an intrinsic method for boolean.and operation, and replace
ANDAND condition with a call to such intrinsic method.

Also added a new test case to make sure existing non-shortcircuit "and" function is not being shadowed by other functions.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/IrIntrinsicMethods.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/IrIntrinsicMethods.kt
@@ -31,6 +31,10 @@ class IrIntrinsicMethods(irBuiltIns: IrBuiltIns) {
 
     val intrinsics = IntrinsicMethods()
 
+    val andandSymbol = irBuiltIns.run { defineOperator(OperatorNames.ANDAND, bool, listOf(bool, bool)) }
+
+    private val andand = andandSymbol.descriptor
+
     private val irMapping = hashMapOf<CallableMemberDescriptor, IntrinsicMethod>()
 
     private fun createPrimitiveComparisonIntrinsics(typeToIrFun: Map<SimpleType, IrSimpleFunctionSymbol>, operator: KtSingleValueToken) {
@@ -55,6 +59,8 @@ class IrIntrinsicMethods(irBuiltIns: IrBuiltIns) {
         irMapping[irBuiltIns.noWhenBranchMatchedException] = IrNoWhenBranchMatchedException()
         irMapping[irBuiltIns.illegalArgumentException] = IrIllegalArgumentException()
         irMapping[irBuiltIns.throwNpe] = ThrowNPE()
+
+        irMapping[andand] = AndAnd
     }
 
     fun getIntrinsic(descriptor: CallableMemberDescriptor): IntrinsicMethod? {
@@ -63,5 +69,9 @@ class IrIntrinsicMethods(irBuiltIns: IrBuiltIns) {
             return intrinsics.getIntrinsic(DescriptorUtils.unwrapFakeOverride(descriptor.correspondingProperty))
         }
         return irMapping[descriptor.original]
+    }
+
+    private object OperatorNames {
+        const val ANDAND = "ANDAND"
     }
 }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
@@ -15,13 +15,12 @@ import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrBlockImpl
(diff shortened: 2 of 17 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -102,6 +102,20 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                 expression.branches.removeIf() {
                     it.condition.isFalseConst() && isCompilerGenerated
                 }
+                // Replace conjunction condition with intrinsic and function call
+                if (expression.type.isBoolean() && expression.branches.size == 2) {
```
max-kammerer: @neetopia Please add '// a && b == if (a) b else false' clarification comment

neetopia:
updated addressing comment.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -102,6 +102,21 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                 expression.branches.removeIf() {
                     it.condition.isFalseConst() && isCompilerGenerated
                 }
+                if (expression.type.isBoolean() && expression.branches.size == 2) {
+                    // Replace conjunction condition with intrinsic "and" function call
+                    // the pattern for conjunction is "a && b == if (a) b else false"
+                    if (expression.branches[1].condition.isTrueConst() && expression.branches[1].result.isFalseConst()) {
```
max-kammerer: @neetopia Sorry, yesterday I also missed next case when debug info for 'false' would be lost and additional redundant (IFEQ) instuction would be generated:
```
fun cond() = false

fun foo() {
    if (cond())
        cond()
    else
         false
}
```
It should be checked that expression origin is IrStatementOrigin.ANDAND
Please also add custom linenumber test for that: compiler/testData/lineNumber/custom

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrBuiltIns.kt
```diff
@@ -198,6 +198,7 @@ class IrBuiltIns(
 
     val eqeqeqSymbol = defineOperator(OperatorNames.EQEQEQ, bool, listOf(anyN, anyN))
     val eqeqSymbol = defineOperator(OperatorNames.EQEQ, bool, listOf(anyN, anyN))
+    val andandSymbol = defineOperator(OperatorNames.ANDAND, bool, listOf(bool, bool))
```
udalov: Since this operator is only used in JVM, could you please declare it in `IrIntrinsicMethods`? Similarly to how for example JS operator intrinsics are declared in [`JsIntrinsics`](https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/JsIntrinsics.kt)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
neetopia: Good catch, the previous condition is too general, addressed.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrBuiltIns.kt
neetopia: Addressed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
max-kammerer: @neetopia Please also add assertion for expected structure (otherwise it could be conflict in future with other optimizations)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/IrIntrinsicMethods.kt
```diff
@@ -64,4 +69,8 @@ class IrIntrinsicMethods(irBuiltIns: IrBuiltIns) {
         }
         return irMapping[descriptor.original]
     }
+
+    object OperatorNames {
```
udalov: `private`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -102,6 +101,22 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                 expression.branches.removeIf() {
                     it.condition.isFalseConst() && isCompilerGenerated
                 }
+                if (expression.origin == IrStatementOrigin.ANDAND) {
+                    assert(expression.type.isBoolean()
```
udalov: Please always add messages with additional information to asserts, including a description of what's wrong and any relevant variables (e.g. `expression.dump()` here)

max-kammerer:
@neetopia Many thanks!

PR #2284, opened 2019-04-19
KT-30975 ''when' has only 'else' branch and should be simplified' inspection removes subject variable definition used in else branch

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30975

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
@@ -19,10 +19,7 @@ import org.jetbrains.kotlin.idea.intentions.branchedTransformations.unwrapBlockO
 import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
-import org.jetbrains.kotlin.psi.psiUtil.containingClass
-import org.jetbrains.kotlin.psi.psiUtil.getChildrenOfType
-import org.jetbrains.kotlin.psi.psiUtil.getParentOfType
-import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.psi.psiUtil.*
 import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsExpression
 import org.jetbrains.kotlin.resolve.calls.callUtil.getType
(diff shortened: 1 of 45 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -157,27 +159,59 @@ private fun KtExpression.enumEntry(): KtEnumEntry? {
...
 fun KtExpression.replaceWithBranch(branch: KtExpression, isUsedAsExpression: Boolean, keepBraces: Boolean = false) {
-    val lastExpression = when {
-        branch !is KtBlockExpression -> replaced(branch)
+    val caretModel = findExistingEditor()?.caretModel
+
+    val subjectVariable = (this as? KtWhenExpression)?.subjectVariable?.let { it ->
+        if (it.annotationEntries.isNotEmpty()) return@let it
+        val initializer = it.initializer ?: return@let it
+        val references = ReferencesSearch.search(it, LocalSearchScope(this)).toList()
+        if (references.size > 1) return@let it
```
dimonchik0036: I think it'd be more useful to replace `toList` with `singleOrNull`.

idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
```diff
@@ -157,27 +159,59 @@ private fun KtExpression.enumEntry(): KtEnumEntry? {
...
+    }
+
+    val factory = KtPsiFactory(this)
+    val parent = this.parent
+    val replaced = when {
+        branch !is KtBlockExpression -> {
+            if (subjectVariable != null) {
+                if (isUsedAsExpression) {
+                    replaced(KtPsiFactory(this).createExpressionByPattern("run { $0\n$1 }", subjectVariable, branch))
+                } else {
```
dimonchik0036: I think we shouldn't expand the scope for subjectVariable.
For example:
``` Kotlin
when (val a = 42) {
    else -> use(a, a)
}

val a = 33
```

t-kameyama: I want to return null if references size is 0.

idea/src/org/jetbrains/kotlin/idea/inspections/ConstantConditionIfInspection.kt
t-kameyama: I want to return null if references size is 0.

idea/src/org/jetbrains/kotlin/idea/inspections/WhenWithOnlyElseInspection.kt
```diff
@@ -24,15 +30,28 @@ class WhenWithOnlyElseInspection : AbstractKotlinInspection() {
 
             val usedAsExpression = expression.isUsedAsExpression(expression.analyze())
 
-            holder.registerProblem(expression,
```
dimonchik0036: I think we shouldn't change this file.

idea/testData/inspectionsLocal/whenWithOnlyElse/subjectVariable/blockElse/noReference.kt.after
```diff
@@ -0,0 +1,10 @@
+fun test() {
+    <caret>use("")
```
dimonchik0036: We can remove `create()` only if it hasn't side effects.

idea/testData/inspectionsLocal/whenWithOnlyElse/subjectVariable/hasSameNameVariable.kt
```diff
@@ -0,0 +1,9 @@
+// PROBLEM: none
+fun test() {
```
dimonchik0036: We should convert to
```kotlin
fun test() {
    run {
        val a = 42
        use(a, a)
    }
    val a = 33
}
```

dimonchik0036:
We do not have good tools to check for side effects now, so this is another issue. I accept this PR.

dimonchik0036:
Thanks! e955dcfc147bffc9121610cf5e9a594f789ba520

PR #2285, opened 2019-04-19
Add ULong.compareTo and UInt.div/rem/compareTo intrinsics

JakeWharton:


JakeWharton:
Looking at the original intrinsics, I'm missing black-box tests. Will re-open once ready.

PR #2288, opened 2019-04-19
Minimal tweaks in IR Validation code just to be able to turn it on in…

alexander-gorshenev:
… Native.


dnpetrov:
Looks good to me. I'll see if we can make IrConst types better.

PR #2289, opened 2019-04-19
Add ULong.compareTo and UInt.div/rem/compareTo intrinsics for Java 8

JakeWharton:


max-kammerer:
@JakeWharton @dnpetrov Is used Java 8 methods supported by d8 desugarer?
Seems it would be same problem with application running on old Android versions as with hashCode optimization that uses JVM 8 api: https://issuetracker.google.com/issues/114711928#comment7

JakeWharton:
I have a desugaring change ready to submit to D8 provided this is accepted. I recently added support to D8 for `Long.divideUnsigned` and `Long.remainderUnsigned` which were already in use.

JakeWharton:
Here's the D8 change, for reference: https://r8-review.googlesource.com/c/r8/+/37540. I didn't want to send it for review before it was clear that more intrinsics to Java 8 methods would be accepted.

max-kammerer:
@JakeWharton  Many Thanks!

udalov:
A follow-up to reduce the number of tests: https://github.com/JetBrains/kotlin/commit/95feb64fee7c6e67938c594d88639ea61b9e6387

PR #2290, opened 2019-04-20
IR lowering for computing trimIndent/trimMargin on constants

JakeWharton:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -107,6 +107,7 @@ val jvmPhases = namedIrFilePhase(
             jvmTypeOperatorLoweringPhase then
             foldConstantLoweringPhase then
             flattenStringConcatenationPhase then
+            computeStringTrimPhase then
```
JakeWharton: I wanted to add this for JS but I don't know how to test it. Would appreciate a pointer so I can follow-up.

And for native I assume I wait until they update to a version of compiler that contains this phrase and then I can update their list?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/StringTrimLowering.kt
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        return IrConstImpl.string(call.startOffset, call.endOffset, call.type, newString)
+    }
+
+    companion object {
+        private fun IrExpression.toConstantString(): String? {
+            return when (this) {
+                is IrConst<*> -> IrConstKind.String.valueOf(this)
+                is IrStringConcatenation -> toConstantString()
```
JakeWharton: I feel like this should be its own lowering phase for all platforms. Currently this phrase gets IR like:
```
$receiver: STRING_CONCATENATION type=kotlin.String
  CONST String type=kotlin.String value="\n        |Hello,\n        |"
  CONST String type=kotlin.String value="Hello,\n        |World"
  CONST String type=kotlin.String value="\n        |World\n    "
```
when it really should only have to care about / check for `IrConst<*>`.

JakeWharton: Just found https://github.com/JetBrains/kotlin/pull/2274 which seems like it should partially handle this, although it omits support for all constants types.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
max-kammerer: @romanart @bashor ^^

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -107,6 +107,7 @@ val jvmPhases = namedIrFilePhase(
             jvmTypeOperatorLoweringPhase then
             flattenStringConcatenationPhase then
             foldConstantLoweringPhase then
+            computeStringTrimPhase then
```
max-kammerer: @JakeWharton 
As order of lowers is sufficient here, please add also some tests with constant inlining and append operations, e.g.:
```
private const val SPACES = "             "


    """
        ${SPACES + SPACES}
        Hello,
        World
        ${SPACES + SPACES}
    """.trimIndent()

(SPACES + """
        ${SPACES + SPACES}
        Hello,
        World
        ${SPACES + SPACES}
    """ + SPACES + ).trimIndent()

```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/StringTrimLowering.kt
```diff
@@ -0,0 +1,86 @@
...
+    override fun visitCall(expression: IrCall): IrExpression {
+        return when {
+            trimIndentMatcher(expression) -> maybeComputeTrimIndent(expression)
+            trimMarginMatcher(expression) -> maybeComputeTrimMargin(expression)
+            else -> super.visitCall(expression)
+        }
+    }
+
+    private fun maybeComputeTrimIndent(call: IrCall): IrExpression {
+        val receiverString = call.extensionReceiver!!.getConstantString() ?: return call
```
max-kammerer: Seems it's also possible to make partial removal of margins and indents if we have constant entries at beginning and ending in IrStringConcatenation

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
bashor: K/JS pipeline is defined in [JsLoweringPhases.kt:349](https://github.com/JetBrains/kotlin/blob/a16ca5e66cd3c17a43d15bd4112da86f1b43059d/compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/JsLoweringPhases.kt#L349)
You can run JS test using following command `./gradlew :js:js.tests:jsIrTest`.

For JS IR BE we don't have yet infrastructure for such tests. For the current BE we have special directives for that, like `CHECK_NOT_CALLED` (e.g. [inlineCallNoInline.kt:4](https://github.com/JetBrains/kotlin/blob/b389e29db567024cb90e7b1f395e8f47e59014af/js/js.translator/testData/box/inline/inlineCallNoInline.kt#L4)) and `CHECK_NOT_CALLED_IN_SCOPE`(e.g. [lambdaInLambda.kt:5](https://github.com/JetBrains/kotlin/blob/b389e29db567024cb90e7b1f395e8f47e59014af/js/js.translator/testData/box/inline/lambdaInLambda.kt#L5)) and we going to support/adopt them for JS IR BE. You can try to do something in that direction but I think you will face issues in other places unrelated to this PR.

bashor: Alternatively, you can write something like that:

```
val f = {
    """
        Hello,
        World
    """.trimIndent()
}

fun box(): String {
    if (f.toString().contains("(")) return "fail"
    return "OK"

}
```

bashor: if you use the last way please leave a TODO that it should be rewritten it using `CHECK_NOT_CALLED`

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/StringTrimLowering.kt
JakeWharton: I'm not sure this is safe. It also leads to complex rewrites where you have to move the trimIndent/Margin call onto the non-constant replacement characters.

Consider
```kotlin
fun foo(a: String): String {
  return """
    indented by 4
      indented by 6
$a
    indented by 4
    """.trimIndent()
}
```
If `a` has an indent of 2 spaces then only 2 spaces should be removed from the prefix/suffix constants. If `a` has an indent of 8 spaces then 4 spaces will correctly be removed from the prefix/suffix constants, but `a` itself will have an indent of 8 spaces which will be removed instead of only 4.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
JakeWharton: Thanks for the info! If you don't mind, I'll investigate this as a follow-up change.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
JakeWharton: Updated!

max-kammerer:
@JakeWharton Do you plan to add JS part in this pr or as separate one, so this one could be merged?

JakeWharton:
Separately, if that's okay. I won't be able to start looking at it for a week or two.

max-kammerer:
Yes, it's ok. Thanks you!

PR #2293, opened 2019-04-24
KT-12291 Override/Implement Members: better member positioning inside the class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12291

### DIFF
--- a/idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
+++ b/idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
@@ -25,13 +25,24 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.DialogWrapper
 import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
+import com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
 import org.jetbrains.kotlin.idea.core.insertMembersAfter
+import org.jetbrains.kotlin.idea.search.usagesSearch.descriptor
 import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.psiUtil.getNonStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.prevSiblingOfSameType
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.resolve.lazy.descriptors.LazyClassDescriptor
+import org.jetbrains.kotlin.resolve.source.getPsi
+import org.jetbrains.kotlin.util.findCallableMemberBySignature
+import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
 
 abstract class OverrideImplementMembersHandler : LanguageCodeInsightActionHandler {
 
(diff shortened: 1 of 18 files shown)


idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
```diff
@@ -106,7 +117,59 @@ abstract class OverrideImplementMembersHandler : LanguageCodeInsightActionHandle
...
+                    it.parent == classBody
+                }
+                if (offsetCursorElement != null && offsetCursorElement != classBody.rBrace) {
+                    insertMembersAfter(editor, classOrObject, selectedMemberDescriptors.keys)
+                    return
+                }
+            }
+            val classLeftBrace = classBody.lBrace
+
+            val allSuperMemberDescriptors = selectedMemberDescriptors.values
```
dimonchik0036: You should use like this:
```kotlin
val allSuperMemberDescriptors = selectedMemberDescriptors.values
    .mapNotNull { it.containingDeclaration as? ClassDescriptor }
    .associateWith {
        DescriptorUtils.getAllDescriptors(it.unsubstitutedMemberScope).filterIsInstance<CallableMemberDescriptor>()
    }
```

dimonchik0036: Otherwise, you will only get functions that are in custom classes. For example: members from `Any` will be ignored.

idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
```diff
@@ -106,7 +117,59 @@ abstract class OverrideImplementMembersHandler : LanguageCodeInsightActionHandle
...
+
+            fun ClassDescriptor.findElement(memberDescriptor: CallableMemberDescriptor): KtDeclaration? {
+                return implementedElements[memberDescriptor]
+                    ?: (findCallableMemberBySignature(memberDescriptor)?.source?.getPsi() as? KtDeclaration)?.also {
+                        implementedElements[memberDescriptor] = it
+                    }
+            }
+
+            fun getAnchor(selectedElement: KtDeclaration): PsiElement? {
+                val lastElement = classOrObject.declarations.lastOrNull()
```
dimonchik0036: You should return `null` by default for case:
```kotlin
class Other {
    fun test() {
        val a = 1
    }
<caret>

    fun otherTest() {

    }

}
```

idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
t-kameyama: Fixed

idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
t-kameyama: This function will not be executed because it is returned here:
https://github.com/t-kameyama/kotlin/blob/9fb07118db46f84598a87326bc2e6410ab3a93d4/idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt#L133-L136

idea/idea-core/src/org/jetbrains/kotlin/idea/core/overrideImplement/OverrideImplementMembersHandler.kt
```diff
@@ -106,7 +117,59 @@ abstract class OverrideImplementMembersHandler : LanguageCodeInsightActionHandle
                 selectedElements: Collection<OverrideMemberChooserObject>,
                 copyDoc: Boolean
         ) {
-            insertMembersAfter(editor, classOrObject, selectedElements.map { it.generateMember(classOrObject, copyDoc) })
+            val selectedMemberDescriptors = selectedElements.associate { it.generateMember(classOrObject, copyDoc) to it.descriptor }
+
+            val classBody = classOrObject.body
+            if (classBody == null) {
```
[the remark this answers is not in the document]
t-kameyama: In `else` branch check is necessary for https://github.com/JetBrains/kotlin/pull/2293#discussion_r317347393

PR #2296, opened 2019-04-25
Implement stub methods generation for Kotlin Immutable Collection cla…

neetopia:
Implement stub methods generation for Kotlin Immutable Collection classes.

This change is to fill the gap between Kotlin Collection
classes(immutable) and Java Collection classes(mutable), to avoid
calling an unsupported operation like remove() on an immutable class in
jvm.

Also extracted a common utility for type erasure to IrUtility


compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -89,3 +92,38 @@ fun IrType.getPrimitiveArrayElementType() = (this as? IrSimpleType)?.let {
 
 fun IrType.isNonPrimitiveArray() =
     (this.isArray() || this.isNullableArray()) && !this.isPrimitiveArray()
+
+fun IrType.eraseTypeParameters() = when (this) {
```
udalov: Please specify explicit return type, provide a few examples in the kdoc and probably rename function to emphasize what exactly this function does since it's not immediately obvious.

Also, the behavior seems very much JVM-specific, consider moving to JVM IR type utils.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -89,3 +92,38 @@ fun IrType.getPrimitiveArrayElementType() = (this as? IrSimpleType)?.let {
...
+            }
+            else -> error("Unknown IrSimpleType classifier kind: $owner")
+        }
+    }
+    else -> error("Unknown IrType kind: $this")
+}
+
+fun IrTypeParameter.upperBoundClass(): IrClass {
+    // Look into super types for possible upperbound
+    val simpleSuperClassifiers = superTypes.asSequence().filterIsInstance<IrSimpleType>().map { it.classifier }
```
udalov: `val superClassifiers = superTypes.mapNotNull { it.classifierOrNull }`

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -89,3 +92,38 @@ fun IrType.getPrimitiveArrayElementType() = (this as? IrSimpleType)?.let {
...
+    return simpleSuperClassifiers
+        .filterIsInstance<IrClassSymbol>()
+        .let {
+            // First try to find a non-interface class in super types, return it as upperbound.
+            it.firstOrNull { !it.owner.isInterface } ?: it.firstOrNull()
```
udalov: 1) Please don't use nested `it`s, declare one of them with a name instead
2) Take a look at [`representativeUpperBound`](https://github.com/JetBrains/kotlin/blob/373424f4304784cfda7fd7219838dbdd728bed54/core/descriptors/src/org/jetbrains/kotlin/types/TypeUtils.kt#L236) which implements similar behavior in compiler frontend. It also checks if the upper bound is an annotation (which is technically an interface on JVM). Also, consider even extracting this part of the function to a new function `IrTypeParameter.representativeUpperBound: IrType` just to be clear that this is exactly the same behavior, and then maybe moving the rest (possibly renaming to `representativeUpperBoundClass`) to JVM-specific utils

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -89,3 +92,38 @@ fun IrType.getPrimitiveArrayElementType() = (this as? IrSimpleType)?.let {
...
+        }?.owner ?: // Try to find first superType of interface as upperbound
+    // If an upperbound for above condition is not found,
+    // upperbound is in type parameter, dive into type parameters for chained type parameters like U:V, V:String
+    simpleSuperClassifiers.filterIsInstance<IrTypeParameterSymbol>().map { it.owner.upperBoundClass() }.first()
```
udalov: Instead of invoking `map` on the whole list and taking only the first element from the result, it'd be better to get the first element from the list and run the lambda on it

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
udalov: UPD: I've noticed this code was already there in BridgeLowering before your changes. In that case, feel free to leave everything as is to avoid breaking something else. However, I feel very nervous about exposing this functionality in `backend.common`, I'd leave it in `backend.jvm` for now

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,216 @@
...
+                    if (mutableMethod.owner.toSignature() !in readOnlyMethodSignatures) {
+                        mutableMethods.add(mutableMethod.owner)
+                    }
+                }
+                StubsForCollectionClass(readOnlyClass, mutableClass, mutableMethods)
+            }
+        }
+    }
+
+    private fun findRelevantSuperCollectionClasses(irClass: IrClass): Collection<StubsForCollectionClass> {
```
gogabr: In a hierarchy of Kotlin classes deriving from a collection, only the most general one requires stub methods. Thus, in
```
class K1: Map
class K2: K1
class K3: K2
```
only `K1` needs stubs.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,216 @@
...
+            parent = target
+        }
+    }
+
+    // Create a map for type parameters between source class (Mutable Collection classes) to
+    // target class (class currently in lowering phase), this map is later used for substituting type parameters in generated functions
+    private fun createTypeMap(sourceClass: IrClass, targetClass: IrClass): Map<IrTypeParameterSymbol, IrType> {
+        var typeList = (targetClass.superTypes.filter { target ->
+            sourceClass.superTypes.any { it.classifierOrNull == target.classifierOrNull }
+        }.firstOrNull() as? IrSimpleType)?.arguments
```
gogabr: Why do you need to search `mutableK` and `K` at all? You already have them when you search for stubs to generate.

The search procedure exploits the fact that for each `K`, `mutableK` is a direct subclass of `K`. You should mention this in a comment.
Moreover, this will only work when `target` is a direct subclass of `K` -- a condition that, as I remark below, should be imposed, but currently is not.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,216 @@
...
+    private val state = context.state
+
+    private val typeMapper = state.typeMapper
+
+    override fun lower(irClass: IrClass) {
+        if (irClass.isInterface) {
+            return
+        }
+
+        val methodStubsToGenerate = computeStubsToGenerate(irClass)
```
gogabr: -- This is the most general comment, everything else only makes sense if you don't go the way described here. --
Why don't you look through _direct_ superclasses of irClass and see if any of them is an immutable collection?
Then you will immediately have the list of methods to override and type arguments to substitute.
(Will only fail if there is an intermediate Java class implementing a Kotlin immutable collection interface.)

gogabr: Sorry, take this back. This simple picture does not take interfaces into account.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
neetopia: Since we have intermediate Java class use case, I just replaced type map calculation logic with a recursive one.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
neetopia: Makes sense, moved to backend.jvm

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,241 @@
...
+                list to mutableList,
+                map to mutableMap,
+                mapEntry to mutableMapEntry,
+                iterable to mutableIterable,
+                iterator to mutableIterator,
+                listIterator to mutableListIterator
+            ).map { (readOnlyClass, mutableClass) ->
+                val mutableMethods = mutableListOf<IrSimpleFunction>()
+                val readOnlyMethodSignatures = readOnlyClass.functions.map { it.owner.toSignature() }.toHashSet()
+                for (mutableMethod in mutableClass.functions) {
```
gogabr: val mutableMethods = mutableClass.functions.filter { it.toSignature() !in readOnlyMethodSignatures }

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,241 @@
...
+                    if (mutableMethod.owner.toSignature() !in readOnlyMethodSignatures) {
+                        mutableMethods.add(mutableMethod.owner)
+                    }
+                }
+                StubsForCollectionClass(readOnlyClass, mutableClass, mutableMethods)
+            }
+        }
+    }
+
+    private fun findRelevantSuperCollectionClasses(irClass: IrClass): HashSet<IrSimpleFunction> {
```
gogabr: If a set of functions is returned, why is the function called `find...Classes()`?
Also, I don't see any reason to specify a particular implementation of set.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,241 @@
...
+    private fun createTypeList(sourceClass: IrClass, targetClass: IrClass): List<IrTypeArgument>? {
+        // First try to find if current class is a direct subtype of source class
+        return (targetClass.superTypes.filter { target ->
+            sourceClass.symbol == target.classifierOrNull
+        }.firstOrNull() as? IrSimpleType)?.arguments ?:
+        // If not direct subtype, recursively look into super types
+        targetClass.superTypes
+            .mapNotNull { it.classOrNull?.owner }
+            .mapNotNull { createTypeList(sourceClass, it) }
+            .firstOrNull()
```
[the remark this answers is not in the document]
gogabr: Come to think of it, you don't need any complicated substitution mechanics in this particular case. We just need the type arguments to one of the standard collection interfaces.
For subtypes of `Collection`, it is sufficient to look at the argument type of `contains()`.
For `Map`, we need to check `containsKey()` and `containsValue`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        val methodStubsToGenerate = findRelevantStubMethods(irClass)
+
+        // We don't need to generate stub for existing methods, but for FAKE_OVERRIDE methods with ABSTRACT modality,
+        // it means an abstract function in superclass that is not implemented yet,
+        // stub generation is still needed to avoid invocation error.
+        val existingMethodSignatures = irClass.functions.filter {
+            it.modality != Modality.ABSTRACT || it.origin != IrDeclarationOrigin.FAKE_OVERRIDE
+        }.associate { it.toSignature() to it }.toMap()
```
udalov: `.associate { it.toSignature() to it }` -> `.associateBy { it.toSignature() }`

Also consider using `filterNot` instead of `filter` above

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,274 @@
...
+    // target class (class currently in lowering phase), this map is later used for substituting type parameters in generated functions
+    private fun createTypeMap(collectionClass: IrClass, mutableClass: IrClass, targetClass: IrClass): Map<IrTypeParameterSymbol, IrType> {
+        // It is guaranteed by compiler that a class inherits from an immutable collection class must have
+        // some methods overridden, therefore we search for such overriddens, and get substitute type for
+        // our stubs based on these functions. For every collection class we select specific function(s), to make sure all
+        // type parameters are covered by our selection.
+        val targetTypeList = when (collectionClass.name.asString()) {
+            "Iterator", "ListIterator" -> {
+                listOf("next")
+                    .map { functionName -> getFunctionForSubstitution(functionName, collectionClass, targetClass) }
```
udalov: Please collapse successive `.map` calls into one for readability and to avoid extra allocations, e.g.:
```
    listOf("next")
        .map { name -> getFunctionForSubstitution(name, collectionClass, targetClass).returnType }
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,274 @@
...
+        return mutableClass.typeParameters.map { it.symbol }.zip(targetTypeList).toMap()
+    }
+
+    // we can write function with same amount of value arguments and same name as those defined in collection interface,
+    // therefore we need to check for overridden symbols to make sure we are looking at the correct one.
+    private fun getFunctionForSubstitution(functionName: String, collectionClass: IrClass, targetClass: IrClass) =
+        targetClass.functions
+            .filter { function -> function.name.asString() == functionName }
+            .filter { function -> overridesFunctionInImmutable(collectionClass, function) }
+            .first()
```
udalov: We usually prefer to use less calls and lambdas if it doesn't make the code heavier or harder to understand. Also, instead of filtering something into a collection and then taking the first element, you can simply take the first element that satisfies the predicate:
```
        targetClass.functions.first { function ->
            function.name.asString() == functionName && overridesFunctionInImmutable(collectionClass, function)
        }
```

udalov: Also, consider using `firstOrNull` instead of `first`, and adding `?: error("...")` with an error description mentioning the class and what happened

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,274 @@
...
+            }
+            "Map" -> {
+                listOf("containsKey", "containsValue")
+                    .map { functionName -> getFunctionForSubstitution(functionName, collectionClass, targetClass) }
+                    .map { function -> function.valueParameters[0].type }
+            }
+            else -> throw IllegalStateException()
+        }
+        if (targetTypeList.isNullOrEmpty() || targetTypeList.size != mutableClass.typeParameters.size) {
+            throw IllegalStateException()
```
udalov: Please always add description and data from the context to explain what went wrong

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,274 @@
...
+            isNoinline
+        ).apply {
+            descriptor.bind(this)
+            parent = target
+        }
+    }
+
+    // Create a map for type parameters between source class (Mutable Collection classes) to
+    // target class (class currently in lowering phase), this map is later used for substituting type parameters in generated functions
+    private fun createTypeMap(collectionClass: IrClass, mutableClass: IrClass, targetClass: IrClass): Map<IrTypeParameterSymbol, IrType> {
```
udalov: I'd prefer to avoid relying on functions with specific hardcoded names to find out which argument is used for the collection's type parameter. I think the code would be easier and more future-proof if we constructed the list of all (substituted) supertypes of `targetClass`, picked the one supertype whose class is `mutableClass`, and extracted all type arguments from it. Pretty much what the existing code in the old backend does here: https://github.com/JetBrains/kotlin/blob/6dbbf0bb7250b1c8ca11a1a925eb3fe468f2e89a/compiler/backend/src/org/jetbrains/kotlin/codegen/CollectionStubMethodGenerator.kt#L97

neetopia: That's also my concern for this approach as well. Regarding the code you pointed, when I looked into that code during initial investigation, I feel the old logic is heavily relying on descriptor, which is what we want to get away with new back end, therefore I am not sure if we want to follow that logic.

udalov: Right. My suggestion was not to use that code directly, but reimplement it for `IrType`

neetopia: @udalov Can you take a look and let me know if latest revision is what you are expecting? Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+    }
+
+    // Compute stubs that should be generated, compare based on signature
+    private fun findRelevantStubMethods(irClass: IrClass): Set<IrSimpleFunction> {
+        val ourStubsForCollectionClasses = preComputedStubs.filter { (readOnlyClass, mutableClass) ->
+            irClass.superTypes.any { superTypeSymbol ->
+                val superType = superTypeSymbol.classOrNull?.owner
+                if (superType == null) {
```
udalov: I'd prefer to merge this condition into the else-branch:
```
val superType = superTypeSymbol.classOrNull?.owner
superType != null
        && (superType.comesFromJava() || superType.isInterface)
        && superTypeSymbol.isSubtypeOfClass(readOnlyClass)
        && !irClass.symbol.isSubtypeOfClass(mutableClass)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+
+        // do a second filtering to ensure only most relevant classes are included.
+        val redundantClasses = ourStubsForCollectionClasses.filter { (readOnlyClass) ->
+            ourStubsForCollectionClasses.any { readOnlyClass != it.readOnlyClass && it.readOnlyClass.isSubtypeOfClass(readOnlyClass) }
+        }.map { it.readOnlyClass }
+
+        // perform type substitution and type erasure here
+        return ourStubsForCollectionClasses.filter { (readOnlyClass) ->
+            readOnlyClass !in redundantClasses
+        }.flatMap {
```
udalov: `}.flatMap { (readOnlyClass, mutableClass, mutableOnlyMethods) ->` maybe to get rid of `it` below

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+            isNoinline
+        ).apply {
+            descriptor.bind(this)
+            parent = target
+        }
+    }
+
+    // Create a map for type parameters between source class (Mutable Collection classes) to
+    // target class (class currently in lowering phase), this map is later used for substituting type parameters in generated functions
+    private fun createTypeMap(readOnlyClass: IrClass, mutableClass: IrClass, targetClass: IrClass): Map<IrTypeParameterSymbol, IrType> {
```
udalov: I propose to use the existing term from our compiler frontend and rename this function to `computeSubstitutionMap` and adapt the comment accordingly

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -119,3 +119,28 @@ fun IrType.substitute(substitutionMap: Map<IrTypeParameterSymbol, IrType>): IrTy
         newAnnotations
     )
 }
+
+fun getImmediateSupertypes(type: IrSimpleType): List<IrSimpleType> {
```
udalov: `private`

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -119,3 +119,28 @@ fun IrType.substitute(substitutionMap: Map<IrTypeParameterSymbol, IrType>): IrTy
...
+    val originalSuperTypes = type.classOrNull?.superTypes() ?: emptyList()
+    val args = type.arguments.mapNotNull { (it as? IrTypeProjection)?.type }
+    return originalSuperTypes
+        .filter { it.classOrNull != null }
+        .map { superType ->
+            superType.substitute(superType.classOrNull!!.owner.typeParameters, args) as IrSimpleType
+        }
+}
+
+fun collectAllSupertypes(type: IrSimpleType, result: HashSet<IrSimpleType>) {
```
udalov: `private`, and also please use `MutableSet` instead of `HashSet`

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -119,3 +119,28 @@ fun IrType.substitute(substitutionMap: Map<IrTypeParameterSymbol, IrType>): IrTy
...
+    val immediateSuperTypes = getImmediateSupertypes(type)
+    result.addAll(immediateSuperTypes)
+    immediateSuperTypes.forEach {
+        collectAllSupertypes(it, result)
+    }
+}
+
+fun getAllSuperTypes(type: IrSimpleType): HashSet<IrSimpleType> {
```
udalov: `HashSet` -> `MutableSet`

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
```diff
@@ -119,3 +119,28 @@ fun IrType.substitute(substitutionMap: Map<IrTypeParameterSymbol, IrType>): IrTy
...
+        .filter { it.classOrNull != null }
+        .map { superType ->
+            superType.substitute(superType.classOrNull!!.owner.typeParameters, args) as IrSimpleType
+        }
+}
+
+fun collectAllSupertypes(type: IrSimpleType, result: HashSet<IrSimpleType>) {
+    val immediateSuperTypes = getImmediateSupertypes(type)
+    result.addAll(immediateSuperTypes)
+    immediateSuperTypes.forEach {
```
udalov: Kotlin style guide:
> prefer using a regular `for` loop instead, unless the receiver of `forEach` is nullable or `forEach` is used as part of a longer call chain

https://kotlinlang.org/docs/reference/coding-conventions.html#using-loops

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/util/IrTypeUtils.kt
udalov: Also "supertype" is one word, so please rename to `getAllSupertypes`

udalov: Also, a type cannot have supertypes, but a class can. Let's change the parameter of these methods to `IrClass`, and either use `defaultType` or `typeParameters` in the implementation, to obtain the type parameters of the target class

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+        // We find the most specific type for the immutable collection class from the inheritance chain of target class
+        // Perform type substitution along searching, then use the type arguments obtained from the most specific type
+        // for type substitution.
+        val readOnlyClassType = getAllSuperTypes(targetClass.defaultType).findMostSpecificTypeForClass(readOnlyClass.symbol)
+        val targetTypeList = (readOnlyClassType as IrSimpleType).arguments.mapNotNull { (it as IrTypeProjection).type }
```
udalov: I suppose you meant `{ (it as? IrTypeProjection)?.type }` here, since otherwise `mapNotNull` could be replaced with `map`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+
+        if (targetTypeList.isEmpty() || targetTypeList.size != mutableClass.typeParameters.size) {
+            throw IllegalStateException("Type argument mismatch between immutable and mutable class")
```
udalov: Please always add context to exception messages. For example, at least `targetClass.fqNameWhenAvailable` here, but I'd also prefer to render some information about either `readOnlyClass` or `mutableClass`, and `targetTypeList`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
udalov: Also I'd rename the variable to `readOnlyClassTypeArguments`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    //TODO: replace with new typeMapper using no descriptor
+    private fun IrSimpleFunction.toSignature() = typeMapper.mapAsmMethod(this.descriptor).toString()
+
+    private fun createStubMethod(function: IrSimpleFunction, irClass: IrClass, typeMap: Map<IrTypeParameterSymbol, IrType>): IrSimpleFunction {
+        return buildFun {
+            name = function.name
+            returnType = function.returnType.substitute(typeMap)
```
udalov: Shouldn't we call `eraseTypeParameters` here as well? Or is it not relevant because none of the methods in mutable collection classes have return types with generic type arguments?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+            origin = IrDeclarationOrigin.IR_BUILTINS_STUB
+            modality = Modality.OPEN
+        }.apply {
+            // Replace Function metadata with the data from class
+            // Add the abstract function symbol to stub function for bridge lowering
+            overriddenSymbols.add(function.symbol)
+            parent = irClass
+            dispatchReceiverParameter = function.dispatchReceiverParameter?.copyWithTypeErasure(this, typeMap)
+            extensionReceiverParameter = function.extensionReceiverParameter?.copyWithTypeErasure(this, typeMap)
+            function.valueParameters.mapIndexed { i, param ->
```
udalov: I suppose you intended to write `mapIndexedTo(valueParameters) { ... }` here, because currently it maps the value parameter list to a list of booleans, adding the new value parameters as a side effect, which is sort of obscure :)

Let's use a simple `for` loop instead:
```
for (parameter in function.valueParameters) {
    valueParameters.add(parameter.copyWithTypeErasure(this, typeMap))
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+                    .map { it.owner }
+                    .filter { it.toSignature() !in readOnlyMethodSignatures }
+                    .toHashSet()
+                StubsForCollectionClass(readOnlyClass, mutableClass, mutableMethods)
+            }
+        }
+    }
+
+    // Compute stubs that should be generated, compare based on signature
+    private fun findRelevantStubMethods(irClass: IrClass): Set<IrSimpleFunction> {
```
udalov: I'd rename `find` to `generate` because it actually generates those methods

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -0,0 +1,211 @@
...
+
+        // We don't need to generate stub for existing methods, but for FAKE_OVERRIDE methods with ABSTRACT modality,
+        // it means an abstract function in superclass that is not implemented yet,
+        // stub generation is still needed to avoid invocation error.
+        val existingMethodSignatures = irClass.functions.filterNot {
+            it.modality == Modality.ABSTRACT && it.origin == IrDeclarationOrigin.FAKE_OVERRIDE
+        }.associateBy { it.toSignature() }
+
+        for (member in methodStubsToGenerate) {
+            if (member.toSignature() in existingMethodSignatures.keys) {
```
udalov: A minor simplification would be to get the value from a map, and then check if it isn't null:
```
val existingMethod = existingMethodSignatures[member.toSignature()]
if (existingMethod != null) {
    // ...
    existingMethod.overriddenSymbols.addAll(member.overriddenSymbols)
} else {
    irClass.declarations.add(member)
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
neetopia: It seems to be not relevant, but I extracted the type erasure logic from bridgelowering when I was using erasure in stub lowering, do you want me to put it back? Or we can keep it as it might be used in other occasions?

neetopia: Well I think I should put it back, since Steven is having a [PR](https://github.com/JetBrains/kotlin/pull/2339) for extracting erasure from bridge lowering already.

gogabr:
Thank you!

PR #2297, opened 2019-04-25
Remove a redundant JVM-specific lowering

pyos:
This case is handled by [this line in codegen](https://github.com/JetBrains/kotlin/blob/c2884f3ca83b11aefd5f268884b68c629ade8f0a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt#L692).


max-kammerer:
Thanks!

PR #2302, opened 2019-04-27
Use spaces instead of tabs to indent -Xlist-phases.

mpetrov:
In Terminal.app and xterm the output columns get misaligned when using tabs.
Tabbed output only renders correctly when `tabstop=1`.


gogabr:
Thank you!

PR #2303, opened 2019-04-28
Fix CoroutineContext documentation

elizarov:
KT-26398 had updated behavior from reference comparison to regular
equality, but the documentation for CoroutineContext was left out.
The whole "by reference" needs to be removed, because it now
follows default behavior of all Kotlin collections and need no
further clarification.


ilya-g:
Missed one in the `CoroutineContext` summary.

elizarov:
Good catch. Fixed.

PR #2305, opened 2019-04-29
Invert the while loop (into if + do-while) that is used when lowering for-loops over progressions that cannot overflow.

punzki:
This ensures that lowered for-loops while keep the same performance characteristics when we migrate Kotlin/Native to use the shared `ForLoopsLowering`. Performance-wise, the impact on JVM is slightly positive.

The code size impact is mixed: There is an additional `if`, but in cases where the upper bound is constant, the surrounding `if` can be optimized away and there is no unconditional `GOTO` at the bottom of the loop.

Since the bytecode for JVM non-IR vs IR backend is different (non-IR uses `while` loop), the bytecode tests had to be forked.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
@@ -149,6 +149,10 @@ internal class ProgressionLoopHeader(
                 //       // Loop body
                 //     } while (loopVar != last)
                 //   }
+                //
+                // This form can only be used when the loop is last-inclusive. We would need to decrement `last` in the right direction
+                // if it were last-exclusive. In any case, the induction variable cannot overflow if it were last-exclusive.
+                assert(headerInfo.isLastInclusive)
                 assert(loopVariable != null)
                 val booleanNotFun = context.irBuiltIns.booleanClass.functions.first { it.owner.name.asString() == "not" }
                 val newCondition = irCallOp(booleanNotFun, booleanNotFun.owner.returnType, irCall(context.irBuiltIns.eqeqSymbol).apply {
@@ -160,22 +164,28 @@ internal class ProgressionLoopHeader(
                     condition = newCondition
                     body = newBody
                 }
-                val notEmptyCheck = irIfThen(buildLoopCondition(builder), newLoop)
+                val notEmptyCheck = irIfThen(buildLoopCondition(this@with), newLoop)
                 LoopReplacement(newLoop, notEmptyCheck)
             } else {
-                // If the induction variable can NOT overflow, use a simple while loop. Loop is lowered into something like:
+                // If the induction variable can NOT overflow, use a do-while loop. Loop is lowered into something like:
                 //
-                //   while (inductionVar <= last) {
-                //       val loopVar = inductionVar
-                //       inductionVar += step
-                //       // Loop body
+                //   if (inductionVar <= last) {
+                //     do {
+                //         val loopVar = inductionVar
+                //         inductionVar += step
+                //         // Loop body
+                //     } while (inductionVar <= last)
                 //   }
-                val newLoop = IrWhileLoopImpl(oldLoop.startOffset, oldLoop.endOffset, oldLoop.type, oldLoop.origin).apply {
+                //
+                // Even though this can be simplified into a simpler while loop, using if + do-while (i.e., doing a loop inversion)
+                // performs better in benchmarks. In cases where `last` is a constant, the `if` may be optimized away.
+                val newLoop = IrDoWhileLoopImpl(oldLoop.startOffset, oldLoop.endOffset, oldLoop.type, oldLoop.origin).apply {
                     label = oldLoop.label
                     condition = buildLoopCondition(this@with)
                     body = newBody
                 }
-                LoopReplacement(newLoop, newLoop)
+                val notEmptyCheck = irIfThen(buildLoopCondition(this@with), newLoop)
+                LoopReplacement(newLoop, notEmptyCheck)
             }
 
             if (!headerInfo.isFirstInclusive) {
--- a/compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInObjectArrayIndicesIR.kt
+++ b/compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInObjectArrayIndicesIR.kt
@@ -1,4 +1,4 @@
-// IGNORE_BACKEND: JVM_IR
+// TARGET_BACKEND: JVM_IR
 fun test() {
     var sum = 0
     for (i in arrayOf("", "", "", "").indices) {
@@ -15,4 +15,5 @@ fun test() {
 // 0 getFirst
 // 0 getLast
 // 1 IF_ICMPGE
-// 1 IF
+// 1 IF_ICMPLT
+// 2 IF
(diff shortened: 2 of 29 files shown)


max-kammerer:
@punzki Could you please a bit describe measurement environment: used JDK versions (maybe hardware specification), do you perform any measurment on Android (Dalvik and Art runtimes)?

punzki:
> @punzki Could you please a bit describe measurement environment: used JDK versions (maybe hardware specification), do you perform any measurment on Android (Dalvik and Art runtimes)? Which benchmark do you use (is it original benchmarks from K/N or there are some additional cases)?

These are the original benchmarks from K/N, under `kotlin-native/performance/ring`. To run the benchmarks under JVM IR, I had to modify the code since there were some bugs (e.g., inlining and default arguments) but that was it. Then I ran the benchmarks with and without these changes. Here's the summary of what the benchmark analyzer tool output ("First score" is with these changes):

```
Performance summary
=================
Regressions: Maximum = 14.96%, Geometric mean = 1.16%
Improvements: Maximum = -19.76%, Geometric mean = -1.92%

Performance details
=================
Benchmark                                         First score              Second score             Percent                  Ratio                    
------------------------------------------------------------------------------------------------------------------------------------------------------
Geometric mean                                    34.7117 ± 0.0461         35.5231 ± 0.0501         -2.02 ± 0.27 %           0.98 ± 0.00              
```

These were only run on JDK, version 1.8.0_202. Didn't measure on Android. Hardware specs:

- CPU: 2x Intel Xeon Gold 6154 / 2x 18-cores (36 cores total) @ 3.0GHz
- Memory: 192GB ECC DDR4 RAM @ 2600 MHz - 6-channel
- Disk: 1024GB PCIe NVMe M.2 disk

max-kammerer:
@punzki There are some thoughts:

1. In most benchmarks there is only one 'for' layer, if additional instructions are added it's worth to also tests nested for loops (with different (and same) pairs ( or tripples for 3 'for' layers) of optimizations)
2. It's worth to accumulate also statistics for loop related benchmarks separately of another ones
3. JVM results could be affected by #2279 
4. It's worth to check also JVM 6 (seems should be easy to make ) and Android targets(could be postponed). Please note that there was some Android related benchmarks that was made by @denisvnukov on Google side (he was contributing to Kotlin some time (1-2 year) ago).
Please also look at https://github.com/JetBrains/kotlin-benchmarks (seems this is an origin of K/N benchmakrs and they are using more suited for JVM JMH infrastructure)

punzki:
> @punzki There are some thoughts:
> 
> 1. In most benchmarks there is only one 'for' layer, if additional instructions are added it's worth to also tests nested for loops (with different (and same) pairs ( or tripples for 3 'for' layers) of optimizations)
> 2. It's worth to accumulate also statistics for loop related benchmarks separately of another ones

I'll add some more benchmarks. I'm currently setting up a separate machine for running benchmarks. Thus far I've been running them on the same machine I use for development so I am not able to write code while running benchmarks :slightly_frowning_face: (also the results may be affected).

> 3. JVM results could be affected by #2279

I don't think it would be affected unless loops were being created inside the benchmark methods (and I don't think they are).

> 4. It's worth to check also JVM 6 (seems should be easy to make ) and Android targets(could be postponed). Please note that there was some Android related benchmarks that was made by @denisvnukov on Google side (he was contributing to Kotlin some time (1-2 year) ago).
>    Please also look at https://github.com/JetBrains/kotlin-benchmarks (seems this is an origin of K/N benchmakrs and they are using more suited for JVM JMH infrastructure)

I'll check on JVM 6. I'll also look into kotlin-benchmarks but I couldn't figure out how to make it use a compiler built from local sources (I am not much of a Maven or Gradle expert).

max-kammerer:
@punzki You can locally install Kotlin plugin by : ```./gradlew install publish```
and then change version in maven pom in kotlin-benchmarks to 
```<kotlin.version>1.3.21</kotlin.version> -> <kotlin.version>1.3-SNAPSHOT</kotlin.version>```
It's best use kotlin-benchmarks cause I don't sure that K/N benchmarks on JVM do proper work

punzki:
@max-kammerer - Thanks for the tip, that was very useful! I added more benchmarks to kotlin-benchmarks (PR here: https://github.com/JetBrains/kotlin-benchmarks/pull/5).

I ran those new benchmarks in kotlin-benchmarks (using JMH), as well as the existing [benchmarks in `LoopBenchmark.kt`](https://github.com/JetBrains/kotlin-benchmarks/blob/master/src/main/kotlin/org/jetbrains/LoopBenchmark.kt). I compared the results with and without these changes, on both JDK 1.8 and JDK 1.6.

**There are no significant differences in performance** when using `if+do-while` vs `while` loops for looping over progressions. If anything, there is a slight improvement in some of the `Long` benchmarks, but I don't have a good explanation.

Benchmark options:
- 2 forks
- 15 warmup iterations
- 10 measurement iterations -- Therefore with 2 forks, there were 20 measurements per benchmark.

New hardware specs (as I mentioned before, I set up a dedicated machine to run these, virtual on GCE):
- CPU: Intel(R) Xeon(R) CPU @ 2.20GHz (4 cores)
- Memory: 30GB of RAM
- Disk: 500GB

punzki:
Actual benchmark results:
- Using `while`, JDK 1.8: [loops-ir-while-1.8.txt](https://github.com/JetBrains/kotlin/files/3143476/loops-ir-while-1.8.txt)
- Using `if+do-while` (i.e., with these changes), JDK 1.8: [loops-ir-do-while-1.8.txt](https://github.com/JetBrains/kotlin/files/3143474/loops-ir-do-while-1.8.txt)

- Using `while`, JDK 1.6: [loops-ir-while-1.6.txt](https://github.com/JetBrains/kotlin/files/3143475/loops-ir-while-1.6.txt)
- Using `if+do-while` (i.e., with these changes), JDK 1.6: [loops-ir-do-while-1.6.txt](https://github.com/JetBrains/kotlin/files/3143473/loops-ir-do-while-1.6.txt)

Where there are significant differences that show up, the error/stddev is also significant.

max-kammerer:
@punzki Is "Using while, JDK XX" corresponds to current JVM backend bytecode generation?

punzki:
> @punzki Is "Using while, JDK XX" correspond to current JVM backend bytecode generation?

These are all using the IR backend. These are comparing the changes in `ForLoopsLowering` in this PR. I figured that's what we're are trying to ensure (no significant difference between while and do-while).

I will run the benchmarks on non-IR JVM backend and post an update, but I expect it to be faster because it uses `IINC` instructions while IR does not (no int increment intrinsic).

punzki:
Results for non-IR JVM backend: [loops-non-ir-1.8.txt](https://github.com/JetBrains/kotlin/files/3149410/loops-non-ir-1.8.txt)

I see a noticeable difference in `intRangeExpressionLoop`, `longRangeExpressionLoop`, and `arrayIndexLoop` where non-IR backend is faster. Must be something about those cases that JIT works better with non-IR (or more likely, worse for IR) but I don't have a better explanation.

punzki:
@max-kammerer Just wondering if you had any other concerns about this PR specifically, separate from the discussion about `until` and `step` progressions. We could also just have one form of the loop over progression, since the only difference between "can overflow" and "cannot overflow" is the do-while condition (`while (loopVar != last)` vs `while (inductionVar <= last)`). I don't see any practical difference between the 2 forms.

punzki:
@max-kammerer This PR is superseded by #2390, which also inverts the loop in addition to changing the handling of `until` progressions. Could you please take a look at that one instead?

max-kammerer:
@punzki Ok

max-kammerer:
Closed cause superseded #2390 was merged

PR #2306, opened 2019-04-29
JVM_IR: remove another redundant lowering

pyos:
Value-returning function calls used in a Unit context are handled by https://github.com/JetBrains/kotlin/blob/865e0cc08023e8a813d20805a6ea82e4f760e79a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt#L395

(This lowering was third-slowest after BridgeLowering and ForLoopsLowering.)


max-kammerer:
Thanks!

PR #2308, opened 2019-04-30
do not use constructor injection for app level extension

develar:
Otherwise IDEA 192 logs warning about it.


yanex:
Thanks! Merged manually (master and 1.3.40-backup branches).

PR #2309, opened 2019-04-30
JVM_IR: attach delegate metadata to properties

pyos:


udalov:
@pyos I had a change that fixed this in another way but it got lost somehow. I've pushed it in 6cc293a0c5. Thank you and I hope you didn't spend too much time on these tests.

PR #2310, opened 2019-04-30
IR parameter fix

sfs:
There are two ways of accessing `this` in the IR:
- In a member function we use the `dispatchReceiverParameter`
- In a constructor we use the `thisReceiver` of the constructed class

Right now there are several passes which produce IR which accesses the `thisReceiver` in member functions, necessitating a workaround in ExpressionCodegen. This patch fixes the passes in question and removes the workaround.

There's also an unrelated fix to the parameter indices in SyntheticAccessorLowering. Currently this pass produces static functions with parameter indices < 0. I didn't want to open a separate PR for that.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
@@ -8,12 +8,14 @@ package org.jetbrains.kotlin.backend.common.lower
 import org.jetbrains.kotlin.backend.common.BackendContext
 import org.jetbrains.kotlin.backend.common.BodyLoweringPass
 import org.jetbrains.kotlin.backend.common.ClassLoweringPass
+import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.builders.irGet
 import org.jetbrains.kotlin.ir.builders.irSetField
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrConstructor
+import org.jetbrains.kotlin.ir.declarations.IrFunction
 import org.jetbrains.kotlin.ir.declarations.IrValueParameter
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.*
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
```diff
@@ -91,6 +95,16 @@ class InnerClassesLowering(val context: BackendContext) : ClassLoweringPass {
                     enclosingConstructor = null
                 }
 
+            override fun visitFunction(declaration: IrFunction): IrStatement =
+                try {
```
gogabr: This way of handling context looks ad-hoc and brittle to me. Why not use `IrElementTransformerVoidWithContext`?

sfs: Good question! I changed it to use `IrElementTransformerVoidWithContext`. This doesn't change the fact that we don't deal with nested functions, but it does make the code more compact.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
```diff
@@ -28,6 +30,7 @@ import org.jetbrains.kotlin.ir.util.transformDeclarationsFlat
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 
+// This pass has to run after LocalDeclarationsLowering, since we don't handle nested functions.
```
gogabr: Now, instead of documenting the dependency in a comment, you could specify it in the `prerequisite` field of the phase.

sfs: The problem is that this particular lowering is run with different phases in the different backends (localDeclarationsLoweringPhase in JS, jvmLocalDeclarationsPhase in JVM, etc.), so I cannot point at a specific phase.

gogabr: I see. We will need to unify those phases at some point.

gogabr:
Thank you!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
udalov: But this variable (`innerClassesPhase`) is used only in JVM IR, could we move it to `JvmLower.kt` for example and specify the requirement on `jvmLocalDeclarationsPhase` there?

(Maybe we need to actually move all other JVM phase declarations from backend.common to backend.jvm exactly to be able to specify JVM-specific phase dependencies.)

sfs: That's a good point, I didn't notice that the phase wasn't shared. I've added the prerequisite as part of #2342.

PR #2311, opened 2019-04-30
Fix type of super

sfs:
In the IR there is no real distinction between `super` and `this`. Instead, this distinction is made at the call site where some calls are marked as super calls. This is in contrast to the representation of super in psi, which has a separate representation.

Currently, when we translate from psi to IR, some of this distinction shines through in that this code
```
class A { fun f() = super.hashCode() }
```
gets turned into IR accessing the dispatch receiver of `f` *with type Any instead of type A*.

For technical reasons (invokespecial) this is unfortunate for the JVM backend, in particular in combination with inline classes. I propose that we change this to treat `super` and `this` interchangeably in the IR instead.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/Clone.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/Clone.kt
@@ -16,21 +16,19 @@
 
 package org.jetbrains.kotlin.backend.jvm.intrinsics
 
-import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.backend.jvm.codegen.*
+import org.jetbrains.kotlin.codegen.AsmUtil
 import org.jetbrains.kotlin.ir.expressions.IrCall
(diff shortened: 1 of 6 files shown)


dnpetrov:
@sfs I see some minor changes in testData, but could you please add an irText test case that demonstrates the motivation for your changes? 
Probably just an extra case in irText/classes/qualifiedSuperCalls.txt and irText/classes/superCalls.txt
That's a minor thing, though. Overall, LGTM.

sfs:
@dnpetrov I've added the case mentioned in the PR, but in fact you can already see the problem in the current irText tests. E.g., we have

```
GET_VAR '<this>: <root>.Derived declared in <root>.Derived.foo' type=<root>.Base origin=null
```

in the current tests, where we read a variable of type Derived as an expression of type Base.

max-kammerer:
@sfs https://teamcity.jetbrains.com/comparisonFailureDiffView.html?buildId=2282272&testId=101830 
after rebase on master

sfs:
@max-kammerer Fixed

dnpetrov:
Looks good to me.

max-kammerer:
Thanks!

PR #2312, opened 2019-05-01
KT-5008 Quick fix to remove empty class initializers/constructors

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-5008


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantEmptyClassInitializerInspection.kt
```diff
@@ -0,0 +1,43 @@
...
+class RedundantEmptyClassInitializerInspection : AbstractKotlinInspection() {
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean) = classInitializerVisitor(fun(initializer) {
+        if (initializer.getStrictParentOfType<KtClassOrObject>() == null) return
+        val body = initializer.body as? KtBlockExpression ?: return
+        val leftBrace = body.lBrace ?: return
+        val rightBrace = body.rBrace ?: return
+        if (leftBrace.getNextSiblingIgnoringWhitespace() != rightBrace) return 
+        holder.registerProblem(
+            initializer.initKeyword,
+            "Redundant empty class initializer",
```
NataliaUkhorskaya: We call them **initializer blocks** (https://kotlinlang.org/docs/reference/classes.html#constructors)

idea/testData/inspectionsLocal/redundantEmptyClassInitializer/notEmpty2.kt
```diff
@@ -0,0 +1,6 @@
+// PROBLEM: none
+class Test {
+    <caret>init {
+        // comment
```
NataliaUkhorskaya: Seems we can report the warning in this case because block is still redundant (comment can be moved to class body)

t-kameyama:
b650c7ab00a4480436bec64e4d8404263aaf3bc3

PR #2314, opened 2019-05-02
Add map() documentation sample

borsini:
Small code sample for the collections.map() function

### DIFF
--- a/libraries/stdlib/common/src/generated/_Maps.kt
+++ b/libraries/stdlib/common/src/generated/_Maps.kt
@@ -59,7 +59,7 @@ public inline fun <K, V, R, C : MutableCollection<in R>> Map<out K, V>.flatMapTo
  * Returns a list containing the results of applying the given [transform] function
  * to each entry in the original map.
  * 
- * @sample samples.collections.Maps.Transformations.mapMap
+ * @sample samples.collections.Maps.Transformations.mapToList
  */
 public inline fun <K, V, R> Map<out K, V>.map(transform: (Map.Entry<K, V>) -> R): List<R> {
     return mapTo(ArrayList<R>(size), transform)
--- a/libraries/stdlib/common/src/generated/_Maps.kt
+++ b/libraries/stdlib/common/src/generated/_Maps.kt
@@ -59,7 +59,7 @@ public inline fun <K, V, R, C : MutableCollection<in R>> Map<out K, V>.flatMapTo
  * Returns a list containing the results of applying the given [transform] function
  * to each entry in the original map.
  * 
- * @sample samples.collections.Collections.Transformations.map
+ * @sample samples.collections.Maps.Transformations.mapMap
  */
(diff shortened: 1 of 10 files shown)


ilya-g:
It's worth to provide different samples for `Map.map` and `CharSequence/String.map`, because their usages are slightly different.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -393,6 +393,12 @@ class Collections {
             assertPrints(chars.joinToString(limit = 5, truncated = "...!") { it.toUpperCase().toString() }, "A, B, C, D, E, ...!")
         }
 
+        @Sample
+        fun map() {
+            val numbers = listOf(75, 111, 116, 108, 105, 110)
+            assertPrints(numbers.map { it.toChar() }, "[K, o, t, l, i, n]")
```
ilya-g: I'd consider using some more simple operation, for example squaring numbers or taking `length` property of string elements.

borsini: Sure I will do that 👍

borsini:
As suggested, I added samples for Map and CharSequence. I also simplified the first example.

borsini:
@ilya-g do the changes look good to you ?

libraries/stdlib/samples/test/samples/collections/maps.kt
```diff
@@ -303,6 +302,12 @@ class Maps {
             assertPrints(props.getProperty("z", "fail"), "fail")
         }
 
+        @Sample
+        fun mapMap() {
+            val peopleToAge = mapOf("alice" to 20, "bob" to 21)
+            assertPrints(peopleToAge.map { it.value }, "[20, 21]")
```
ilya-g: I propose to add another example with entry argument destructuring:
`peopleToAge.map { (name, age) -> "$name is $age years old" }`

libraries/stdlib/common/src/generated/_Maps.kt
```diff
@@ -58,6 +58,8 @@ public inline fun <K, V, R, C : MutableCollection<in R>> Map<out K, V>.flatMapTo
 /**
  * Returns a list containing the results of applying the given [transform] function
  * to each entry in the original map.
+ * 
+ * @sample samples.collections.Maps.Transformations.mapMap
```
ilya-g: to many `map`s in the sample qualified name :)
`mapToList` would be better.

libraries/stdlib/samples/test/samples/collections/maps.kt
borsini: Indeed, I placed it first as it is the obvious way to use it.

libraries/stdlib/common/src/generated/_Maps.kt
borsini: Done

ilya-g:
Thanks, I've merged it.

PR #2315, opened 2019-05-04
Change order of line number geneartion for function call, and also force  line number generation after visiting inline functions

neetopia:
Currently, line number for a function call is only generated for the line when start entering function call, before visiting arguments, therefore missing line number after all function arguments are visited.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -268,7 +268,6 @@ class ExpressionCodegen(
         visitStatementContainer(expression, data).coerce(expression.asmType)
 
     override fun visitFunctionAccess(expression: IrFunctionAccessExpression, data: BlockInfo): PromisedValue {
-        expression.markLineNumber(startOffset = true)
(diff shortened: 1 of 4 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -366,6 +369,11 @@ class ExpressionCodegen(
             }
         }
 
+        //Reset line number if this call has non-const argument
```
madsager: This doesn't seem like the right fix to me. It seems to me that we should always mark the line number here for the call instead of marking it in the callers so that we get the line number for the call after evaluating the parameters. The other issue for these tests seems to be that we do not get the last line number emitted by the inline codegen correctly reflected here and therefore, we thing that the last line number emitted is still the line number of the call even though it isn't.

madsager: I think if we move the line number generation for a call to after the evaluation of the arguments, the other thing we have to do is implement ExpressionCodegen.markLineNumberAfterInlineIfNeeded.

max-kammerer:
@neetopia please rebase on master

PR #2316, opened 2019-05-06
JVM_IR: run lowerings on entire module fragments

pyos:
Otherwise, the order of lowerings does not actually guarantee anything, as lowering A that is before B may see B's effects through inter-file references (or it may not, if it got lucky with the ordering of files). Lowering the entire module, by contrast, ensures that A will *never* see the effects of B (although B might still see IR not lowered by A if it was reconstructed from another module's reflection metadata.)


max-kammerer:
@udalov @gogabr Is this change invariant for any incremental compilation problem?

udalov:
I have an absurd feeling that we need to postpone this change as much as possible, and then only push it if we have no other options. The reason is because solving the highlighted problem in another way, which is to rewrite lowerings in a way that they _don't rely_ on the IR being either lowered or unlowered, will probably help us in the long-term. For example, if we adapt lowerings this way, we can even maybe avoid lowering IR loaded from dependencies altogether, and further guarantee that backend will work exactly the same in a partial compilation scenario vs in the full build. Also, there's a possibility (only theoretical at the moment) to run lowerings on different files in parallel in the future.

@pyos Is this change still required for the updated #2318? I feel like the fact that we've caught a problematic (in terms of partial compilation on JVM) implementation of EnumWhenLowering is telling and it's actually good to have this weird behavior right now.

pyos:
@udalov it is still required because in the lowered IR enum accesses look like arbitrary static field reads. This could be worked around by carefully looking at the receivers `IrGetField`s, of course, but I feel like requiring each lowering to be capable of handling *three* different states (unlowered, lowered by passes before it, and lowered by all passes) is a bit too much.

>we can even maybe avoid lowering IR loaded from dependencies altogether

I'm not sure what you mean by that, since IR stubs reconstructed from class files aren't lowered now either, and this case is the exact opposite of having to deal with the fully lowered IR.

>Also, there's a possibility (only theoretical at the moment) to run lowerings on different files in parallel in the future.

I believe this will result in data races unless you completely separate the files from each other. In fact, I had considered copying each `IrFile` before lowering and patching references in other `IrFile`s to point to the unlowered version as an alternative to this PR, but decided to use a less complex approach, at least for now. (Note that isolating `IrFile`s still only requires two modes from each lowering, "unlowered" and "lowered up to this point", not three.)

pyos:
>I admit that I haven't bothered checking the reason they were failing before. I'll have a look at these later as well.

These tests all contain calls to inline functions from other files in the same module. In this case, since the class files do not exist yet, the inliner attempts to generate bytecode from unlowered IR. Or it would, had these functions not been defined at top-level scope, breaking the very first assumption the inliner makes (that the parent of any JVM function is a class).

Combined with the fact that lowering IR file-by-file produces outright incorrect IR (the previous version of EnumWhenLowering did not work because it encountered a reference to a declaration that is not a child of its parent), I don't really see any alternative to this change.

udalov:
> > we can even maybe avoid lowering IR loaded from dependencies altogether
> 
> Do you mean serializing lowered IR for inline functions? That might result in even more cruft, as you'd be essentially declaring the effects of each lowering to be part of ABI.

No, by "IR loaded from dependencies" I mean lazy stub IR which is constructed from deserialized descriptors loaded from metadata. In a partial compilation scenario, this IR loaded from a compiled file from the same module is basically indistinguishable from IR loaded from a compiled file from a library. Therefore, if we managed to succeed in making all lowerings work regardless of whether the IR from other files is lowered or not, we'd avoid the need to lower this lazy stub IR at all. (Right now we also don't lower lazy stub IR, but that isn't fully justified. We were actually discussing the fact that it's helpful to perform structural lowerings on IR from dependencies, to simplify code of certain lowerings. But it would be nice if this wasn't needed.)

> I believe this will result in data races unless you completely separate the files from each other.

You're right, this is why it's theoretical at the moment. The copying approach could work though.

> These tests all contain calls to inline functions from other files in the same module. In this case, since the class files do not exist yet, the inliner attempts to generate bytecode from unlowered IR. Or it would, had these functions not been defined at top-level scope, breaking the very first assumption the inliner makes (that the parent of any JVM function is a class).

This actually is fixable in a simpler way. The problem is that the bytecode inliner relies on the fact that all source files are lowered. We can fix exactly that by lowering all files before invoking codegen on them: split `JvmBackend.generateFile` into two functions, one which lowers the file and another which invokes the codegen, and then lower all files in `JvmBackendFacade.doGenerateFilesInternal` first, and afterwards invoke codegen on all lowered files.

> Combined with the fact that lowering IR file-by-file produces outright incorrect IR (the previous version of EnumWhenLowering did not work because it encountered a reference to a declaration that is not a child of its parent), I don't really see any alternative to this change.

Right. I'm arguing that this incorrect IR is a sort of a safeguard for us here. As soon as any lowering depends on the fact that the other file is lowered up to a certain point, you have to think very carefully whether this won't backfire in the partial compilation scenario, where this other file is loaded from dependencies. If there are no dependencies of such kind, it's possible to observe incorrect IR, but the right course of action there would be to modify the lowering for it to avoid depending on the exact format of IR in other files. Of course, this won't prevent all cases where we could illegitimately depend on content of other files, but I think it's still useful (it helped with the EnumWhenLowering).

---

Also note that in JS backend, lowering are performed per-module right now, but there's a plan (and it's work in progress) to convert them to per-file instead, specifically to support incremental compilation.

udalov:
I've pushed e72388895b7 where I've fixed the problem with inline mentioned above and unmuted the same tests.

I'll close this pull request for now, since we're not ready to depend on this behavior yet. We might come back to it at some point in the future though. Thank you!

PR #2317, opened 2019-05-06
JVM_IR: wrap performInline in enterIntoInlining/exitFromInliningOf

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -93,7 +93,12 @@ class IrInlineCodegen(
         expression: IrFunctionAccessExpression
     ) {
         val typeArguments = expression.descriptor.typeParameters.keysToMap { expression.getTypeArgumentOrDefault(it) }
-        performInline(typeArguments, callDefault, codegen)
+        state.globalInlineContext.enterIntoInlining(null)
```
max-kammerer: Please add TODO, that  IrFunctionAccessExpression should be somehow passed for cycle check

PR #2318, opened 2019-05-06
JVM_IR: implement the when(enum) optimization

pyos:
The base EnumWhenLowering is mostly copied from Kotlin/Native; the only functional difference is that a nullable subject is converted into ordinal -1 when null rather than never being optimized at all, and the null checks are transformed accordingly. K/N tests pass when using this version of the lowering, so its copy can be removed after the next revision sync. On top of that, a JVM-specific subclass of EnumWhenLowering remaps the ordinals using WhenMappings.


gogabr:
Thank you!

PR #2319, opened 2019-05-06
Remove generated sources and classes for full KAPT build

gavra0:
https://youtrack.jetbrains.com/issue/KT-31322

When running incremental annotation processing in KAPT,
even if incremental flag is enabled, and full rebuild should
be performed (e.g. in presence of dynamic non-incremental APs),
make sure generated sources and classes are removed.

 #KT-31322 fixed


yanex:
Thank you for the contribution! Merged manually (master and 1.3.40-backup branches).

PR #2320, opened 2019-05-07
Implement special cases for the hashCode intrinsic in JVM_IR

sfs:
In the jvm backend the behavior of Object.hashCode applied to primitive types depends on the context. 

Inside of the auto generated hashCode method of data or inline classes we inline the implementation of hashCode for primitive types. Otherwise we either call Object.hashCode or the static hashCode methods for primitive types (Integer.hashCode, etc.) on targets where they exist.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
@@ -16,24 +16,39 @@
 
 package org.jetbrains.kotlin.backend.jvm.intrinsics
 
-import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.backend.jvm.codegen.*
 import org.jetbrains.kotlin.codegen.AsmUtil
-import org.jetbrains.kotlin.codegen.Callable
-import org.jetbrains.kotlin.codegen.CallableMethod
-import org.jetbrains.kotlin.ir.expressions.IrCall
+import org.jetbrains.kotlin.config.JvmTarget
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.expressions.IrFunctionAccessExpression
-import org.jetbrains.kotlin.ir.expressions.IrMemberAccessExpression
-import org.jetbrains.kotlin.resolve.jvm.AsmTypes
-import org.jetbrains.kotlin.resolve.jvm.jvmSignature.JvmMethodSignature
+import org.jetbrains.kotlin.ir.util.render
 import org.jetbrains.org.objectweb.asm.Opcodes
 import org.jetbrains.org.objectweb.asm.Type
-import org.jetbrains.org.objectweb.asm.commons.InstructionAdapter
 
+// TODO Implement hashCode on primitive types as a lowering.
 object HashCode : IntrinsicMethod() {
-
-    override fun toCallable(expression: IrFunctionAccessExpression, signature: JvmMethodSignature, context: JvmBackendContext): IrIntrinsicFunction {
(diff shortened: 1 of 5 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
```diff
@@ -16,24 +16,38 @@
...
+        val target = context.state.target
+        when {
+            irFunction.origin == IrDeclarationOrigin.GENERATED_INLINE_CLASS_MEMBER || irFunction.origin == IrDeclarationOrigin.GENERATED_DATA_CLASS_MEMBER ->
+                AsmUtil.genHashCode(mv, mv, result.type, target)
+            target == JvmTarget.JVM_1_6 -> {
+                result.coerce(AsmUtil.boxType(result.type)).materialize()
+                mv.visitMethodInsn(Opcodes.INVOKEVIRTUAL, "java/lang/Object", "hashCode", "()I", false)
+            }
+            else -> {
+                val boxedType = AsmUtil.boxType(result.type)
```
max-kammerer: @sfs Please add box test with 'hashCode' invoked on interface instance, please test would it work with // JVM_TARGET: 1.8.
Result test shouldn't contain JVM_TARGET (it would be tested in " Test Codegen on different platforms (JDKs/Android)" build configuration)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
```diff
@@ -16,24 +16,38 @@
...
-
-    override fun toCallable(expression: IrFunctionAccessExpression, signature: JvmMethodSignature, context: JvmBackendContext): IrIntrinsicFunction {
-        return IrIntrinsicFunction.create(expression, signature, context, AsmTypes.OBJECT_TYPE) {
-            it.visitMethodInsn(Opcodes.INVOKEVIRTUAL, "java/lang/Object", "hashCode", "()I", false)
+    override fun invoke(expression: IrFunctionAccessExpression, codegen: ExpressionCodegen, data: BlockInfo) = with(codegen) {
+        val receiver = expression.dispatchReceiver ?: error("No receiver for hashCode: ${expression.render()}")
+        val result = receiver.accept(this, data).materialized
+        val target = context.state.target
+        when {
+            irFunction.origin == IrDeclarationOrigin.GENERATED_INLINE_CLASS_MEMBER || irFunction.origin == IrDeclarationOrigin.GENERATED_DATA_CLASS_MEMBER ->
```
max-kammerer: Seems it's best to add separate lowering for this, please add //TODO...

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
sfs: @max-kammerer can you elaborate? The HashCode intrinsic is only ever used when hashCode is called on primitive types. If you call hashCode on an interface type it will go through the normal generateCall machinery in ExpressionCodegen.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
sfs: Definitely. I've added the TODO.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/HashCode.kt
max-kammerer: @sfs Oh, I missed it

max-kammerer:
Thanks!

PR #2321, opened 2019-05-08
Fix compareCustom function in IrLineNumberTest class, to exclude comment

neetopia:
Fix compareCustom function in IrLineNumberTest class, to exclude comment lines that are not related to line number verification.

Unmute a test case should have been unmuted but shadowed by this bug.

### DIFF
--- a/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractIrLineNumberTest.kt
+++ b/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractIrLineNumberTest.kt
@@ -25,7 +25,7 @@ abstract class AbstractIrLineNumberTest : AbstractLineNumberTest() {
     override fun compareCustom(psiFile: KtFile, wholeFile: File) {
         val fileText = psiFile.text
         val expectedLineNumbers = normalize(
-            fileText.substring(fileText.indexOf("//") + 2)
+            fileText.substring(Regex("// \\d+").find(fileText)!!.range.start + 2)
                 .trim().split(" ").map { it.trim() }.toMutableList()
         )
         val actualLineNumbers = normalize(extractActualLineNumbersFromBytecode(classFileFactory, false))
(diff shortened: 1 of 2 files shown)


compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractIrLineNumberTest.kt
```diff
@@ -24,10 +24,13 @@ abstract class AbstractIrLineNumberTest : AbstractLineNumberTest() {
 
     override fun compareCustom(psiFile: KtFile, wholeFile: File) {
         val fileText = psiFile.text
-        val expectedLineNumbers = normalize(
-            fileText.substring(fileText.indexOf("//") + 2)
-                .trim().split(" ").map { it.trim() }.toMutableList()
-        )
+        val expectedLineNumbers = fileText.split("\n".toRegex()).filter { line ->
```
madsager: How about simplifying this to:

        val expectedLineNumbers = normalize(
            fileText.substring(Regex("// \\d+").find(fileText)!!.range.start + 2)
                .trim().split(" ").map { it.trim() }.toMutableList()
        )

Then we are looking for a line that starts with "// " with a number of digits after it. That should work to exclude the comment lines with text. That looks consistent with what the old backend expects from these tests.

neetopia: Looks reasonable since this is how old test is doing. I was a little bit overthinking this, was considering the case where there is a trailing comment line after line numbers

neetopia: I am also considering adding an optional "+" check before digit check, since there are line numbers start with a "+", although such case is not present in current test data yet.

max-kammerer: It wouldn't presented in future (answered in slack)

max-kammerer:
Thanks!

PR #2322, opened 2019-05-08
KT-10215 Kotlin classes are listed after Java classes in the navigation bar

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-10215


goodwinnk:
Thank you! I have merged this PR manually (https://github.com/JetBrains/kotlin/commit/0a4d21c51ffddb0b008b32ece26641eb8f769543) with the another commit (https://github.com/JetBrains/kotlin/commit/ef70e9db1921f3870bfc0c3f493f108ceb92dfb9) that removes using light classes. Light classes are meant to be used for interop with Java and if possible shouldn't be used for implementing general purpose features.

PR #2324, opened 2019-05-09
JVM: Generate LVT entry for this in @JvmOverload methods.

madsager:
Since we generate local variable entries for the rest of the arguments
we should do it for the receiver as well.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/DefaultParameterValueSubstitutor.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/DefaultParameterValueSubstitutor.kt
@@ -244,8 +244,10 @@ class DefaultParameterValueSubstitutor(val state: GenerationState) {
         val methodEnd = Label()
         mv.visitLabel(methodEnd)
 
+        val thisType = functionDescriptor.dispatchReceiverParameter?.type?.asmType(typeMapper)
+
         FunctionCodegen.generateLocalVariablesForParameters(
-            mv, signature, functionDescriptor, null, methodBegin, methodEnd, remainingParameters, isStatic, state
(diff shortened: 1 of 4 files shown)


madsager:
@udalov @max-kammerer what do you think? Matching the current behavior in the JVM_IR backend seemed inconsistent so I would rather make this change. Another option would be to not generate a local variable table for these methods at all. There are no line numbers in the methods anyway so I don't think you can inspect or change these locals from the debugger anyway?

max-kammerer:
I would prefer to not generate LVT at all
@udalov Any objections?

madsager:
@udalov @max-kammerer based on my understanding it also seems silly to me to generate the LCT for these. Updated this change to avoid the generation.

max-kammerer:
NB: it will change debugger experience a bit, cause these variables wouldn't available in stack frames

madsager:
That is true. Since the method is not synthetic and shouldn't be this will not be filtered or greyed out from stack traces and being able to inspect |this| and all the passed variables here might be nice (even though you can do that if you just select the frame for the actual target method).

I'm still leaning towards not generating the table. Let me know which one you prefer. Both changes are simple. :)

max-kammerer:
@goodwinnk @yanex Any thoughts from debugger perspective?

udalov:
We need to generate LVT in `@JvmOverloads` method adapters in order for Java compiler and analysis in IDE to see the parameter names and render them nicely in things like completion, error messages, etc. See [KT-7319](https://youtrack.jetbrains.com/issue/KT-7319)

madsager:
Thanks @udalov that does make sense. Went back to generating the full local variable table (including |this|) for @JvmOverload methods.

max-kammerer:
Thanks!

PR #2313, opened 2019-05-01
Split up generation of test data for codegen/box/ranges into signed and unsigned.

punzki:
The tests pass in JVM_IR for signed ranges/progressions, but fail for unsigned due to inlining being broken. We can disable the JVM_IR tests for unsigned, while keeping them enabled for signed, to get better test coverage in the interim until inlining is fixed.

PR #2326, opened 2019-05-10
Implement inline classes in the JVM_IR backend

sfs:
This PR implements inline classes in the JVM_IR backend in a way that's compatible with the jvm backend. The implementation is split between a lowering pass (JvmInlineClassLowering), which is analogous to the existing InlineClassDeclarationLowering passes in JS_IR and native, and additional coercions implemented in ExpressionCodegen.

Unfortunately, there is not much code that can be shared between backends at the moment. Inline classes on the JVM are substantially different in order to interoperate with Java (name mangling rules, replacement functions have to be static or not depending on the situation, etc.).

The second part of the implementation in ExpressionCodegen is basically the only way to produce the same bytecode as the jvm backend, since it has to run concurrently with bytecode inlining.

The implementation is almost complete, apart from the integration with bytecode inlining for external functions (the inliner gets the bytecode for external declarations based on their descriptors and does not know what to do with WrappedDescriptors) and some issues in bridge lowering.

Both of these issues are fairly self-contained and will be handled in different PRs.

The first 5 commits are from #2310 and #2311.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -360,19 +393,22 @@ class ExpressionCodegen(
...
-            expression is IrDelegatingConstructorCall -> voidValue
+            expression is IrConstructorCall ->
+                MaterialValue(this, asmType, expression.type)
+            expression is IrDelegatingConstructorCall ->
+                voidValue
             expression.type.isUnit() ->
                 // NewInference allows casting `() -> T` to `() -> Unit`. A CHECKCAST here will fail.
-                MaterialValue(mv, callable.returnType).discard().coerce(expression.asmType)
-            else -> MaterialValue(mv, callable.returnType).coerce(expression.asmType)
+                MaterialValue(this, callable.returnType).discard().coerce(expression.type)
```
pyos: If IrType is nullable because of void, it would probably be best to be consistent and specify it whenever the asm type is non-void, even if it is unused.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -265,23 +266,47 @@ class ExpressionCodegen(
...
+            ?.invoke(expression, this, data)?.let { return it.coerce(expression.type) }
+
+        if (expression.isErasedNewCall)
+            return generateErasedNewCall(expression, data)
+
+        if (expression.symbol.owner.isInlineClassFieldGetter) {
+            val argument = expression.dispatchReceiver!!
+            val type = expression.symbol.owner.dispatchReceiverParameter!!.type
+            val value = argument.accept(this, data).coerce(type).materialized
+            value.irType = expression.type
```
pyos: Doesn't `expression.type` map to the argument's type? Would `.coerce(expression.type)` do something different?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -599,13 +640,21 @@ class ExpressionCodegen(
...
+            mv.newarray(typeMapper.boxType(elementIrType))
         } else {
-            val type = typeMapper.mapType(arrayType)
-            mv.newarray(correctElementType(type))
+            mv.newarray(correctElementType(arrayType.asmType))
         }
     }
 
+    // Copied from FinallyBlocksLowering
+    private val IrReturnTarget.returnType: IrType
```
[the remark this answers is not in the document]
sfs: The problem with this is that `returnType` is specific to the JVM backend in the case of constructors.

Changing the IR by adding new fields or methods also needs input from JS_IR and native, so whenever possible I'd rather avoid it and just add extension methods. Otherwise I agree that this should be a property.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -687,30 +735,32 @@ class ExpressionCodegen(
...
             IrTypeOperator.IMPLICIT_CAST, IrTypeOperator.IMPLICIT_INTEGER_COERCION -> {
-                expression.argument.accept(this, data).coerce(expression.asmType)
+                expression.argument.accept(this, data).coerce(expression.type)
             }
 
             IrTypeOperator.CAST, IrTypeOperator.SAFE_CAST -> {
-                expression.argument.accept(this, data).coerce(AsmTypes.OBJECT_TYPE).materialize()
-                val boxedType = boxType(asmType)
+                val result = expression.argument.accept(this, data)
+                val boxedLeftType = result.irType?.let { typeMapper.boxType(it) } ?: OBJECT_TYPE
```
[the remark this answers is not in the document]
sfs: In principle every PromisedValue except for void should have an IrType. The problem is that there is no IrType which maps to void.

I'll refactor this to make it clearer.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -55,27 +70,76 @@ fun PromisedValue.discard(): MaterialValue {
...
 }
 
 // On materialization, cast the value to a different type.
-fun PromisedValue.coerce(target: Type) = when (target) {
-    type -> this
-    else -> object : PromisedValue(mv, target) {
-        // TODO remove dependency
-        override fun materialize() = StackValue.coerce(this@coerce.materialized.type, type, mv)
+fun PromisedValue.coerce(target: Type, irTarget: IrType? = null): PromisedValue {
+    val irType = irType
```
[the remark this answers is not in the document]
sfs: It was just there for an implicit cast in the code below. I'm making irType read only, which also allows us to get rid of this.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/JavaClassProperty.kt
```diff
@@ -20,20 +20,29 @@ import org.jetbrains.kotlin.backend.jvm.codegen.*
...
-        if (value.type == Type.VOID_TYPE) {
-            return invokeWith(value.coerce(AsmTypes.UNIT_TYPE))
-        }
-        if (isPrimitive(value.type)) {
-            value.discard()
-            value.mv.getstatic(boxType(value.type).internalName, "TYPE", "Ljava/lang/Class;")
-        } else {
-            value.materialize()
-            value.mv.invokevirtual("java/lang/Object", "getClass", "()Ljava/lang/Class;", false)
+        val irType = value.irType
```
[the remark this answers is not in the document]
sfs: It was for another smart cast. It's redundant when irType is read-only.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,364 @@
...
+                    else -> {
+                        memberDeclaration.accept(this, null)
+                        listOf(memberDeclaration)
+                    }
+                }
+            }
+
+            if (declaration.isInline) {
+                val irConstructor = declaration.constructors.single { it.isPrimary }
+                declaration.declarations.removeIf { it is IrConstructor && it.isPrimary || it is IrFunction && it.isInlineClassFieldGetter }
```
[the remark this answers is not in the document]
sfs: Sure!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
sfs: Yes it would - there is an implicit coercion between irTypes going on here, which does not change the underlying asm types. I changed the code to do this coercion in an intrinsic instead, which should make it clearer.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
sfs: You're right, I'll make it more consistent.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -265,23 +266,47 @@ class ExpressionCodegen(
...
+
+    private val IrFunctionAccessExpression.isErasedNewCall: Boolean
+        get() = symbol.owner.isPrimaryInlineClassConstructor && irFunction.origin != IrDeclarationOrigin.SYNTHETIC_INLINE_CLASS_MEMBER
+
+    private fun generateErasedNewCall(expression: IrFunctionAccessExpression, data: BlockInfo): PromisedValue {
+        val function = expression.symbol.owner
+        val argument = expression.getValueArgument(0) ?: function.valueParameters[0].defaultValue!!
+        val argumentType = function.valueParameters[0].type
+        val value = argument.accept(this, data).coerce(argumentType).materialized
+        value.irType = function.returnType
```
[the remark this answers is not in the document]
sfs: I added it as another intrinsic (UnsafeCoerce) to make it clearer where this happens and what it does.

Admittedly, the logic for coerce is convoluted, it is probably better to refine PromisedValue instead. I'll investigate this further.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -616,13 +618,13 @@ class KotlinTypeMapper @JvmOverloads constructor(
         }
 
         // Special methods for inline classes.
-        if (InlineClassDescriptorResolver.isSynthesizedBoxMethod(descriptor)) {
+        if (!isIrBackend && InlineClassDescriptorResolver.isSynthesizedBoxMethod(descriptor)) {
```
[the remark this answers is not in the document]
sfs: In this case I think it makes sense to have a separate function for the different backends. This makes it clear what the IrTypeMapper has to do in this case.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -383,7 +383,7 @@ class KotlinTypeMapper @JvmOverloads constructor(
         resolvedCall: ResolvedCall<*>? = null
     ): CallableMethod {
         // we generate constructors of inline classes as usual functions
-        if (descriptor is ConstructorDescriptor && kind !== OwnerKind.ERASED_INLINE_CLASS) {
+        if (descriptor is ConstructorDescriptor && (kind !== OwnerKind.ERASED_INLINE_CLASS || isIrBackend)) {
```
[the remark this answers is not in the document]
sfs: I don't think it makes sense to completely separate the two here, since there is only this single special case in which they differ. On the other hand, it definitely makes sense to separate the name mangling rules, since this is much simpler for the IrBackend.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -631,17 +633,17 @@ class KotlinTypeMapper @JvmOverloads constructor(
...
+            if (!isIrBackend && kind === OwnerKind.ERASED_INLINE_CLASS) {
                 newName = JvmAbi.ERASED_INLINE_CONSTRUCTOR_NAME
             } else {
                 return name
             }
         }
 
         val suffix = getInlineClassSignatureManglingSuffix(descriptor)
-        if (suffix != null) {
+        if (!isIrBackend && suffix != null) {
```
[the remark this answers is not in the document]
sfs: Done!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/ScopedValueMappingTransformer.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.kotlin.ir.expressions.IrGetValue
+import org.jetbrains.kotlin.ir.expressions.IrSetVariable
+import org.jetbrains.kotlin.ir.expressions.impl.IrGetValueImpl
+import org.jetbrains.kotlin.ir.expressions.impl.IrSetVariableImpl
+import org.jetbrains.kotlin.ir.symbols.IrSymbol
+import org.jetbrains.kotlin.ir.symbols.IrValueSymbol
+import org.jetbrains.kotlin.ir.symbols.IrVariableSymbol
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+
+open class ScopedValueMappingTransformer() : IrElementTransformerVoid() {
```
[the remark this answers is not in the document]
sfs: Done

sfs:
Did some refactoring:
- Introduced UnsafeCoerce intrinsic
- Made irType in PromisedValue more consistent (it's now always present unless type is void)
- Some small fixes (see comments)
- Simplified TypeMapper for IrBackend
- Rebased to latest master (#2310 got merged)
- Mute two regressed tests (linenumber changes hitting an assert in IrSourceCompilerForInline)

Going forward, I'll try to split the PR to get some of the IR fixes merged and will look into simplifying the coercion logic.

max-kammerer:
@sfs Please note that #2313 affects testdata

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/DataClassMembersGenerator.kt
```diff
@@ -254,13 +256,17 @@ class DataClassMembersGenerator(
                     irLetS(
```
madsager: I think with your comments to irLetS this might not be needed. It looks like irLetS will just return the symbol for the IrGet and therefore I don't think it is needed for this?

pyos: irGet here produces a call to a getter, not a read of a variable, so irLetS is necessary for exactly-once evaluation.

compiler/tests/org/jetbrains/kotlin/generators/tests/GenerateRangesCodegenTestData.java
```diff
@@ -119,7 +119,7 @@ private static String renderTemplate(String template, int number, String element
                 .replace("\n", LineSeparator.getSystemLineSeparator().getSeparatorString());
     }
 
-    private static final List<String> WHITELISTED_FOR_JVM_IR_BACKEND = Collections.singletonList("overflowZeroDownToMaxValue.kt");
```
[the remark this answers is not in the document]
sfs: Inlining mangled external declarations is currently broken.

Inlining external declarations uses descriptors to look up the underlying bytecode for inlining. Since I'm doing name mangling for inline class functions at the IR level, these descriptors are replaced by WrappedDescriptors which the inliner currently does not handle.

The fact that this particular test was working before was a bit of a happy coincidence - it didn't really depend on the inline class machinery at all.

I'm planning to submit a separate PR for this later, since the inliner seems to be very much a work in progress.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+val IrTypeParameter.erasedUpperBound: IrClass
+    get() {
+        // Pick the (necessarily unique) non-interface upper bound if it exists
+        for (type in superTypes) {
+            val irClass = type.classOrNull?.owner ?: continue
+            if (!irClass.isJvmInterface) return irClass
+        }
+
+        // Otherwise, choose either the first IrClass supertype or recurse.
```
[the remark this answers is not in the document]
sfs: There really are two cases though, it's a coincidence that they can be combined into a single case in the code (and in the Java spec).

- If a type parameter has a type parameter upper bound we have to recurse.
- If a type parameter has only interface upper bounds we (arbitrarily) pick the first one.

The case distinction mentioned in the comment happens in `IrType.erasedUpperBound`.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -910,7 +910,8 @@ class KotlinTypeMapper @JvmOverloads constructor(
         val isConstructor = isConstructor(jvmSignature)
         val descriptor = getDefaultDescriptor(
             jvmSignature,
-            if (isStaticMethod(kind, functionDescriptor) || isConstructor) null else ownerType.descriptor,
+            if (isStaticMethod(kind, functionDescriptor) || isStaticDeclaration(functionDescriptor) || isConstructor)
+                null else ownerType.descriptor,
```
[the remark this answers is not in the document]
sfs: Done

sfs:
The first 10 commits are now in separate PRs (#2339, #2340, #2341).

sfs:
@max-kammerer Thanks for the heads up. I've updated the tests.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -615,6 +617,9 @@ class KotlinTypeMapper @JvmOverloads constructor(
             return name
         }
 
+        if (isIrBackend)
```
madsager: Looks like this ends up duplicating some code. Could we refactor this along these lines:

```
return if (isIrBackend) mangleNonInlineClassMemberNames() else mangleMemberNames()
```

And then have:

```
mangleMemberNames() {
  // all the early returns for inline class members.
  ...
  return mangleNonInlinClassMemberNames()
}
```

The we have sharing of the common parts in `mangleNonInlineClassMemberNames()`?

compiler/testData/codegen/boxInline/noInline/extensionReceiver.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
madsager: If you understand why this is now failing, can you add a comment.

compiler/testData/codegen/box/ranges/unsigned/expression/overflowZeroDownToMaxValue.kt
```diff
@@ -1,4 +1,7 @@
 // KJS_WITH_FULL_RUNTIME
+// TODO: muted automatically, investigate should it be ran for JVM_IR or not
```
madsager: Do you understand why these two tests are now failing. If so, can you replace this auto mute comment with an actual comment explaining what is going on?

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/DataClassMembersGenerator.kt
madsager: Hah! Thanks Denis, yes, I totally misunderstood that. :)

compiler/testData/codegen/box/ranges/unsigned/expression/overflowZeroDownToMaxValue.kt
sfs: Will do. Mark was wondering the same thing earlier. I'll add a comment to the generator for these tests.

compiler/testData/codegen/boxInline/noInline/extensionReceiver.kt
sfs: It seems like its just a missing test in IrInlineCodegen. Fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -1161,12 +1185,12 @@ class ExpressionCodegen(
...
 fun DefaultCallArgs.generateOnStackIfNeeded(callGenerator: IrCallGenerator, isConstructor: Boolean, codegen: ExpressionCodegen): Boolean {
     val toInts = toInts()
-    if (!toInts.isEmpty()) {
+    if (toInts.isNotEmpty()) {
         for (mask in toInts) {
             callGenerator.putValueIfNeeded(Type.INT_TYPE, StackValue.constant(mask, Type.INT_TYPE), ValueKind.DEFAULT_MASK, -1, codegen)
         }
 
-        val parameterType = if (isConstructor) AsmTypes.DEFAULT_CONSTRUCTOR_MARKER else AsmTypes.OBJECT_TYPE
+        val parameterType = if (isConstructor) AsmTypes.DEFAULT_CONSTRUCTOR_MARKER else OBJECT_TYPE
```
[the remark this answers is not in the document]
sfs: Fixed

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -167,7 +166,7 @@ class ExpressionCodegen(
                 irFunction.markLineNumber(startOffset = irFunction is IrConstructor && irFunction.isPrimary)
             }
             val returnType = typeMapper.mapReturnType(irFunction)
-            result.coerce(returnType).materialize()
+            result.coerce(returnType, if (irFunction !is IrConstructor) irFunction.returnType else null).materialize()
```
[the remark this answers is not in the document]
sfs: In this particular case, yes. Since this was confusing for a while now I've split the implementation of voidValue into two cases (a non-materialized value of type unit, and a promised default value).

This also made it possible to make the IrType field on PromisedValue non-nullable, but on the flip-side we now distinguish between materialized values on the stack and immaterial values.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrTypeMapper.kt
```diff
@@ -57,9 +58,15 @@ class IrTypeMapper(val kotlinTypeMapper: KotlinTypeMapper) {
     fun mapType(irType: IrType, sw: JvmSignatureWriter, mode: TypeMappingMode) =
         kotlinTypeMapper.mapType(irType.toKotlinType(), sw, mode)
 
+    fun mapTypeAsDeclaration(irType: IrType) =
```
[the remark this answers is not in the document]
sfs: It seems hard to extract the part of the code that is actually affected (name mangling) without a lot of refactoring in the TypeMapper. Since there is a rewrite in progress I would like to hold off on that.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
sfs: I changed it to get rid of the duplication.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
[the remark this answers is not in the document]
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -55,27 +70,73 @@ fun PromisedValue.discard(): MaterialValue {
...
+    if (!isFromTypeInlineClass && !isToTypeInlineClass) return null
+
+    val isFromTypeUnboxed = isFromTypeInlineClass && typeMapper.mapType(irType.unboxed) == type
+    val isToTypeUnboxed = isToTypeInlineClass && typeMapper.mapType(irTarget.unboxed) == target
+
+    return when {
+        isFromTypeUnboxed && !isToTypeUnboxed -> object : PromisedValue(codegen, target, irTarget) {
+            override fun materialize() {
+                this@coerceInlineClasses.materialize()
+                // TODO: This is broken for type parameters
```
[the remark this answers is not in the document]
sfs: Yes exactly. Rather than fixing it in the type mapper I changed the code to always go through the erased upper bound for inline class boxing instead.

It would be nice if we could always do this, but unfortunately the type mapper has additional side effects when processing generic types.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -55,27 +70,73 @@ fun PromisedValue.discard(): MaterialValue {
...
+                StackValue.boxInlineClass(irType.toKotlinType(), mv)
+            }
+        }
+        !isFromTypeUnboxed && isToTypeUnboxed -> object : PromisedValue(codegen, target, irTarget) {
+            override fun materialize() {
+                val value = this@coerceInlineClasses.materialized
+                StackValue.unboxInlineClass(value.type, irTarget.toKotlinType(), mv)
```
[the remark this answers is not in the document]
sfs: Fixed as well

sfs:
Rebased to latest master since #2341 got merged and fixed generated not-null assertions for inline class parameters.

udalov:
@sfs Please rebase on master once again. Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -503,7 +520,7 @@ internal class CallableReferenceLowering(val context: JvmBackendContext) : FileL
                 val irBuilder = context.createIrBuilder(function.symbol, startOffset, endOffset)
                 body = irBuilder.irBlockBody(startOffset, endOffset) {
                     +irReturn(
-                        IrConstImpl.string(-1, -1, context.irBuiltIns.stringType, callee.name.asString())
+                        IrConstImpl.string(-1, -1, context.irBuiltIns.stringType, callee.originalName.asString())
```
udalov: Probably, the same change will be useful in `PropertyReferenceLowering`. However, I don't fully understand how is it relevant for inline classes?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    private val IrFunction.propertyName: Name
+        get() = (this as IrSimpleFunction).correspondingPropertySymbol!!.owner.name
+
+    private fun hashSuffix(irFunction: IrFunction) =
+        md5base64(irFunction.fullValueParameterList.joinToString { it.type.eraseToString() })
+
+    private fun IrType.eraseToString() = buildString {
+        append('L')
+        append(erasedUpperBound.fqNameSafe)
```
udalov: Please use `.fqNameWhenAvailable!!`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,133 @@
...
+        if (isNullable()) append('?')
+        append(';')
+    }
+
+    private fun md5base64(signatureForMangling: String): String {
```
udalov: Let's simply make existing `md5base64` from `inlineClassManglingUtils.kt` public and reuse it

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -46,36 +67,92 @@ abstract class BooleanValue(mv: InstructionAdapter) : PromisedValue(mv, Type.BOO
...
     get() {
         materialize()
-        return MaterialValue(mv, type)
+        return MaterialValue(codegen, type, irType)
     }
 
 // Materialize and disregard this value. Materialization is forced because, presumably,
 // we only wanted the side effects anyway.
-fun PromisedValue.discard(): MaterialValue {
+fun PromisedValue.discard(): PromisedValue {
```
udalov: I believe the comment on this declaration is inaccurate now, since materialization will only happen on coercion

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.jetbrains.kotlin.ir.symbols.IrTypeParameterSymbol
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeImpl
+import org.jetbrains.kotlin.ir.types.impl.IrStarProjectionImpl
+import org.jetbrains.kotlin.ir.util.isNullable
+
+/**
+ *  Perform as much type erasure as is significant for JVM signature generation.
+ */
+fun IrType.eraseTypeParameters() = when (this) {
```
udalov: I see that this function was moved from BridgeLowering, but I'd appreciate if you added several examples of how it behaves on different types into the kdoc

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt
```diff
@@ -32,6 +32,7 @@ interface IrDeclarationOrigin {
     object FILE_CLASS : IrDeclarationOriginImpl("FILE_CLASS")
     object GENERATED_DATA_CLASS_MEMBER : IrDeclarationOriginImpl("GENERATED_DATA_CLASS_MEMBER")
     object GENERATED_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("GENERATED_INLINE_CLASS_MEMBER")
+    object SYNTHETIC_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("SYNTHETIC_INLINE_CLASS_MEMBER", isSynthetic = true)
```
udalov: Let's add this to `JvmLoweredDeclarationOrigin` instead. These origins are used in IR serialization format in Native and JS (`KotlinIr.proto`) and it's probably better to avoid adding JVM specifics here at this point

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,133 @@
...
+import org.jetbrains.kotlin.ir.util.*
+import org.jetbrains.kotlin.load.java.JvmAbi
+import org.jetbrains.kotlin.name.Name
+import java.security.MessageDigest
+import java.util.*
+
+/**
+ * Replace inline classes by their underlying types.
+ */
+fun IrType.unbox() = InlineClassAbi.unboxType(this) ?: this
```
udalov: I'd rename it to `unboxInlineClass` to avoid confusion with boxed/unboxed primitives

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Check if a function can be replaced.
+ */
+val IrFunction.isBoxed: Boolean
```
udalov: Maybe consider naming this `hasInlineClassParameters` or something more explicit, or make it `internal`.

Also I didn't understand the kdoc

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,133 @@
...
+        if (underlyingType.isNullable() || underlyingType.isPrimitiveType())
+            return null
+        return underlyingType.makeNullable()
+    }
+
+    /**
+     * Get the underlying type of an inline class based on the single argument to its
+     * primary constructor. This is what the current jvm backend does.
+     *
+     * Looking for a backing field does not work for built-in inline classes (UInt, etc.),
```
udalov: I'd replace "built-in inline classes (UInt, etc.)" with "unsigned types" to avoid confusion with types that are named built-in (`KotlinBuiltIns`) in the sense that any program has them in dependencies (a program can successfully compile without `UInt` in dependencies)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
sfs: Right now it's not needed for `PropertyReferenceLowering`, because it runs before inline class lowering and already uses descriptors to get the property name. If you want I can change it, though.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
sfs: I've added some more explanations and examples to the comment.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt
sfs: Thanks for the explanation, I wasn't aware of this! Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

sfs:
@udalov Thanks for the comments. I've addressed them and rebased to the current master again.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -1617,7 +1621,8 @@ class KotlinTypeMapper @JvmOverloads constructor(
             val containingDeclaration = descriptor.containingDeclaration
             return containingDeclaration.isInlineClass() &&
                     descriptor.kind == CallableMemberDescriptor.Kind.SYNTHESIZED &&
-                    descriptor.name == InlineClassDescriptorResolver.BOX_METHOD_NAME
+                    (descriptor.name == InlineClassDescriptorResolver.BOX_METHOD_NAME ||
+                     descriptor.name.asString() == BOX_JVM_METHOD_NAME)
```
max-kammerer: Is it possible to add test for old backend in this case?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -42,20 +43,20 @@ class IrInlineCodegen(
...
         codegen: ExpressionCodegen,
         blockInfo: BlockInfo
     ) {
-        if (irValueParameter?.isInlineParameter() == true && isInlineIrExpression(argumentExpression)) {
-            val irReference: IrFunctionReference =
-                (argumentExpression as IrBlock).statements.filterIsInstance<IrFunctionReference>().single()
-            rememberClosure(irReference, parameterType, irValueParameter) as IrExpressionLambdaImpl
-        } else {
-            putValueOnStack(argumentExpression, parameterType, irValueParameter?.index ?: -1, blockInfo)
+        if (irValueParameter.isInlineParameter() && isInlineIrExpression(argumentExpression)) {
```
max-kammerer: Please pass flag here: is parameter extensionReceiver or not, and check it in 'if'.
Now non-null 'irValueParameter' is passed and inliner tries to inline extension receiver argument as lambda one

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
sfs: Nice catch. Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
sfs: Is this the only situation where `isInlineParameter` and `isInlineIrExpression` are both satisfied and yet the argument does not have the right shape?

Assuming that this is the case, I simplified the code by moving the check to the `isInlineParameter` function.

max-kammerer: Yes, receiver inlining is not supported in old backend and in IR for now

sfs: Good to know! Fixed.

max-kammerer:
@sfs There are some new conflicts could you please rebase again

sfs:
@max-kammerer Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+                }
+            }
+
+            declaration.transformDeclarationsFlat { memberDeclaration ->
+                when (memberDeclaration) {
+                    is IrFunction ->
+                        transformFunctionFlat(memberDeclaration)
+                    else -> {
+                        memberDeclaration.accept(this, null)
+                        listOf(memberDeclaration)
```
udalov: Please use `null` here (also consider converting `when` to `if`)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Keeps track of replacement functions and inline class box/unbox functions.
+ */
+class InlineClassManager {
+    private val storageManager = LockBasedStorageManager("inline-class-manager")
+
+    /**
+     * Get a replacement for a function or a constructor.
+     */
+    val getReplacementFunction: (IrFunction) -> IrReplacementFunction? =
```
udalov: I'm having trouble understanding why should anything be lazy in this implementation. I'd suggest maybe using the following approach instead of laziness: first, lower all function declarations to their corresponding replacements, and record the mapping (function -> replacement). Second, traverse the whole file again and lower all usages (calls, references, etc), using the recorded mapping. Basically, `JvmInlineClassLowering.lower` would consist of two passes, with different visitors, with data flowing naturally from the first to the second. What do you think?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+        declaration.delegate.transformChildrenVoid()
+        declaration.getter = transformFunctionFlat(declaration.getter)[0] as IrFunction
+        declaration.setter?.let { setter ->
+            declaration.setter = transformFunctionFlat(setter)[0] as IrFunction
+        }
+        return declaration
+    }
+
+    private fun buildPrimaryInlineClassConstructor(irClass: IrClass, irConstructor: IrConstructor) {
+        irClass.declarations += buildConstructor {
```
udalov: Please consider using `addConstructor`/`addFunction`/`addValueParameter`/... family of functions instead of `buildConstructor` et al, throughout the file. The advantage is that they automatically set `parent` and add the new declaration to parent's `declarations`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,121 @@
...
+
+internal val IrFunction.fullValueParameterList: List<IrValueParameter>
+    get() = listOfNotNull(extensionReceiverParameter) + valueParameters
+
+internal val IrClass.inlineClassFieldName: Name
+    get() = constructors.single { it.isPrimary }.valueParameters.single().name
+
+val IrFunction.isInlineClassFieldGetter: Boolean
+    get() = (parent as? IrClass)?.isInline == true && this is IrSimpleFunction &&
+            correspondingPropertySymbol?.let { it.owner.getter == this && it.owner.name == parentAsClass.inlineClassFieldName } == true
```
udalov: It's not enough to check property name, consider:
```
inline class Z(val s: String) {
    val Int.s: Int get() = 42
}
```

(upon fixing this, please add a test if no other test becomes unmuted)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,121 @@
...
+            (this is IrSimpleFunction || this is IrConstructor && constructedClass.isInline && !isPrimary)
+
+internal val IrFunction.hasMethodReplacement: Boolean
+    get() = dispatchReceiverParameter != null && this is IrSimpleFunction && (parent as? IrClass)?.isInline != true
+
+internal val IrFunction.fullValueParameterList: List<IrValueParameter>
+    get() = listOfNotNull(extensionReceiverParameter) + valueParameters
+
+internal val IrClass.inlineClassFieldName: Name
+    get() = constructors.single { it.isPrimary }.valueParameters.single().name
```
udalov: There are so many `constructors.single { it.isPrimary }` usages now in this change and elsewhere, that I'd consider adding something like this to the common IR utils:
```
val IrClass.primaryConstructor: IrConstructor?
    get() = constructors.singleOrNull(IrConstructor::isPrimary)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    private fun IrType.eraseToString() = buildString {
+        append('L')
+        append(erasedUpperBound.fqNameWhenAvailable!!)
+        if (isNullable()) append('?')
+        append(';')
+    }
+}
+
+internal val IrFunction.hasInlineClassParameters: Boolean
```
udalov: I propose to move all these properties which are used only in `InlineClassManager` closer to the usage and make them private

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
```diff
@@ -0,0 +1,145 @@
...
+
+        val parameterMap = mutableMapOf<IrValueParameterSymbol, IrValueParameter>()
+        val replacement = buildReplacement(function) {
+            annotations += function.annotations
+            metadata = function.metadata
+            overriddenSymbols.addAll(overrides)
+
+            for ((index, parameter) in function.explicitParameters.withIndex()) {
+                val name = if (parameter == function.extensionReceiverParameter) Name.identifier("\$receiver") else parameter.name
+                val newParameter = parameter.copyTo(this, index = index - 1, name = name)
```
udalov: Why `index - 1`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/ScopedValueMappingTransformer.kt
```diff
@@ -0,0 +1,92 @@
...
+            ).apply {
+                value = expression.value.transform(this@ScopedValueMappingTransformer, null)
+            }
+        }
+        return super.visitSetVariable(expression)
+    }
+}
+
+class ScopedValueDeclarationMap {
+    private val scopeStack = mutableListOf<MutableMap<IrValueSymbol, IrValueDeclaration>>(mutableMapOf())
```
udalov: Could you please explain (with sample code maybe) why is a stack needed here and why is it not just a big mutable map? I got an impression that keys (`IrValueSymbol` instances) are unique here, which could allow us to simplify code by not bothering to remove entries, but I think I'm missing something

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+
+        // Don't create a wrapper for functions which are only used in an unboxed context
+        if (function.overriddenSymbols.isEmpty() || worker.dispatchReceiverParameter != null)
+            return listOf(worker)
+
+        // Replace the function body with a wrapper
+        scoped {
+            context.createIrBuilder(function.symbol).run {
+                val fullParameterList =
+                    listOfNotNull(function.dispatchReceiverParameter, function.extensionReceiverParameter) + function.valueParameters
```
udalov: `function.explicitParameters`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
udalov: Sorry, I figured that this approach won't work because we need to replace even usages of functions that are declared outside of our file.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
```diff
@@ -0,0 +1,145 @@
...
+
+class IrReplacementFunction(
+    val function: IrFunction,
+    val valueParameterMap: Map<IrValueParameterSymbol, IrValueParameter>
+)
+
+/**
+ * Keeps track of replacement functions and inline class box/unbox functions.
+ */
+class InlineClassManager {
```
udalov: I suggest using some other name, something like `MemoizedInlineClassReplacements`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+                                    return expression
+
+                                return irReturn(irBlock(expression.startOffset, expression.endOffset) {
+                                    +expression.value
+                                    +irGet(thisVar)
+                                })
+                            }
+                        }, null)
+                        .transform(this@JvmInlineClassLowering, null)
+                        .patchDeclarationParents(worker)
```
udalov: Why is `patchDeclarationParents` needed here?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+                            // the return expression. Afterwards we return "thisVar".
+                            // For example, the following is a valid inline class declaration.
+                            //
+                            //     inline class Foo(val x: String) {
+                            //       constructor(y: Int) : this(y.toString()) {
+                            //         if (y == 0) return throw java.lang.IllegalArgumentException()
+                            //         if (y == 1) return
+                            //         return Unit
+                            //       }
+                            //     }
```
udalov: Please ensure there's a test, or add a new one, that checks side effects here

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -0,0 +1,402 @@
...
+    private fun coerceInlineClasses(argument: IrExpression, from: IrType, to: IrType) =
+        IrCallImpl(UNDEFINED_OFFSET, UNDEFINED_OFFSET, to, context.ir.symbols.unsafeCoerceIntrinsicSymbol).apply {
+            putTypeArgument(0, from)
+            putTypeArgument(1, to)
+            putValueArgument(0, argument)
+        }
+
+    private fun visitPrimaryInlineClassConstructorCall(expression: IrMemberAccessExpression, constructor: IrConstructor): IrExpression {
+        val parameter = constructor.valueParameters.single()
+        val arg = expression.getValueArgument(0) ?: parameter.defaultValue!!.expression.deepCopyWithSymbols()
```
udalov: Loading parameter's default value won't work with an inline class loaded from dependencies. Can't we simply leave this call without an argument if it didn't have one, since `defaultArgumentStubPhase` is running right after this lowering?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Nice catch - this code was ported from the current backend and this is indeed a bug even in the current backend. Specifically, the code fails with a verification error.

I'll submit an issue for this and add a test.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Returning `null` from `transformDeclarationsFlat` will remove the declaration in question, which is not what we want.

I changed the `when` to an `if`, though. There used to be more cases. ;)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: I've changed it here, but cannot change it in `build(Un)BoxFunction` or `InlineClassManager.kt` since these functions potentially run before the lowering and for classes in other files. In this case we don't want to add the functions to the class (yet).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
sfs: This method is only called on functions with non-zero dispatch receiver. Thus the iterator for `function.explicitParameters.withIndex()` starts with the dispatch receiver at index `0`, while it should get index `-1`. At the same time we are turning the extension receiver into a normal value parameter, which should be at index `0` instead of `-1`. This is why we can't just use `parameter.index` and instead have to use `index - 1`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/ScopedValueMappingTransformer.kt
sfs: Looking at this code again, I don't think it's necessary anymore. I've removed it and simplified the code.

For completeness though, the use case is explained in the comment to `ScopedValueMappingTransformer`. It's for the situation where you copy parts of one function into another. In this case you want to remap the value parameters in the replacement function but not in the body of the original, yet you have to traverse both.

Previously, the lowering also added coercions and changed types. That was where the second traversal came from. Right now the lowering is simpler so that a map suffices. I've changed it and removed the transformer.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassManager.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Because we copy the statement from the constructor to a static replacement function. It's possible to do this inline in one of the other traversals, but this just ends up duplicating code which is already in `patchDeclarationParents`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Already done! I've added `testSimpleSecondaryConstructor` in the inline class tests which checks this case.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: That's a good point. The problem here is that calls to the inline class constructor are replaced by the `unsafeCoerce` intrinsic, which is not treated in `defaultArgumentStubPhase`.

The way this is handled in the current backend is by always creating a `constructor-impl` function for the primary constructor (which doesn't do anything except nullability checks). This function is in turn lowered by `defaultArgumentStubPhase` which produces the correct behavior.

It's probably a good idea to reproduce this behavior anyway, because otherwise we might miss some null checks. I've added this to the IR lowering.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
udalov: Maybe I misunderstand something, but the kdoc says explicitly:
https://github.com/JetBrains/kotlin/blob/c67517cb2bc99844d5393a655ef264ce46d05b73/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/transform.kt#L32

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: I've opened an issue on [YouTrack](https://youtrack.jetbrains.com/issue/KT-31927) and added a corresponding [test](https://github.com/sfs/kotlin/blob/148590b45fde3eb5a74064974a8a43e713afaa24/compiler/testData/codegen/box/inlineClasses/fieldNameClash.kt).

Unfortunately, while I've fixed this in the IR backend, there seems to be something wrong in the type mapper. Specifically, calling `mapType` on class `Z` in your example above results in an incorrect type. I'll need to look into the type mapper code a bit more before fixing this.

sfs:
@udalov Thanks for all the comments. I've addressed them and (once again) rebased to master.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Huh, you're right. No idea why I thought otherwise. I'll change it momentarily.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -484,4 +503,4 @@ fun IrClass.addFakeOverrides() {
 }
 
 fun IrValueParameter.isInlineParameter() =
```
max-kammerer: @sfs As I see this method is only used in JVM backend so it's best to move it back to JVM utils cause extension inlining could depend on backend: ```index >=0```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
sfs: Done

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
```diff
@@ -40,6 +40,9 @@ inline fun IrBuilderWithScope.irLetS(
     nameHint: String? = null,
     body: (IrValueSymbol) -> IrExpression
 ): IrExpression {
+    if (value is IrGetValue)
```
max-kammerer: @sfs Is this optimization still required?
As I see this method only used in DataClassMembersGenerator and a getter call is always passed

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
sfs: That's a good point, I've moved the function to the IrUtils file in the jvm backend.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
sfs: You're right, it's no longer required. I would still like to leave it in, since it will probably come up again in the future (e.g., when generating a safe call in the IR). What do you think?

max-kammerer: We need some ir-text test here, but it's not clear how test anything in coming up changes. Maybe best is to move this commit to separate pr for now and discuss details there

max-kammerer:
@sfs Please rebase once more, there is new conflict afrer #2398 merge

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
sfs: I've removed it from this PR. I will try to commit it as part of #2362, where the optimization is actually required for reasonable codegen.

sfs:
@max-kammerer Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
udalov: OK, thanks! I suppose it can be fixed separately from this PR then.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
udalov: Great, thanks. Note that in the current implementation the bytecode produced for `constructor-impl` after this lowering contains an unnecessary creation of the inline class:
```
    Code:
      stack=3, locals=1, args_size=1
         0: aload_0
         1: ldc           #71                 // String s
         3: invokestatic  #17                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
         6: new           #2                  // class Z
         9: dup
        10: aload_0
        11: invokespecial #73                 // Method "<init>":(Ljava/lang/String;)V
        14: invokevirtual #38                 // Method "unbox-impl":()Ljava/lang/String;
        17: areturn
```

But I suppose it can be fixed later

sfs: Thank you for looking at the generated code. I've just fixed it and will add a test.

sfs: Done

udalov:
Merged manually with minor changes to new tests (ignore on Native and/or JS).

@sfs Thank you. Great work!

PR #2327, opened 2019-05-10
JVM_IR: Support access flag jvm annotations on fields and methods.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -342,7 +344,9 @@ private val IrClass.flags: Int
 private val IrField.flags: Int
     get() = origin.flags or visibility.flags or
             (if (isFinal) Opcodes.ACC_FINAL else 0) or
-            (if (isStatic) Opcodes.ACC_STATIC else 0)
+            (if (isStatic) Opcodes.ACC_STATIC else 0) or
+            (annotations.findAnnotation(VOLATILE_ANNOTATION_FQ_NAME)?.let { Opcodes.ACC_VOLATILE } ?: 0) or
```
max-kammerer: Maybe `if` with `hasAnnotation`?

madsager: Thanks! :)

max-kammerer:
Thanks!

PR #2329, opened 2019-05-13
JVM_IR: support references to Java fields

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -258,120 +272,76 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
-                }
 
-                expression.setter?.owner?.let { setter ->
-                    buildOverride(superClass.functions.single { it.name.asString() == "set" }) { dispatchReceiverParameter, valueParameters ->
-                        val type = setter.valueParameters.last().type
-                        val value = irImplicitCast(irGet(valueParameters.last()), type)
-                        irSet(setter.returnType, null, setter.symbol, value).apply {
-                            copyTypeArgumentsFrom(expression)
-                            setReceiversOn(this, dispatchReceiverParameter, valueParameters)
+                    if (!field.isFinal) {
```
max-kammerer: Maybe best ```field.correspondingPropertySymbol.owner.isVar ```?

pyos: correspondingPropertySymbol is null for Java fields.

max-kammerer:
Thanks!

PR #2330, opened 2019-05-14
Replicate old backend's naming logic for captured receiver parameter and extension receiver parameters.

neetopia:
Replicate old backend's naming logic for captured receiver parameter in constructors and extension receiver parameters.

Currently names for captured receiver parameter and extension receiver parameter are not transformed.

Also added a LVT test for extension receiver parameter (Only captured receiver parameter test is available now)

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
@@ -11,6 +11,7 @@ import org.jetbrains.kotlin.backend.common.descriptors.*
 import org.jetbrains.kotlin.backend.common.ir.copyTo
 import org.jetbrains.kotlin.backend.common.ir.copyTypeParametersFrom
 import org.jetbrains.kotlin.descriptors.Modality
+import org.jetbrains.kotlin.descriptors.ReceiverParameterDescriptor
 import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.descriptors.Visibility
 import org.jetbrains.kotlin.ir.IrElement
@@ -53,6 +54,8 @@ val IrDeclaration.parents: Sequence<IrDeclarationParent>
     get() = parentsWithSelf.drop(1)
 
 object BOUND_VALUE_PARAMETER: IrDeclarationOriginImpl("BOUND_VALUE_PARAMETER")
+
+object BOUND_RECEIVER_PARAMETER: IrDeclarationOriginImpl("BOUND_RECEIVER_PARAMETER")
 
 class LocalDeclarationsLowering(
     val context: BackendContext,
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
```diff
@@ -583,8 +585,17 @@ class LocalDeclarationsLowering(
...
                 IrValueParameterImpl(
-                    p.startOffset, p.endOffset, BOUND_VALUE_PARAMETER, IrValueParameterSymbolImpl(parameterDescriptor),
-                    suggestNameForCapturedValue(p), i, p.type, null, false, false
+                    p.startOffset,
+                    p.endOffset,
+                    if (p.descriptor is ReceiverParameterDescriptor && newDeclaration is IrConstructor)
+                        BOUND_RECEIVER_PARAMETER else BOUND_VALUE_PARAMETER,
+                    IrValueParameterSymbolImpl(parameterDescriptor),
+                    suggestNameForCapturedValue(p)
+                    , i,
```
yanex: Could you please fix the formatting on this line?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -202,6 +208,21 @@ class ExpressionCodegen(
...
+    private fun writeReceiverParameterInLocalVariableTable(param: IrValueParameter, startLabel: Label, endLabel: Label) {
+
+        //TODO: get rid of descriptor
+        val name = getNameForReceiverParameter(
+            irFunction.descriptor,
+            typeMapper.kotlinTypeMapper.bindingContext,
+            context.configuration.languageVersionSettings)
+
+        val type = typeMapper.mapType(param)
+        typeMapper.kotlinTypeMapper.bindingContext
```
yanex: Looks like this line has no side-effects. Was it added on purpose?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -202,6 +208,21 @@ class ExpressionCodegen(
         )
     }
 
+    private fun writeReceiverParameterInLocalVariableTable(param: IrValueParameter, startLabel: Label, endLabel: Label) {
```
yanex: As name is the only thing that changes, I'd move the code back to `writeValueParameterInLocalVariableTable()` and added a `when` (or moved only the name calculation to the separate function).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
neetopia: > Looks like this line has no side-effects. Was it added on purpose?

Should be a mistake I reckon, I'll remove it.

yanex:
Merged into master manually (https://github.com/JetBrains/kotlin/commit/2c102ecd570e156994da543fe3c3b755abda0c2b). Thanks for the contribution!

PR #2331, opened 2019-05-14
MoveCompanionObjectFields: handle more cases

ting-yuan:
Constant / @JvmField / lateinit fields in companion objects should be
moved to corresponding classes and interfaces.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -83,9 +84,15 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
         companion.declarations.removeAll { it is IrAnonymousInitializer }
     }
 
-    private fun IrClass.allFieldsAreJvmField() =
-        declarations.filterIsInstance<IrProperty>()
-            .mapNotNull { it.backingField }.all { it.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME) }
+    private fun shouldExposeProperty(irClass: IrClass, irProperty: IrProperty): Boolean {
+        if (!irClass.isInterface && !irClass.isAnnotationClass)
```
punzki: A little bit confused on this... I thought the property _should_ be moved from the companion object of the current class _is_ an interface or annotation class?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -83,9 +84,15 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
         companion.declarations.removeAll { it is IrAnonymousInitializer }
     }
 
-    private fun IrClass.allFieldsAreJvmField() =
-        declarations.filterIsInstance<IrProperty>()
-            .mapNotNull { it.backingField }.all { it.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME) }
+    private fun shouldExposeProperty(irClass: IrClass, irProperty: IrProperty): Boolean {
```
punzki: What does "exposing" the property mean? I suggest renaming it (`shouldMoveProperty`?) since "expose" isn't used anywhere else.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -83,9 +84,15 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
...
+        if (!irClass.isInterface && !irClass.isAnnotationClass)
+            return true
+
+        return irProperty.backingField?.let {
+            it.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME) || it.correspondingPropertySymbol?.owner?.isConst == true ||
```
punzki: Is `it.correspondingPropertySymbol?.owner` the same as `irProperty.symbol`? If so, could you just use the latter? It is not nullable and you don't have to use `== true`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
sfs:
```kotlin
if (!irClass.isJvmInterface)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
sfs: `it.corresponginPropertySymbol?.owner` should be the same thing as `irProperty`. I guess the check is in the body of the let to exclude the case of delegated properties?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -83,9 +84,15 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
...
+                    it.correspondingPropertySymbol?.owner?.isLateinit == true
```
sfs: I don't see any tests with lateinit properties. Is this really required? In general, could you please add a comment to this method explaining why only these particular cases should be moved?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -66,12 +66,13 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
...
         companion.declarations.forEach {
             when (it) {
                 is IrProperty -> {
-                    val newField = movePropertyFieldToStaticParent(it, companion, irClass, fieldReplacementMap)
-                    if (newField != null) irClass.declarations.add(newField)
+                    if (shouldExposeProperty(irClass, it)) {
+                        val newField = movePropertyFieldToStaticParent(it, companion, irClass, fieldReplacementMap)
+                        if (newField != null) irClass.declarations.add(newField)
+                    }
                 }
```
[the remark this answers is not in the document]
ting-yuan: Good idea. Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: It's used in:
https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#static-fields

meaning that exposing a Kotlin property as a Java field.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: Good catch! Thanks.

No, it's just me not knowing that they are the same thing; I started this patch from somewhere else dealing with IrField and found later that they are mostly done in this phase except for these cases.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: I'll use shouldMoveProperty and document the link in comments.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: @punzki My understanding is that they should be moved into the parent class/interface whenever possible. A JVM specific limitation is that interfaces can only contain `public static final` fields so the checks have to be there. Does that make sense?

@sfs Thanks, that is neat.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: I followed [the doc](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#static-fields) blindly. They are probably not needed as it doesn't make sense for interface fields, which are `public static final`, to be lateinit.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
punzki: @ting-yuan That makes sense now, thanks for clarifying. I think a code comment could add clarity, since it's confusing why these particular conditions are used (especially the `!isJvmInterface`).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
punzki: Why did you remove the `isLateinit == true` condition in the latest version? Yes for interfaces it doesn't make sense, but you can have a `lateinit` property in a companion object for non-interface classes.

ting-yuan: Yes, `lateinit` properties will be copied for non-interfaces classes. They are enabled by the first condition `!irClass.isJvmInterface`. I've updated the comments.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: Please take a look at the updated comments and let me know if it should be improved.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -66,26 +67,25 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
...
             }
         }
         companion.declarations.removeAll { it is IrAnonymousInitializer }
     }
 
-    private fun IrClass.allFieldsAreJvmField() =
-        declarations.filterIsInstance<IrProperty>()
-            .mapNotNull { it.backingField }.all { it.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME) }
+    // Properties in companions are always copied into their declaration parents except for interfaces, which can only host
+    // public static final fields. Only constants and values annotated with @JvmField are copied into interfaces*.
```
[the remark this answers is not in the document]
ting-yuan: Done. Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -66,26 +67,25 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
...
+                else -> null
             }
         }
         companion.declarations.removeAll { it is IrAnonymousInitializer }
     }
 
-    private fun IrClass.allFieldsAreJvmField() =
-        declarations.filterIsInstance<IrProperty>()
-            .mapNotNull { it.backingField }.all { it.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME) }
+    // Properties in companions are always copied into their declaration parents except for interfaces, which can only host
```
max-kammerer: @ting-yuan Seems it's a bug in documentation around interface companion properties.
Actually by default all properties in interface companion are not moved to outer interface. 'Const' properties just copied to it (so there are 2 representation of them).
If interface companion have all fields marked as @JvmFiled then only in such case all fields are moved.

ting-yuan: @max-kammerer So, for example, if a companion object in an interface has a field marked with @JvmField and another field without, both of them shouldn't be moved nor copied into the outer interface. Is that correct? If that is the case, how do Java refer to the one with @JvmField?

I uploaded a patch implementing that. Please take a look if this is the desired behavior.

max-kammerer: @ting-yuan Yes it's correct, @JvmField in such cases is prohibeted (it's not applicable by default in interface companion). Doc team should update documentation

max-kammerer:
It's worth to rename MoveCompanionObjectFieldsLowering  -> MoveOrCopyCompanionObjectFieldsLowering

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -83,6 +86,16 @@ private class MoveCompanionObjectFieldsLowering(val context: CommonBackendContex
         companion.declarations.removeAll { it is IrAnonymousInitializer }
     }
 
+    private fun copyConsts(irClass: IrClass) {
+        val companion = irClass.declarations.find {
+            it is IrClass && it.isCompanion
+        } as IrClass? ?: return
+        val unused = mutableMapOf<IrFieldSymbol, IrFieldSymbol>()
+        companion.declarations.filter { it is IrProperty && it.isConst }.mapNotNullTo(irClass.declarations) {
+            movePropertyFieldToStaticParent(it as IrProperty, companion, irClass, unused)
```
max-kammerer: In such cases old field should not be replaced, all referenced should be kept. New copy is not used in Kotlin, it's created only for Java

ting-yuan: @max-kammerer Thanks for the explanation. I misunderstood how the old field is held.

In the case of copy, the latest patch simply doesn't touch `irProperty.backingField`, so that the old field is retained. The new field is held directly by the interface and doesn't have a `correspondingPropertySymbol`. The references are not updated, of course. Does that make sense?

Please take a look again. I also verified with an example and saw that both new and old const exist.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -31,19 +32,21 @@ import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 import org.jetbrains.kotlin.load.java.JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME
 
-internal val moveCompanionObjectFieldsPhase = makeIrFilePhase(
-    ::MoveCompanionObjectFieldsLowering,
+internal val moveOrCopyCompanionObjectFieldsPhase = makeIrFilePhase(
+    ::MoveOrCopyCompanionObjectFieldsLowering,
     name = "MoveCompanionObjectFields",
     description = "Move companion object fields to static fields of companion's owner"
```
max-kammerer: Please also update description and name

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
max-kammerer: Yes, everything is fine except obsolete name and desc

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
ting-yuan: Done!

max-kammerer:
Thank you!

PR #2333, opened 2019-05-14
Check classpath entries exist before reading them.

brianduff:
This fixes a potential crash for nonexistent kapt3 classpath entries. Prior to the incremental annotation processing feature being introduced, these would be silently ignored. Even when the incremental annotation processing flag is off, these classpath entries are read, and nonexistent ones trigger an IOException.


abelkov:
Do we have a related YouTrack issue at https://youtrack.jetbrains.com/issues/KT ? If not, please consider creating one.

brianduff:
https://github.com/JetBrains/kotlin/pull/2334 is the right fix for this.

PR #2334, opened 2019-05-14
If incremental KAPT is disabled, do not analyze classpath

gavra0:
When incremental apt mode in KAPT is disabled, do not analyze
classpath to determine the type of the annotation processors. Instead,
just mark them all as non-incremental.


yanex:
Thank you for the contribution! Merged manually (master and 1.3.40-backup branches).

PR #2335, opened 2019-05-14
Generate refs for vals initialized in lambdas

pyos:


compiler/testData/codegen/box/contracts/exactlyOnceNotInline.kt
```diff
@@ -1,6 +1,6 @@
 // !LANGUAGE: +AllowContractsForCustomFunctions +UseCallsInPlaceEffect +ReadDeserializedContracts
 // !USE_EXPERIMENTAL: kotlin.contracts.ExperimentalContracts
-// IGNORE_BACKEND: JVM_IR, NATIVE, JS_IR
```
romanart: Could you also please unmute JS_IR tests?

compiler/testData/codegen/box/properties/lateinit/local/implicitLocalLateinit.kt
```diff
@@ -0,0 +1,7 @@
+fun box(): String {
```
max-kammerer: This test is redundant: lateinit logic exists in localLateinit.kt, non lateinit in exactlyOnceCrossinline.kt and definiteValInitialization.kt

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
```diff
@@ -80,7 +80,8 @@ class SharedVariablesLowering(val context: BackendContext) : FunctionLoweringPas
                 override fun visitVariable(declaration: IrVariable, data: IrDeclarationParent?) {
                     declaration.acceptChildren(this, data)
 
-                    if (declaration.isVar) {
+                    // `val` locals can be initialized in a lambda inlined into `run` or similar.
+                    if (declaration.isVar || declaration.initializer == null) {
```
max-kammerer: What about such cases:
```
 val z: Int
    z = 1
```
seems we also should check kind of IrValueAccessExpression

compiler/testData/codegen/box/properties/lateinit/local/implicitLocalLateinit.kt
pyos: This test is much simplier than exactlyOnceCrossinline.kt and definiteValInitialization.kt, though, both of which also check contracts and inline functions (and fail on JVM_IR right now due to other inliner bugs).

pyos:
@max-kammerer can you have a look at this again, please?

(I closed this PR because I had assumed that the approach was incorrect, as the non-IR backend does not have any refs in its output, but apparently they were removed by CapturedVarsOptimizationMethodTransformer at the very end; not sure why it doesn't work on the output of the IR backend, though.)

max-kammerer:
It's a bit tricky story. As I remember old backend didn't create Ref wrapper for captured vals at all and to support late variable initialization through simple inline functions with contracts  there was made some magic in inliner - MethodInliner.kt:632, that allows to write into local slots directly.  
From somepoint there was made 8a01da6ec641d4f94866efaea9d1e45fd781694a commit that allow val wraps for non-inline cases but this commit also add some new case support with crossinline arguments in inline functions. Maybe @ilmirus can clarify situation with contract and crossinline arguments. But at least in simple cases Ir backend should avoid Ref wrapping generation (e.g. definiteValInitialization.kt)

ilmirus:
Well, the idea is simple: `val`s and `var`s should behave the same if they are captured by `EXACTLY_ONCE` lambdas, even when they are `noinline` or `crossinline` lambdas. Weird case of `crossinline` and `EXACTLY_ONCE` is merely a bug. Since the bug is now a blocker for you, I will work on a fix ASAP.

pyos:
Closing because @udalov committed pretty much the same thing in https://github.com/JetBrains/kotlin/commit/66e19b13ce42b21bede4fedf9a0b0fd331233878.

udalov:
@pyos Sorry, somehow I completely missed your changes here. :(

I though it's better to have `EXACTLY_ONCE` lambdas and val initialization supported first, even suboptimally, and investigate how to get rid of `Ref` afterwards. @max-kammerer I've added a test to ensure that we don't forget to optimize `Ref`s here: https://github.com/JetBrains/kotlin/blob/66e19b13ce42b21bede4fedf9a0b0fd331233878/compiler/testData/codegen/bytecodeText/capturedVarsOptimization/capturedValInLambdaInitializedInside.kt. I'm not sure if it's possible to use the same trick (writing to the local directly) in the IR backend, so I think we should look into optimizing this in `CapturedVarsOptimizationMethodTransformer`, as I mentioned in a comment in the test.

PR #2336, opened 2019-05-14
KT-12567 "Introduce 'when' subject" intention does not work for "this" in extension function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12567


mglukhikh:
Merged 👍

PR #2338, opened 2019-05-15
KT-14369 "Replace elvis expression with 'if" intention produces boilerplate code for 'return' in RHS

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14369

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/ElvisToIfThenIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/ElvisToIfThenIntention.kt
@@ -78,6 +78,22 @@ class ElvisToIfThenIntention : SelfTargetingRangeIntention<KtBinaryExpression>(K
         val right = KtPsiUtil.safeDeparenthesize(element.right!!)
 
         val leftSafeCastReceiver = left.findSafeCastReceiver(context)
+        if (leftSafeCastReceiver == null) {
+            val property = (KtPsiUtil.safeDeparenthesize(element).parent as? KtProperty)
+            val propertyName = property?.name
+            if ((right is KtReturnExpression || right is KtBreakExpression || right is KtContinueExpression || right is KtThrowExpression)
+                && propertyName != null
+            ) {
+                val parent = property.parent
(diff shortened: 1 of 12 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/ElvisToIfThenIntention.kt
```diff
@@ -77,6 +77,18 @@ class ElvisToIfThenIntention : SelfTargetingRangeIntention<KtBinaryExpression>(K
         val left = KtPsiUtil.safeDeparenthesize(element.left!!)
         val right = KtPsiUtil.safeDeparenthesize(element.right!!)
 
+        val property = (KtPsiUtil.safeDeparenthesize(element).parent as? KtProperty)
+        val propertyName = property?.name
+        if (propertyName != null && right is KtReturnExpression) {
```
mglukhikh: We can do the same for `break`, `continue` and `throw`

idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/intentions/ElvisToIfThenIntention.kt
```diff
@@ -77,6 +77,18 @@ class ElvisToIfThenIntention : SelfTargetingRangeIntention<KtBinaryExpression>(K
...
+            val parent = property.parent
+            val factory = KtPsiFactory(element)
+            factory.createExpressionByPattern("if ($0 == null) $1", propertyName, right)
+            parent.addAfter(factory.createExpressionByPattern("if ($0 == null) $1", propertyName, right), property)
+            parent.addAfter(factory.createNewLine(), property)
+            element.replace(left)
+            return
+        }
+
```
mglukhikh: Could you please also check how it works in safe cast situation (see code about safe cast receiver below)? For example
```
class My(val x: Int)
fun foo(arg: Any?) {
    val y = (arg as? My)?.x ?: return
    // Before your commit, it gives "val y = if (arg is My) arg.x else return", what after?
}
```

mglukhikh:
Merged, thank you!

PR #2339, opened 2019-05-15
Ir type fixes

sfs:


compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/DataClassMembersGenerator.kt
```diff
@@ -254,13 +256,17 @@ class DataClassMembersGenerator(
...
                     ) { variable ->
+                        val get = irGet(irPropertyType, variable)
                         irIfNull(
                             context.irBuiltIns.intType,
                             irGet(irPropertyType, variable),
                             irInt(0),
                             getHashCodeOf(
                                 propertyType,
-                                irGet(irPropertyType, variable)
+                                if (irPropertyType.isMarkedNullable())
```
dnpetrov: Why is this implicit cast required?

sfs: Good catch - it's not since there is an implicit coercion between the branches of a when expression and the result. This was from a separate branch where I experimented with removing the implicit coercions between IR elements. Removed.

sfs:
@dnpetrov Is there anything else that prevents this PR from being merged?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/ToArrayLowering.kt
```diff
@@ -141,7 +139,7 @@ private class ToArrayLowering(private val context: JvmBackendContext) : ClassLow
                 toArrayName,
                 Visibilities.PUBLIC,
                 Modality.OPEN,
-                returnType = irBuiltIns.arrayClass.typeWith(irBuiltIns.anyType),
+                returnType = irBuiltIns.arrayClass.typeWith(irBuiltIns.anyNType),
```
max-kammerer: Could we extract nullability information from collection type?

sfs: Sure! I'm not sure how useful it would be though, given that a user would see the frontend type `Any?` from [toArray](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-abstract-collection/to-array.html). What do you think?

max-kammerer: Seems not. But it's look like there could be some inconsistency if 'toArray' on call site would be binded to  lowered one with a bit wrong type. But it's not clear would it cause any problem in further ir or bytecode generation.

max-kammerer: Seem it's very minor for now

max-kammerer:
Thank you!

PR #2340, opened 2019-05-15
Fix type parameters in WrappedClassConstructorDescriptor

sfs:
`IrConstructorCall` has type parameters from the constructed class as well as the constructor type parameters. This is currently not implemented in `WrappedClassConstructorDescriptor`, which breaks constructor calls with type arguments generated in an IR lowering pass.

Among other places, you can see the problem in the IR generated by fir2ir, which is why the tests have to be updated.


dnpetrov:
Looks good to me.

max-kammerer:
Thanks!

PR #2341, opened 2019-05-15
Fix test for static default methods in KotlinTypeMapper

sfs:
The function `isStaticMethod` makes its decision based on the kind and annotation, but doesn't check whether the function has a null dispatch receiver. Apparently this case never happens in the current backend, but I had to add this test for #2326.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
@@ -910,7 +910,8 @@ class KotlinTypeMapper @JvmOverloads constructor(
         val isConstructor = isConstructor(jvmSignature)
         val descriptor = getDefaultDescriptor(
             jvmSignature,
-            if (isStaticMethod(kind, functionDescriptor) || isConstructor) null else ownerType.descriptor,
+            if (isStaticMethod(kind, functionDescriptor) || isStaticDeclaration(functionDescriptor) || isConstructor)
+                null else ownerType.descriptor,
             functionDescriptor.unwrapFrontendVersion()
         )
 


compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -910,7 +910,8 @@ class KotlinTypeMapper @JvmOverloads constructor(
         val isConstructor = isConstructor(jvmSignature)
         val descriptor = getDefaultDescriptor(
             jvmSignature,
-            if (isStaticMethod(kind, functionDescriptor) || isConstructor) null else ownerType.descriptor,
+            if (isStaticMethod(kind, functionDescriptor) || isStaticDeclaration(functionDescriptor) || isConstructor)
```
max-kammerer: Please add `isIrBackend &&` here so code would be cleaned after switching to new typeMapper

sfs: Are you sure that this isn't a bug in the current TypeMapper? The function in question will produce the wrong result if you call it with a descriptor for a static function which is not at top-level (or in an inline class) and not marked with the `@JvmStatic` annotation.

Apparently the jvm backend does not currently run into this case, but it's possible that it will with future language extensions.

Checking for static declarations (`dispatchReceiver == null`) in addition to the other special cases seems more robust to me.

max-kammerer: Yes, you are right: it would cause some problems in future. Thanks!

udalov: FYI we've met some problems after bootstrapping on the compiler with this change: https://teamcity.jetbrains.com/viewLog.html?buildId=2319700&buildTypeId=Kotlin_dev_CompilerAndPlugin_191. I did not have the time yet to investigate, and the problem did not reproduce on a small sample. So meanwhile I've restored the old behavior for the non-IR backend here in 4e1effd0655eac4a971d5f23e94652d08e544774.

In general, let's please be more careful with such changes in the future. Even though we have a lot of codegen tests, they unfortunately don't cover everything, so even if the change itself is correct in theory (as in this case), interaction with code from some other dark corners of our codebase may lead to such errors. Specifically, I'd prefer if we didn't change the behavior of the old backend "speculatively", i.e. without understanding of the real cases where behavior changes.

sfs: That's humbling. I'll be more careful about touching the old backend in the future...

udalov: @sfs Thanks!

PR #2342, opened 2019-05-15
Move jvm specific phases to JvmLower

sfs:
Suggested by @udalov in this comment: https://github.com/JetBrains/kotlin/pull/2310#discussion_r283411966

This allows us to add prerequisites on backend specific phases.


gogabr:
I would prefer to have each phase object as close to its code as possible, to make source more self-contained.
But for now, it seems this solution works better.
Thank you!

PR #2343, opened 2019-05-15
JVM_IR: No EnclosingMethod attribute for member classes.

madsager:
The comment in the code is correct that EnclosingMethod
attributes should only be generated for local and
anonymous classes. We were generating them for member
classes as well which leads to invalid class files.

With this change I had to mute one more tests. That is
because we lose the parent method and therefore we
see a class as a member class instead of a local class.
With the old descriptor based check that test still
passes.


madsager:
@punzki This is what is leading to the invalid class files that you cannot dex.

@udalov @max-kammerer can you take a look? :)

max-kammerer:
Thanks!

PR #2345, opened 2019-05-15
Change ContractDescription to use KotlinBuiltIns as map keys,

juan-chen:
instead of ModuleDescriptor.

This change is to address memory issues due to ContractDescription holding large object trees.
See https://youtrack.jetbrains.com/issue/KT-31454


dsavvinov:
Thanks for the PR!

While it indeed mitigates the initial issue, the whole situation is a bit more complicated. We've discussed it internally, and have found another solution that involves more deep changes in the compiler code. So, we're closing this particular PR and going to implement proper change as soon as possible (here's the YouTrack issue to track progress: https://youtrack.jetbrains.com/issue/KT-31364)

Thanks again, you've really helped us to pin down the issue :)

PR #2348, opened 2019-05-17
tailrec: handle default parameters

ting-yuan:
More precisely, don't copy default parameters as they are always filled
by foo$default.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
@@ -36,9 +36,6 @@ val tailrecPhase = makeIrFilePhase(
 
 /**
  * This pass lowers tail recursion calls in `tailrec` functions.
- *
- * Note: it currently can't handle local functions and classes declared in default arguments.
- * See [deepCopyWithVariables].
  */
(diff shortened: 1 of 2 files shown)


ting-yuan:
@punzki Would you mind to proof-read the idea? I'm assuming that all references to functions with default parameters are replaced with their $default equivalents.

neetopia:
LGTM, default arguments are taken care of by jvmDefaultArgumentsPhase already, but I think another assumption here is that unspecified arguments are all default arguments.

ting-yuan:
Thanks, I also checked the implementation of genCall() in IrCallGenerator.kt and my assumption should be true. 

Also added back the assertion after discussed with @neetopia.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
```diff
@@ -130,31 +127,11 @@ private class BodyTransformer(
             +irSetVar(parameterToVariable[parameter]!!.symbol, argument)
         }
 
+        // Unspecified arguments do not exist; Calls to, e.g. foo, are replaced with foo$default.
```
max-kammerer: This piece of code is totally unclear: foo$default always calls foo function
With tailrec optimization foo$default call from foo should be avoided

PR #2349, opened 2019-05-17
KT-25465 "Redundant 'suspend' modifier" with suspend operator invoke

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25465


mglukhikh:
Merged 👍

PR #2351, opened 2019-05-20
KT-29731 Don't suggest `Add val/var to parameter` at expect class constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29731


dimonchik0036:
We already have a fix, but anyway, thanks for the attempt!

PR #2352, opened 2019-05-20
Remove imports of javafx from Gradle build files

pyos:
Added by this commit for some reason: https://github.com/JetBrains/kotlin/commit/ced973b7074f4207859d9709375f2bf28b3e2c55

### DIFF
--- a/compiler/daemon/daemon-common-new/build.gradle.kts
+++ b/compiler/daemon/daemon-common-new/build.gradle.kts
@@ -8,8 +8,6 @@
  * that can be found in the license/LICENSE.txt file.
  */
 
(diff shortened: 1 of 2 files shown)


compiler/daemon/daemon-common-new/build.gradle.kts
```diff
@@ -8,8 +8,6 @@
  * that can be found in the license/LICENSE.txt file.
  */
```
udalov: @Ololoshechkin I suppose copyrights in `build.gradle.kts` files added in your commits are also not needed

udalov:
Thanks!

PR #2353, opened 2019-05-20
KT-27822 Don't suggest `might be const` on `actual` member declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27822


dimonchik0036:
We already have a fix, but anyway, thanks for the attempt!

PR #2355, opened 2019-05-20
JVM_IR: generate non-null assertions for arguments

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -148,7 +148,15 @@ class JvmDeclarationFactory(
             createStaticFunctionWithReceivers(
                 defaultImpls, name, interfaceFun,
                 dispatchReceiverType = parent.defaultType,
-                origin = JvmLoweredDeclarationOrigin.DEFAULT_IMPLS
+                // If `interfaceFun` is not a real implementation, then we're generating stubs in a descendant
+                // interface's DefaultImpls. For example,
+                //
+                //     class I1 { fun f() { ... } }
```
gogabr: `interface I1`?

compiler/testData/codegen/bytecodeText/inlineClasses/noAssertionsForInlineClassesBasedOnNullableTypes.kt
```diff
@@ -1,4 +1,5 @@
 // !LANGUAGE: +InlineClasses
+// IGNORE_BACKEND: JVM_IR
```
gogabr: What happens here?

pyos: There should be no assertion because an inline class is unboxed into a parameter of type `Any?`, but the inline class PR is only a draft right now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -148,7 +148,15 @@ class JvmDeclarationFactory(
...
+                //     class I2 : I1
+                //
+                // is supposed to allow using `I2.DefaultImpls.f` as if it was inherited from `I1.DefaultImpls`.
+                // The classes are not actually related and `I2.DefaultImpls.f` is not a fake override but a bridge.
+                origin = interfaceFun.resolveFakeOverride()!!.origin
```
max-kammerer: Maybe we need new origin kind? Actually it would be source origin with new code?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: Yes, thanks

gogabr:
Thank you!

PR #2356, opened 2019-05-20
Use FQ names for string plus intrinsic in FlattenStringConcatenationLowering

sfs:
This makes the check more precise, in case there are future extensions to the standard library, as discussed [here](https://github.com/JetBrains/kotlin/pull/2244#discussion_r284272433).

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
@@ -16,7 +16,7 @@ import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrStringConcatenation
 import org.jetbrains.kotlin.ir.expressions.impl.IrStringConcatenationImpl
 import org.jetbrains.kotlin.ir.types.isStringClassType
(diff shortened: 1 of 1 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
```diff
@@ -68,6 +72,13 @@ val flattenStringConcatenationPhase = makeIrFilePhase(
 class FlattenStringConcatenationLowering(val context: CommonBackendContext) : FileLoweringPass, IrElementTransformerVoid() {
 
     companion object {
+        // There are two versions of String.plus in the library. One for nullable and one for non-nullable strings.
+        // The version for nullable strings has FqName kotlin.plus, the version for non-nullable strings
+        // is a member function of kotlin.String (with FqName kotlin.String.plus)
+        private val PARENT_NAMES = setOf(
+            KotlinBuiltIns.BUILT_INS_PACKAGE_FQ_NAME,
+            KotlinBuiltIns.BUILT_INS_PACKAGE_FQ_NAME.child(Name.identifier("String"))
```
udalov: `KotlinBuiltIns.FQ_NAMES.string.toSafe()`

sfs: Done.

udalov:
@sfs Thanks!

PR #2357, opened 2019-05-20
Fix type nullability check

sfs:
There is a bug in `TypeUtils.isNullableType` and its IrType analogue `IrType.isNullable`.

Both functions are supposed to check that a given type is not a subtype of `Any`. For type parameters the implementation checks whether there is *any* nullable upper bound. This is the wrong way around --- when we have  a type parameter with multiple upper bounds such as `T : Any, T : I?` both
functions return true, even though `T` is declared to be a subtype of `Any`.

Similarly `IrType.makeNotNull` does not work at all in the case of type parameters.

This PR fixes both problems.

Since type parameters with mutliple upper bounds are rare, neither bug causes any test failures whatsoever. I have added a test case where the bug will lead to a missing parameter null-check and a test for another situation involving recursive upper bounds.

Since `IrType.makeNotNull` is declared in ir.tree I had to move WrappedDescriptors and `IrType.isNullable` to this package. This implies that we need a corresponding change in kotlin-native.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
@@ -24,7 +24,7 @@ import org.jetbrains.kotlin.ir.expressions.impl.IrWhileLoopImpl
 import org.jetbrains.kotlin.ir.symbols.IrSymbol
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.functions
-import org.jetbrains.kotlin.ir.util.isNullable
+import org.jetbrains.kotlin.ir.types.isNullable
 import org.jetbrains.kotlin.name.Name
 
 /**
@@ -336,7 +336,7 @@ internal class HeaderProcessor(
     }
 
     private fun DeclarationIrBuilder.ensureNotNullable(expression: IrExpression) =
-        if (expression.type is IrSimpleType && expression.type.isNullable()) {
+        if (expression.type is IrSimpleType && expression.type.isNullable) {
             irImplicitCast(expression, expression.type.makeNotNull())
         } else {
             expression
--- a/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt
+++ b/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt
@@ -8,8 +8,12 @@ package org.jetbrains.kotlin.ir.types
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrTypeParameter
+import org.jetbrains.kotlin.ir.declarations.impl.IrTypeParameterImpl
+import org.jetbrains.kotlin.ir.descriptors.WrappedTypeParameterDescriptor
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.symbols.IrClassifierSymbol
+import org.jetbrains.kotlin.ir.symbols.IrTypeParameterSymbol
+import org.jetbrains.kotlin.ir.symbols.impl.IrTypeParameterSymbolImpl
 import org.jetbrains.kotlin.ir.types.impl.*
 import org.jetbrains.kotlin.types.*
 import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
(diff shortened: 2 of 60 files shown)


udalov:
Although the fix in `TypeUtils.isNullableType` looks correct to me, sadly this is a breaking change (existing code might start to throw exception or may result in a compilation error), which means we can't push it right away in the patch release. We need to get approval of the [Language Committee](https://kotlinlang.org/foundation/language-committee-guidelines.html) and probably also postpone the change itself until the next major release, which is 1.4 -- the latter part will depend on the exact decision of the committee.

Please create an [issue](https://kotl.in/issue) with the code from `multipleUpperBounds.kt` that starts to throw an exception. We'll then complete it with details relevant for the committee (this can take significant time because we need to identify the myriad of places where `TypeUtils.isNullableType` is used and specify how the change will affect that), discuss the plan internally, wait for the committee's decision and decide what exactly we'll do.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt
```diff
@@ -41,14 +45,40 @@ val IrType.classifierOrNull: IrClassifierSymbol?
...
+    if (!this.isNullable || this !is IrSimpleType) {
+        // This is wrong for IrDynamicType, but there's nothing we can do about it.
+        return this
+    }
+
+    val newClassifier = if (classifier is IrTypeParameterSymbol) {
+        val descriptor = WrappedTypeParameterDescriptor()
+        val symbol = IrTypeParameterSymbolImpl(descriptor)
+        val newParameter = with(classifier.owner as IrTypeParameter) {
+            IrTypeParameterImpl(startOffset, endOffset, origin, symbol, name, index, isReified, variance).also {
```
udalov: I'm not sure if this is the right approach. I'd prefer just always creating some sort of `IrNotNullTypeImpl` which is a subtype of `IrSimpleType` (or `IrSimpleTypeImpl`) and returns `true` on `isNullable`. Similarly to how it's done for `KotlinType` implementations (see [`KotlinTypeFactory.kt`](https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/types/KotlinTypeFactory.kt)). Creating fake IR (with fake symbols) is the one thing that we want to avoid where it's possible.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -336,7 +336,7 @@ internal class HeaderProcessor(
     }
 
     private fun DeclarationIrBuilder.ensureNotNullable(expression: IrExpression) =
-        if (expression.type is IrSimpleType && expression.type.isNullable()) {
+        if (expression.type is IrSimpleType && expression.type.isNullable) {
```
udalov: In the future, please extract refactorings that don't change behavior (function -> property conversion, renames, moves between packages) into separate commits. This simplifies review a lot.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt
sfs: Yes, that is a better approach and also fixes an issue in psi2ir (there's a workaround in the PR). I didn't consider it, because I didn't want to change the IR, but if you agree I would like to change the implementation in this manner.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
sfs: Will do! To be honest I only refactored the code since the PR was touching every file which used this utility function anyway.

sfs:
I've created an [issue](https://youtrack.jetbrains.com/issue/KT-31545?project=kt). Please let me know if I can help with the rest of this process.

For now I will make a separate PR fixing `IrType.makeNotNull`, since this should not be a breaking change.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypes.kt
udalov: @dnpetrov @romanart 
1) Do you have any objections?
2) Would this conflict with the ongoing branch on IrType subtyping very much? I suppose it'd make sense to wait until that branch is merged in that case

dnpetrov: I don't mind conflicts as soon as we have corresponding psi2ir tests.

udalov:
Since the decision on [KT-32343](https://youtrack.jetbrains.com/issue/KT-32343) (the backend part of [KT-31545](https://youtrack.jetbrains.com/issue/KT-31545)) is to change the behavior for progressive mode in 1.3.* and for production since 1.4, I think this change can be split to:
1) For IR backends: implement the new correct behavior unconditionally in `IrType.isNullable`.
2) For non-IR backends and frontend: add a new `LanguageFeature` with `sinceVersion` = 1.4, add `LanguageVersionSettings` to `TypeUtils.isNullableType` and use it to determine whether to use the old or new behavior. (This could be a bit challenging because `isNullableType` is sometimes invoked in static contexts where there's no easy way to obtain language version settings.)
3) Support a special "definitely non-null type" in `IrType.makeNotNull` for type parameters with nullable upper bounds.

From these, I wonder if we still need p.3? Because the exact implementation details would probably need more discussion and some time to converge, so maybe there's an easier way to handle this part in the JVM IR backend?

In any case, I suppose it's best to close this pull request since it's somewhat out of date. @sfs Please feel free to work on any (or none) of p.1 or p.2, we don't plan to do it ourselves in the nearest future. Thanks!

udalov:
Closing because the changes are pretty obsolete by now. The issue with `isNullable`/`isNullableType` is still relevant, but not so critical in terms of pushing JVM IR to production because behavior stays the same between JVM and JVM_IR right now.

PR #2358, opened 2019-05-20
Move tailrecPhase before defaultArgumentStubPhase

ting-yuan:
So that default values can be copied to the recursive call sites.

### DIFF
--- a/compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects.kt
+++ b/compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects.kt
@@ -1,0 +1,34 @@
+// DONT_RUN_GENERATED_CODE: JS
+// IGNORE_BACKEND: JVM
+
+var counter = 0
+fun inc() = counter++
+
+tailrec fun test(x: Int = 0, y: Int = inc(), z: Int = inc()) {
+    if (x * 2 != y || z - y != 1)
+        throw IllegalArgumentException("x=$x y=$y z=$z")
+
+    if (x < 100000)
+        test(x + 1)
+}
+
+fun box() : String {
+    test()
+
+    counter = 4
+    test(x = 1, y = 2, z = 3)
+
+    counter = 0
+    test(-1, -2, -1)
+
+    counter = 3
+    test(1, 2)
+
+    counter = 1
+    test(y = 0)
+
+    counter = 2
+    test(x = 1)
+
+    return "OK"
+}
--- a/compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects2.kt
+++ b/compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects2.kt
@@ -1,0 +1,22 @@
+// DONT_RUN_GENERATED_CODE: JS
+
+var counter = 0
+fun inc() = counter++
+
+tailrec fun test(x: Int = 0, y: Int = inc(), z: Int = inc()) {
+    if (x * 2 != y + 1 || z - y != -1)
+        throw IllegalArgumentException("x=$x y=$y z=$z")
+
+    if (x < 100000)
+        test(z = inc(), x = (counter - 1)/2 + 1)
+}
+
+fun box() : String {
+    counter = 4
+    test(1, 1, 0)
+
+    counter = 0
+    test(-1, -3, -4)
+
+    return "OK"
+}
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
@@ -112,6 +112,7 @@ val jvmPhases = namedIrFilePhase<JvmBackendContext>(
             propertiesPhase then
             renameFieldsPhase then
             annotationPhase then
+            tailrecPhase then
 
             defaultArgumentStubPhase then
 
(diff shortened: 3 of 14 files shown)


max-kammerer:
@ting-yuan I can't understand whole intention: `tailrec` optimization should eliminate tail calls, so it wouldn't any information for 'default' lower to insert any additional arguments

ting-yuan:
Actually, it's default lowering removed some information; The default value in the parameter IR is replaced with IrError (or something like that) after the default lowering. Moving tailrec before it makes the code, which the other PR tried to remove, works.

Since the removed default value (in the form of IR) is lowered into foo$bar, it's harder to recover the info than just moving tailrec to a place where it is still directly available. Does that make sense?

max-kammerer:
Yes,  it's correct. So then original pr would be corrected?

ting-yuan:
Do you mean: in the original PR, instead of deleting the copying logic, we should recover the default value initializer from the generated default stubs (foo$default)? There are some drawbacks:
1. The IR in foo$default may itself be transformed by other phases in the middle. The recovering process is inherently not reliable.
2. Compilation overhead.

Or, we may keep the initializer and try to mark it "don't generate", if possible. Still, moving tailrec earlier (this PR) is much simpler and subjectively cleaner.

max-kammerer:
Sorry for my complications, yes, please keep this pr it's easiest and rightest solution. Please also add mentioned test

ting-yuan:
@max-kammerer Please take a look at the newly added tests.

Interestingly, defaultArgsWithSideEffects.kt seemed to discover a bug in the old backend; Removing `tailrec` (and reduce the iterations) make it pass. It is disabled for the old backend at the moment and I'll report on youtrack later.

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects2.kt
```diff
@@ -0,0 +1,33 @@
...
+
+var counter = 0
+fun inc() = counter++
+
+tailrec fun test(x: Int = 0, y: Int = inc(), z: Int = inc()) {
+    if (x * 2 != y || z - y != 1)
+        throw IllegalArgumentException()
+
+    if (x < 100000)
+        test(x + 1, y + 2, z + 2)
```
max-kammerer: Please switch this one to  `test(x + 1,  z = counter + 1)`

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects3.kt
```diff
@@ -0,0 +1,33 @@
+// DONT_RUN_GENERATED_CODE: JS
```
max-kammerer: Seems this and next one tests are redundant

max-kammerer:
Actually it's a bug in TailRecursionCodegen.assignParameterValues - last default is initialized first

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects.kt
```diff
@@ -0,0 +1,34 @@
+// DONT_RUN_GENERATED_CODE: JS
+// IGNORE_BACKEND: JVM
+
+var counter = 0
+fun inc() = counter++
+
+tailrec fun test(x: Int = 0, y: Int = inc(), z: Int = inc()) {
+    if (x * 2 != y || z - y != 1)
+        throw IllegalArgumentException()
```
udalov: Please add something like `"x=$x y=$y z=$z"` to the exception message here and in other tests

max-kammerer:
Actually I filed issue https://youtrack.jetbrains.com/issue/KT-31541
and draft for LC: https://youtrack.jetbrains.com/issue/KT-31540

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects2.kt
max-kammerer: test(z = sideEffect, x = sideEffect())

ting-yuan: Did you mean `test(y = inc(), x = x + 1)`? The check expects that:
1. `counter` is approximately `2 * x`. So `x = inc()` makes the the test unhappy.
2. `y = inc()` happens before `z = inc()`. So we cannot write `z = sideEffect()` without specifying `y`.

I'm uploading `test(y = inc(), x = x + 1)` for now.

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects3.kt
ting-yuan: Done.

compiler/testData/codegen/box/diagnostics/functions/tailRecursion/defaultArgsWithSideEffects2.kt
max-kammerer: Actually I want to test argument initializion order in mixed case of named arguments and default values.
Please change this test to 
```
// DONT_RUN_GENERATED_CODE: JS

var counter = 0
fun inc() = counter++

tailrec fun test(x: Int = 0, y: Int = inc(), z: Int = inc()) {
    if (x * 2 != y + 1 || z - y != -1)
        throw IllegalArgumentException("x=$x y=$y z=$z")

    if (x < 100000)
        test(z = inc(), x = (counter - 1)/2 + 1)
}

fun box() : String {
    counter = 4
    test(1, 1, 0)

    counter = 0
    test(-1, -3, -4)

    return "OK"
}
```

ting-yuan: Done. Thanks for elaborating this!

max-kammerer:
Thank you!

PR #2359, opened 2019-05-21
Fix signature generation for calls to enum and inner class constructors

ting-yuan:
The synthesized arguments caused the size of default value mask off by
one when it is close to the boundary of Int.SIZE, which in turn
resulted in wrong signature at call sites.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
@@ -912,7 +912,15 @@ class KotlinTypeMapper @JvmOverloads constructor(
             jvmSignature,
             if (isStaticMethod(kind, functionDescriptor) || isStaticDeclaration(functionDescriptor) || isConstructor)
                 null else ownerType.descriptor,
-            functionDescriptor.unwrapFrontendVersion()
+            functionDescriptor.unwrapFrontendVersion(),
+            if (isIrBackend && isConstructor) {
+                val containingDeclaration = functionDescriptor.containingDeclaration
+                when {
+                    isEnumClass(containingDeclaration) -> 2
+                    (containingDeclaration as? ClassDescriptor)?.isInner == true -> 1
+                    else -> 0
+                }
+            } else 0
         )
 
         return Method(if (isConstructor) "<init>" else jvmSignature.name + JvmAbi.DEFAULT_PARAMS_IMPL_SUFFIX, descriptor)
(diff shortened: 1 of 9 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -1199,6 +1199,29 @@ class KotlinTypeMapper @JvmOverloads constructor(
...
+        dispatchReceiverDescriptor: String?,
+        callableDescriptor: CallableDescriptor
+    ): String {
+        val descriptor = method.descriptor
+        val extraArgsShift = when {
+            isIrBackend && isConstructor(method) && isEnumClass(callableDescriptor.containingDeclaration) -> 2
+            isIrBackend && isConstructor(method) && (callableDescriptor.containingDeclaration as? ClassDescriptor)?.isInner == true -> 1
+            else -> 0
+        }
+        val maskArgumentsCount = (callableDescriptor.valueParameters.size - extraArgsShift + Integer.SIZE - 1) / Integer.SIZE
```
ting-yuan: The real change is around `extraArgsShfit`; I have to move this function out of companion in order to use `isIrBackend`.

neetopia: > The real change is around `extraArgsShift`; I have to move this function out of companion in order to use `isIrBackend`.

Thanks for elaborating on this, I was wondering why change location.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -1199,6 +1199,29 @@ class KotlinTypeMapper @JvmOverloads constructor(
             ?: computeInternalName(classDescriptor, typeMappingConfiguration, isIrBackend)
     }
 
+    private fun getDefaultDescriptor(
```
max-kammerer: Could you keep old position of method and just pass neccesery parameters to it: actually there would be new TypeMapper for IR so all IR related logic would be removed from current one

max-kammerer:
It's looks strange that there are no any unmuted test for inner classes

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
ting-yuan: Will do.

ting-yuan:
I'm trying to find related test cases but chances are that they might not exist; There has to be exactly 31/63/... parameters for the inner class constructor. If there is none, I'll add one.

ting-yuan:
@max-kammerer There seems to be no test case for inner class constructor with 32 parameters and some default values so I created one in the PR.

compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt
```diff
@@ -908,10 +908,17 @@ class KotlinTypeMapper @JvmOverloads constructor(
...
+            isIrBackend && isConstructor && isEnumClass(containingDeclaration) -> 2
+            isIrBackend && isConstructor && (containingDeclaration as? ClassDescriptor)?.isInner == true -> 1
+            else -> 0
+        }
         val descriptor = getDefaultDescriptor(
             jvmSignature,
             if (isStaticMethod(kind, functionDescriptor) || isConstructor) null else ownerType.descriptor,
-            functionDescriptor.unwrapFrontendVersion()
+            functionDescriptor.unwrapFrontendVersion(),
+            extraArgsShift
```
max-kammerer: Please move new logic here:
```
       if (isIrBackend && isConstructor) {
                val containingDeclaration = functionDescriptor.containingDeclaration
                when {
                    isEnumClass(containingDeclaration) -> 2
                    (containingDeclaration as? ClassDescriptor)?.isInner == true -> 1
                    else -> 0
                }
            } else 0
```

ting-yuan: Done.

max-kammerer:
Thanks!

PR #2360, opened 2019-05-21
Improve constants analysis in KAPT

gavra0:
When analyzing source files in KAPT for incremental annotation processing,
avoid creating TreePath instances for initializers that are literals. E.g.
if it is just a boolean/int literal, we do not need to visit the tree nor
create expensive TreePath instance.

For classes that contain only constants (such as R.java for Android projects)
this is a significant improvement. Class that contains 16000 constants took
more than 10 seconds to analyze, now it is 70ms.

This is an improvement for https://youtrack.jetbrains.com/issue/KT-23880

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
@@ -47,7 +47,7 @@ private enum class Visibility {
     ABI, NON_ABI
 }
 
-private class TypeTreeVisitor(val elementUtils: Elements, val trees: Trees, val compilationUnit: CompilationUnitTree,  val sourceStructure: SourceFileStructure) :
+private class TypeTreeVisitor(val elementUtils: Elements, val trees: Trees, val compilationUnit: CompilationUnitTree, val sourceStructure: SourceFileStructure) :
     SimpleTreeVisitor<Void, Visibility>() {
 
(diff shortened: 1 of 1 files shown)


plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
```diff
@@ -196,12 +198,23 @@ private class TypeTreeVisitor(val elementUtils: Elements, val trees: Trees, val
...
+    Tree.Kind.STRING_LITERAL
+)
+
 /**
  * Visits a constant initializer expression, and extracts all references to constants, either through field select (A.MY_FIELD) or
  * identifier (MY_FIELD, which happens if A.MY_FIELD is statically imported).
  */
 private class ConstantTreeVisitor(val sourceStructure: SourceFileStructure) : TreePathScanner<Void, Void>() {
 
+    fun ifTreeIsLiteral(tree: Tree) = tree.kind !in literalKinds
```
yanex: Either the name or the function body looks counterintuitive to me :)
Maybe rename it to `isNonTrivialLiteralExpression()` or something similar?

gavra0: Thanks for spotting this. I've renamed it to `isTrivialLiteralExpression`, to make it more readable.

yanex:
Merged manually (master + 1.3.40). Thanks for the contribution!

PR #2361, opened 2019-05-21
Support isolating APs that report multiple originating elements

gavra0:
Support isolating APs that report multiple originating elements from the same source file

It is possible for isolating annotation processor to report two or more
originating elements from the same source file when generating
sources/classes/resources. This commit makes sure the source file are
de-duped, so assertion that there is a single source file does not fail.

Test: IsolatingIncrementalProcessorsTest.testIsolatingWithMultipleOriginatingElements

This is improvement to https://youtrack.jetbrains.com/issue/KT-23880


yanex:
Merged manually (master + 1.3.40). Thanks for the contribution!

PR #2362, opened 2019-05-22
Implement a lowering for type operators in the JVM_IR backend

sfs:
This PR moves some of the logic from ExpressionCodegen/codegenUtils to the IR level, where it can potentially be shared with [JS_IR](https://github.com/JetBrains/kotlin/blob/72f7287ad2897fd97da3f24d256076b3ced3ca75/compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/lower/TypeOperatorLowering.kt) and [native](https://github.com/JetBrains/kotlin-native/blob/4da64036c188e3e29dff00ed4dac071fb4412d34/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/lower/TypeOperatorLowering.kt) in the future.

I've also added two tests for the error messages around not-null assertions. They are currently disabled on JVM_IR, since the IMPLICIT_NOTNULL casts are missing. With these casts, "testErrorMessage" succeeds while "testStaticCallErrorMessage" fails, since the source offsets for a static call don't include the class name.

Finally, I've changed "testReifiedSafeAsWithMutable" to use the result of the safe cast - otherwise the test checks for suboptimal code and fails when using the lowering.

The first 5 commits in this PR are from #2339.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeOperatorLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeOperatorLowering.kt
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.jvm.lower
+
+import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.lower.IrBuildingTransformer
+import org.jetbrains.kotlin.backend.common.lower.at
+import org.jetbrains.kotlin.backend.common.lower.irNot
+import org.jetbrains.kotlin.backend.common.lower.irThrow
+import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.backend.jvm.codegen.fileParent
+import org.jetbrains.kotlin.backend.jvm.ir.erasedUpperBound
+import org.jetbrains.kotlin.config.ApiVersion
+import org.jetbrains.kotlin.ir.IrElement
+import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
+import org.jetbrains.kotlin.ir.builders.*
+import org.jetbrains.kotlin.ir.declarations.IrDeclaration
+import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.expressions.IrExpression
+import org.jetbrains.kotlin.ir.expressions.IrStatementOrigin
+import org.jetbrains.kotlin.ir.expressions.IrTypeOperator
+import org.jetbrains.kotlin.ir.expressions.IrTypeOperatorCall
+import org.jetbrains.kotlin.ir.expressions.impl.IrCompositeImpl
+import org.jetbrains.kotlin.ir.symbols.IrFunctionSymbol
+import org.jetbrains.kotlin.ir.symbols.IrTypeParameterSymbol
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.util.render
+import org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid
+import org.jetbrains.kotlin.ir.visitors.acceptChildrenVoid
+import org.jetbrains.kotlin.ir.visitors.acceptVoid
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+// After this pass runs there are only four kinds of IrTypeOperatorCalls left:
+//
+// - IMPLICIT_CAST
+// - SAFE_CAST with reified type parameters
+// - INSTANCEOF with non-nullable type operand or reified type parameters
+// - CAST with non-nullable argument, nullable type operand, or reified type parameters
+//
+// The latter two correspond to the instanceof/checkcast instructions on the JVM, except for
+// the presence of reified type parameters.
+internal val typeOperatorLowering = makeIrFilePhase(
+    ::TypeOperatorLowering,
+    name = "TypeOperatorLowering",
+    description = "Lower IrTypeOperatorCalls to (implicit) casts and instanceof checks"
+)
+
+private class TypeOperatorLowering(private val context: JvmBackendContext) : FileLoweringPass, IrBuildingTransformer(context) {
+    override fun lower(irFile: IrFile) = irFile.transformChildrenVoid()
+
+    private fun IrExpression.transformVoid() = transform(this@TypeOperatorLowering, null)
+
+    private val IrType.isReifiedTypeParameter: Boolean
+        get() = classifierOrNull?.safeAs<IrTypeParameterSymbol>()?.owner?.isReified == true
+
+    private fun lowerInstanceOf(argument: IrExpression, type: IrType) = with(builder) {
+        when {
+            type.isReifiedTypeParameter ->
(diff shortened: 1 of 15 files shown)


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
```diff
@@ -179,26 +182,44 @@ fun IrBuilderWithScope.irEqeqeq(arg1: IrExpression, arg2: IrExpression) =
...
+        IrStatementOrigin.OROR
+    )
+
+fun IrBuilderWithScope.irAndAnd(arg1: IrExpression, arg2: IrExpression) =
+    irIfThenElse(
+        context.irBuiltIns.booleanType,
+        arg1,
+        arg2,
+        irFalse(),
+        IrStatementOrigin.ANDAND
```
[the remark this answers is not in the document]
sfs: In this case I want to have short-circuit behavior. Is there any use case where we might want to disable it?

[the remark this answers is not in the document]
neetopia: My concern is if someone else will use it in future whether they will realize this will be short circuited. But since all ANDAND origin have short circuited behavior, it should be fine.

max-kammerer:
@sfs Please rebase on master

sfs:
@max-kammerer rebased to master

sfs:
Rebased to latest master and moved the `NOT_INSTANCEOF` statement origin to `JvmLoweredDeclarationOrigin`.

sfs:
Rebased on master again.

@dnpetrov Is there anything you'd like me to change with the PR?

sfs:
Rebased on master

sfs:
Rebased on master and incorporated 480313210a510a7d65b9d431b53bccd8f73db0dd and e207c963365af6d25e8ea77564eb2438f2d4e633 (#KT-22275)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
```diff
@@ -40,6 +40,9 @@ inline fun IrBuilderWithScope.irLetS(
     nameHint: String? = null,
     body: (IrValueSymbol) -> IrExpression
 ): IrExpression {
+    if (value is IrGetValue)
```
udalov: Let's remove `irLet` above to avoid different behavior for a similar-looking function (it's unused both in Kotlin and kotlin-native)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeOperatorLowering.kt
```diff
@@ -0,0 +1,194 @@
...
+                    endOffset = element.endOffset
+            }
+        })
+        return startOffset to endOffset
+    }
+
+    private fun sourceViewFor(declaration: IrDeclaration) =
+        context.psiSourceManager.getKtFile(declaration.fileParent)!!.viewProvider.contents
+
+    // Since Kotline 1.4 we throw NullPointerExceptions instead of more specialized exception classes.
```
udalov: typo in "Kotlin"

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeOperatorLowering.kt
sfs: Fixed

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
```diff
@@ -20,26 +20,15 @@ import org.jetbrains.kotlin.utils.addToStdlib.assertedCast
...
-}
-
 inline fun IrBuilderWithScope.irLetS(
     value: IrExpression,
     origin: IrStatementOrigin? = null,
     nameHint: String? = null,
     body: (IrValueSymbol) -> IrExpression
 ): IrExpression {
+    if (value is IrGetValue)
+        return body(value.symbol)
```
udalov: Looks like this change might not be correct in general, if the `value` is an access to a mutable variable which is changed inside the `body`. (Credit to @SvyatoslavScherbina for noticing.) Maybe we could use this logic in a separate method where it'd explicitly say that mutating the subject inside the body is not allowed (not sure how to name it though)?

sfs: Nice catch! Since this is only an optimization, I've just gotten rid of it for now.

It might be better to have a separate cleanup pass for the IR to get rid of unnecessary temporaries instead.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeOperatorLowering.kt
```diff
@@ -0,0 +1,194 @@
...
+            IrTypeOperator.IMPLICIT_INTEGER_COERCION -> {
+                irImplicitCast(expression.argument.transformVoid(), expression.typeOperand)
+            }
+
+            IrTypeOperator.CAST ->
+                lowerCast(expression.argument.transformVoid(), expression.typeOperand)
+
+            IrTypeOperator.SAFE_CAST -> {
+                irLetS(expression.argument.transformVoid(), IrStatementOrigin.SAFE_CALL) { valueSymbol ->
+                    irIfThenElse(
```
max-kammerer: Everything looks good, except minor impact of bytecode size of reified inline functions with 'as?' : +4 instuctions

sfs: Yes, previously we would generate a `reifiedOperationMarker` with a kind argument indicating a safe call. Upon inlining this would then be turned into the code we generate here. If you'd prefer, I could change it back to the old behavior?

sfs: Now that I think about it, this actually makes a difference when calling the code from Java.
```kotlin
inline fun <refified T>f(x: Any?) = x as? T
```
Right now, calling `f(null)` from Java fails with an exception. With these changes the code runs without exceptions. So it turns out that this isn't just a size issue, we actually need the old behavior for backwards compatibility.

udalov: @sfs Note that inline functions with reified parameters shouldn't be called from Java, since we generate them with ACC_SYNTHETIC in the current backend (and with ACC_PRIVATE before 1.3.40), and IIRC in JVM IR too. So the issue with the behavior inside the function body is only relevant for access via reflection, which should fail. In any case, I think it'd be nice to keep the same bytecode here as in the current backend

max-kammerer: Totally agree with @udalov. Please note that ACC_SYNTHETIC flag now is missed in IR

sfs: I've implemented the changes in question and added an additional test to check that we generate reified operation markers for safe casts to reified types.

max-kammerer:
@sfs Many thanks!

PR #2363, opened 2019-05-22
JVM_IR: generate more accessors for use in inline functions

pyos:
(+2 trivial fixes which I didn't bother separating into their own PRs.)

The new logic generates accessors for private/protected (including Java protected-package) entities used in inline functions if
 * the function is public;
 * the function is protected and the target is a superclass member (if the function is called from a subclass, invokespecial would do the wrong thing).

In addition, if the entity is protected-package and static, an accessor is inserted into the current class, not into the entity's owner (which is in Java and non-modifiable). This case is currently broken in the non-IR backend, by the way; see the new test, `inlineCallsStaticMethodFromOtherPackage.kt`. Since this access mode is deprecated, not sure if it's worth fixing.


compiler/testData/codegen/box/statics/inlineCallsStaticMethodFromOtherPackage.kt
```diff
@@ -0,0 +1,25 @@
+// TARGET_BACKEND: JVM
+// The non-IR backend attempts to call a non-existent accessor in class Test.
```
max-kammerer: @pyos Could you file an issue for this? With test name note

compiler/testData/codegen/boxInline/delegatedProperty/local.kt
```diff
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
```
max-kammerer: It's not clear why this test is affected, should it be merged with one of previous commits?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -110,6 +105,14 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
             else -> error("Unknown subclass of IrFunctionSymbol")
         }
 
+    // In case of Java `protected static`, access could be done from a public inline function in the same package,
+    // requiring an accessor, which we cannot add to the Java class.
+    private fun IrDeclarationWithVisibility.accessorParent() =
+        if (visibility == JavaVisibilities.PROTECTED_STATIC_VISIBILITY)
+            allScopes.findLast { it.irElement is IrClass }!!.irElement as IrClass
```
max-kammerer: maybe `currentClass`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -418,24 +429,25 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
-        if (!withSuper && !Visibilities.isPrivate(declaration.visibility) && declaration.visibility != Visibilities.PROTECTED) return true
+        // `internal` maps to public and requires no accessor.
+        if (!withSuper && !declaration.visibility.isPrivate && !declaration.visibility.isProtected) return true
 
         // If local variables are accessible by Kotlin rules, they also are by Java rules.
         val symbolDeclarationContainer = (declaration.parent as? IrDeclarationContainer) as? IrElement ?: return true
 
         // Within inline functions, we have to assume the worst.
-        if (inlinedLambdasCollector.isInlineNonpublicContext(allScopes))
+        val function = allScopes.lastOrNull { it.irElement is IrFunction }?.irElement as IrFunction?
```
max-kammerer: same with 'currentFunction'?

compiler/testData/codegen/boxInline/delegatedProperty/local.kt
pyos: It's affected because the delegated property's accessors are extracted by LocalDeclarationsLowering into private methods, which with this commit are called through public accessors as an accidental workaround. The lambda methods should probably be inlined or at least have the visibility of the function itself.

compiler/testData/codegen/box/statics/inlineCallsStaticMethodFromOtherPackage.kt
pyos: KT-31592

max-kammerer:
Thank you!

PR #2364, opened 2019-05-23
KT-31362 'Move variable declaration into `when`' quickfix comments left brace with EOL comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31362

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
@@ -11,13 +11,17 @@ import com.intellij.codeInspection.ProblemDescriptor
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiComment
 import com.intellij.psi.PsiElementVisitor
+import com.intellij.psi.PsiWhiteSpace
 import com.intellij.psi.SmartPsiElementPointer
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
```diff
@@ -96,6 +100,21 @@ private class VariableDeclarationIntoWhenFix(
         val property = descriptor.psiElement as? KtProperty ?: return
         val subjectExpression = subjectExpressionPointer.element ?: return
         val newElement = transform(property)?.copy() ?: return
+
+        val lastChild = newElement.lastChild
+        if (lastChild is PsiComment && lastChild.node.elementType == KtTokens.EOL_COMMENT) {
+            val leftBrace = subjectExpression.siblings(withItself = false).firstOrNull { it.node.elementType == KtTokens.LBRACE }
+            val whiteSpaceBeforeComment = lastChild.prevSibling?.takeIf { it is PsiWhiteSpace }
+            if (leftBrace != null) {
```
dimonchik0036:
```kotlin
fun foo(style: Int?): Int {
    val a = style ?: 0  // comment
    return when (a) { 0 -> 0
        else -> a
    }
}
```
Look at `CommentSaver`.

t-kameyama: Added test for above case.

dimonchik0036:
Thanks! 3bca1a50e37f00e2fb6cbdebcf3c418c3f81846c

PR #2365, opened 2019-05-23
KT-16476 Extend selection (Select Word) doesn't select just KDoc if cursor is just before the KDoc

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16476


goodwinnk:
Thank you! Merged as https://github.com/JetBrains/kotlin/commit/c442d69db6996d19973f69d0915feb3eb85ca7a6.

PR #2366, opened 2019-05-23
KT-31499 "Extend selection" selects escaped identifier name together with backticks

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31499


goodwinnk:
Thank you! Merged as https://github.com/JetBrains/kotlin/commit/8afa938723782a5b0726a4af14f31c31f4437b46.

PR #2367, opened 2019-05-23
Clarify the comments for withSign functions

majormajors:


ilya-g:
I believe that changing "sign bit" to "sign" might create confusion, because the sign of double is not the same as the sign bit of double.

majormajors:
The javadoc for `Math.copySign` doesn't mention the word "bit" and I think this method should be agnostic of underlying system used to represent the values.

I guess my question would be: What is the contract of this method? Is the expectation simply that it will make the sign bits match or that it will return a value of the same magnitude but with the sign of the parameter value?

ilya-g:
The confusion might stem from the presence of `sign` extension property in Kotlin: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.math/sign.html

For zero `0.0.sign` returns `0.0`, but what should be the expected result of `value.withSign(0.0).sign` if the documentation says that it copies the _sign_?

PR #2368, opened 2019-05-23
KAPT: do not discover incremental APs only if cache is not set

gavra0:
KaptFlag.INCREMENTAL_APT is used to indicate that KAPT run
should try to be incremental because changes to input files and
classpath have been detected. However, first run of the Gradle
task will set this flag to false, which means that all APs will be
detected as non-incremental in the first run. Further on, this means
that the dependency caches will be invalid, and all subsequent runs
will be non-incremental as well.

This commit uses existence of dependency cache to determine if incremental
APs should be discovered. The regression was introduced in
72fdc648ffcb598656c7c318a5fa6726e74d64bd.

Test: ./gradlew :kotlin-gradle-plugin-integration-tests:testAdvanceGradleVersion --tests *KaptIncrementalWith*\
 && ./gradlew :kotlin-gradle-plugin-integration-tests:test --tests *KaptIncrementalWith*


gavra0:
@yanex Should we close this PR, as it seems to be cherry-picked to 1.3.40 and master?

yanex:
Sure. Thanks!

PR #2369, opened 2019-05-23
JVM_IR: Optimize disjunction condition with intrinsic function call.

neetopia:
This is a follow up for [conjunction optimization](https://github.com/JetBrains/kotlin/pull/2283)

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
@@ -114,10 +114,30 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                         expression.startOffset,
                         expression.endOffset,
                         context.irBuiltIns.booleanType,
-                        context.irIntrinsics.andandSymbol
+                        context.irBuiltIns.andandSymbol
                     ).apply {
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -120,6 +120,26 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+                    }
+                    return IrCallImpl(
+                        expression.startOffset,
+                        expression.endOffset,
+                        context.irBuiltIns.booleanType,
+                        context.irIntrinsics.ororSymbol
+                    ).apply {
+                        dispatchReceiver = expression.branches[0].condition
+                        putValueArgument(0, expression.branches[1].result)
+                    }
```
ting-yuan: Although this call doesn't exist in bytecode, is there any reason preferring dispatchReceiver + arg0 over arg0 + arg1? That seems to be inconsistent with the parameters passed to `Key()`.

neetopia: It was intended to make it read like a.and(b), but I defined its symbol in the format of 2 value arguments, so should be changed to align with it. Was not captured before as Intrinsic does not go through expressionCodegen. I will change it to be 2 value arguments here, for andand as well.

max-kammerer:
Thank you!

PR #2370, opened 2019-05-24
psi2ir: keep the type of `when` when possible

ting-yuan:
Ideally, the type of `IrWhen` should be provided by type inference for
a consistent behavior. `USED_AS_EXPRESSION` from CFG isn't always
consistent with type inference, unfortunately.

The behavior is now aligned with `if`. The type of `when` is kept when
it *can* be an expression, instead of whether it is used or not.

### DIFF
--- a/compiler/testData/codegen/box/constants/constantsInWhen.kt
+++ b/compiler/testData/codegen/box/constants/constantsInWhen.kt
@@ -1,7 +1,3 @@
-// Even before any IR lowerings, the type of `when` is determined to be
-// Unit even though the outer `if` still returns `Int?`. This results
-// in a ClassCastException when that Unit is converted into a Number.
-// IGNORE_BACKEND: JVM_IR
 fun test(
         b: Boolean,
         i: Int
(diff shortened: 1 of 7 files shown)


compiler/testData/codegen/box/constants/constantsInWhen.kt
```diff
@@ -1,7 +1,6 @@
 // Even before any IR lowerings, the type of `when` is determined to be
 // Unit even though the outer `if` still returns `Int?`. This results
 // in a ClassCastException when that Unit is converted into a Number.
-// IGNORE_BACKEND: JVM_IR
```
udalov: The comment above is probably also not needed after this change?

ting-yuan: Oops. Thanks for catching that!

dnpetrov:
Please, add corresponding psi2ir tests (compiler/testData/ir/irText).

ting-yuan:
@dnpetrov Done.

BTW, it looks like that fir2ir works different than psi2ir; The inferred type is Unit. Although the IR tree is consistent and looks reasonable to me, I'm not sure whether this is expected. I just copied and pasted what is generated by fir2ir.

dnpetrov:
LGTM.
I'll take a look into fir2ir vs psi2ir differences here.

max-kammerer:
Thank you!

PR #2371, opened 2019-05-24
JVM_IR: sidestep defective getMethodAsmFlags when inlining lambdas

pyos:
It uses isStaticMethod to determine whether to set ACC_STATIC, which is not correct (see PR #2341). This results in using incorrectly typed opcodes (as all arguments are shifted by 1) when modifying the inlined lambda's bytecode. For example, in the test added by this commit, these opcodes are inserted to spill the stack into locals before calling another inline function. Without the patch, `ILOAD 0` for the lambda's `it` is considered to refer to a nonexistent `this`, which is spilled with an `ASTORE` instead of `ISTORE`.

Because getMethodAsmFlags is used by the non-IR backend (see PR #2341 again for why changing stuff might not be a good idea), the proposed solution is to ditch it completely and override generateLambdaBody in IrExpressionLambdaImpl to use FunctionCodegen's IR-based flag computation logic.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
@@ -38,7 +38,6 @@ class IrSourceCompilerForInline(
     private val data: BlockInfo
 ) : SourceCompilerForInline {
 
-
     //TODO
     override val lookupLocation: LookupLocation
         get() = NoLookupLocation.FROM_BACKEND
@@ -59,18 +58,23 @@ class IrSourceCompilerForInline(
     override val lazySourceMapper: DefaultSourceMapper
         get() = codegen.classCodegen.getOrCreateSourceMapper()
 
-    override fun generateLambdaBody(adapter: MethodVisitor, jvmMethodSignature: JvmMethodSignature, lambdaInfo: ExpressionLambda): SMAP {
-        lambdaInfo as? IrExpressionLambdaImpl ?: error("Expecting ir lambda, but $lambdaInfo")
-
-        val functionCodegen = object : FunctionCodegen(lambdaInfo.function, codegen.classCodegen, true) {
+    private fun makeInlineNode(function: IrFunction, classCodegen: ClassCodegen, marker: CallSiteMarker?): SMAPAndMethodNode {
+        var node: MethodNode? = null
+        val functionCodegen = object : FunctionCodegen(function, classCodegen, isInlineLambda = marker == null) {
             override fun createMethod(flags: Int, signature: JvmMethodGenericSignature): MethodVisitor {
-                return adapter
+                val asmMethod = signature.asmMethod
+                node = MethodNode(Opcodes.API_VERSION, flags, asmMethod.name, asmMethod.descriptor, signature.genericsSignature, null)
+                return wrapWithMaxLocalCalc(node!!)
             }
         }
+        lazySourceMapper.callSiteMarker = marker
         functionCodegen.generate()
-
-        return SMAP(codegen.classCodegen.getOrCreateSourceMapper().resultMappings)
+        lazySourceMapper.callSiteMarker = null
+        return SMAPAndMethodNode(node!!, SMAP(classCodegen.getOrCreateSourceMapper().resultMappings))
     }
+
+    override fun generateLambdaBody(lambdaInfo: ExpressionLambda): SMAPAndMethodNode =
+        makeInlineNode((lambdaInfo as IrExpressionLambdaImpl).function, codegen.classCodegen, null)
 
     override fun doCreateMethodNodeFromSource(
         callableDescriptor: FunctionDescriptor,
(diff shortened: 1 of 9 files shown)


compiler/testData/codegen/boxInline/simple/inlineCallInInlineLambda.kt
```diff
@@ -0,0 +1,8 @@
+inline fun f(g: (Int) -> Unit) = g(0)
```
max-kammerer: Please split test into two FILE sections (see nearby tests), it will also allow to test inliing against binaries

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -158,6 +162,23 @@ class IrExpressionLambdaImpl(
...
+        val functionCodegen = object : FunctionCodegen(function, sourceCompiler.classCodegen, true) {
+            override fun createMethod(flags: Int, signature: JvmMethodGenericSignature): MethodVisitor {
+                val asmMethod = signature.asmMethod
+                methodNode = MethodNode(Opcodes.API_VERSION, flags, asmMethod.name, asmMethod.descriptor, null, null)
+                methodAdapter = wrapWithMaxLocalCalc(methodNode)
+                return methodAdapter
+            }
+        }
+        functionCodegen.generate()
+        methodAdapter.visitMaxs(-1, -1)
```
max-kammerer: This should be perfromed in FunctionCodegen before visitEnd

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -158,6 +162,23 @@ class IrExpressionLambdaImpl(
     override val invokeMethodDescriptor: FunctionDescriptor = function.descriptor
 
     override val hasDispatchReceiver: Boolean = false
+
+    override fun generateLambdaBody(sourceCompiler: SourceCompilerForInline, reifiedTypeInliner: ReifiedTypeInliner) {
+        require(sourceCompiler is IrSourceCompilerForInline)
+        lateinit var methodNode: MethodNode
```
max-kammerer: Is it possible to move this logic and related one from ExpressionLambda to sourceCompiler? Seems all adapter related logic could be shared there

pyos: Thanks, haven't noticed the similar code in doCreateMethodNodeFromSource

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -59,19 +58,27 @@ class IrSourceCompilerForInline(
     override val lazySourceMapper: DefaultSourceMapper
         get() = codegen.classCodegen.getOrCreateSourceMapper()
 
-    override fun generateLambdaBody(adapter: MethodVisitor, jvmMethodSignature: JvmMethodSignature, lambdaInfo: ExpressionLambda): SMAP {
-        lambdaInfo as? IrExpressionLambdaImpl ?: error("Expecting ir lambda, but $lambdaInfo")
+    override fun generateLambdaBody(adapter: MethodVisitor, jvmMethodSignature: JvmMethodSignature, lambdaInfo: ExpressionLambda): SMAP =
+        throw AssertionError("unreachable; IrExpressionLambdaImpl overrides generateLambdaBody")
```
max-kammerer: This method still looks strange: base implementation is used from only one place (ExpressionLambda). So maybe switch adapter here to [Ir]ExpressionLambda and move adapter related logic from ExpressionLambda also to sourceCompilerForInline

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -99,31 +106,8 @@ class IrSourceCompilerForInline(
...
-
         assert(codegen.lastLineNumber >= 0)
-        lazySourceMapper.callSiteMarker = CallSiteMarker(codegen.lastLineNumber)
-        functionCodegen.generate()
-        lazySourceMapper.callSiteMarker = null
-        maxCalcAdapter!!.visitMaxs(-1, -1)
-        maxCalcAdapter!!.visitEnd()
-
-        return SMAPAndMethodNode(node!!, SMAP(fakeClassCodegen.getOrCreateSourceMapper().resultMappings))
+        return makeInlineNode(irFunction, FakeClassCodegen(irFunction, codegen.classCodegen), CallSiteMarker(codegen.lastLineNumber))
```
max-kammerer: FakeClassCodegen should only be used for inline function compilation not  lambda

max-kammerer: This will affect source further mapping support

pyos: It still is, generateLambdaBody passes the original ClassCodegen, not the fake one.

max-kammerer: Yes indeed, sorry

compiler/testData/codegen/boxInline/simple/inlineCallInInlineLambda.kt
```diff
@@ -0,0 +1,12 @@
+// FILE: 1.kt
+
```
max-kammerer: package test

max-kammerer: Please add package directive and corresponding import in second section

pyos: Done

max-kammerer:
Thank you!

PR #2372, opened 2019-05-25
JVM_IR: generate default constructor

ting-yuan:
Quoted from https://kotlinlang.org/docs/reference/classes.html

"On the JVM, if all of the parameters of the primary constructor have
 default values, the compiler will generate an additional parameterless
 constructor which will use the default values. This makes it easier to
 use Kotlin with libraries such as Jackson or JPA that create class
 instances through parameterless constructors."


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmDefaultConstructorLowering.kt
```diff
@@ -0,0 +1,60 @@
...
+        val primaryConstructor = irClass.constructors.firstOrNull { it.isPrimary } ?: return
+
+        if (!primaryConstructor.valueParameters.all { it.hasDefaultValue() })
+            return
+
+        // Skip if the default constructor is already defined by user.
+        if (irClass.constructors.any { it.valueParameters.isEmpty() })
+            return
+
+        irClass.addConstructor().apply {
```
max-kammerer: Seems this one would be shorter:
```
irClass.addConstructor().apply {
            val irBuilder = context.createIrBuilder(this.symbol, startOffset, endOffset)
            body = irBuilder.irBlockBody {
                +irDelegatingConstructorCall(primaryConstructor).apply {
                    passTypeArgumentsFrom(irClass)
                    passTypeArgumentsFrom(primaryConstructor, irClass.typeParameters.size)
                }
            }            
        }
```

ting-yuan: Done and tested. Thanks for the hints!

max-kammerer:
Thank you!

PR #2373, opened 2019-05-25
KT-9792 Don't propose the same name for arguments of lambda on completion of function call with lambda template

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-9792


goodwinnk:
Great! Thank you!

Merged as https://github.com/JetBrains/kotlin/commit/3a77b63c8588f437cfba628cb07ef2ce7b5a0cab.

PR #2374, opened 2019-05-27
build: add kotlin-compiler-embeddable to BOM

gastaldi:
Added missing library in BOM

### DIFF
--- a/libraries/tools/kotlin-bom/pom.xml
+++ b/libraries/tools/kotlin-bom/pom.xml
@@ -167,6 +167,11 @@
             </dependency>
             <dependency>
                 <groupId>${project.groupId}</groupId>
+                <artifactId>kotlin-compiler-embeddable</artifactId>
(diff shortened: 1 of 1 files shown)


ilya-g:
Could you elaborate on why do you need this dependency listed here?

gastaldi:
This library is used in a project (Quarkus). The BOM is missing other dependencies too

ilya-g:
> The BOM is missing other dependencies too

What other dependencies do you mean?

gastaldi:
I don't have a complete list but I miss `org.jetbrains.kotlin:kotlin-maven-allopen` too, which is used as a dependency to the Maven plugin :smile:

ilya-g:
AFAIK dependency versions listed in the `dependencyManagement` section do not affect neither the plugin versions nor dependencies of plugins.

gastaldi:
Cool, then I think this is the only one missing

PR #2375, opened 2019-05-27
JVM_IR: Enable loadJava tests.

madsager:
I'm enabling these tests in preparation for changing inner class
attribute emission in connection with annotations. These tests
are the only ones that fail if you get it wrong with the current
backend.


compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/ir/javac/AbstractIrLoadJavaUsingJavacTest.kt
```diff
@@ -0,0 +1,37 @@
...
+package org.jetbrains.kotlin.jvm.compiler.ir.javac
+
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.jvm.compiler.ir.AbstractIrLoadJavaTest
+import org.jetbrains.kotlin.test.KotlinTestUtils
+import java.io.File
+
+abstract class AbstractIrLoadJavaUsingJavacTest : AbstractIrLoadJavaTest() {
```
max-kammerer: Why not extend AbstractLoadJavaUsingJavacTest?

madsager: Good point. Done!

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/ir/javac/AbstractIrLoadJavaUsingJavacTest.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.jvm.compiler.ir.javac
+
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.jvm.compiler.javac.AbstractLoadJavaUsingJavacTest
+
+abstract class AbstractIrLoadJavaUsingJavacTest : AbstractLoadJavaUsingJavacTest() {
```
udalov: I think we actually don't need this test for now. I'd prefer if we only had `AbstractIrLoadJavaTest`.

The reason is that we aren't working on javac integration (subsystem that uses javac internals to load source & binary Java classes, instead of intellij-core) for a long time already, and don't plan to in the near future. The existing `AbstractLoadJavaUsingJavacTest` is being run only to verify that we don't break what's already supported. It may well be the case that when we get to working on the javac integration again, the IR backend will already be in production, and all the non-IR tests will be removed :)

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/ir/AbstractIrLoadJavaTest.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.jvm.compiler.ir
+
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.jvm.compiler.AbstractLoadJavaTest
+
+abstract class AbstractIrLoadJavaTest : AbstractLoadJavaTest() {
```
udalov: Let's inherit from `AbstractLoadJavaWithFastClassReadingTest` instead. This might be a bit confusing, but "fast class reading" (loading classes directly from class files by ASM) is used in the _compiler_. The other class reading mode (loading classes by querying `PsiClass` constructed by IntelliJ) is used in the _IDE_ and is tested in `AbstractLoadJavaTest`.

For the IR backend scenario, I think it makes more sense to test the compiler behavior.

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/ir/javac/AbstractIrLoadJavaUsingJavacTest.kt
madsager: Thanks for the context Alex. That makes perfect sense; I wasn't actually sure what the difference was between these. :)

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/ir/AbstractIrLoadJavaTest.kt
madsager: Done. Thanks again for the context, that sounds right to me as well.

udalov:
Thanks! Merged manually in 5be0bd4a56b.

PR #2377, opened 2019-05-27
KT-31639: Fix Iterables.drop integer overflow

EdwarDDay:
https://youtrack.jetbrains.com/issue/KT-31639

### DIFF
--- a/libraries/stdlib/test/collections/CollectionTest.kt
+++ b/libraries/stdlib/test/collections/CollectionTest.kt
@@ -521,6 +521,7 @@ class CollectionTest {
         val coll = listOf("foo", "bar", "abc")
         assertEquals(listOf("bar", "abc"), coll.drop(1))
         assertEquals(listOf("abc"), coll.drop(2))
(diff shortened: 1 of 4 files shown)


libraries/stdlib/test/collections/CollectionTest.kt
```diff
@@ -521,6 +521,7 @@ class CollectionTest {
         val coll = listOf("foo", "bar", "abc")
         assertEquals(listOf("bar", "abc"), coll.drop(1))
         assertEquals(listOf("abc"), coll.drop(2))
+        assertEquals(listOf(2147483647L, 2147483648), (0L..2147483648).drop(2147483647))
```
ilya-g: - I expect this test running for some noticeable time, so better to place it in stdlib/jvm/testLongRunning/collections/IndexOverflowJVMTest.kt
- I believe we can express these numbers with `Int.MAX_VALUE` constant

EdwarDDay: good points, I changed it

ilya-g:
Thanks, I've merged it manually with the commit c885cadde434debcc98e82b23d19b9e201fbed73

PR #2380, opened 2019-05-28
JVM_IR: Restore lost top-level declaration check in annotation gen.

madsager:
When moving away from descriptors, this top-level declaration check
was lost. That leads to many useless annotation elements in
inner classes attributes for annotations that are not inner classes.


madsager:
Tests that check that we still have the right inner classes information are in: https://github.com/JetBrains/kotlin/pull/2375

I haven't found a good standalone way to check that the annotations are not there. There are bytecode listing tests that will show this, but they are all failing for different reasons right now, so I'm not enabling those here.

madsager:
@gogabr I think this top-level check was lost in the removal of descriptors from this code.

gogabr:
You are right, I see this in the corresponding commit
(bfe148efd588c01b82fdbdf4c0c935e536923297)

-    override fun addInnerClassInfoFromAnnotation(irClass: IrClass) {
-        var current: DeclarationDescriptor? = irClass.descriptor
-        while (current != null && !isTopLevelDeclaration(current)) {
-            if (current is ClassDescriptor) {
+    override fun addInnerClassInfoFromAnnotation(innerClass: IrClass) {
+        var current: IrDeclaration? = innerClass
+        while (current != null) {
+            if (current is IrClass) {

My mistake, sorry!

gogabr:
Thanks!

PR #2381, opened 2019-05-28
KT-31127: add only generated Java sources to JavaCompile inputs

gavra0:
This commit changes how generated sources are added to the Gradle JavaCompile
task. Previously, directory was added to sources which caused issues with
annotation processors that only generate Kotlin sources. I.e. JavaCompile task
was executed but no java sources existed.

Now, only generated Java sources are added to the JavaCompile task. This
means that if only Kotlin sources are generated JavaCompile task will be
skipped.

Fixes: https://youtrack.jetbrains.com/issue/KT-31127
Test: Kapt3IT.testKotlinProcessorUsingFiler

### DIFF
--- a/libraries/examples/annotation-processor-example/src/main/kotlin/example/KotlinFilerGeneratingProcessor.kt
+++ b/libraries/examples/annotation-processor-example/src/main/kotlin/example/KotlinFilerGeneratingProcessor.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package example
+
+import javax.annotation.processing.AbstractProcessor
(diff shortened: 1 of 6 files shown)


yanex:
Merged into master and 1.3.40. Thanks for the contribution!

technoir42:
@gavra0 @yanex I believe there is a separate code path for Android projects that needs a similar fix.
Might be this one https://github.com/JetBrains/kotlin/blob/71bd88bdb3faebb266c58c9b99669bd91175fb72/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/android/Android25ProjectHandler.kt#L193-L195

Right now with 1.3.40-eap-105 if I disable BuildConfig generation in pure Kotlin `com.android.library` module I get the same error saying:
```
Execution failed for task ':compileDebugJavaWithJavac'.
> no source files
```

gavra0:
Thanks for spotting this @technoir42. I've uploaded https://github.com/JetBrains/kotlin/pull/2410 to fix this issue.

PR #2382, opened 2019-05-29
KT-31668 Complete statement for class declaration: add '()' to supertype

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31668

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/editor/KotlinSmartEnterHandler.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/editor/KotlinSmartEnterHandler.kt
@@ -19,6 +19,7 @@ import org.jetbrains.kotlin.idea.editor.fixers.*
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
 
 class KotlinSmartEnterHandler : SmartEnterProcessorWithFixers() {
     init {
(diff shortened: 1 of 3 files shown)


goodwinnk:
There is a problem with the fix that it introduces an new resolve call at the UI thread and that might cause freezes. On the other hand implemented behavior is very expected and desired. Maybe we can avoid resolve by looking to indexes only?

t-kameyama:
I think that looking to indexes only is not enough. I want to check that super type is not interface.

vladimirdolzhenko:
Thanks a lot @t-kameyama for this PR and sorry for the long delay in response.

I have several concerns and would like to discuss them:

### Freeze 
 As @goodwinnk mentioned earlier this change introduces extra resolve on UI thread that is dangerous (potential freeze) 
Potential solution could be to resolve parent candidate in a background thread with a progress bar and execute it synchronously to UI actions (using `ProgressManager.getInstance().runProcessWithProgressSynchronously(...)`) 
- it protects from UI freeze
- it would not show modal progress dialog in case of fast (less than [300 ms](https://github.com/JetBrains/intellij-community/blob/2e9c58f56b507fedf8d650dc1e5d6e39300e54ec/platform/platform-impl/src/com/intellij/openapi/progress/util/ProgressWindow.java#L41)) resolve (it could cover more than 50% cases)
- but I assume we need to cancel action in case of caret change (user has no patience to wait and can fix it by fix own) 

### Parent class with non empty parameters ctor
Let's say there is 
`open class A(val v: Int)` 
and we would like to extend it with class B:
`class B: A<caret>` 
in case of _complete current statement_ action I think it is reasonable to get smth like
```
class B: A(<caret>) {
}
```

In case of empty parent ctor we can put caret right between curly brackets: ```class B: A() {<caret>
}```.

t-kameyama:
@vladimirdolzhenko please review

vladimirdolzhenko:
Thanks a lot for the changes

I'm so sorry I have not realised that it is happened under write action and therefore modal dialog could not be shown and it is executed synchronously (to prove that you can add `Thread.sleep(5000)` within a task, lets say before resolve).

It looks like smart enter completion requires platform changes to support this.

vladimirdolzhenko:
Sorry for a long pause. While it is a potential cause of UI freeze it is worth to merge this request (after talks and consideration).

UI Freeze is forced by a platform and it has to be addressed with https://youtrack.jetbrains.com/issue/IDEA-233579 and it should not block this PR.

PR is merged manually, Thanks a lot for your contribution.

PR #2383, opened 2019-05-30
KT-30775 Inspection for the case when one lateinit var overrides another lateinit var

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30775

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -3384,6 +3384,15 @@
                      language="kotlin"
     />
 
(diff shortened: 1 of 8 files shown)


mglukhikh:
Thank you, I think it can be accepted after testing. See also: https://youtrack.jetbrains.com/issue/KT-30775#focus=streamItem-27-3520818.0-0

mglukhikh:
For some reason, this inspection works slowly and consumes a lot of memory. Will try to investigate it a little bit.

mglukhikh:
Refactored a bit and merged manually

PR #2386, opened 2019-05-31
KT-31714: Close files after analysis in incremental KAPT

gavra0:
Artifact transform used to analyze classpath was not closing files, which
caused java.nio.file.FileSystemException. This commit fixes that, and all
input streams are closed after use.

Fixes: https://youtrack.jetbrains.com/issue/KT-31714


yanex:
Merged into master and 1.3.40. Thanks for the contribution!

PR #2387, opened 2019-05-31
Moved the common part of klib infrastructure to util-klib

alexander-gorshenev:


compiler/util-klib/src/org/jetbrains/kotlin/library/KonanLibraryVersioning.kt
```diff
@@ -0,0 +1,29 @@
+package org.jetbrains.kotlin.library
```
udalov: @alexander-gorshenev Please add copyright to all new files

settings.gradle
```diff
@@ -22,6 +22,8 @@ pluginManagement {
 include ":kotlin-build-common",
         ":compiler",
         ":compiler:util",
+        ":kotlin-util-io",
+        ":kotlin-util-klib",
```
udalov: I'd prefer if these projects were named `:compiler:util-io` and `:compiler:util-klib` which is more similar to `:compiler:util` and reflects directory structure, which means less guesswork needed when trying to do anything with our build

PR #2388, opened 2019-06-03
JVM_IR: Allow delegated companion objects.

madsager:
This currently leads to a compilation error due to metadata
generation.


madsager:
@udalov I noticed this while enabling the loadJava tests. With this change, the metadata generated for the test case is identical to what it is for the old backend.

udalov:
Looks good, thank you!

PR #2389, opened 2019-06-03
JVM_IR: Maintain annotations in SAM conversion.

madsager:
Copy method and parameter annotations to the generated SAM
implementation method.


madsager:
@pyos, could you have a look to double check that this makes sense?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -180,7 +189,7 @@ class SingleAbstractMethodLowering(val context: CommonBackendContext) : ClassLow
             // common, as `Interface { something }` is a local function + a SAM-conversion of a reference
             // to it into an implementation.
             private fun createFunctionProxy(superType: IrType, reference: IrFunctionReference): IrClass =
-                implement(superType, reference.arguments.mapNotNull { it?.type }) { fields ->
+                implement(superType, reference.arguments.mapNotNull { it?.type }, reference.symbol.owner) { fields ->
```
pyos: This would also copy annotations from non-lambdas (e.g. in `f(::g)`, `f` would see `g`'s annotations), and the non-IR backend does not do that.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -137,16 +144,18 @@ class SingleAbstractMethodLowering(val context: CommonBackendContext) : ClassLow
 
             // Construct a class that wraps an invokable object into an implementation of an interface:
             //     class sam$n(private val invokable: F) : Interface { override fun method(...) = invokable(...) }
-            private fun createObjectProxy(superType: IrType, invokableType: IrType): IrClass =
-                implement(superType, listOf(invokableType)) { fields ->
+            private fun createObjectProxy(superType: IrType, invokableType: IrType): IrClass {
+                val invokableClass = invokableType.classifierOrFail.owner as IrClass
+                val function = invokableClass.functions.single { it.name == OperatorNameConventions.INVOKE }
+                return implement(superType, listOf(invokableType), function) { fields ->
```
pyos: This probably has no effect at all. CallableReferenceLowering would not add any annotations to invoke().

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
madsager: Thanks! Yes, extending the tests show that this is indeed the case. I'll have a look if I can replicate the behavior of the old backend here.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
madsager: Ack. You are right, that doesn't make sense.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
madsager: Added more tests and implemented this in `createFunctionProxyInstance` for lambdas and anonymous functions only. That seems to match the behavior of the old backend.

@pyos, can you have another look?

compiler/testData/codegen/box/annotations/annotatedLambda/samFunExpression.kt
```diff
@@ -35,15 +34,28 @@ fun testMethod(method: Method, name: String) {
...
+}
+
+@Ann("OK")
+fun annotatedFunction(@Ann("OK0") x: String, @Ann("OK1") y: String) {
+}
 
+fun box(): String {
+    testClass(Test.apply(@Ann("OK") fun(){}), "1", ::testMethod)
+    testClass(Test.applyABC(@Ann("OK") fun(@Ann("OK0") x: String, @Ann("OK1") y: String){}), "2", ::testMethod)
+    testClass(Test.applyABC(::annotatedFunction), "3", ::testMethodNoAnnotations)
```
pyos: Maybe it's best to add this as a separate test instead?

max-kammerer: @madsager Please move to separate test

madsager: Done, thanks!

max-kammerer:
Thank you!

PR #2390, opened 2019-06-03
Decrement "last" for last-exclusive progressions.

punzki:
Make all progression headers inclusive, and decrement last for last-exclusive progressions (i.e., `until` progressions and loop over array indices). Also invert the while loop (into if + do-while) that is used when lowering for-loops over progressions that cannot overflow.

Results from running the benchmarks show no regressions under JVM IR. Also for native, the performance is on par with the `ForLoopsLowering` in native, except for `step` progressions which will be covered in subsequent changes.


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -138,7 +153,7 @@ internal class ProgressionLoopHeader(
 
     override fun buildLoop(builder: DeclarationIrBuilder, oldLoop: IrLoop, newBody: IrExpression?): LoopReplacement {
         with(builder) {
-            var (newLoop, replacementExpression) = if (headerInfo.canOverflow) {
+            return if (headerInfo.canOverflow) {
```
sfs: Nit, but could you use an expression body here?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -149,44 +164,38 @@ internal class ProgressionLoopHeader(
...
-                    putValueArgument(1, irGet(last))
-                })
                 val newLoop = IrDoWhileLoopImpl(oldLoop.startOffset, oldLoop.endOffset, oldLoop.type, oldLoop.origin).apply {
                     label = oldLoop.label
-                    condition = newCondition
+                    condition = irNotEquals(irGet(loopVariable!!), irGet(last))
                     body = newBody
                 }
-                val notEmptyCheck = irIfThen(buildLoopCondition(builder), newLoop)
+                val notEmptyCheck = irIfThen(buildLoopCondition(this@with), newLoop)
```
sfs: Why can you ignore the additional not empty condition here?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -149,44 +164,38 @@ internal class ProgressionLoopHeader(
...
+                //     do {
+                //         val loopVar = inductionVar
+                //         inductionVar += step
+                //         // Loop body
+                //     } while (inductionVar <= last)
                 //   }
-                val newLoop = IrWhileLoopImpl(oldLoop.startOffset, oldLoop.endOffset, oldLoop.type, oldLoop.origin).apply {
+                //
+                // Even though this can be simplified into a simpler while loop, using if + do-while (i.e., doing a loop inversion)
+                // performs better in benchmarks. In cases where `last` is a constant, the `if` may be optimized away.
```
sfs: This is a mystery.

Do you have any idea why this is faster? It shouldn't be because of branching - the loop inversion in question would be performed anyway (without code duplication) by, e.g., MachineBlockPlacement in llvm. Is it because we don't remove empty loops, but do remove empty conditionals? In either case, it seems like this is a performance problem stemming from somewhere else and should be fixed there.

This is especially strange since you only perform a loop inversion for progressions, not for array loops.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -82,18 +84,148 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
...
+        //   infix fun Long.until(to: Int): LongRange
+        //   infix fun Long.until(to: Long): LongRange
+        //
+        // The combinations where the range element type is strictly larger than the argument type do NOT need the additional condition.
+        // In such combinations, there is no possibility of underflow when the argument (casted to the range element type) is decremented.
+        // For unexpected combinations that currently don't exist (e.g., Int until Char), we assume the check is needed to be safe.
+        // TODO: Include unsigned types
+        return with(context.irBuiltIns) {
+            when (receiverKotlinType) {
+                char -> true
```
sfs: Why not use the IrTypes directly and compare against charType, byteType, etc.?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.types.getClass
+import org.jetbrains.kotlin.ir.types.toKotlinType
+import org.jetbrains.kotlin.ir.util.functions
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.util.OperatorNameConventions
+
+internal fun IrExpression.castIfNecessary(targetType: IrType, numberCastFunctionName: Name): IrExpression {
+    return if (type.toKotlinType() == targetType.toKotlinType()) {
```
sfs: Why not `type == targetType`?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+                dispatchReceiver = this@negate
+            }
+        }
+    }
+}
+
+/** Return `this - 1` if the expression is const, otherwise call dec(). */
+internal fun IrExpression.decrement(): IrExpression {
+    val thisValue = (this as? IrConst<*>)?.value
+    return when (thisValue) {
```
pyos: Are byte & short intentionally not handled here? They may appear from const field reads in ConstLowering.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
```diff
@@ -44,22 +44,24 @@ val forLoopsPhase = makeIrFilePhase(
...
- *       inductionVar++  // (`inductionVar--` if the progression is decreasing)
- *       // Loop body
+ *   val last = B
+ *   if (inductionVar <= last) {  // (`inductionVar >= last` if the progression is decreasing)
+ *       // Loop is not empty
+ *       do {
+ *           val loopVar = inductionVar
+ *           inductionVar++  // (`inductionVar--` if the progression is decreasing)
+ *           // Loop body
+ *       } while (inductionVar <= last)
```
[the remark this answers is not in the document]
punzki: That was in fact something I tried but I saw a significant difference in the `downTo` loops and I am not sure why; somehow JIT compiler is not able to optimize it as well with the `do {...} while (loopVar != last)` form.

To address your comment about why the loop inversion is required: This ensures that lowered for-loops while keep the same performance characteristics when we migrate Kotlin/Native to use the shared `ForLoopsLowering`. K/N is slightly worse *without* the inversion, so it was important to do this for that reason. JVM is fine with either, performance-wise, and a very small impact on code size: There is an additional `if`, but in cases where the upper bound is constant, the surrounding `if` can be optimized away and there is no unconditional `GOTO` at the bottom of the loop.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
punzki: Unfortunately I don't have a good idea of why this is so. In any case, one of my goals is to converge on a single shared `ForLoopsLowering`, and for this I have to track the current `ForLoopsLowering` in K/N as closely as possible, which does have do the inversion at this stage. And K/N also does the inversion for progressions but not array loops. ¯\_(ツ)_/¯ My top priority is to ensure that when we switch over K/N to the shared version, that we are very confident that there aren't any performance regressions.

I am inclined to converge on the lowering first, and investigate this mystery later. Perhaps there is also a shortcoming in the existing benchmarks.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
punzki: Actually I should also use it here, but since it is only currently used in `until` progressions which go through the "cannot overflow" code path then we didn't expose this error. I'll updates this. Good catch!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
punzki: My recollection could be wrong, but when I started working on for-loops, I was not able to compare IrTypes directly. Anyway, updated to use IrTypes, and also updated elsewhere in the code where I was using KotlinType.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
punzki: Updated

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
punzki: These are intentionally not handled here. `decrement()` is only called on expressions which are guaranteed to be one of the 3 progression element types (Int, Long, Char). If you look at UntilHandler, it is called on the argument to `until` *after* at has been casted to the progression element type.

max-kammerer:
@punzki Thanks! PR was merged manually with bytecode text combine via https://github.com/JetBrains/kotlin/commit/496765f41e26ea05a203a92e3ac39fea08e6604e

PR #2392, opened 2019-06-04
Extension Points Discussion

jimgoog:
This PR is NOT intended for merging, and merely serves to seed discussions about the extension points as currently used in Compose.  Several of these extension points are very rough and would require substantial discussion/cleanup before they could get merged into upstream.

I broke the change into two commits.  The first contains all the extension points as they currently stand on our end.  The second attempts to remove anything that looked related to the KTX syntax which is being removed in favor of the Function Call Syntax.  I attempted to be relatively careful, but it is possible the second commit may have been a bit overzealous, so the first commit can be a useful reference for understanding where/why changes were made.


jimgoog:
Closing, since we don't actually intend to merge.  Just created the PR to have a place for discussion.

PR #2393, opened 2019-06-04
KT-31645 Fix yarn downloading on Windows

ilgonmic:
https://youtrack.jetbrains.com/issue/KT-31645

According to docs (https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileCopyDetails.html#getPath--)  FileCopyDetails uses '/' as separator regardless of platform, on Windows standard separator is '\'
Addditionally, it returns the path relative to the root of destination
So separator is unnecessary to remove


snrostov:
Thanks!

PR #2395, opened 2019-06-05
Enable debugger file ranking tests for JVM_IR

neetopia:
11 out of 13 existing tests are passing, 2 failing tests disabled for now.


idea/testData/debugger/fileRanking/lambdas.kt
```diff
@@ -33,4 +33,6 @@ class A {
             block { block { val x = 4 }}
         }
     }
-}
\ No newline at end of file
+}
+
+// IGNORE_BACKEND: JVM_IR
```
ting-yuan: I usually see this in the top of the test files. Is there any consideration putting it here?

neetopia: There will be line number check for some of the test cases, although such is not present in this test case, just make it a convention to reduce future impact on line numbers.

yanex:
I think it would make sense to update also bunch files (`GenerateTests.kt.183` and others) to minimize differences between them.

neetopia:
> I think it would make sense to update also bunch files (`GenerateTests.kt.183` and others) to minimize differences between them.

@yanex Thanks for pointing out, done. Actually that file is the only file I found with usage of FileRank tests.

Do you know what is the purpose of such files? I am also adding other tests in other PRs, just want to make sure I didn't miss anything.

yanex:
@neetopia We build plugins for multiple IDE versions (several versions of IDE + several versions of Android Studio). So we have to manage changes in the API somehow. Before we used separate branches and rebased them from time to time, but that turned out to be exhausting and error-prone.
So we made an utility called [bunch](https://github.com/JetBrains/bunches) that allows to keep several versions of files in the same branch.

yanex:
Also, bunch is mentioned in the Kotlin readme. There are instructions about how to switch to some different branch. https://github.com/JetBrains/kotlin#building-for-different-versions-of-intellij-idea-and-android-studio

neetopia:
I see, thanks for elaborating.. I've updated .183 as well, please take a look.

yanex:
Hello @neetopia! Sorry for the long wait.
I updated the tests to the new debugger test infrastructure (https://github.com/JetBrains/kotlin/commit/ce21a5594374b9567f92b6e7a79f4d3f502f998c). Unfortunately, I had to revert the commit (https://github.com/JetBrains/kotlin/commit/a19f8ffe6a6450b6caee55bdfab9e26ec2ee88c6) because of the wrong implementation in `BuilderFactoryForDuplicateSignatureDiagnostics.groupMembersDescriptorsBySignature`.

udalov:
I guess this is now obsolete, since test data for debugger tests is in the intellij repo.

PR #2397, opened 2019-06-05
[KT-31820] Added isNegative, isZero, and isPositive

Fleshgrinder:
[KT-31820](https://youtrack.jetbrains.com/issue/KT-31820)


ilya-g:
I think it's better to start with an issue in the tracker (https://kotl.in/issue) analyzing what use cases these proposed functions make better. For example, what advantage brings `isPositive`/`isNegative` over a comparison with the zero literal.

Fleshgrinder:
@ilya-g done 😎

PR #2385, opened 2019-05-30
JVM_IR: support @JvmDefaults in compatibility mode.

ting-yuan:
It's easier to start with a small PR. Some TODOs to be done in later PRs.
1. -Xjvm-default=enabled will be implemented later.
2. There are also a dozen of related tests failing and need to be fixed with other phases, such as accessor.

There are some changes in BridgeLowering.kt because otherwise it breaks some existing tests.

PR #2398, opened 2019-06-06
JVM_IR: handle JvmDefault for bridges

ting-yuan:
First commit is in https://github.com/JetBrains/kotlin/pull/2385.


ting-yuan:
@max-kammerer Looks like `irFunction.findConcreteSuperDeclaration()` only works for concrete functions. The function we are checking is always abstract.

Anyway, it's unrelated to JvmDefault and there is a test case covering it. Will address it in a separate PR.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
```diff
@@ -74,7 +75,7 @@ private class BridgeLowering(val context: JvmBackendContext) : ClassLoweringPass
         if (irFunction.isMethodOfAny()) return
 
         if (irFunction.origin === IrDeclarationOrigin.FAKE_OVERRIDE &&
-            irFunction.overriddenSymbols.all { it.owner.modality !== Modality.ABSTRACT && !it.owner.comesFromJava() }
+            irFunction.overriddenSymbols.all { (if ((it.owner.parent as? IrClass)?.isInterface == true) it.owner.hasJvmDefault() else it.owner.modality !== Modality.ABSTRACT) && !it.owner.comesFromJava() }
```
max-kammerer: `!it.owner.comesFromJava() ` looks wrong: java also generated bridges in interfaces and classes

ting-yuan: @max-kammerer Fixed in latest patch. I didn't include it in previous patch because it breaks 2 bytecodeText tests and I'm still investigating whether it is the IR backend or tests to blame:

`abstract class A8 : java.util.ArrayList<String>()` had a few bridges before this patch. Now it has no bridges. Although it looks unnecessary to me, the old backend (and the IR backend before this patch) expects the bridges. Are they necessary?

I've disabled those 2 tests for JVM_IR.

max-kammerer: Oh, yes, it's necessery: it's a bridge beetwen Kotlin  and Java collections - so it should be more wise filtering for java declarations here or later with restored '!it.owner.comesFromJava()' . //cc @gogabr

ting-yuan: @max-kammerer Done. Please take a look again.

max-kammerer:
Thank you!

PR #2402, opened 2019-06-07
Reduce compiler runtime overhead of MaxStackFrameSizeAndLocalsCalculator

jvgogh:
Profiling the compilation of kotlinx.serialization, MaxStackFrameSizeAndLocalsCalculator
causes ~7% of the runtime to be spent in java.lang.Object.hashCode

This is through two uses:
- visitMaxs(..) has a pushed hashSet that causes ~2%
- labelWrappersMap used to attach additional data to asm Labels, causes ~ 5%

visitMaxs was fixed by reusing the existing SmartHashSet class from IntelliJ Collections

Analysis of labelWrappersMap creation & sizes:

| What               | Amount |
| ------------------ | ------ |
| hashtables created | 4006   |
| max entries        | 175    |
| median entries     | 5      |
| average entries    | 6.10   |
| stdev entries      | 8.28   |
| 90 percentile      | 11     |

testing with a non hash based map using an array for keys and an array for values
showed that the cost of MaxStackFrameSizeAndLocalsCalculator became neglible to
the overall running time.

SmartHashMap implements the MutableMap interface. As long as there are 10 or less
entries in the map, it will use key & value arrays. Once the map grows beyond 10,
it will convert to a HashMap and will stay as such until clear() is called.
It will not convert back on remove() calls.

Between the two changes, compilation of kotlinx.serialization through kotlinc
commandline decreased from 14 seconds to 11 seconds on my machine.

### DIFF
--- a/compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
+++ b/compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
+
+const val ARRAY_UNTIL_SIZE = 10
+
+/**
+ * [SmartHashMap] is a Map implementation that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than 10.
+ * At that point it switches to using a HashMap and stays that way until [clear] is called.
+ * It does not auto convert back if the number of entries decreases through [remove].
+ */
+class SmartHashMap<K, V>() : MutableMap<K, V> {
+
+    private var keysArray: MutableList<K>? = arrayListOf()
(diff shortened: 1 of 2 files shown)


compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+
+    override val keys: MutableSet<K>
+        get() {
+            return keysArray?.let {
+                val ms = mutableSetOf<K>()
+                ms.addAll(it)
+                return ms
+
```
ting-yuan: `Iterable` has a handy extension `toMutableSet()`. So this can be just `return keysArray.toMutableSet()`

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+            } ?: largeMap!!.keys
+        }
+
+    override val size: Int
+        get() = keysArray?.size ?: largeMap!!.size
+
+    override val values: MutableCollection<V>
+        get() {
+            return valuesArray?.let {
+                val ms = mutableSetOf<V>()
```
ting-yuan: Values can be repeated. Maybe use `valuesArray.toMutableList()`?

PR #2403, opened 2019-06-07
Reduce compiler runtime overhead of MaxStackFrameSizeAndLocalsCalculator

jvgogh:
Profiling the compilation of kotlinx.serialization, MaxStackFrameSizeAndLocalsCalculator
causes ~7% of the runtime to be spent in java.lang.Object.hashCode

This is through two uses:
- visitMaxs(..) has a pushed hashSet that causes ~2%
- labelWrappersMap used to attach additional data to asm Labels, causes ~ 5%

visitMaxs can use the existing SmartSet (not to be confused with SmartHashSet)

Analysis of the visitMaxs HashSet creation & sizes:

| What               | Amount |
| ------------------ | ------ |
| calls to visitMaxs | 4416   |
| max pushed	     | 158    |
| median pushed	     | 4      |
| average pushed     | 5.20   |
| stddev pushed	     | 7.66   |
| 90 percentile      | 10     |

Analysis of labelWrappersMap creation & sizes:

| What               | Amount |
| ------------------ | ------ |
| hashtables created | 4006   |
| max entries        | 175    |
| median entries     | 5      |
| average entries    | 6.10   |
| stdev entries      | 8.28   |
| 90 percentile      | 11     |

testing with a non hash based map using an array for keys and an array for values
showed that the cost of MaxStackFrameSizeAndLocalsCalculator became neglible to
the overall running time.

SmartHashMap implements the MutableMap interface. As long as there are 10 or less
entries in the map, it will use key & value arrays. Once the map grows beyond 10,
it will convert to a HashMap and will stay as such until clear() is called.
It will not convert back on remove() calls.

Between the two changes, compilation of kotlinx.serialization through kotlinc
commandline decreased from 14 seconds to 11 seconds on my machine
(less than 7% due to startup time overhead)

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
@@ -46,8 +46,9 @@
 
 package org.jetbrains.kotlin.codegen.inline;
 
-import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.kotlin.utils.SmartSet;
+import org.jetbrains.kotlin.utils.SmartIdentityTable;
 import org.jetbrains.org.objectweb.asm.*;
 
 import java.util.*;
@@ -106,7 +107,7 @@ public class MaxStackFrameSizeAndLocalsCalculator extends MaxLocalsCalculator {
     private int maxStack;
 
     private final Collection<ExceptionHandler> exceptionHandlers = new LinkedList<>();
-    private final Map<Label, LabelWrapper> labelWrappersMap = new HashMap<>();
+    private final SmartIdentityTable<Label, LabelWrapper> labelWrappersTable = new SmartIdentityTable<>();
 
     public MaxStackFrameSizeAndLocalsCalculator(int api, int access, String descriptor, MethodVisitor mv) {
         super(api, access, descriptor, mv);
@@ -336,7 +337,7 @@ public class MaxStackFrameSizeAndLocalsCalculator extends MaxLocalsCalculator {
          */
         int max = 0;
         Stack<LabelWrapper> stack = new Stack<>();
-        Set<LabelWrapper> pushed = new HashSet<>();
+        Set<LabelWrapper> pushed = SmartSet.create();
 
         stack.push(firstLabel);
         pushed.add(firstLabel);
@@ -434,7 +435,7 @@ public class MaxStackFrameSizeAndLocalsCalculator extends MaxLocalsCalculator {
     // ------------------------------------------------------------------------
 
     private LabelWrapper getLabelWrapper(Label label) {
-        return ContainerUtil.<Label, LabelWrapper>getOrCreate(labelWrappersMap, label, () -> new LabelWrapper(label));
+        return labelWrappersTable.getOrCreate(label, () -> new LabelWrapper(label));
     }
 
     private void increaseStackSize(int variation) {
--- a/compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
+++ b/compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.util
+
+import com.intellij.openapi.util.Factory
+import org.jetbrains.kotlin.utils.SmartIdentityTable
+import org.junit.Assert
+import org.junit.Test
+
+class SmartIdentityTableTest {
+    class Key(val number: Int) {
+        override fun equals(other: Any?): Boolean {
+            Assert.fail("equals Should not be called")
+            return false
+        }
+
+        override fun hashCode(): Int {
+            Assert.fail("equals Should not be called")
+            return 0
+        }
+    }
+
+    class Value(val number: Int) {
+        override fun equals(other: Any?): Boolean {
+            Assert.fail("equals Should not be called")
+            return false
+        }
+
+        override fun hashCode(): Int {
+            Assert.fail("equals Should not be called")
+            return 0
+        }
+    }
+
+    @Test
+    fun basicTest() {
+        val key1 = Key(1)
+        val key2 = Key(2)
+        val key3 = Key(3)
+        val val1 = Value(1)
+        val val2 = Value(2)
+        val val3 = Value(3)
+
+        val table = SmartIdentityTable<Key, Value>()
+
+        // insert two keys and ensure that those are present
+        table[key1] = val1
+        table[key2] = val2
+
+        Assert.assertEquals(2, table.size)
+        Assert.assertTrue(table[key1] === val1)
+        Assert.assertTrue(table[key2] === val2)
+
+        // replace existing key's value
+        table[key1] = val3
+
+        // expect size to stame the same
+        Assert.assertEquals(2, table.size)
+
+        // values should be updated for key1 and same for other key
+        Assert.assertTrue(table[key1] === val3)
+        Assert.assertTrue(table[key2] === val2)
+
+        // add a new key with existing value
+        table[key3] = val2
+
+        // new key should be added and existing keys should maintain their values
+        Assert.assertEquals(3, table.size)
+        Assert.assertTrue(table[key1] === val3)
+        Assert.assertTrue(table[key2] === val2)
+        Assert.assertTrue(table[key3] === val2)
+
+        // create a key that has the same contents but a different reference identity, it should not be found in the table.
+        val secondKey1 = Key(1)
+        Assert.assertTrue(table[secondKey1] === null)
+    }
+
+    @Test
+    fun growToMapTest() {
+        val table = SmartIdentityTable<Key, Value>()
+
+        // insert enough data to trigger conversion of the Table to use a Map
+        val keys = mutableListOf<Key>()
+        for (i in 0 until 15) {
+            val key = Key(i)
+            val value = Value(i)
+            keys.add(key)
+            table[key] = value
+        }
+
+        Assert.assertEquals(15, table.size)
+        for (key in keys) {
+            val value = table[key]
+            Assert.assertNotNull(value)
+            Assert.assertEquals(key.number, table[key]!!.number)
+        }
+    }
+
+    @Test
+    fun getOrCreateTest() {
+        val table = SmartIdentityTable<Key, Value>()
+        val f = { Value(5) }
+        table.getOrCreate(Key(5), f)
+    }
+}
--- a/compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
+++ b/compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
+import java.util.IdentityHashMap
+
+/**
+ * [SmartIdentityTable] is a Map like structure that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than [ARRAY_UNTIL_SIZE].
+ * At that point it switches to using an IdentityHashMap.
+ *
+ * This structure can be used instead of [HashMap] when reference identity can be used and
+ * the number of entries inserted is small (<= [ARRAY_UNTIL_SIZE]) on average, drastically reducing the overhead
+ * of calls to [Object.hashCode].
+ *
+ * The implementation of [SmartIdentityTable] is not synchronized.
+ */
+class SmartIdentityTable<K, V> {
+
+    private var keysArray: MutableList<K>? = ArrayList(ARRAY_UNTIL_SIZE)
+    private var valuesArray: MutableList<V>? = ArrayList(ARRAY_UNTIL_SIZE)
+    private var largeMap: IdentityHashMap<K, V>? = null
+
+    val size: Int
+        get() = keysArray?.size ?: largeMap!!.size
+
+    operator fun get(key: K): V? {
+        return keysArray?.let {
+            for ((index, k) in it.withIndex()) {
+                if (k === key) {
+                    return valuesArray!![index]
+                }
+            }
+            return null
+        } ?: largeMap!![key]
+    }
+
+    operator fun set(key: K, value: V): V? {
+        val ka = keysArray
+        if (ka != null) {
+            val va = valuesArray!!
+            // scan for existing keys in array
+            for (i in 0 until ka.size) {
+                if (ka[i] === key) {
+                    val tmp = va[i]
+                    va[i] = value
+                    return tmp
+                }
+            }
+            // if a new key, and array has room
+            if (ka.size < ARRAY_UNTIL_SIZE) {
+                ka.add(key)
+                va.add(value)
+                return null
+            }
+            convertToHashMap()
+        }
+        // all other cases, fallback to IdentityHashMap implementation
+        return largeMap!!.put(key, value)
+    }
+
+    private fun convertToHashMap() {
+        val map = IdentityHashMap<K, V>()
+        val ka = keysArray!!
+        val va = valuesArray!!
+        for (i in 0 until ka.size) {
+            map[ka[i]] = va[i]
+        }
+        largeMap = map
+        keysArray = null
+        valuesArray = null
+    }
+
+    fun getOrCreate(key: K, factory: () -> V): V {
+        return this[key] ?: factory().also {
+            this[key] = it
+        }
+    }
+
+    companion object {
+        private const val ARRAY_UNTIL_SIZE = 10
+    }
+}


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
```diff
@@ -47,7 +47,9 @@
 package org.jetbrains.kotlin.codegen.inline;
 
 import com.intellij.util.containers.ContainerUtil;
+import com.intellij.util.containers.SmartHashSet;
```
punzki: Unused now

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * [SmartHashMap] is a Map implementation that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than 10.
+ * At that point it switches to using a HashMap and stays that way until [clear] is called.
+ * It does not auto convert back if the number of entries decreases through [remove].
+ */
+class SmartHashMap<K, V>() : MutableMap<K, V> {
+
+    private var keysArray: MutableList<K>? = arrayListOf()
```
punzki: It will be faster if you just use `Array` directly of size `ARRAY_UNTIL_SIZE`. e.g., `private var keysArray: Array<K?>? = arrayOfNulls(ARRAY_UNTIL_SIZE)`. Use a simple `Int` property to keep track of the number of elements in the array. It will be tricky what to do about removals but perhaps you can use null to track empty slots. (keys cannot be null anyway)

Ditto for `valuesArray`

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
+
+const val ARRAY_UNTIL_SIZE = 10
+
+/**
+ * [SmartHashMap] is a Map implementation that uses reference identity for keys.
```
punzki: Also add a note that this class is not thread-safe

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+            } ?: largeMap!!.entries
+        }
+
+
+    override val keys: MutableSet<K>
+        get() {
+            return keysArray?.let {
+                val ms = mutableSetOf<K>()
+                ms.addAll(it)
+                return ms
```
ting-yuan: `Iterable` has a handy extension `toMutableSet()`. So this can be just `return keysArray.toMutableSet()`.

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+            } ?: largeMap!!.keys
+        }
+
+    override val size: Int
+        get() = keysArray?.size ?: largeMap!!.size
+
+    override val values: MutableCollection<V>
+        get() {
+            return valuesArray?.let {
+                val ms = mutableSetOf<V>()
```
ting-yuan: Values can be repeated. Maybe use `valuesArray.toMutableList()`?

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+package org.jetbrains.kotlin.utils
+
+const val ARRAY_UNTIL_SIZE = 10
+
+/**
+ * [SmartHashMap] is a Map implementation that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than 10.
+ * At that point it switches to using a HashMap and stays that way until [clear] is called.
```
punzki: I think you should switch to use IdentityHashMap instead of HashMap, to keep it consistent that this map uses reference identity for keys.

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+ */
+class SmartHashMap<K, V>() : MutableMap<K, V> {
+
+    private var keysArray: MutableList<K>? = arrayListOf()
+    private var valuesArray: MutableList<V>? = arrayListOf()
+    private var largeMap: HashMap<K, V>? = null
+
+    override val entries: MutableSet<MutableMap.MutableEntry<K, V>>
+        get() {
+            return keysArray?.let {
```
punzki: Look into using `mapTo()` to make this (and `keys`, `values`, and `convertToHashMap()`) more "Kotlin-ish" (or `mapNotNullTo()` if you take my suggestion of using Array. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null-to.html

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+    override fun containsKey(key: K): Boolean = keysArray?.contains(key) ?: largeMap!!.containsKey(key)
+
+
+    override fun containsValue(value: V): Boolean = valuesArray?.contains(value) ?: largeMap!!.containsValue(value)
+
+    override fun get(key: K): V? {
+        return keysArray?.let {
+
+            for (i in 0 until it.size) {
+                if (it[i] === key) {
```
ting-yuan: The underlying HashMap and MutableList compare with `.equals()` rather than identity so most of SmartHashMap still use equality except here and below.

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    override fun containsKey(key: K): Boolean = keysArray?.contains(key) ?: largeMap!!.containsKey(key)
+
+
+    override fun containsValue(value: V): Boolean = valuesArray?.contains(value) ?: largeMap!!.containsValue(value)
+
+    override fun get(key: K): V? {
+        return keysArray?.let {
+
+            for (i in 0 until it.size) {
```
punzki: Use `find()` (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/find.html). Ditto for `put()` and `remove()`.

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: that is a bit annoying as you then have to defrag the array manually. instead I used ArrayList<K>(ARRAY_UNTIL_SIZE) to preallocate the underlying array to 10, but still maintain the size semantics.

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: that doesn't work unfortunately, as the index operates over both the key and value arrays

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: cleaned the whole file for this pattern, using IdentityHashMap as Mark suggested and replaced the calls to contains with any{x===y}

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: cleaned up a different way

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,149 @@
...
+        }
+        largeMap!!.putAll(from)
+    }
+
+    override fun remove(key: K): V? {
+        val ka = keysArray
+        if (ka != null) {
+            val va = valuesArray!!
+            for (i in 0 until ka.size) {
+                if (ka[i] == key) {
```
ting-yuan: Should this be `===`?

jvgogh: whoops, fixed

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,149 @@
...
+package org.jetbrains.kotlin.utils
+
+import java.util.IdentityHashMap
+
+const val ARRAY_UNTIL_SIZE = 10
+
+/**
+ * [SmartHashMap] is a Map implementation that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than 10.
+ * At that point it switches to using a HashMap and stays that way until [clear] is called.
```
punzki: Nit: HashMap -> IdentityHashMap

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
punzki: Ah didn't notice that. In which case you can use `withIndex`:
```
for ((index, k) in it.withIndex()) {
  if (k === key) {
    return valuesArray!![index]
  }
}
```

I think you can also use it in `put` and `remove`. `for (i in 0 until someCollection.size)` generally should be replaceable with `withIndex`

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,149 @@
...
+            for (i in 0 until ka.size) {
+                if (ka[i] === key) {
+                    val tmp = va[i]
+                    ka.removeAt(i)
+                    va.removeAt(i)
+                    return tmp
+                }
+            }
+            return null
+        } else {
```
punzki: Nit: else not needed

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
+
+import java.util.IdentityHashMap
+
+const val ARRAY_UNTIL_SIZE = 10
```
dzharkov: I'd probably make it private or even put within a companion for `SmartHashMap` to avoid its appearing in other scopes

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
```diff
@@ -336,10 +337,10 @@ public void visitMaxs(int maxStack, int maxLocals) {
          */
         int max = 0;
         Stack<LabelWrapper> stack = new Stack<>();
-        Set<LabelWrapper> pushed = new HashSet<>();
+        Map<LabelWrapper, LabelWrapper> pushed = new SmartHashMap<>();
```
udalov: Why not use `org.jetbrains.kotlin.utils.SmartSet` here?

udalov:
@jvgogh I've played with a slightly different implementation of a smart hash map some time ago, which I never pushed to master since I couldn't find a place where it improved performance significantly when used instead of the standard hash map:
https://github.com/udalov/kotlin/blob/b58136485fa69573c91fc772ba5e668d68667c8d/core/util.runtime/src/org/jetbrains/kotlin/utils/SmartMap.kt

I mostly optimized for memory, but if we change it to allocate the array at creation instead of growing it on `add`, it should have reasonable time performance too. Please take a look. Since this is such a core utility API, it may soon be used in other places too, therefore I propose to make it as correct and optimal as is reasonable.

Key points which I think could be incorporated into your implementation:
* My implementation explicitly doesn't support `remove`. Since the only possible use site at `MaxStackFrameSizeAndLocalsCalculator` doesn't need it (and actually I suspect that most code in our project doesn't use `MutableMap.remove`), the implementation is untested, and I actually think it's more reliable to throw UOE and explicitly state that `remove` is not supported in the kdoc
* Using raw arrays instead of lists should be better both in terms of memory and time (will be useful in scenarios where we create a lot of small maps). If we don't support `remove`, [your concern here](https://github.com/JetBrains/kotlin/pull/2403#discussion_r292029937) is no longer applicable and the implementation will be easier
* I use a single field both for the small array of keys+values and for the larger map, to save space. I'm not sure if checkcasts of that field to array/map on each access hurt execution performance or not, compared to the case of three explicit fields, so maybe this should be investigated
* My implementation uses `Object.equals/hashCode` for equality, and fallbacks to `LinkedHashMap` when grows large. Since this is the contract of `HashMap`, I think this should be the default behavior of `SmartHashMap`, and we could add a subclass `SmartIdentityHashMap` where equality/hash code logic is overridden and is identity-based.
* `Map.keys`/`Map.values`/`Map.entries` should actually be views into the map, meaning that if the original map changes, the changes should be visible through those views and vice versa. Currently your implementation returns a new mutable collection _almost_ every time, which could lead to very subtle bugs. BTW `java.util.HashMap` actually stores all three collections in fields, which I avoided to save space. (On the other hand, for some reason I implemented `containsKey`/`containsValue` via those views, and that is suboptimal because it creates an object each time.)

compiler/util/src/org/jetbrains/kotlin/utils/SmartHashMap.kt
jvgogh: Done

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
jvgogh: See my PR notes/commit message. SmartSet is only optimized for the 0/1 case and then fallsback to HashSet. From my data analysis, this doesn't make the pushed set a good candidate for SmartSet:

| What               | Amount |
| ------------------ | ------ |
| hashtables created | 4006   |
| max entries        | 175    |
| median entries     | 5      |
| average entries    | 6.10   |
| stdev entries      | 8.28   |
| 90 percentile      | 11     |

It would fallback to a Hashtable in almost all cases, making this a cost instead of a savings

jvgogh:
@udalov Re: SmartMap. I'm not sure that would help in this case, as the real overhead is the time spent in Object.hashCode My implementation explicitly tries to avoid that by looking at object reference equality only. 

Re:maintaining Map semantics. Mabye we should just not implement the java.util.Map interface here. Both use cases, remove, keys, values etc.. are never needed. We can save on code, complexity and confusion by just implementing the bare minimum to make the performance work well here.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MaxStackFrameSizeAndLocalsCalculator.java
udalov: It's actually optimized for sizes <= 5, see https://github.com/JetBrains/kotlin/blob/master/core/util.runtime/src/org/jetbrains/kotlin/utils/SmartSet.kt. Maybe you're confusing it with `SmartHashSet` from intellij-core, which is indeed not applicable here?

jvgogh: Ah yes, let me give SmartSet a try

jvgogh: confirmed that SmartSet removes the performance overhead on object.hashCode from visitMaxs(..) as well. I've applied the change.

udalov:
@jvgogh 
> Re: SmartMap. I'm not sure that would help in this case, as the real overhead is the time spent in Object.hashCode My implementation explicitly tries to avoid that by looking at object reference equality only.

As I mentioned above, we can override this behavior in a subclass. So, `SmartHashMap` could be an open class with open method `areEqual` which uses `Object.equals`, overridden in the subclass `SmartIdentityHashMap` where it uses `===`. Alternatively, you could simply rename the class to `SmartIdentityHashMap`. But it'll probably be useful somewhere else at some point, so I suggested to extract the "default" equals-based superclass right away.

> Re:maintaining Map semantics. Mabye we should just not implement the java.util.Map interface here. Both use cases, remove, keys, values etc.. are never needed. We can save on code, complexity and confusion by just implementing the bare minimum to make the performance work well here.

I'm OK with removing `Map` superinterface and simplifying everything, but I'd suggest renaming the class then to something like `SmartIdentityTable` to avoid the cognitive association with maps and explicitly mentioning that it uses identity equals/hashCode.

My original point was that since this class is so centrally placed, it makes sense to make it as general and correct as possible, since it inevitably will end up being used in other places. (An alternative would be to just make it package-private near its only use site, but that would be kind of short-sighted IMHO.)

jvgogh:
Sorry for the late reply, was in a summit most of the day.
Let me rework the CL to not implement MutableMap, rename it SmartIdentityTable and then add a bunch of tests to ensure the correct behavior. I'll update the PR in a few days.

jvgogh:
updated to SmartIdentityTable and tests added

compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
```
udalov: Please change package to `org.jetbrains.kotlin.util`, as per the inspection that is reported here

compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
```diff
@@ -0,0 +1,107 @@
...
+            return 0
+        }
+    }
+
+    @Test
+    fun basicTest() {
+        val key1 = Key(1)
+        val key2 = Key(2)
+        val key3 = Key(3)
+        var val1 = Value(1)
```
udalov: `val`

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
```diff
@@ -0,0 +1,86 @@
...
+ * This structure can be used instead of [HashMap] when reference identity can be used and
+ * the number of entries inserted is small (<= [ARRAY_UNTIL_SIZE]) on average, drastically reducing the overhead
+ * of calls to [Object.hashCode].
+ *
+ * The implementation of [SmartIdentityTable] is not synchronized.
+ */
+class SmartIdentityTable<K, V> {
+
+    private var keysArray: MutableList<K>? = ArrayList<K>(ARRAY_UNTIL_SIZE)
+    private var valuesArray: MutableList<V>? = ArrayList<V>(ARRAY_UNTIL_SIZE)
```
udalov: Please remove explicit `<K>` and `<V>` as per the warnings

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
```diff
@@ -0,0 +1,86 @@
...
+        val va = valuesArray!!
+        for (i in 0 until ka.size) {
+            map[ka[i]] = va[i]
+        }
+        largeMap = map
+        keysArray = null
+        valuesArray = null
+    }
+
+    fun getOrCreate(key: K, factory: Factory<out V>): V {
```
udalov: We usually prefer to use the Kotlin standard library instead of intellij-core when possible. Here a functional type `() -> V` would probably look better both at declaration and use site

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
```diff
@@ -0,0 +1,86 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.utils
+import com.intellij.openapi.util.Factory
+import java.util.IdentityHashMap
+
+/**
+ * [SmartIdentityTable] is a Map like structure that uses reference identity for keys.
+ * It uses 2 arrays to store keys & values until the number of entries stored is larger than 10.
```
udalov: Maybe `[ARRAY_UNTIL_SIZE]` instead of 10

udalov:
Looks great now, modulo minor remarks. I'll merge this as soon as they're fixed and the remote run build is successful.

compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
jvgogh: Ah, the test package is util and the production code package is utils. 
Fixed.

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
jvgogh: done

compiler/tests/org/jetbrains/kotlin/util/SmartIdentityTableTest.kt
jvgogh: done

udalov:
Merged manually in 1dbe48707794d6fe61525edea1197ad5aa7c4b34.

@jvgogh Thank you!

PR #2404, opened 2019-06-08
 IDE, wizards: update Kotlin/JS support

snrostov:


idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/GradleKotlinFrameworkSupportProvider.kt
```diff
@@ -187,21 +187,59 @@ open class GradleKotlinJavaFrameworkSupportProvider(
...
+}
+
+open class GradleKotlinJSNodeFrameworkSupportProvider(
+    frameworkTypeId: String = "KOTLIN_JS_NODE",
+    displayName: String = "Kotlin/JS for nodejs"
+) : GradleKotlinJSFrameworkSupportProvider(frameworkTypeId, displayName) {
+    override val jsSubTargetName: String
+        get() = "nodejs"
+
+    override fun getDescription() = "A Kotlin library or application targeting JavaScript for nodejs"
```
h0tk3y: Minor: maybe stylize nodejs as Node.js (the official stylization)?

snrostov: Indeed. Will do.

idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/GradleKotlinFrameworkSupportProvider.kt
```diff
@@ -187,21 +187,59 @@ open class GradleKotlinJavaFrameworkSupportProvider(
...
+        super.addSupport(buildScriptData, module, sdk, specifyPluginVersionIfNeeded, explicitPluginVersion)
+
+        buildScriptData.addOther(
+            """
+            kotlin {
+                target {
+                    $jsSubTargetName {
+                    }
+                }
+            }""".trimIndent()
```
h0tk3y: Minor: is there a reason to emit six lines with nested scopes instead of one or two (`kotlin.target.browser`)?

snrostov: User also may want to add `npm` dependencies into `sourceSets.main.depdencies`. But currently we generate `dependency` section above anyway. So I agree that shortcut is more preferable for now. Will fix.

idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/GradleKotlinFrameworkSupportProvider.kt
snrostov: Fixed

idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/GradleKotlinFrameworkSupportProvider.kt
snrostov: Fixed

snrostov:
Merged manually.

PR #2405, opened 2019-06-10
KT-30782 'Show Method Separators' does not separate expression body Kotlin functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30782


goodwinnk:
There was an idea that method separators might be to heavy for expression functions. But I think this fix leads to more logical behavior. Thank you for the fix!

Merged as https://github.com/JetBrains/kotlin/commit/bd661fcf82a232beef4d5e69a929bbbe8fe4d46d.

PR #2406, opened 2019-06-10
Make `KotlinCompile` task compatible with experimental instant execution mode

bamboo:
We are currently experimenting with a new build execution strategy that improves repeated build execution performance by caching configured task graphs. We call it [instant execution](https://gradle.github.io/instant-execution/).

The new strategy imposes a few limitations on how tasks manage state and what sort of state `@TaskAction` methods have access to:
- tasks must be isolated from each other, i.e., no referencing a task instance from another during execution
- required `Project` state (including `Configuration` values) must be explicitly captured in fields otherwise it won't be available for instant execution 

This PR includes an integration test and some examples of the sort of changes that would be necessary to make vanilla `KotlinCompile` tasks compatible with instant execution.  

The biggest challenge seems to be converting the computations performed by `org.jetbrains.kotlin.compilerRunner.GradleCompilerRunner.Companion#buildModulesInfo` and the resulting state to be represented via [Gradle lazy configuration](https://docs.gradle.org/current/userguide/lazy_configuration.html).

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
@@ -88,8 +88,8 @@ abstract class AbstractKotlinCompileTool<T : CommonToolArguments>
 
     @get:Classpath
     @get:InputFiles
-    internal val computedCompilerClasspath: List<File>
-        get() = compilerClasspath?.takeIf { it.isNotEmpty() }
+    internal val computedCompilerClasspath: List<File> by lazy {
+        compilerClasspath?.takeIf { it.isNotEmpty() }
             ?: compilerJarFile?.let {
                 // a hack to remove compiler jar from the cp, will be dropped when compilerJarFile will be removed
                 listOf(it) + findKotlinCompilerClasspath(project).filter { !it.name.startsWith("kotlin-compiler") }
@@ -108,6 +108,7 @@ abstract class AbstractKotlinCompileTool<T : CommonToolArguments>
                 findKotlinCompilerClasspath(project)
             }
             ?: throw IllegalStateException("Could not find Kotlin Compiler classpath")
+    }
 
 
     protected abstract fun findKotlinCompilerClasspath(project: Project): List<File>
@@ -155,8 +156,9 @@ abstract class AbstractKotlinCompile<T : CommonCompilerArguments>() : AbstractKo
 
     @get:Classpath
     @get:InputFiles
-    val pluginClasspath: FileCollection
-        get() = project.configurations.getByName(PLUGIN_CLASSPATH_CONFIGURATION_NAME)
+    val pluginClasspath: FileCollection by lazy {
+        project.configurations.getByName(PLUGIN_CLASSPATH_CONFIGURATION_NAME)
+    }
 
     @get:Internal
     internal val pluginOptions = CompilerPluginOptions()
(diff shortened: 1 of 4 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
```diff
@@ -168,14 +170,14 @@ abstract class AbstractKotlinCompile<T : CommonCompilerArguments>() : AbstractKo
         get() = (classpath + additionalClasspath)
             .filterTo(LinkedHashSet(), File::exists)
 
-    private val sourceFilesExtensionsSources: MutableList<Iterable<String>> = mutableListOf()
+    private val sourceFilesExtensionsSources = project.objects.listProperty(String::class.java)
```
alexey-tsvetkov: `listProperty` is introduced in Gradle `4.3`, but we support all versions starting from `4.1`

bamboo: Thank you for the quick feedback, @alexey-tsvetkov. I'll see what I can do about it.

alexey-tsvetkov:
@bamboo is the feature gonna work if some tasks in graph don't meet the criteria above? Does it checks tasks (i.e. gives meaningful error messages) or fails on execution?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
alexey-tsvetkov: @bamboo our compatibility policy is to support Gradle versions at least for 1 year, so we can bump the minimal supported version to 4.3 (in theory even to 4.8, but we don't want to deprecate older versions without reason)

alexey-tsvetkov:
So far, the PR looks good to me (API compatibility issue can be solved by dropping support for older versions of Gradle). 
@bamboo do you want us to merge it? (I'm asking, because you've submitted the PR as draft)

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
big-guy: The initial value for listProperty changed a couple of times (empty list, not set, empty list), so if you want to keep compatibility, you should explicitly set the initial value to an empty list.

bamboo: That's great to hear @alexey-tsvetkov, thanks.

bamboo: Thanks, @big-guy, will do.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/Tasks.kt
```diff
@@ -168,14 +170,14 @@ abstract class AbstractKotlinCompile<T : CommonCompilerArguments>() : AbstractKo
...
 
     @get:Input
     val sourceFilesExtensions: List<String>
-        get() = DEFAULT_KOTLIN_SOURCE_FILES_EXTENSIONS + sourceFilesExtensionsSources.flatten()
+        get() = DEFAULT_KOTLIN_SOURCE_FILES_EXTENSIONS + sourceFilesExtensionsSources.get()
 
     internal fun sourceFilesExtensions(extensions: Iterable<String>) {
-        sourceFilesExtensionsSources.add(extensions)
+        sourceFilesExtensionsSources.addAll(extensions)
```
alexey-tsvetkov: Another problem is that `ListProperty#addAll` was added in Gradle 4.10. We cannot drop support for all versions below 4.10.

bamboo:
I'm happy this work is now in the very capable hands of the Kotlin team.

Closing it as superseded by [KT-33908](https://youtrack.jetbrains.com/issue/KT-33908).

PR #2407, opened 2019-06-10
add links to Scope Functions doc from with, run, let, apply, and also…

kevpeek:

Updates the doc block for the Any class's scope extension functions so they link to the documentation here: https://kotlinlang.org/docs/reference/scope-functions.html


ilya-g:
I think this will do, thanks.

PR #2409, opened 2019-06-13
JVM_IR: support vararg & defaults in function references

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -158,6 +146,12 @@ internal class CallableReferenceLowering(val context: JvmBackendContext) : FileL
         })
     }
 
+    private val arrayGetFun =
```
max-kammerer: @pyos Please move these declaration into FunctionReferenceBuilder where they are used: this will allow to avoid overhead with loading lazy data when it is not necessery.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -168,10 +162,18 @@ internal class CallableReferenceLowering(val context: JvmBackendContext) : FileL
...
+
         private val callee = irFunctionReference.symbol.owner
         private val calleeParameters = callee.explicitParameters
         private val boundCalleeParameters = irFunctionReference.getArgumentsWithIr().map { it.first }
-        private val unboundCalleeParameters = calleeParameters - boundCalleeParameters
+
+        // The type of the reference is KFunction<in A1, ..., in An, out R>
+        private val argumentTypes = (irFunctionReference.type as IrSimpleType).arguments.dropLast(1).map { (it as IrTypeProjection).type }
+        private val returnType = ((irFunctionReference.type as IrSimpleType).arguments.last() as IrTypeProjection).type
+        private val useVararg = (argumentTypes.size > MAX_ARGCOUNT_WITHOUT_VARARG)
```
max-kammerer: Please redefine MAX_ARGCOUNT_WITHOUT_VARARG with `FunctionInvokeDescriptor.BIG_ARITY - 1` or use BIG_ARITY instead

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
pyos: @max-kammerer done

gogabr:
Merged manually after regenerating tests.
Thank you!

PR #2410, opened 2019-06-13
KT-31127: register only generated Java sources to AGP

gavra0:
When adding generated Java sources to AGP using Variant API,
add only Java sources. This was already done in
a6ae4494167c9d4cca23995457c59fb2daedad39 for Java projects, and this
commit fixes the same issue for Android projects.

Test: Kapt3Android33IT.testKotlinProcessorUsingFiler


emartynov:
What is usual ETA of getting these small updates merged to the main stream?

yanex:
Merged into master manually (https://github.com/JetBrains/kotlin/commit/f60bf5cb97cfd940edf8b40715f92834ef0b5708). Thanks for the contribution!

PR #2412, opened 2019-06-14
Generate synthetic functions for local functions with default values.

punzki:
Generate synthetic functions for local functions with default values, by re-ordering the lowering phases.

The changes in InterfaceLowering are necessary so that IrElements that target the removed functions are re-targeted to the new functions in DefaultImpls. This affects local functions in interface functions since now LocalDeclarationsLowering comes before InterfaceLowering.

This supercedes #2128, which took a different approach to generate the functions. #2128 has changes that affected the non-IR backend, and also fixed fewer disabled tests (e.g., KT-5589: local functions with default values from closures).


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -465,46 +470,58 @@ fun irCall(
...
+    firstArgumentsAsReceivers: Boolean = false
 ) {
     copyTypeArgumentsFrom(call)
 
     var toValueArgumentIndex = 0
     var fromValueArgumentIndex = 0
 
     when {
-        dispatchReceiverAsFirstArgument -> {
+        receiversAsFirstArguments && call.descriptor.dispatchReceiverParameter != null -> {
```
neetopia: Is there any difference between call.dispatchReceiver and call.descriptor.dispatchReceiverParameter in this case? Same question to other usage of descriptors.

punzki: `call.descriptor.dispatchReceiverParameter` is the parameter in the function definition (i.e., in the function signature) and is null if it's a static function.`call.dispatchReceiver` is the actual argument value (for the parameter) in the call.

The reason I added the descriptor checks is to make sure we only copy the receivers from the "from" call if there are receiver params in the "from" function. Ditto for the other case: Only copy to receivers in the "to" call if there are receiver params in the "to" function.

gogabr: We are getting rid of descriptor usage in lowerings. You can use `call.symbol.owner.dispatchReceiverParameter` for an IR counterpart of `call.descriptor.dispatchReceiverParameter`.

punzki: @gogabr Done. I had to refactor a little bit because `call` is `IrMemberAccessExpression` which doesn't have the `symbol` property, but `IrFunctionReference` and `IrFunctionAccessExpression` do.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -100,6 +107,59 @@ private class InterfaceLowering(val context: JvmBackendContext) : IrElementTrans
...
+        )
+    }
+
+    override fun visitCall(expression: IrCall): IrExpression {
+        val newFunction = removedFunctions[expression.symbol]?.owner
+        return super.visitCall(
+            if (newFunction != null) {
+                with(expression) {
+                    if (this is IrCallWithShallowCopy) {
+                        shallowCopy(origin, symbol, superQualifierSymbol)
```
pyos: Shouldn't this be `newFunction.symbol`?

punzki: You are correct, good catch! Actually, thanks for catching this, "correcting" it actually surfaced some errors here. `IrCallWithShallowCopy` is implemented by 4 classes:
1. `IrBinaryPrimitiveImpl`
1. `IrNullaryPrimitiveImpl`
1. `IrGetterCallImpl`
1. `IrSetterCallImpl`

The first 2 will never be present in an interface and moved by `InterfaceLowering`. The latter 2 do exist for properties in interfaces, and will be moved to `DefaultImpls`. However, these will no longer have the dispatch receiver parameters so doing the shallow copy does not work. Simplifying the code to:
```
with(expression) {
    irCall(this, newFunction, receiversAsFirstArguments = true)
}
```
...works and we can now enable 4 more tests. :slightly_smiling_face:

pyos: Maybe make it even shorter then? :)

```
irCall(expression, newFunction, receiversAsFirstArguments = true)
```

I also have doubts about the other uses of `with`...

punzki: Done. I use `with` just so I don't have to do `expression.` for offsets, type, etc. That said, I missed a few spots and fixed those now.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -465,46 +470,72 @@ fun irCall(
...
+) = copyTypeAndValueArgumentsFrom(src, src.symbol.owner, symbol.owner, receiversAsFirstArguments, firstArgumentsAsReceivers)
+
 private fun IrMemberAccessExpression.copyTypeAndValueArgumentsFrom(
-    call: IrMemberAccessExpression,
-    dispatchReceiverAsFirstArgument: Boolean = false,
-    firstArgumentAsDispatchReceiver: Boolean = false
+    src: IrMemberAccessExpression,
+    srcFunction: IrFunction,
+    destFunction: IrFunction,
+    receiversAsFirstArguments: Boolean = false,
```
max-kammerer: Could you a bit explain why extension receiver should be moved to value parameters?
Seems both schemes should work properly.

NB: This renaming should also affect native code.

compiler/testData/codegen/box/functions/localFunctions/kt3978.kt
```diff
@@ -1,11 +1,20 @@
-// IGNORE_BACKEND: JVM_IR
-fun box() : String {
-
+class C() {
```
max-kammerer: Please keep old test and add new one

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
gogabr: Test that fails without this change: compiler/testData/codegen/box/traits/traitWithPrivateExtension.kt

max-kammerer: @gogabr Oh thanks. Now I caught the case. 
@punzki Please check that renaming and new logic not affect K/N

punzki: That's right, that's the test case that fails.

@max-kammerer - This param is not used in K/N.

compiler/testData/codegen/box/functions/localFunctions/kt3978.kt
punzki: Done. I split it up but also cleaned up the old test a bit.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
punzki: Modified this so there is a separate set of parameters for the extension receiver. Looks like it is needed to address this failure: https://teamcity.jetbrains.com/viewLog.html?buildId=2383285&buildTypeId=Kotlin_dev_Compiler. InlineClassLowering creates functions where the dispatch receiver is moved to an argument but not the extension receiver.

max-kammerer: It could be unexpected behavior or bug in InlineClassLowering
@skuzmich @romanart Could you take a look?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -440,18 +440,29 @@ fun irConstructorCall(
...
 
 fun irCall(
-    call: IrMemberAccessExpression,
+    call: IrFunctionAccessExpression,
     newFunction: IrFunction,
-    dispatchReceiverAsFirstArgument: Boolean = false,
-    firstArgumentAsDispatchReceiver: Boolean = false
+    dispatchReceiverAsArgument: Boolean = false,
+    argumentAsDispatchReceiver: Boolean = false,
+    extensionReceiverAsArgument: Boolean = dispatchReceiverAsArgument,
```
skuzmich: These defaults look confusing. Any objections on changing it to `false` to make intent explicit?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
punzki: The relevant lines are here: https://github.com/JetBrains/kotlin/blob/d14859b44e1e1241a8e3c9d01b40b0b8175cf0c1/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InlineClassDeclarationLowering.kt#L267-L273. Extension receiver is copied as-is, but dispatch receiver is copied as a parameter. It seemed a little odd to me, but I don't know much about this lowering.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
punzki: I don't have any objections to making it false by default, since that was essentially the original behavior anyway. I will make this change.

The reason I did this is that I _thought_ it is more typical that both corresponding arguments (e.g., `dispatchReceiverAsArgument` and `extensionReceiverAsArgument`) would have the same value when this function is used. For example, in JVM, when methods are moved from an interface into the "DefaultImpls" class as a static method, both dispatch receiver and extension receiver become parameters.

I don't know enough about InlineClassLowering to determine why it needs to keep the extension receiver as-is. AFAIK extension receivers become parameters eventually anyway. But I don't have any objections to just making it false by default.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
max-kammerer: @skuzmich seems extensionReceiverParameter also should be passes as value parameter

max-kammerer: NB: createStaticBodilessMethod seems could be merged with createStaticFunctionWithReceivers

skuzmich: This was intentional. I thought keeping extension receiver would be less intrusive. Now I see that this is odd because conventional ordering receivers is changed. 

We can fix is here or in separate commit, but IrFunction and IrCall transformations have to be in sync either way.

max-kammerer: @punzki Would you take a look? Additional parameters for extension receiver become unnecessary

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
skuzmich: Great! These functions are originated in JS IR BE where they are used extensively to translate missing built-ins (Any, Int, etc.) member calls into runtime plain function calls. These are concrete functions with known signature, AFAIK without extension receivers.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
punzki: @max-kammerer @skuzmich Something like this? (see latest commit)

max-kammerer: @punzki Yes! Many thanks!

skuzmich: @punzki LGTM

max-kammerer:
@punzki Thank you!

PR #2413, opened 2019-06-14
KT-30824 No highlighting of declaration/usage of function with functional-type (lambda) parameter on its usage

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30824


goodwinnk:
This is an important fix, thank you! Merged manually: https://github.com/jetbrains/kotlin/commit/6b36833ee21231046190ecdd9128e0abdf42199b.

I have also asked Intellij IDEA team for providing API for testing `highlightReferences()` method behavior.

PR #2417, opened 2019-06-19
Handle `step` progressions in ForLoopsLowering.

punzki:
The first commit is from #2390.

I still need to add a ton more tests, but I'm sending this out for review of the rest of the logic which is mostly in `StepHandler.kt`. You can see the direction I'm going w.r.t. testing in the currently added box tests, and I'm going to add more bytecode text tests as well. I also moved the handlers into separate files since the single file was getting quite large.

This version handles more cases than the `ForLoopsLowering` in Kotlin/Native (such as the nested `step` case and combinations of `reverse`+`step`) and minimizes temporary variable use.


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/matchers/IrFunctionMatcher.kt
```diff
@@ -97,3 +99,13 @@ internal fun IrFunctionMatcherContainer.singleArgumentExtension(
     fqName { it == fqName }
     return this
 }
+
+internal fun IrFunctionMatcherContainer.singleArgumentExtensionBySimpleType(
```
pyos: This looks suspicious, why not change `progressionClassesTypes` to contain IR types?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -221,6 +217,302 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
...
+internal class StepHandler(
+    private val context: CommonBackendContext,
+    private val visitor: IrElementVisitor<HeaderInfo?, Nothing?>
+) : ProgressionHandler {
+
+    private val symbols = context.ir.symbols
+
+    override val matcher = SimpleCalleeMatcher {
+        singleArgumentExtensionBySimpleType(FqName("kotlin.ranges.step"), symbols.progressionClassesTypes)
+        parameterCount { it == 1 }
```
pyos: This is redundant.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -221,6 +217,302 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
...
+            // We can return the nested info if its step is constant and its absolute value is the same as the step argument. Examples:
+            //
+            //   1..10 step 1               // Nested step is 1, argument is 1. Equivalent to `1..10`.
+            //   10 downTo 1 step 1         // Nested step is -1, argument is 1. Equivalent to `10 downTo 1`.
+            //   10 downTo 1 step 2 step 2  // Nested step is -2, argument is 2. Equivalent to `10 downTo 1 step 2`.
+            if (stepArg.constLongValue != null && nestedInfo.step.constLongValue?.absoluteValue == stepArg.constLongValue) {
+                return nestedInfo
+            }
+
+            // To reduce local variable usage, we create and use temporary variables only if necessary.
```
pyos: Does this actually have any effect? At least on the JVM, `JvmBuiltinOptimizationLowering` would remove these variables anyway.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -221,6 +217,302 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
...
+                        irGet(nestedStepVar)
+                    } else {
+                        nestedStep
+                    }
+                    val nestedStepPositiveCheck = irCall(stepGreaterFun).apply {
+                        putValueArgument(0, nestedStepExpression)
+                        putValueArgument(1, zeroStep)
+                    }
+
+                    checkedStepVar = scope.createTemporaryVariable(checkedStepExpression, nameHint = "checkedStep")
```
pyos: The only variable that is created even if the value is constant - is this for a reason? Looks like it can be constant...

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/handlers/UntilHandler.kt
```diff
@@ -0,0 +1,168 @@
...
+import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.name.FqName
+
+/** Builds a [ProgressionHeaderInfo] for progressions built using the `until` extension function. */
+internal class UntilHandler(private val context: CommonBackendContext, private val progressionElementTypes: Collection<IrType>) :
+    ProgressionHandler {
+
+    override val matcher = SimpleCalleeMatcher {
+        singleArgumentExtension(FqName("kotlin.ranges.until"), progressionElementTypes)
+        parameterCount { it == 1 }
```
pyos: Also redundant.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/matchers/IrFunctionMatcher.kt
pyos: Ah, already done, sorry.

punzki: Oops, fixed.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
punzki: It does make a difference. `JvmBuiltinOptimizationLowering` does not:
1. Eliminate the variable or its initializer. Not sure if these get optimized out at the bytecode level later on, I don't believe so.
2. Replace gets of variables that are initialized by gets of variables.

Now these _could_ be addressed in the lowering or some other lowering (e.g., constant propagation or something similar).

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
punzki: Missed that, thanks!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/handlers/StepHandler.kt
```diff
@@ -0,0 +1,290 @@
...
+
+            // The `step` standard library function only accepts positive values, and performs the following check:
+            //
+            //   if (step > 0) step else throw IllegalArgumentException("Step must be positive, was: $step.")
+            //
+            // We insert this check in the lowered form only if necessary.
+            val stepType = data.stepType(context.irBuiltIns)
+            val stepGreaterFun = context.irBuiltIns.greaterFunByOperandType[stepType.toKotlinType()]!!
+            val zeroStep = if (data == ProgressionType.LONG_PROGRESSION) irLong(0) else irInt(0)
+            val throwIllegalStepExceptionCall = irCall(context.irBuiltIns.illegalArgumentExceptionSymbol, stepType).apply {
```
madsager: Maybe move the construction of this to the case where it is needed to not build the IR for it when it is not needed?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/handlers/UntilHandler.kt
```diff
@@ -0,0 +1,167 @@
...
+            // However, `B` may be an expression with side-effects that should only be evaluated once, and `A` may also have side-effects.
+            // They are evaluated once and in the correct order (`A` then `B`), the final lowered form is:
+            //
+            //   // Additional variables
+            //   val untilReceiverValue = A
+            //   val untilArg = B
+            //   // Standard form of loop over progression
+            //   var inductionVar = untilReceiverValue
+            //   val last = untilArg - 1
+            //   if (inductionVar <= last && untilFunArg != MIN_VALUE) {
```
madsager: untilFunArg -> untilArg

compiler/testData/codegen/bytecodeText/forLoop/primitiveProgression.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
madsager: Could you add comments to the tests that now fail on JVM_IR, so it is easy for us to see why these fail and if that is by design?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/handlers/StepHandler.kt
punzki: Easy enough, converted to lambda

punzki:
@max-kammerer Updated the PR. Had to resolve merge conflicts from @madsager 's changes (#2541, #2547, #2552, #2560).

Note that one of the commits (10ab88450216c8089670d5135aa148a25cd9cc56) split up the `ProgressionHandlers.kt` file which was getting pretty big. It might be easier to review without that commit.

punzki:
> Note that one of the commits ([10ab884](https://github.com/JetBrains/kotlin/commit/10ab88450216c8089670d5135aa148a25cd9cc56)) split up the `ProgressionHandlers.kt` file which was getting pretty big. It might be easier to review without that commit.

Updated the PR without the last commit to split up, to make it easier to review and manage potential conflicts. I'll split it up separately.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt
```diff
@@ -108,6 +105,12 @@ abstract class Symbols<out T : CommonBackendContext>(val context: T, private val
     val progressionClasses = listOf(charProgression, intProgression, longProgression)
     val progressionClassesTypes = progressionClasses.map { it.descriptor.defaultType }.toSet()
 
+    val getProgressionLastElementByReturnType = builtInsPackage("kotlin", "internal").getContributedFunctions(
+        Name.identifier("getProgressionLastElement"),
+        NoLookupLocation.FROM_BACKEND
+    ).filter { it.containingDeclaration !is BuiltInsPackageFragment }
+        .mapNotNull { Pair(it.returnType!!, symbolTable.referenceSimpleFunction(it)) }.toMap()
```
max-kammerer: Always not null values

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -219,6 +220,272 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
...
+            // Retrieve the HeaderInfo from the underlying progression (if any).
+            val nestedInfo = expression.extensionReceiver!!.accept(visitor, null) as? ProgressionHeaderInfo
+                ?: return null
+
+            val stepArg = expression.getValueArgument(0)!!
+            // We can return the nested info if its step is constant and its absolute value is the same as the step argument. Examples:
+            //
+            //   1..10 step 1               // Nested step is 1, argument is 1. Equivalent to `1..10`.
+            //   10 downTo 1 step 1         // Nested step is -1, argument is 1. Equivalent to `10 downTo 1`.
+            //   10 downTo 1 step 2 step 2  // Nested step is -2, argument is 2. Equivalent to `10 downTo 1 step 2`.
```
max-kammerer: Is it possible to add next failing test (if they not present yet)
with  ```10 downTo 1 step 2 step -2```  and  ```1..10 step -2 step 2``` ?

max-kammerer:
@punzki Maybe it worth to write some test generator similar to GenerateInRangeExpressionTestData/GeneratePrimitiveVsObjectEqualityTestData

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
```diff
@@ -3,16 +3,9 @@ fun foo() {
     val range = 1 until 2
     for (i in range) {
     }
-
-    for (i in 1..2 step 4) {}
```
max-kammerer: Why it's removed?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
punzki: I have both already; see `compiler/testData/codegen/box/ranges/stepped/expression/downTo/legalStepThenIllegalStep.kt` and `compiler/testData/codegen/box/ranges/stepped/expression/rangeTo/illegalStepThenLegalStep.kt` for all various combinations. Is that what you're referring to?

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
punzki: I removed it because the behavior is different now with JVM IR (no calls to getFirst/Last). But since now we have the JVM_IR_TEMPLATES I'll just change the file.

punzki:
> @punzki Maybe it worth to write some test generator similar to GenerateInRangeExpressionTestData/GeneratePrimitiveVsObjectEqualityTestData

It's going to be tricky -- the tests aren't perfectly "symmetrical" and many of the values are hand-picked to cover the edge cases -- but I'll see what I can do. It'll take a while though.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt
max-kammerer: @punzki Could you change mapNotNull to simple map?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
max-kammerer: Yes)

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
max-kammerer: Yes, please

punzki:
@max-kammerer I updated the PR but am still working on the test generator. In the meantime, I found the new tests caused problems with JS (non-IR). Filed this bug: https://youtrack.jetbrains.com/issue/KT-34166. For now, I disabled the affected tests for JS. I also had to fix duplicate IR element errors.

max-kammerer:
@punzki Thank you for a bug) Actually test generator could be postponed for another PR so this one could be merged

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
```diff
@@ -12,7 +12,19 @@ fun foo() {
 // 1 INVOKEVIRTUAL kotlin/ranges/IntRange.getLast \(\)I
 // 0 INVOKEVIRTUAL kotlin/ranges/IntRange.getLast \(\)Ljava/lang/Integer;
 
+// JVM non-IR does NOT specifically handle "step" progressions. The stepped progression in the above code are constructed and its
+// first/last/step properties are retrieved.
+// JVM IR has an optimized handler for "step" progressions and elides the construction of the stepped progressions.
+
+// JVM_TEMPLATES
 // 1 INVOKEVIRTUAL kotlin/ranges/IntProgression.getFirst \(\)I
```
max-kammerer: Please correct templates to use simple function name for 0 count (like in primitiveProgression.kt):
e.g. 
//0 INVOKEVIRTUAL kotlin/ranges/IntProgression.getFirst \(\)Ljava/lang/Integer; 
->
//0 getFirst

punzki: It seems that the origin and purpose of the test is to make sure the return types for `getFirst/last/step` are not boxed. I'm splitting this up into 2 test files (range and progression) so that it's clearer.

punzki:
@max-kammerer - Finally finished the test generator! Hope it looks good. 🙂 It actually caught some copy-paste issues and a missing test.

I also added the existing generators (`GenerateInRangeExpressionTestData` and `GeneratePrimitiveVsObjectEqualityTestData`) in `GenerateCompilerTests` so that they are invoked during test generation. From what I can tell, they used to have to be invoked manually. I couldn't find any code references to them anywhere -- maybe I missed it so please correct if I'm mistaken about them!

punzki:
@max-kammerer - Gentle ping... Let me know if you are able to look and have any other feedback!

max-kammerer:
@punzki Thank you for ping and sorry for defaly, will look in next days!

max-kammerer:
> I also added the existing generators (GenerateInRangeExpressionTestData and GeneratePrimitiveVsObjectEqualityTestData) in GenerateCompilerTests so that they are invoked during test generation. From what I can tell, they used to have to be invoked manually. I couldn't find any code references to them anywhere -- maybe I missed it so please correct if I'm mistaken about them!

Yes, they was called manually. Thank for change!

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
max-kammerer: Minor, but I would preferer
```
// 1 INVOKEVIRTUAL kotlin/ranges/IntProgression.getFirst \(\)I
// 1 getFirst
and so on
```
and same in intProgressionNoBoxing.kt

compiler/tests/org/jetbrains/kotlin/generators/tests/GenerateInRangeExpressionTestData.kt
```diff
@@ -28,17 +17,18 @@ object GenerateInRangeExpressionTestData {
...
+        elementExpressions: List<String>,
+        header: String = ""
     ) {
         PrintWriter(File(GENERATED_DIR, fileName)).use {
             it.generateTestCaseBody(header, rangeExpressions, elementExpressions)
         }
     }
 
     private fun PrintWriter.generateTestCaseBody(header: String, rangeExpressions: List<String>, elementExpressions: List<String>) {
+        println("// KJS_WITH_FULL_RUNTIME")
```
max-kammerer: Why this change is required? Are affected tests was broken?

max-kammerer: Or this line was added manually for non affected?

max-kammerer:
@punzki Is it possible to merge tests in more bigger pieces to decrease test count?

max-kammerer:
Please also note that  after #2657 merge there is a problem with test
```
Caused by: java.lang.AssertionError: Unexpected IR element found during code generation. Either code generation for it is not implemented, or it should have been lowered: STRING_CONCATENATION type=kotlin.String
```

udalov:
In addition to Mike's question above, I wonder if it's possible to minimize the total amount of test code in the PR? Specifically, can we reduce the number of cases tested for different combinations of steps in succession, and in combination with `reversed`? Also, are all tests equally relevant both for expression and literal forms?

My main concern is the execution time of box tests, both locally and on the build server. It's probably not a big deal if we do need that many tests, but it'd be nice if they didn't take major time on each run of all black box tests.

compiler/tests/org/jetbrains/kotlin/generators/tests/GenerateInRangeExpressionTestData.kt
punzki: It was added manually before for the "in range" tests, instead of through this test generator. It was only added for the tests that needed it; there are 6 affected tests. Now we want this generator to be called during test generation. It was cleaner to add the directive for all 14 tests generated by this generator—whether it was needed or not for the test—rather than adding it only for the 6 tests that needed it.

punzki:
> @punzki Is it possible to merge tests in more bigger pieces to decrease test count?

Do you mean something like combining all the "illegal step" test files (there are currently 5) into a single test file? And maybe similarly for nested, step+reversed, etc? Would that make a 

> In addition to Mike's question above, I wonder if it's possible to minimize the total amount of test code in the PR? Specifically, can we reduce the number of cases tested for different combinations of steps in succession, and in combination with `reversed`? Also, are all tests equally relevant both for expression and literal forms?

There was some bug mentioned in the native `ForLoopsLowering` (now deleted) which described a bug in handling nested step loops, which is why I added those tests. I combined those with `reversed` to make sure we handle the calculations correctly. The common lowering now handles those nested-call progressions, and the native version did not.

The handling of `step` is different when the underlying progression is a literal vs some other expression, because the direction of the underlying progression is either known (for literal) or unknown. The certainty (or lack of) in direction propagates across additional `step` and `reverse` calls. So the tests are relevant.

There is certainly some redundant coverage across literal and expression forms, even across `rangeTo/..` and `until`, but I am not comfortable picking and choosing the minimum number of tests to ensure enough coverage (also I don't think I'm smart enough to choose wisely :slightly_smiling_face: ). In addition, there can be changes in the future in which the reduced coverage can come back and bite us.

> My main concern is the execution time of box tests, both locally and on the build server. It's probably not a big deal if we do need that many tests, but it'd be nice if they didn't take major time on each run of all black box tests.

I experimented with taking out some tests and ran all the codegen box tests (IR and non-IR), and measured the test time only:

- Baseline, no changes: 9757 tests, took 2m 50s
- Without "expression" tests: 9563 tests, took 2m 47s (2% fewer tests, 1.8% faster)
- Without all "stepped" ("expression" and "literal") tests: 9359 tests, took 2m 45s (4% fewer tests, 3% faster)

Do you think the difference is significant? The tests are running in parallel by test class (i.e., test data directories). I can continue to experiment in consolidating the tests (as Mike suggested), or even splitting up the tests into more directories.

punzki:
> Please also note that after #2657 merge there is a problem with tests in master
> 
> ```
> Caused by: java.lang.AssertionError: Unexpected IR element found during code generation. Either code generation for it is not implemented, or it should have been lowered: STRING_CONCATENATION type=kotlin.String
> ```

Fixed. I moved ForLoopsLowering before JvmStringConcatenationLowering in the JVM phases. ForLoopsLowering may produce IrStringConcatenations which should be
eliminated before codegen.

udalov:
Fair enough, thanks for the explanation! Let's keep the tests, but it'd be awesome if some of them can be merged like you suggest.

One other thing: consider ignoring the directories with generated tests for `AbstractLightAnalysisModeTest` in `GenerateCompilerTests`. These tests check the light analysis mode (where method bodies are not generated) on different kinds of declarations, but they won't do anything useful here because each test only consists of the single function `box`.

compiler/testData/codegen/bytecodeText/intRangeNoBoxing.kt
punzki: @max-kammerer Updated the test

punzki:
> Fair enough, thanks for the explanation! Let's keep the tests, but it'd be awesome if some of them can be merged like you suggest.

Hmm, the generator I wrote doesn't support multiple tests per file. I could refactor it but I think it would be messy. It'd be easier to understand and debug tests in the future if there's one per file. Hope it's ok that I just left it as-is.

> One other thing: consider ignoring the directories with generated tests for `AbstractLightAnalysisModeTest` in `GenerateCompilerTests`. These tests check the light analysis mode (where method bodies are not generated) on different kinds of declarations, but they won't do anything useful here because each test only consists of the single function `box`.

Didn't know that; thanks for the info. Done!

max-kammerer:
@punzki Great work! Many thanks!

max-kammerer:
@punzki Could you take a look: there is a problem with test regeneration on Windows because of another line separator - all generated files are marked as changed cause of it. Problem is only with new generator

punzki:
> @punzki Could you take a look: there is a problem with test regeneration on Windows because of another line separator - all generated files are marked as changed cause of it. Problem is only with new generator

Can you describe the issue further? Do you have a TeamCity log link, or something like that? I do not have a Windows machine easily available. Also I don't really understand what the issue is...

max-kammerer:
@punzki Actually it's reproduced on attempt to run test generator on Windows machine (this task is not executed on Teamcity). Actually i'm using Linux too: so look like println insert \n\r on Windows that affects all generated testdata for new stepped tests

PR #2419, opened 2019-06-19
KT-17916: Import popup does not indicate deprecated classes

Ghedeon:
https://youtrack.jetbrains.com/issue/KT-17916

Currently, the `KotlinAddImportAction` doesn't highlight deprecated Psi elements. At the moment I just copy-pasted the java implementation from `AddImportAction.java`.

|Before|After|
|-----|------|
|class_before|class_kt|
|function_before|function_kt|

Something to consider: feels like common logic from `KotlinAddImportAction.kt`, `AddImportAction.java` and `StaticImportMethodQuestionAction.java` could be abstracted in one place but somebody with better knowledge about the project should evaluate this.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
@@ -21,11 +21,11 @@ import com.intellij.codeInsight.daemon.impl.ShowAutoImportPass
 import com.intellij.codeInsight.daemon.impl.actions.AddImportAction
 import com.intellij.codeInsight.hint.HintManager
 import com.intellij.codeInsight.hint.QuestionAction
+import com.intellij.ide.util.DefaultPsiElementCellRenderer
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.module.ModuleUtilCore
 import com.intellij.openapi.project.Project
-import com.intellij.openapi.ui.popup.JBPopupFactory
 import com.intellij.openapi.ui.popup.PopupStep
 import com.intellij.openapi.ui.popup.util.BaseListPopupStep
 import com.intellij.psi.PsiDocumentManager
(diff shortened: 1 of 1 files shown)


idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
```diff
@@ -155,7 +161,31 @@ class KotlinAddImportAction internal constructor(
             return true
         }
 
-        JBPopupFactory.getInstance().createListPopup(getVariantSelectionPopup()).showInBestPositionFor(editor)
+        val popup = object : ListPopupImpl(getVariantSelectionPopup()) {
+            @Suppress("UNCHECKED_CAST")
```
Ghedeon: Thoughts on how to avoid this?

idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
```diff
@@ -155,7 +161,31 @@ class KotlinAddImportAction internal constructor(
...
+                                index,
+                                isSelected,
+                                cellHasFocus
+                            )
+                        )
+                    }
+                }
+            }
+        }
+        NavigationUtil.hidePopupIfDumbModeStarts(popup, project)
```
Ghedeon: Present in java action but was missing here. Is it still required?

idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
dimonchik0036: `as PopupListElementRenderer`

idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
dimonchik0036: No, it is unnecessary.

idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
Ghedeon: fixed

idea/src/org/jetbrains/kotlin/idea/actions/KotlinAddImportAction.kt
Ghedeon: fixed

dimonchik0036:
Thanks!

PR #2420, opened 2019-06-19
KT-32084 Optimize standard library collections extensions that return `List` 

ajalt:
https://youtrack.jetbrains.com/issue/KT-32084

The first commit in this PR fixes a bug that caused mixed line endings in code generation on Windows, which was required for me to generate the main changes in this PR. Let me know if you'd like that moved to a separate PR.

### DIFF
--- a/libraries/stdlib/common/src/generated/_Collections.kt
+++ b/libraries/stdlib/common/src/generated/_Collections.kt
@@ -1311,11 +1311,8 @@ public inline fun <T, K> Iterable<T>.groupingBy(crossinline keySelector: (T) -> 
  */
 public inline fun <T, R> Iterable<T>.map(transform: (T) -> R): List<R> {
     val collectionSize = collectionSizeOrDefault(10)
-    return when (collectionSize) {
-        0 -> emptyList()
-        1 -> listOf(transform(iterator().next()))
-        else -> mapTo(ArrayList<R>(collectionSize), transform)
-    }
+    if (collectionSize == 0) return emptyList()
+    return mapTo(ArrayList<R>(collectionSize), transform).optimizeReadOnlyList()
 }
 
 /**
@@ -1326,11 +1323,8 @@ public inline fun <T, R> Iterable<T>.map(transform: (T) -> R): List<R> {
  */
 public inline fun <T, R> Iterable<T>.mapIndexed(transform: (index: Int, T) -> R): List<R> {
     val collectionSize = collectionSizeOrDefault(10)
-    return when (collectionSize) {
-        0 -> emptyList()
-        1 -> listOf(transform(0, iterator().next()))
-        else -> mapIndexedTo(ArrayList<R>(collectionSize), transform)
-    }
+    if (collectionSize == 0) return emptyList()
+    return mapIndexedTo(ArrayList<R>(collectionSize), transform).optimizeReadOnlyList()
 }
 
 /**
--- a/libraries/stdlib/common/src/generated/_Collections.kt
+++ b/libraries/stdlib/common/src/generated/_Collections.kt
@@ -663,14 +663,14 @@ public inline fun <T> Iterable<T>.dropWhile(predicate: (T) -> Boolean): List<T> 
             list.add(item)
             yielding = true
         }
-    return list
+    return list.optimizeReadOnlyList()
 }
 
 /**
  * Returns a list containing only elements matching the given [predicate].
  */
 public inline fun <T> Iterable<T>.filter(predicate: (T) -> Boolean): List<T> {
-    return filterTo(ArrayList<T>(), predicate)
+    return filterTo(ArrayList<T>(), predicate).optimizeReadOnlyList()
 }
 
 /**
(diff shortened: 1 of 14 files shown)


libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -217,6 +217,7 @@ public inline fun <C, R> C.ifEmpty(defaultValue: () -> R): R where C : Collectio
 @kotlin.internal.InlineOnly
 public inline fun <@kotlin.internal.OnlyInputTypes T> Collection<T>.containsAll(elements: Collection<T>): Boolean = this.containsAll(elements)
 
+@PublishedApi
```
ilya-g: `optimizeReadOnlyList` was introduced after the release of Kotlin 1.0. We cannot make it public and reference it from use code (via inlining) while we support `-api-version 1.0` option.

libraries/stdlib/common/src/generated/_Collections.kt
```diff
@@ -1320,7 +1325,12 @@ public inline fun <T, R> Iterable<T>.map(transform: (T) -> R): List<R> {
  * and returns the result of the transform applied to the element.
  */
 public inline fun <T, R> Iterable<T>.mapIndexed(transform: (index: Int, T) -> R): List<R> {
-    return mapIndexedTo(ArrayList<R>(collectionSizeOrDefault(10)), transform)
+    val collectionSize = collectionSizeOrDefault(10)
+    return when (collectionSize) {
+        0 -> emptyList()
+        1 -> listOf(transform(0, iterator().next()))
+        else -> mapIndexedTo(ArrayList<R>(collectionSize), transform)
```
ilya-g: Such change makes `transform` lambda body inlined twice into the call site. It can lead to a significant bytecode size growth and is undesirable.

libraries/stdlib/src/kotlin/collections/Collections.kt
ajalt: Ah, I didn't know that. So should I manually copy its implementation into the inline functions I added it to, or is there a better solution?

ilya-g: One solution is to wait when we drop support for `-api-version 1.0`.

Another option is to make an inline function that will either call `optimizeReadOnlyList` or do nothing, depending on the api version specified during the compilation. This technique is used here for example: https://github.com/JetBrains/kotlin/blob/v1.3.40/libraries/stdlib/jvm/src/generated/_ArraysJvm.kt#L393

ilya-g:
I think it'd better to extract the commit about line endings from this PR and discuss it separately.

libraries/stdlib/src/kotlin/collections/Collections.kt
ajalt: The api version check sounds good, but `apiVersionIsAtLeast` is JVM-only, so it can't currently be used from these functions. I can make an `expect` declaration for the new inline function to have the JVM implementation do a version check, and have the other platforms do nothing. That doesn't feel like a great solution, though. Is there a better way forward?

ilya-g: `expect`/`actual` is a valid option. Other platforms have `optimizeReadOnlyList` since they become stable so there's no need to use `apiVersionIsAtLeast` there at all, just call that function.

ajalt:
I added a wrapper around `optimizeReadOnlyList` to preserve binary compatibility. I followed the existing example of `checkIndexOverflow`, and added an `actual` implementation to the `jvm` and `js` source sets. If there are any other source sets that need an `actual` implementation, please let me know.

I named the wrapper `optimizeReadOnlyListCompat`, which is not a great name. I'd love a suggestion on something better to call it.

libraries/stdlib/js/src/kotlin/collections.kt
```diff
@@ -164,7 +165,10 @@ internal actual inline fun <T> Array<out T>.copyToArrayOfAny(isVarargs: Boolean)
     else
         this.copyOf()
 
-
+@PublishedApi
+@SinceKotlin("1.3")
```
ajalt: Should this be `"1.4"`, or some other value?

libraries/stdlib/jvm/src/kotlin/collections/CollectionsJVM.kt
```diff
@@ -45,6 +45,14 @@ internal actual fun <T> Array<out T>.copyToArrayOfAny(isVarargs: Boolean): Array
     else
         java.util.Arrays.copyOf(this, this.size, Array<Any?>::class.java)
 
+@PublishedApi
+@SinceKotlin("1.3")
+@InlineOnly
+internal actual inline fun <T> List<T>.optimizeReadOnlyListCompat(): List<T> =
+    if (apiVersionIsAtLeast(1, 1, 0))
```
ajalt: I wasn't sure what version `optimizeReadOnlyList` was added. If you do, let me know and I'll change this check to the correct value.

PR #2163, opened 2019-03-08
KT-30343 Add new Quarkus preset to all-open compiler plugin

geoand:

PR #2421, opened 2019-06-20
KT-30343 Make quarkus compiler plugin available

geoand:
Follow up of #2163

### DIFF
--- a/libraries/tools/kotlin-maven-allopen/src/main/kotlin/AllOpenMavenPluginExtensions.kt
+++ b/libraries/tools/kotlin-maven-allopen/src/main/kotlin/AllOpenMavenPluginExtensions.kt
@@ -57,3 +57,22 @@ class KotlinSpringMavenPluginExtension : KotlinMavenPluginExtension {
         return listOf(PluginOption("spring", ALLOPEN_COMPILER_PLUGIN_ID, PRESET_ARG_NAME, "spring"))
     }
 }
+
+@Component(role = KotlinMavenPluginExtension::class, hint = "quarkus")
+class KotlinQuarkusMavenPluginExtension : KotlinMavenPluginExtension {
+    private companion object {
+        val PRESET_ARG_NAME = "preset"
+    }
+
+    override fun getCompilerPluginId() = ALLOPEN_COMPILER_PLUGIN_ID
+
+    @Requirement
+    lateinit var logger: Logger
+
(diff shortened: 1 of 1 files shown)


geoand:
Would it be possible to include this in a `1.3.41` release if that happens?

geoand:
Any news on this folks?

Dufgui:
I ping you to be more viewable. @yanex @abelkov

yanex:
Hello @geoand and @Dufgui!
While we have out-of-the-box support for Spring, we decided we are not ready to support additional frameworks right in the Kotlin project. What do you think about making the Quarkus plugin as a community one? It seems it only needs to provide a couple of compiler arguments to kotlinc, so it should be possible even in the current Kotlin & Maven/Gradle plugins.

geoand:
@yanex Hi!

I see, however #2163 was merged so it wouldn't it make sense to go all the way and add this as well?

yanex:
Well, we hadn't discussed this team-wide until now. If you think that keeping 'quarkus' preset alone doesn't make sense, we can move the annotation list to the community plugin as well.
Please contact us if you need any help. I persontally don't have much expertize in Maven plugin development, but other people from the team might be able to help you.

geoand:
I see. Thanks for the info. Hopefully we will get around to it soon and we need help we'll reach out :)

jreznot:
@geoand As far as I know, Micronaut projects with Kotlin are configured with:
```
<configuration>
  <compilerPlugins>
    <plugin>all-open</plugin>
    <plugin>no-arg</plugin>
    <plugin>jpa</plugin>
  </compilerPlugins>
  <pluginOptions>
    <option>all-open:annotation=io.micronaut.aop.Around</option>
  </pluginOptions>
</configuration>
```

Examples: https://github.com/jreznot/micronaut-microservices-poc/blob/master/documents-service/pom.xml#L232

geoand:
@jreznot Thanks for the info. We are doing something similar in the generated kotlin projects. However I would like to make it even easier for users - just like the spring compiler plugin where the user doesn't have to specify anything extra, just that they want the spring "profile" and everything else is added magically :)

yanex:
Unfortunately, I have to close the pull request as the team decided not to add more framework-specific wrappers on top of the all-open plugin.
@geoand, please ping me if you encounter any problems during the implementation of the community plugin.

geoand:
@yanex I will let you know, thanks

PR #2422, opened 2019-06-20
KT-31840: Caching with KAPT incremental annotation processing

gavra0:
Fix caching for incremental annotation processing

Artifact transform that uses ASM to analyze KAPT classpath stored absolute
paths in the output artifact. This resulted in remote build cache misses.

This commit changes how analysis is stored. Actual analysis file is the
output of the transform, and there is not need to use a marker file any more.
Output does not store the classpath entry absolute path. Instead, it uses
task action incremental information to find analysis outputs that changed.

Additionally, class that handles analysis snapshot comparison has been
simplified, and lazy loading of the structural information is handled in
a more straightforward way.


yanex:
Merged manually to 1.3.50 (293eb56a9fe4e3c5ba25bbb8609af7a32650653e, 226bdfe993c498e063ac1789a8cd59a333fb7740) and master (251d8ccd97b65ad9f3dc7733d5a960f4c6dc2cef and 5cab6fd6cf430fb91128e5dbdce7a2936f7b449d). Thanks for the contribution!

PR #2423, opened 2019-06-21
KT-16602: Provide samples for sorting API usage

Dattish:
https://youtrack.jetbrains.com/issue/KT-16602


libraries/stdlib/samples/test/samples/collections/sort.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        @Sample
+        fun sortRangeOfArray() {
+            val intArray = intArrayOf(4, 3, 2, 1)
+
+            assertPrints(
+                intArray.joinToString(),
+                "4, 3, 2, 1"
+            )
```
ilya-g: `assertPrints` invocations should be kept in a single line. They are post-processed into `println()` calls when being converted to runnable samples.

libraries/stdlib/samples/test/samples/collections/sort.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        @Sample
+        fun sortRangeOfComparable() {
+            class Envelope(val sent: LocalDateTime) : Comparable<Envelope> {
+                override fun compareTo(other: Envelope): Int {
+                    return sent.compareTo(other.sent)
+                }
+
+                override fun toString(): String {
```
ilya-g: The class can be made `data class`, that would generate the same `toString` implementation

libraries/stdlib/samples/test/samples/collections/sort.kt
```diff
@@ -0,0 +1,138 @@
...
+
+            assertPrints(
+                mail.joinToString(),
+                "Envelope(sent=2019-06-21T10:39), Envelope(sent=2019-06-21T10:40), Envelope(sent=2019-06-21T10:41)"
+            )
+        }
+
+        @Sample
+        fun sortRangeOfComparable() {
+            class Envelope(val sent: LocalDateTime) : Comparable<Envelope> {
```
ilya-g: The class can be simplified by using another comparable type for its fields, that doesn't require platform-specific imports. For example, `Person(val firstName: String, val lastName: String)`

libraries/stdlib/samples/test/samples/collections/sort.kt
```diff
@@ -0,0 +1,138 @@
...
+            mutableList.sort()
+
+            assertPrints(
+                mutableList.joinToString(),
+                "1, 2, 3, 4"
+            )
+        }
+
+        @Sample
+        fun sortComparable() {
```
ilya-g: Better to name `sortArrayOfComparable`, because it's _array_ being sorted, and not a comparable type.

libraries/stdlib/samples/test/samples/collections/sort.kt
```diff
@@ -0,0 +1,138 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package samples.collections
+
+import samples.*
+import java.time.LocalDateTime
+
+@RunWith(Enclosed::class)
+class Sort {
```
ilya-g: Please use another placement schema, similar to other collection and array samples.
Collection and List samples should go into `collections.kt`, and `Sorting` nested class can be created there, so its full name would be `samples.collections.Collections.Sorting`. Ditto for the array sorting samples.

Dattish:
I think I managed to address your concerns in my latest commit. Please let me know if I missed or misunderstood anything.

ilya-g:
The follow-up commit with improvements: https://github.com/JetBrains/kotlin/commit/4fc942c5e910552e26eeb8dc9360cdc13a426e9a

PR #2424, opened 2019-06-21
KT-31291: Wrap ASM version in a method to avoid having it inlined

gavra0:
This commit adds a method to get the value of org.jetbrains.org.objectweb.asm.Oopcodes.API_VERSION.
In general, this should not be a problem, but some Gradle plugins package
Jetbrains ASM version. At runtime, their version may be loaded which may be
different than the one this code was compiled against.


yanex:
Merged manually to 1.3.50 (d19fdb2008a5e94acae7c55e07142134face1cdc) and master (4caddd2c1ca080a3d51f595fe260046d88ce23c9) branches. Thank you for the contribution!

PR #2426, opened 2019-06-24
KT-27550 "Redundant explicit this" false positive with subclass and extension lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27550


mglukhikh:
Merged 👍

PR #2427, opened 2019-06-24
JVM_IR: do not generate two nullary constructors

pyos:
when the primary constructor with @JvmOverloads has default values for all arguments.

Bugfix for #2372.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
@@ -115,7 +115,6 @@ val jvmPhases = namedIrFilePhase<JvmBackendContext>(
             annotationPhase then
             tailrecPhase then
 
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -146,6 +145,7 @@ val jvmPhases = namedIrFilePhase<JvmBackendContext>(
             collectionStubMethodLowering then
             bridgePhase then
             jvmOverloadsAnnotationPhase then
+            jvmDefaultConstructorPhase then
```
max-kammerer: Please add 'jvmOverloadsAnnotationPhase' prerequisite for jvmDefaultConstructorPhase phase

pyos: done

max-kammerer:
Thank you!

PR #2429, opened 2019-06-24
Enable inlining of external functions with inline class parameters in the JVM IR backend

sfs:
In particular, this fixes unsigned types in the JVM IR backend, which in turn fixes a number of tests.


sfs:
Updated to unmute an additional test.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
```diff
@@ -126,6 +131,12 @@ fun IrFunctionBuilder.buildConstructor(): IrConstructor {
     }
 }
 
+inline fun buildFunWithDescriptor(originalDescriptor: FunctionDescriptor, b: IrFunctionBuilder.() -> Unit): IrFunctionImpl =
```
max-kammerer: Please rename 'b' to something meaningful: builder?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
```diff
@@ -149,7 +150,7 @@ class MemoizedInlineClassReplacements {
     }
 
     private fun buildReplacement(function: IrFunction, body: IrFunctionImpl.() -> Unit) =
-        buildFun {
+        buildFunWithDescriptor(function.descriptor) {
```
max-kammerer: Could you clarify why descriptor is reuqired here?

sfs: I changed the function to use a more descriptive name and added a comment to explain what the descriptor does.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
sfs: I added another commit to do this refactoring in the whole file, since it seems inconsistent to only use a different name in this one function.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
```diff
@@ -56,6 +57,13 @@ object InlineClassAbi {
      * to avoid clashes between overloaded methods.
      */
     fun mangledNameFor(irFunction: IrFunction): Name {
+        val suffix = when {
```
max-kammerer: @ilmirus Could you take a look there is similar code that is used in MemoizedInlineClassReplacements from https://github.com/JetBrains/kotlin/commit/7a096b305f11bece9e39ffd8875a766db5df677a

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
max-kammerer: @sfs Thanks for renaming!

But it's still not clear for me why descriptor should be passed and its container source should be keept.
Could it be fixed purely with ir approach? 
Maybe same JvmInlineClassLowering should be execute  for external declarations (at least for their inline parts, as we do in InterfaceDelegationLowering.generateDelegationToDefaultImpl: default method stubs are also constructed for external declarations without actual lowering of external declaration)? cc @udalov

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/InlineClassAbi.kt
sfs: There's still a problem with this commit. The methods of the result class have to be mangled (e.g. the compiled function is called `Result.getOrNull-impl` instead of `Result.getOrNull`). Right now it just disables mangling entirely. On the other hand, my patch did replace functions unnecessarily in some cases. I've updated the PR to fix both problems.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
sfs: The culprit here is `InlineCodegen`, which relies on descriptors to recognize external declarations (see [InlineCodegen.kt#L557](https://github.com/JetBrains/kotlin/blob/728459a5046759a91075f3e03dba21076133feb5/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt#L557)).

Alternatively we could override this method in IrInlineCodegen, but this would end up copying some code right now.

The inline class lowering already executes (lazily) for external declarations.

max-kammerer:
@sfs There are some failing tests after rebase on master: https://teamcity.jetbrains.com/viewLog.html?buildId=2388320&buildTypeId=Kotlin_dev_Compiler
Could be linked with https://github.com/JetBrains/kotlin/commit/7a096b305f11bece9e39ffd8875a766db5df677a
mentioned above

sfs:
@max-kammerer I noticed, that's what the force push to 7239812 fixed. The explanation is in my answer to your comment above. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
max-kammerer: @sfs Actually I would prefer to avoid descriptor usage cause at some point we want to get rid  descriptors off. But this work seems would depend on https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3069 cause otherwise it would be impossible to extract proper facade name

max-kammerer: And from another point this will allow us to support another cases of inlining from binaries (especially to support CompileKotlinAgainstInlineKotlinTestGenerated in Ir in future)

sfs: I don't think the usage of descriptors here is such a problem - the changes will simply become obsolete once IrInlineCodegen no longer depends on descriptors.

In the meantime though, this PR allows us to start fixing tests that rely on inlined external declaration with inline class arguments (such as all coroutine tests using `Result<T>`) and to actually use inline classes and inline functions together with the IR backend.

max-kammerer: @sfs After short investigation I've realized that this work will require new type mapper so pr could be merged as is. And further changes in inliner I will made by myself

max-kammerer:
@sfs Many thanks!

PR #2431, opened 2019-06-24
Reduce calls to Object.hashCode from ConstraintSystem(Builder)Impl

jvgogh:
Building on the work in #2403, shaves 1% of compilation time on kotlinx.serialization
by using SmartIdentityTable instead of hashMap in ConstraintSystemImpl and ConstraintSystemBuilderImpl

Metrics on the prior hashMap use in this code:

ConstraintSystemImpl:

| What      | Count |
| --------- | ----- |
| instances | 3223  |
| min       | 0     |
| max       | 3     |
| mean      | 1     |
| stddev    | 0.57  |
| median    | 1     |
| 90p       | 2     |
| 95p       | 2     |

ConstraintSystemBuilderImpl:

| What      | Count |
| --------- | ----- |
| instances | 3227  |
| min       | 0     |
| max       | 3     |
| mean      | 0     |
| stddev    | 1.04  |
| median    | 1     |
| 90p       | 2     |
| 95p       | 2     |

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
@@ -44,6 +44,7 @@ import org.jetbrains.kotlin.types.model.TypeSystemInferenceExtensionContext
 import org.jetbrains.kotlin.types.typeUtil.builtIns
 import org.jetbrains.kotlin.types.typeUtil.defaultProjections
 import org.jetbrains.kotlin.types.typeUtil.isDefaultBound
+import org.jetbrains.kotlin.utils.SmartIdentityTable
 import org.jetbrains.kotlin.utils.addToStdlib.cast
 import java.lang.IllegalArgumentException
 import java.lang.IllegalStateException
@@ -64,7 +65,7 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
         EQUAL(EXACT_BOUND)
     }
 
-    internal val allTypeParameterBounds = LinkedHashMap<TypeVariable, TypeBoundsImpl>()
+    internal val allTypeParameterBounds = SmartIdentityTable<TypeVariable, TypeBoundsImpl>()
     internal val usedInBounds = HashMap<TypeVariable, MutableList<TypeBounds.Bound>>()
     internal val errors = ArrayList<ConstraintError>()
     internal val initialConstraints = ArrayList<Constraint>()
@@ -99,11 +100,11 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
             }
         }
 
-        for ((_, typeVariable) in typeParameters.zip(typeVariables)) {
-            allTypeParameterBounds.put(typeVariable, TypeBoundsImpl(typeVariable))
+        for (typeVariable in typeVariables) {
+            allTypeParameterBounds[typeVariable] = TypeBoundsImpl(typeVariable)
         }
 
-        for ((typeVariable, _) in allTypeParameterBounds) {
+        for (typeVariable in allTypeParameterBounds.keys) {
             for (declaredUpperBound in typeVariable.freshTypeParameter.upperBounds) {
                 if (declaredUpperBound.isDefaultBound()) continue //todo remove this line (?)
                 val context = ConstraintContext(TYPE_BOUND_POSITION.position(typeVariable.originalTypeParameter.index))
@@ -405,7 +406,7 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
                         "Calls of the first system: ${typeVariableSubstitutors.keys}, second: ${other.typeVariableSubstitutors.keys}"
             )
         }
-        if (!Collections.disjoint(other.allTypeParameterBounds.keys, allTypeParameterBounds.keys)) {
+        if (!Collections.disjoint(other.allTypeParameterBounds.keys.toList(), allTypeParameterBounds.keys.toList())) {
             throw IllegalArgumentException(
                 "Combining two constraint systems only makes sense when they have no common variables. " +
                         "First system variables: ${allTypeParameterBounds.keys}, second: ${other.allTypeParameterBounds.keys}"
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
@@ -30,17 +30,15 @@ import org.jetbrains.kotlin.types.*
 import org.jetbrains.kotlin.types.TypeUtils.DONT_CARE
 import org.jetbrains.kotlin.types.checker.KotlinTypeChecker
 import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
-import java.util.*
+import org.jetbrains.kotlin.utils.SmartIdentityTable
 
 internal class ConstraintSystemImpl(
-    private val allTypeParameterBounds: Map<TypeVariable, TypeBoundsImpl>,
+    private val allTypeParameterBounds: SmartIdentityTable<TypeVariable, TypeBoundsImpl>,
     private val usedInBounds: Map<TypeVariable, MutableList<TypeBounds.Bound>>,
     private val errors: List<ConstraintError>,
     private val initialConstraints: List<ConstraintSystemBuilderImpl.Constraint>,
     private val typeVariableSubstitutors: Map<CallHandle, TypeSubstitutor>
 ) : ConstraintSystem {
-    private val localTypeParameterBounds: Map<TypeVariable, TypeBoundsImpl>
-        get() = allTypeParameterBounds.filterNot { it.key.isExternal }
 
     override val status = object : ConstraintSystemStatus {
         // for debug ConstraintsUtil.getDebugMessageForStatus might be used
@@ -81,10 +79,12 @@ internal class ConstraintSystemImpl(
 
         override fun hasViolatedUpperBound() = !isSuccessful() && filterConstraintsOut(TYPE_BOUND_POSITION).status.isSuccessful()
 
-        override fun hasConflictingConstraints() = localTypeParameterBounds.values.any { it.values.size > 1 }
+        override fun hasConflictingConstraints() =
+            allTypeParameterBounds.entries.any { !it.key.isExternal && it.value.values.size > 1 }
 
         override fun hasUnknownParameters() =
-            localTypeParameterBounds.values.any { it.values.isEmpty() } || hasTypeParameterWithUnsatisfiedOnlyInputTypesError()
+            allTypeParameterBounds.entries.any { !it.key.isExternal && it.value.values.isEmpty() } ||
+                    hasTypeParameterWithUnsatisfiedOnlyInputTypesError()
 
         override fun hasParameterConstraintError() = errors.any { it is ParameterConstraintError }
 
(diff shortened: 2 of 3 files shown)


compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -101,33 +101,41 @@ internal class ConstraintSystemImpl(
...
                 else getDefaultType(variable)
-            substitutionContext.put(typeConstructor, TypeProjectionImpl(type))
+            substitutionContext[typeConstructor] = TypeProjectionImpl(type)
         }
         return substitutionContext
     }
 
     override val typeVariables: Set<TypeVariable>
-        get() = allTypeParameterBounds.keys
+        get() = allTypeParameterBounds.keys.toSet()
```
ting-yuan: nit: probably better to do this in `SmartIdentityTable`

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
```diff
@@ -100,10 +101,10 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
         }
 
         for ((_, typeVariable) in typeParameters.zip(typeVariables)) {
-            allTypeParameterBounds.put(typeVariable, TypeBoundsImpl(typeVariable))
+            allTypeParameterBounds[typeVariable] = TypeBoundsImpl(typeVariable)
         }
 
-        for ((typeVariable, _) in allTypeParameterBounds) {
+        for (typeVariable in allTypeParameterBounds.keys) {
```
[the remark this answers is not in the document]
jvgogh: not the same, other code in this class also adds entries to allTypeParameterBounds

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
```diff
@@ -100,10 +101,10 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
         }
 
         for ((_, typeVariable) in typeParameters.zip(typeVariables)) {
```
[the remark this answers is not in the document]
jvgogh: done

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -81,10 +79,12 @@ internal class ConstraintSystemImpl(
 
         override fun hasViolatedUpperBound() = !isSuccessful() && filterConstraintsOut(TYPE_BOUND_POSITION).status.isSuccessful()
 
-        override fun hasConflictingConstraints() = localTypeParameterBounds.values.any { it.values.size > 1 }
+        override fun hasConflictingConstraints() =
+            allTypeParameterBounds.keys.any { !it.isExternal && allTypeParameterBounds[it]!!.values.size > 1 }
 
         override fun hasUnknownParameters() =
-            localTypeParameterBounds.values.any { it.values.isEmpty() } || hasTypeParameterWithUnsatisfiedOnlyInputTypesError()
+            allTypeParameterBounds.keys.any { !it.isExternal && allTypeParameterBounds[it]!!.values.isEmpty() } ||
```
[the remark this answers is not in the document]
jvgogh: done

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
jvgogh: SmartIdentityTable does not implement the Map interface, so keys does not return a Set

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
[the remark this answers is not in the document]
jvgogh: done

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -175,8 +188,9 @@ internal class ConstraintSystemImpl(
 
     override fun toBuilder(filterConstraintPosition: (ConstraintPosition) -> Boolean): ConstraintSystem.Builder {
         val result = ConstraintSystemBuilderImpl()
-        for ((typeParameter, typeBounds) in allTypeParameterBounds) {
-            result.allTypeParameterBounds.put(typeParameter, typeBounds.filter(filterConstraintPosition))
+        for (typeParameter in allTypeParameterBounds.keys) {
+            val typeBounds = allTypeParameterBounds[typeParameter]!!
```
ting-yuan: This can use `entries`, too:
`for ((typeParameter, typeBounds) in allTypeParameterBounds.entries) {`

jvgogh: done

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
```diff
@@ -27,6 +27,27 @@ class SmartIdentityTable<K, V> {
     val size: Int
         get() = keysArray?.size ?: largeMap!!.size
 
+
+    val keys: Sequence<K>
+        get() = keysArray?.asSequence() ?: largeMap!!.keys.asSequence()
+
+
+    val entries: Sequence<Entry<K, V>>
```
udalov: I'm somewhat concerned that this method behaves slightly different depending on the size of the table. Namely, if there are no more than 10 elements, it returns a new collection, otherwise it returns a view into the current collection. Which means that if you save the result of `entries` into a variable `x` and then add more elements to the table, you'll observe the newly added elements in `x` if and only if the table size was > 10 at the point where `x` was defined. This could lead to obscure bugs in certain situations. Note that `keys` doesn't suffer from this problem, it always returns a view.

I suggest either providing a view even if the table is small (more preferably, for consistency with `keys`), or return a new collection even if it's large. In any case, it might be worth to document the specific behavior (keys vs new collection) both on `keys` and `entries`

compiler/util/src/org/jetbrains/kotlin/utils/SmartIdentityTable.kt
```diff
@@ -80,7 +101,15 @@ class SmartIdentityTable<K, V> {
...
+        for (key in other.keys) {
+            this[key] = other[key]!!
+        }
+    }
+
     companion object {
         private const val ARRAY_UNTIL_SIZE = 10
     }
+
+    data class Entry<K, V>(val key: K, val value: V)
```
udalov: Might be nice to inherit this from `Map.Entry<K, V>` and make `entries` return `Sequence<Map.Entry<K, V>>` to avoid rewrapping map entries into these entries in the implementation

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -175,8 +188,8 @@ internal class ConstraintSystemImpl(
 
     override fun toBuilder(filterConstraintPosition: (ConstraintPosition) -> Boolean): ConstraintSystem.Builder {
         val result = ConstraintSystemBuilderImpl()
-        for ((typeParameter, typeBounds) in allTypeParameterBounds) {
-            result.allTypeParameterBounds.put(typeParameter, typeBounds.filter(filterConstraintPosition))
+        for (typeParameter in allTypeParameterBounds.entries) {
```
udalov: Since `entries` returns a destructurable class, I suppose you can keep the `for ((typeParameter, typeBounds) ...` syntax here

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -145,12 +148,22 @@ internal class ConstraintSystemImpl(
...
-                    TypeConstructorSubstitution.createByConstructorsMap(parameterToInferredValueMap)
+                    createTypeConstructorSubstitutionFromTable(parameterToInferredValueMap)
                 )
             )
         )
     }
 
+    private fun createTypeConstructorSubstitutionFromTable(
+        table: SmartIdentityTable<TypeConstructor, TypeProjection>,
+        approximateCapturedTypes: Boolean = false
```
udalov: I suggest to inline this parameter since it's always `false`

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -101,33 +101,36 @@ internal class ConstraintSystemImpl(
...
             val type =
                 if (value != null && !TypeUtils.contains(value, DONT_CARE)) value
                 else getDefaultType(variable)
-            substitutionContext.put(typeConstructor, TypeProjectionImpl(type))
+            substitutionContext[typeConstructor] = TypeProjectionImpl(type)
         }
         return substitutionContext
     }
 
     override val typeVariables: Set<TypeVariable>
```
udalov: Note that behavior was sneakily changed here in a similar way: a new collection is returned instead of a view. This is probably not a big deal, but we could've changed the `ConstraintSystem` interface instead so that `typeVariables` returned a `Sequence` to keep the old behavior

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -101,33 +101,36 @@ internal class ConstraintSystemImpl(
...
-        typeParameterBounds: Map<TypeVariable, TypeBoundsImpl>,
+        typeParameterBounds: SmartIdentityTable<TypeVariable, TypeBoundsImpl>,
         getDefaultType: (TypeVariable) -> KotlinType,
         substituteOriginal: Boolean
-    ): Map<TypeConstructor, TypeProjection> {
-        val substitutionContext = HashMap<TypeConstructor, TypeProjection>()
-        for ((variable, typeBounds) in typeParameterBounds) {
+    ): SmartIdentityTable<TypeConstructor, TypeProjection> {
+        val substitutionContext = SmartIdentityTable<TypeConstructor, TypeProjection>()
+        for (variable in typeParameterBounds.keys) {
```
udalov: Maybe `for ((variable, typeBounds) in typeParameterBounds.entries)`

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemImpl.kt
```diff
@@ -81,10 +79,12 @@ internal class ConstraintSystemImpl(
 
         override fun hasViolatedUpperBound() = !isSuccessful() && filterConstraintsOut(TYPE_BOUND_POSITION).status.isSuccessful()
 
-        override fun hasConflictingConstraints() = localTypeParameterBounds.values.any { it.values.size > 1 }
+        override fun hasConflictingConstraints() =
+            allTypeParameterBounds.entries.any { !it.key.isExternal && it.value.values.size > 1 }
```
udalov: Please use destructuring declaration here and below

compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/inference/ConstraintSystemBuilderImpl.kt
```diff
@@ -64,7 +65,7 @@ open class ConstraintSystemBuilderImpl(private val mode: Mode = ConstraintSystem
         EQUAL(EXACT_BOUND)
     }
 
-    internal val allTypeParameterBounds = LinkedHashMap<TypeVariable, TypeBoundsImpl>()
+    internal val allTypeParameterBounds = SmartIdentityTable<TypeVariable, TypeBoundsImpl>()
```
udalov: Unfortunately we'll now lose the order of iteration reflecting the sequence in which the type variables were added, as soon as the size of this map becomes greater than 10, which I think could lead to very hard-to-debug problems (those in which you can't minimize the test case because it starts to behave correctly)

udalov:
@jvgogh Could you please clarify what exactly brings the highest performance gain -- avoiding the use of `hashCode` or not creating a map when the expected size is small? It looks like these are two separate issues and maybe we can limit the optimization to exactly one of them

udalov:
Closing since there's been no movement on this. Please feel free to reopen if it's still relevant.

PR #2433, opened 2019-06-25
JVM_IR: Handle annotation KClass fields in external declarations

sfs:
This PR implements the special cases around KClass and Array<KClass> fields in annotation classes in ExpressionCodegen instead of as a lowering. Since the special cases in question are already implemented as part of the type mapper this results in less code overall while also handling external declarations correctly.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
@@ -6,159 +6,35 @@
 package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
-import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
-import org.jetbrains.kotlin.backend.jvm.intrinsics.KClassJavaProperty
 import org.jetbrains.kotlin.ir.IrStatement
-import org.jetbrains.kotlin.ir.builders.declarations.buildField
-import org.jetbrains.kotlin.ir.builders.irCall
-import org.jetbrains.kotlin.ir.builders.irGet
-import org.jetbrains.kotlin.ir.declarations.*
-import org.jetbrains.kotlin.ir.expressions.IrCall
-import org.jetbrains.kotlin.ir.expressions.IrExpression
-import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
-import org.jetbrains.kotlin.ir.symbols.IrSimpleFunctionSymbol
-import org.jetbrains.kotlin.ir.types.*
-import org.jetbrains.kotlin.ir.types.impl.makeTypeProjection
-import org.jetbrains.kotlin.ir.util.*
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.declarations.IrConstructor
+import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.util.isAnnotationClass
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
-import org.jetbrains.kotlin.types.Variance
 
-internal val annotationPhase = makeIrFilePhase(
-    ::AnnotationLowering,
+internal val annotationPhase = makeIrFilePhase<JvmBackendContext>(
+    { AnnotationLowering() },
     name = "Annotation",
-    description = "Remove constructors and modify field types in annotation classes"
+    description = "Remove constructors of annotation classes"
 )
 
 /**
- * Remove the constructors from annotation classes and change the types of KClass
- * and Array<KClass> fields to use java.lang.Class instead. This phase also rewrites
- * the uses of annotation class fields appropriately.
+ * Remove the constructors from annotation classes.
  */
-private class AnnotationLowering(private val context: JvmBackendContext) : FileLoweringPass, IrElementTransformerVoid() {
+private class AnnotationLowering : FileLoweringPass, IrElementTransformerVoid() {
+    override fun lower(irFile: IrFile) = irFile.transformChildrenVoid(this)
 
-    override fun lower(irFile: IrFile) {
-        irFile.transformChildrenVoid(this)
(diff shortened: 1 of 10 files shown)


gogabr:
In general, I'd like the codegen (and type mapper) to be as dumb and straightforward as possible, so moving code from lowerings into codegen does not seem a good idea to me.
Maybe in this case simplification overrides that.

sfs:
@gogabr I agree in principle. In fact, I'm the one who implemented this as a lowering in the first place since it seemed cleaner.

This works fine for local declarations, but is a lot of additional work for external declarations as we would need to lazily replace IrFunctions in order to change their return types.

gogabr:
@sfs Applying lowerings (or some similar transformations) to external declarations is a recurring theme.
We would like to find a general solution.

One idea currently discussed is to treat external package fragments as stripped versions of normal files, and make sure lowerings tolerate this (functions without bodies, values without initializers). Then all files could be lowered, presenting a coherent picture at each step.
But nothing is decided yet.

udalov:
I'm a bit concerned that the Class<->KClass transformation is happening unconditionally and unnoticeably everywhere now, not just when handling returned values of annotation methods, which was the original scope of the feature. Can we guarantee that this won't change semantics by avoiding CCE and wrapping the Class object instead, in _some_ places? It looks like it won't because `coerce` is supposed to cover only backend-specific implicit conversions, and all other conversions should've been handled by frontend (reporting errors where not applicable) and psi2ir, and because we use custom code instead of `coerce` when generating `as` operator. But I'm not 100% confident. BTW this is the reason why this feature was implemented carefully in the old backend, by calling `wrapJavaClassIntoKClass` manually where needed instead of adding the corresponding conversions directly into `StackValue.coerce`.

sfs:
I can add an additional check that the underlying IrTypes are both instances of KClass. This should always be the case for code coming from the frontend, but it ensures that this code will only run when the type mapper mapped from KClass to Class for some reason.

This is still assuming that this is the *only* case in which the type mapper turns a KClass into a Class. Is this assumption correct?

udalov:
Yes, it would be very nice to check `IrType`s in this way.

> This is still assuming that this is the only case in which the type mapper turns a KClass into a Class. Is this assumption correct?

Yes, I believe so.

sfs:
Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
```diff
@@ -6,167 +6,35 @@
...
 )
 
 /**
- * Remove the constructors from annotation classes and change the types of KClass
- * and Array<KClass> fields to use java.lang.Class instead. This phase also rewrites
- * the uses of annotation class fields appropriately.
+ * Remove the constructors from annotation classes.
  */
-private class AnnotationLowering(private val context: JvmBackendContext) : FileLoweringPass, IrElementTransformerVoid() {
+private class AnnotationLowering : FileLoweringPass, IrElementTransformerVoid() {
```
udalov: I suppose such a trivial lowering is not needed as well, since it can be handled by a simple check in `ClassCodegen.generateMethod`

sfs: @udalov I've just had a look and it's a bit more complicated since we also need to ensure that no other lowering touches the constructor of an annotation class.

Specifically, default argument stub generation needs a special case to avoid generating a stub for the constructor (which also replaces the default arguments with IrErrorExpressions) and enum class lowering needs a special case to avoid lowering the GetEnumValue constructs. Right now this only works since the IrConstructor is no longer being visited as it is only accessible through the initializers of corresponding properties. This is then how AnnotationCodegen gets the relevant information for default values in the backend.

I can certainly remove AnnotationLowering and add these three special cases, but in general this is a bit fragile since it depends on parts of the ir not being lowered. At least for the jvm ir backend I would rather go the other direction and replace annotation classes in the ir with something more specialized which is not a subtype of IrClass. What do you think?

udalov: You're right, it does make sense to leave this lowering right now. But what do you mean specifically by
> something more specialized which is not a subtype of IrClass

?

sfs: I mean having special constructs in the ir for annotation classes and instantiations.

Right now annotations are just special IrClasses with lots of implicit restrictions. This does allow for some code sharing, but adds a number of special cases. For example, annotation constructor calls have completely different codegen, default arguments for annotations are completely different from default arguments for normal constructors, annotation fields don't really have to be lowered and have very restricted types, etc.

In the end we don't really want to run all that many lowerings on annotation classes and we have to add special cases for them during codegen anyway. This begs the question if they shouldn't be represented with a different structure in the first place.

sfs:
Rebased PR to latest master.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnnotationLowering.kt
udalov: We don't have a platform-specific IR at the moment, and while your points are totally valid, I'm not sure if annotations alone are a good reason to diverge the IR in this way. I agree that annotation classes are a special case most of the time though.

Maybe we could rather have this lowering transform annotation classes to interfaces, since they are represented as interfaces on JVM, storing the fact that the original class was an annotation in `IrDeclaration.origin` (for the codegen to generate ACC_ANNOTATION)? We'll still need to find a way to store default argument values somewhere (a mutable map in `JvmBackendContext` would be the first approximation), but at least we'll keep using the common IR and all lowerings will behave more consistently seeing annotation classes as interfaces rather than normal classes.

In any case, even if this works, it's not critical and can be postponed. For now, we can simply remove annotation constructors in this lowering.

sfs: Are annotation classes less special in the other backends? :)

Just to be clear though, I totally agree that this is not something we should focus on now. The current implementation works (or will work after this PR gets merged anyway) and it's not going to break with any new lowering that's using the existing IrElementVisitor apis.

udalov:
@sfs Thank you!

udalov:
BTW I've moved `isKClass` and `isKClassArray` out of `IrTypeUtils.kt` to `irTypePredicates.kt` and `PromisedValue.kt` respectively, see motivation in https://github.com/JetBrains/kotlin/commit/361c468c339e8779f210ca43e63c0cb3b5eb5838

PR #2435, opened 2019-06-25
JVM_IR: a few small fixes

pyos:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
```diff
@@ -744,15 +744,17 @@ class LocalDeclarationsLowering(
...
 
                     localClassConstructors[declaration] = LocalClassConstructorContext(declaration)
                 }
 
                 override fun visitClass(declaration: IrClass) {
                     super.visitClass(declaration)
 
-                    if (declaration.isInner) return
+                    // Inner classes inside local classes have local visibility for some reason.
+                    if (declaration.visibility != Visibilities.LOCAL || declaration.isInner) return
```
max-kammerer: Actually LocalDeclarationLower was rewriten a bit and would be merged today: see rr/mb/ldl branch, this commit would be redundant

compiler/testData/codegen/box/callableReference/property/kt6870_privatePropertyReference.kt
```diff
@@ -1,8 +1,9 @@
-// IGNORE_BACKEND: JVM_IR
 // IGNORE_BACKEND: JS_IR
 // TODO: muted automatically, investigate should it be ran for JS or not
 // IGNORE_BACKEND: JS, NATIVE
 
+// FULL_JDK
```
max-kammerer: Why FULL_JDK directive is required? Current backend works withoit it

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -164,7 +164,8 @@ internal class CallableReferenceLowering(val context: JvmBackendContext) : FileL
         val irFunctionReference: IrFunctionReference
     ) {
 
-        private val isLambda = irFunctionReference.origin == IrStatementOrigin.LAMBDA
+        private val isLambda = irFunctionReference.origin == IrStatementOrigin.LAMBDA ||
```
max-kammerer: Please rename to isLambdaOrAnonymousFunction

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
pyos: I'll wait until it is merged then, one of the tests unmuted here is dependent on this change plus another one.

compiler/testData/codegen/box/callableReference/property/kt6870_privatePropertyReference.kt
pyos: Thread has a method annotated with sun.reflect.CallerSensitive, which cannot be resolved without FULL_JDK, breaking fake override generation in psi2ir. The non-IR backend works because it doesn't actually use the unresolved descriptor for anything.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt
max-kammerer: @pyos  Just merged

compiler/testData/codegen/box/callableReference/property/kt6870_privatePropertyReference.kt
max-kammerer: @gogabr Could you take a look? Could we add some kind of lazyness here?

udalov: I think we should rather ignore unresolved annotation descriptors in psi2ir completely, something like this: https://github.com/JetBrains/kotlin/commit/888f7dda968b7b7debcc40af4310dc8312114d0f#diff-625389343bc48b023a77d0639f1b73b4R122. Please check if my change helps this test.

In any case, I think we shouldn't add or remove dependencies in existing codegen tests because it can mask a real problem (exactly as happened here) which will be hard to discover afterwards

udalov: I've pushed https://github.com/JetBrains/kotlin/commit/385366384c29ac776bbfea55f7d825c3021aacd7, so `FULL_JDK` is no longer needed in this test

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
```diff
@@ -79,35 +79,37 @@ class InnerClassesLowering(val context: BackendContext) : ClassLoweringPass {
...
-                        // Might be before a super() call (e.g. an argument to one), in which case the JVM bytecode verifier will reject
-                        // an attempt to access the field. Good thing we have a local variable as well.
-                        IrGetValueImpl(startOffset, endOffset, function.valueParameters[0].symbol, origin)
-                    } else {
-                        val outerThisField = context.declarationFactory.getOuterThisField(innerClass)
-                        IrGetFieldImpl(startOffset, endOffset, outerThisField.symbol, outerThisField.type, irThis, origin)
+                val receiverParameters = if (function == null || function is IrConstructor)
+                    listOfNotNull(irClass.thisReceiver)
+                else
+                    listOfNotNull(function.dispatchReceiverParameter, function.extensionReceiverParameter)
```
max-kammerer: `function.extensionReceiverParameter ` is redundant

max-kammerer: Actually as I see there is an `Outer.<anonymous>(): () -> String` passed to 'run' function. And during local declaration lowering this lambda is transformed to function with Outer extension receiver, that explicitly passed to it. So lower shouldn't try to lower any operation with this receiver.

pyos: Actually, the extension receiver is of type `Inner` here, so `ok` *should* be lowered to `this$0.$outer.ok`. That's why this change fixes two tests; if the receiver was `Outer`, this change would have no effect as the first iteration of the loop would notice that and return the same IrGetValue.

max-kammerer: @pyos Next sample works incorrectly in JVM_IR:
```
class Outer {
    val z = 1;

    inner class Inner {
        fun Outer.foo() {
            z
        }
    }
}
```

max-kammerer: Seems that there are wrong  assumptions in `IrValueSymbol.classForImplicitThis` cause extension receiver also have '<this>' name

pyos: Looks like you're right after all, the extension receiver in the lambda is `Outer`.

max-kammerer:
@pyos I've made another approach to distinguish dispatch and extension receivers: https://github.com/JetBrains/kotlin/commits/rr/mb/inner

PR #2436, opened 2019-06-25
Make ReturnableBlockLowering common

pyos:
and remove special handling in JVM_IR codegen.


pyos:
@max-kammerer can you have a look at this, please?

max-kammerer:
@pyos Yes, of course, sorry for delay

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/ReturnableBlockLowering.kt
```diff
@@ -69,40 +69,25 @@ import org.jetbrains.kotlin.ir.visitors.IrElementTransformer
...
-private class ReturnableBlockLoweringContext(val containingDeclaration: IrDeclarationParent) {
-    var labelCnt = 0
-    val returnMap = mutableMapOf<IrReturnableBlockSymbol, (IrReturn) -> IrExpression>()
-}
-
-private class ReturnableBlockTransformer(
-    val context: CommonBackendContext
-) : IrElementTransformer<ReturnableBlockLoweringContext> {
+class ReturnableBlockLowering(val context: CommonBackendContext) : FileLoweringPass, IrElementTransformerVoidWithContext() {
+    private var labelCnt = 0
```
max-kammerer: Counter was file specific but now it's  module specific.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/ReturnableBlockLowering.kt
```diff
@@ -69,40 +69,25 @@ import org.jetbrains.kotlin.ir.visitors.IrElementTransformer
...
+    private val returnMap = mutableMapOf<IrReturnableBlockSymbol, (IrReturn) -> IrExpression>()
```
max-kammerer: same here

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/ReturnableBlockLowering.kt
pyos: Thanks, haven't noticed that the JS backend only constructs file lowerings once, unlike the JVM one.

max-kammerer:
Thanks!

PR #2437, opened 2019-06-25
KT-32237 Always use LF line endings in generated stdlib code 

ajalt:
`appendln` uses `SystemProperties.LINE_SEPARATOR`, which in `\r\n` on Windows. This causes generated files to have mixed LF and CRLF line endings when `:tools:kotlin-stdlib-gen` is run on Windows. This commit removes the usages of `appendln` in Writers.kt and appends `\n` instead. `appendln` was only used in one function; the rest of the code that generates the stdlib uses `\n`.


ilya-g:
I'm for bringing some consistency in line endings here. `\n` is used more often than a platform dependent separator here, so it's indeed easier to avoid using the platform separator. 

However I feel that eradicating every usage of `appendln` is unnecessary and quite error-prone: there are no guarantees that it won't appear in the generator code again. Instead we can specify `line.separator` JVM system property when starting the generator by adding `systemProperty 'line.separator', '\n'` option to `JavaExec` task in `kotlin-stdlib-gen/build.gradle` file.

We should place `.gitattributes` with the following content in the directories with generated kt files:
```
*.kt eol=lf
```
So that if `core.autocrlf` git config setting is on, it won't replace LF with CRLF in these files on checkout, and won't consider them modified after regenerating them.

ajalt:
Yes, that's definitely a better solution. I reverted the previous change, implemented your suggestion, and squashed the branch.

PR #2440, opened 2019-06-29
Allow omitting version of standard artifacts when kotlin-android plugin is used

technoir42:


technoir42:
@h0tk3y PTAL

h0tk3y:
@technoir42 Thanks for the PR! Merged manually as 15feeddc4618.

PR #2441, opened 2019-07-02
Enable not null assertions tests on the JVM_IR backend

sfs:
This PR splits the custom `GenerateNotNullAssertionsTest` class into standard `box`, `boxAgainstJava`, `bytecodeText`, and `compileKotlinAgainstKotlin` tests. This allows us to run the same tests on the JVM_IR backend.

The second commit in the PR fixes some of these tests which were failing due to missing support for `-Xno-param-assertions`. The remaining test failures are due to missing IMPLICIT_NOTNULL casts in the IR backend.


compiler/testData/compileKotlinAgainstKotlin/noAssertionsForKotlin.kt
```diff
@@ -0,0 +1,28 @@
+// TARGET_BACKEND: JVM
```
udalov: Looks like this test doesn't check what was tested previously in `testNoAssertionsForKotlinFromBinary`: that there are no assertions on calls to Kotlin declarations from dependencies. I'm afraid we'll have to keep that test until we have the necessary infrastructure to migrate it, namely multi-module bytecode text tests.

sfs: You're right, there's no way to test here that there are no intrinsic calls. I'll keep this one test in `GenerateNotNullAssertionsTest` and add a comment.

compiler/testData/codegen/boxAgainstJava/notNullAssertions/doGenerateParamAssertions.kt
```diff
@@ -15,3 +17,20 @@ public static void runTest(doGenerateParamAssertions a) {
         throw new AssertionError("Fail: IllegalArgumentException expected");
     }
 }
+
+// FILE: B.kt
+import test.doGenerateParamAssertions as C
```
udalov: Would be nice to rename the class itself now that it's possible. Previously the Java class was named this way because its file was compiled directly by javac, but box tests copy everything to tmpdir with new names, so we can use the short name for convenience

sfs: Done

compiler/testData/compileKotlinAgainstKotlin/noAssertionsForKotlin.kt
sfs: I've moved this test to tests-common and at least enabled it for both the JVM and JVM_IR backends, with a TODO to remove the test class once multi-module bytecode text tests become available.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractGenerateNotNullAssertionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+        myEnvironment =
+                KotlinCoreEnvironment.createForTests(getTestRootDisposable(), configuration, EnvironmentConfigFiles.JVM_CONFIG_FILES);
+        myFiles = null;
+    }
+
+    private void loadSource(@NotNull String fileName) {
+        loadFileByFullPath(KotlinTestUtils.getTestDataPathBase() + "/codegen/" + getPrefix() + "/" + fileName);
+    }
+
+    public void doTestNoAssertionsForKotlinFromBinary(String binaryDependencyFilename, String testFilename) {
```
udalov: Please make it protected to fix the IDE inspection

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractGenerateNotNullAssertionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.jetbrains.kotlin.test.TestJdkKind;
+import org.jetbrains.org.objectweb.asm.ClassReader;
+import org.jetbrains.org.objectweb.asm.ClassVisitor;
+import org.jetbrains.org.objectweb.asm.MethodVisitor;
+import org.jetbrains.org.objectweb.asm.Opcodes;
+
+import java.io.File;
+
+// TODO Remove this class once general multi-module bytecode text tests are implemented.
+public class AbstractGenerateNotNullAssertionsTest extends CodegenTestCase {
```
udalov: Please make the class abstract to fix the IDE inspection

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractGenerateNotNullAssertionsTest.java
sfs: Done

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractGenerateNotNullAssertionsTest.java
sfs: Done

udalov:
@sfs Thank you!

PR #2443, opened 2019-07-02
KT-32324 Formatter doesn't insert space after safe cast operator `as?`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32324


goodwinnk:
Merged as https://github.com/JetBrains/kotlin/commit/9f6e88d5713ab66af049110c55197bbe6b49d719. Thank you!

PR #2444, opened 2019-07-03
Not throw exception when cidr has been registered already.

juan-chen:
This is to avoid failures from some Android Studio integration tests depending on cidr.


goodwinnk:
I would like to understand more about how building of Android Studio is done, because those additional checks were added to remove the whole `registerAdditionalResourceBundleInTests` hack eventually. But that's OK for now. Thank you!

PR #2445, opened 2019-07-03
KT-32277 Space before by delegate keyword on property is not formatted

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32277


goodwinnk:
Merged as https://github.com/JetBrains/kotlin/commit/44e9090931bbbee041c4f4dd8a1bd38610941072. Thank you!

PR #2446, opened 2019-07-04
Add intrinsic for ULong/UInt.toString on Java 8+

JakeWharton:
Unfortunately this cannot currently be done for the extension overload which accepts a radix due to behavior difference with regard to invalid radix values.


max-kammerer:
Thank you!

PR #2447, opened 2019-07-04
KT-32178 Autocompletion of 'suspend' should not add the 'fun' keyword when writing a function type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32178

### DIFF
--- a/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KeywordCompletion.kt
+++ b/idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KeywordCompletion.kt
@@ -104,7 +104,8 @@ object KeywordCompletion {
             var keyword = keywordToken.value
 
             val nextKeyword = when {
-                keywordToken == SUSPEND_KEYWORD && ((position.containingFile as? KtFile)?.isScript() == true) -> null
+                keywordToken == SUSPEND_KEYWORD && ((position.containingFile as? KtFile)?.isScript() == true 
+                        || position.getStrictParentOfType<KtTypeReference>() != null) -> null
                 else -> COMPOUND_KEYWORDS[keywordToken]
             }
             var applicableAsCompound = false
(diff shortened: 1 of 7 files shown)


LouisCAD:
Is there a reason this PR still hasn't been merged?

idea/idea-completion/src/org/jetbrains/kotlin/idea/completion/KeywordCompletion.kt
```diff
@@ -103,7 +103,8 @@ object KeywordCompletion {
             var keyword = keywordToken.value
 
             val nextKeyword = when {
-                keywordToken == SUSPEND_KEYWORD && ((position.containingFile as? KtFile)?.isScript() == true) -> null
+                keywordToken == SUSPEND_KEYWORD && ((position.containingFile as? KtFile)?.isScript() == true
```
fedochet: It seems like there are no actual need to prevent `suspend` or `suspend fun` completion in `.kts` files. Could you please remove this check all along, so it will be something like: 

```kt
val nextKeyword = when {
    keywordToken == SUSPEND_KEYWORD && position.getStrictParentOfType<KtTypeReference>() != null -> null
    else -> COMPOUND_KEYWORDS[keywordToken]
}
```

idea/idea-completion/testData/keywords/InFunctionTypePosition.kt
```diff
@@ -2,5 +2,5 @@ fun foo() {
     val test : <caret>
```
fedochet: Could you please add a test for completion in function argument type parameter? 

Like this: 

```kt
fun foo(a: <caret>) {}
```

fedochet:
@LouisCAD Hi, I am going to merge this soon enough

fedochet:
Merged, thanks @t-kameyama!

I've squashed your commits into ba640be81defa304ac2ed2396fc83249834a0f0d, changed added test a little and added ticket ID to commit message

PR #2448, opened 2019-07-04
KT-32318 "Remove argument name" intention does not remove square braces for annotation vararg argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32318


dimonchik0036:
Thanks!

PR #2449, opened 2019-07-05
KT-32365 "Convert to sealed class" intention should not be suggested when no "class" keyword

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32365


dimonchik0036:
Thanks! https://github.com/JetBrains/kotlin/commit/1f89c0f7302b48f8bf03e1eaf1c66c97dcdd39b8

PR #2450, opened 2019-07-05
Support -Xassertions in the JVM_IR backend

sfs:
In legacy mode, assert is an ordinary inline function. In particular, it always evaluates its arguments and only fails if assertions are enabled for the kotlin package.

This PR adds a lowering which implements the other three assertion modes. This allows assertions to be globally disabled or enabled at compile time, or to match the jvm behavior which allows assertions to be enabled/disabled per class and package and also does not evaluate the assertion arguments unless assertions are enabled.


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -556,3 +557,11 @@ fun createStaticFunctionWithReceivers(
         metadata = oldFunction.metadata
     }
 }
+
+val IrBody.statements: List<IrStatement>
+    get() = when (this) {
+        is IrBlockBody -> statements
+        is IrExpressionBody -> listOf(expression)
+        is IrSyntheticBody -> error("Synthetic body contains no statements $this")
+        else -> throw IllegalStateException()
```
neetopia: maybe add some text on exception?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,177 @@
...
+
+private class AssertionLowering(private val context: JvmBackendContext) :
+    FileLoweringPass,
+    IrElementTransformer<AssertionLowering.ClassInfo?>
+{
+    // Keeps track of the $assertionsDisabled field, which we generate lazily for classes containing
+    // assertions when compiled with -Xassertions=jvm.
+    private inner class ClassInfo(val irClass: IrClass) {
+        private val lazyAssertionsDisabledField: Lazy<IrField> = lazy {
+            val descriptor = WrappedFieldDescriptor()
```
pyos: Using `buildField` would make this a bit shorter

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,177 @@
...
+                initializer = context.createIrBuilder(irClass.symbol).run {
+                    at(this@apply)
+                    irExprBody(irNot(irCall(this@AssertionLowering.context.ir.symbols.desiredAssertionStatus).apply {
+                        dispatchReceiver = getJavaClass(irClass)
+                    }))
+                }
+            }
+        }
+
+        val assertionsDisabledField by lazyAssertionsDisabledField
```
pyos: IMO since you're only using this from getAssertionDisabled in the same class, a simple nullable field would be cleaner

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,177 @@
...
+
+        val needsAssertionDisabledField: Boolean
+            get() = lazyAssertionsDisabledField.isInitialized()
+
+        fun getAssertionDisabled(irBuilder: IrBuilderWithScope): IrExpression =
+            irBuilder.irGetField(null, assertionsDisabledField)
+
+        private fun getJavaClass(irClass: IrClass): IrExpression {
```
pyos: This looks like a third copy of this code, maybe it should be extracted into a common method of CrIrType, for example?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,177 @@
...
+                    checkAssertion(assertCondition, lambdaArgument)
+                )
+            }
+        }
+    }
+
+    private fun IrBuilderWithScope.checkAssertion(assertCondition: IrExpression, lambdaArgument: IrExpression?) =
+        irBlock {
+            val lambda = lambdaArgument?.asSimpleLambda()
+            val invokeVar = if (lambda == null && lambdaArgument != null) irTemporary(lambdaArgument) else null
```
pyos: I don't think you need a temporary variable here.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
```diff
@@ -211,11 +211,11 @@ fun IrBuilderWithScope.irEquals(arg1: IrExpression, arg2: IrExpression, origin:
         arg1, arg2
     )
 
+fun IrBuilderWithScope.irNot(arg: IrExpression, origin: IrStatementOrigin? = null) =
```
pyos: There's already an irNot in LowerUtils.kt.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrInlineUtils.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// TODO use a generic inliner (e.g. JS/Native's FunctionInlining.Inliner)
+// Inline simple function calls without type parameters, default parameters, or varargs.
+fun IrFunction.inline(arguments: List<IrValueDeclaration> = listOf()): IrReturnableBlock {
+    require(body != null)
+    val argumentMap = valueParameters.zip(arguments).toMap()
+    val blockSymbol = IrReturnableBlockSymbolImpl(descriptor)
+    val block = IrReturnableBlockImpl(startOffset, endOffset, returnType, blockSymbol, null, symbol)
+    val remapper = object : AbstractVariableRemapper() {
```
pyos: `: VariableRemapper(argumentMap)` would make the override of `remapVariable` redundant.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: You're right, done!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: I've added it as an extension method on `JvmBackendContext` in `CallableReferenceLowering`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: Because there can be side effects when evaluating the argument to a function, see `compiler/testData/codegen/box/assert/jvm/LambdaNotEvaluated.kt`.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt
sfs: Thanks, I've changed the code to use that one instead.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrInlineUtils.kt
sfs: Done!

sfs:
Note: This PR does not implement the special handling of assertions in the inliner. That part depends on having the correct `inlineCallSiteInfo` in `IrSourceCompilerForInline`, which is currently not implemented. This is why I had to mute 3 boxInline tests.

By the way, it seems that the desired behavior here is for an inlined assertion to be enabled if the enclosing class *at the call-site* has assertions enabled. Is there anywhere to read up on the rationale for this?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,156 @@
...
+            initializer = context.createIrBuilder(irClass.symbol).run {
+                at(this@apply)
+                irExprBody(irNot(irCall(this@AssertionLowering.context.ir.symbols.desiredAssertionStatus).apply {
+                    dispatchReceiver = getJavaClass(irClass)
+                }))
+            }
+        }
+
+    private fun getJavaClass(irClass: IrClass) =
+        context.buildJavaClassReference(context.state.typeMapper.mapClass(irClass.descriptor))
```
udalov: Could you please avoid type mapping here and generate a proper class reference with the default type of the class, instead of CrIrType of its ASM type? I've removed CrIrType in fc5e9f5c60 (not in master yet) since it was pretty much a hack, and would be grateful if it wasn't used in more places

sfs: Absolutely! This was actually on my todo list anyway. I'll fix it as soon as your commit is merged.

sfs:
Changed the code to avoid CrIrType (fc5e9f5) and to look past facade classes when detecting assertions.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: Done!

compiler/testData/codegen/box/assert/jvm/InitializerAssertionsEnabled.kt
```diff
@@ -0,0 +1,23 @@
+// TARGET_BACKEND: JVM
```
max-kammerer: Please name file with lower letter

compiler/testData/codegen/box/assert/jvm/LambdaNotEvaluated.kt
```diff
@@ -0,0 +1,50 @@
+// TARGET_BACKEND: JVM
```
max-kammerer: ..and here

compiler/testData/codegen/boxInline/assert/jvmAssertInlineFunctionAssertionsEnabled.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: Seems this could be supported via `IrInlineCodegen.generateAssertFieldIfNeeded`.
It's not required for this PR

compiler/testData/codegen/box/assert/jvm/InitializerAssertionsEnabled.kt
sfs: Done

compiler/testData/codegen/box/assert/jvm/LambdaNotEvaluated.kt
sfs: Done

sfs:
Rebased to master (minor fix in `ArrayConstructorLowering`)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    fun getAssertionDisabled(irBuilder: IrBuilderWithScope, data: ClassInfo): IrExpression {
+        if (data.assertionsDisabledField == null)
+            data.assertionsDisabledField = buildAssertionsDisabledField(data.irClass)
+        return irBuilder.irGetField(null, data.assertionsDisabledField!!)
+    }
+
+    private fun buildAssertionsDisabledField(irClass: IrClass) =
+        buildField {
+            name = Name.identifier("\$assertionsDisabled")
```
max-kammerer: Please use 'ASSERTIONS_DISABLED_FIELD_NAME' from AssertCodegenUtil.kt

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/Primitives.kt
```diff
@@ -30,7 +30,7 @@ fun primitiveOp1(
     startOffset: Int, endOffset: Int,
     primitiveOpSymbol: IrSimpleFunctionSymbol,
     primitiveOpReturnType: IrType,
-    origin: IrStatementOrigin,
+    origin: IrStatementOrigin?,
```
max-kammerer: Is it neccessery?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,158 @@
...
+        if (context.state.assertionsMode != JVMAssertionsMode.LEGACY)
+            irFile.transformChildren(this, null)
+    }
+
+    override fun visitClass(declaration: IrClass, data: ClassInfo?): IrStatement {
+        val info = ClassInfo(declaration)
+
+        super.visitClass(declaration, info)
+
+        // Note that it's necessary to add this member at the end of the class, behind all user-visible initializers,
```
max-kammerer: Could you point to specification?
Actually I see totatly opposite situation for Java code:
```
class A2 {

    static long z = System.currentTimeMillis();

    public static void main(String foo) {
        assert foo != null : "123";
    }
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    override fun visitCall(expression: IrCall, data: ClassInfo?): IrElement {
+        val function = expression.symbol.owner
+        if (!function.isAssert)
+            return super.visitCall(expression, data)
+
+        val mode = context.state.assertionsMode
+        if (mode == JVMAssertionsMode.ALWAYS_DISABLE)
+            return IrGetObjectValueImpl(
```
max-kammerer: Is it possible to return empty expression here, e.g. IrCompositeImpl?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: Done

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/Primitives.kt
sfs: Good catch, I've removed it.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: It's a bit more subtle than I thought. The java spec says (se7, section 14.10):

> An assert statement that is executed before its class has completed initialization is enabled.

However, the idea here is that the assertion status is initialized as soon as the initializer runs. The rule is only relevant with circular initializers for which we run static code before running the initializer at all. I've changed the comment and code accordingly and expanded the corresponding test.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AssertionLowering.kt
sfs: You're right, that's cleaner. Changed it.

sfs:
I just noticed that calling `getTopLevelClass` in `JvmSymbols.kt` breaks in firMode. In particular it breaks `testFirHello`, so I've added a workaround...

max-kammerer:
Please also rebase on master: #2483 is merged

sfs:
Rebased on master

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
```diff
@@ -0,0 +1,53 @@
+// TARGET_BACKEND: JVM
```
max-kammerer: Could you also add 'initializerAssertionsDisable.kt' test by 'interfaceAssertionsDisabled.kt/interfaceAssertionsEnabled.kt' analogue?

Actually compiler also moves all initializer from class companion to outer class. And here is an interesting corner case with assertions. @ilmirus Seems this  code is not supported in current backend:
```
open class Bar {    
    companion object {
        val barAssertionThrown = try {
            assert(false) 
            false
        } catch(error: java.lang.AssertionError) {
            true
        }
    }
}
...
I don't see any $assertionsDisabled neither in companion nor in Bar. And if understand correctly this field initialization should  be performed only in companion clinit (that require some trick at least in IR backend cause all initialization is poped up in outer class)

Comment (max-kammerer) [reply compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt]: @sfs NB:  corner case support is not necessery for this PR

Comment (ilmirus) [reply compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt]: This code is, in fact, supported in current backend. The `$assertionsDisabled` field is generated in `Bar` class and not in companion. This, it seems, is in line with intended behavior for inner classes: if `assert` is used in inner class, we shall generate `$assertionsDisabled` in outer class.

Comment (ilmirus) [reply compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt]: However, in cases of nested and inner classes the old backend generates the field in inner class. This is known bug.

File: compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt [hunk 17]
```diff
@@ -0,0 +1,53 @@
...
+            return "OK"
+        }
+    }
+}
+
+class Dummy
+
+fun disableAssertions(): Checker {
+    val loader = Dummy::class.java.classLoader
+    loader.setPackageAssertionStatus("initializerAssertionsEnabled", false)
```
max-kammerer: Sorry for confusion by previous comment about 'initializerAssertionsDisable.kt' test , but seems it's best to rename this one to address name with ' // Assertions which run before the class initializer are always checked ' comment

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
sfs: For comparison: Javac generates a field in an inner class, but initializes the assertion status based on the outer class. In either case, there's no way to selectively enable/disable assertions in an inner class.

I will change the code to generate the field in the outer class and add some tests for this. We will still need to add support for this in `IrInlineCodegen.generateAssertFieldIfNeeded`, though.

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
sfs: Right, I'll think of a better name, this one is misleading.

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
sfs: I've added 5 more tests for setting the assertion status per class: for toplevel, companion, local, nested, and inner classes. Currently, local, nested, and inner classes do not work in the JVM backend.

At the same time I've fixed these cases in the JVM_IR backend.

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
sfs: Renamed the test to `testAssertionsEnabledBeforeClassInitializers`.

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
max-kammerer: @sfs Thanks, I will merge PR as is after rr. 

@ilmirus Actually as mentioned by @sfs javac generates field in each nested class but initializes it with outer class state:
```
    LDC LOuter;.class
    INVOKEVIRTUAL java/lang/Class.desiredAssertionStatus ()Z
    PUTSTATIC Inner.$assertionsDisabled : Z
```
As I remember `LDC Class` doesn't trigger class initialization

max-kammerer: @sfs last request could you also add muted bytecode test with proper `$assertionsDisabled` initialization though 'desiredAssertionStatus(outer)' ?

sfs: > As I remember `LDC Class` doesn't trigger class initialization

That's true ([jvm spec](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-5.html#jvms-5.5)) and explains the Java behavior. For compatibility we should do the same in Kotlin, but this is not a backwards compatible change. For example, consider the following code.

```kotlin
class Outer {
  companion object {
    init { error("") }
  }

  class Inner {
    init { assert(true) }
  }
}

fun main() {
  Outer.Inner()
}
```

When compiled with `-Xassertions=jvm`, `main` will fail with an exception, since we trigger class initialization for `Outer` (through `getstatic Outer.$assertionsDisabled` in `Outer$Inner.<init>`). The equivalent code in Java --- or for that matter the Kotlin code with any other assertions mode --- will run just fine.

Do you know in which language version `-Xassertions=jvm` was introduced? Is the feature still experimental? Otherwise I can file a language committee issue for this.

> could you also add muted bytecode test with proper nested class $assertionsDisabled initialization though 'desiredAssertionStatus(outer)' ?

Will do! For the reasons above I don't want to codify the current behavior, but I can at least add a test which checks that we only look at the assertion status in the Outer class.

sfs: Correction: the code example above doesn't currently work in the JVM backend anyway (since the assertion status is initialized based on the inner class). Did it work in a previous version?

sfs: @max-kammerer I've added a simple bytecode test in `compiler/testData/codegen/bytecodeText/assert/jvmNestedClass.kt`. It's not very precise and I would like to expand on this once it's clear what we should do about the issue above.

udalov: All modes of `-Xassertions` are experimental, so no language committee approval is needed, we're free to change behavior as we please here. From my (slightly out-of-context) view, it'd be ideal to generate bytecode similar to javac both in the old and in the IR backend, but we can basically postpone this until the feature is graduated (which I think is not yet planned for `-Xassertions` in 1.4).

max-kammerer: @sfs Yes, this feature is still experimental and could be changed without compatibility respect.  There is a lack of proper nested/local class support in current backend. The test looks good, please just add 
'
// 0 PUTSTATIC Outer
// 1 PUTSTATIC Outer$Inner.\$assertionsDisabled
'
 to be sure that field is placed in correct class

sfs: > All modes of `-Xassertions` are experimental, so no language committee approval is needed, we're free to change behavior as we please here.

> Yes, this feature is still experimental and could be changed without compatibility respect.

That's perfect, in this case I'll just change the lowering to mimic the javac behavior and add more tests.

sfs: @max-kammerer I've changed the lowering as discussed, made the bytecode text test more precise, and added a new box test to ensure that we don't run class initializers just because we use assertions in a nested class. Except for inlining, everything should work as in javac now.

Incidentally, the bytecode text test for asserts in crossinline lambda arguments to inline functions (`compiler/testData/codegen/bytecodeText/assert/jvmCrossinlineAssertInLambda.kt`) doesn't seem correct to me. The test checks that the assertion status is set based on the generated lambda class (`LDC LA\$inlineSite\$\$inlined\$inlineMe\$1;.class`), instead of based on the top level class.

max-kammerer: Yes, it's wrong test

compiler/testData/codegen/bytecodeText/assert/jvmNestedClass.kt
```diff
@@ -0,0 +1,16 @@
...
+    }
+}
+
+// We set the assertion status based on top-level classes.
+// 0 LDC LOuter\$Inner;.class
+// Outer\$Inner.<clinit>:
+// 1 LDC LOuter;.class\s*INVOKEVIRTUAL java/lang/Class.desiredAssertionStatus \(\)Z
+// 1 PUTSTATIC Outer\$Inner.\$assertionsDisabled : Z
+// Outer\$Inner.f:
+// 1 GETSTATIC Outer\$Inner.\$assertionsDisabled
```
max-kammerer: Please also add:
// 1 desiredAssertionStatus
// 1 public final static Z \$assertionsDisabled

sfs: Done!

compiler/testData/codegen/box/assert/jvm/initializerAssertionsEnabled.kt
sfs: Ok, I've fixed the test as well.

compiler/testData/codegen/bytecodeText/assert/jvmNestedClass.kt
max-kammerer: @sfs Sorry, but actually in java all '$assertionsDisabled' fields are package local and synthetic:
```
final static synthetic Z $assertionsDisabled
```

sfs: Thanks, that's good to know. I've changed the test.

Currently the IR backend has problems with field visibility and this is just one more test case which we'll have to fix down the line...

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
```diff
@@ -41,6 +41,7 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object GENERATED_SAM_IMPLEMENTATION : IrDeclarationOriginImpl("GENERATED_SAM_IMPLEMENTATION", isSynthetic = true)
     object ENUM_MAPPINGS_FOR_WHEN : IrDeclarationOriginImpl("ENUM_MAPPINGS_FOR_WHEN", isSynthetic = true)
     object SYNTHETIC_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("SYNTHETIC_INLINE_CLASS_MEMBER", isSynthetic = true)
+    object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD")
```
max-kammerer: 'isSynthetic = true' should add synthetic modifier to bytecode

sfs: Thanks! I changed this and made the check in `testJvmStatic` more precise to ensure that it doesn't pass without the right field visibility.

max-kammerer:
@sfs Many thanks!

PR #2453, opened 2019-07-09
docs: added KDoc for Comparator, fixed KDoc for comparison binarySeach()

p7nov:


libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -327,12 +327,12 @@ public inline fun <T, K : Comparable<K>> List<T>.binarySearchBy(
 
 /**
  * Searches this list or its range for an element for which [comparison] function returns zero using the binary search algorithm.
- * The list is expected to be sorted into ascending order according to the provided [comparison],
- * otherwise the result is undefined.
+ * The list is expected to be sorted so that the integer return values of the provided [comparison]
+ * form an ascending sequence on the list elements. Otherwise the result is undefined.
```
ilya-g: "ascending sequence" that is sufficient but not necessary requirement. The necessary requirement is that the sign is ascending, i.e.  negative values come before zero and zeroes come before positive values.

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -327,12 +327,12 @@ public inline fun <T, K : Comparable<K>> List<T>.binarySearchBy(
...
  *
  * If the list contains multiple elements for which [comparison] returns zero, there is no guarantee which one will be found.
  *
- * @param comparison function that compares an element of the list with the element being searched.
+ * @param comparison function that maps elements to integer values.
```
ilya-g: This is a restatement of the parameter function signature that doesn't give an idea how this function should be implemented.

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -326,13 +326,16 @@ public inline fun <T, K : Comparable<K>> List<T>.binarySearchBy(
 
 
 /**
- * Searches this list or its range for an element for which [comparison] function returns zero using the binary search algorithm.
- * The list is expected to be sorted into ascending order according to the provided [comparison],
- * otherwise the result is undefined.
+ * Searches this list or its range for an element for which the given [comparison] function returns zero using the binary search algorithm.
+ * The list is expected to be sorted so that the signs of the [comparison] function's return values ascend on the list elements,
```
ilya-g: I propose moving this to details (i.e. adding a blank line after summary).

ilya-g:
While we here, it may make sense also to document `Comparator` function: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-comparator.html

libraries/stdlib/common/src/kotlin/KotlinH.kt
```diff
@@ -9,11 +9,29 @@ import kotlin.annotation.AnnotationTarget.FIELD
 import kotlin.annotation.AnnotationTarget.PROPERTY
 
 
-
+/**
+ * Provides a [comparison] function for imposing a total ordering between instances of a class.
```
ilya-g: I'll change this to the following upon merge:
```suggestion
 * Provides a comparison function for imposing a total ordering between instances of the type [T].
```

ilya-g:
Squashed and merged as https://github.com/JetBrains/kotlin/commit/2577dfa06abd2350a287d456fbeb44293ddabcd4

PR #2454, opened 2019-07-10
KT-30136 False negative "Redundant explicit 'this'" with local variable

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30136


idea/src/org/jetbrains/kotlin/idea/inspections/ExplicitThisInspection.kt
```diff
@@ -60,7 +61,13 @@ class ExplicitThisInspection : AbstractKotlinInspection() {
 
             //we avoid overload-related problems by enforcing that there is only one candidate
             val name = referenceExpression.getReferencedNameAsName()
-            val candidates = scope.getAllAccessibleVariables(name) + scope.getAllAccessibleFunctions(name)
+            val candidates = if (reference is KtCallExpression
+                || (expression is KtCallableReferenceExpression && reference.mainReference.resolve() is KtFunction)
+            ) {
+                scope.getAllAccessibleFunctions(name)
```
mglukhikh: I'm sorry but looks like it's not so easy :(
```
class Some {
    operator fun invoke() {}
}

class Other {
    fun foo() {
        val a = Some()
        this.a() // Resolved to member a()
        a() // Resolved to invoke on local a()
    }

    fun a() {}
}
```

mglukhikh:
Merged 👍

PR #2455, opened 2019-07-11
Disable cross compilation from mingw to linux-arm64

alexander-gorshenev:
as it is disfunctional now


sbogolepov:
Did you check that it is not working?

alexander-gorshenev:
seems to work on master, actually

PR #2458, opened 2019-07-12
JS API to access DOM elements by ID via delegated properties

jonnyzzz:
See [KT-32552](https://youtrack.jetbrains.com/issue/KT-32552) for more details.

 Provides a short access to document elements by ID via delegated
 property syntax. Received element is not cached and received
 directly from the [Document] by calling [Document.getElementById]
 function on every property access. Throws an exception if element
 is not found or has different type
 To access an element with `theId` ID use the following property declaration
 ```
 val theID by document.gettingElementById
 ```
 To access an element of specific type, just add it to the property declaration
 ```
 val theID : HTMLImageElement by document.gettingElementById
 ```

### DIFF
--- a/libraries/stdlib/js/src/kotlin/dom/Dom.kt
+++ b/libraries/stdlib/js/src/kotlin/dom/Dom.kt
@@ -37,7 +37,7 @@ public val Node.isElement: Boolean
  *
  * To access an element of specific type, just add it to the property declaration
  * ```
- * val theID : HTMLImageElement by document.gettingElementById
+ * val theID: HTMLImageElement by document.gettingElementById
  * ```
  */
 inline val Document.gettingElementById get() = DocumentGettingElementById(this)
--- a/libraries/stdlib/js/src/kotlin/dom/Dom.kt
+++ b/libraries/stdlib/js/src/kotlin/dom/Dom.kt
@@ -32,12 +32,12 @@ public val Node.isElement: Boolean
  *
  * To access an element with `theId` ID use the following property declaration
  * ```
- * val theID by document.gettingElementById
+ * val theId by document.gettingElementById
  * ```
  *
  * To access an element of specific type, just add it to the property declaration
  * ```
- * val theID: HTMLImageElement by document.gettingElementById
+ * val theId: HTMLImageElement by document.gettingElementById
  * ```
  */
 inline val Document.gettingElementById get() = DocumentGettingElementById(this)
--- a/libraries/stdlib/js/src/kotlin/dom/Dom.kt
+++ b/libraries/stdlib/js/src/kotlin/dom/Dom.kt
@@ -5,7 +5,10 @@
 
 package kotlin.dom
 
-import org.w3c.dom.*
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+import org.w3c.dom.Node
+import kotlin.reflect.KProperty
 
 /**
  * Gets a value indicating whether this node is a TEXT_NODE or a CDATA_SECTION_NODE.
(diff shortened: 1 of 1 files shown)


libraries/stdlib/js/src/kotlin/dom/Dom.kt
```diff
@@ -18,3 +21,41 @@ public val Node.isText: Boolean
...
+/**
+ * Implementation details of [Document.gettingElementById]
+ * @see Document.gettingElementById
+ */
+inline class DocumentGettingElementById(val document: Document) {
+    /**
+     * Implementation details of [Document.gettingElementById]. Delegated property
+     * @see Document.gettingElementById
+     */
+    inline operator fun <reified T : Element> getValue(x: Any?, kProperty: KProperty<*>): T {
```
ilya-g: What reification of type parameter gives over having it at the call site (in the generated property getter)?

Could it be a problem for using external interfaces here?

libraries/stdlib/js/src/kotlin/dom/Dom.kt
```diff
@@ -18,3 +21,41 @@ public val Node.isText: Boolean
...
+
+
+/**
+ * Provides a short access to document elements by ID via delegated
+ * property syntax. Received element is not cached and received
+ * directly from the [Document] by calling [Document.getElementById]
+ * function on every property access. Throws an exception if element
+ * is not found or has different type
+ *
+ * To access an element with `theId` ID use the following property declaration
```
ilya-g: is the id case insensitive? Otherwise `theId` element won't be found by `theID` property

libraries/stdlib/js/src/kotlin/dom/Dom.kt
jonnyzzz: Reification is used to allow the cast to T (otherwise type erasure wipes the cast). I assume such a cast make sense on the JS backend.

The generated code seems to try checking types as I see it
``` var tmp$, tmp$_0, tmp$_1;
    var id = leftImage_metadata.callableName;
    tmp$ = leftImage.document.getElementById(id);
    if (tmp$ == null) {
      throw new NullPointerException('Element ' + id + ' is not found');
    }
    var element = tmp$;
    tmp$_1 = Kotlin.isType(tmp$_0 = element, HTMLCanvasElement) ? tmp$_0 : null;
    if (tmp$_1 == null) {
      throw new ClassCastException('Element ' + id + ' is not of type ' + toString(getKClass(HTMLCanvasElement).simpleName));
    }
    return tmp$_1;
```

jonnyzzz: > in the generated property getter 
What do you mean by it?

libraries/stdlib/js/src/kotlin/dom/Dom.kt
jonnyzzz: pushed the fix, thanks :)

libraries/stdlib/js/src/kotlin/dom/Dom.kt
ilya-g: I meant the JVM behavior where CHECKCAST happens as soon as a generic result being assigned to concrete type, that usually happens in generic getter. In JS there is indeed no such behavior and reified cast is required to achieve the same behavior.

The following snippet illustrates this difference between JVM and JS:
JVM https://pl.kotl.in/JwBgIiNBH
JS https://pl.kotl.in/YSE8ZhtUX

ilya-g: The concern about external interfaces is still actual: https://pl.kotl.in/GiIJeQJS-
They cannot be used as reified type arguments. Can there be a use case where we would need to delegate property of such external interface type?

jonnyzzz: As we discussed in a voice call, it looks like it will work OK with the current approach. It something would go wrong we may remove `reified` in the declaration. Yet another solution could be to add another declaration (with a different name) to implement an _unchecked_ way of accessing elements without `reified` declaration.

The `reified` generics do not work with `external` interfaces (and do work with external classes), Kotlin interfaces/classes implementing external interfaces are supported too. We do have 
```
public external abstract class Element
```
as the base class constraint for our generic. 


The good part of the `reified` generic is that we do have runtime checks. It is something where Kotlin could be safer. It looks good to still have it (and thus to constraint the API a bit)

jonnyzzz:
I've moved the PR to the kotlinx.html repository. Please fill free to close it

PR #2459, opened 2019-07-15
KT-15143 Kotlin: Colors&Fonts -> "Enum entry" should use Language Default -> Classes - Static field

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15143


klunnii:
Merged. Thanks!

PR #2462, opened 2019-07-16
Handle classpath snapshot changes better

gavra0:
Because Kapt Gradle task uses PathSensitivity.NONE for the input representing
the classpath structure, there are cases when paths of the files may change,
but because content is the same, there will be no incremental changes.

Classpath snapshot comparison did not handle this case correctly, but this
commit fixes that. In more details:
- classpath entries with the same path and that are not reported as changed
will have their information loaded from the previous snapshot
- any other entry (changed path or changed content) will be reloaded

Test: ClasspathSnapshotTest

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/incremental/ClasspathSnapshot.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/incremental/ClasspathSnapshot.kt
@@ -54,36 +54,20 @@ open class ClasspathSnapshot protected constructor(
             return KaptClasspathChanges.Unknown
         }
 
-        loadEntriesFor(changedFiles)
+        val unchangedBetweenCompilations = dataForFiles.keys.intersect(previousSnapshot.dataForFiles.keys).filter { it !in changedFiles }
+        val currentToLoad = dataForFiles.keys.filter { it !in unchangedBetweenCompilations }.also { loadEntriesFor(it) }
+        val previousToLoad = previousSnapshot.dataForFiles.keys.filter { it !in unchangedBetweenCompilations }
 
-        val currentHashAbiSize = changedFiles.sumBy { dataForFiles[it]!!.classAbiHash.size }
-        val currentHashesToAnalyze =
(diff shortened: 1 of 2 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/incremental/ClasspathSnapshot.kt
```diff
@@ -54,31 +54,30 @@ open class ClasspathSnapshot protected constructor(
...
-        val previousHashAbiSize = previousChanged.sumBy { previousSnapshot.dataForFiles.get(it)?.classAbiHash?.size ?: 0 }
+        val currentHashAbiSize = currentToLoad.sumBy { dataForFiles[it]!!.classAbiHash.size }
+        val currentHashesToAnalyze =
+            HashMap<String, ByteArray>(currentHashAbiSize).also { hashes ->
+                currentToLoad.forEach {
+                    hashes.putAll(dataForFiles[it]!!.classAbiHash)
+                }
+            }
+
+        val previousHashAbiSize = previousToLoad.sumBy { previousSnapshot.dataForFiles.get(it)?.classAbiHash?.size ?: 0 }
```
alexey-tsvetkov: Seems impossible that `previousSnapshot` has nullable entries for `previousToLoad`.
If this is the case, then creating a map for hashes could be extracted:
```
private fun hashesFor(filesToLoad: Collection<File>): Map<String, ByteArray> {
        val size = filesToLoad.sumBy { dataForFiles[it]!!.classAbiHash.size }
        return HashMap<String, ByteArray>(size).also { hashes ->
            filesToLoad.forEach {
                hashes.putAll(dataForFiles[it]!!.classAbiHash)
            }
        }
}
```

gavra0: Agree, it is more readable with your comment.

yanex:
Merged manually (master – 5a2ff8669132c9fedd99762f0d5fac5f4b7f7ea8, 1.3.50 – ef14310bc29f036a1ff82d550fcb8ba95f7efb4d).
Thanks!

PR #2464, opened 2019-07-17
Mark property as output instead of local state when needed

gavra0:
Kapt task incremental annotation processing cache directory
was using @LocalState annotation that was added in Gradle 4.3.
However, minimum supported version is 4.2, so this commit registers
this directory as output of a task for pre 4.3 versions. This is safe
as caching is enabled only for Gradle 4.3+.

Test: verified manually


yanex:
Merged manually (master – a108af76d663b8dba21545359ec88d2d950c9f14, 1.3.50 – ac16bb21d0522b61ae68e3f44fc3ebb155dab8d6).
Thanks!

Kelventri:
> Combinado manualmente (maestro - [a108af7](https://github.com/JetBrains/kotlin/commit/a108af76d663b8dba21545359ec88d2d950c9f14) , 1.3.50 - [ac16bb2](https://github.com/JetBrains/kotlin/commit/ac16bb21d0522b61ae68e3f44fc3ebb155dab8d6) ). ¡Gracias!

PR #2465, opened 2019-07-17
KT-31295 Scratches as project files

fedochet:
Allows to open project files with `.ws.kts` extension (`ws` is for `worksheet`) as Kotlin Scratches.

All existing scratch functionality works except for the module selection - it is automatically set to the current module (i.e. to the module in which `.ws.kts` file is located).

https://youtrack.jetbrains.com/issue/KT-31295

### DIFF
--- a/idea/tests/org/jetbrains/kotlin/idea/scratch/AbstractScratchRunActionTest.kt
+++ b/idea/tests/org/jetbrains/kotlin/idea/scratch/AbstractScratchRunActionTest.kt
@@ -192,7 +192,9 @@ abstract class AbstractScratchRunActionTest : FileEditorManagerTestCase() {
         val (_, scratchPanel) = getEditorWithScratchPanel(myManager, myFixture.file.virtualFile)
             ?: error("Couldn't find scratch panel")
 
-        configureOptions(scratchPanel, text, myFixture.module)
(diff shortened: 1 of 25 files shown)


idea/tests/org/jetbrains/kotlin/idea/scratch/AbstractScratchRunActionTest.kt
```diff
@@ -75,23 +96,27 @@ abstract class AbstractScratchRunActionTest : FileEditorManagerTestCase() {
...
-        doReplTest(mainFileName)
+        doTest(mainFileName, isRepl = true, isWorksheet = isWorksheet)
 
         ModuleRootModificationUtil.updateModel(myFixture.module) { model ->
             model.getModuleExtension(CompilerModuleExtension::class.java).inheritCompilerOutputPath(true)
         }
     }
 
-    fun doTest(fileName: String, isRepl: Boolean) {
+    fun doTest(fileName: String, isRepl: Boolean, isWorksheet: Boolean) {
```
fedochet: @NataliaUkhorskaya We were talking about splitting this method into two, but as I understand there's no need in that since we use in-memory editor with `configureByText`. But I may be wrong

NataliaUkhorskaya:
Merged manually

PR #2466, opened 2019-07-17
KT-23880: Enable incremental APT mode with KAPT by default

gavra0:
The feature has been present since 1.3.30 behind a flag. This
commit enables it by default.

Before this PR is merged, https://github.com/JetBrains/kotlin/pull/2462 should be merged as KaptIncrementalIT.testAddNewLine test fails without it.


yanex:
Merged manually (master – ad352355deb20283d0f8c91f749faf4e77b520a0, 1.3.50 – 89dcc4bba616531cb5e2918a76722347ff26d0e9).
Thanks!

PR #2468, opened 2019-07-19
KT-14758 Move statement up shouldn't move top level declarations above package and import directives

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14758

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinDeclarationMover.java
+++ b/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinDeclarationMover.java
@@ -287,8 +287,7 @@ public class KotlinDeclarationMover extends AbstractKotlinUpDownMover {
 
         PsiElement sibling = getLastNonWhiteSiblingInLine(firstNonWhiteSibling(sourceRange, down), editor, down);
 
-        // Either reached last sibling, or jumped over multi-line whitespace
-        if (sibling == null)  {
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinDeclarationMover.java
```diff
@@ -287,8 +287,7 @@ public boolean checkAvailable(@NotNull Editor editor, @NotNull PsiFile file, @No
 
         PsiElement sibling = getLastNonWhiteSiblingInLine(firstNonWhiteSibling(sourceRange, down), editor, down);
 
-        // Either reached last sibling, or jumped over multi-line whitespace
-        if (sibling == null)  {
+        if (sibling == null || sibling instanceof KtPackageDirective || sibling instanceof KtImportList)  {
```
NataliaUkhorskaya: This is also actual for expressions because of scripts.

idea/testData/codeInsight/moveUpDown/classBodyDeclarations/class/classAtImport.kt
```diff
@@ -0,0 +1,5 @@
+// MOVE: up
+// IS_APPLICABLE: false
+import foo.Bar
+
+<caret>class A
```
NataliaUkhorskaya: Add a test for script too

NataliaUkhorskaya:
Merged manually, thank you!

PR #2469, opened 2019-07-19
KT-10478  Move-statement doesn't work for methods with single-expression body and lambda as returning type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-10478


NataliaUkhorskaya:
Merged manually. Thank you for contributing!

PR #2470, opened 2019-07-19
Provide Array.fill in common stdlib

JakeWharton:
KT-32781

### DIFF
--- a/libraries/stdlib/common/src/generated/_Arrays.kt
+++ b/libraries/stdlib/common/src/generated/_Arrays.kt
@@ -6339,6 +6339,51 @@ public expect fun BooleanArray.copyOfRange(fromIndex: Int, toIndex: Int): Boolea
 public expect fun CharArray.copyOfRange(fromIndex: Int, toIndex: Int): CharArray
 
 /**
+ * Fills original array with the provided value.
+ */
+public expect fun <T> Array<T>.fill(element: T, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun ByteArray.fill(element: Byte, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun ShortArray.fill(element: Short, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun IntArray.fill(element: Int, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun LongArray.fill(element: Long, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun FloatArray.fill(element: Float, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun DoubleArray.fill(element: Double, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun BooleanArray.fill(element: Boolean, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun CharArray.fill(element: Char, fromIndex: Int = 0, toIndex: Int = size): Unit
+
+/**
  * Returns the range of valid indices for the array.
  */
 public val <T> Array<out T>.indices: IntRange
--- a/libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
+++ b/libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
@@ -6,8 +6,8 @@
 package templates
 
 import templates.Family.*
-import templates.Ordering.stableSortNote
 import templates.Ordering.appendStableSortNote
+import templates.Ordering.stableSortNote
 
 object ArrayOps : TemplateGroupBase() {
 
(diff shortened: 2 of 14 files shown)


libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
```diff
@@ -1273,17 +1273,40 @@ object ArrayOps : TemplateGroupBase() {
...
 
+        on(Platform.JVM) {
+            suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")
+            body {
+                """
+                java.util.Arrays.fill(this, fromIndex, toIndex, element)
+                """
+            }
+        }
+        on(Platform.JS) {
```
JakeWharton: JS has https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill but not sure it's worth delegating to or not.

libraries/stdlib/common/src/generated/_Arrays.kt
```diff
@@ -6338,6 +6338,51 @@ public expect fun BooleanArray.copyOfRange(fromIndex: Int, toIndex: Int): Boolea
  */
 public expect fun CharArray.copyOfRange(fromIndex: Int, toIndex: Int): CharArray
 
+/**
+ * Fills original array with the provided value.
+ */
+public expect fun <T> Array<T>.fill(element: T, fromIndex: Int = 0, toIndex: Int = size): Unit
```
Prototik: Isn't `Array<in T>` enough here?

JakeWharton: This is the original signature of the function. The PR just moves it to common.

If we want a signature change I would prefer to do it separately.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
ilya-g: I think it's definitely worth it, however we should still check argument values to enforce our contract of this method.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
```diff
@@ -1273,17 +1273,40 @@ object ArrayOps : TemplateGroupBase() {
...
+            if (fromIndex !in 0..size) {
+              throw IndexOutOfBoundsException("fromIndex ${'$'}fromIndex out of range [0, ${'$'}size]")
+            }
+            if (toIndex > size) {
+              throw IndexOutOfBoundsException("toIndex ${'$'}toIndex out of range [${'$'}fromIndex, ${'$'}size]")
+            }
+            require(fromIndex <= toIndex) { "fromIndex(${'$'}fromIndex) > toIndex(${'$'}toIndex)" }
+            for (index in fromIndex until toIndex) {
+                this[index] = element
+            }
```
ilya-g: Given that Kotlin/Native has its own implementation and Kotlin/JS will be delegating to array's method, there seem to be no need in this manual implementation.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
JakeWharton: Ok.

 > we should still check argument values to enforce our contract of this method

Yes JS seems to just silently ignore out-of-bounds and to < from 😑

> We'll need to provide polyfills though, as we target ES 5 that doesn't have these methods in the standard.

Do you know of a similar API does this that I can copy? I can go searching for one, but thought you might know one off-hand.

libraries/stdlib/common/src/generated/_Arrays.kt
ilya-g: @Prototik, the current signature won't benefit from changing `Array<T>` to `Array<in T>`.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
JakeWharton: Oh, ok. Do I point to its implementation somehow? Or remove it from being generated? I don't really understand how native tracks these kind of changes being in a separate repo.

ilya-g: Yes, you need to call implementation methods (e.g. https://github.com/JetBrains/kotlin-native/blob/9e26d32/runtime/src/main/kotlin/kotlin/collections/ArrayUtil.kt#L78) here, however building Kotlin/Native together with Kotlin can be quite complicated, so feel free to leave this exercise to maintainers :-)

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
ilya-g: The MDN article you have linked contains a polyfill implementation. We keep 'em collection of polyfills used in stdlib here: https://github.com/JetBrains/kotlin/blob/1.3.50/libraries/stdlib/js-v1/src/js/polyfills.js

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
JakeWharton: Ok I left it as an `arrayFill` function call, but obviously implementations will need to be added to that file similar to `arrayCopy`.

It's not clear to me whether the responsibility for the argument checks should live in the template for native or in the implementation. If it's in the template, let me know and I'll update.

libraries/tools/kotlin-stdlib-gen/src/templates/Arrays.kt
JakeWharton: Updated. PTAL!

ilya-g:
Thanks, I've merged it manually with several edits, see the commit range https://github.com/JetBrains/kotlin/compare/1a6069382ecafc4c6d79bc109c83f67f0aca6188~1...2e445ebad6b95bee38749d85274abe3742baae68

JakeWharton:
Thanks! I hope these PRs help more than they just obligate you to do more work.

PR #2471, opened 2019-07-20
Implement List.firstIndex to match List.lastIndex

IvanMwiruki:


ilya-g:
This proposal needs some justification.

What are the benefits of having this extension when it can be replaced with a literal `0`?
I see that you have defined it that it's not always zero, but that looks questionable. What is the reason of having `firstIndex` equal to -1 for an empty list? I'd expect that the range `list.firstIndex..list.lastIndex` is equal to or at least returns the same values as the range `list.indices`.

IvanMwiruki:
Your points are fair. Resolving.

PR #2472, opened 2019-07-20
KT-23461 `Move statement up/down` attaches a comment block to the function being moved

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23461

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.java
+++ b/idea/src/org/jetbrains/kotlin/idea/codeInsight/upDownMover/KotlinExpressionMover.java
@@ -350,6 +350,28 @@ public class KotlinExpressionMover extends AbstractKotlinUpDownMover {
             }
         }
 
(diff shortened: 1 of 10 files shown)


goodwinnk:
Thank you for this pull request. This was a tough one for the review. 

I have accepted the commit (https://github.com/JetBrains/kotlin/commit/ed28a062854d338f92d3e614a750651db1844ba4), did some refactoring with function extraction (https://github.com/JetBrains/kotlin/commit/11de99f9d72b01d799e8a142b214c2bcb72994aa) and fixed the bug with moving comments (https://github.com/JetBrains/kotlin/commit/fc4865a440944fed5b146111ba6abe180aba5a9b). 

At the end I have found that the same problem can be reproduced with additional tests :(

```
// MOVE: up
fun foo() {
    // comment
    if (true) {}
    <caret>fun nested() {}
}
```

As a conclusion, I think the whole mover should get a deep refactoring in the future.

PR #2473, opened 2019-07-20
KT-28471: "Add initializer" quickfix initializes non-null variable with null

devbridie:
https://youtrack.jetbrains.com/issue/KT-28471

Remove call to `CodeInsightUtils.defaultInitializer(descriptor.type)` in favor of `TODO()`.


idea/src/org/jetbrains/kotlin/idea/quickfix/InitializePropertyQuickFixFactory.kt
```diff
@@ -57,9 +57,7 @@ object InitializePropertyQuickFixFactory : KotlinIntentionActionsFactory() {
 
         override fun invoke(project: Project, editor: Editor?, file: KtFile) {
             val element = element ?: return
-            val descriptor = element.resolveToDescriptorIfAny() as? PropertyDescriptor ?: return
-            val initializerText = CodeInsightUtils.defaultInitializer(descriptor.type) ?: "null"
```
dimonchik0036: May be we should replace only the default value `null` with TODO()`?

devbridie:
Suggestion applied.

PR #2476, opened 2019-07-22
psi2ir: precompute sort keys when generating members

pyos:
On kotlinx.serialization, this improves the performace of the compiler with -Xuse-ir by about 5% (average of 30 runs after 10 warmup runs: 3096 ms -> 2970 ms, minimum: 3037 ms -> 2884 ms). Flamegraphs for comparison (careful, huge SVG images):

https://drive.google.com/open?id=1BlIFN4ivSKrXjY6vZ0xkZymwucGc7bSI
https://drive.google.com/open?id=1xIW2IvW7LMCaAe4P2dOGOG8nCcpWI9PW

Note that calling `sortedBy` with `DESCRIPTOR_RENDERER::render` would be equivalent to the code removed by this commit -- perhaps a version that uses additional space to save time should be added to the standard library?


dnpetrov:
Looks good to me.

max-kammerer:
Thanks!

PR #2477, opened 2019-07-22
Report non-incremental annotation processors correctly

gavra0:
This commit fixes an issue when all APs would be reported as non-incremental,
even if only a single one is non-incremental. Now, additional declared type
is added that is used to denote processors that are incremental, but have
been forced to run non-incrementally in presence of non-incremental APs.

This means that only APs that do not support incremental annotation processing,
or APs that are dynamic and are non-incremental at runtime will be reported.


yanex:
Merged manually (master – be3fe9495cbcb0b49019ef9bfa9a8ecae7d89c3c, 1.3.50 – da8668ad90ce7b6bc172fad2f11dedc91b61a200).
Thanks!

PR #2478, opened 2019-07-22
KT-32797 Guard clause (check/require/...) inspection: handle case with if/else correctly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32797


mglukhikh:
Merged 👍

PR #2479, opened 2019-07-23
KT-32496 "Problems Detected" dialog message about conflicting declarations on moving file to another package is absolutely unreadable

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32496


Alefas:
From YT ticket resolution: Thank you for your work but I decided not to pull it. The reason is simple - your PR is pretty small and I made my own commit with some additional refactoring and improvement so it was most simple way to merge theese changes without your PR.

PR #2481, opened 2019-07-23
KT-32479 False positive "Redundant overriding method" with derived property and base function starting with "get" (Accidental override)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32479

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
@@ -10,13 +10,19 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ClassKind
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.load.java.descriptors.JavaMethodDescriptor
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.*
 import org.jetbrains.kotlin.resolve.BindingContext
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -77,6 +79,14 @@ class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLoc
         val superCallMethodName = superSelectorExpression.getCallNameExpression()?.text ?: return false
         return function.name == superCallMethodName
     }
+    
+    private fun KtNamedFunction.hasDerivedProperty(): Boolean {
+        if (valueParameters.isNotEmpty()) return false
+        val functionName = nameAsName ?: return false
+        if (!functionName.asString().startsWith("get")) return false
+        val propertyName = SyntheticJavaPropertyDescriptor.propertyNameByGetMethodName(functionName)?.asString() ?: return false
+        return containingClassOrObject?.body?.children?.any { it is KtProperty && it.name == propertyName } == true
```
dimonchik0036: Use `containingClassOrObject?.declarations?.find { it is KtProperty && it.name == propertyName } != null`.

idea/src/org/jetbrains/kotlin/idea/inspections/KotlinRedundantOverrideInspection.kt
```diff
@@ -77,6 +83,18 @@ class KotlinRedundantOverrideInspection : AbstractKotlinInspection(), CleanupLoc
...
+    
+    private fun KtNamedFunction.hasDerivedProperty(functionDescriptor: FunctionDescriptor?): Boolean {
+        if (functionDescriptor == null) return false
+        val functionName = nameAsName ?: return false
+        if (!canBePropertyAccessor(functionName.asString())) return false
+        val isSetter = functionDescriptor.returnType?.isUnit() == true
+        if (isSetter && valueParameters.size != 1 || !isSetter && valueParameters.isNotEmpty()) return false
+        return SyntheticJavaPropertyDescriptor.propertyNamesByAccessorName(functionName).any {
+            val propertyName = it.asString()
+            containingClassOrObject?.declarations?.find { d -> d is KtProperty && d.name == propertyName } != null
```
dimonchik0036: We should also check a type. This is a false negative:
```kotlin
open class A {
    open fun isFoo(): Boolean = true
}
class B : A() {
    private val isFoo: Int = 42
    override fun isFoo(): Boolean = super.isFoo()
}
```

PR #2482, opened 2019-07-23
KT-19744 "Change Signature" is not available on `constructor` keyword in primary constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19744


darthorimar:
Thank you!

PR #2483, opened 2019-07-23
JVM_IR: Change CallableReferenceLowering to match JVM backend

sfs:
There are some differences in the compilation of callable references between the JVM and JVM_IR backends. In particular, the JVM backend inlines the bodies of lambdas into the generated `Lambda` subclasses and stores bound receivers into the receiver field of the `CallableReference` class. The former seems to be expected by the inliner and both are required for reflection to work properly.

This PR refactors `CallableReferenceLowering` to implement these changes and additionally moves `CallableReferenceLowering` before `LocalDeclarationsLowering` to reduce code duplication. Previously, both lowerings contained code to add fields for captured variables.

These changes exposed some problems in `LocalDeclarationsLowering` (non-deterministic statement order) and `SyntheticAccessorLowering` (accessors for super calls were broken and accessor names could clash), which are also fixed.


sfs:
Note that 2c0f9fc contains some code which could be simplified/removed once fc5e9f5 is merged.

There is also still some code duplication between `CallableReferenceLowering` and `FunctionNVarargInvokeLowering` which could be removed now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -150,38 +150,38 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
         val source = this
         return buildFun {
             origin = JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR
             name = source.accessorName()
             visibility = Visibilities.PUBLIC
             isSuspend = false // do not generate state-machine for synthetic accessors
         }.also { accessor ->
-            accessor.parent = if ((source.parent as? IrClass)?.isInterface == true) {
+            // Resolve the actual parent for a super call
+            val parent = expression.superQualifierSymbol?.let { expression.dispatchReceiver!!.type.classOrNull!!.owner } ?: parent
```
madsager: This code seems pretty convoluted. I'm having a hard time figuring out what is happening here. Maybe you can expand the comment to explain what the receiver of makeSimpleFunctionAccessor is in these two cases and show by examples why this parent computation does the right thing?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -124,6 +124,7 @@ private val jvmFilePhases =
 
         enumWhenPhase then
         singletonReferencesPhase then
+        callableReferencePhase then
```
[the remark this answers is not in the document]
sfs: Right, the new order assumes that CallableReferenceLowering runs before LocalDeclarationsLowering. I've added the constraint there.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -29,155 +17,153 @@ import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
...
+
+        if (expression.valueArgumentsCount < FunctionInvokeDescriptor.BIG_ARITY ||
+            !expression.symbol.owner.parentAsClass.defaultType.isFunctionOrKFunction())
+            return expression
+
+        return IrCallImpl(
+            expression.startOffset, expression.endOffset,
+            expression.type, functionNInvokeFun.symbol, functionNInvokeFun.descriptor,
+            1, expression.origin,
+            (expression as? IrCall)?.superQualifier?.let { context.ir.symbols.externalSymbolTable.referenceClass(it) }
```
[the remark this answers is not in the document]
sfs: There shouldn't be and you're right, the code is suspect. Right now, `superQualifier(Symbol)` is always null in this line and I've remove the code.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -29,155 +17,153 @@ import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
...
-                        (expression as? IrCall)?.superQualifier?.let { context.ir.symbols.externalSymbolTable.referenceClass(it) }
-                    ).apply {
-                        putTypeArgument(0, expression.type)
-                        dispatchReceiver = expression.dispatchReceiver
-                        extensionReceiver = expression.extensionReceiver
-                        putValueArgument(0, vararg)
-                    }
-                } else expression
+    // This pass ignores inlined function references and (some) SAM conversions.
+    private val ignoredFunctionReferences = mutableSetOf<IrFunctionReference>()
```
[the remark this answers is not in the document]
sfs: I agree in this case. Moved to the top and expanded the comment.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -150,38 +150,38 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
             accessor.copyTypeParametersFrom(source, JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR)
             // When a method is defined in class C1 but called on C1's subclass C2, source.dispatchReceiverParameter.type can be resolved
             // to C1, while the method symbol still bound to C2, which expects a receiver of type C2. Therefore, we need to specify
-            // dispatchReceiver's type using super qualifier's type explicitly.
-            accessor.copyValueParametersToStatic(
-                source,
-                JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR,
-                superQualifierSymbol?.owner?.defaultType ?: source.dispatchReceiverParameter?.type
-            )
+            // dispatchReceiver's type using the disptach receiver's type explicitly (the super qualifier points to C1, not C2).
```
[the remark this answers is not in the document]
sfs: Fixed

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
sfs: I've simplified the code and added some comments. To be honest, this code *is* fragile and currently relies on having correct IR types. I don't know how to find the correct enclosing class for super calls otherwise.

sfs:
Rebased to latest master (which removed CrIrType and simplified some things) and addressed the comments.

compiler/testData/codegen/boxInline/smap/anonymous/lambda.kt
```diff
@@ -1,5 +1,4 @@
 // FILE: 1.kt
-// IGNORE_BACKEND: JVM_IR
 package builders
```
max-kammerer: @sfs Please just add new line here and in next file too

compiler/testData/lineNumber/custom/functionCallWithLambdaParam.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: Please move ignore directive at file end

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -150,38 +150,59 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
         }.also { accessor ->
-            accessor.parent = if ((source.parent as? IrClass)?.isInterface == true) {
-                // Accessors for interfaces are only for private methods. They should always be placed in DefaultImpls. The only exception
-                // is private methods annotated with @JvmDefault; Only in such cases JVM default method is generated.
-                // TODO: Handle the exception after targeting Java 8 or newer, where JVM default methods are available.
-                context.declarationFactory.getDefaultImplsClass(source.parent as IrClass)
-            } else {
-                source.accessorParent()
-            }
+            // Find the right container to insert the accessor. Simply put, when we call a function on a class A,
```
max-kammerer: @dnpetrov @gogabr Please take a look to changes in this lower

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -391,22 +412,25 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
+    // Counter used to disambiguate accessors to functions with the same base name.
+    private var nameCounter = 0
+
     private fun IrFunction.accessorName(): Name {
         val jvmName = context.state.typeMapper.mapFunctionName(
             descriptor,
             OwnerKind.getMemberOwnerKind(parentAsClass.descriptor)
         )
-        return Name.identifier("access\$$jvmName")
+        return Name.identifier("access\$$jvmName\$${nameCounter++}")
```
max-kammerer: @gogabr @sfs Could we use old naming scheme from current backend? ('getAccessorNameSuffix' in syntheticAccessorUtil.kt) This will allow to unmute a bunch of bytecode tests.

max-kammerer: If yes it could be postponed for another PR

sfs: I considered this, but right now there's a bug in the old naming scheme which can lead to name clashes (reported as [KT-32870](https://youtrack.jetbrains.com/issue/KT-32870)). Is this naming scheme part of any public ABI?

However, I didn't see that there were tests which depend on the naming scheme. In that case it might be better to use it now and fix it later. I'll have a look.

compiler/testData/codegen/boxInline/smap/anonymous/lambda.kt
sfs: Done

compiler/testData/lineNumber/custom/functionCallWithLambdaParam.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
max-kammerer: There are some tests, e.g: bytecodeText/companion, bytecodeText/jvm8/jvmDefault/compatibility.

It's a part of public api in a sence that inlined code wouldn't work after binary library update. Change in naming scheme should be accepted by Language Committee.
Seems as code generated by IR backend should be binary compatible with old one we should support old scheme in any way,  should it @udalov?

sfs: It turns out that this is not straightforward to implement.

For one, the naming scheme depends on the origin of fields (e.g., moved from companion classes, generated for locals) which is not recorded in the ir backend. For another, the "super call" suffix also needs to be added from field accesses (IrGetField, IrSetField), which currently don't record a `superQualifierSymbol` for static fields. The latter ends up breaking `testProtectedStatic2`, which was why I had to put in the counter in the first place.

If possible, I'd like to postpone this change.

max-kammerer: @sfs Actually protected member calls from public inline functions would be prohibited in 1.4 (https://youtrack.jetbrains.com/issue/KT-21178). So most of inline cases around synthetic accessors shouldn't be considered as public ABI, except ones with @PublishedApi internal functions

max-kammerer: Update comment above^

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -32,140 +20,151 @@ import org.jetbrains.kotlin.backend.jvm.codegen.isInlineIrExpression
...
-                    +irCall(loweredFunctionReference.functionReferenceConstructor.symbol).apply {
-                        expression.getArguments().forEachIndexed { index, argument ->
-                            putValueArgument(index, argument.second)
-                        }
-                    }
-                }
+            if (valueArgument is IrFunctionReference) {
+                ignoredFunctionReferences.add(valueArgument)
+            } else if (valueArgument is IrBlock) {
+                ignoredFunctionReferences.addIfNotNull(valueArgument.statements.firstIsInstanceOrNull())
```
max-kammerer: Actually it should be `singletIsInstanceOrNull`

sfs: And wouldn't that be a nice function to have in the standard library? :-)

Fixed!

max-kammerer:
@sfs Could you rebase on master (there is a compilation error on it)?
Please also extract SMAP related changes to separate PR (I need to look deeper into this change, maybe you can write something about difference between old and new backend)?

sfs:
@max-kammerer Done. What happened was that I rebased earlier in response to f66b475867501ed86ef57cc0d08b9fdeab6df0f2, but I missed that this was then reverted in 3729c4e7705f686fcf8161f683ddef8a22e28b61.

The change in SMAP.kt isn't entirely independent - not having it breaks inlining in 6 tests involving anonymous objects. I've split it off into a separate PR: #2494

sfs:
As for the difference between the old and the new backend, as far as I can tell the old backend never reaches this part of the inliner with an `IdenticalSourceMapper`, which is why it didn't break when calling the unimplemented `mapLineNumber` override.

max-kammerer:
> As for the difference between the old and the new backend, as far as I can tell the old backend never reaches this part of the inliner with an IdenticalSourceMapper, which is why it didn't break when calling the unimplemented mapLineNumber override.

Yes, but seems there is some difference in code generation or codegen construction for inline methods. I will look into it deeper.
Thanks for changes!

compiler/testData/codegen/boxInline/smap/anonymous/lambda.kt
```diff
@@ -1,5 +1,6 @@
-// FILE: 1.kt
 // IGNORE_BACKEND: JVM_IR
+// FILE: 1.kt
+
```
max-kammerer: New line causes smap data shift:  https://teamcity.jetbrains.com/viewLog.html?buildId=2450360&buildTypeId=Kotlin_dev_CompilerWithTests

max-kammerer: It's best to revert to original:
```
// FILE: 1.kt
// IGNORE_BACKEND: JVM_IR
package builders
``

File: compiler/testData/lineNumber/custom/functionCallWithLambdaParam.kt [hunk 12]
```diff
@@ -13,3 +13,4 @@ fun foo(f: () -> Unit) {
 }
 
 // 2 6 9 12 13 3 4 7 8
+// IGNORE_BACKEND: JVM_IR
```
max-kammerer: Could you look into this test too: it's fails on Windows: https://teamcity.jetbrains.com/comparisonFailureDiffView.html?buildId=2450360&testId=453380
Maybe additional empty line is required to separate numbers from ignore directive

compiler/testData/codegen/boxInline/smap/anonymous/lambda.kt
sfs: Done.

compiler/testData/lineNumber/custom/functionCallWithLambdaParam.kt
sfs: It fails because of how `AbstractLineNumberTest#compareCustom` is implemented. The line number comment has to be the last (non-empty) line in the file for the non-IR test.

I've fixed it by swapping the comments for now, although it might make sense to change `AbstractLineNumberTest` in the future.

sfs:
> Yes, but seems there is some difference in code generation or codegen construction for inline methods. I will look into it deeper.

Sorry that I can't be of more help in this. I didn't look too deeply into the inliner code.

max-kammerer:
Many Thanks!

NB: The last major difference between backends in lambda generation is variable capturing scheme

PR #2485, opened 2019-07-24
KT-32210: Handle long compiler plugin options

gavra0:
Some of the KAPT compiler plugin options were longer than 0xFFFF bytes in
UTF-8 encoding, causing UTFDataFormatException when serializing them. This
commit avoids that issue by encoding changed files and changed classpath types
one by one instead of joining them to a single string.


yanex:
Merged manually (master – 1c63d3aa2f42c8dc89123d88d512f230852bdce4, 1.3.50 – 9141fe93a54c6c2d74be04ba899a1c54fb4f9214).

PR #2486, opened 2019-07-25
Avoid eager deserialization of array classes in ArrayConstructorLowering

sfs:
This is a minor performance fix. Currently, `ArrayConstructorLowering` always looks at the declarations contained in the array classes (`Array<T>`, `IntArray`, ...), which leads to some unnecessary deserialization.


max-kammerer:
Thanks!

PR #2487, opened 2019-07-25
Fix gradle integration tests on Windows

gavra0:
Android Gradle Plugin 3.0 and 3.1 use aapt that does not handle
long paths on Windows. This commit uses shorter paths for the project
working directory on Windows.

Test: running integration tests on Windows


altsvetkov:
Merged manually with minor changes. Thanks a lot for fixing the tests!

PR #2490, opened 2019-07-26
KT-28824 Adds artifact for jvm-abi-gen for use with embeddable compiler

tyvsmith:
Adds ":plugin:jvm-abi-gen-embeddable" Gradle target in plugins/jvm-abi-gen/embeddable dir.

Testing: Built artifact locally and used with embeddable compiler.

Resolves https://youtrack.jetbrains.com/issue/KT-28824


alexey-tsvetkov:

PR #2493, opened 2019-07-30
KT-32616 "To ordinary string literal" doesn't remove indents, newlines and `trimIndent`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32616


idea/src/org/jetbrains/kotlin/idea/intentions/ToOrdinaryStringLiteralIntention.kt
```diff
@@ -65,4 +66,42 @@ class ToOrdinaryStringLiteralIntention : SelfTargetingOffsetIndependentIntention
...
+
+        val stringTemplateText = entries
+            .joinToString(separator = "") { it.text }
+            .let { if (marginPrefix != null) it.trimMargin(marginPrefix) else it.trimIndent() }
+            .escape()
+
+        return TrimIndentCall(qualifiedExpression, stringTemplateText)
+    }
+
+    data class TrimIndentCall(
```
darthorimar: Should be private class

darthorimar:
@t-kameyama thank you for your pull request ^_^

P.S. could you please do not use force push in your pull request? It very hard to follow what changed since the last update

PR #2495, opened 2019-07-31
KT-33056: Fix how parent class loader for KAPT processors is found

gavra0:
When instantiating annotation processors, the class loader where they
are defined needs to have the correct parent class loader set up. In
order to find the parent class loader, we need to find the most specific
class loader that contains javac compiler classes (and does not contain
e.g. KAPT classes). This commit changes how that search is done, and it
simply returns the class loader in which the com.sun.tools.javac.util.Context
is defined, as the is the class we need on the annotation processor
classpath.


yanex:
Merged into `master`. Thanks for the contribution!

PR #2496, opened 2019-07-31
FIR build through LightTree

vana06:


mglukhikh:
I asked to remove merge commits from PR and to create a new one.

PR #2500, opened 2019-08-02
spec docs: Fix Type parameters position

satoshun:


dzharkov:

PR #2501, opened 2019-08-02
KT-33028, KT-33050: Fix how kapt invokes javac on JDK 9+

gavra0:
This commit fixes KT-33028 by not setting empty boot classpath when
running on JDK9+. When compiling with -source 8 and below, this allows
classes from java.base module to be available in the boot classpath.

This commit also fixes KT-33050 by passing the source level of Java compile
task in the KAPT javac options. This is important as some annotation processors
are using ProcessingEnvironment.getSourceVersion() in order to decide what
code to generate.

Test: Kapt3IT test added for worker and non-worker invocations

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
@@ -257,7 +257,7 @@ class Kapt3KotlinGradleSubplugin : KotlinGradleSubplugin<KotlinCompile> {
     }
 
     // This method should be called no more than once for each Kapt3SubpluginContext
-    private fun Kapt3SubpluginContext.buildOptions(aptMode: String): List<SubpluginOption> {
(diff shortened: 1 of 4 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
```diff
@@ -428,14 +428,18 @@ class Kapt3KotlinGradleSubplugin : KotlinGradleSubplugin<KotlinCompile> {
             kaptTask.annotationProcessorOptionProviders.add(it)
         }
 
+        val dslJavacOptions = kaptExtension.getJavacOptions().toMutableMap()
+        if (javaCompile != null && "-source" !in dslJavacOptions) {
```
technoir42: Probably need to check for `--release N` too which combines `-source N`, `-target N` and `-bootclasspath <from-N>` http://openjdk.java.net/jeps/247

gavra0: Good catch, I was not aware of this option.

yanex:
Merged into `master`. Thanks for the contribution!

PR #2503, opened 2019-08-03
KT-9065 Wrong result when move statement through if block with call with lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-9065

Also fixed:
[KT-17887 Moving statement (Ctrl/Cmd+Shift+Down) messes with use block](https://youtrack.jetbrains.com/issue/KT-17887)


vladimirdolzhenko:
Thank you for you contribution, merged.

PR #2504, opened 2019-08-04
KT-14757 Move statement up breaks code in function parameter list

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14757


klunnii:
Thank you! merged

PR #2505, opened 2019-08-04
Improve error message

LouisCAD:
Fore the case where a target has been created or configured twice with different presets,
avoid showing "null" inside the error message if existing target has not preset,
and append a point.


h0tk3y:
Thanks!

h0tk3y:
@LouisCAD Sorry, I've just noticed that this pull request targeted the `1.3.50` branch. I've manually cherry-picked it to `master` as well.
Please submit pull requests to `master` instead, since:
* commits from the release branches are usually not intended for porting to `master`, and such a commit may accidentally end up lost in a single release branch;
* the release branches are considered to be under stabilizing process, and we carefully cherry-pick commits from `master` into the release branches.

PR #2506, opened 2019-08-04
Use clearer syntax for lambda execution

LouisCAD:
That way, it's easier to see that we refer to the property just above and execute the passed lambda.


LouisCAD:
Closed in favor of #2508.

PR #2507, opened 2019-08-05
Fix handling of inline classes in DefaultParameterInjector

sfs:
There are some cases where `DefaultParameterInjector` adds unnecessary boxing for inline class arguments, since `IrClass.underlyingType` doesn't recursively unfold inline classes. Since the rules for unfolding inline classes differ by backend I've fixed this by adding a new function in the `Ir` class and removing the `IrClass.underlyingType` function.

I've added a test for this, as well as for another issue where default vararg arguments allocate an unnecessary array. I've got a different fix in the works for the latter, but we might as well add the test now.


gogabr:
Thank you!

PR #2508, opened 2019-08-05
Use receiver syntax for lambda execution

LouisCAD:
That way, it's easier to see that we refer to the property just above and execute the passed lambda.

Closes #2506


h0tk3y:
Thanks!

PR #2509, opened 2019-08-06
Add Pair to Triple extention method

richarddd:
This extension method allows for easy creation of Triples, i.e:
```kotlin
1 to 2 to 3 //Triple(1,2,3)
```


ilya-g:
Currently the expression `1 to 2 to 3` has type `Pair<Pair<Int, Int>, Int>`. The proposed function will change it to `Triple<Int, Int, Int>` and thus is not a source compatible change.

ilya-g:
I don't see how to resolve this problem of the proposed function, so I'm closing this for now.

PR #2510, opened 2019-08-06
JVM IR: Add a lowering to replace varargs with arrays

sfs:
Currently the JVM IR backend handles varargs directly in `ExpressionCodegen`. This PR instead adds a lowering to replace all vararg calls with explicit array arguments and removes the special handling from the rest of the compiler. As a side benefit, I've added a utility for building arrays in the IR (which is JVM specific right now, because of the way we handle arrays with spreads).

This fixes some corner cases, such as the `arrayOf` function not generating reified operation markers. Incidentally, this bug caused `testKt8047` to pass before, but this was a fluke. There is an anonymous object in this test which was not specialized. I've made this more clear in `testKt8047_2`, which uses a reified type check instead of `arrayOf`.

The only difference in the generated code should be that arrays are copied using `System.arraycopy`, instead of `Arrays.copyOf`, because this was easier in the IR.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -323,4 +323,57 @@ class JvmSymbols(
...
+        }
+
+        klass.addFunction("add", irBuiltIns.unitType).apply {
+            addValueParameter("element", irBuiltIns.anyNType)
+        }
+
+        klass.addFunction("size", irBuiltIns.intType)
+
+        klass.addFunction("toArray", irBuiltIns.arrayClass.typeWith(irBuiltIns.anyNType)).apply {
+            addValueParameter("a", irBuiltIns.arrayClass.typeWith(irBuiltIns.anyNType))
```
ting-yuan: Would it be simpler to add `toArray(): Array<T>` in `SpreadBuilder`, instead of using `toArray(array<T>): Array<T>`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/VarargLowering.kt
```diff
@@ -0,0 +1,115 @@
...
+        val function = expression.symbol
+
+        // Replace empty varargs with empty arrays
+        for (i in 0 until expression.valueArgumentsCount) {
+            if (expression.getValueArgument(i) != null)
+                continue
+
+            val parameter = function.owner.valueParameters[i]
+            if (parameter.varargElementType != null && !parameter.hasDefaultValue()) {
+                val arrayType = parameter.type.substitute(expression.typeSubstitutionMap)
```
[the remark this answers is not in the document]
sfs: Actually we do want the array type here. I've added a comment to make it more clear.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/VarargLowering.kt
```diff
@@ -0,0 +1,115 @@
...
+                expression.putValueArgument(i, createBuilder().irArrayOf(arrayType))
+            }
+        }
+
+        // Lower arrayOf calls
+        return if (function.isArrayOf) expression.getValueArgument(0)!! else expression
```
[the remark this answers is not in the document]
sfs: It's actually safe, since the call to `function.isArrayOf` ensures that the function in question only has a single vararg parameter and the code above ensure that this parameter is not null. I've added a comment to make it less confusing.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
sfs: It would be, but the jvm doesn't allow us to have nice things. Briefly, arrays have limited subtyping - you cannot cast `[Ljava/lang/Object;` to, e.g., `[Ljava/lang/Integer;` unless the array was originally allocated with this type. And since we compile generics through erasure, a function like `toArray<T>(): Array<T>` doesn't have enough information to allocate an array of the correct type.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -72,24 +72,17 @@ internal class CallableReferenceLowering(private val context: JvmBackendContext)
...
-            return expression
-
-        return IrCallImpl(
-            expression.startOffset, expression.endOffset,
-            expression.type, functionNInvokeFun.symbol, functionNInvokeFun.descriptor,
-            1, expression.origin
-        ).apply {
-            putTypeArgument(0, expression.type)
-            dispatchReceiver = expression.dispatchReceiver
-            extensionReceiver = expression.extensionReceiver
```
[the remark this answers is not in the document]
sfs: Because the function in question doesn't actually have type parameters or an extension receiver parameter. Previously the IrCall was just built manually, with an incorrect number of type arguments.

sfs:
Rebased on master, since #2511 was merged. Also due to unrelated changes, `testKt8047` and `testKt8047_2` are now unmuted.

compiler/testData/codegen/bytecodeText/inlineClasses/passInlineClassesWithSpreadOperatorToVarargs.kt
```diff
@@ -17,8 +21,6 @@ fun test(u1: UInt, u2: UInt, us: Array<UInt>) {
 // 2 INVOKESTATIC UInt\.box
 // 0 INVOKEVIRTUAL UInt.unbox
 
-// 2 CHECKCAST \[LUInt
```
max-kammerer: @sfs Please customize test for both backend via:
...TEMPLATES ...

// JVM_TEMPLATES
// 2 CHECKCAST \[LUInt

// JVM_IR_TEMPLATES
// 0 CHECKCAST \[LUInt
// Maybe smth else

max-kammerer: like forInObjectArrayIndices.kt for example

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrArrayBuilder.kt
```diff
@@ -0,0 +1,171 @@
...
+fun JvmIrBuilder.irArrayOf(arrayType: IrType, elements: List<IrExpression> = listOf()): IrExpression =
+    irArray(arrayType) { elements.forEach { +it } }
+
+private class IrArrayElement(val expression: IrExpression, val isSpread: Boolean)
+
+class IrArrayBuilder(val builder: JvmIrBuilder, val arrayType: IrType) {
+    // We build unboxed arrays for inline classes (UIntArray, etc) by first building
+    // an unboxed array of the underlying primitive type, then coercing the result
+    // to the correct type.
+    val unwrappedArrayType = arrayType.unboxInlineClass()
```
max-kammerer: It's not clear should this utility perfrom any logic related to inline classes or it should be covered by inline class lowering.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -348,84 +348,55 @@ class ExpressionCodegen(
...
-            }
+            require(arg != null) { "Null argument in ExpressionCodegen for parameter ${irParameter.render()}" }
+            callGenerator.genValueAndPut(irParameter, arg, parameterType, this, data)
         }
 
         expression.markLineNumber(true)
 
         // Do not generate redundant markers in continuation class.
-        if (callee.isSuspend && !irFunction.isInvokeSuspendInContinuation()) {
+        if (callee.isSuspend && !irFunction.isInvokeSuspendInContinuation())
```
gogabr: We prefer to keep the braces in such cases. (Makes it harder to accidentally botch the scope of the conditional.)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrArrayBuilder.kt
sfs: This is a special case for unsigned types that exists in the handling of varargs. Specifically, we can pass unsigned types as varargs through unwrapped arrays (e.g., `UIntArray` for `UInt` varargs). The array type in question is a wrapper around the underlying array of primitives (e.g., `IntArray`) and does not have a corresponding SpreadBuilder. In the example, we have to use `IntSpreadBuilder` directly and coerce the result to `UIntArray`.

We cannot implement this case in inline class lowering, since vararg lowering has to run early in the pipeline in order to implement the array copy optimizations (`testDoNotCopyImmediatelyCreatedArrays`) and inline class lowering does not actually change the IR types (inline class boxing and primitive type boxing are both implemented in the same way in `ExpressionCodegen`).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
sfs: Done!

compiler/testData/codegen/bytecodeText/inlineClasses/passInlineClassesWithSpreadOperatorToVarargs.kt
sfs: Thanks for the pointer, I didn't know about TEMPLATES. Fixed!

max-kammerer:
@sfs Many thanks!

PR #2511, opened 2019-08-06
JVM IR: Remove bridge construction from CallableReferenceLowering

sfs:
Currently, there is some code duplication between `CallableReferenceLowering`, `BridgeLowering`, and `FunctionNVarargInvokeLowering`. Specifically, `CallableReferenceLowering` builds a bridge method to `invoke`, which is either an ordinary bridge as generated in `BridgeLowering`, or a specialized vararg bridge as in `FunctionNVarargInvokeLowering`.

Furthermore, `FunctionNVarargInvokeLowering` implements some JVM IR specific behavior for functions with a large number of arguments, which is incompatible with the changes outlined in [KT-32698](https://youtrack.jetbrains.com/issue/KT-32698) as explained in [KT-32701](https://youtrack.jetbrains.com/issue/KT-32701). It also generates custom error messages, resulting in a lot of bytecode which is probably unnecessary.

This PR generalizes `FunctionNVarargInvokeLowering` (renamed to `FunctionNVarargBridgeLowering`) to handle the cases produced by `CallableReferenceLowering` and removes all bridge generation from `CallableReferenceLowering`.

This almost works perfectly, except that `BridgeLowering` currently doesn't remove the `overriddenSymbols` from the bridged method, resulting in unnecessary boxed return types. I've fixed this as much as it was relevant for `CallableReferenceLowering`, but the issue probably still persists in other cases.


compiler/testData/codegen/box/functions/bigArity/callFunViaVararg.kt
```diff
@@ -1,6 +1,7 @@
 // !LANGUAGE: +FunctionTypesWithBigArity
```
gogabr: What happens to this test?
Do we not want to support this kind of code any more?
If so, the test should probably just be removed.

sfs: That was my understanding based on [KT-32701](https://youtrack.jetbrains.com/issue/KT-32701). Instead of removing the test, I've changed it to have only a single large arity supertype.

Going forward we still want to support this case, at least after [KT-32698](https://youtrack.jetbrains.com/issue/KT-32698) is fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FunctionNVarargBridgeLowering.kt
```diff
@@ -0,0 +1,180 @@
...
+// adds a bridge method for such large arity functions, which checks the number of arguments
+// dynamically.
+private class FunctionNVarargBridgeLowering(val context: JvmBackendContext) :
+    FileLoweringPass, IrElementTransformerVoidWithContext() {
+    override fun lower(irFile: IrFile) = irFile.transformChildrenVoid(this)
+
+    // Change calls to big arity invoke functions to vararg calls.
+    override fun visitFunctionAccess(expression: IrFunctionAccessExpression): IrExpression {
+        if (expression.valueArgumentsCount < FunctionInvokeDescriptor.BIG_ARITY ||
+            !expression.symbol.owner.parentAsClass.defaultType.isFunctionOrKFunction()
```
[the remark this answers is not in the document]
sfs: Thanks, fixed!

gogabr:
Thank you!

PR #2512, opened 2019-08-07
KT-14946 Move statement up/down (with Ctrl+Shift+Up/Down) messes with empty lines

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14946


klunnii:
Accepted. Thank you!

PR #2513, opened 2019-08-07
Delete output when unable to run KAPT incrementally

gavra0:
When comparing previous and current classpath snapshots, and it is not
possible to compute the types that changed, make sure to delete the outputs
before invoking KAPT.

Test: KaptIncrementalWithAggregatingApt.testIncompatibleClasspathChanges


yanex:
Merged into master manually. Thanks!

PR #2514, opened 2019-08-07
Run KAPT non-incrementally if annotation processor classpath changes

gavra0:
Annotation processor classpath for a KAPT run is recorded, and the current
run is compared against the previous one. If those differ, KAPT should
run non-incrementally.

Test: KaptIncrementalWithIsolatingApt.testChangingAnnotationProcessorClasspath


yanex:
Merged into master manually. Thanks!

PR #2515, opened 2019-08-07
JVM IR: Convert interface companion fields to static fields.

sfs:
`MoveCompanionObjectFieldsLowering` replaces fields in a companion object with static fields in the enclosing class. Currently the lowering ignores fields in interface companions (unless they are marked with `@JvmField`). This PR adds support for fields in interface companions, which we do not want to move, but still have to mark as static.


max-kammerer:
@sfs  Thank you!

PR #2518, opened 2019-08-08
KT-33204 False positive "flatMap call could be simplified to flatten()" with Array

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33204

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallInspection.kt
@@ -11,8 +11,10 @@ import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.KtNodeTypes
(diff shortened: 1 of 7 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallInspection.kt
```diff
@@ -87,6 +89,8 @@ class SimplifiableCallInspection : AbstractKotlinInspection() {
                 val reference = lambdaExpression.singleStatement() ?: return null
                 val lambdaParameterName = lambdaExpression.singleLambdaParameterName() ?: return null
                 if (!reference.isNameReferenceTo(lambdaParameterName)) return null
+                val receiverType = callExpression.receiverType() ?: return null
+                if (KotlinBuiltIns.isArray(receiverType) || KotlinBuiltIns.isPrimitiveArray(receiverType)) return null
```
MaximPestryakov: Will it work on an array of arrays, for example:
```kotlin
arrayOf(
    arrayOf(1),
    arrayOf(2, 3),
    arrayOf(4, 5, 6)
).flatMap { it }
```

PR #2519, opened 2019-08-08
JVM IR: Fix field visibility and ACC_DEPRECATED flags

sfs:
This PR adds a lowering to set field visibility and adds more handling for `@Deprecated` annotations. 

Currently, psi2ir produces fields with the same visibility as the underlying property, which is usually correct except for a few special cases. I've implemented this as a separate lowering, since it's potentially useful to set the visibility in other lowerings, instead of relying on special cases in `ClassCodegen`. For example, `AssertionLowering` should produce package private fields.

Going forward, it might be a good idea to add this into psi2ir directly, but it's not clear that this is the desired behavior for the other backends.

For the second part, there's already some handling for `ACC_DEPRECATED` in `FunctionCodegen`, this PR just fills in more cases.

The two commits depend on each other to fix some tests (e.g., `testTopLevelProperty`), otherwise I would have split this into two PRs.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FieldVisibilityLowering.kt
```diff
@@ -0,0 +1,37 @@
...
+    name = "FieldVisibilityLowering",
+    description = "Set correct visibility flags for fields."
+)
+
+private class FieldVisibilityLowering(val context: JvmBackendContext) : ClassLoweringPass {
+    override fun lower(irClass: IrClass) {
+        for (irProperty in irClass.declarations.filterIsInstance<IrProperty>()) {
+            if (irProperty.origin == IrDeclarationOrigin.FAKE_OVERRIDE) continue
+            val field = irProperty.backingField ?: continue
+            if (irProperty.isConst || field.hasAnnotation(JVM_FIELD_ANNOTATION_FQ_NAME)) continue
```
udalov: I wonder if it makes more sense to implement most of the field visibility rules directly in psi2ir when creating the fields, so that all backends would observe the same IR structure and we'd have less trouble adapting common lowerings to be run on all backends. Namely, everything except the `@JvmField` logic here looks as if it should be performed by psi2ir. This lowering would still be needed, but only to change visibility of `@JvmField` fields from private to public. What do you think?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -559,3 +568,18 @@ fun getLabeledThisName(callableName: String, prefix: String, defaultName: String
         defaultName
     } else prefix + mangleNameIfNeeded(callableName)
 }
+
+val IrAnnotationContainer.deprecationFlags: Int
+    get() {
+        val annotation = annotations.findAnnotation(FQ_NAMES.deprecated) ?: return 0
+        val isHidden = (annotation.getValueArgument(2) as? IrGetEnumValue)?.symbol?.owner?.name?.asString() == "HIDDEN"
```
udalov: Maybe `DeprecationLevel.HIDDEN.name()` instead of `"HIDDEN"`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -559,3 +568,18 @@ fun getLabeledThisName(callableName: String, prefix: String, defaultName: String
...
+        return Opcodes.ACC_DEPRECATED or if (isHidden) Opcodes.ACC_SYNTHETIC else 0
+    }
+
+val IrFunction.deprecationFlags: Int
+    get() {
+        val isPropertyAnnotationStub = origin == JvmLoweredDeclarationOrigin.SYNTHETIC_METHOD_FOR_PROPERTY_ANNOTATIONS
```
udalov: Unfortunately this doesn't work for `$annotations` methods that are moved to `DefaultImpls` classes for properties declared in interfaces because their origin is rewritten to `DEFAULT_IMPLS`: https://github.com/JetBrains/kotlin/blob/0065236bdebefa72c06aa7b32fe78bec6eb65781/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt#L275

Consider extracting that check and the comment and reusing it here, until we find a solution to this problem

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FieldVisibilityLowering.kt
sfs: I've updated the PR to make this change in psi2ir. Instead of adding a separate lowering for backend specific visibility exceptions, I've added a new function to `GeneratorExtensions` to compute field visibilities. This is then overridden in the jvm backend to handle `@JvmFields` and fields overridden from Java and in the js backend to handle `@JsExport`.

The reason is that we don't generate fake overrides for private fields (this would break during the IR verification phase in the JS IR backend), so a lowering to modify field visibility after the fact would have to add fake overrides as well. It's simpler to start out with the correct visibilities coming from psi2ir.

Unfortunately this affects a lot of IrTextTests. I've updated the tests in a separate commit.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
sfs: That's irritating, but thanks for the pointer. I've added the same workaround for now.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt
```diff
@@ -171,5 +173,11 @@ class PropertyGenerator(declarationGenerator: DeclarationGenerator) : Declaratio
         val variableDescriptor = getOrFail(BindingContext.VARIABLE, ktProperty)
         return variableDescriptor as? PropertyDescriptor ?: TODO("not a property: $variableDescriptor")
     }
+
+    private val Visibility.admitsFakeOverride: Boolean
+        get() = this != Visibilities.PRIVATE && this != Visibilities.PRIVATE_TO_THIS && this != Visibilities.INVISIBLE_FAKE
```
udalov: Looks like this can be simplified to `!Visibilities.isPrivate(this)` and inlined because invisible fakes have been filtered out in the first line of `generateFakeOverrideProperty`?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/impl/IrFieldImpl.kt
```diff
@@ -53,10 +53,20 @@ class IrFieldImpl(
...
+    ) :
+            this(startOffset, endOffset, origin, symbol, type, symbol.descriptor.visibility)
+
+    constructor(
+        startOffset: Int,
+        endOffset: Int,
+        origin: IrDeclarationOrigin,
+        symbol: IrFieldSymbol,
+        type: IrType,
+        visibility: Visibility
```
udalov: I'd prefer to add a default parameter to an existing constructor

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt
```diff
@@ -285,6 +285,25 @@ open class SymbolTable : ReferenceSymbolTable {
 
     val unboundEnumEntries: Set<IrEnumEntrySymbol> get() = enumEntrySymbolTable.unboundSymbols
 
+    fun declareFieldWithVisibility(
+        startOffset: Int,
+        endOffset: Int,
+        origin: IrDeclarationOrigin,
+        descriptor: PropertyDescriptor,
+        type: IrType,
+        visibility: Visibility,
```
udalov: Similarly, would be nice to have a default value in `declareField` instead. Also consider specifying the actual argument value for clarity (when it's for example obviously private or public) in as many call sites as possible

compiler/testData/ir/irText/classes/enum.txt
```diff
@@ -319,9 +319,6 @@ FILE fqName:<root> fileName:/enum.kt
               public final fun <get-ordinal> (): kotlin.Int declared in <root>.TestEnum4
             $this: VALUE_PARAMETER name:<this> type:kotlin.Enum<<root>.TestEnum4>
         PROPERTY FAKE_OVERRIDE name:x visibility:public modality:FINAL [val]
-          FIELD FAKE_OVERRIDE name:x type:kotlin.Int visibility:public [final]
```
udalov: Is it expected that this fake field override was removed?

compiler/ir/serialization.js/src/org/jetbrains/kotlin/ir/backend/js/JsGeneratorExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package org.jetbrains.kotlin.ir.backend.js
+
+import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.Visibility
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi2ir.generators.GeneratorExtensions
+
+class JsGeneratorExtensions : GeneratorExtensions() {
+    override fun computeFieldVisibility(descriptor: PropertyDescriptor): Visibility =
```
udalov: @skuzmich Please take a look

udalov:
A couple of tests are failing in `writeFlags/property/syntheticAnnotationsMethod` after rebasing on my recent [changes](https://github.com/JetBrains/kotlin/commit/ea0142da60ede3163440db9f6fca1096604ec4e8):
* `privateProperty.kt`, `protectedProperty.kt`, `publicProperty.kt` -- I'll fix these soon with a separate commit
* `withGetterJvmName.kt` -- please unmute

We also have tests on incremental compilation with IR now, and one of them fails after these changes: `IrIncrementalJvmCompilerRunnerTestGenerated.WithJava.Other.testDefaultValueInConstructorAdded`. Please take a look, but if you don't succeed, you can simply add a `directives.txt` text file with `// IGNORE_BACKEND: JVM` and we'll look into it afterwards. Basically this test compiles the given project in several steps, where all `.kt` files represent the first step and `.new`/`.delete` files represent the next steps, and compares the resulting files with the clean build scenario.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt
sfs: That's what I wrote originally! However, if you have a lateinit field with a private setter then the field visibility in a subclass is "invisible fake". This is because we take the visibility of the setter as the visibility of the field.

So there's either going to be a special case in `computeFieldVisibility` or we add a utility function which sees invisible fakes as private. I would prefer the latter.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/impl/IrFieldImpl.kt
sfs: Done

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt
sfs: Done, except for specifying the argument value in the other call sites, since it's actually unused.

This function is strange to begin with: Most arguments are only used in the default value for `fieldFactory`. However, all but two call-sites pass a different value for `fieldFactory`. This applies to pretty much all `declare*` functions in this file.

A better interface would be to have a `declare*` function which just takes the descriptor and factory and a convenience method which produces the default factory from some other arguments with reasonable defaults.

I'd really like to refactor this code, just not in this PR.

compiler/testData/ir/irText/classes/enum.txt
sfs: Yes, the field we're overriding here is now private and as far as I can tell should not generate fake overrides. That's one of the properties that [`CheckIrElementVisitor`](https://github.com/JetBrains/kotlin/blob/e8ac22e238f3b108a11215292f92e39726691964/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/CheckIrElementVisitor.kt#L307) tests for. Do you agree?

sfs:
I'm currently updating the PR after rebasing on master. I didn't see the incremental compilation test before - I'll have a look now.

sfs:
The incremental compilation test failed because we did not generate annotations for synthetic methods. Since we now mark hidden deprecated methods as synthetic, this lead to a missing deprecated annotation. Call resolution ignores methods with `@Deprecated` annotations with level `HIDDEN`, leading to a mismatch between full compilation (which sees the annotation in the IR) and incremental compilation (which sees no annotations in the deserialized declarations).

The logic for annotations in `FunctionCodegen` is fragile. Ideally, we'd take care of this in the individual lowerings instead, but right now this fails because of dependencies between lowerings. Concretely, we need to keep `@JvmStatic` annotations on default argument stubs, since we cannot move `jvmStaticAnnotationPhase` before `defaultArgumentStubPhase` without it leading to more problems.

For now I've added a workaround. Or more accurately, changed the existing workarounds in `FunctionCodegen` to handle this additional case.

compiler/ir/serialization.js/src/org/jetbrains/kotlin/ir/backend/js/JsGeneratorExtensions.kt
skuzmich: JS IR backend can export properties but not fields. All fields should be private.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/GeneratorExtensions.kt
```diff
@@ -32,4 +30,11 @@ open class GeneratorExtensions {
 
         companion object Instance : SamConversion()
     }
+
+    open fun computeFieldVisibility(descriptor: PropertyDescriptor): Visibility =
+        when {
```
udalov: Suggestion to move this logic inside the `PropertyDescriptor.fieldVisibility` in `PropertyGenerator` and return null in this method, to have only truly platform-specific code in `GeneratorExtensions` implementations

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt
udalov: Thanks! I agree. I'd still prefer replacing two checks with `Visibilities.isPrivate`, but I don't insist

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt
udalov: Thanks! I agree that this needs refactoring and can be postponed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -51,17 +50,22 @@ open class FunctionCodegen(
         var methodVisitor = createMethod(flags, signature)
 
         val hasSyntheticFlag = flags.and(Opcodes.ACC_SYNTHETIC) != 0
-        generateParameterNames(irFunction, methodVisitor, signature, state, hasSyntheticFlag)
+        if (state.generateParametersMetadata && !hasSyntheticFlag) {
+            generateParameterNames(irFunction, methodVisitor, signature, state)
+        }
+
+        if (irFunction.origin != IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER) {
```
udalov: I wonder if we should use the same condition to generate parameter annotations (below)? I really hope that there's no case in the current backend where we generate method annotations, but not parameter annotations, or vice versa. (Note BTW that `$default` method was exactly this case before 1.3.40 where we deliberately decided not to generate any annotations on it: [KT-30289](https://youtrack.jetbrains.com/issue/KT-30289).)

compiler/testData/ir/irText/classes/enum.txt
udalov: Yes, absolutely. Thanks!

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt
sfs: Done

compiler/ir/serialization.js/src/org/jetbrains/kotlin/ir/backend/js/JsGeneratorExtensions.kt
sfs: That's refreshingly simple, thanks! But can you take a look at `testEval` and `testTopLevelCallables`? Both access a top level exported field through reflection and fail if this field is not public. As a compromise, the code currently still marks `@JsExport`ed fields as public. Is it necessary to support access to fields through reflection like this? If not, marking every field as private would work just fine.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/GeneratorExtensions.kt
sfs: Done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
sfs: There shouldn't be, but there is one more special case that we need to handle: When we regenerate anonymous objects in the inliner, the constructor should not have parameter annotations, since we can end up with fewer parameters after inlining (see, e.g., the lambda class corresponding to `{ s() }()` in `testKt17091`).

I suppose we should fix this in the inliner instead, but the code is sufficiently complex that I want to think a bit more about it before making any changes.

udalov: Right, anonymous object copy should take a different code path since it's being done via bytecode. I don't mind postponing this change though. But it'd be nice to update the TODO below if possible, just to prevent someone else spending time researching why specific (and different) origins are handled here and below

compiler/ir/serialization.js/src/org/jetbrains/kotlin/ir/backend/js/JsGeneratorExtensions.kt
skuzmich: Tests fail due to bug in JS code generator. They supposed to access top level properties but instead, by mistake, fields are exported with the same name. I'll change `JsGeneratorExtensions#computeFieldVisibility` to return `Visibilities.PRIVATE` once its fixed. Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
sfs: Sure, I've added a comment to that effect.

udalov:
@sfs Thank you! I've squashed the 1st and 3rd commits (since it's usually a bit easier to navigate history when tracking changes in specific tests, seeing also the original change) and pushed to master in a90ac2438d and fd2ad89799

sfs:
@udalov Thanks! I only separated the commits for easier review. It's definitely more logical like this.

PR #2520, opened 2019-08-08
KT-22412 J2K: Intention to replace if(...) throw IAE with require

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22412


darthorimar:
LGTM, thank you! :smiley_cat:

PR #2521, opened 2019-08-08
Annotate object INSTANCE field as @NotNull in bytecode

JakeWharton:
[KT-33226](https://youtrack.jetbrains.com/issue/KT-33226)

I believe a change needs to be made to the light classes, but I'm not sure where that needs to take place. Perhaps `ByJvmSignatureIndexer`? But it's not clear to me how to denote annotations there.

### DIFF
--- a/compiler/testData/asJava/lightClasses/object/SimpleObject.java
+++ b/compiler/testData/asJava/lightClasses/object/SimpleObject.java
@@ -5,6 +5,7 @@ public final class A {
     public static final int cc = 1;
     @org.jetbrains.annotations.NotNull
     public static final java.lang.String cv = "A";
+    @org.jetbrains.annotations.NotNull
     public static final pack.A INSTANCE;
 
     public final int getC() { /* compiled code */ }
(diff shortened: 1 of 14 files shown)


compiler/testData/asJava/lightClasses/object/SimpleObject.java
```diff
@@ -5,6 +5,7 @@
     public static final int cc = 1;
     @org.jetbrains.annotations.NotNull
     public static final java.lang.String cv = "A";
+    @org.jetbrains.annotations.NotNull
```
JakeWharton: ...as a result, this test currently fails.

udalov:
Hi @JakeWharton! Sorry for the delay.

To support this in light classes, please take a look at `KtLightNullabilityAnnotation` and how it's used. I think you need to add a NotNull annotation instance to `KtLightModifierList.computeAnnotations` for this annotation to be visible on the light field.

We also have a new faster implementation of light classes (called appropriately _ultra_ light classes) which is going to replace the default one soon. Please make sure that tests in `UltraLightClassLoadingTestGenerated` also pass after change above, or otherwise there might be some changes required around `KtUltraLightElementWithNullabilityAnnotation` to support it.

JakeWharton:
I'm afraid I won't have time to look at this for a bit. If it's _really_ easy to fix up feel free to take it over. Otherwise I'll try to circle back in a week or two.

Thanks for the pointers regardless!

dnpetrov:
NB this PR actually generates @NotNull annotations on static INSTANCE fields for lambdas and callable references (but not for objects and companion objects). I'm doing similar changes now, and will incorporate some of the code. PR, however, is not quite relevant at this point.

PR #2522, opened 2019-08-09
JVM IR: Don't move non-public const fields out of interface companions

sfs:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -31,6 +31,7 @@ import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 import org.jetbrains.kotlin.load.java.JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME
+import org.jetbrains.kotlin.synthetic.isVisibleOutside
```
max-kammerer: Redundant import

sfs: Thanks! Removed.

max-kammerer:
Thanks!

PR #2523, opened 2019-08-09
JVM IR: Handle special case visibility rules for callable references

sfs:
A class for a callable reference is package private by default and public when it is used inside of an inline function.

Since we are moving functions around during compilation, we have to record this information early in the compilation pipeline. Fortunately, `InventNamesForLocalClasses` already provides infrastructure to record information about local classes and we can reuse this to mark classes which were defined in the body of an inline function.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
@@ -9,6 +9,7 @@ import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.backend.common.ir.ir2string
 import org.jetbrains.kotlin.backend.common.lower.allOverridden
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns.FQ_NAMES
 import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
@@ -135,15 +136,15 @@ private val visibilityToAccessFlag = mapOf(
     JavaVisibilities.PACKAGE_VISIBILITY to AsmUtil.NO_FLAG_PACKAGE_PRIVATE
 )
 
-private fun IrDeclaration.getVisibilityAccessFlagForAnonymous(): Int =
-    if (isInlineOrContainedInInline(parent as? IrDeclaration)) Opcodes.ACC_PUBLIC else AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+private fun IrClass.getVisibilityAccessFlagForAnonymous(context: JvmBackendContext): Int =
+    if (context.getLocalClassInfo(this)?.inInlineScope == true) Opcodes.ACC_PUBLIC else AsmUtil.NO_FLAG_PACKAGE_PRIVATE
 
 fun IrClass.calculateInnerClassAccessFlags(context: JvmBackendContext): Int {
     val isLambda = superTypes.any { it.safeAs<IrSimpleType>()?.classifier === context.ir.symbols.lambdaClass }
     val visibility = when {
-        isLambda -> getVisibilityAccessFlagForAnonymous()
+        isLambda -> getVisibilityAccessFlagForAnonymous(context)
         visibility === Visibilities.LOCAL -> Opcodes.ACC_PUBLIC
-        else -> getVisibilityAccessFlag()
+        else -> getVisibilityAccessFlag(context)
     }
     return visibility or
             if (origin.isSynthetic) Opcodes.ACC_SYNTHETIC else 0 or
@@ -167,13 +168,13 @@ private fun IrClass.innerAccessFlagsForModalityAndKind(): Int {
     return 0
 }
 
-private fun IrDeclarationWithVisibility.getVisibilityAccessFlag(kind: OwnerKind? = null): Int =
-    specialCaseVisibility(kind)
+fun IrDeclarationWithVisibility.getVisibilityAccessFlag(context: JvmBackendContext, kind: OwnerKind? = null): Int =
+    specialCaseVisibility(context, kind)
         ?: visibilityToAccessFlag[visibility]
         ?: throw IllegalStateException("$visibility is not a valid visibility in backend for ${ir2string(this)}")
 
 
-private fun IrDeclarationWithVisibility.specialCaseVisibility(kind: OwnerKind?): Int? {
+private fun IrDeclarationWithVisibility.specialCaseVisibility(context: JvmBackendContext, kind: OwnerKind?): Int? {
 //    if (JvmCodegenUtil.isNonIntrinsicPrivateCompanionObjectInInterface(memberDescriptor)) {
 //        return ACC_PUBLIC
 //    }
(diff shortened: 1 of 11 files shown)


max-kammerer:
NB: actually current backend do more complex thing: for CR used in inline it generates public class with a bit different name and private class if CR is used outside inline. So it could be two different classes for same CR inside one file. But it's not required to support such behaviour now, this logic would be revised and maybe changed near 1.4

sfs:
@max-kammerer I was wondering about this myself with respect to the IR backend. Can you give an example of the behavior in the current backend?

From what I've seen, I thought that the current backend always creates a new class for callable references. For lambdas we cannot do better because of debugging information. In contrast to that, the IR backend caches and reuses property references and SAM conversions (although I don't see how to get SAM conversion to trigger on a function reference). The flags for the latter would currently be broken, e.g., for a property reference used in both an inline function and a non-inline function we might not generate a public class.

It's on my todo list anyway, so if you have any more insights to share about the behavior in the current backend, or indeed the intended behavior going forward, it would be much appreciated. :)

max-kammerer:
@sfs Oh, yes, sorry for confusion: there was a clash with SAM wrappers in my head when I wrote about current backend approach. It's required cause of class cashing (that is also performed sometimes in IR as you mentioned):
```
public class Foo {
    public void run (Runnable r) { }
}

fun main() {
    val z: () -> Unit = { 123  } 
    Foo().run(z) //one cached wrapper with package visibility for non-inline  and...
}

inline fun foo() {
    val z: () -> Unit = {
        123
    }
    Foo().run(z)  //...another cached wrapper for inline with public visibility
}
```


> for a property reference used in both an inline function and a non-inline function we might not generate a public class

Actually this could cause problem with current inliner logic cause it's attempt to avoid additional class creation/regeneration (but we think to revisit this logic cause of unclear semantic). So if `foo` would be inlined to another package without regeneration it's implies some restriction to class  and it constructor visibilities.
If inliner would perform unconditional class regeneration then it would be possible to avoid additional public class generation.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -418,23 +411,20 @@ internal fun getSignature(
...
-    /* Original had a check for SyntheticClassDescriptorForJava, never invoked in th IR backend. */
-    if (isOptionalAnnotationClass()) {
-        return AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+fun IrClass.getVisibilityAccessFlagForClass(context: JvmBackendContext): Int =
+    when {
+        isOptionalAnnotationClass() -> AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+        kind === ClassKind.ENUM_ENTRY -> AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+        visibility === Visibilities.PRIVATE -> AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+        isClassForCallableReference -> getVisibilityAccessFlagForAnonymous(context)
+        else -> Opcodes.ACC_PUBLIC
```
udalov: I'd prefer to list all supported visibilities explicitly (`PUBLIC`, `PROTECTED`, `LOCAL`, `INTERNAL`), keeping the TODO about local visibility and reflection, and throw ISE otherwise if the visibility is unknown. This will allow this code to fail faster in case we use some new visibility (such as `JavaVisibilities.PACKAGE_VISIBILITY`) in some new lowering

sfs: Done! And a good thing, too, since it turns out that there are cases where we have classes with package visibility...

udalov:
To elaborate on this point raised by @max-kammerer above,

> attempt to avoid additional class creation/regeneration (but we think to revisit this logic cause of unclear semantic)

The issue here is that semantics of non-inlined lambdas and callable references (and also anonymous objects and local classes) _inside inline functions_ are unclear even in the today's stable JVM backend, not mentioning other backends and the IR. Here's the issue where we'll track the progress: https://youtrack.jetbrains.com/issue/KT-28064.

(It's not directly related to these changes, just a window into a long-standing problem where the solution might or might not affect this code in the backend.)

sfs:
@max-kammerer Thanks for the explanation. I've looked into the logic for sam wrappers and implemented it in #2531 except for the special handling of classes used inside of inline functions. I think it's probably best to fix this part in a separate PR, same as for property references.

Unconditional class regeneration seems like it might produce significantly more code, even if it has clearer semantics with regards to object identity. For now I'd like to try and match the behavior of the old backend as closely as possible.

sfs:
Rebased on master

@max-kammerer @udalov Should I move forward with this PR, or should I fix all the remaining cases (cached property references, SAM wrappers for inline functions, etc.) first? Or are there already talks about changing the semantics of inlining which will make these changes superfluous?

udalov:
@sfs Thanks, we'll take a look soon. Sorry for this delay.

> Or are there already talks about changing the semantics of inlining which will make these changes superfluous?

Not yet

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -408,23 +401,26 @@ internal fun getSignature(
...
+        visibility === Visibilities.LOCAL -> Opcodes.ACC_PUBLIC
+        visibility === Visibilities.INTERNAL -> Opcodes.ACC_PUBLIC
+        visibility === JavaVisibilities.PACKAGE_VISIBILITY -> AsmUtil.NO_FLAG_PACKAGE_PRIVATE
+        else -> error("Unexpected visibility: $visibility")
+    }
+
+private val IrClass.isClassForCallableReference: Boolean
+    get() = origin == JvmLoweredDeclarationOrigin.LAMBDA_IMPL ||
+            origin == JvmLoweredDeclarationOrigin.FUNCTION_REFERENCE_IMPL ||
+            origin == JvmLoweredDeclarationOrigin.GENERATED_SAM_IMPLEMENTATION ||
```
udalov: Looks like we already generate `GENERATED_SAM_IMPLEMENTATION` and `GENERATED_PROPERTY_REFERENCE` classes as `PUBLIC`, so these two lines are redundant

udalov:
I wonder if generating function reference classes with the right visibility (`PUBLIC` if inside inline) instead of always `LOCAL` could be done in the lowering instead of retroactively trying to deduce this information in the codegen. Basically, set visibility to `PUBLIC` if there are inline functions among our parents in `CallableReferenceLowering.FunctionReferenceBuilder.functionReferenceClass`. This would seem to work because this would be computed before `localDeclarationsPhase` where we move declarations around, and it wouldn't require storing the `inInlineScope` flag and passing the context around in visibility flag computation in codegen. What do you think?

sfs:
This is actually what I tried at first. I only changed it since `localDeclarationsPhase` only handles classes with `LOCAL` visibility and changing this was a bit delicate. But yes, it would be cleaner. I'll look into it again.

sfs:
It's possible to produce the correct visibility directly for FunctionReferences. The trade-off is a more complicated check for local classes in `LocalDeclarationsLowering`.

The main problem is that we need the same logic for PropertyReferences and (the constructors of) anonymous objects. This makes it seem like we should have a separate lowering for this. Ideally we would run the lowering after `LocalDeclarationsLowering` so that we could consistently use `LOCAL` visibility beforehand and fill in the correct visibility afterwards. However, this won't work, since `LocalDeclarationsLowering` moves functions around and we forget whether we are in the scope of an inline function. 

I'm not sure which approach would be simpler here. Generally speaking I want to move special cases out of codegen, but in this case we would have essentially the same code, *and* slightly more complicated code in `LocalDeclarationsLowering`. Which one would you prefer?

udalov:
Thanks for the investigation!

If I understand everything correctly, I still kind of prefer to compute visibility correctly from the start, even at the cost of complicating `LocalDeclarationsLowering` with another platform-specific check. The main reason is that it doesn't need another dependency of codegen on mutable state stored in `JvmBackendContext`. My other goal is to keep `InventNamesForLocalClasses` as simple as possible, which includes not doing stuff it's not directly responsible for (which is computing JVM names for everything). Also, it's nice that subsequent lowerings will see the correct visibility for the newly generated declarations and can make decisions upon that. Let's try the new approach and discuss if it ends up more complicated than I imagine?

sfs:
@udalov I've reimplemented this PR in #2558 to avoid changing `InventNamesForLocalClasses` and to move the actual logic into `LocalDeclarationsLowering`. Turns out that it just needed some fixes in other lowerings.

udalov:
Closing in favor of #2558.

PR #2525, opened 2019-08-09
KT-20357Added flatMap() sample

Dmitry-Borodin:

### DIFF
--- a/libraries/tools/kotlin-stdlib-gen/src/templates/Mapping.kt
+++ b/libraries/tools/kotlin-stdlib-gen/src/templates/Mapping.kt
@@ -314,6 +314,7 @@ object Mapping : TemplateGroupBase() {
         specialFor(ArraysOfUnsigned) { inlineOnly() }
 
         doc { "Appends all elements yielded from results of [transform] function being invoked on each ${f.element} of original ${f.collection}, to the given [destination]." }
+        sample("samples.collections.Collections.Transformations.flatMap")
         specialFor(Sequences) {
             signature("flatMapTo(destination: C, transform: (T) -> Sequence<R>)")
         }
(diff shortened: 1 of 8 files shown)


libraries/tools/kotlin-stdlib-gen/src/templates/Mapping.kt
```diff
@@ -314,6 +314,7 @@ object Mapping : TemplateGroupBase() {
         specialFor(ArraysOfUnsigned) { inlineOnly() }
 
         doc { "Appends all elements yielded from results of [transform] function being invoked on each ${f.element} of original ${f.collection}, to the given [destination]." }
+        sample("samples.collections.Collections.Transformations.flatMap")
```
ilya-g: The sample is for `flatMap` function, but it's placed into `flatMap`'s function template. 
But no worries: I'll fix it before merging.

ilya-g:
Thanks for the sample, merged it manually.

If you're interested in further improvement of this sample, you can provide more a specialized  sample for `String.flatMap` and `Map.flatMap` overloads.

Dmitry-Borodin:
@ilya-g I checked and now this sample presented only in generated folder. How can we keep it so net time samples will be generated this comment will still be generated?
I thought that is the reason to have it in templates.

ilya-g:
Sorry, I don't get the question. The sample reference is put into a template, and then that template is used to generate a lot of stdlib functions. If the question is how to provide different sample references for different overloads generated from a single template, then you can take a look here for an example of how to do it: https://github.com/JetBrains/kotlin/blob/f334ad2ffc26be755d9752da9993af2f40292143/libraries/tools/kotlin-stdlib-gen/src/templates/Snapshots.kt#L208-L212

PR #2526, opened 2019-08-09
KT-28343 J2K: Automatic conversion from @lombok.Data to data class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28343


darthorimar:
@t-kameyama thank you for your PR :) 

According to official documentation (https://projectlombok.org/features/all), even Lombok's data and value classes have a lot of features. Many of it depends on Lombok config file of the project. Support all of it may be really routine and possibly error-prone. So why not to use Lombok plugin for IntelliJ IDEA for converting Lombok classes to Kotlin?

As I can see, Lombok plugin generates some fake PSI methods to the classes bodies which we can use in J2K to generate corresponding declarations. In post-processing getters and setters pairs will be joined together as properties and moved to the primary constructor. Also, we may have some additional post-processing/conversion to detect if `toString`/`hashCode`/`equals` have default implementations and convert the class to the data one :)

darthorimar:
Closing for now as inactive. Feel free to create a new one :)

PR #2527, opened 2019-08-10
[MD] Fix typos and improve grammar in stdlib readmes

Inego:


libraries/stdlib/samples/ReadMe.md
```diff
@@ -2,8 +2,8 @@
 
 This project contains samples for the standard library functions. 
 They are located in the [test](test) source root and each sample is written like a small unit test.
-Then these samples are referenced in the documentation of the standard library functions 
-with the [`@sample`](https://kotlinlang.org/docs/reference/kotlin-doc.html#block-tags) tag and are embedded in the generated documentation as runnable samples.
+These samples can be referenced in the documentation of the standard library functions
```
ilya-g: IMO, this somewhat changes the meaning: it's not that samples can (optionally) be referenced from library functions, but they are/must be referenced, otherwise it makes sense to write them there.

Inego: You're right. I'll rewrite this. Thanks for the clarification!

ilya-g:
Thanks for the fixes.

PR #2528, opened 2019-08-13
KT-21504 J2K: Convert Long.parseLong(s) to s.toLong()

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21504


darthorimar:
@t-kameyama Thank you!

PR #2529, opened 2019-08-13
Improve reliability of simulator test runner

benasher44:
With a recent Xcode (10.3) or macOS (10.14.6) upgrade, I've been consistently running into issues getting the test runner to work (example: https://dev.azure.com/autodeskoss/coroutineworker/_build/results?buildId=24&view=logs). Switching to standalone mode, which I believe has something to do with hinting to simctl that it doesn't need to fully boot a simulator (i.e. just execute the binary), makes multiplatform/native tests run more reliably in these scenarios.


benasher44:
Confirmed that this avoids booting the sim, among other things. It could cause some services not to work, but it's unclear what those are and why you'd need them in unit tests (likely depends on platform-specific APIs you're using).

benasher44:
this is fixed in Kotlin 1.3.70 with the new built-in sim test task

PR #2531, opened 2019-08-15
JVM IR: Use correct naming scheme for SAM wrappers

sfs:
There's two parts to this PR:
- SAM conversion applied to callable references or lambdas does not generate a wrapper, but instead generates a callable reference class.
- SAM wrappers are cached based on the file class (if present) or the enclosing top-level class and in particular are always local to a top-level class.
- This means that it's actually easier to generate correct class names for SAM wrappers in the corresponding lowering than in `InventNamesForLocalClasses`.

There's one more corner case, involving SAM wrappers used in the scope of an inline function, which is not handled by this PR (mostly because it depends on #2523).

Finally, I've added a couple of additional tests, which check that a bound receiver in a callable reference is only evaluated once.

Additionally the code now handles SAM conversion of references to functions with default arguments or varargs. This is currently rejected by the frontend, because of the way type inference for SAM arguments works, but it ends up reducing code duplication here.


sfs:
Rebased on master

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -282,16 +309,15 @@ internal class CallableReferenceLowering(private val context: JvmBackendContext)
                             //     f(::h) == g(::h)
                             //
                             parameter.isVararg && unboundIndex < argumentTypes.size && parameter.type == valueParameters[unboundIndex].type ->
-                                irArray(parameter.type) { addSpread(irGet(valueParameters[unboundIndex++])) }
+                                irGet(valueParameters[unboundIndex++])
```
max-kammerer: Why spread is removed?

sfs: Because the current backend fails to copy the array here. :(

I also think that this is not the behavior we want, but technically it would be a breaking change and needs to go through the language committee. I'll file an issue about it later.

max-kammerer: Than 'In all other cases, excess arguments are packed into a new array' comment is obsolete

sfs: This part of the comment refers to the next case. I've changed it to be less confusing.

max-kammerer:
Many thanks!

PR #2534, opened 2019-08-19
JVM_IR: lower calls to @JvmStatic functions from other files.

pyos:
Note: this currently results in invalid IR (but valid bytecode) if the @JvmStatic function is imported, because its IR representation is unlowered and therefore has a dispatch receiver, but the call will not.


gogabr:
Thank you!

PR #2535, opened 2019-08-21
Psi2Ir: Mark accessors with no bodies as default accessors.

madsager:
They are currently marked as defined even when they get a default implementation. That makes it hard to figure out if the accessor should be removed when introducing a backing field in the JVM_IR backend.

The only actual change here is to psi2ir/generators/FunctionGenerator.kt. The rest are test changes to make sure that we run all the non-lateinit property annotation tests on the JVM_IR backend. The JVM_IR backend has different behavior from the current backend (on purpose) for lateinit properties and therefore the test has to be split to make sure we do not regress on the parts that work the same.

### DIFF
--- a/compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
+++ b/compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
@@ -1,0 +1,31 @@
+// WITH_REFLECT
+// IGNORE_BACKEND: JVM_IR
+// TARGET_BACKEND: JVM
+
+// Please make sure that this test is consistent with the diagnostic test "annotationsTargetingLateinitAccessor.kt"
+
+import kotlin.reflect.KAnnotatedElement
+
+annotation class Ann
+
+fun check(element: KAnnotatedElement, annotationExists: Boolean) {
+    require(element.annotations.isNotEmpty() == annotationExists) { "Fail: $element" }
+}
+
+class LateinitProperties {
+    @get:Ann
+    lateinit var x0: String
+
+    @get:Ann
+    private lateinit var x1: String
+
+    fun test() {
+        check(::x0.getter, annotationExists = true)
+        check(::x1.getter, annotationExists = false)
+    }
+}
+
+fun box(): String {
+    LateinitProperties().test()
+    return "OK"
+}
--- a/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/FunctionGenerator.kt
+++ b/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/FunctionGenerator.kt
@@ -112,7 +112,7 @@ class FunctionGenerator(declarationGenerator: DeclarationGenerator) : Declaratio
         declareSimpleFunctionInner(
             descriptor,
             ktAccessor ?: ktProperty,
-            if (ktAccessor != null) IrDeclarationOrigin.DEFINED else IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR
+            if (ktAccessor != null && ktAccessor.hasBody()) IrDeclarationOrigin.DEFINED else IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR
         ).buildWithScope { irAccessor ->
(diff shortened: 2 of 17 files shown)


madsager:
@dnpetrov @udalov @max-kammerer Could you have a quick look and see if this makes sense to you? :)

compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
```diff
@@ -0,0 +1,33 @@
...
+// IGNORE_BACKEND: JVM_IR
+// TARGET_BACKEND: JVM
+
+// Please make sure that this test is consistent with the diagnostic test "annotationsTargetingLateinitAccessor.kt"
+
+import kotlin.reflect.KAnnotatedElement
+import kotlin.reflect.KProperty
+
+annotation class Ann
+annotation class AnnRepeat
```
udalov: Looks like this annotation is not used here and in the corresponding diagnostic test, let's remove it

compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
```diff
@@ -0,0 +1,33 @@
+// WITH_REFLECT
+// IGNORE_BACKEND: JVM_IR
+// TARGET_BACKEND: JVM
+
+// Please make sure that this test is consistent with the diagnostic test "annotationsTargetingLateinitAccessor.kt"
+
+import kotlin.reflect.KAnnotatedElement
+import kotlin.reflect.KProperty
```
udalov: This import is unused

udalov:
Looks good to me from the JVM backend perpective, since we use [similar logic](https://github.com/JetBrains/kotlin/blob/d59f2bcc800f537657b56ac9a5d61721b1c489e1/compiler/backend/src/org/jetbrains/kotlin/codegen/PropertyCodegen.java#L148) to determine whether we need to generate accessor in the current backend.

I wonder if it's correct to consider this a part of the Kotlin language and not just JVM-specific behavior though. Since Dmitry is on vacation, @SvyatoslavScherbina @skuzmich do you have any objections from Native and JS side respectively?

compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
madsager: Good catch, done.

compiler/testData/codegen/box/annotations/annotationsOnLateinitAccessors.kt
madsager: Thanks, removed.

SvyatoslavScherbina:
> do you have any objections from Native and JS side respectively?

I don't have any objections from Native side.

Could you also clarify
> I wonder if it's correct to consider this a part of the Kotlin language and not just JVM-specific behavior though. 

what is "this"? How `origin` here affects lowerings/codegen?

udalov:
> what is "this"? How `origin` here affects lowerings/codegen?

I'm talking about the criteria that we use to determine that a property accessor is "default" or not. For example, on JVM the generated bytecode for
```
private val x = 42
```
and for
```
private val x = 42
    get
```
will be the same (no property accessor in the bytecode) precisely because we chose to use the following logic: generate an accessor only if it _has body_ (as opposed to _is declared in source code_). In the following case there will be an accessor generated, because the property accessor is not "default":

```
private val x = 42
    get() = field
```

In the IR, this "default"-ness of property accessors is stored in `origin`. If other backends use `origin` to determine if they need to generate a property accessor in the resulting binary (whatever it might mean), then behavior after this change will differ (less property accessors will be generated).

My question is then merely if it's OK to extend this, seemingly JVM-specific, criteria to the common IR. It seems as though it shouldn't matter much since it's only relevant for non-public-API properties (we always need to generate accessors for public/protected/internal properties), which are not accessible from other files, so I'm just hoping to confirm that this is the case. But on JVM for example it matters even for private properties because it affects the size of the bytecode, and how the class is visible via reflection.

SvyatoslavScherbina:
Thanks for the detailed clarification!

> My question is then merely if it's OK to extend this, seemingly JVM-specific, criteria to the common IR.

Yes, I suppose so.

skuzmich:
> do you have any objections from Native and JS side respectively?

No objections.

udalov:
@madsager Thanks!

PR #2536, opened 2019-08-22
Disable endorsed libraries in plugin

LepilkinaElena:


ilya-g:
See also the PR where this key is introduced in kotlin/native compiler: https://github.com/JetBrains/kotlin-native/pull/3215

PR #2538, opened 2019-08-23
Add new native targets.

olonho:


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/cocoapods/KotlinCocoapodsPlugin.kt
```diff
@@ -134,7 +134,8 @@ open class KotlinCocoapodsPlugin: Plugin<Project> {
 
         if (requestedTargetName == KOTLIN_TARGET_FOR_DEVICE) {
             // We create a fat framework only for device platforms: iosArm64 and iosArm32.
-            val devicePlatforms = listOf(KonanTarget.IOS_ARM64, KonanTarget.IOS_ARM32)
+            val devicePlatforms = listOf(KonanTarget.IOS_ARM64, KonanTarget.IOS_ARM32,
```
ilmat192: The comment above also needs to be updated.

ilmat192:
We also need to provide Gradle DSL shortcuts for these targets.

PR #2539, opened 2019-08-23
JVM, JVM IR: Specialize equality for inline classes to avoid boxing

sfs:
Currently, when we compare values of an inline class type we always box both arguments and call `Intrinsics.areEqual`. This also applies to the `Result` type which is used in coroutines. There is a test (`testResultApiDoesntBox`) which claims to check that this is not the case, but unfortunately the test is currently broken.

This PR implements support for the `equals-impl0` method which is present in inline classes and uses it to avoid boxing when comparing values of an inline class type. Additionally, there are many cases where comparing a value of an inline class type to `null` can use `IFNULL` and related instructions, instead of falling back to `Intrinsics.areEqual`. This fixes [KT-30785](https://youtrack.jetbrains.com/issue/KT-30785), and some of the changes will also be necessary to support [KT-24874](https://youtrack.jetbrains.com/issue/KT-24874).

Concretely, the PR only specializes calls of the form `x == y` where `x` is an unboxed value of inline class type. If `y` is similarly unboxed, we use the `equals-impl0` method, otherwise we use `equals-impl`. There is one case where we have to add additional null-checks at the moment, namely when `x` is a nullable value of an inline class wrapper around a reference type. Strictly speaking, this is not necessary right now, since the `equals-impl0` method is generated by the compiler and handles the case where `x` is `null`, but it might cause problems in the future if we want to permit user defined equality checks.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ErasedInlineClassBodyCodegen.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ErasedInlineClassBodyCodegen.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.kotlin.codegen
 
+import org.jetbrains.kotlin.codegen.AsmUtil.genTotalOrderEqualsForExpressionOnStack
 import org.jetbrains.kotlin.codegen.context.ClassContext
 import org.jetbrains.kotlin.codegen.state.GenerationState
 import org.jetbrains.kotlin.codegen.state.KotlinTypeMapper
@@ -17,6 +18,7 @@ import org.jetbrains.kotlin.resolve.jvm.AsmTypes
 import org.jetbrains.kotlin.resolve.jvm.diagnostics.Synthetic
 import org.jetbrains.kotlin.resolve.jvm.jvmSignature.JvmMethodGenericSignature
 import org.jetbrains.kotlin.resolve.jvm.jvmSignature.JvmMethodSignature
+import org.jetbrains.org.objectweb.asm.Type
 
 class ErasedInlineClassBodyCodegen(
     aClass: KtClass,
@@ -109,10 +111,16 @@ class ErasedInlineClassBodyCodegen(
 
 
                 override fun doGenerateBody(codegen: ExpressionCodegen, signature: JvmMethodSignature) {
-                    val iv = codegen.v
-                    iv.aconst(null)
-                    iv.athrow()
+                    val firstIndex = codegen.frameMap.getIndex(specializedEqualsDescriptor.valueParameters[0])
+                    val secondIndex = codegen.frameMap.getIndex(specializedEqualsDescriptor.valueParameters[1])
+                    val asmType = signature.valueParameters[0].asmType
+                    val left = StackValue.local(firstIndex, asmType)
+                    val right = StackValue.local(secondIndex, asmType)
+                    genTotalOrderEqualsForExpressionOnStack(left, right, asmType).put(Type.BOOLEAN_TYPE, codegen.v)
+                    codegen.v.areturn(Type.BOOLEAN_TYPE)
                 }
+
+                override fun skipNotNullAssertionsForParameters(): Boolean = true
             }
         )
     }
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
@@ -3390,15 +3390,10 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
         Type leftType = pregeneratedSubject != null ? pregeneratedSubject.type : expressionType(left);
         Type rightType = expressionType(right);
 
-        boolean leftIsInlineClassWithPrimitiveEquality = leftIsInlineClass && isInlineClassTypeWithPrimitiveEquality(leftKotlinType);
-        boolean rightIsInlineClassWithPrimitiveEquality = rightIsInlineClass && isInlineClassTypeWithPrimitiveEquality(rightKotlinType);
-
-        boolean leftHasPrimitiveEquality = isPrimitive(leftType) && (!leftIsInlineClass || leftIsInlineClassWithPrimitiveEquality);
-        boolean rightHasPrimitiveEquality = isPrimitive(rightType) && (!rightIsInlineClass || rightIsInlineClassWithPrimitiveEquality);
-
-        if (leftIsInlineClass && !leftIsInlineClassWithPrimitiveEquality || rightIsInlineClass && !rightIsInlineClassWithPrimitiveEquality) {
-            return generateEqualityWithInlineClassesBoxing(left, leftType, right, rightType, opToken, pregeneratedSubject);
-        }
+        boolean leftHasPrimitiveEquality =
+                isPrimitive(leftType) && (!leftIsInlineClass || isInlineClassTypeWithPrimitiveEquality(leftKotlinType));
+        boolean rightHasPrimitiveEquality =
+                isPrimitive(rightType) && (!rightIsInlineClass || isInlineClassTypeWithPrimitiveEquality(rightKotlinType));
 
         if (KtPsiUtil.isNullConstant(left)) {
             return genCmpWithNull(right, opToken, null);
@@ -3406,6 +3401,17 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
 
         if (KtPsiUtil.isNullConstant(right)) {
             return genCmpWithNull(left, opToken, pregeneratedSubject);
+        }
+
+        if (leftIsInlineClass || rightIsInlineClass) {
+            boolean leftIsUnboxed = leftIsInlineClass && StackValue.isUnboxedInlineClass(leftKotlinType, leftType);
+            boolean rightIsUnboxed = rightIsInlineClass && StackValue.isUnboxedInlineClass(rightKotlinType, rightType);
+            if (!leftIsUnboxed || !rightIsUnboxed || !leftHasPrimitiveEquality || !rightHasPrimitiveEquality) {
+                return genEqualsForInlineClasses(
+                        left, right, opToken, pregeneratedSubject, leftKotlinType, rightKotlinType,
+                        leftType, rightType, leftIsUnboxed, rightIsUnboxed
+                );
+            }
         }
 
         if (isIntZero(left, leftType) && rightHasPrimitiveEquality && isIntPrimitive(rightType)) {
@@ -3428,35 +3434,33 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
             return genCmpPrimitiveToSafeCall(left, leftType, (KtSafeQualifiedExpression) right, opToken, pregeneratedSubject);
         }
 
-        if (!leftIsInlineClass && !rightIsInlineClass) {
-            if (BoxedToPrimitiveEquality.isApplicable(opToken, leftType, rightType)) {
-                return BoxedToPrimitiveEquality.create(
-                        opToken,
-                        genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
-                        genLazy(right, rightType), rightType,
-                        myFrameMap
-                );
-            }
+        if (BoxedToPrimitiveEquality.isApplicable(opToken, leftType, rightType)) {
+            return BoxedToPrimitiveEquality.create(
+                    opToken,
+                    genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
+                    genLazy(right, rightType), rightType,
+                    myFrameMap
+            );
+        }
 
-            if (PrimitiveToBoxedEquality.isApplicable(opToken, leftType, rightType)) {
-                return PrimitiveToBoxedEquality.create(
-                        opToken,
-                        genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
-                        genLazy(right, rightType), rightType
-                );
-            }
+        if (PrimitiveToBoxedEquality.isApplicable(opToken, leftType, rightType)) {
+            return PrimitiveToBoxedEquality.create(
+                    opToken,
+                    genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
+                    genLazy(right, rightType), rightType
+            );
+        }
 
-            if (PrimitiveToObjectEquality.isApplicable(opToken, leftType, rightType)) {
-                return PrimitiveToObjectEquality.create(
-                        opToken,
-                        genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
-                        genLazy(right, rightType), rightType
-                );
-            }
+        if (PrimitiveToObjectEquality.isApplicable(opToken, leftType, rightType)) {
+            return PrimitiveToObjectEquality.create(
+                    opToken,
+                    genLazyUnlessProvided(pregeneratedSubject, left, leftType), leftType,
+                    genLazy(right, rightType), rightType
+            );
         }
 
 
-        if (!leftIsInlineClass && !rightIsInlineClass && isPrimitive(leftType) != isPrimitive(rightType)) {
+        if (isPrimitive(leftType) != isPrimitive(rightType)) {
             leftType = boxType(leftType);
             rightType = boxType(rightType);
         }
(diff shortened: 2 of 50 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -244,14 +249,85 @@ private class JvmInlineClassLowering(private val context: JvmBackendContext) : F
...
+
+    private fun IrClass.getEqualsImpl(): IrFunction {
+        val equals = functions.single { it.name.asString() == "equals" && it.overriddenSymbols.isNotEmpty() }
+        return manager.getReplacementFunction(equals)!!.function
+    }
+
+    private fun IrBuilderWithScope.specializeEqualsCall(left: IrExpression, right: IrExpression): IrExpression? {
+        // There's already special handling for null-comparisons in Equals.kt.
+        // We cannot specialize calls for which the first argument is already boxed.
+        if (left.isNullConst() || right.isNullConst() || left.type.unboxInlineClass() === left.type)
```
udalov: We usually avoid using identity equality (`===`) unless it's the specifically required semantics. It's not a big deal here (and below), but it'd probably be more sensible to use `==` to avoid dependency on the fact that `unboxInlineClass` returns the original type in case the class is not `inline`.

sfs: Fixed!

udalov:
@sfs Looks like there's a minor compatibility problem. Currently `equals-impl0` throws exception, so if we call it unconditionally on `==`, we'll get exception for all inline classes compiled with Kotlin 1.3.50 and earlier.

Normally it's allowed to break this because inline classes is an experimental feature without any compatibility guarantees. However, there's one inline class whose API is exposed in the standard library, so we keep its API and ABI stable: [`kotlin.Result`](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/util/Result.kt). Therefore, new code should not call old `Result.equals-impl0` to avoid breaking runtime behavior. And even besides Result, we should probably somehow simplify migration for users of the experimental feature if possible.

Regarding the Result, I see no other solution than to disable this optimization unless API version (`-api-version`) is greater or equal than 1.4, because 1.4 is the least Kotlin release which is sure to have the correct behavior of `Result.equals-impl0` (assuming we land this change) and because binaries of 1.3.x must be fully compatible both ways.

Regarding all other inline classes, we could do the following: distinguish "old" classes from the new ones, for example via the Kotlin metadata version in the class file, and only use specialized equals on the new classes. This can be done by advancing [`JvmMetadataVersion`](https://github.com/JetBrains/kotlin/blob/master/core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/JvmMetadataVersion.kt) to `1.1.16` and then checking if `DeserializedClassDescriptor.metadataVersion` is not less than that when generating a call site.

Of course, the above only concerns the stable JVM backend. Since JVM IR is not yet in production, we can avoid this complexity there. So maybe (but I'm not sure here) it'd make sense to extract the JVM IR implementation to a separate pull request

compiler/backend/src/org/jetbrains/kotlin/codegen/ErasedInlineClassBodyCodegen.kt
```diff
@@ -109,11 +111,16 @@ class ErasedInlineClassBodyCodegen(
 
 
                 override fun doGenerateBody(codegen: ExpressionCodegen, signature: JvmMethodSignature) {
-                    val iv = codegen.v
-                    iv.aconst(null)
-                    iv.athrow()
+                    val asmType = signature.valueParameters[0].asmType
+                    val secondParameterIndex = asmType.getSize()
```
udalov: I think you could use `codegen.frameMap` to get JVM parameter index from a parameter descriptor directly. But it probably wouldn't be easier, so I don't insist

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3476,34 +3480,106 @@ private StackValue generateEquals(
...
+    @NotNull
+    private StackValue genEqualsForInlineClasses(
             @NotNull KtExpression left,
-            @NotNull Type leftType,
             @NotNull KtExpression right,
-            @NotNull Type rightType,
             @NotNull IElementType opToken,
-            @Nullable StackValue pregeneratedSubject
+            @Nullable StackValue pregeneratedSubject,
+            KotlinType leftKotlinType,
```
udalov: Please add `@NotNull`s

compiler/backend/src/org/jetbrains/kotlin/codegen/ErasedInlineClassBodyCodegen.kt
sfs: Oh, thank you! I was actually looking for how to map from parameters to indices and only put in the constants since I didn't find this function... it's fixed now.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: Done!

sfs:
I like the idea of using the metadata version to avoid calling `equals-impl0` methods generated by old compiler versions. I've implemented it, but I'm not sure if it's entirely correct; are there other descriptors which could wrap a `DeserializedClassDescriptor`?

Also, I forgot about `Result`! I wonder if we shouldn't just special case equality for `Result` and compare the underlying types directly. Right now we are already doing this for unsigned types and the alternative is to use the same metadata version check in the IR backend. What do you think?

udalov:
@sfs Thanks! The metadata version check looks good. I think it should be enough to check `DeserializedClassDescriptor`.

Regarding `Result`, we can't generate `equals-impl0` call in 1.3.x because then the binaries and stdlibs compiled by Kotlin 1.3.x would not be interchangeable. Therefore the metadata version check makes no sense for `Result` specifically. We can only generate `equals-impl0` if the stdlib that user depends on is _at least 1.4_, which is approximated by the fact that the API version used in the compilation is at least 1.4. So I propose to use this check (basically `state.getLanguageVersionSettings().getApiVersion().compareTo(ApiVersion.KOTLIN_1_4) >= 0`) instead in case we're dealing with equality on `Result`.

Comparing the underlying values would also work, but that sort of introduces another requirement -- we won't be able to add a custom `equals` implementation to `Result` transparently, so that old code would automatically work. Of course, the possibility that we add a custom `Result.equals` is very low, but it's not zero. :)

Also, this change got me thinking that this makes `Result.equals-impl0` a part of public ABI since 1.4, so I've reported [KT-33678](https://youtrack.jetbrains.com/issue/KT-33678) just in case. I suspect that there should be no problem, but if we discuss and decide to avoid using this method, we'll switch to comparing the underlying values. For now, let's use the API version check approach.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3476,34 +3480,109 @@ private StackValue generateEquals(
...
+            if (!useUnboxedEquals) {
+                StackValue.coerce(leftType, leftKotlinType, OBJECT_TYPE, nullableAnyType, v);
+                afterTopType = OBJECT_TYPE;
+            }
 
             rightValue.put(rightType, rightKotlinType, v);
-            StackValue.coerce(rightType, rightKotlinType, AsmTypes.OBJECT_TYPE, nullableAnyType, v);
+            //noinspection SuspiciousNameCombination
+            Type topType = rightType;
+            if (!useUnboxedEquals || !rightIsUnboxed) {
```
udalov: Interesting that we always box the RHS if the LHS is boxed. Maybe we could swap the arguments with `AsmUtil.swap` and call `equals-impl`, to avoid boxing in cases like this?
```
inline class Z(val s: Long)

fun test(z1: Z?, z2: Z): Boolean {
    //  current:   Intrinsics.areEqual(z1, Z.box-impl(z2))
    // proposed:   Z.equals-impl(z2, z1)
    return z1 == z2
}
```

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3476,34 +3480,109 @@ private StackValue generateEquals(
...
-        KotlinType leftKotlinType = kotlinType(left);
-        KotlinType rightKotlinType = kotlinType(right);
-
         StackValue leftValue = genLazyUnlessProvided(pregeneratedSubject, left, leftType, leftKotlinType);
         StackValue rightValue = genLazy(right, rightType, rightKotlinType);
 
         return StackValue.operation(Type.BOOLEAN_TYPE, v -> {
             KotlinType nullableAnyType = state.getModule().getBuiltIns().getNullableAnyType();
+            Label endLabel = new Label();
+            boolean generateEndLabel = false;
```
udalov: I think we can always generate `endLabel` to simplify code, it shouldn't have any effect on the resulting bytecode

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3476,34 +3480,109 @@ private StackValue generateEquals(
...
+                        generateEndLabel = true;
+                        AsmUtil.dup(v, topType);
+                        Label nonNullLabel = new Label();
+                        v.ifnonnull(nonNullLabel);
+                        AsmUtil.pop2(v, topType, afterTopType);
+                        v.iconst(flipComparison ? 1 : 0);
+                        v.goTo(endLabel);
+                        v.visitLabel(nonNullLabel);
+                    }
+                    String descriptor = Type.getMethodType(Type.BOOLEAN_TYPE, leftType, leftType).toString();
```
udalov: `Type.getMethodDescriptor`

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntUseBox.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM
```
udalov: Maybe it's better to extract the failing part of this test into a separate file and file a YouTrack issue, since we don't have a habit of looking for tests ignored for _JVM_ backend and fixing those

udalov: Also, after changing the behavior for Result, consider adding a separate test checking that we don't invoke `Result.equals-impl0` in the bytecode (that test will change its behavior in 1.4)

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: I implemented this in the beginning! Unfortunately, it's not a sound optimization, see [equalsCallsLeftArgument.kt](https://github.com/JetBrains/kotlin/pull/2539/files#diff-85b28675505ffc8ed17d13bf53dba531).

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: Good to know, done!

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: Fixed

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntUseBox.kt
sfs: Done

sfs:
I've changed the code to avoid `Result.equals-impl0` before version 1.4. However, I'm still thinking that it would be better to compare the underlying values directly for `Result`, simply because that is what we're already doing for unsigned values.

I've also filed an issue for the remaining cases where we end up boxing `Result` under [KT-33722](https://youtrack.jetbrains.com/issue/KT-33722).

udalov:
Some tests using `Result` are failing in our compiler bootstrap build, please take a look: https://teamcity.jetbrains.com/viewLog.html?buildId=2516392&tab=buildResultsDiv&buildTypeId=Kotlin_dev_BootstrapTest

(To run them "honestly", you can set up a second project copy and point `kotlinBootstrapFrom` to `BootstrapOption.Local` with the path to the first project build's repo, which is created with `./gradlew dist install publish`. But since this is a pretty simple test, I suggest to just copy the sources of `Result` and the corresponding test `ResultTest` to a new project or compile/run it with CLI.)

> However, I'm still thinking that it would be better to compare the underlying values directly for `Result`, simply because that is what we're already doing for unsigned values.

I sympathize with your preference. However, there are two key differences of `Result` with unsigned types:
* unsigned types are still experimental, so we can change their behavior in any way
* I'm 99.99% sure unsigned types won't have `equals` which has any behavior other than `equals` on underlying values. Which means that even if they weren't experimental, we could've basically committed to the fact that they won't have custom `equals` and could freely call `equals` on underlying values

One of the goals of [KT-33678](https://youtrack.jetbrains.com/issue/KT-33678) is therefore to verify that we don't intend to introduce custom `equals` to `Result`. If we agree on that in the Kotlin team, we can of course switch to using `equals` on underlying values for `Result` too. Hope this makes sense.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
udalov: That is an awesome test! You're right, it's incorrect in the general case.

I wonder if it makes sense to perform this optimization if it's statically known that the RHS is an _inline_ class, where we know that the equality is symmetric? That would remove boxing from the most popular case, where we compare values of the same inline class type, without changing behavior of others.

This approach could supposedly backfire only in a situation where the RHS's inline class is independently recompiled to a non-inline class with non-symmetric `equals`, and behavior of the `==` check will be different at runtime compared to if it was recompiled. I'm not sure if we need to tolerate this though, because seeing the word "inline", a user _could_ be required to accept that any ABI-affecting change will need to cause clients to recompile. I think this is another design question which I'll raise internally...

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntCallSpecializedEquals.kt
```diff
@@ -0,0 +1,13 @@
+// WITH_COROUTINES
```
udalov: This is really minor, but maybe these tests will succeed even with `WITH_RUNTIME`, now that `Result` in the standard library.

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntCallSpecializedEquals.kt
```diff
@@ -0,0 +1,13 @@
...
+// FILE: test.kt
+fun test() {
+    val result = Result.success("yes!")
+    val other = Result.success("nope")
+    if (result == other) println("==")
+    if (result != other) println("!=")
+    if (result.equals(other)) println("equals")
+    if (!result.equals(other)) println("!equals")
```
udalov: Judging by [KT-33722](https://youtrack.jetbrains.com/issue/KT-33722), these two lines should fail. Probably they don't because we only call `Result.equals-impl0` if API version >= 1.4. So let's add `!API_VERSION: LATEST` to this test, so that it tests something useful. :)

But also, I don't really think the explicit `equals` call is such critical behavior that we should check it at this point in a codegen test. Having it reported as KT-33722 is probably enough. So I'd probably just remove these two lines from the test.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3486,34 +3490,103 @@ private StackValue generateEquals(
...
-        KotlinType leftKotlinType = kotlinType(left);
-        KotlinType rightKotlinType = kotlinType(right);
-
         StackValue leftValue = genLazyUnlessProvided(pregeneratedSubject, left, leftType, leftKotlinType);
         StackValue rightValue = genLazy(right, rightType, rightKotlinType);
 
         return StackValue.operation(Type.BOOLEAN_TYPE, v -> {
             KotlinType nullableAnyType = state.getModule().getBuiltIns().getNullableAnyType();
+            Label endLabel = new Label();
+            boolean flipComparison = opToken == KtTokens.EXCLEQ || opToken == KtTokens.EXCLEQEQEQ;
```
udalov: Maybe it's worth adding a test on identity equals (`===`/`!==`)? I haven't found one

compiler/testData/codegen/bytecodeText/inlineClasses/equalsDoesNotBox.kt
```diff
@@ -0,0 +1,74 @@
...
+// equals-impl
+fun isEqualAnyLeftC(s: C, t: Any?) = s == t
+// boxes
+// fun isEqualAnyRightC(s: Any?, t: C) = s == t
+// Intrinsics.areEqual
+fun isEqualSameNullableC(s: C?, t: C?) = s == t
+fun isEqualAnyNullableLeftC(s: C?, t: Any?) = s == t
+fun isEqualAnyNullableRightC(s: Any?, t: C?) = s == t
+
+// @TestKt.class:
```
udalov: Please consider adding `// FILE: a.kt`, `// FILE: b.kt`,`// FILE: c.kt` to this test and splitting the expected counts with `// @AKt.class` etc., for locality and easier understanding where things like 6 `areEqual` calls are found

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntCallSpecializedEquals.kt
sfs: Indeed they do, fixed!

compiler/testData/codegen/bytecodeText/inlineClasses/resultApiDoesntCallSpecializedEquals.kt
sfs: Right, I don't think specializing explicit `equals` calls is very relevant either. I've removed this part and also got rid of `testResultApiEqualsCallDoesntBox`.

Apart from that, I've changed the test so that it explicitly runs with an old API version (and doesn't use `equals-impl0`), while `testResultApiEquals` runs with the latest API version (and doesn't box).

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: Reference equality is not allowed for values of inline class type, that's why there's no test for it. I've added a comment to this effect in the IR backend.

compiler/testData/codegen/bytecodeText/inlineClasses/equalsDoesNotBox.kt
sfs: Good idea, done!

sfs:
@udalov I've addressed most of your comments and fixed the bug with the `Result` tests. Could you please try running the tests again?

The one remaining thing is the potential for optimizing `x == y` when `y` is unboxed and `x` is boxed. Do you want me to implement this as `y.equals-impl(x)`? It could be potentially confusing once we allow user defined `equals` methods on inline classes and the user could always rewrite the code to `y == x` instead. To be honest, I'd prefer the latter - maybe we could instead add a diagnostic message for this case?

sfs:
By the way, I've added a box test (`testResultRunCatchingOrElse`) for the case that broke in the standard library, since I usually don't run the stdlib tests.

udalov:
@sfs Thanks! All compiler-relevant tests have passed at https://teamcity.jetbrains.com/viewLog.html?buildId=2533807&buildTypeId=Kotlin_dev_AggregateBranch.

Regarding the "boxed equals unboxed" case: you're right, swapping the arguments might be counterintuitive in some rare cases of non-symmetric `equals` implementations. Instead, couldn't we check the left-hand side for null then, unbox it and call `equals-impl0`? As if the user actually wrote `x != null && x == y` and the smart cast made it possible to unbox and use a specialized equals call in the bytecode.

I feel like without this part, users will still complain about redundant boxing where there shouldn't be any. It's not a blocker though, so if you feel like it's too much work, we can merge this PR and postpone the rest of the story for later.

sfs:
> I feel like without this part, users will still complain about redundant boxing where there shouldn't be any.

You're probably right about that.

In general, though, we'll need to add a type check instead of comparing against `null`:
```kotlin
inline class I(val v: Int)

val x: Any? = ...
val y: I = I(0)
x == y
```

In this case we can avoid boxing `y` if `x` is of type `I`, e.g, by creating code for
```kotlin
if (x is I)
   I.`equals-impl0`(x, y)
else
   x.equals(I.`box-impl`(y))
```
For all we know `x` might be something weird, like
```kotlin
val x = object { override fun equals(other: Any?) = other is I }
```
so in the worst case we actually need to box.

udalov:
Oh, I don't suggest to do anything for calls that are resolved to `Any.equals`, I'm only talking about the case where we know that `equals` is from an inline class and thus can be optimized, which is all the cases you've covered in this PR + the case of a nullable inline class on LHS with a non-null inline class on RHS.

sfs:
I see, that makes sense. I'll look into it!

sfs:
@udalov I've implemented additional optimizations for comparisons of the form `x:A? == y` where `A` is an inline class. I've also added some more tests for this case.

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3488,34 +3492,112 @@ private StackValue generateEquals(
...
+                        v.goTo(endLabel);
+                        v.visitLabel(nonNullLabel);
+                    }
+                    if (!leftIsUnboxed) {
+                        AsmUtil.swap(v, topType, afterTopType);
+                        KotlinType nonNullableLeftKotlinType = TypeUtils.makeNotNullable(leftKotlinType);
+                        Type nonNullableLeftType = typeMapper.mapType(nonNullableLeftKotlinType);
+                        StackValue.coerce(leftType, leftKotlinType, nonNullableLeftType, nonNullableLeftKotlinType, v);
+                        afterTopType = nonNullableLeftType;
+                        AsmUtil.swap(v, afterTopType, topType);
```
udalov: The resulting bytecode after all dup/pop instructions here is scary :) I'd imagine that we'd simply load the first argument, check if it's null and jump, then load the second argument and call `equals-impl0`. But it's not a problem, since this is in the old backend. I think we can wait for real complaints before touching this code again

udalov:
@sfs Great work, thank you!

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: Note that left/right could have side effects, so we'd need to store their results in a new temporary variable. That's probably what we should do, it's just that I don't know my way around the old backend.

PR #2540, opened 2019-08-26
fix for KT-29471

kevin1e100:
Proposed fix for KT-29471: preserve objects in scope of inline functions so they can be referenced by the inlined code.  Include test case that fails without this change.


altsvetkov:

PR #2541, opened 2019-08-27
[IR] For loop lowering for iteration over CharacterSequences.

madsager:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -261,8 +264,28 @@ internal class HeaderProcessor(
...
+        //
+        // If the `iterator` method is an extension method, make sure that we are calling a known extension
+        // method in the library such as `kotlin.text.StringsKt.iterator` with no value arguments. Other
+        // extension methods could return user-defined iterators.
+        val iterable = (variable.initializer as? IrCall)?.let {
+            val extensionReceiver = it.extensionReceiver
+            if (extensionReceiver != null) {
+                val function = it.symbol.owner
+                if (it.valueArgumentsCount == 0
+                    && function.name == Name.identifier("iterator")
```
udalov: `OperatorNameConventions.ITERATOR`

core/descriptors/src/org/jetbrains/kotlin/builtins/KotlinBuiltIns.java
```diff
@@ -210,6 +210,8 @@ private Primitives(
         public final FqNameUnsafe intRange = rangesFqName("IntRange");
         public final FqNameUnsafe longRange = rangesFqName("LongRange");
 
+        public final FqNameUnsafe stringsKt = textFqName("StringsKt");
```
udalov: This is a JVM-specific name, so I prefer not to declare it here, but rather store it as a constant at the use site

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
madsager: See reply below. Changed this.

core/descriptors/src/org/jetbrains/kotlin/builtins/KotlinBuiltIns.java
madsager: I ended up changing this with function.fqNameWhenAvailable == FqName("kotlin.text.StringsKt.iterator")

However, this is in a common lowering, so when you say that StringsKt is a JVM specific name that makes me wonder if I cannot check it this way for JS and Native backends?

madsager:
Thanks Alex! Could you have another look? I also changed to share a bit more code since the generated IR is almost the same for arrays and character sequences.

core/descriptors/src/org/jetbrains/kotlin/builtins/KotlinBuiltIns.java
udalov: You're right. Unfortunately, it won't work for Native. So it seems like you need to actually compare for both cases: a member function declared in class `kotlin.text.StringsKt` for JVM, and a top level function declared in package `kotlin.text` for Native. You can use `IrDeclaration.isTopLevel` extension which takes this into account, and then `IrElement.getPackageFragment` extension to acquire the containing package and check if it's `kotlin.text`. See [`isLateinitIsInitializedPropertyGetter`](https://github.com/JetBrains/kotlin/blob/e9b2e74b15ac4536915fb2b74ba8bf9dc93b80eb/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt#L241) for usage example.

madsager: Done, thanks. Now checking that the function is the no value argument top-level method in kotlin.text with the name "iterator". The matcher will then check that the extension receiver is a character sequence and we can apply the optimization.

I'll see if I can test this on kotlin native tomorrow as well to make sure it works.

madsager:
I verified that this version also work in kotlin native. We never got the common lowering used in kotlin native, we should make sure that happens. That will be a change in kotlin-native, I'll start looking into that.

Can you have another look at this one Alex. I think this is ready to go and is only used for JVM right now.

max-kammerer:
@madsager NB: there is still @punzki frozen work in #2417 around additional optimization. As i remember he wanted to make K/N transition on them at some point in future

madsager:
Thanks Mike, looks like that is orthogonal to this change which does character sequences. Therefore, I don't think it matters if this one lands before or after @punzki's change.

I have another change coming that will optimize local variables not only for for loops lowering but for more cases as well. That might make some of the work that @punzki  did in that area unnecessary, but time will tell.

@punzki will be back from leave soon and we can push on this together.

max-kammerer:
@madsager Thanks, I was just aware about K/N migration work that you mentioned ("That will be a change in kotlin-native, I'll start looking into that."). It doen't matter for this change of course.

I sent pr to remote run on TeamCity and will merge it after it.

max-kammerer:
Thanks!

PR #2543, opened 2019-08-27
KT-30643 J2K: wrong position of TYPE_USE annotation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30643

### DIFF
--- a/nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
+++ b/nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
@@ -16,6 +16,7 @@
 
 package org.jetbrains.kotlin.nj2k
 
+import com.intellij.codeInsight.AnnotationTargetUtil
 import com.intellij.lang.jvm.JvmModifier
 import com.intellij.psi.*
 import com.intellij.psi.JavaTokenType.SUPER_KEYWORD
@@ -619,7 +620,7 @@ class JavaToJKTreeBuilder constructor(
                 classKind(),
                 typeParameterList?.toJK() ?: JKTypeParameterList(),
                 createClassBody(),
-                annotationList(this, emptyArray()),
+                annotationList(this),
                 otherModifiers(),
                 visibility(),
                 modality()
@@ -718,13 +719,12 @@ class JavaToJKTreeBuilder constructor(
             visibility(referenceSearcher) { ast, psi -> ast.withFormattingFrom(psi) }
 
         fun PsiField.toJK(): JKField {
-            val type = type
-            val typeAnnotations = type.annotations
+            val typeElement = typeElement
             return JKField(
-                JKTypeElement(type.toJK(), typeAnnotations.toJK()).withFormattingFrom(typeElement),
+                JKTypeElement(type.toJK(), typeElement.annotationList()).withFormattingFrom(typeElement),
                 nameIdentifier.toJK(),
                 with(expressionTreeMapper) { initializer.toJK() },
-                annotationList(this, typeAnnotations),
+                annotationList(this),
                 otherModifiers(),
                 visibility(),
                 modality(),
@@ -736,23 +736,22 @@ class JavaToJKTreeBuilder constructor(
             }
         }
 
-        fun <T : PsiModifierListOwner> T.annotationList(
-            docCommentOwner: PsiDocCommentOwner?,
-            typeAnnotations: Array<out PsiAnnotation>
-        ): JKAnnotationList {
+        fun <T : PsiModifierListOwner> T.annotationList(docCommentOwner: PsiDocCommentOwner?): JKAnnotationList {
             val deprecatedAnnotation = docCommentOwner?.docComment?.deprecatedAnnotation()
             val plainAnnotations = annotations.mapNotNull { annotation ->
                 when {
                     annotation !is PsiAnnotation -> null
                     annotation.qualifiedName == DEPRECATED_ANNOTAION_FQ_NAME && deprecatedAnnotation != null -> null
-                    annotation in typeAnnotations -> null
+                    AnnotationTargetUtil.isTypeAnnotation(annotation) -> null
                     else -> annotation.toJK()
                 }
             }
             return JKAnnotationList(plainAnnotations + listOfNotNull(deprecatedAnnotation))
         }
 
-        private fun Array<out PsiAnnotation>.toJK() = JKAnnotationList(map { it.toJK() })
+        fun PsiTypeElement?.annotationList(): JKAnnotationList {
+            return JKAnnotationList(this?.applicableAnnotations?.map { it.toJK() }.orEmpty())
+        }
 
         fun PsiAnnotation.toJK(): JKAnnotation =
             JKAnnotation(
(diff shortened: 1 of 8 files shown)


nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -719,11 +717,12 @@ class JavaToJKTreeBuilder constructor(
             visibility(referenceSearcher) { ast, psi -> ast.withFormattingFrom(psi) }
 
         fun PsiField.toJK(): JKField {
+            val (annotationList, typeUseAnnotationList) = annotationList(this)
             return JKField(
-                JKTypeElement(type.toJK()).withFormattingFrom(typeElement),
+                JKTypeElement(type.toJK(), typeUseAnnotationList).withFormattingFrom(typeElement),
```
darthorimar: Please, take annotations directly from `PsiField.typeElement`

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -735,18 +734,29 @@ class JavaToJKTreeBuilder constructor(
             }
         }
 
-        fun <T : PsiModifierListOwner> T.annotationList(docCommentOwner: PsiDocCommentOwner?): JKAnnotationList {
+        fun <T : PsiModifierListOwner> T.annotationList(docCommentOwner: PsiDocCommentOwner?): Pair<JKAnnotationList, JKAnnotationList> {
```
darthorimar: And in this function just filter out type-use annotations by checking `owner` property

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -735,18 +736,23 @@ class JavaToJKTreeBuilder constructor(
...
+        fun <T : PsiModifierListOwner> T.annotationList(
+            docCommentOwner: PsiDocCommentOwner?,
+            typeAnnotations: Array<out PsiAnnotation>
+        ): JKAnnotationList {
             val deprecatedAnnotation = docCommentOwner?.docComment?.deprecatedAnnotation()
             val plainAnnotations = annotations.mapNotNull { annotation ->
                 when {
                     annotation !is PsiAnnotation -> null
                     annotation.qualifiedName == DEPRECATED_ANNOTAION_FQ_NAME && deprecatedAnnotation != null -> null
+                    annotation in typeAnnotations -> null
```
t-kameyama: `it.owner` returns PsiModifierList, not PsiTypeElement. Could you tell me how to filter out by `it.owner`?

darthorimar: As far, as I can see `com.intellij.psi.PsiAnnotation#getOwner` has type `PsiAnnotationOwner` and `PsiTypeElement` is subtype of it

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -810,16 +816,19 @@ class JavaToJKTreeBuilder constructor(
...
         fun PsiMethod.toJK(): JKMethod {
+            val returnType = returnType
+            val returnTypeAnnotations = returnType?.annotations.orEmpty()
             return JKMethodImpl(
                 JKTypeElement(
                     returnType?.toJK()
                         ?: JKJavaVoidType.takeIf { isConstructor }
-                        ?: JKNoType),
+                        ?: JKNoType,
+                    returnTypeAnnotations.toJK()),
```
darthorimar: Let's create `PsiTypeElement.toJk()` which will also retrieve annotations from `PsiTypeElement`

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
t-kameyama: I think `getOwner()` returns `PsiModifierList` here:
https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/java/PsiAnnotationImpl.java#L120-L124

darthorimar: You are right :) then `com.intellij.codeInsight.AnnotationTargetUtil#isTypeAnnotation` is your friend here

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -719,8 +719,9 @@ class JavaToJKTreeBuilder constructor(
             visibility(referenceSearcher) { ast, psi -> ast.withFormattingFrom(psi) }
 
         fun PsiField.toJK(): JKField {
+            val typeElement = typeElement
```
darthorimar: Why we need this variable?

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
```diff
@@ -862,9 +868,10 @@ class JavaToJKTreeBuilder constructor(
                 }
 
 
-        fun PsiLocalVariable.toJK(): JKLocalVariable =
-            JKLocalVariable(
-                JKTypeElement(type.toJK()).withFormattingFrom(typeElement),
+        fun PsiLocalVariable.toJK(): JKLocalVariable {
+            val typeElement = typeElement
```
darthorimar: Redundant var?

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
t-kameyama: `this.typeElement` calls: 
https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/PsiFieldImpl.java#L132-L139

We should avoid unnecessary method calls.
https://github.com/t-kameyama/kotlin/blob/07054bdee360b173175d4df6ade67a209ad504cd/nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt#L724

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
t-kameyama: Same as above.

nj2k/src/org/jetbrains/kotlin/nj2k/JavaToJKTreeBuilder.kt
darthorimar: This call is very cheap, so there is no need for that :)

But for avoiding this you can move all functionality related to working with type elements to separated function like `PsiTypeElement.toJk(): JKtypeElement` which will retrieve type, annotations, and formating from type element

darthorimar:
@t-kameyama, two tests are failing now:

1. `NewJavaToKotlinConverterSingleFileTestGenerated.ArrayInitializerExpression#testDoubleArrayWithVariables`
2. `NewJavaToKotlinConverterSingleFileTestGenerated.ArrayInitializerExpression#testOneDimWithVariables`

Please, consider fixing them

darthorimar:
@t-kameyama thanks!

PR #2547, opened 2019-08-29
[JVM_IR, IR] Remove more needless temporary variables.

madsager:
Avoid using a separate origin for temporary variables introduced
for for loops. That doesn't add anything and gives one more case
for optimizations to deal with.

Extend the JVM specific optimizations to remove temporary
variables to deal with more cases encountered in for loops
lowering.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
@@ -310,8 +310,7 @@ internal class HeaderProcessor(
                         progressionType.elementCastFunctionName
                     ),
                     nameHint = "inductionVariable",
-                    isMutable = true,
-                    origin = IrDeclarationOrigin.FOR_LOOP_IMPLICIT_VARIABLE
+                    isMutable = true
                 )
 
                 // Due to features of PSI2IR we can obtain nullable arguments here while actually
@@ -325,8 +324,7 @@ internal class HeaderProcessor(
                             progressionType.elementCastFunctionName
                         )
                     ),
-                    nameHint = "last",
-                    origin = IrDeclarationOrigin.FOR_LOOP_IMPLICIT_VARIABLE
+                    nameHint = "last"
                 )
 
                 val stepValue = scope.createTemporaryVariable(
@@ -336,8 +334,7 @@ internal class HeaderProcessor(
                             progressionType.stepCastFunctionName
                         )
                     ),
-                    nameHint = "step",
-                    origin = IrDeclarationOrigin.FOR_LOOP_IMPLICIT_VARIABLE
+                    nameHint = "step"
                 )
 
                 return when (headerInfo) {
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
@@ -178,8 +178,7 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                         statement.initializer is IrConst<*>
             }
 
-            override fun visitBlock(expression: IrBlock): IrExpression {
-                expression.transformChildrenVoid(this)
+            private fun removeUnnecessaryTemporaryVariables(statements: MutableList<IrStatement>) {
                 // Remove declarations of immutable temporary variables with constant values.
                 // IrGetValue operations for such temporary variables are replaced
                 // by the initializer IrConst. This makes sure that we do not load and
@@ -208,9 +207,10 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
                 //
                 // which allows the equality check to be simplified away and we end up with
                 // just a const string load.
-                expression.statements.removeIf {
+                statements.removeIf {
                     isImmutableTemporaryVariableWithConstantValue(it)
                 }
+
                 // Remove a block that contains only two statements: the declaration of a temporary
                 // variable and a load of the value of that temporary variable with just the initializer
                 // for the temporary variable. We only perform this transformation for compiler generated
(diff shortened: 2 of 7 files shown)


madsager:
Note that this has a minor conflict with: https://github.com/JetBrains/kotlin/pull/2541

They should not be landed together as this will change the expectations in one of the character sequence tests in terms of the amount of local variables used. :)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -285,8 +285,7 @@ internal class HeaderProcessor(
                         progressionType.elementCastFunctionName
                     ),
                     nameHint = "inductionVariable",
-                    isMutable = true,
-                    origin = IrDeclarationOrigin.FOR_LOOP_IMPLICIT_VARIABLE
```
madsager: I left the declaration of this origin in the source code as kotlin-native is still using it until we unify the for loops lowerings.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -222,17 +225,34 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+                return body
+            }
+
+            override fun visitContainerExpression(expression: IrContainerExpression): IrExpression {
+                expression.transformChildrenVoid(this)
+                removeUnnecessaryTemporaryVariables(expression.statements)
+                return expression
+            }
+
+            override fun visitBlock(expression: IrBlock): IrExpression {
```
udalov: Looks like this override is redundant, since super method delegates to `visitContainerExpression` which does the same?

udalov:
I agree, it doesn't look useful to have this origin separate from the standard `IR_TEMPORARY_VARIABLE`. I even had a similar experience (resulting in a [weird workaround](https://github.com/JetBrains/kotlin/blob/cb2e68feced642a5888dd890da2c49a0ac1f04c6/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InventNamesForLocalClasses.kt#L93)) with the related `FOR_LOOP_ITERATOR`, which I hope we can some day also get rid of.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
madsager: Good catch! Removed. :)

compiler/testData/codegen/bytecodeText/forLoop/forInStringSpecialized.kt
```diff
@@ -9,5 +9,3 @@ fun test() {
 // 0 hasNext
 // 0 nextChar
 // 0 INVOKEINTERFACE
-// 1 ISTORE 4
```
madsager: I opted for removing these. I'm not sure why we would want to check that we store and load integers in local 4? With this optimization we don't anymore. What was in local 4 was the constant 1 for the step, and we just use const_1 for that when needed now.

madsager:
Rebased after landing the character sequence change. This is ready for another look and should be safe to land if you agree with the changes. :)

compiler/testData/codegen/bytecodeText/forLoop/forInStringSpecialized.kt
max-kammerer: Maybe @dnpetrov remember an aim for this.
@madsager Could we add // 1 charAt (I)C and // 1 length ()I ?

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/LoopExpressionGenerator.kt
```diff
@@ -197,7 +197,7 @@ class LoopExpressionGenerator(statementGenerator: StatementGenerator) : Statemen
                     irNextCall
                 )
             } else {
-                scope.createTemporaryVariable(irNextCall, "loop_parameter", origin = IrDeclarationOrigin.FOR_LOOP_IMPLICIT_VARIABLE)
+                scope.createTemporaryVariable(irNextCall, "loop_parameter")
```
max-kammerer: @madsager Maybe we should switch K/N to common for-lowers in first cause FOR_LOOP_IMPLICIT_VARIABLE is used in K/N HeaderProcessor. But from opposite I don't see any usage of it outside of them so looks like there would be safe to perform it even now.

@homuroll Would it be a problem with serialized ir if we remove FOR_LOOP_IMPLICIT_VARIABLE is some feature?

udalov: @max-kammerer https://github.com/JetBrains/kotlin-native/commit/e24062155083feef3838791f6e69cb72ceceadda

madsager: @max-kammerer Switch to common for-loop lowering happened here: https://github.com/JetBrains/kotlin-native/pull/3294 :-)

There is still one occurrence of FOR_LOOP_IMPLICIT_VARIABLE in the bitcode generator which can be removed once this lands here and is pulled into kotlin-native. I therefore believe this is safe.

compiler/testData/codegen/bytecodeText/forLoop/forInStringSpecialized.kt
madsager: Thanks! Added charAt and length lines. Looks like the matching doesn't like having the signature there though, so I had to leave that out.

JakeWharton: I believe it's a regex so you'd have to escape the parenthesis.

max-kammerer: Yes, it should be escaped \( \)

madsager: Awesome. Thanks both! :)

max-kammerer:
@madsager Please rebase on master

madsager:
Rebased.

max-kammerer:
Thanks!

PR #2549, opened 2019-08-29
KT-33052: Fix KAPT stub generation for enums on JDK11

gavra0:
When using JDK11, generating stubs for enums is broken.
This commit handles the faulty generation by JDK11 by
fully printing enums in custom implementation of tree
printing.

Test: added JDK11 tests for KAPT stub  generation

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/util/java9Utils.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/util/java9Utils.kt
@@ -24,6 +24,16 @@ import com.sun.tools.javac.util.List as JavacList
 import org.jetbrains.kotlin.kapt3.base.plus
 
 fun isJava9OrLater(): Boolean = !System.getProperty("java.version").startsWith("1.")
+fun isJava11OrLater(): Boolean {
+    val majorVersion = System.getProperty("java.version").substringBefore(".", "")
+    if (majorVersion.isEmpty()) return false
+
+    return try {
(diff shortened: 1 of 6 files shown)


plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/util/java9Utils.kt
```diff
@@ -24,6 +24,16 @@ import com.sun.tools.javac.util.List as JavacList
 import org.jetbrains.kotlin.kapt3.base.plus
 
 fun isJava9OrLater(): Boolean = !System.getProperty("java.version").startsWith("1.")
+fun isJava11OrLater(): Boolean {
```
JakeWharton: Alternatively,
```kotlin
fun isJava11OrLater(): Boolean = try {
  Runtime.Version.parse(System.getProperty("java.version")).major() >= 11
} catch (_: NoClassDefFoundError) {
  false
}
```

JakeWharton: Same could be done for the above method

gavra0: The issue is that KAPT compiles against JDK8, and `Runtime.Version` is added in JDK9. Once it migrates to JDK9, using this is definitely better.

yanex:
Merged manually (0cdfe89a26e16a44c93219ee666d2d4c7f98456d). Thanks!

PR #2552, opened 2019-08-30
[IR] For loop lowering for CharacterSequence.indices.

madsager:


max-kammerer:
Thank you!

madsager:
Thanks @max-kammerer 

Turns out that this way of getting the size property doesn't work for CharSequence for subtypes with type parameters. I'll upload a fix for that shorty. 

@homuroll, FYI

PR #2553, opened 2019-08-30
Consistently copyAttributes in DeepCopyIrTreeWithSymbols

sfs:
Right now, the call to `copyAttributes` is missing for `IrFunctionReference` and `IrTypeOperator` in `DeepCopyIrTreeWithSymbols`, which can cause problems with names for local classes.

To avoid the same bug popping up again in the future, I've added `copyAttributes` to all relevant classes in `DeepCopyIrTreeWithSymbols`, even though this is currently only used for a handful of cases.


udalov:
Looks great, thank you!

PR #2555, opened 2019-08-30
Fix some errors in JVM_IR inlining

pyos:


pyos:
@max-kammerer ping?

max-kammerer:
@pyos will look soon

max-kammerer:
@pyos Is it possible to split this pr to smaller ones?

At least I would prefer to have separate pr that combine changes in old inliner core part  AnonymousObjectTransformer/InliningContext and so on, e.g. at least next commits:
JVM_IR: do not regenerate anonymous objects in lambdas. 
Shift parameter annotations on constructors of local classes
Always capture locals for inline IR lambdas directly 
Reset captured parameter indices when copying descriptions 
Remove the old copy of captured params after adding them to the end

pyos:
@max-kammerer ok, I'll try to.

pyos:
^ split into separate PRs.

PR #2558, opened 2019-09-02
JVM IR: Implement visibility rules for CallableReferences

sfs:
This is a new implementation of #2523, which moves the logic for setting the visibility of callable references into several lowerings instead of keeping it in the backend.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
@@ -16,6 +16,7 @@ import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.backend.jvm.ir.createJvmIrBuilder
 import org.jetbrains.kotlin.backend.jvm.ir.irArrayOf
+import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.builders.*
 import org.jetbrains.kotlin.ir.builders.declarations.addFunction
@@ -29,6 +30,7 @@ import org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeImpl
 import org.jetbrains.kotlin.ir.types.impl.makeTypeProjection
 import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+import org.jetbrains.kotlin.load.java.JavaVisibilities
 import org.jetbrains.kotlin.load.java.JvmAbi
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.types.Variance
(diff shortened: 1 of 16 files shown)


sfs:
The largest change here is actually in `PropertyReferenceLowering`. Previously, in the IR backend we would cache both property reference classes and instances.

The former is problematic, since we want different visibilities for property reference classes, depending on whether they are used in the scope of an inline function (also the reflection metadata would be incorrect).

This then also makes the second kind of caching superfluous. This kind of caching is also problematic since it breaks binary compatibility (property reference instances are accessed by index into the `$$delegatedProperties` array, which can change when adding new property references even in private functions).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -165,20 +165,17 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
                 // For delegated properties, the getter and setter contain a reference each as the second argument to getValue
                 // and setValue. Since it's highly unlikely that anyone will call get/set on these, optimize for space.
                 val useReflectedImpl = expression.origin == IrStatementOrigin.PROPERTY_REFERENCE_FOR_DELEGATE
+
+                // Reflected implementation does not support partial application; also, cannot cache instances with arguments.
+                if (!useReflectedImpl || expression.dispatchReceiver != null || expression.extensionReceiver != null)
+                    return createSpecializedKProperty(expression)
+
                 return context.createIrBuilder(currentScope!!.scope.scopeOwnerSymbol, expression.startOffset, expression.endOffset).run {
                     val (_, index) = kProperties.getOrPut(PropertyCacheKey(expression.symbol, useReflectedImpl)) {
```
udalov: Looks like `useReflectedImpl` is always true here now. And then `PropertyCacheKey` is not needed and `kProperties` can simply be a list, since reflected KProperty objects are only created for delegated properties, and we never meet any delegated property more than once.

However, I wonder if getting rid of caching is correct in the long term, given that [KT-15690](https://youtrack.jetbrains.com/issue/KT-15690) is a popular issue. In any case, we should probably decide where such a cached property reference class is generated and what name it has first, before implementing this optimization. Keeping them in `$$delegatedProperties` before your changes was sort of a hack, since that array was only supposed to be used for delegated properties. So I suppose your changes are OK, and we can revisit the caching later

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -165,20 +165,17 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
-                // Reflected implementation does not support partial application; also, cannot cache instances with arguments.
-                if (expression.dispatchReceiver != null || expression.extensionReceiver != null)
-                    return createSpecializedKProperty(expression)
-
                 // For delegated properties, the getter and setter contain a reference each as the second argument to getValue
                 // and setValue. Since it's highly unlikely that anyone will call get/set on these, optimize for space.
                 val useReflectedImpl = expression.origin == IrStatementOrigin.PROPERTY_REFERENCE_FOR_DELEGATE
+
+                // Reflected implementation does not support partial application; also, cannot cache instances with arguments.
+                if (!useReflectedImpl || expression.dispatchReceiver != null || expression.extensionReceiver != null)
```
udalov: I think property reference for delegate is guaranteed not to have receivers, so it should be enough to check `!useReflectedImpl` (and inline it)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
sfs: Thanks, I've simplified the code. It turns out that we do potentially meet every delegated property twice (once in the getter and once in the setter), so the code ended up only a little bit simpler.

As for caching, there are two issues here.

Right now we are caching *instances* in `$$delegatedProperties`, which causes some problems with inlining and leads to an unstable ABI. I don't think we want to do this - we may want to generate singleton classes for unbound property references instead, but that is a separate issue.

The issue described in KT-15690 is about caching generated classes which we should definitely look into. We can certainly share generated classes within the same scope. However, right now the code shares generated classes whenever possible. Sharing the same class between different scopes causes problems with reflection and inlining and we should probably avoid it.

Unfortunately, it's also not entirely straightforward to check which scope we will end up in after inlining (e.g., I think the generated classes in `fun f() { run { A::x }; A::x }` could be shared), so I'd like to defer this change to a separate PR.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
sfs: Good to know! Fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
udalov: Thanks! This makes total sense and I agree that the second issue can be addressed later.

udalov:
Thank you!

PR #2559, opened 2019-09-03
A bit more native targets

sbogolepov:
Android x86 and 32-bit watchOS.


konan/utils/src/org/jetbrains/kotlin/konan/target/KonanTarget.kt
```diff
@@ -14,11 +14,19 @@ enum class Family(val exeSuffix:String, val dynamicPrefix: String, val dynamicSu
                   val staticPrefix: String, val staticSuffix: String) {
     OSX     ("kexe", "lib", "dylib", "lib", "a"),
     IOS     ("kexe", "lib", "dylib", "lib", "a"),
+    TVOS     ("kexe", "lib", "dylib", "lib", "a"),
```
olonho: Better align like others.

PR #2560, opened 2019-09-03
[IR] Fix issue getting the length property from CharSequence bounded …

madsager:
…type parameter

Add regression tests for the cases that failed.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
@@ -17,7 +17,6 @@ import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.IrCall
 import org.jetbrains.kotlin.ir.expressions.IrExpression
-import org.jetbrains.kotlin.ir.symbols.IrSimpleFunctionSymbol
 import org.jetbrains.kotlin.ir.symbols.IrSymbol
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.*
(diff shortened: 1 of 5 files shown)


madsager:
@max-kammerer @homuroll 

Sorry about the breakage. :)

madsager:
Updated to use the class when available and otherwise revert back to the CharSequence interface. This allows the use of INVOKEVIRTUAL of java/lang/String.length and java/lang/String.charAt in the JVM_IR backend instead of INVOKEINTERFACE of the CharSequence method.

max-kammerer:
@madsager Could you add a bytecode test for INVOKEVIRTUAL?

madsager:
@max-kammerer  It is already there in the sense that the currently failing forInStringSpecialized expects no INVOKEINTERFACE and expects a call to charAt and to length (once all the changes in flight are landed). Should I make the test explicitly check for `// 2 INVOKEVIRTUAL` as well?

max-kammerer:
@madsager Oh, no. I forgot about forInStringSpecialized
Thank you!!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -375,16 +380,22 @@ internal class ArrayIterationHandler(context: CommonBackendContext) : IndexedGet
         get() = getClass()!!.properties.first { it.name == Name.identifier("size") }.getter!!
 
     override val IrType.getFunction
-        get() = getClass()!!.functions.first { it.name.asString() == "get" }
+        get() = getClass()!!.functions.first { it.name == Name.identifier("get") }
```
max-kammerer: I would prefer to use 'name.asString() == "get"' instead of Name object creation or at least extract Name object to global properties

madsager: Done, thanks!

udalov: @madsager Note `asString`, not `toString` :) There's no difference right now, but we don't usually call `toString` in production code

madsager: d'oh! ;-)

Done!

max-kammerer:
Thanks!

PR #2561, opened 2019-09-04
[IR] Avoid duplicate IR elements in for loops lowering.

madsager:


madsager:
@max-kammerer @homuroll 

This should fix the duplicate IR elements noticed when using these lowerings in kotlin-native.

homuroll:
Confirm that the PR fixes the [IR VALIDATION] warnings.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -102,12 +102,12 @@ internal sealed class ForLoopHeader(
                     //   // (use `<` if last is exclusive)
                     //   (step > 0 && inductionVar <= last) || (step < 0 || last <= inductionVar)
                     val stepKotlinType = progressionType.stepType(builtIns).toKotlinType()
-                    val zero = if (progressionType == ProgressionType.LONG_PROGRESSION) irLong(0) else irInt(0)
+                    val isLong = progressionType == ProgressionType.LONG_PROGRESSION;
```
olonho: Semicolon may be considered a nostalgic reminiscence here.

madsager: Whoops. Old habits die hard. ;-) Removed.

max-kammerer:
Thanks!

PR #2563, opened 2019-09-04
YarnSetupTask: Fix setup() to be a Gradle TaskAction

sschuberth:

PR #2564, opened 2019-09-04
YarnSetupTask: Remove a too coarse "onlyIf" in the init block

sschuberth:
"installationDir" is the non-version specific top-level Yarn
installation directory simply called "yarn". Checking for its existence
is not sufficient to be able to skip the task, as another version than
the requested one could be installed below the "yarn" directory.

Instead of taking the version into account in this check, simply remove
it as the version-specific installation directory at "env.home" is
already marked as an "OutputDirectory", so Gradle itself already takes
care of skipping task execution if the output files were created before.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/yarn/YarnSetupTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/yarn/YarnSetupTask.kt
@@ -28,10 +28,6 @@ open class YarnSetupTask : DefaultTask() {
     init {
         group = NodeJsRootPlugin.TASKS_GROUP_NAME
         description = "Download and install a local yarn version"
-
-        onlyIf {
-            !settings.installationDir.exists()
-        }
     }
 
     @Action


snrostov:
@sschuberth Thanks for the contribution!

Changes itself looks good, but we need tests for that case. Please add on at `libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle`
Also, please create issue in youtrack with problem description. This is required for our QA process and it will appear in changelog. 

Please tell if you can do it (we can do it for you if you cannot).

sschuberth:
Looks like @ilgonmic already created a ticket at https://youtrack.jetbrains.com/issue/KT-33716.

ilgonmic:
@sschuberth If you will, It can be useful to see similar test about YarnSetupTask, related with https://github.com/JetBrains/kotlin/pull/2563
https://github.com/JetBrains/kotlin/blob/c615541461096a60eba6f1d777724dff1cf995d8/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/Kotlin2JsGradlePluginIT.kt#L401-L419

ilgonmic:
@sschuberth I merged this PR and will test it on my own
Thank you for contribution!

sschuberth:
Thanks @ilgonmic, it would have taken me some more time to write a test...

PR #2565, opened 2019-09-05
JVM_IR: Fix null-constant comparison with primitive types.

kandersen:
```
1.equals(null)
```
Caused generation of invalid bytecode.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
@@ -13,12 +13,14 @@ import org.jetbrains.kotlin.codegen.intrinsics.Not
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.declarations.IrFile
+import org.jetbrains.kotlin.ir.declarations.IrFunction
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrBlockImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
 import org.jetbrains.kotlin.ir.types.isBoolean
+import org.jetbrains.kotlin.ir.types.isNullableAny
 import org.jetbrains.kotlin.ir.types.isPrimitiveType
 import org.jetbrains.kotlin.ir.types.toKotlinType
 import org.jetbrains.kotlin.ir.util.*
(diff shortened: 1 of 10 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -43,56 +45,72 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
-                    } else {
-                        IrBlockImpl(expression.startOffset, expression.endOffset, expression.type, expression.origin).apply {
-                            statements.add(nonNullArgument.coerceToUnitIfNeeded(nonNullArgument.type.toKotlinType(), context.irBuiltIns))
-                            statements.add(constFalse)
+
+                if (isNegation(expression, context) && isNegation(expression.dispatchReceiver!!, context)) {
+                    return (expression.dispatchReceiver as IrCall).dispatchReceiver!!
+                }
+
+                getOperandsIfCallToEqeqOrEquals(expression)?.let {
```
max-kammerer: (left, right) ->

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -43,56 +45,72 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+
                         }
+
+                        else -> expression
                     }
-                } else if (isNullCheckOfConstant(expression, context)) {
-                    if (expression.getValueArgument(0)!!.isNullConst() && expression.getValueArgument(1)!!.isNullConst()) {
-                        IrConstImpl.constTrue(expression.startOffset, expression.endOffset, context.irBuiltIns.booleanType)
-                    } else {
-                        IrConstImpl.constFalse(expression.startOffset, expression.endOffset, context.irBuiltIns.booleanType)
```
max-kammerer: false optimization is lost

compiler/testData/codegen/box/primitiveTypes/equalityWithObject/intEqualsNull.kt
```diff
@@ -0,0 +1,2 @@
+fun box(): String =
```
max-kammerer: Please also add bytecode test or convert this one to it: use as example compiler/testData/codegen/bytecodeText/conditions/ tests

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
kandersen: I believe the `isNullCheckOfConstant` check is subsumed by the call to `hasNoSideEffectsForNullCompare`. I.e. paths that hit line 89 (`null == null`) end at 82. Paths ending at 91 (`null == "constant"`, `4.equals(null)`, ...) now end at 90.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
kandersen: Great, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmBuiltinOptimizationLowering.kt
```diff
@@ -43,56 +45,72 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+                        left.isNullConst() && right.isNullConst() ->
+                            IrConstImpl.constTrue(expression.startOffset, expression.endOffset, context.irBuiltIns.booleanType)
+
+                        it.second.isNullConst() && it.first.type.unboxInlineClass().isPrimitiveType() ||
+                                it.first.isNullConst() && it.second.type.unboxInlineClass().isPrimitiveType() -> {
+                            val nonNullArgument = if (left.isNullConst()) right else left
+                            val constFalse =
+                                IrConstImpl.constFalse(expression.startOffset, expression.endOffset, context.irBuiltIns.booleanType)
+                            if (hasNoSideEffectsForNullCompare(nonNullArgument)) {
+                                constFalse
```
max-kammerer: It's for null == primitive, but null == "non-primitive" is lost cause of 'case' condition

max-kammerer: Bytecode test is required

kandersen: You are absolutely correct. Added bytecode tests to showcase presence of optimization (and the absence of a possible optimization when we find an easier way to pick out the default equals method.

compiler/testData/codegen/bytecodeText/conditions/nullCompareConst.kt
```diff
@@ -0,0 +1,11 @@
+// IGNORE_BACKEND: JVM
```
max-kammerer: Please note that you can customize bytecode tests by backend via 
..common_templates...
// JVM_TEMPLATES
...
// JVM_IR_TEMPLATES
...
, see for example forInReversedArrayIndices.kt

compiler/testData/codegen/bytecodeText/conditions/nullCompareConst.kt
```diff
@@ -0,0 +1,11 @@
...
+fun f(): Boolean = "non-primitive" == null
+fun g(): Boolean = null == "non-primitive"
+fun h(): Boolean = "non-primitive".equals(null)
+//fun i(): Boolean = null.equals("non-primitive")
+//see nullCompareNonPrimitiveConst
+
+// 0 ACONST_NULL
```
max-kammerer: Please also check 

// 0 IF

compiler/testData/codegen/bytecodeText/conditions/nullCompareConst.kt
max-kammerer: So IGNORE_BACKEND could be removed

kandersen: Ah, neat! And that is preferred over the individual tests?

max-kammerer: Yes, to avoid test data duplication and decrease number of IGNORE directives

kandersen:
@max-kammerer So, as the PR stands, there are two new bytecode tests. `nullCompareConst.kt` documenting the (correct as of this PR) behaviour of JVM and JVM_IR backends. `nullCompareNonPrimitiveConst.kt` (with `IGNORE_BACKEND: JVM, JVM_IR`) demonstrates an unexploited oppportunity for optimization.

Both fine, or would you prefer the second elided from this PR?

max-kammerer:
@kandersen It's best to extract nullCompareNonPrimitiveConst.kt from pr and file an issue to YT for further optimizations around this case

kandersen:
@max-kammerer Will do!

max-kammerer:
Thank you!

PR #2567, opened 2019-09-05
Support symlinked JAVA_HOME

saschpe:
When testing for the JAVA_HOME environment variable, expand potential
symlinks first. This is almost guaranteed to be the case in all modern
Linux distributions putting the JDK / JRE behind update-alternatives to
be able to select the *current* version [0], e.g. on Fedora:

  $ ls -l $JAVA_HOME
  lrwxrwxrwx. ... /usr/lib/jvm/java -> /etc/alternatives/java_sdk

[0] https://linux.die.net/man/8/update-alternatives


4u7:
Thanks for the contribution!

ilya-g:
@saschpe Could you remind what was the problem when JDK path was symlinked to another directory?

PR #2568, opened 2019-09-05
Git: Ignore dependencies/dependencies.properties

saschpe:
It seems to be a user-specific generated file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /android.tests.dependencies
 /confluence/target
 /dependencies/repo
+/dependencies/dependencies.properties
 /dist
 /local
 /gh-pages


4u7:
Just wondering where did it come from? I don't have any in my working copy

saschpe:
Judging from it's contents, it's IntelliJ (I'm using 2019.2.1):
```
#
#Thu Sep 05 13:27:32 CEST 2019
markdown.build.id=0.1.25
idea.build.id=191.6707.61
```

4u7:
Found that It is generated by `writeVersions` task in https://github.com/JetBrains/kotlin/blob/master/prepare/ide-lazy-resolver/build.gradle.kts
We don't run this task during development process, only on CI. Looks like you somehow triggered it. Have you called `writeVersions` manually?

saschpe:
Could have been `./gradlew build` or `./gradlew install` as well.

4u7:
I got the problem and will fix it slightly other way. The problem is dependencies dir is not a best place for this artifact, I'll move it to the better place. Thank you!

PR #2569, opened 2019-09-05
Update Jetbrains Java annotations to 17.0.0

saschpe:
Fixes an incompatibility when using kotlin-stdlib and Testcontainers,
i.e. in a Kotlin/Multiplatform project:

  Could not determine the dependencies of task ':core:lint'.
  > Could not resolve all artifacts for configuration ':core:debugAndroidTestRuntimeClasspath'.
     > Could not resolve org.jetbrains:annotations:{strictly 13.0}.
       Required by:
           project :core
        > Cannot find a version of 'org.jetbrains:annotations' that satisfies the version constraints:
             Dependency path 'foo:core:0.2.4' --> 'org.jetbrains.kotlin:kotlin-stdlib:1.3.50' --> 'org.jetbrains:annotations:13.0'
             Constraint path 'foo:core:0.2.4' --> 'org.jetbrains:annotations:{strictly 13.0}' because of the following reason: debugRuntimeClasspath uses version 13.0
             Dependency path 'foo:core:0.2.4' --> 'org.testcontainers:testcontainers:1.12.0' --> 'org.jetbrains:annotations:17.0.0'

No real upstream changelog provided [0].

[0] https://github.com/JetBrains/java-annotations/releases


JakeWharton:
As of version 15 or 16 the annotations are built with Java 8 which means this change would make the Kotlin stdlib incompatible with running on Java 6 or 7.

JakeWharton:
Per the error message, the problem seems to be whatever `:foo:core` is, not the kotlin-stdlib. Gradle would have happily automatically upgraded the kotlin-stdlib dependency to 17 automatically were it not for `:foo:core`'s strict dependency on 13.

saschpe:
> As of version 15 or 16 the annotations are built with Java 8 which means this change would make the Kotlin stdlib incompatible with running on Java 6 or 7.

I didn't think of that...

> Per the error message, the problem seems to be whatever `:foo:core` is, not the kotlin-stdlib. Gradle would have happily automatically upgraded the kotlin-stdlib dependency to 17 automatically were it not for `:foo:core`'s strict dependency on 13.

According to this (heavily snipped) output of `$ ./gradlew :core:dependencies`, it's a direct dep of only the *kotlin-stdlib* artifact. The only other case is Android Lint but that isn't an issue.

```
iosX64Default
+--- io.ktor:ktor-client-core:1.2.4
|    +--- org.jetbrains.kotlinx:atomicfu:0.12.11
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.3.50
|    |    |    \--- org.jetbrains:annotations:13.0

jvmDefault
+--- com.squareup.sqldelight:runtime:1.2.0
|    \--- com.squareup.sqldelight:runtime-jvm:1.2.0
|         +--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.3.50
|         |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.3.50
|         |         \--- org.jetbrains:annotations:13.0

kotlinCompilerClasspath
\--- org.jetbrains.kotlin:kotlin-compiler-embeddable:1.3.50
     +--- org.jetbrains.kotlin:kotlin-stdlib:1.3.50
     |    \--- org.jetbrains:annotations:13.0

kotlinCompilerPluginClasspath
\--- org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:1.3.50
     +--- org.jetbrains.kotlin:kotlin-scripting-compiler-impl-embeddable:1.3.50
     |    +--- org.jetbrains.kotlin:kotlin-scripting-common:1.3.50
     |    |    +--- org.jetbrains.kotlin:kotlin-reflect:1.3.50
     |    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.3.50
     |    |    |         \--- org.jetbrains:annotations:13.0

lintClassPath - The lint embedded classpath
\--- com.android.tools.lint:lint-gradle:26.5.0
     +--- com.android.tools:sdk-common:26.5.0
     |    +--- com.android.tools:sdklib:26.5.0
     |    |    +--- com.android.tools.layoutlib:layoutlib-api:26.5.0
     |    |    |    +--- com.android.tools:common:26.5.0
     |    |    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.41
     |    |    |    |         +--- org.jetbrains.kotlin:kotlin-stdlib:1.3.41
     |    |    |    |         |    \--- org.jetbrains:annotations:13.0
     |    |    |    \--- org.jetbrains:annotations:13.0

releaseCompileClasspath - Resolved configuration for compilation for variant: release
+--- org.jetbrains.kotlin:kotlin-stdlib:1.3.50
|    \--- org.jetbrains:annotations:13.0
```

For now, the current workaround is:
```gradle
implementation("org.testcontainers:testcontainers:1.12.0") {
    exclude(group = "org.jetbrains", module = "annotations")
}
```
It's much easier this way than *exclude*-it from all those kotlin artifacts. Using *exclude* because *KotlinDependencyHandler* from kotlin-multiplatform doesn't support `constraints {}` yet...

ilya-g:
We're not going to merge this PR for the reasons already mentioned by @JakeWharton:
> this change would make the Kotlin stdlib incompatible with running on Java 6 or 7

Also upgrading stdlib dependency version doesn't fix the root cause: strict dependency version constraint preventing Gradle version conflict resolution. Should a new annotation library version come and testcontainers depend on it, the same version conflict will happen.

I suggest you to find who's adding that strict constraint and why. It can be Android Gradle plugin, who requires the same version of artifact in compileClasspath and runtimeClasspath configurations and therefore adds a strict constraint to the latter configuration. You can try adding annotations:17.0.0 as a direct implementation dependency, so that the right version gets into the compileClasspath configuration.

vlsi:
For historical reasons. I just got the same "org.jetbrains:annotations:13.0 strictly", and it was caused by 

```kotlin
java {
    consistentResolution { 
        useCompileClasspathVersions()
     }
 }
```

In other words, compile classpath was using an old annotations library, and some of the runtime libraries "failed" to update the annotation version.

Of course, the solution is to add a common constraint with the desired `org.jetbrains:annotations` version.

PR #2572, opened 2019-09-06
KT-19578: Add radians and degrees conversion functions

saschpe:
Many functions in the math package, especially trigonometric ones,
expect radians. So far, conversion functions from and to degrees where
missing.

### DIFF
--- a/libraries/stdlib/common/src/kotlin/MathH.kt
+++ b/libraries/stdlib/common/src/kotlin/MathH.kt
@@ -477,8 +477,19 @@ public expect fun Double.roundToInt(): Int
 @SinceKotlin("1.2")
 public expect fun Double.roundToLong(): Long
 
-// endregion
+/**
+ * Converts a degree angle to radians.
+ */
+@SinceKotlin("1.3")
+public expect fun radians(degrees: Double): Double
 
+/**
+ * Converts a radian angle to degrees.
+ */
+@SinceKotlin("1.3")
+public expect fun degrees(radians: Double): Double
+
+// endregion
 
 
 // region ================ Float Math ========================================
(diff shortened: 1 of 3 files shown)


altavir:
Unit conversion is rather complicated matter and is required in different fields (for example in new time API). I oppose the idea of function that multiply number by 2*PI/360. Those could be added as extensions wherever it is needed. I think in should be discussed further. Probably better to add a type-safe way like inline classes wrappers. The problem with this approach is that one needs to redefine all operations for those inline classes.

saschpe:
> Unit conversion is rather complicated matter and is required in different fields (for example in new time API). I oppose the idea of function that multiply number by 2*PI/360.

Not sure what approach you have in mind but this commit just exposes what is already available in java.lang.Math. Those functions didn't seem to have caused any issues since their introduction in Java 1.2.

> Those could be added as extensions wherever it is needed.

Extensions to Float and Double for the most part, yes. But both classes mostly contain comparison and type conversion functions. Since this is a unit conversion function it better belonged to the *math* package and kept it similar to what is already there. It's a matter of taste and I'd be fine either way. Let's see what others have to say...

> Probably better to add a type-safe way like inline classes wrappers. The problem with this approach is that one needs to redefine all operations for those inline classes.

Exactly. And AFAIR inline classes are still experimental. Also, we'd have to introduce *Radian* and *Degree* types first.

altavir:
In java, we do not have a way to provide type-safe units solution. In Kotlin we do so it does not make sense to bloat stdlib with half-way solutions. At least names should be something like `degreeToRadian`, but I do not think that those methods make sense at all since they could be implemented by simple extension on user side when it is needed.

PR #2575, opened 2019-09-09
KT-15700 "Convert lambda to reference" does not work with backtick-escaped references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15700


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
```diff
@@ -278,7 +279,12 @@ open class ConvertLambdaToReferenceIntention(text: String) :
             }
         }
 
-        private fun KtCallExpression.getCallReferencedName() = (calleeExpression as? KtNameReferenceExpression)?.getReferencedName()
+        private fun KtCallExpression.getCallReferencedName() = (calleeExpression as? KtNameReferenceExpression)?.getSafeReferencedName()
+        
+        private fun KtNameReferenceExpression.getSafeReferencedName(): String {
```
dimonchik0036: I think you should replace the body of this function with `getReferencedNameAsName().render()`.

PR #2578, opened 2019-09-10
KT-17209 Provide intention to fix platform declaration clash (CONFLICTING_JVM_DECLARATIONS)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17209

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/AddJvmNameAnnotationFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/AddJvmNameAnnotationFix.kt
@@ -10,12 +10,13 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.diagnostics.Diagnostic
 import org.jetbrains.kotlin.idea.KotlinBundle
-import org.jetbrains.kotlin.idea.core.CollectingNameValidator
 import org.jetbrains.kotlin.idea.core.KotlinNameSuggester
+import org.jetbrains.kotlin.idea.core.NewDeclarationNameValidator
 import org.jetbrains.kotlin.idea.util.addAnnotation
 import org.jetbrains.kotlin.idea.util.findAnnotation
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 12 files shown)


Takhion:
Could the generated name be based on the type parameters (like in this [example](https://youtrack.jetbrains.com/issue/KT-34406))?

Advantages:
- stable/reproducible name generation
  - not tied to function order
  - less potential VCS conflicts
- less likely to be renamed
  - captures the actual difference between the overloads
  - meaningful to invoke from other JVM languages

idea/src/org/jetbrains/kotlin/idea/quickfix/AddJvmNameAnnotationFix.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.kotlin.diagnostics.Diagnostic
+import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.core.CollectingNameValidator
+import org.jetbrains.kotlin.idea.core.KotlinNameSuggester
+import org.jetbrains.kotlin.idea.util.addAnnotation
+import org.jetbrains.kotlin.idea.util.findAnnotation
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.*
+
+class AddJvmNameAnnotationFix(element: KtElement, private val jvmName: String) : KotlinQuickFixAction<KtElement>(element) {
```
dimonchik0036, line 20: Shouldn't we replace `jvmName` with `functionName` and move the logic of `jvmName` inside `invoke`?

t-kameyama: If `KotlinNameSuggester.suggestNameByName` call is moved inside `invoke`, Tests fails here:
https://github.com/JetBrains/kotlin/blob/3a5f42cc5ea618dae04bcb954bb1fea738a54a4c/idea/tests/org/jetbrains/kotlin/idea/quickfix/AbstractQuickFixTest.kt#L170-L172

Should we add `AddJvmNameAnnotationFix` to `idea/testData/quickfix/allowResolveInWriteAction.txt`?

dimonchik0036: You right, we shouldn't do this.

PR #2582, opened 2019-09-13
KT-26047 Refactor -> Rename: Overridden method renaming in generic class doesn't rename base method

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26047


t-kameyama:
I think it was fixed in https://github.com/JetBrains/kotlin/commit/0c4134470dfcd09805499b1a0834ea1410c6501b, so added tests.

t-kameyama:
I think it was fixed in https://github.com/JetBrains/kotlin/commit/0c4134470dfcd09805499b1a0834ea1410c6501b, so this pull request is tests only.

igoriakovlev:
We could't accept this because there are some conflicts exists.
Please resolve it.

igoriakovlev:
Thank you for your work.

PR #2583, opened 2019-09-13
[IR] Support Collection<*>.indices in for loop lowering.

madsager:
Additionally unmute a couple of tests that are working as intended.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
@@ -222,8 +222,6 @@ internal class UntilHandler(private val context: CommonBackendContext, private v
 /** Builds a [HeaderInfo] for progressions built using the `indices` extension property. */
 internal abstract class IndicesHandler(protected val context: CommonBackendContext) : ProgressionHandler {
 
-    // TODO: Handle Collection<*>.indices
-
     override fun build(expression: IrCall, data: ProgressionType, scopeOwner: IrSymbol): HeaderInfo? =
         with(context.createIrBuilder(scopeOwner, expression.startOffset, expression.endOffset)) {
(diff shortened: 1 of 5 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -247,7 +245,7 @@ internal abstract class IndicesHandler(protected val context: CommonBackendConte
 
 internal class ArrayIndicesHandler(context: CommonBackendContext) : IndicesHandler(context) {
```
punzki: Suggestion: Rename to `CollectionIndicesHandler`

compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInCollectionIndices.kt
```diff
@@ -12,6 +11,16 @@ fun test() {
 // 0 getFirst
 // 0 getLast
 
+// JVM non-IR uses while.
+// JVM IR uses if + do-while.
+
+// JVM_TEMPLATES
 // 0 IF_ICMPGT
```
punzki: These two lines (`// 0 IF_...`) can be removed; they are made redundant by the `// 1 IF`

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
madsager: Good idea, done.

compiler/testData/codegen/bytecodeText/forLoop/forInIndices/forInCollectionIndices.kt
madsager: Thanks. Done.

udalov:
@madsager Thanks!

PR #2584, opened 2019-09-14
KT-24352 Method separators: displayed between properties, not displayed between companion object and function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24352


vladimirdolzhenko:
LGTM. Thank you for your contribution

PR #2586, opened 2019-09-15
KT-29364 "Extend selection" can't select lambda body with parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29364


darthorimar:
Thank you!

PR #2588, opened 2019-09-16
JVM IR: Generate public SAM wrappers in inline function scope

sfs:
This PR fixes the visibility of generated SAM wrappers (public inside the scope of an inline function and package private otherwise) and makes the caching scheme match the current backend.

There is still some work left to do in order to make `singleAbstractMethodPhase` produce correct IR types, but at least the generated code is correct.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
@@ -92,7 +92,7 @@ class SingleAbstractMethodLowering(val context: CommonBackendContext) : FileLowe
                     val instance = irCall(implementation.constructors.single()).apply {
                         putValueArgument(0, irGet(invokableVariable))
                     }
-                    irIfNull(superType, irGet(invokableVariable), irNull(), instance)
+                    +irIfNull(superType, irGet(invokableVariable), irNull(), instance)
                 }
             } else {
                 irCall(implementation.constructors.single()).apply { putValueArgument(0, invokable) }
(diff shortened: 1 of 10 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -14,76 +14,102 @@ import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
...
+            parentClass.declarations += wrapper
         }
-        return result
     }
 
-    override fun lower(irFile: IrFile) {
-        val fileClass = irFile.declarations.filterIsInstance<IrClass>().find { it.origin == IrDeclarationOrigin.FILE_CLASS }
-        withCacheFor(fileClass) { irFile.transformChildrenVoid() }
+    override fun visitClassNew(declaration: IrClass): IrStatement {
+        enclosingClass = enclosingClass ?: declaration
```
udalov: I'd prefer to refactor this line to
```
if (enclosingClass == null)
    enclosingClass = declaration
```

compiler/testData/codegen/bytecodeText/sam/samWrapperForNullableInitialization.kt
```diff
@@ -17,8 +16,16 @@ fun test() {
     JFoo.foo2({}, runnable())
 }
 
+// JVM_TEMPLATES
 // @TestKt.class:
```
max-kammerer: Minor:
```
// @TestKt.class
// 2 NEW
``` could be extracted in common part

sfs: Done

sfs:
Rebased on master and unmuted an additional test (`testCopySamOnInline2`)

udalov:
Thanks!

PR #2590, opened 2019-09-16
psi2ir: Optimize generated data class members

sfs:
This PR changes `DataClassMembersGenerator` to
- Access underlying fields directly, instead of using accessors. This is safe since data classes are always final.
- Don't generate a temporary variable in hashCode to use as the accumulator.

Both optimizations are effectively present in the current JVM backend and with these changes the generated code is much closer to what the current backend generates.

Most of the changes are just updating `IrText` tests.


dnpetrov:
Looks good to me.

udalov:
@sfs Thanks!

PR #2592, opened 2019-09-17
KT-33586 "Constructors are not allowed for objects" diagnostic needs quickfix to change object to class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33586


dimonchik0036:
Thanks! https://github.com/JetBrains/kotlin/commit/d86e87d35ee698086dd6d5827e0b240412271ba2

PR #2597, opened 2019-09-19
KT-31999 "Variable declaration could be moved into `when`" inspection suggests to inline expression containing return (throw) statement

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31999


idea/src/org/jetbrains/kotlin/idea/inspections/MoveVariableDeclarationIntoWhenInspection.kt
```diff
@@ -75,6 +77,10 @@ private fun KtWhenExpression.findDeclarationNear(): KtProperty? {
     return previousProperty.takeIf { !it.isVar && it.hasInitializer() && it.nameIdentifier?.text == subjectExpression?.text }
 }
 
+private fun KtProperty.hasReturnOrThrowExpression(): Boolean {
+    return this.initializer?.anyDescendantOfType<KtExpression> { it is KtReturnExpression || it is KtThrowExpression } == true
```
mglukhikh: I think the same should be true for `break` & `continue`, so for all kinds of jumps.

mglukhikh:
Merged 👍

PR #2599, opened 2019-09-19
JVM IR: Cache inline class replacements at the module level

sfs:
Currently, we are caching inline class replacements at the file level, which breaks when you have an inline function taking inline class arguments defined in a different file in the same module. This PR just moves the cache to the backend context, effectively caching at the level of modules instead.


udalov:
Thank you!

PR #2600, opened 2019-09-19
Minor: Fix some test cases

sfs:
As part of #2111, @ting-yuan muted some bytecode text tests for the JVM_IR backend which were optimized away. At the time, the problem was that we were also removing line numbers, and the tests remained so that we would know to come back to this issue again.

Well, turns out that this was fixed at some point (probably before the PR got merged), but the superfluous tests never got removed.

### DIFF
--- a/compiler/testData/codegen/bytecodeText/deadCodeElimination/simpleConstructorNotRedundant.kt
+++ b/compiler/testData/codegen/bytecodeText/deadCodeElimination/simpleConstructorNotRedundant.kt
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
 class A
 fun box() {
     val x: A? = A()
@@ -9,12 +8,17 @@ fun box() {
     else {
         y = null
     }
-    
+
     y!!
 }
 
 // 0 IFNULL
-// 1 IFNONNULL
-// 1 throwNpe
 // 0 ATHROW
 // 0 checkNotNull
+// 1 throwNpe
+
+// JVM_TEMPLATES:
+// 1 IFNONNULL
+
+// JVM_IR_TEMPLATES:
+// 0 IFNONNULL
--- a/compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
+++ b/compiler/testData/codegen/bytecodeText/deadCodeElimination/boxing.kt
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
 class A
 
 fun foo(x: Any?) {}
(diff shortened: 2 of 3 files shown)


udalov:
As far as I understood from the discussion in https://github.com/JetBrains/kotlin/pull/2111#discussion_r260763583, these tests are not superfluous since they at least check the old backend's behavior in the case of "optimizable" comparisons (which the old backend doesn't optimize). I'd propose to make the following changes instead:
* `boxing.kt`: leave the test, but remove `IGNORE_BACKEND: JVM_IR` and add `JVM_IR_TEMPLATES`
* `boxingNotOptimizable.kt`: leave the test without modification

And similarly to other tests.

Also, could you please explain what do you mean by

> the problem was that we were also removing line numbers

?

sfs:
What I got from the discussion was that at the time we turned, e.g., the `box` function in `primitiveToString.kt` into a function with an empty body *without NOPs and linenumbers*. The reason to keep the tests was so that we could fix that behavior down the line. That part is no longer a problem if you look at the bytecode generated for these tests - we still optimize the constant expression away, but we keep a NOP + linenumber information.

Do we actually want to check that the old backend doesn't optimize constant expressions? In that case, I'll just change the tests in the way you indicated.

udalov:
Do you mean that we actually generated empty body without a single instruction, and that caused the test to fail? Otherwise I don't understand how line numbers would be relevant for a bytecode text test which doesn't check them. Could you link to a thread where that's discussed?

> Do we actually want to check that the old backend doesn't optimize constant expressions?

Yes, in general we'd like to keep existing tests that verify how the old backend works, even if it's suboptimal at the moment.

compiler/testData/codegen/bytecodeText/deadCodeElimination/simpleConstructorNotRedundant.kt
```diff
@@ -9,12 +8,19 @@ fun box() {
...
+// JVM_TEMPLATES:
 // 0 IFNULL
 // 1 IFNONNULL
 // 1 throwNpe
 // 0 ATHROW
 // 0 checkNotNull
+// JVM_IR_TEMPLATES:
+// 0 IFNULL
+// 0 IFNONNULL
+// 0 throwNpe
```
udalov: This should probably be `// 1 throwNpe`

udalov: Also note that you can extract common parts for additional clarity:
```
// 0 IFNULL
// 1 throwNpe
// 0 ATHROW
// 0 checkNotNull

// JVM_TEMPLATES:
// 1 IFNONNULL
// JVM_IR_TEMPLATES:
// 0 IFNONNULL
```

sfs: Right, fixed and refactored!

udalov:
@sfs Thanks.

PR #2604, opened 2019-09-20
KT-33235 Remove "Replace guard clause with kotlin's function call" inspection and tranform it to J2K post-processing

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33235


mglukhikh:
Merged 👍

PR #2605, opened 2019-09-20
Proposal API: Add scoping controls for parsing Km types

ZacSweers:
Right now, Kotlinx-metadata's KmClass types eagerly parse every possibly bit of information. Not every case requires every possible bit of information though, and some performance could be improved by only deserializing what's requested from the underlying proto.

I've started an implementation below with what I'm describing, and can continue figuring this out if this proposal is deemed acceptable/welcomed.

Example use cases (mostly from https://github.com/square/kotlinpoet/issues/755)
- [Moshi kotlin codegen (annotation processor)](https://github.com/square/moshi)
  - Only cares about constructors, properties, and class signature/hierarchy (type parameters, visibility).
  - Doesn't care about enums, inner classes, companion objects, functions.
- [Dagger (annotation processor)](https://github.com/google/dagger)
  - Is just starting kotlin support and cares a lot about processor performance. Right now its support is limited to reading properties, so it could skip almost everything else in KmClass until they add more functionality.
- [CopyDynamic (annotation processor)](https://github.com/zacsweers/copydynamic)
  - Only cares about constructors and class signature
- [Androidx Room (annotation processor)](https://developer.android.com/topic/libraries/architecture/room)
  - Only cares about constructors, properties, and class signature/hierarchy
- [Moshi-sealed (annotation processor)](https://github.com/zacsweers/moshi-sealed)
  - Only cares about sealed class signatures and sealed class hierarchies
- API jar generation
  - Private/internal APIs are parsed currently but don't need to be in the case of API jar generation tools. Examples  Configuration to strip these could allow parsing of only public API (and also turn that around to generate API stubs/jars).

CC @udalov

### DIFF
--- a/libraries/kotlinx-metadata/jvm/src/kotlinx/metadata/jvm/KotlinClassMetadata.kt
+++ b/libraries/kotlinx-metadata/jvm/src/kotlinx/metadata/jvm/KotlinClassMetadata.kt
@@ -35,18 +35,23 @@ sealed class KotlinClassMetadata(val header: KotlinClassHeader) {
 
         /**
          * Visits metadata of this class with a new [KmClass] instance and returns that instance.
+         *
+         * @param scope the scopes to visit.
          */
-        fun toKmClass(): KmClass =
-            KmClass().apply(this::accept)
+        fun toKmClass(scope: Scope.Class = Scope.Class.ALL): KmClass =
+            KmClass().apply {
+                accept(this, scope)
+            }
 
         /**
          * Makes the given visitor visit metadata of this class.
          *
          * @param v the visitor that must visit this class
+         * @param scope the scopes to visit.
          */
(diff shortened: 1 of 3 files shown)


udalov:
Do I understand correctly that the main motivation for your proposal is performance improvement? In that case, I think we need to have specific benchmarks to confirm that this change is useful.

My main doubt comes from the fact that underneath, the whole protobuf message is still being deserialized regardless of the scope (and it doesn't seem possible to avoid that, except introducing a custom protobuf deserializer), and I suspect that that takes the majority of execution time.

ZacSweers:
Right that's the main motivation. I didn't realize the whole proto was being deserialized regardless under the hood, so in that case I'll just close this out 👍 . May be worth revisiting if the format ever changes to something lazier like FlatBuffers

PR #2609, opened 2019-09-20
Fix gradle integeration and KAPT tests

gavra0:
Fix test failure on Windows, where backslash was not escaped
in a build file. Also, remove obsolete test for Gradle 4.2,
because 4.3 is the minimum supported version.

Also, run KAPT JDK 11 tests only if there is JDK_11 path specified.


yanex:
Merged manually (6e7f0464719b663cc48f8f05ae01b596ff68be23, 573e7f2eee923eb072f6bb4521027daa26f826f5). Thanks for the contribution!

PR #2611, opened 2019-09-22
KT-18772 "Introduce subject to when": don't choose an object or a constant as the subject

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18772

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/branchedTransformationUtils.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/branchedTransformationUtils.kt
@@ -6,6 +6,7 @@
 package org.jetbrains.kotlin.idea.intentions.branchedTransformations
 
 import org.jetbrains.kotlin.idea.core.replaced
+import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.idea.util.CommentSaver
 import org.jetbrains.kotlin.idea.util.psi.patternMatching.matches
 import org.jetbrains.kotlin.lexer.KtTokens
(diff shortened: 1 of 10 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/branchedTransformationUtils.kt
```diff
@@ -68,7 +71,11 @@ private fun KtExpression?.getWhenConditionSubjectCandidate(): KtExpression? = wh
         val lhs = left
         when (operationToken) {
             KtTokens.IN_KEYWORD, KtTokens.NOT_IN -> lhs
-            KtTokens.EQEQ -> lhs as? KtNameReferenceExpression ?: right
+            KtTokens.EQEQ -> {
+                lhs as? KtNameReferenceExpression
+                    ?: (lhs as? KtQualifiedExpression)?.takeIf { it.selectorExpression is KtNameReferenceExpression }
```
mglukhikh: It's better to apply here the same checks for the right part, if the left one does not satisfy our conditions.

idea/testData/inspectionsLocal/branched/introduceWhenSubject/qualified.kt
```diff
@@ -0,0 +1,13 @@
...
+
+sealed class Platform {
+    object JvmPlatform : Platform()
+    class Another(val name: String) : Platform()
+}
+
+class ModuleInfo(val platform: Platform)
+
+fun foo(moduleInfo: ModuleInfo) = <caret>when {
+    moduleInfo.platform == JvmPlatform -> 1
```
mglukhikh: Please check how it works if the condition is flipped: `JvmPlatform == moduleInfo.platform`.

idea/src/org/jetbrains/kotlin/idea/intentions/branchedTransformations/branchedTransformationUtils.kt
```diff
@@ -82,6 +88,15 @@ private fun KtExpression?.getWhenConditionSubjectCandidate(): KtExpression? = wh
     else -> null
 }
 
+private fun KtExpression.hasCandidateNameReferenceExpression(): Boolean {
+    val nameReferenceExpression = this as? KtNameReferenceExpression
+        ?: (this as? KtQualifiedExpression)?.selectorExpression as? KtNameReferenceExpression
+        ?: return false
+    val resolved = nameReferenceExpression.mainReference.resolve()
```
mglukhikh: @t-kameyama This change breaks "if to when" intention because resolve of factory-created elements is not possible. Yet I forbid these checks during "if to when", but you may try to make a nicer fix.

mglukhikh: See this commit: https://github.com/JetBrains/kotlin/commit/d45e83b27d20cfc4b708ed15d68d6973258e49a3

mglukhikh:
Merged manually (but see my comment about if-to-when).

PR #2612, opened 2019-09-23
JVM IR: Implement -Xjvm-default=enable

kandersen:


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/impl/IrClassImpl.kt
```diff
@@ -99,4 +99,7 @@ class IrClassImpl(
         typeParameters.transform { it.transform(transformer, data) }
         declarations.transform { it.transform(transformer, data) }
     }
+
+    override fun toString(): String = descriptor.toString()
```
max-kammerer: Why you need toString override here and below? 
We have several utilities for debug purpose: ir2string() and dump()

kandersen: The debugger uses toString for formatting the little labels next to objects in the value explorer. It saves you 3-4 mouseclicks through the descriptors/names/owners etc in order to see e.g. the class name. 

I am completely open to suggestions!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceDelegationLowering.kt
```diff
@@ -222,7 +225,7 @@ internal val interfaceDefaultCallsPhase = makeIrFilePhase(
 )
 
 private class InterfaceDefaultCallsLowering(val context: JvmBackendContext) : IrElementTransformerVoid(), FileLoweringPass {
-
+    // TODO If there are no default _implementations_ we can avoid generating defaultImpls class entirely by moving default arg dispatchers to the interface class
```
max-kammerer: NB: if any annotation on property is present it's stored on synthetic methods in DI (and we can support interface as storage class)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -94,7 +97,7 @@ private class InterfaceLowering(val context: JvmBackendContext) : IrElementTrans
         }
     }
 
-    private fun shouldRemoveFunction(function: IrFunction): Boolean =
+    private fun mustMoveToDefaultImpls(function: IrFunction): Boolean =
         Visibilities.isPrivate(function.visibility) && function.name != clinitName ||
                 function.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER ||
```
max-kammerer: Please note that for @JvmDefault function  '$default' companion is not moved to DefaultImpls class (except maybe compatibility mode)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/impl/IrClassImpl.kt
max-kammerer: Actually you can specify custom rendered for any class in IJ debugger)

gogabr: In any case, please don't use descriptor's `toString()` here. 
- This will not work properly for declarations created or moved in lowerings.
- In general, we are moving away from descriptor use.

kandersen: Max's suggestion is precisely what I needed. Will drop both toString's from this PR.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
kandersen: OK, I think I understand the behaviour. In the case of @JvmDefault+compat, $default is implemented on the interface itself with DefaultImpls containing a simple delegation to that implementation.

With this PR, JVM_IR is on par with the old backend.

Could you clarify your intent with this comment? I think neither name (shouldRemove/mustMove) is true to the behaviour for FUNCTION_FOR_DEFAULT_PARAMETER.

max-kammerer: Seems this comment is more suitable for line 56 ``` if(function.hasJvmDefault() && !context.state.jvmDefaultMode.isCompatibility && !mustMoveToDefaultImpls(function)) continue```. Maybe flag about @JvmDefault should be passed to this funtion

max-kammerer: @kandersen My point was to fix a bit behaviour with  '$default' companion function, but it's could be postponed for another commit.

kandersen: I have a major revision of this entire PR in the works, I think I might have adressed this in there. We'll see once I land it :)

kandersen:
@max-kammerer @sfs Did a major rewrite and simplification of the logic spawned by a desire to not generate DefaultImpls companion if it contains no members. This revealed that the logic in a later pass depended on that empty class being present in order to fire. I collected this logic in the same pass in order to remove this implicit dependency.

With things more local, it allowed a simplification of the logic which should more explicitly document correct handling of default implementations in interfaces, with and without @JvmDefault implementations.

I think I also caught that note you mentioned about $default methods on companions -- I'll point it out in the file.

 Please have a look, looking forward to your comments. Will branch from here and take a further look at the 4 interface passes, because later passes re-targeting super and interface calls are making the same case distinction all over again, just from a call site rather than at time of declaration. I think this can be simplified such that the logic of "should this function be moved" can all be gathered in one, single place.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -22,59 +25,95 @@ import org.jetbrains.kotlin.ir.expressions.IrReturn
...
 
-                val element = context.declarationFactory.getDefaultImplsFunction(function).also {
-                    if (mustMoveToDefaultImpls(function))
-                        removedFunctions[function.symbol] = it.symbol
+                /**
+                 * 3) Private methods, default parameter dispatchers (without @JvmDefault)
+                 *    and $annotation methods are always moved without bridges
+                 */
+                Visibilities.isPrivate(function.visibility)
+                        || (function.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER && !function.hasJvmDefault())
```
kandersen: @max-kammerer This is the point you made earlier, no?

max-kammerer:
@kandersen Please rebase on master cause of conflict reported by github in compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
max-kammerer: Look like yes (cause of changed code). But it's not clear is JvmDefault annotation presented on $default companion (shouldn't we check original function)?

max-kammerer:
@kandersen Looks cool. 
Please note that in compatibility mode delegation is reversed (but it's not necessary for now):

```
public abstract interface A {

  public default foo(I)I
    ILOAD 1
    IRETURN

  public static synthetic foo$default(LA;IILjava/lang/Object;)I
    ... 
    INVOKEINTERFACE A.foo (I)I (itf)


  // access flags 0x1009
  public static synthetic access$foo$jd(LA;I)I
    INVOKESPECIAL A.foo (I)I (itf)
}

public final class A$DefaultImpls {
  
  public static foo(LA;I)I
  ...
    INVOKESTATIC A.access$foo$jd (LA;I)I (itf)

  public static synthetic foo$default(LA;IILjava/lang/Object;)I 
   ...
    INVOKESTATIC A.foo$default (LA;IILjava/lang/Object;)I (itf)

}
```

kandersen:
@max-kammerer 
> Please note that in compatibility mode delegation is reversed (but it's not necessary for now):
Thanks for pointing this out, this has indeed come up doing our design of this refactor. There doesn't seem to be any tests that constrain this choice -- any particular reason the existing backend went with implementations on the interface + accessors?

max-kammerer:
Actually @JvmDefault delegates method dispatch to runtime (as java default methods works), that could changes semantic of statically linked DefaultImpls methods (in some cases with binary library update). So main point was to keep this runtime dispatch in binary compatibile mode.

e.g.
```
library 
interface A {
  fun foo()  = "A"
}

interface Left : A {}

interface Right : A {
 // override  fun foo()  = "Right"
}

//usage
interface C : Left, Right {
  implicit delegation to Left.foo (or Right.foo in case of interface C : Right, Left)
}
```
If then Right will add binary compatible override it wouldn't cause any effect to statically linked C during binary library update

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
kandersen: Yep, [we copy all annotations on to the $default method](https://github.com/JetBrains/kotlin/blob/ea56a5e8b1eebcbd6648db2105712daf9ee7ace0/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt#L458).

kandersen:
@max-kammerer I rebased, should be good to merge.

I am looking at setting up a test that will reveal this behaviour, as it's currently underspecified by the test suite.

max-kammerer:
Many thanks!

kandersen:
Likewise!

PR #2613, opened 2019-09-23
Start scope for destructuring variables after initialization.

madsager:
If the scope for a local variable is started before a value
has been written, another value from a previous use of the local
slot can be present. That value could have a different type which
would lead to weird debugging situations and also leads to other
tools (such as D8) rejecting the locals information as it is
invalid.

Fixes KT-33959.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/range/forLoop/AbstractForLoopGenerator.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/range/forLoop/AbstractForLoopGenerator.kt
@@ -101,13 +101,17 @@ abstract class AbstractForLoopGenerator(
             })
         }
 
-        v.visitLabel(destructuringStartLabel)
-
         codegen.initializeDestructuringDeclarationVariables(
             destructuringDeclaration,
             TransientReceiver(elementType),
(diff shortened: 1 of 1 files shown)


madsager:
@ilmirus @udalov I'm still looking into adding a regression test for this that is not *very* specific to the exact code generated. The best way to test this would probably be using a debugging test infrastructure that actually steps through the code on a JVM. We are adding such infrastructure, but it is not there yet. I'll see what I can come up with. :)

ilmirus:
Thanks for the fix! To be honest, it took me a while to figure out, why this change was needed, since JVM_IR generates code, which does not have the bug. Only old BE is affected.

As for the regression test, you can reuse existing debugger facility from ide tests, located at `testData/debugger/tinyApp/evaluate/singleBreakpoint/`, as I did for my `LINENUMBER` related changes.

For example, you can write something like
```
class C {
    val map: Map<String, Any> = mapOf<String, Any>("1", "23")
    fun f() {
        // Breakpoint!
        for ((a, b) // (1)
                in map) {
            println("$a $b")
        }
    }
}
```
and below the code something like
```
// EXPRESSION: a + b
// RESULT: "123": Ljava/lang/String;
```

PR #2615, opened 2019-09-23
JVM IR: Generate synthetic methods for type aliases with annotations

kandersen:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeAliasAnnotationMethodsLowering.kt
```diff
@@ -0,0 +1,75 @@
...
+        declaration.acceptChildrenVoid(this)
+        declaration.visitTypeAliases()
+    }
+
+    private val IrTypeAlias.syntheticAnnotationMethodName
+        get() = Name.identifier(JvmAbi.getSyntheticMethodNameForAnnotatedTypeAlias(name))
+
+    private fun IrDeclarationContainer.visitTypeAliases() {
+        val annotatedAliases = declarations.filterIsInstance<IrTypeAlias>().filter { !it.descriptor.annotations.isEmpty() }
+        for( alias in annotatedAliases ) {
```
max-kammerer: Please reformat

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt
```diff
@@ -42,6 +42,8 @@ interface IrDeclarationOrigin {
     object IR_BUILTINS_STUB : IrDeclarationOriginImpl("IR_BUILTINS_STUB")
     object BRIDGE : IrDeclarationOriginImpl("BRIDGE", isSynthetic = true)
     object BRIDGE_SPECIAL: IrDeclarationOriginImpl("BRIDGE_SPECIAL")
+    object SYNTHETIC_METHOD_FOR_TYPEALIAS_ANNOTATIONS :
```
max-kammerer: As it's only jvm-specific please move it to JvmLoweredDeclarationOrigin

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/TypeAliasAnnotationMethodsLowering.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    override fun lower(irFile: IrFile) {
+        irFile.acceptVoid(this)
+    }
+
+    override fun visitElement(element: IrElement) {
+        element.acceptChildrenVoid(this)
+    }
+
+    override fun visitFile(declaration: IrFile) {
```
max-kammerer: This override is redundant cause of fileClassPhase, that moves file content into class

max-kammerer: ...and lower could be switched to ClassLoweringPass

kandersen: Is the same true of PropertiesLowering?

kandersen: It is! Just tried tried out the refactoring of PropertiesLowering to a ClassLoweringPass. Do you want that in this PR, or should I extract that to a separate one?

max-kammerer: It's best to make separate ibe

max-kammerer: Oh, no. PropertiesLowering is common lowering. And as i remeber Native and JS don't perform fileClassPhase

kandersen: Ah, right! I just spotted the same as I went to check the changes in and I noticed the path was different.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
```diff
@@ -157,7 +157,7 @@ inline fun buildFun(builder: IrFunctionBuilder.() -> Unit): IrFunctionImpl =
         buildFun()
     }
 
-inline fun IrDeclarationContainer.addFunction(builder: IrFunctionBuilder.() -> Unit): IrSimpleFunction =
+inline fun IrDeclarationContainer.addFunction(builder: IrFunctionBuilder.() -> Unit): IrFunctionImpl =
```
max-kammerer: Is this change is neccessery for pr?

kandersen: Yes, it allows line 54 in TypeAliasAnnotationMethodsLowering.kt. It's, as far as I can tell, an unnecessary restriction to the superclass.

max-kammerer:
Many thanks!

PR #2617, opened 2019-09-23
Fix running incremental KAPT on JDK 9

gavra0:
Incremental annotation processing was relying on internal JDK types/method that were removed in JDK 9. This PR fixes that are removes dependency on `BaseFileObject` and `JavacProcessingEnvironment.validImportStringToPattern`.

Fixes KT-33889


yanex:
Merged (594f4a6f560..f89885224ef). Thanks for the contribution!

PR #2620, opened 2019-09-25
Do not cache "last" value in lowered for-loop over CharSequences

punzki:
CharSequences may be mutable (e.g., StringBuilder) and therefore its contents and length can change within the loop.

This issue was introduced in #2541. This fix matches the optimization in current JVM backend (https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/range/forLoop/ForInCharSequenceLoopGenerator.kt#L56).


max-kammerer:
@punzki Please look at failed tests: https://teamcity.jetbrains.com/viewLog.html?buildId=2553594&buildTypeId=Kotlin_dev_CompilerWithTests

punzki:
> @punzki Please look at failed tests: https://teamcity.jetbrains.com/viewLog.html?buildId=2553594&buildTypeId=Kotlin_dev_CompilerWithTests

Thanks for pointing that out. I was using functions not available in JS. Fixed.

max-kammerer:
Many thanks!

PR #2621, opened 2019-09-25
Update deprecated doc about mac build environment requirements

Matts966:
It is not JDK 6 but JDK 9 that is needed for build, so added installation guide of jdk9 instead of 6.


vladimirdolzhenko:
Sorry for a long delay,

Actually java6 is available at caskroom/versions: https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/java6.rb

so, instructions could be like
```
$ brew tap caskroom/versions
$ brew cask install java6
```

Matts966:
@vladimirdolzhenko 
Thank you. I updated the doc.

vladimirdolzhenko:
LGTM. Thanks

PR #2622, opened 2019-09-26
KT-21172 Join declaration and assignment should place the result at the assignment, not at declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21172


idea/src/org/jetbrains/kotlin/idea/intentions/JoinDeclarationAndAssignmentIntention.kt
```diff
@@ -76,25 +77,33 @@ class JoinDeclarationAndAssignmentIntention : SelfTargetingRangeIntention<KtProp
...
 
-        val initializerBlock = assignment.parent.parent as? KtAnonymousInitializer
-        assignment.delete()
-        if (initializerBlock != null && (initializerBlock.body as? KtBlockExpression)?.isEmpty() == true) {
-            initializerBlock.delete()
+        val grandParent = assignment.parent.parent
+        val initializerBlock = grandParent as? KtAnonymousInitializer
+        val secondaryConstructor = grandParent as? KtSecondaryConstructor
+        val newProperty = if (initializerBlock != null || secondaryConstructor != null) {
+            assignment.delete()
```
mglukhikh: In this branch, *two* different cases are possible, and it seems to me you consider only one of them:

- *local variable* declared *and* assigned in anonymous initializer / secondary constructor, in this case result should be placed at the assignment
- *property* of the class which is assigned in anonymous initializer / secondary constructor, in this case result should be placed at the declaration

idea/src/org/jetbrains/kotlin/idea/intentions/JoinDeclarationAndAssignmentIntention.kt
```diff
@@ -76,25 +77,33 @@ class JoinDeclarationAndAssignmentIntention : SelfTargetingRangeIntention<KtProp
...
+            if ((initializerBlock?.body as? KtBlockExpression)?.isEmpty() == true) initializerBlock.delete()
```
mglukhikh: Why can't you do the same for the secondary constructor?

idea/src/org/jetbrains/kotlin/idea/intentions/JoinDeclarationAndAssignmentIntention.kt
```diff
@@ -75,25 +77,34 @@ class JoinDeclarationAndAssignmentIntention : SelfTargetingRangeIntention<KtProp
...
-        assignment.delete()
-        if (initializerBlock != null && (initializerBlock.body as? KtBlockExpression)?.isEmpty() == true) {
-            initializerBlock.delete()
+        val grandParent = assignment.parent.parent
+        val initializerBlock = grandParent as? KtAnonymousInitializer
+        val secondaryConstructor = grandParent as? KtSecondaryConstructor
+        val newProperty = if (!element.isLocal && (initializerBlock != null || secondaryConstructor != null)) {
+            assignment.delete()
+            if ((initializerBlock?.body as? KtBlockExpression)?.isEmpty() == true) initializerBlock.delete()
+            if ((secondaryConstructor?.bodyBlockExpression)?.isEmpty() == true) secondaryConstructor.delete()
```
mglukhikh: Careful please. Constructor can have delegation, in this case you may delete its block but not constructor itself.

Before
```
constructor(...): this(...) {
    x = 42
}
```

After

```
constructor(...): this(...)
```

mglukhikh:
Merged manually, thank you

PR #2624, opened 2019-09-26
Introduce local variable type checker.

madsager:
CheckLocalVariablesTableTests will now check the validity of
the locals table against types of locals computed based on the
bytecode.

These checks and the new destructuringInFor test act as a
regression test for the changes in
https://github.com/JetBrains/kotlin/pull/2613

These checks also caught a similar issue for destructuring
lambda parameters, where the local is introduced before the
value has been written to the local slot. This change also
fixes that.

Finally, this change fixes the asmLike tests to correctly
look up the name of parameters in the locals table.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ClosureGenerationStrategy.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ClosureGenerationStrategy.kt
@@ -27,7 +27,8 @@ class ClosureGenerationStrategy(
 ) : FunctionGenerationStrategy.FunctionDefault(state, declaration) {
 
     override fun doGenerateBody(codegen: ExpressionCodegen, signature: JvmMethodSignature) {
-        initializeVariablesForDestructuredLambdaParameters(codegen, codegen.context.functionDescriptor.valueParameters)
+        initializeVariablesForDestructuredLambdaParameters(
+            codegen, codegen.context.functionDescriptor.valueParameters, codegen.context.methodEndLabel)
         if (declaration is KtFunctionLiteral) {
             recordCallLabelForLambdaArgument(declaration, state.bindingTrace)
         }
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
@@ -29,7 +29,6 @@ import org.jetbrains.kotlin.descriptors.*;
 import org.jetbrains.kotlin.descriptors.annotations.Annotated;
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor;
 import org.jetbrains.kotlin.descriptors.impl.AnonymousFunctionDescriptor;
-import org.jetbrains.kotlin.descriptors.impl.ValueParameterDescriptorImpl;
 import org.jetbrains.kotlin.load.java.BuiltinMethodsWithSpecialGenericSignature;
 import org.jetbrains.kotlin.load.java.JvmAbi;
 import org.jetbrains.kotlin.load.java.SpecialBuiltinMembers;
@@ -767,7 +766,7 @@ public class FunctionCodegen {
         generateLocalVariablesForParameters(mv,
                                             jvmMethodSignature, functionDescriptor,
                                             thisType, methodBegin, methodEnd, functionDescriptor.getValueParameters(),
-                                            AsmUtil.isStaticMethod(ownerKind, functionDescriptor), state, shiftForDestructuringVariables
+                                            AsmUtil.isStaticMethod(ownerKind, functionDescriptor), state
         );
     }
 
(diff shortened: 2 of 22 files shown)


madsager:
@ilmirus @max-kammerer can you have a look when you have the time. Thanks! :)

@yanex since I'm changing the AsmLike tests.

compiler/backend/src/org/jetbrains/kotlin/codegen/ClosureGenerationStrategy.kt
```diff
@@ -27,7 +27,8 @@ class ClosureGenerationStrategy(
 ) : FunctionGenerationStrategy.FunctionDefault(state, declaration) {
 
     override fun doGenerateBody(codegen: ExpressionCodegen, signature: JvmMethodSignature) {
-        initializeVariablesForDestructuredLambdaParameters(codegen, codegen.context.functionDescriptor.valueParameters)
+        initializeVariablesForDestructuredLambdaParameters(
```
ilmirus: Minor. Can you, please, remove default argument from `initializeVariablesForDestructuredLambdaParameters`, since this was the only place, which used it?

madsager: Yes! Done. :)

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
```diff
@@ -845,30 +844,7 @@ else if (kind == JvmMethodParameterKind.RECEIVER) {
             mv.visitLocalVariable(parameterName, type.getDescriptor(), null, methodBegin, methodEnd, shift);
             shift += type.getSize();
         }
-
         shift += shiftForDestructuringVariables;
```
max-kammerer: Could be deleted

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
```diff
@@ -46,6 +46,7 @@ abstract class AbstractCheckLocalVariablesTableTest : CodegenTestCase() {
             assertNotNull("Couldn't find class file for pattern $classFileRegex in: $pathsString", outputFile)
 
             val actualLocalVariables = readLocalVariable(ClassReader(outputFile.asByteArray()), methodName)
+            checkLocalVariableTypes(ClassReader(outputFile.asByteArray()), methodName, actualLocalVariables)
```
max-kammerer: Could we reuse class reader from previous line?

madsager: Yes, good idea. Done.

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
madsager: Yes! Done.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
```diff
@@ -120,13 +134,296 @@ abstract class AbstractCheckLocalVariablesTableTest : CodegenTestCase() {
...
+
+                            Type.DOUBLE_TYPE -> {
+                                localsAtEntry[parameterIndex++] = "D"
+                                // A double takes up two slots.
+                                parameterIndex++
+                            }
+                            Type.LONG_TYPE -> {
+                                localsAtEntry[parameterIndex++] = "J"
+                                // A double takes up two slots.
+                                parameterIndex++
```
max-kammerer: All increments could be changed to 'parameterIndex += it.size'

max-kammerer: and then localsAtEntry[...] could be pushed up from when

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
```diff
@@ -120,13 +134,296 @@ abstract class AbstractCheckLocalVariablesTableTest : CodegenTestCase() {
...
+            class BasicBlock(
+                val successors: MutableSet<BasicBlock> = mutableSetOf(),
+                var endsWithUnconditionalJump: Boolean = false,
+                var localsTable: MutableMap<Int, String> = mutableMapOf(),
+                var localsAtEntry: MutableMap<Int, String> = mutableMapOf(),
+                val localsInstructions: MutableList<Instruction> = mutableListOf()
+            ) {
+                fun addInstruction(index: Int, opcode: Int) {
+                    when (opcode) {
+                        Opcodes.ILOAD -> localsInstructions.add(Load(index, "I"))
```
max-kammerer: ...and same here:
```
localsInstructions.add(
when (opcode) {
     Opcodes.ILOAD -> Load(index, "I")
       ....
)
```

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
madsager: Done, thanks!

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
madsager: Done, thanks.

max-kammerer:
Thank you!

PR #2625, opened 2019-09-26
JVM IR: Hide constructors with inline class parameters

sfs:
On the JVM backend, we hide constructors taking inline class parameters and call them through public bridge methods. This PR implements the same feature in the JVM IR backend.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
@@ -5,18 +5,12 @@
 
 package org.jetbrains.kotlin.backend.common.lower
 
-import org.jetbrains.kotlin.backend.common.BackendContext
-import org.jetbrains.kotlin.backend.common.BodyLoweringPass
-import org.jetbrains.kotlin.backend.common.ClassLoweringPass
-import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
+import org.jetbrains.kotlin.backend.common.*
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.builders.irGet
 import org.jetbrains.kotlin.ir.builders.irSetField
-import org.jetbrains.kotlin.ir.declarations.IrClass
-import org.jetbrains.kotlin.ir.declarations.IrConstructor
-import org.jetbrains.kotlin.ir.declarations.IrFunction
-import org.jetbrains.kotlin.ir.declarations.IrValueParameter
+import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.*
 import org.jetbrains.kotlin.ir.symbols.IrConstructorSymbol
@@ -122,9 +116,9 @@ val innerClassConstructorCallsPhase = makeIrFilePhase(
     description = "Handle constructor calls for inner classes"
 )
 
-class InnerClassConstructorCallsLowering(val context: BackendContext) : BodyLoweringPass {
-    override fun lower(irBody: IrBody) {
-        irBody.transformChildrenVoid(object : IrElementTransformerVoid() {
+class InnerClassConstructorCallsLowering(val context: BackendContext) : FileLoweringPass {
+    override fun lower(irFile: IrFile) {
+        irFile.transformChildrenVoid(object : IrElementTransformerVoid() {
             override fun visitConstructorCall(expression: IrConstructorCall): IrExpression {
                 expression.transformChildrenVoid(this)
 
@@ -190,12 +184,11 @@ class InnerClassConstructorCallsLowering(val context: BackendContext) : BodyLowe
                 }
 
                 newReference.let {
+                    it.copyTypeArgumentsFrom(expression)
+                    // TODO: This is wrong, since we moved all parameters into value parameters,
+                    //       but changing it breaks JS IR in CallableReferenceLowering.
                     it.dispatchReceiver = expression.dispatchReceiver
                     it.extensionReceiver = expression.extensionReceiver
-                    for (t in 0 until expression.typeArgumentsCount) {
-                        it.putTypeArgument(t, expression.getTypeArgument(t))
-                    }
-
                     for (v in 0 until expression.valueArgumentsCount) {
                         it.putValueArgument(v, expression.getValueArgument(v))
                     }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
@@ -9,12 +9,15 @@ import org.jetbrains.kotlin.backend.common.lower.IrLoweringContext
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.JvmSymbols
 import org.jetbrains.kotlin.backend.jvm.codegen.isJvmInterface
+import org.jetbrains.kotlin.backend.jvm.lower.inlineclasses.hasMangledParameters
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.builders.IrBuilderWithScope
 import org.jetbrains.kotlin.ir.builders.Scope
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.descriptors.IrBuiltIns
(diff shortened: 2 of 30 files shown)


sfs:
Small update: I got rid of a hack to compute signatures in `CallableReferenceLowering`. Instead there is now an intrinsic with an IrFunctionReference argument to compute the signature during codegen. This is still not ideal, but at least a step in the right direction...

core/reflection.jvm/src/kotlin/reflect/jvm/internal/KDeclarationContainerImpl.kt
```diff
@@ -155,7 +155,7 @@ internal abstract class KDeclarationContainerImpl : ClassBasedDeclarationContain
 
         if (functions.size != 1) {
             val allMembers = members.joinToString("\n") { descriptor ->
-                DescriptorRenderer.DEBUG_TEXT.render(descriptor) + " | " + RuntimeTypeMapper.mapSignature(descriptor)
+                DescriptorRenderer.DEBUG_TEXT.render(descriptor) + " | " + RuntimeTypeMapper.mapSignature(descriptor).asString()
```
udalov: Thanks for fixing this. Please also fix the similar code above where `RuntimeTypeMapper.mapPropertySignature` is invoked

sfs: Thanks for pointing this out, I didn't notice the other code. Fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -49,6 +55,53 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
 
     override fun lower(irFile: IrFile) {
         inlineLambdaToCallSite.putAll(InlineReferenceLocator.scan(context, irFile).lambdaToCallSite)
+
+        // Unconditionally add bridges for hidden constructors
+        irFile.transformChildrenVoid(object: IrElementTransformerVoid() {
```
udalov: Why not override `visitConstructor` & `visitConstructorCall` in `SyntheticAccessorLowering` itself, instead of an extra object?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InnerClassesLowering.kt
```diff
@@ -190,12 +184,11 @@ class InnerClassConstructorCallsLowering(val context: BackendContext) : BodyLowe
                 }
 
                 newReference.let {
+                    it.copyTypeArgumentsFrom(expression)
+                    // TODO: This is wrong, since we moved all parameters into value parameters,
+                    //       but changing it breaks JS IR in CallableReferenceLowering.
```
udalov: @romanart FYI

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
```diff
@@ -375,14 +376,14 @@ internal class CallableReferenceLowering(private val context: JvmBackendContext)
...
-            body = context.createIrBuilder(symbol, startOffset, endOffset).run {
-                // TODO do not use descriptors
-                val declaration = codegenContext.referenceFunction(
-                    DescriptorUtils.unwrapFakeOverride(irFunctionReference.symbol.descriptor).original
-                ).owner
-                val method = codegenContext.methodSignatureMapper.mapAsmMethod(declaration)
-                irExprBody(irString(method.name + method.descriptor))
+            body = context.createJvmIrBuilder(symbol, startOffset, endOffset).run {
+                irExprBody(irCall(backendContext.ir.symbols.signatureStringIntrinsic).apply {
+                    putValueArgument(0, with(irFunctionReference) {
```
udalov: Maybe `irFunctionReference.deepCopyWithSymbols()`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -96,6 +99,12 @@ fun IrType.getArrayElementType(irBuiltIns: IrBuiltIns): IrType =
     else
         irBuiltIns.primitiveArrayElementTypes.getValue(this.classOrNull!!)
 
+val IrStatementOrigin?.isLambda
```
udalov: Would make sense to delete similar utility from `CallableReferenceLowering`

Also please add explicit return type

udalov: Also, this function can be used in `IrBlock.isInlineIrBlock` now

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -49,6 +55,53 @@ private class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElem
...
+            private val hiddenDeclarations = mutableSetOf<IrConstructor>()
+
+            private fun handleConstructor(declaration: IrConstructor): IrConstructorSymbol? {
```
udalov: Could you explain the change in this function in the commit "JVM IR: Don't compute signatures in CallableReferenceLowering"? I don't quite get why do we keep the `hiddenDeclarations` set, if `functionMap.getOrPut` already guarantees that we don't create more than one accessor per constructor

udalov:
@sfs Nice solution with the `SignatureString` intrinsic! I was thinking of removing generation of `getSignature`/`getName`/`getOwner` from the callable reference lowering and moving them to codegen altogether, since they aren't used from IR anyway and the correct IR-based implementation is not trivial. Maybe we'll do this one day, but for now the intrinsic approach would work really well already.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
sfs: Because we need to generate bridges unconditionally, even if there is no call to the constructor in the current module.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CallableReferenceLowering.kt
sfs: Fixed, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
sfs: I must have missed the one in `CallableReferenceLowering`, thanks for pointing it out! Also changed `isInlineIrBlock` to use this.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
sfs: Right, I refactored the code to reuse `functionMap` for this purpose - the problem is that we change the visibility of the constructor to private, after which they no longer have to be hidden.

However, if we assume that the `functionMap` is empty before we generate bridges for hidden constructors then we can use it to determine if we already generated a bridge.

sfs:
@udalov Thanks! I totally agree that it would be easier to generate the `getSignature`/`getName`/`getOwner` methods during codegen, though. In the long run it might be a good idea to build some infrastructure in codegen to handle more of these low-level JVM specific details.

sfs:
Updated to latest master and simplified the code around parameter annotations.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/RemoveDeclarationsThatWouldBeInlinedLowering.kt
```diff
@@ -36,8 +36,10 @@ private class RemoveDeclarationsThatWouldBeInlinedLowering(val context: JvmBacke
             override fun visitElement(element: IrElement) = element.acceptChildrenVoid(this)
 
             override fun visitFunctionReference(expression: IrFunctionReference) {
-                assert(expression.origin == IrStatementOrigin.LAMBDA || expression.origin == IrStatementOrigin.ANONYMOUS_FUNCTION)
-                loweredLambdasToDelete.add(expression.symbol.owner)
+                if (expression.origin.isLambda) {
```
udalov: Could you explain why did you replace `assert` with an `if`?

udalov: Oh right, I see. Thanks.

udalov:
Please also update test data for `SerializationIrBytecodeListingTestGenerated.testBasic`

sfs:
Rebased on master, updated the test and refactored the code in SyntheticAccessorLowering.

udalov:
Merged, thanks!

PR #2627, opened 2019-09-27
KT-17735 Simplify boolean expression produces incorrect result for floats comparison

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17735


zarechenskiy:
Hi @t-kameyama !

Regardless of the correctness, this change can produce incompatible change.
For example, consider the following code
```
const val a = -0.0 < 0.0

fun main() {
    println(a)
}
```

Currently, it prints `true` and after your changes, it'll print `false` (actually, it's [KT-24565](https://youtrack.jetbrains.com/issue/KT-24565)), so, unfortunately, we can't just accept your PR :(

The common problem is a bit bigger: we don't use IEEE754 comparison for constant evaluation and before start using it we should understand all places, which can change, then all such places need to be discussed with the Language committee (see more in https://kotlinlang.org/foundation/language-committee-guidelines.html) and only then we might introduce changes in the major update of the Kotlin compiler (1.4, for example)

So, I see two options:
- postpone with the changes until we fix this issue in the compiler
- fix this issue only on the IDEA side without fixing this in the compiler, so it'll be a totally compatible change

What do you think?

t-kameyama:
I will try to fix this issue only on the IDEA side.

t-kameyama:
@yanex 
I found 9f192ba85e29c46e3364080d6b985b5c95c1a5e5. Do you no longer need this pull request?

yanex:
@t-kameyama Well, yes. During internal discussions, we decided that turning off the inspection for this precise case is safer, considering that it's infrequent in real code. Thanks for your help anyway!

PR #2628, opened 2019-09-27
New script dependency resolvers

niktolik:
Added new script dependency resolvers that are supposed to replace legacy resolvers in kotlin-script-util. Main reason for new resolvers is to distinguish between two situations: when resolver is not suitable for artifact resolution (based on artifact syntax) and when resolution failed. When dependency resolution failed, a set of resolution attempts with tried addresses and failure reasons are returned in order to provide correct diagnostics for user.

### DIFF
--- a/libraries/scripting/dependencies/src/kotlin/script/experimental/DirectDependenciesResolver.kt
+++ b/libraries/scripting/dependencies/src/kotlin/script/experimental/DirectDependenciesResolver.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package kotlin.script.experimental
+
+import org.jetbrains.kotlin.script.util.resolvers.experimental.GenericArtifactCoordinates
+import org.jetbrains.kotlin.script.util.resolvers.experimental.GenericRepositoryCoordinates
+import java.io.File
+import java.lang.Exception
+
+class DirectDependenciesResolver : GenericDependenciesResolver {
+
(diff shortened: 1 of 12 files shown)


libraries/scripting/dependencies/src/kotlin/script/experimental/DirectDependenciesResolver.kt
```diff
@@ -0,0 +1,32 @@
...
+        return ResolveArtifactResult.Success(listOf(file))
+    }
+
+    override fun addRepository(repositoryCoordinates: GenericRepositoryCoordinates) =
+        throw Exception("DirectDependenciesResolver doesn't support adding repositories")
+
+    override fun accepts(repositoryCoordinates: GenericRepositoryCoordinates): Boolean = false
+
+    override fun accepts(artifactCoordinates: GenericArtifactCoordinates): Boolean =
+        !artifactCoordinates.string.isBlank() && !artifactCoordinates.string.contains(':')
```
ligee: Seems the second part of the condition will reject all full paths on Windows.

libraries/scripting/dependencies/src/kotlin/script/experimental/FlatLibDirectoryDependenciesResolver.kt
```diff
@@ -0,0 +1,68 @@
...
+                !file.isFile && !file.isDirectory -> resolveAttempts.add(ResolveAttemptFailure(file.canonicalPath, "Path is neither file nor directory"))
+                else -> return ResolveArtifactResult.Success(listOf(file))
+            }
+        }
+        return ResolveArtifactResult.Failure(resolveAttempts)
+    }
+
+    override fun accepts(artifactCoordinates: GenericArtifactCoordinates): Boolean {
+        return artifactCoordinates.string.takeUnless(String::isBlank)?.let { path ->
+            localRepos.any { File(it, path).exists() }
```
ligee: I think this is too strict. It is enough to check syntactic validity here.

PR #2631, opened 2019-09-30
Exclude additional generated testData/out-min dir in js/js.translator.

punzki:
This was causing re-indexing in IntelliJ after every time the JS tests are run.

(Credit goes to @sfs for finding the cause and fix! :clap:)

### DIFF
--- a/js/js.translator/build.gradle.kts
+++ b/js/js.translator/build.gradle.kts
@@ -28,6 +28,6 @@ sourceSets {
 
 configure<IdeaModel> {
     module {
-        excludeDirs = excludeDirs + files("testData/out-min")
+        excludeDirs = excludeDirs + files("testData/out", "testData/out-min")
     }
 }


bashor:
Hi! Sorry for delay!

It's strange that message mentions `testData/out-min`, but actually it was already excluded and commit excludes `testData/out`.

Just to be clear:
I think main source of pain in your case is `testData/out/irBox` because it contains a lot of JS code now. The reason why have preferred to not exclude out dirs is that it's very convenient to find out files by name and this IDE feature unfortunately doesn't work for excluded dirs.

I think we can apply this PR to take away your pain :), and later we'll try improve our cases in our side without returning known problems :)

bashor:
I've merged it w/o any change (Github usually asks to change commit message before final push, but in some reason it didn't in this case 🤷‍♂️)

punzki:
I see. Sorry if the commit message was confusing. It would be great if we could somehow have best of both worlds (being able to find files without causing re-indexing). Thanks for the review!

PR #2632, opened 2019-09-30
Avoid duplicate IR elements in ForLoopsLowering (accidentally introduced in #2620).

punzki:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -43,12 +44,16 @@ internal data class LoopReplacement(
...
     val inductionVariable: IrVariable,
-    val lastExpression: IrExpression,
+    lastExpression: IrExpression,
     val step: IrVariable,
     var loopVariable: IrVariable? = null,
     val isLastInclusive: Boolean,
     val declarations: List<IrVariable>
 ) {
+    val lastExpression: IrExpression = lastExpression
+        // Always copy `lastExpression` is it may be used in multiple conditions.
```
madsager: is it -> as it

madsager:
@homuroll This should fix the warnings you made us aware of yesterday.

max-kammerer:
Thanks!

PR #2636, opened 2019-10-01
Commonized metadata infrastructure between native, js and native ide.

alexander-gorshenev:


idea/idea-native/src/org/jetbrains/kotlin/ide/konan/NativePlatformKindResolution.kt
```diff
@@ -116,32 +149,34 @@ private fun createKotlinNativeBuiltIns(moduleInfo: ModuleInfo, projectContext: P
...
     val deserializationConfiguration = CompilerDeserializationConfiguration(languageSettings)
 
-    val libraryProto = konanLibrary.moduleHeaderData
+    val libraryProto = parseModuleHeader(konanLibrary.moduleHeaderData)
 
-    val stdlibFragmentProvider = KonanFactories.DefaultPackageFragmentsFactory.createPackageFragmentProvider(
+    val stdlibFragmentProvider = KonanFactories.DefaultDeserializedDescriptorFactory.createPackageFragmentProvider(
         konanLibrary,
-        null,
+        CachingIdeKonanLibraryMetadataLoader,
```
ddolovov: `CachingIdeKonanLibraryMetadataLoader` is no required here.

idea/idea-native/src/org/jetbrains/kotlin/ide/konan/NativePlatformKindResolution.kt
```diff
@@ -158,11 +193,7 @@ private fun ModuleInfo.findNativeStdlib(project: Project): NativeLibraryInfo? =
 
 class NativeLibraryInfo(project: Project, library: Library, root: File) : LibraryInfo(project, library) {
 
-    private val nativeLibrary = createKonanLibrary(
-        root,
-        KOTLIN_NATIVE_CURRENT_ABI_VERSION,
-        metadataReader = CachingIdeMetadataReaderImpl
-    )
+    private val nativeLibrary = createKotlinLibrary(root)
```
ddolovov: But here `CachingIdeKonanLibraryMetadataLoader` must be present in order to cache contents of any KLIB loaded in IDE.

idea/idea-native/src/org/jetbrains/kotlin/ide/konan/NativePlatformKindResolution.kt
```diff
@@ -183,6 +214,6 @@ class NativeLibraryInfo(project: Project, library: Library, root: File) : Librar
     override fun toString() = "Native" + super.toString()
 
     companion object {
-        val NATIVE_LIBRARY_CAPABILITY = ModuleDescriptor.Capability<KonanLibrary>("KonanLibrary")
+        val NATIVE_LIBRARY_CAPABILITY = ModuleDescriptor.Capability<KotlinLibrary>("KotlinLibrary")
```
ddolovov: Maybe keep "KotlinNativeLibrary" as the name of capability? "KotlinLibrary" can be misunderstood.

compiler/util-klib-metadata/src/org/jetbrains/kotlin/library/metadata/impl/KlibMetadataModuleDescriptorFactoryImpl.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.jetbrains.kotlin.library.metadata.parseModuleHeader
+import org.jetbrains.kotlin.library.unresolvedDependencies
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.resolve.CompilerDeserializationConfiguration
+import org.jetbrains.kotlin.serialization.deserialization.*
+import org.jetbrains.kotlin.storage.StorageManager
+
+val ModuleDescriptorImpl.isStdlibModule
+    get() = (this.klibModuleOrigin as? DeserializedKlibModuleOrigin)
```
ddolovov: This check may work improperly for user libs. Ex: a klib that user creates for their own needs.

alexander-gorshenev: In practivce all klibs created by konanc have at least stdlib in the depends clause. On the other hand, js's stdlib doesn't have 'stdlib' name. So the above is the least common denominator until we have a common policy.

ddolovov: OK

PR #2610, opened 2019-09-20
Use `getPropertyGetter()` in ForLoopsLowering to retrieve property getters

punzki:
...getters instead of directly retrieving the property first. When the IR backend is used to compile the standard library, the progression classes (in sources) are lowered, and therefore do not have properties anymore (only fields and functions).

PR #2637, opened 2019-10-01
JVM_IR: fix const reads from other files in same module

pyos:
If the property happens to be lowered before the call is, then `property.getter` is null and the comparison fails.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/ConstLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/ConstLowering.kt
@@ -41,8 +41,7 @@ class ConstLowering(val context: CommonBackendContext) : IrElementTransformerVoi
     override fun visitCall(expression: IrCall): IrExpression {
         val function = (expression.symbol.owner as? IrSimpleFunction) ?: return super.visitCall(expression)
         val property = function.correspondingPropertySymbol?.owner ?: return super.visitCall(expression)
-        if (function != property.getter)
-            return super.visitCall(expression)
+        // If `constantValue` is not null, `function` can only be the getter because the property is immutable.
         return property.backingField?.constantValue() ?: super.visitCall(expression)
     }
 
--- a/compiler/testData/codegen/box/properties/const/anotherFile.kt
+++ b/compiler/testData/codegen/box/properties/const/anotherFile.kt
@@ -1,0 +1,7 @@
+// FILE: 1.kt
+// File names are important! This file should come before the other one
+// in a lexicographic order.
+const val x = "OK"
+
+// FILE: 2.kt
+fun box() = x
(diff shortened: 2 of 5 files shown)


compiler/testData/codegen/box/properties/const/anotherFile.kt
```diff
@@ -0,0 +1,7 @@
+// FILE: 1.kt
+// File names are important! This file should come before the other one
```
pyos: There are also other tests that fail with different orderings, e.g.

```
// FILE: 1.kt
fun box() = A.f()
// FILE: 2.kt
object A {
    @JvmStatic
    fun f(x: String = "OK") = x
}
```

If having all lowerings do the right thing on both lowered and unlowered inputs is still a goal, perhaps all multi-file tests should run twice (once with the original file names, and once with files renamed to be in reverse order)?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/ConstLowering.kt
```diff
@@ -30,7 +30,7 @@ fun IrField.constantValue(implicitConst: Boolean = false): IrConst<*>? {
     // Kotlin fields are only inlined if explicitly `const` to avoid making the values part of the ABI.
     val inline = correspondingPropertySymbol?.owner?.isConst == true ||
             (isStatic && isFinal && (origin == IrDeclarationOrigin.IR_EXTERNAL_JAVA_DECLARATION_STUB || implicitConst))
-    return if (inline) initializer?.expression as? IrConst<*> else null
+    return if (inline) (initializer?.expression as? IrConst<*>)?.copy() else null
```
max-kammerer: Is it required for ir-validation? Or is it another reason?

pyos: Yeah, that's for validation.

>NB: maybe it's worth to rewrite using #2610 approach?

Not sure what you mean, as in that PR the task is to locate a property by name while here we already have the property/field.

max-kammerer: Than my question doesn't matter

pyos: Rolled back to avoid conflicts with #2640

compiler/testData/codegen/box/properties/const/anotherFile.kt
udalov: I like this idea. My only concern is that it'd increase the IR tests run time by a few hundred test cases, where majority of those cases supposedly don't test anything useful. Maybe an explicit directive in the test source would be a better approach

pyos:
ping?

udalov:
@pyos Sorry for delay, was on vacation. Merged manually, thank you!

PR #2638, opened 2019-10-01
JVM IR: Use WrappedClassDescriptor in class builder.

sfs:
We don't lower descriptors in the JVM IR backend, but we still use them for ClassBuilders. This causes a spurious diagnostic related to inline classes which breaks one test.

This PR changes ClassCodegen to pass a WrappedDescriptor to the ClassBuilder instead.


max-kammerer:
Thank you!

PR #2640, opened 2019-10-01
[JVM IR] Enable IR validation and fix all duplication errors

punzki:
@gogabr - I cherry-picked your change into my branch. Thanks for sharing that! I had to disable the IR validation _before_ lowering because it fails one of the FIR tests (`org.jetbrains.kotlin.cli.CliTestGenerated$Jvm.testFirHello`).


gogabr:
Re-enabled IR validation at the start and committed.
(I will mute the failing test and assign it to people responsible for FIR).
Thank you!

PR #2641, opened 2019-10-02
KT-18536 Provide proper quick fix for accidental override error

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18536

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/MakePrivateAndOverrideMemberFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/MakePrivateAndOverrideMemberFix.kt
@@ -7,28 +7,40 @@ package org.jetbrains.kotlin.idea.quickfix
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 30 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/MakePrivateAndOverrideMemberFix.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.kotlin.lexer.KtTokens
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.containingClassOrObject
+import org.jetbrains.kotlin.resolve.jvm.diagnostics.ErrorsJvm
+
+class MakePrivateAndOverrideMemberFix(
+    propertyOrParameter: KtDeclaration,
+    private val memberNameToOverride: String,
+    private val makePrivate: Boolean,
+    private val implement: Boolean
```
dimonchik0036: I'll be prefer to replace these arguments with one `private val memberObject: OverrideMemberChooserObject`.

PR #2642, opened 2019-10-02
KT-7940 J2K: convert Integer.MAX_VALUE to Int.MAX_VALUE

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-7940


darthorimar:
Thank you ^_^

PR #2643, opened 2019-10-02
JVM_IR inliner fixes: MethodInliner and object transformers

pyos:
Part 1 of #2555


pyos:
Rebased on top of #2646 and unmutedd 57 more tests. (Another 16 require #2666.)

max-kammerer:
> Not sure what Steven's change has to do with this, as CallableReferenceLowering completely ignores inline lambdas. If you're planning on generating classes for them, the situation will become even worse as classes can't have reified type parameters.

Another possible approach to keep IrFunctionExpression as long as possible (now it's unrolled into function+function reference in provisionalFunctionExpressionPhase): seems it will allow to keep reified parameters as long as possible. But this approach conflicts with logic in  inlineCallableReferenceToLambdaPhase...

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -62,14 +62,11 @@ open class FunctionCodegen(
...
         //        since otherwise anonymous object regeneration can fail with an ArrayIndexOutOfBounds exception if the number
-        //        or arguments to the constructor changes.
-        if (!hasSyntheticFlag ||
-            irFunction.origin == JvmLoweredDeclarationOrigin.SYNTHETIC_METHOD_FOR_PROPERTY_ANNOTATIONS ||
-            //TODO: investigate this case: annotation here is generated twice in lowered function and in interface method overload
-            irFunction.origin == JvmLoweredDeclarationOrigin.GENERATED_SAM_IMPLEMENTATION
-        ) {
+        //        of arguments to the constructor changes. This hack can be removed after at least one version is released
+        //        that includes `ShiftParameterAnnotationsVisitor`.
+        if (irFunction !is IrConstructor || !(hasSyntheticFlag || irFunction.parentAsClass.isAnonymousObject)) {
```
max-kammerer: @pyos Actually there shouldn't be any annotation on anonymous object constructor, so looks like after adding `irFunction.parentAsClass.isAnonymousObject` all other changes in first 'Shift parameter annotations on constructors of local classes pyos' commit are redundant  (sorry if I missed smth in previous pr discussions)

pyos: Yeah, now that I think about it, it really doesn't make sense to support them in the inliner.

max-kammerer: As this commit unmutes a bunch of tests I could merge it if you extract it from this pr

pyos: Extracted to #2681

pyos:
@max-kammerer any progress reviewing this PR?

max-kammerer:
@pyos Sorry for delay, there was a bunch of other work last week. Will try to finish on this week

max-kammerer:
Wrote comment for "Do not regenerate anonymous objects in receiverless lambdas" directly

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -496,11 +496,11 @@ class AnonymousObjectTransformer(
...
         val capturedLambdasToInline = HashMap<String, LambdaInfo>() //captured var of inlined parameter
         val allRecapturedParameters = ArrayList<CapturedParamDesc>()
-        val addCapturedNotAddOuter =
-            parentFieldRemapper.isRoot || parentFieldRemapper is InlinedLambdaRemapper && parentFieldRemapper.parent!!.isRoot
+        val addCapturedNotAddOuter = capturedLambdas.all { !it.hasDispatchReceiver } ||
+                parentFieldRemapper.isRoot || parentFieldRemapper is InlinedLambdaRemapper && parentFieldRemapper.parent!!.isRoot
         val alreadyAdded = HashMap<String, CapturedParamInfo>()
         for (info in capturedLambdas) {
-            if (addCapturedNotAddOuter) {
+            if (!info.hasDispatchReceiver || addCapturedNotAddOuter) {
```
max-kammerer: Seems next condition with !addCapturedNotAddOuter should be also updated. Wrote about new semantic flag privately

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
```diff
@@ -413,9 +413,18 @@ class MethodInliner(
 
         val capturedParamsSize = parameters.capturedParametersSizeOnStack
         val realParametersSize = parameters.realParametersSizeOnStack
+        val newArgumentList = if (!inliningContext.isInliningIrLambda) {
+            // Add captured parameters at the end to be used instead of the receiver.
+            Type.getArgumentTypes(node.desc) + parameters.capturedTypes
+        } else {
+            // Already have parameters for captured values, but they are actually at the beginning.
+            // The order of arguments in IR inline lambdas is extension?, captured*, real*;
```
max-kammerer: I would prefer to change order of parameters in IR because 'extension?, captured*, real*' looks too weird

max-kammerer:
Could you clarify a bit 'Reset captured parameter indices when copying descriptions' commit it's required cause of different order of parameters or smth else?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
pyos: The problem is that real parameters can have default values:

```kotlin
fun f(x: Int) {
    fun g(y: Int = x) = y
    return g()
}
```

->

```kotlin
fun `f$g`(`$x`: Int, y: Int = `$x`) = y
fun f(x: Int) = `f$g`(x)
```

Swapping the order of arguments would result in invalid IR.

pyos:
^ split into separate PRs with improvements.

PR #2646, opened 2019-10-02
JVM_IR inliner fixes: codegen

pyos:
Part 3 of #2555


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -152,23 +147,14 @@ class IrExpressionLambdaImpl(
     }
 
     companion object {
-        private var counter: Int = 123//TODO: pass proper type
+        private var counter: Int = 1//TODO: pass proper type
     }
 
-    override val lambdaClassType: Type = Type.getObjectType("test${counter++}")
+    override val lambdaClassType: Type = Type.getObjectType("$typePrefix\$lambda-${counter++}")
```
max-kammerer: Maybe best to use precalculated old name ```codegen.context.getLocalClassInfo(reference)```

pyos: There still needs to be a fallback in case the reference was generated by a lowering after InventNamesForLocalClasses (e.g. by InlineCallableReferenceToLambdaPhase) and has no LocalClassInfo. Since this name never appears in the output and any unique string is fine, IMO it's not worth the effort.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -82,6 +84,11 @@ open class ClassCodegen protected constructor(
             else -> null
         }
 
+    fun getRegeneratedObjectNameGenerator(function: Name) =
+        context.regeneratedObjectNameGenerators.getOrPut(irClass to function) {
```
max-kammerer: Actually it's possible to keep name generator in ClassCodegen with class related prefix (it would be more local) and for other purposes create subgenerator via NameGenerator.subGenerator

pyos: That's unfortunately what the new test (boxInline/lambdaTransformation/regeneratedLambdaName2.kt) is about. Since currently IrSourceCompilerForInline creates a new (Fake)ClassCodegen for every call to an inline function for which source code is available, caching name generators at ClassCodegen level would cause the second call to `sameName` to overwrite the anonymous object created by the first one with a different implementation.

pyos: Added a TODO about this.

max-kammerer:
@pyos Actually I need to clarify situation with assertions in local/private classes internally, so it's best to extract first commit if possible to separate pr to avoid postponing of other changes.
NB: #2644 was merged to master

pyos:
This commit is kind of necessary as otherwise the inliner cannot regenerate the constructor (due to lambdas being passed to a non-inline function).

>NB: #2644 was merged to master

Yes, thanks, I've already rebased the other two on top of that and unmuted a few more tests.

pyos:
In fact, since it's possible to take a reference to a local named type's constructor and pass it to Java code as a KFunction, that's probably exactly what I should do. Thanks for noticing!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
max-kammerer: Actually seems it's only InlineCallableReferenceToLambdaPhase lower that was writen especcially for inline purpose. It could be simply patched in 2 places via:
```
+IrFunctionReferenceImpl(...).also {
                        this@InlineCallableReferenceToLambdaPhase.context.putLocalClassInfo(
                            it,
                            this@InlineCallableReferenceToLambdaPhase.context.getLocalClassInfo(expression)!!
                        )
                    }
```
It's worth for debugging purpose in case of two backend support

max-kammerer: Seems best to use: ```it.copyAttributes(expression)```

pyos: Done

max-kammerer:
> In fact, since it's possible to take a reference to a local named type's constructor and pass it to Java code as a KFunction, that's probably exactly what I should do. Thanks for noticing!

Could you please add test for such case?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -196,15 +196,18 @@ class ExpressionCodegen(
         if (state.isParamAssertionsDisabled)
             return
 
-        val isSyntheticOrBridge = irFunction.origin.isSynthetic ||
+        val notCallableFromJava = isInlineLambda ||
+                Visibilities.isPrivate(irFunction.visibility) ||
+                irFunction.origin.isSynthetic ||
+                (irFunction is IrConstructor && irFunction.parentAsClass.isAnonymousObject) ||
```
max-kammerer: As I understand there would be similar problem with redundant assertions on captured variables in other cases,e.g.:
```
fun foo() {
    val captured = "123"
    
    class Z(val withAssertion: String) {
           
        val x = captured
    }
    
    Z("12")
}
```
So maybe it's best to specify special (synthetic?) kind for captured variables and variables passed to super constructor and don't generate for such variable assertions

pyos: The important part here is not the assertion itself, which is never triggered but still valid code, but how the inliner handles it. Specifically, when the captured variable is an inline lambda, the load of that lambda before the null-checking intrinsic causes the variable remapper to throw an exception. Since local named classes in inline functions are currently rejected by the frontend, only the anonymous object case really needs to be handled right now.

pyos:
>Could you please add test for such case?

Done; however, it turns out this code doesn't actually matter as the generated reference class has an assertion in its invoke method.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -82,6 +83,13 @@ open class ClassCodegen protected constructor(
             else -> null
         }
 
+    fun getRegeneratedObjectNameGenerator(function: IrFunction): NameGenerator {
+        val name = function.name.takeIf { !it.isSpecial }
```
pyos: This should fix tests on Windows.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: Could you add TODO about possible further logic generalization for mentioned cases?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendContext.kt
```diff
@@ -71,6 +73,13 @@ class JvmBackendContext(
         localClassInfo[container.attributeOwnerId] = value
     }
 
+    // TODO cache these at ClassCodegen level. Currently, sharing this map between classes in a module is required
+    //      because IrSourceCompilerForInline constructs a new (Fake)ClassCodegen for every call to
+    //      an inline function in the same module. Thus, if two inline functions happen to have the same name
+    //      and call a third inline function that has an anonymous object, the one which is called last
+    //      will overwrite the other's regenerated copy. (Or don't recompile the inline function for every call.)
+    internal val regeneratedObjectNameGenerators = mutableMapOf<Pair<IrClass, Name?>, NameGenerator>()
```
max-kammerer: Would such logic stable in case of incremental compilation?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
max-kammerer: maybe smth like val name = function.name.takeIf { !it.isSpecial } ?: "_special_"

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -70,39 +71,39 @@ class IrSourceCompilerForInline(
...
         asmMethod: Method
     ): SMAPAndMethodNode {
-        assert(callableDescriptor == callElement.descriptor.original) { "Expected $callableDescriptor got ${callElement.descriptor.original}" }
+        assert(callableDescriptor == callee.descriptor.original) { "Expected $callableDescriptor got ${callee.descriptor.original}" }
         assert(codegen.lastLineNumber >= 0) { "lastLineNumber shall be not negative, but is ${codegen.lastLineNumber}" }
 
-        val irFunction = getFunctionToInline(callElement as IrCall, jvmSignature, callDefault)
-        return makeInlineNode(irFunction, FakeClassCodegen(irFunction, codegen.classCodegen), CallSiteMarker(codegen.lastLineNumber))
+        val irFunction = getFunctionToInline(jvmSignature, callDefault)
+        lazySourceMapper.callSiteMarker = CallSiteMarker(codegen.lastLineNumber)
```
max-kammerer: Intresting catch: could we move marker set/reset logic a bit  further (similar to old backend logic)?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -529,7 +538,7 @@ class ExpressionCodegen(
         )
 
     override fun visitClass(declaration: IrClass, data: BlockInfo): PromisedValue {
-        classCodegen.generateLocalClass(declaration, irFunction.isInline).also {
+        classCodegen.generateLocalClass(declaration, generateSequence(this) { it.inlinedInto }.any { it.irFunction.isInline }).also {
```
max-kammerer: Please note there is also withinInline check inside generateLocalClass, maybe it's a bug with it?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendContext.kt
pyos: I believe so. If the file containing the overloaded inline functions (`sameName` in the test case) is recompiled, then so should be all the use sites; all regenerated classes will be created again with the same names as in a full compilation scenario. If the file is not recompiled, then call sites will read bytecode for the inline functions from the class file instead of going through FakeClassCodegen, and no potentionally clashing regenerations happen in the first place.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: No, this fix is for cases like this:

```kotlin
inline fun f() {
    someOtherInlineFunction {
        object : { ... }
    }
}
```

Here we have a ClassCodegen for the file class, then a FunctionCodegen+ExpressionCodegen for `f`, then finally a FunctionCodegen+ExpressionCodegen for the lambda which calls `classCodegen.generateLocalClass`. There are no ClassCodegens between the two ExpressionCodegens.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
pyos: done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
pyos: I'm actually not sure what this marker is for at all, as generating code for an inline function through FakeClassCodegen shouldn't touch lazySourceMapper, i.e. the current class' SMAP, at all. Removing these two lines doesn't seem to break any tests.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: done

pyos:
Rebased on top of #2655 and unmuted 3 more tests

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: As I see it should be ClassCodegen + FunctionCodegen+ExpressionCodegen(1) for 'f' than same for lambda(2) and anonymous object(3). and withinInline flag should be passed from ExpressionCodegen(1) to ClassCodegen(2) to FC(2) to EC(2) and finally to CC(3)

max-kammerer: Oh seems it's cause of change and a bit new semantic of 
```val isInlineLambda: Boolean = false``` -> ```val inlinedInto: ExpressionCodegen```
will look more deeply

File: compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt [hunk 8]
```diff
@@ -72,32 +73,24 @@ class IrInlineCodegen(
                 activeLambda = null
             }
         } else {
-            putValueOnStack(argumentExpression, parameterType, irValueParameter.index, blockInfo, irValueParameter.type)
+            // Reusing the same local reduces the amount of debug information available when breaking
+            // inside the inline function. Here we replicate the behavior of the non-IR backend, which at least
+            // preserves the receivers.
+            val onStack = if (irValueParameter.index >= 0)
```
max-kammerer: @pyos Could you clarify a bit this line and comment for it 'which at least preserves the receivers.': 
as i remember receivers have negative index so 'if (irValueParameter.index >= 0)' looks wrong

max-kammerer: ...Or is it assumes that all receivers are materialized on stack?

max-kammerer: If so then 'if (irValueParameter.index >= 0)' check could be removed: it was impossible to make such optimization in old backend cause generally speaking both receivers (dispatch+extension) were incapsulated in one StackValue/or materialized before inline stuff

pyos: The condition is correct: if index < 0 (=> receiver) we materialize the value (== call `gen`), otherwise we ask ExpressionCodegen to return an unmaterialized `StackValue.local` when possible. It's OK to always use `genOrGetLocal` here from the correctness perspective, yes, although it would cause the debugger to not show there receiver when stopped at a breakpoint in an inline method, and also change the output of some bytecode text tests.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -1001,7 +1012,26 @@ class ExpressionCodegen(
...
+        //     object {
+        //         inline fun <reified T : Any> f() = run { T::class.java.getName() }
+        //     }
+        //
+        // LocalDeclarationsLowering would extract that lambda into a method of the enclosing type, but will not create
+        // a reified type parameter in it (in fact, the lambda method isn't even marked as inline):
+        //
+        //     object {
+        //         /* static */ private fun `f$lambda-0`() = T::class.java.getName()
+        //         inline fun <reified T : Any> f() = run(::`f$lambda-0`)
```
max-kammerer: @pyos Could you extract commit with this change ("JVM_IR: "fix" reified type parameter scope check" and linked one  "JVM_IR: generate SMAPs for classes local to inline lambdas...") to separate pr as I need more time for investigation.
Actually we have back logic in lowering made by @sfs  that merges f$lambda-0 fun into invoke method of lambda class/function reference, so maybe problem could be easier hacked here on class construction: https://github.com/JetBrains/kotlin/commit/c77fa1ecd6040be2dadc1d66e9013aa2aa49f594#diff-852213ef8c0292f5b7339ac0b276eab3R232

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
max-kammerer: Than it's ok, more precise tuning and debugger support could be made late. Could you just add comment about additional optimization?

pyos: Rewrote the comment so that it is hopefully more clear

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: Removed the commits from this branch. I'll create a separate PR later as there's a small conflict in ExpressionCodegen in a place that's also touchedby "JVM_IR: omit nullability assertions in anonymous object constructors".

Not sure what Steven's change has to do with this, as CallableReferenceLowering completely ignores inline lambdas. If you're planning on generating classes for them, the situation will become even worse as classes can't have reified type parameters.

max-kammerer:
Thank you! Great work!

PR #2649, opened 2019-10-02
The friend module identification has regressed recently

alexander-gorshenev:


bashor:
@alexander-gorshenev @skuzmich is it merged?

PR #2654, opened 2019-10-04
JVM_IR: omit annotations on outer this parameters

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -207,8 +207,6 @@ open class FunctionCodegen(
                 } else {
                     frameMap.enterTemp(parameter.asmType)
                 }
-            } else if (parameter.kind != JvmMethodParameterKind.VALUE) {
```
max-kammerer: Please add assertion that there are only VALUE or RECEIVER kinds

max-kammerer:
Could you please add some description to commit to clarify made changes

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
pyos: The point of removing the branch is precisely that there's also OUTER now, but it corresponds to an IR value parameter like VALUE does and therefore does not need a call to `enterTemp`. If I'm correct in assuming that the IR type mapper is supposed to be lightweight and should not insert any parameters that were not present in the IR in the first place, "RECEIVER" and "everything else" are in fact the only cases that should ever be differentiated.

max-kammerer: > If I'm correct in assuming that the IR type mapper is supposed to be lightweight and should not insert any parameters that were not present in the IR in the first place, "RECEIVER" and "everything else" are in fact the only cases that should ever be differentiated.

@udalov please comment

udalov: Yes, `IrTypeMapper` & `MethodSignatureMapper` are a part of the codegen (technically not yet, but I _hope_ we'll get rid of their usages in lowerings in the future), and the general desire is to make codegen as simple as possible, moving all the complexity to lowerings. Currently `MethodSignatureMapper.mapSignature` only produces `RECEIVER` and `VALUE` parameter kinds, and it'd be even nicer to remove this distinction some day too.

pyos: Good point, done.

max-kammerer: Looks like change in MethodSignatureMapper is now redundant, isn't it?

pyos: That change is to make the `kind.isSkippedInGenericSignature` checks in `generateParameterAnnotations` detect outer `this`, so it's actually the main point of the PR. Now that I think about it, it's probably better to just replace these checks with `origin == JvmLoweredDeclarationOrigin.FIELD_FOR_OUTER_THIS`.

pyos: On second thought, this also affects the flags generated for this parameter, so OUTER is more correct for now.

max-kammerer: Sorry, but code looks a bit tricky for me cause OUTER recognized as `VALUE` parameter in `IrCallableMethod.valueParameterTypes` and `RECEIVER` is not (yet). Maybe it's possible to avoid usage of `jvmSignature` in `generateParameterAnnotations`?

pyos: >code looks a bit tricky for me cause OUTER recognized as VALUE parameter in IrCallableMethod.valueParameterTypes and RECEIVER is not (yet)

That's because RECEIVER, corresponds to `irFunction.extensionReceiverParameter` while everything else is in `irFunction.valueParameters`, and that's what `valueParameterTypes corresponds to. Or, at least, that's how it's used in ExpressionCodegen.

>Maybe it's possible to avoid usage of jvmSignature in generateParameterAnnotations?

Sure, but the parameter should still be OUTER to generate the correct Signature attribute in the bytecode.

max-kammerer: @pyos Thank you for clarification,  I will merge as is

max-kammerer:
Many thanks!

PR #2655, opened 2019-10-04
Generate unique parameter names in LocalDeclarationsLowering

pyos:


max-kammerer:
@pyos 

1. Could you add simple box test based on next snippet:
```
fun String.foo() {    
    fun String.bar(): String {
       return {
            this@foo + this@bar
        }()
    }    
   ....
} 
```

2. As this clash could be occurred only for captured receivers could you use function name as suffix?

pyos:
Added your test, and also the case from okhttp that actually made me look at this code:

```kotlin
inline fun f(g: () -> String) = g()

class A(val x: String) {
    inner class B(val y: String) {
        fun h() = f { x + y } // captures `this@B` and `this@A`
    }
}
```

With #2646 (specifically the commit that changes how `IrExpressionLambdaImpl.capturedVars` is constructed), this test fails because of two `CapturedDesc`s with the same name `$this`: `y` is incorrectly retrieved from an instance of A instead of A$B.

>As this clash could be occurred only for captured receivers could you use function name as suffix: $foo or $bar?

Could you give me an example of what names the parameters in the above test should have, which only has one function? (Maybe this is a separate bug in LocalDeclarationsLowering and `x` should be transformed into `<captured B>.this$0.x` instead of `<captured A>.x`?)

max-kammerer:
@pyos Oh yes, old backend works as you described <captured B>.this$0.x, but new one is not. And for other reasons I think that we should support old capturing scheme in JVM_IR backend (not sure that it would be accepted in other backends).
So for now next naming scheme should solve problem:  basename + $function/or class name + $index, shouldn't it?

max-kammerer:
@pyos Seems that your test should fail even with non-inline function `f`:
```fun f(g: () -> String) = g()```

so it could be made as simple box

pyos:
The test works with non-inline `f` because the JVM does not care about name conflicts so long as descriptors are different: `$this : A` is not the same as `$this : A$B`. It's the inliner that incorrectly maps them to the same local.

pyos:
Added the parent's name to `$this` parameters. This results in `$this$A` and `$this$h` in my test, which I guess is close enough.

max-kammerer:
Many thanks!

PR #2656, opened 2019-10-04
JVM IR: Respect InlineConstVals

sfs:
I'm honestly not sure if this feature is relevant for the IR backends - as far as I can tell, this is merely for compatibility with Kotlin before version 1.1


max-kammerer:
@ Thank you!
I think it worth!

PR #2657, opened 2019-10-04
JVM IR: Lower IrStringConcatenation

sfs:
- Add a lowering to replace `IrStringConcatenation` with string builders instead of handling this in `ExpressionCodegen`.
- Add some missing optimizations from the JVM backend (inline classes shouldn't box in string templates).
- Fix constant folding of `toString` with unsigned arguments.
- Unify the handling to `toString`, by turning direct calls to `toString` into `IrStringConcatenations` (and ensuring that there is no special handling for super calls).


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
```diff
@@ -76,21 +77,42 @@ class FlattenStringConcatenationLowering(val context: CommonBackendContext) : Fi
...
-                            expression.valueArgumentsCount == 1 &&
-                            function.name == PLUS_NAME &&
-                            function.fqNameWhenAvailable?.parent() in PARENT_NAMES
+                    when (function.name.asString()) {
+                        "toString" -> {
+                            // Check that this is a call to method Any.toString or to the extension function
+                            // Any?.toString defined in the kotlin package.
+                            expression.valueArgumentsCount == 0
+                                    && ((expression.dispatchReceiver != null
+                                    && (expression.type.isAny() || (function as IrSimpleFunction).overriddenSymbols.isNotEmpty()))
```
udalov: Why `isAny()` and not `isString()` here?

Also, I think we need to check that extension receiver is null in this case. Overall, consider splitting this to two functions for easier reading

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
```diff
@@ -0,0 +1,146 @@
...
+            putValueArgument(0, argument)
+        }
+    }
+
+    private fun JvmIrBuilder.lowerInlineClassArgument(expression: IrExpression): IrExpression {
+        if (expression.type.unboxInlineClass() == expression.type)
+            return expression
+
+        val toStringFunction = expression.type.classOrNull?.owner?.functions?.singleOrNull {
+            it.name.asString() == "toString" && it.dispatchReceiverParameter != null
```
udalov: Maybe either use the `nameToString` value here as well, or inline it

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.jetbrains.kotlin.name.Name
+
+internal val jvmStringConcatenationLowering = makeIrFilePhase(
+    ::JvmStringConcatenationLowering,
+    name = "StringConcatenation",
+    description = "Replace IrStringConcatenation with string builders"
+)
+
+/**
+ * This lowering pass replaces [IrStringConcatenation]s with StringBuilder appends.
```
udalov: Please replace or supplement this comment with the note that this is based on the common `StringConcatenationLowering`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
udalov: Also, there's similar code in `FlattenStringConcatenationLowering`, would be nice to extract to a utility in backend.common

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -427,6 +434,36 @@ class JvmSymbols(
     }
 
     val systemArraycopy: IrSimpleFunctionSymbol = systemClass.functionByName("arraycopy")
+
+    val javaLangString: IrClassSymbol = createClass(FqName("java.lang.String")) { klass ->
```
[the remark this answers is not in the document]
sfs: Done, although this might be worse in terms of performance, since we wouldn't normally deserialize `java.lang.String` and it's fairly large.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/FlattenStringConcatenationLowering.kt
sfs: That was supposed to be a test on the receiver parameter, fixed! I also split the function to make it easier to read.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private val defaultAppendFunction = stringBuilder.functions.single {
+        it.name == nameAppend &&
+                it.valueParameters.size == 1 &&
+                it.valueParameters.single().type.isNullableAny()
+    }
+
+    private val appendFunctions: Map<IrType, IrSimpleFunction?> =
+        typesWithSpecialAppendFunction.map { type ->
```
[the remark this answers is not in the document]
sfs: Thanks, fixed!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
sfs: Done!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
sfs: Done!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
```diff
@@ -0,0 +1,165 @@
...
+    private fun typeToAppendFunction(type: IrType): IrSimpleFunction =
+        appendFunctions[type] ?: defaultAppendFunction
+
+    private val defaultValueOfFunction = context.ir.symbols.javaLangString.owner.functions.single {
+        it.name == nameValueOf && it.valueParameters.singleOrNull()?.type?.isNullableAny() == true
+    }
+
+    private val valueOfFunctions: Map<IrType, IrSimpleFunction?> =
+        typesWithSpecialStringFunction.associateWith { type ->
+            context.ir.symbols.javaLangString.owner.functions.singleOrNull {
```
udalov: Currently we'll iterate over all `java.lang.String` members (of which there are ~100) 9 times for each source file. Maybe we could do this traversal only once per compilation? Basically store this info somewhere in `JvmSymbols`, or `JvmBackendContext`

sfs: Done! I've moved the code in question to JvmSymbols.

udalov:
Merged manually. Thanks once more!

PR #2658, opened 2019-10-07
KT-33123 False positive "Redundant qualifier name" with inner class as constructor parameter for outer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33123


idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantQualifierNameInspection.kt
```diff
@@ -126,6 +128,16 @@ private fun KtUserType.applicableExpression(context: BindingContext): KtUserType
...
     val originalDescriptor = referenceExpression.mainReference.resolveToDescriptors(context).firstOrNull() ?: return null
 
+    if (originalDescriptor is ClassDescriptor && originalDescriptor.isInner) {
+        val outerClassDescriptor = originalDescriptor.containingDeclaration as? ClassDescriptor
+        if (outerClassDescriptor?.typeConstructor != null) {
+            if (getStrictParentOfType<KtPrimaryConstructor>() != null
+                || getStrictParentOfType<KtSuperTypeEntry>() != null
+                || containingClass()?.descriptor != outerClassDescriptor
+            ) return null
+        }
```
dimonchik0036: I'd prefer replace this with `if (outerClassDescriptor?.typeConstructor != null) return null`.
Support inner classes are better left for later.

PR #2660, opened 2019-10-07
KT-34203 'Add constructor parameter' fix does not add generics

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34203


mglukhikh:
Merged 👍

PR #2661, opened 2019-10-07
[JVM_IR] Map suspend lambdas to original function for inlining.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -153,7 +154,13 @@ class IrSourceCompilerForInline(
         get() = callElement.descriptor as FunctionDescriptor
 
     override fun getContextLabels(): Set<String> {
-        return setOf(codegen.irFunction.name.asString())
+        val name = codegen.irFunction.name.asString()
+        if (name == INVOKE_SUSPEND_METHOD_NAME) {
```
madsager: Should this check for invoke as well as invokeSuspend?

madsager:
@ilmirus

ilmirus:
Thanks!

PR #2663, opened 2019-10-07
KT-34218 "Merge 'else if' intention"

DmiitriyJarosh:
https://youtrack.jetbrains.com/issue/KT-34218

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.idea.intentions
+
+import com.intellij.openapi.editor.Editor
+import org.jetbrains.kotlin.psi.*
+
+
+class MergeElseIfIntention : SelfTargetingIntention<KtIfExpression>(KtIfExpression::class.java, "Merge 'else if'") {
+    override fun isApplicableTo(element: KtIfExpression, caretOffset: Int): Boolean {
+        val elseBody = element.`else` ?: return false
+        val nestedIf = elseBody.nestedIf() ?: return false
+        if (nestedIf.`else` != null) {
+            return false
+        }
+        return true
+    }
+
+    override fun applyTo(element: KtIfExpression, editor: Editor?) {
+        applyTo(element)
+    }
+
+    companion object {
+        fun applyTo(element: KtIfExpression): Int {
+            val nestedIf = element.`else`?.nestedIf() ?: return -1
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import com.intellij.openapi.editor.Editor
+import org.jetbrains.kotlin.psi.*
+
+
+class MergeElseIfIntention : SelfTargetingIntention<KtIfExpression>(KtIfExpression::class.java, "Merge 'else if'") {
+    override fun isApplicableTo(element: KtIfExpression, caretOffset: Int): Boolean {
+        val elseBody = element.`else` ?: return false
+        val nestedIf = elseBody.nestedIf() ?: return false
+        if (nestedIf.`else` != null) {
```
goodwinnk: Return can be lifted up.

idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return true
+    }
+
+    override fun applyTo(element: KtIfExpression, editor: Editor?) {
+        applyTo(element)
+    }
+
+    companion object {
+        fun applyTo(element: KtIfExpression): Int {
```
goodwinnk: Can function be inlined? Also what is the purpose of return value?

idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
```diff
@@ -0,0 +1,46 @@
...
+            val nestedIf = element.`else`?.nestedIf() ?: return -1
+            val condition = nestedIf.condition ?: return -1
+            val nestedBody = nestedIf.then ?: return -1
+
+            val factory = KtPsiFactory(element)
+            val newBody = element.`else`?.replace(
+                factory.createExpressionByPattern("if ($0) $1", condition, nestedBody)
+            ) ?: return -1
+            return newBody.textRange!!.startOffset
```
goodwinnk: Please avoid using `!!`. If you sure that some nullable expression can't be null, be explicit in reasoning and use `?: error("An explanation why value can't be null")`.

idea/src/org/jetbrains/kotlin/idea/intentions/MergeElseIfIntention.kt
DmiitriyJarosh: Return value was unnecessary. Fixed it.

goodwinnk:
Thank you for the pull request and congrats on becoming a Kotlin contributor! I have merged your commit manually as https://github.com/JetBrains/kotlin/commit/8c4dd1f331f304ccc9bb5d61137dcc4de67abb8f.

I did some modifications in commit that might be important to know for future contributions:
- Please add ticket number to commit message
- Don't forget to commit generated test files
- Avoid warnings in the new code (there was a one because of using inline function)

PR #2664, opened 2019-10-08
KT-33211 Quickfix "add parameter" for method references should infer functional type instead of KFunction

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33211


darthorimar:
Thank you!

PR #2665, opened 2019-10-08
[IR] Deal with forward references in default argument lambdas.

madsager:
Rely on the frontend weeding out cases that are not supported.

In psi2ir, introduce all the parameters before processing default
values.

Change the DefaultArgumentStubGenerator to generate code that
matches the behavior of the current backend.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
@@ -11,7 +11,6 @@ import org.jetbrains.kotlin.backend.common.ir.*
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.builders.*
-import org.jetbrains.kotlin.ir.builders.declarations.addValueParameter
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.declarations.impl.IrConstructorImpl
 import org.jetbrains.kotlin.ir.declarations.impl.IrFunctionImpl
@@ -84,6 +83,20 @@ open class DefaultArgumentStubGenerator(
                 variables[it] = newIrFunction.extensionReceiverParameter!!
             }
 
+            // In order to deal with forward references in default value lambdas,
+            // accesses to the parameter before it has been determined if there is
+            // a default value or not is redirected to the actual parameter of the
+            // $default function. This is to ensure that examples such as:
+            //
+            // fun f(f1: () -> String = { f2() },
+            //       f2: () -> String = { "OK" }) = f1()
+            //
+            // works correctly so that `f() { "OK" }` returns "OK" and
+            // `f()` throws a NullPointerException.
+            irFunction.valueParameters.associateWithTo(variables) {
+                newIrFunction.valueParameters[it.index]
+            }
+
             for (valueParameter in irFunction.valueParameters) {
                 val parameter = newIrFunction.valueParameters[valueParameter.index]
                 val remapped = if (valueParameter.defaultValue != null) {
@@ -129,7 +142,6 @@ open class DefaultArgumentStubGenerator(
                     passTypeArgumentsFrom(newIrFunction.parentAsClass)
                     passTypeArgumentsFrom(newIrFunction)
                     dispatchReceiver = newIrFunction.dispatchReceiverParameter?.let { irGet(it) }
-
                     params.forEachIndexed { i, variable -> putValueArgument(i, irGet(variable)) }
                 }
                 is IrSimpleFunction -> +irReturn(dispatchToImplementation(irFunction, newIrFunction, params))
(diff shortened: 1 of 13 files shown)


madsager:
@max-kammerer @udalov This is a bit of a strange case encountered by Compose users. This makes it work as in the current backend.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/FunctionGenerator.kt
```diff
@@ -289,21 +289,21 @@ class FunctionGenerator(declarationGenerator: DeclarationGenerator) : Declaratio
         }
 
         val bodyGenerator = createBodyGenerator(irFunction.symbol)
+
+        // Declare all the value parameters up first.
         functionDescriptor.valueParameters.mapTo(irFunction.valueParameters) { valueParameterDescriptor ->
             val ktParameter = DescriptorToSourceUtils.getSourceFromDescriptor(valueParameterDescriptor) as? KtParameter
-            generateValueParameterDeclaration(valueParameterDescriptor, ktParameter, bodyGenerator, withDefaultValues, irFunction)
+            declareParameter(valueParameterDescriptor, ktParameter, irFunction)
```
max-kammerer: @madsager Please add irText test (see  testData/ir/irText/declarations/parameters) for mentioned case

compiler/testData/codegen/box/defaultArguments/useNextParamInLambda.kt
```diff
@@ -0,0 +1,16 @@
...
+): String = f1()
+
+fun box(): String {
+    var result = "fail"
+    try {
+        f()
+    } catch (e : Exception) {
+        result = "OK"
+    }
+    return f { result }
```
pyos: This looks like it does `f(f1 = { result })` while you also probably want to test `f(f2 = { result })`.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -84,6 +83,23 @@ open class DefaultArgumentStubGenerator(
                 variables[it] = newIrFunction.extensionReceiverParameter!!
             }
 
+            // In order to deal with forward references in default value lambdas,
```
pyos: An easier way to achieve the same effect is to do

```kotlin
irFunction.valueParameters.associateWithTo(variables) {
    newIrFunction.valueParameters[it.index]
}
```

so that reads of `f1` before the temporary is created are redirected to the original argument.

compiler/testData/codegen/box/defaultArguments/useNextParamInLambda.kt
max-kammerer: Seems it's a bit confusing place cause of default arguments, but https://kotlinlang.org/docs/reference/lambdas.html#passing-a-lambda-to-the-last-parameter
so ```f ( { result })``` not equals to ```f { result }```

compiler/testData/codegen/box/defaultArguments/useNextParamInLambda.kt
```diff
@@ -0,0 +1,16 @@
+// IGNORE_BACKEND: JS, JS_IR
+
+fun f(
+    f1: () -> String = { f2() },
+    f2: () -> String = { "OK" }
```
max-kammerer: Maybe use hier non "OK" value cause it clashes with result value in ```f { result ("OK") }``` call

compiler/testData/codegen/box/defaultArguments/useNextParamInLambda.kt
madsager: Yeah, this is testing passing to f2. But it is a good point. Let me change the test to check what happens both when passing to f1 and f2 explicitly. Thanks!

compiler/testData/codegen/box/defaultArguments/useNextParamInLambda.kt
madsager: Yes! Thanks. :)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
madsager: That is a great point and that is indeed much simpler. Thanks!

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/FunctionGenerator.kt
madsager: Done. I'm unsure if I should add these expectations for the fir2ir test? These expectations say that f2 does not resolve which is incompatible with the current system. Is that OK since fir is so early right now, or should I rather put in some failure marker?

madsager: Just as a data point: it seems that there are other fir.txt files here that have ERROR_CALL in them where the normal .txt does not.

max-kammerer: Seems we can keep current variant with IrErrorTypes //cc @mglukhikh @semoro

mglukhikh: Error types in `.fir.txt` files aren't a problem right now.

max-kammerer:
Many thanks!

PR #2666, opened 2019-10-08
JVM_IR inliner fixes: inline lambdas in inline functions

pyos:
Extracted from #2646


max-kammerer:
> Not sure what Steven's change has to do with this, as CallableReferenceLowering completely ignores inline lambdas. If you're planning on generating classes for them, the situation will become even worse as classes can't have reified type parameters.

Another possible approach to keep IrFunctionExpression as long as possible (now it's unrolled into function+function reference in provisionalFunctionExpressionPhase): seems it will allow to keep reified parameters as long as possible. But this approach conflicts with logic in  inlineCallableReferenceToLambdaPhase...

pyos:
I still believe the correct approach is to make LocalDeclarationsLowering handle captured type parameters regardless of exactly what happens to inline lambdas. Even with non-reified ones, although the current logic works, it's kind of weird to have an IR equivalent of

```
fun <T> f() = g()
fun g() { /* T used here */ }
```

max-kammerer:
Yes, it's the rightest  approach. But not sure that the easiest.

max-kammerer:
Many thanks!

PR #2667, opened 2019-10-09
KT-24281 Importing of invoke() from the same file is reported as unused even if it isn't

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24281


vladimirdolzhenko:
LGTM

PR #2668, opened 2019-10-09
KT-34190 Inline Function: False positive error "not at the end of the body" with anonymous object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34190


igoriakovlev:
Thank you for your work.

PR #2669, opened 2019-10-09
JVM IR: Overrides + Incremental Compilation Bug

kandersen:
This commit introduces two tests illustrating the behaviour of the old
and new JVM backends pertaining to override resolution and its
interaction with separate/incremental compilation of interfaces and
clients of that interface. 

In particular, it reveals a flaw in code generation in the IR backend.

Note that this commit introduces a new, failing test. This PR is intended
as a "ticket".


kandersen:
I suppose as an afterthought this should go in YouTrack. At any rate, now there's concrete code to link to from YouTrack.

PR #2671, opened 2019-10-10
KT-19574 Code with inferred default parameters and parameter vs property name clashes

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19574

### DIFF
--- a/nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
+++ b/nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
@@ -9,6 +9,7 @@ import com.intellij.psi.PsiMethod
 import org.jetbrains.kotlin.nj2k.*
 import org.jetbrains.kotlin.nj2k.symbols.JKUniverseMethodSymbol
 import org.jetbrains.kotlin.nj2k.tree.*
+import org.jetbrains.kotlin.nj2k.types.JKNoType
 
 
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
@@ -91,6 +92,8 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
                 .zip(calledMethod.parameters)
                 .drop(method.parameters.size)
 
+            val parameters = defaults.map { it.second }
+            val declarations = element.declarations
 
             for ((defaultValue, parameter) in defaults) {
                 fun remapParameterSymbol(on: JKTreeElement): JKTreeElement {
@@ -102,7 +105,12 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
                             return JKFieldAccessExpression(newSymbol)
                         }
                     }
-
+                    if (on is JKExpression) {
+                        val newExpression = on.addThisReceiverIfNeeded(parameters, declarations)
+                        if (newExpression != null) {
+                            return newExpression
+                        }
+                    }
                     return applyRecursive(on, ::remapParameterSymbol)
                 }
                 parameter.initializer = remapParameterSymbol(defaultValue) as JKExpression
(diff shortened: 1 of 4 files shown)


nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
```diff
@@ -102,7 +105,12 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
                             return JKFieldAccessExpression(newSymbol)
                         }
                     }
-
+                    if (on is JKExpression) {
```
darthorimar: Please, let's don't break functionallity with remapping we already have. Join your `addThisReceiverIfNeeded` with `remapParameterSymbol`, so we will be able to do both and not just taking over one case and forgetting about other one

nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
```diff
@@ -125,6 +133,52 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
 
     }
 
+    private fun JKExpression.addThisReceiverIfNeeded(
+        parameters: List<JKParameter>,
+        declarations: List<JKDeclaration>
+    ): JKExpression? = when (this) {
+        is JKFieldAccessExpression, is JKCallExpression -> {
+            if (parameters.any { it.name.value == identifier?.name } && declarations.any { it == identifier?.target }) {
+                parent?.also { this.detach(it) }
```
darthorimar: Please, don't use detach directly. The `element::child.detached()` syntax is better to be used as it also check parent-child relation consistency. I will deprecate this method to avoid confusions

nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
```diff
@@ -125,6 +133,52 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
...
+                JKQualifiedExpression(JKThisExpression(JKLabelEmpty(), JKNoType), this)
+            } else {
+                null
+            }
+        }
+        is JKQualifiedExpression -> {
+            selector.detach(this)
```
darthorimar: The same with detach here and below, please use `this::selector.detached()`

nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
```diff
@@ -125,6 +133,52 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
 
     }
 
+    private fun JKExpression.addThisReceiverIfNeeded(
```
darthorimar: Also, please use `applyRecursive` for creating recursive conversions as it will not unnecessary recreate tree elements

nj2k/src/org/jetbrains/kotlin/nj2k/conversions/DefaultArgumentsConversion.kt
```diff
@@ -91,18 +92,31 @@ class DefaultArgumentsConversion(context: NewJ2kConverterContext) : RecursiveApp
...
                         if (target is JKParameter) {
                             val newSymbol =
                                 symbolProvider.provideUniverseSymbol(calledMethod.parameters[method.parameters.indexOf(target)])
                             return JKFieldAccessExpression(newSymbol)
                         }
+                        if (on.identifier.isNeedThisReceiver()) {
+                            return JKQualifiedExpression(JKThisExpression(JKLabelEmpty(), JKNoType), JKFieldAccessExpression(on.identifier))
+                        }
+                    }
+                    if (on is JKCallExpression && on.identifier.isNeedThisReceiver()) {
```
darthorimar: Why do we need this check? As far as I know, there should be no conflict with parameter/function names

In the following example `x` is resolved fine to the function `x`:
```
fun x() = 10
fun f(x: Int, y: Int = x()) = 42
```

t-kameyama: This is necessary for `isD: Boolean = this.isD` in:
https://github.com/t-kameyama/kotlin/blob/dfdf07a5cf21fcc557a5e19cce0a4f933a34c7bf/nj2k/testData/newJ2k/overloads/ConflictParameterName.kt#L23

darthorimar: You are right :) In that way let's conver the similar case when `getSmth` becomes `smth` like:

```
public class Test {
    int getX() { return 1; }

    void foo() { foo(getX()); }
    void foo(int x) {}
}
```
Becomes
```
class Test {
    internal val x: Int internal get() = 1
    @JvmOverloads
    internal fun foo(x: Int = x/*error*/) {}
}
```

t-kameyama: Fixed

darthorimar:
Thank you :)

PR #2672, opened 2019-10-10
JVM: generate $assertionsDisabled before inlining the node

pyos:
This fixes the problem where compiling a class initializer that contains a call to an `assert`ing function in a separate module causes the assertion to always be enabled (i.e. the attached test used to fail in CompileKotlinAgainstInlineKotlin mode).


max-kammerer:
@ilmirus Please take a look. Do we have any issue for this?

ilmirus:
Thanks!

PR #2673, opened 2019-10-10
KT-34285 False positive "Redundant companion reference" for enum static function `values` and companion object's function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34285

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.analysis.analyzeAsReplacement
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.idea.intentions.isReferenceToBuiltInEnumFunction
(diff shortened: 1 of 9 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantCompanionReferenceInspection.kt
```diff
@@ -65,6 +66,9 @@ class RedundantCompanionReferenceInspection : AbstractKotlinInspection() {
 
             val containingClass = objectDeclaration.containingClass() ?: return false
             if (reference.containingClass() != containingClass && reference == parent.receiverExpression) return false
+            val calleeText = ((selectorExpression as? KtCallExpression) ?: (grandParent as? KtQualifiedExpression)?.callExpression)
```
nikitabobko, line 69: First: this function is already too long, you can't add more logic into it. You should extract 3 lines of code you added into separate function + split code of existing `isRedundantCompanionReference` into subfunctions

Second: `values` and `valueOf` are hardcoded strings. There are already present in the project. At least you can find them in `RemoveRedundantQualifierNameInspection` and `UnusedSymbolInspection`. Please, extract those hardcoded strings into util class/file and reuse them

nikitabobko:
Merged to new repo:
https://github.com/JetBrains/intellij-kotlin/commit/c36fd4056e3b5bc12318d921726b7f6922ee598e
https://github.com/JetBrains/intellij-kotlin/commit/09cddfbf9bb626f2de6be7fdf78497abcc2f4713

PR #2674, opened 2019-10-11
KT-24615 "Extract property" generates useless extension property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24615


igoriakovlev:
Thank you for your work.
Please pay attention on that we have now another ticket in this area: https://youtrack.jetbrains.com/issue/KT-36504

PR #2675, opened 2019-10-11
KT-34297 "Add 'replaceWith' argument" inserts positional instead of named argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34297


mglukhikh:
Merged 👍

PR #2676, opened 2019-10-11
JVM: Fix KT-34268

sfs:
The current JVM backend doesn't have a type for the left hand side of an equality in a when expression with a declaration, so the following code breaks when trying to specialize the implicit equality check `x == 1UL`.
```kotlin
when (val x = 1UL) {
  1UL -> True
  else -> False
}
```


udalov:
Many thanks for the timely fix!

PR #2678, opened 2019-10-12
KT-33951 ReplaceWith quickfix with unqualified object member call doesn't substitute argument for parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33951


idea/src/org/jetbrains/kotlin/idea/codeInliner/CodeToInlineBuilder.kt
```diff
@@ -164,6 +165,7 @@ class CodeToInlineBuilder(
             }
 
             if (expression.getReceiverExpression() == null) {
+                val targetCallable = (targetCallable as? FunctionImportedFromObject)?.callableFromObject ?: targetCallable
```
dimonchik0036: Should we use `ImportedFromObjectCallableDescriptor<*>` instead?

PR #2679, opened 2019-10-13
KT-11865 "Create secondary constructor" quick fix always inserts parameter-less call to `this()`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-11865


vladimirdolzhenko:
There is a false positive case:

```
class CtorPrimary(val f1: String) {}

val v6 = CtorPrimary(1, 2) // Error.
```

New secondary ctor is created incorrectly:
```
class CtorPrimary(val f1: String) {
    constructor(f1: Int, i: Int) : this(f1)
}
```

I'd say add `this` with parameters iff they are compatible by types, otherwise - add `this(<caret>)` as it behaves now.

I've added this note to the ticket.

t-kameyama:
PTAL

vladimirdolzhenko:
LGTM 👍 Merged

PR #2683, opened 2019-10-14
JVM Backends: Support parameterless main methods

kandersen:
This commit:

- introduces tests explicating what is and isn't considered a
  proper main method on the JVM backends.
- implements support for parameterless main methods on the JVM IR
  backend.
- identifies omissions in main method detection on the JVM backend.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.backend.jvm.lower
+
+import org.jetbrains.kotlin.backend.common.ClassLoweringPass
+import org.jetbrains.kotlin.backend.common.ir.allParameters
+import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
+import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
+import org.jetbrains.kotlin.backend.jvm.ir.getJvmNameFromAnnotation
+import org.jetbrains.kotlin.config.LanguageFeature
+import org.jetbrains.kotlin.config.languageVersionSettings
+import org.jetbrains.kotlin.descriptors.Modality
+import org.jetbrains.kotlin.descriptors.Visibilities
+import org.jetbrains.kotlin.ir.builders.declarations.addFunction
+import org.jetbrains.kotlin.ir.builders.declarations.addValueParameter
+import org.jetbrains.kotlin.ir.builders.irBlockBody
+import org.jetbrains.kotlin.ir.builders.irCall
+import org.jetbrains.kotlin.ir.builders.irReturn
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
+import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
+import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
+import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.util.functions
+import org.jetbrains.kotlin.ir.util.getAnnotation
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.resolve.DescriptorUtils
+import org.jetbrains.kotlin.types.Variance
+
+internal class MainMethodGenerationLowering(val context: JvmBackendContext) : ClassLoweringPass {
+
+    override fun lower(irClass: IrClass) {
+        if (!context.configuration.languageVersionSettings.supportsFeature(LanguageFeature.ExtendedMainConvention)) return
+        if (irClass.origin != IrDeclarationOrigin.FILE_CLASS) return
+
+        val parameterlessMain = irClass.functions.find { it.isParameterlessMainMethod() } ?: return
+
+        if (irClass.functions.any { it.isMainMethod() }) return
+
+        generateMainMethod(irClass, parameterlessMain)
+    }
+
+    private fun generateMainMethod(irClass: IrClass, parameterlessMain: IrSimpleFunction) {
+        irClass.addFunction {
+            name = Name.identifier("main")
+            visibility = Visibilities.PUBLIC
+            returnType = context.irBuiltIns.unitType
+            modality = Modality.FINAL
+            origin = JvmLoweredDeclarationOrigin.GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD
+        }.apply {
+            addValueParameter {
+                name = Name.identifier("args")
+                type = context.irBuiltIns.arrayClass.typeWith(context.irBuiltIns.stringType)
+            }
+            body = context.createIrBuilder(this.symbol).irBlockBody {
+                +irReturn(this.irCall(parameterlessMain))
+            }
+        }
+    }
+
+    private fun IrSimpleFunction.isParameterlessMainMethod(): Boolean =
+        typeParameters.isEmpty() && valueParameters.isEmpty() && returnType.isUnit() && name.asString() == "main"
+
+
+    private fun IrSimpleFunction.isMainMethod(): Boolean {
+        if (getJvmNameFromAnnotation() ?: name.asString() != "main") return false
+        if (!returnType.isUnit()) return false
+
+        val parameter = allParameters.singleOrNull() ?: return false
+        if (!parameter.type.isArray() && !parameter.type.isNullableArray()) return false
+
+        val argType = (parameter.type as IrSimpleType).arguments.first()
+        return when (argType) {
+            is IrTypeProjection -> {
+                (argType.variance != Variance.IN_VARIANCE) && argType.type.isStringClassType()
+            }
+            else -> false
+        }
+    }
+}
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
@@ -43,4 +43,5 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object SYNTHETIC_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("SYNTHETIC_INLINE_CLASS_MEMBER", isSynthetic = true)
     object INLINE_CLASS_GENERATED_IMPL_METHOD : IrDeclarationOriginImpl("INLINE_CLASS_GENERATED_IMPL_METHOD")
     object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD", isSynthetic = true)
+    object GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD : IrDeclarationOriginImpl("GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD", isSynthetic = true)
 }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
@@ -172,8 +172,15 @@ private val syntheticAccessorPhase = makeIrFilePhase(
     prerequisite = setOf(objectClassPhase, staticDefaultFunctionPhase, interfacePhase)
 )
 
+private val mainMethodGenerationPhase = makeIrFilePhase(
+    ::MainMethodGenerationLowering,
+    name = "MainMethodGeneration",
+    description = "Identify parameterless main methods and generate bridge main-methods"
+)
+
 @Suppress("Reformat")
 private val jvmFilePhases =
+        mainMethodGenerationPhase then
         typeAliasAnnotationMethodsPhase then
         stripTypeAliasDeclarationsPhase then
         provisionalFunctionExpressionPhase then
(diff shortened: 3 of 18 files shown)


kandersen:
Backpointer to [KT-34338](https://youtrack.jetbrains.com/issue/KT-34338).

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/AbstractWriteSignatureTest.kt
```diff
@@ -220,7 +230,7 @@ abstract class AbstractWriteSignatureTest : CodegenTestCase() {
             ).joinToString("\n") { "// $it" }
         }
 
-        val expectationRegex = Regex("^// (class|method|field): *([^:]+)(::(.+))? *(//.*)?")
+        val expectationRegex = Regex("^// (class|method|field|!method): *([^:]+)(::(.+))? *(//.*)?")
```
udalov: Is the new expectation kind used somewhere? I can't find usages in the tests in this PR

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+            body = context.createIrBuilder(this.symbol).irBlockBody {
+                +irReturn(this.irCall(parameterlessMain))
+            }
+        }
+    }
+
+    private fun IrSimpleFunction.isParameterlessMainMethod(): Boolean {
+        if (typeParameters.isNotEmpty()) return false
+        if (valueParameters.size != 0) return false
```
udalov: `isNotEmpty()`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+            else -> false
+        }
+    }
+
+    private val IrTypeProjection.isSubtypeOfNullableString
+        get() = type.isSubtypeOf(context.irBuiltIns.stringType.makeNullable(), context.irBuiltIns)
+
+
+    private val IrSimpleFunction.hasJvmNameMain
+        get() = name.asString() == "main"
```
udalov: I think you should only check `name` after assuring that there's no JvmName annotation. Otherwise `hasJvmNameMain` incorrectly returns true on this function:
```
@JvmName("notMain")
fun main() {}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+        return when (argType) {
+            is IrTypeProjection -> {
+                (argType.variance != Variance.IN_VARIANCE) && argType.isSubtypeOfNullableString
+            }
+            else -> false
+        }
+    }
+
+    private val IrTypeProjection.isSubtypeOfNullableString
+        get() = type.isSubtypeOf(context.irBuiltIns.stringType.makeNullable(), context.irBuiltIns)
```
udalov: Probably would be more correct to just call `type.isStringClassType()`, and inline this function, to prevent `Array<Nothing>` from being recognized as a string array

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+        return true
+    }
+
+    private fun IrSimpleFunction.isMainMethod(): Boolean {
+        if (!hasJvmNameMain) return false
+        if (!returnType.isUnit()) return false
+        if (allParameters.size != 1) return false
+
+        val parameter = allParameters.first()
+        if (!parameter.type.isArray()) return false
```
udalov: `MainFunctionDetector.isMain` also recognizes nullable array as a possible parameter type

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+        if (name.asString() != "main") return false
+        return true
+    }
+
+    private fun IrSimpleFunction.isMainMethod(): Boolean {
+        if (!hasJvmNameMain) return false
+        if (!returnType.isUnit()) return false
+        if (allParameters.size != 1) return false
+
+        val parameter = allParameters.first()
```
udalov: Suggestion to collapse with the previous line with:
```
val parameter = allParameters.singleOrNull() ?: return false
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+internal class MainMethodGenerationLowering(val context: JvmBackendContext) : ClassLoweringPass {
+
+    override fun lower(irClass: IrClass) {
+        if (!context.configuration.languageVersionSettings.supportsFeature(LanguageFeature.ExtendedMainConvention)) return
+        if (irClass.origin != IrDeclarationOrigin.FILE_CLASS) return
+
+        val parameterlessMain = irClass.functions.find { it.isParameterlessMainMethod() }
+        val hasRegularMain = irClass.functions.any { it.isMainMethod() }
+
+        if (parameterlessMain == null || hasRegularMain) return
```
udalov: I think we can split this check and do it after corresponding steps above (to avoid looking for regular main, if no parameterless main was found)

compiler/tests-common/tests/org/jetbrains/kotlin/jvm/compiler/AbstractWriteSignatureTest.kt
kandersen: You are completely right! That's a remnant from earlier testing. The absence of the generated main is now confirmed through the bytecode tests without patterns, detecting a duplicate JVM signature in case it IS generated when it shouldn't.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,97 @@
...
+                name = Name.identifier("args")
+                type = context.irBuiltIns.arrayClass.typeWith(context.irBuiltIns.stringType)
+            }
+            body = context.createIrBuilder(this.symbol).irBlockBody {
+                +irReturn(this.irCall(parameterlessMain))
+            }
+        }
+    }
+
+    private fun IrSimpleFunction.isParameterlessMainMethod(): Boolean {
```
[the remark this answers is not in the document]
kandersen: Sure thing! I guess it made it easier to work with, adding/removing and shuffling lines around.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Thanks, done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Good catch, flipped.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Of course! :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Done. Let me know how you feel about the revised version.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: Yes, great catch. Added a test.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
udalov: I think it's still incorrect on the same example, isn't it? If the JvmName is not "main", we shouldn't check the Kotlin name at all

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
udalov: Looks good, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
kandersen: You are completely right! Added a test to catch this, and fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -0,0 +1,92 @@
...
+                (argType.variance != Variance.IN_VARIANCE) && argType.type.isStringClassType()
+            }
+            else -> false
+        }
+    }
+
+
+    private val IrSimpleFunction.hasJvmNameMain: Boolean
+        get() {
+            val jvmAnnotatedName =
```
udalov: Minor, but this would probably be simpler:
```
val jvmName = (getAnnotation(DescriptorUtils.JVM_NAME)?.getValueArgument(0) as? IrConst<*>)?.value
        ?: name.asString()
return jvmName == "main"
```
Maybe even extracting and reusing `getJvmNameFromAnnotation` from `MethodSignatureMapper` would make sense

kandersen: Hah, yeah, that'd make sense. What's a good place to lift that out to? org.jetbrains.kotlin.backend.common.ir.IrUtils?

kandersen: And agreed on your review comment.

udalov: > What's a good place to lift that out to? 

Something JVM-specific, e.g. `compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt`

kandersen: Easy :) Done.

compiler/testData/codegen/bytecodeText/parameterlessMain/dontGenerateOnJvmNameExtensionNullable.kt
```diff
@@ -0,0 +1,9 @@
+// IGNORE_BACKEND: JVM
```
udalov: Actually I think it'd be better to avoid adding tests where `JVM` is disabled, and just have these logged in KT-34338 (which I think all of them are already). The reason is that we usually keep the compiler test data as the current production state, and report all cases which don't work as intended to YouTrack, because it somewhat simplifies working with test data.

I know there are already several tests that pass on JVM_IR and are ignored for JVM, but I don't want that to be the rule. For non-major issues like this one, IMHO it's OK to keep the difference in behavior underspecified, and come back to it only in the future when JVM IR becomes the default backend.

udalov: Thanks for asking. I actually thought about it too. Our current position is that we don't invest in fixing tests for language versions 1.0, 1.1 and 1.2. Unless it's extraordinarily easy (as in this PR, for example), in which case it's more convenient to add a LanguageFeature check, than to cause hassle with ignoring or moving test data. The reason is that we're aiming to ship JVM IR backend as the _default_ in 1.5, and starting from 1.4, we'll only support two previous language versions ([KT-25823](https://youtrack.jetbrains.com/issue/KT-25823)). So Kotlin 1.5 is only expected to support language versions 1.3 and 1.4.

`DONT_TARGET_EXACT_BACKEND` actually seems OK to me as a solution, since there's not a lot of pre-1.3 test data. Your approach also looks good to me, but I'd maybe extend it to moving _all_ outdated box test data to a separate folder, in which case I'd prefer it over `DONT_TARGET_EXACT_BACKEND`.

kandersen:
That should do it for this PR! :)

Will see about picking out language version 1.0, 1.1 and 1.2 tests out into a separate folder.

kandersen:
Rebased on master for ease of merge :)

udalov:
Thanks, looks very good now! I'll run a remote run and merge soon.

If you don't mind, I'll change "JVM Backends" in the commit message to "JVM IR" since there are actually no changes in the old JVM backend

udalov:
Merged as https://github.com/JetBrains/kotlin/commit/28b6913a25994fdb1f611b7130e4fd41fe53a643. Thanks once again.

PR #2687, opened 2019-10-15
KT-33617: Handle non-hierarchical URIs for incremental annotation processing

gavra0:
JDK9 may decide to load .java files from jar when resolving a type, even
when .class file exists in the same jar. Further on, these files will be
parsed and analyzed, which triggers the listener used by incremental
annotation processing. This commit avoids analyzing those files, and
makes sure URI passed from the compiler are such that a File instance
can be created.

Test: KaptIncrementalWithIsolatingApt.testSourcesInCompileClasspathJars


yanex:
Merged into master (8f94a71de7a70764e9ebbd272fb9a41321fa18e9). Thanks for your contribution!

PR #2692, opened 2019-10-16
JVM_IR: fix synthetic accessor generation (again)

pyos:
Turns out private inline functions *can* be called from other classes.


max-kammerer:
Thanks!

PR #2693, opened 2019-10-16
KT-34194: Pass error/NonExistentClass for incremental KAPT run

gavra0:
When running KAPT incrementally, make sure to pass error/NonExistentClass.java
when there is at least one source file to process. This is to make sure
APs that must resolve all types are able to do so.

Test: updated KaptIncrementalWithAggregating/WithIsolating


yanex:
Merged manually (https://github.com/JetBrains/kotlin/commit/f4c6f57354088c8ff542378da0e51e227e9bf3b9). Ivan, thanks for the contribution!

PR #2694, opened 2019-10-16
[JVM_IR] Contain 'invokeSuspend' rewriting to the suspend function.

madsager:
Do not rewrite returns in methods in nested lambda implementations.
That leads to unintended non-local returns from nested non-suspend
lambdas.


madsager:
@ilmirus

madsager:
Friendly ping. :)

ilmirus:
Thanks!

PR #2695, opened 2019-10-16
JVM_IR: never create temporaries in $default stubs

pyos:
Given the strict pattern-matching in the inliner, this is the only way to make it not crash when attempting to inline these stubs. Note that the IR backend does not currently use the inliner's default method stub handling; the crash only occurs when a module compiled with the non-IR JVM backend is attempting to call an inline function with default arguments defined in a module that was compiled with the IR backend.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantGotoMethodTransformer.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantGotoMethodTransformer.kt
@@ -31,6 +31,7 @@ class RedundantGotoMethodTransformer : MethodTransformer() {
      *  (1) subsequent labels
      *      ...
      *      goto Label (can be removed)
+     *      nop        (any number of them, or maybe none; will be removed by RedundantNopsCleanupMethodTransformer)
      *      Label:
      *      ...
      *  (2) indirect goto
@@ -49,6 +50,10 @@ class RedundantGotoMethodTransformer : MethodTransformer() {
         var pendingGoto: JumpInsnNode? = null
 
         for (insn in insns) {
+            // We can remove jumps over labels, NOPs, GOTOs with the same target, and fake
+            // instructions used to describe the current frame state. We have to keep jumps
+            // over line numbers, though, as otherwise something like an `if` with an empty
+            // `else` will trigger a breakpoint inside the `else` even when the condition is true.
             when {
                 insn is LabelNode -> {
                     currentLabels.add(insn)
@@ -62,8 +67,7 @@ class RedundantGotoMethodTransformer : MethodTransformer() {
                         currentLabels.clear()
                     }
                 }
-                insn is LineNumberNode -> pendingGoto = null
-                insn.isMeaningful -> {
+                insn is LineNumberNode || (insn.isMeaningful && insn.opcode != Opcodes.NOP) -> {
                     currentLabels.clear()
                     pendingGoto = null
                 }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
@@ -378,6 +378,18 @@ class JvmSymbols(
             returnType = dst.defaultType
         }.symbol
 
+    val reassignParameterIntrinsic: IrSimpleFunctionSymbol =
+        buildFun {
+            name = Name.special("<set-parameter>")
+            origin = IrDeclarationOrigin.IR_BUILTINS_STUB
+        }.apply {
+            parent = kotlinJvmInternalPackage
+            val type = addTypeParameter("T", irBuiltIns.anyNType)
+            addValueParameter("parameter", type.defaultType) // must be IrGetValue of an IrValueParameter
+            addValueParameter("value", type.defaultType)
+            returnType = irBuiltIns.unitType
+        }.symbol
+
     private val collectionToArrayClass: IrClassSymbol = createClass(FqName("kotlin.jvm.internal.CollectionToArray")) { klass ->
         klass.origin = JvmLoweredDeclarationOrigin.TO_ARRAY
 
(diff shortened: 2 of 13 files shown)


pyos:
@udalov I accidentally compiled half a project with JVM_IR and another half with JVM and found a backwards compatibility problem. This begs the question: are there any plans to introduce a kind of CompileKotlinAgainstKotlin tests where each module is compiled with a different backend?

udalov:
@pyos Good idea, but we don't have any specific plans for that right now. The only thing I'm worried about (as with the [different file order tests proposal](https://github.com/JetBrains/kotlin/pull/2637#discussion_r330506153)) is the major increase in the number of tests. Maybe we could have these kinds of tests outside of the main test suite, just like `:compiler:tests-different-jdk`. Also, it'd be interesting to see how many failures are unique (compared to the main test failures on the same test data). But in the long run, it makes sense to have such tests on `compileKotlinAgainstKotlin` and `boxInline` test data, both testing JVM against JVM_IR and vice versa.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ReassignParameter.kt
```diff
@@ -0,0 +1,36 @@
...
+//
+// This is used to optimize default method stubs (and, more importantly, produce bytecode that
+// the inliner can actually handle) in `JvmDefaultArgumentStubGenerator`.
+object ReassignParameter : IntrinsicMethod() {
+    override fun invoke(expression: IrFunctionAccessExpression, codegen: ExpressionCodegen, data: BlockInfo): PromisedValue {
+        val parameterGet = expression.getValueArgument(0) as? IrGetValue
+        val parameter = parameterGet?.symbol as? IrValueParameterSymbol
+            ?: throw AssertionError("${expression.getValueArgument(0)} is not a get of a parameter")
+        codegen.setVariable(parameter, expression.getValueArgument(1)!!, data)
+        return codegen.defaultValue(expression.type)
```
max-kammerer: This line a bit tricky maybe use VOID type explicitly?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -342,6 +342,18 @@ class JvmSymbols(
...
         }.symbol
 
+    val reassignParameterIntrinsic: IrSimpleFunctionSymbol =
+        buildFun {
+            name = Name.special("<set-parameter>")
+            origin = IrDeclarationOrigin.IR_BUILTINS_STUB
+        }.apply {
+            parent = kotlinJvmInternalPackage
+            val type = addTypeParameter("T", irBuiltIns.anyNType)
+            addValueParameter("p", type.defaultType) // must be IrGetValue of a parameter
```
max-kammerer: Please use full name: parameter and value

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantGotoMethodTransformer.kt
```diff
@@ -62,8 +62,7 @@ class RedundantGotoMethodTransformer : MethodTransformer() {
                         currentLabels.clear()
                     }
                 }
-                insn is LineNumberNode -> pendingGoto = null
-                insn.isMeaningful -> {
+                insn is LineNumberNode || (insn.isMeaningful && insn.opcode != Opcodes.NOP) -> {
```
max-kammerer: Could you clarify a bit this change?

pyos: As the comment in JvmDefaultArgumentStubGenerator says, you cannot have a jump between the store into the argument and the "had explicit parameter" label. Problem is, the IR codegen may insert a bunch of NOPs, resulting in something like

```
    IFEQ Ln
    ...
    STORE n
    GOTO Ln
    NOP
 Ln
```

The NOP is removed by RedundantNopsCleanupMethodTransformer, but it only runs after RedundantGotoMethodTransformer (as otherwise it may decide LINENUMBER+NOP+GOTO can be shortened to just LINENUMBER+GOTO, but then removing the GOTO leaves no instruction to attach LINENUMBER to). Obviously, however, jumping over NOPs is unnecessary, so this change removes the GOTO even though there's a NOP between it and the target. (Clearing `currentLabels` on a line number is necessary to avoid jumping over LINENUMBER+NOP pairs.)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ReassignParameter.kt
pyos: Done, it always returns Unit now

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantGotoMethodTransformer.kt
pyos: Added a small comment.

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantGotoMethodTransformer.kt
```diff
@@ -31,6 +31,7 @@ class RedundantGotoMethodTransformer : MethodTransformer() {
      *  (1) subsequent labels
      *      ...
      *      goto Label (can be removed)
+     *      nop        (will be removed by RedundantNopsCleanupMethodTransformer)
```
max-kammerer: @pyos Please mark nop as optional

pyos: done

max-kammerer:
@pyos Thank you! 
Please note that @dnpetrov would work on supporting writable parameters in IR.

PR #2697, opened 2019-10-16
JVM_IR: discard results of nonexhaustive whens' branches

pyos:


max-kammerer:
Thank you!

udalov:
@pyos Looks like this assertion leads to the failure in cases like this:
```
fun box(): String {
    var s = "2"
    var res = "Fail"
    if (s == "1") {
        false
    } else if (s == "2") {
        res = "OK"
    }
    return res
}
```
Here, psi2ir infers `Any` as the type of the when-expression, presumably because the current frontend infers `Any` there ([Generator.kt:48](https://github.com/JetBrains/kotlin/blob/eb744cc9e14f6055da89911b06553661b90f4bd5/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/Generator.kt#L48)). Since fixing the current frontend might be dangerous, maybe we could remove this assertion?

pyos:
Huh, thanks for the interesting example. I've uploaded a follow-up PR: #2706

PR #2699, opened 2019-10-17
Skip generating (some) method bodies in ABI mode

kevin1e100:
Fixes https://youtrack.jetbrains.com/issue/KT-33604 by skipping code generation for (some) method bodies.  This avoids issues with inlining code into method bodies we would drop later anyway.  As a (welcome) side effect this also avoids generating local and lambda classes contained in the methods being skipped.


altsvetkov:
Merged manually with minor changes. 
@kevin1e100 thanks for the contribution!

PR #2700, opened 2019-10-17
Avoid PsiFile searching when resolving imported scripts descriptors.

jdemeulenaere:
Doing this allows to import FileBasedScriptSource's and correctly resolve imported descriptors.

An even better fix would be to also correctly support StringScriptSource's.


ligee:
Merged manually

devLiuGit:
https://youtrack.jetbrains.com/issue/KTIJ-16352
Scripts / IDE: unresolved reference when trying to import file in main.kts scripts with @Import annotation

PR #2701, opened 2019-10-17
KT-33796 INVISIBLE_SETTER: quick fix "Make '<set-attribute>' public" does not remove redundant setter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33796


idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeVisibilityFix.kt
```diff
@@ -51,6 +54,11 @@ open class ChangeVisibilityFix(
         }
 
         element?.setVisibility(visibilityModifier)
```
dimonchik0036: Use `SmartPsiElementPointer`. Otherwise, you can catch `PsiInvalidElementAccessException`.
```kotlin
val pointer = element?.createSmartPointer()

val originalElement = pointer?.element
if (originalElement is KtDeclaration) {
    originalElement.runOnExpectAndAllActuals { it.setVisibility(visibilityModifier) }
}

pointer?.element?.setVisibility(visibilityModifier)

val propertyAccessor = pointer?.element as? KtPropertyAccessor
```

PR #2703, opened 2019-10-17
Make useIR available from Gradle KotlinJvmOptions

jimgoog:
Make useIR available from Gradle KotlinJvmOptions.


jimgoog:
cc @erokhins

PR #2705, opened 2019-10-18
Removed configuration that isn't needed anymore for composite build o…

LepilkinaElena:
…f Kotlin/Native


LepilkinaElena:
Found cases when resolve doesn't work, so closed

PR #2710, opened 2019-10-21
Add launcher script for K2MetadataCompiler

Foso:
This will add a launcher script for the K2MetadataCompiler

### DIFF
--- a/compiler/cli/bin/kotlinc-metadata
+++ b/compiler/cli/bin/kotlinc-metadata
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Copyright 2010-2015 JetBrains s.r.o.
+#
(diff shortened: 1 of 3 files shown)


udalov:
Hi! Thanks for the contribution. Could you elaborate why do you need this? AFAIK we don't intend to make this compiler a public part of the CLI, since we don't target or support the case of building multiplatform projects outside of Gradle right now.

Foso:
Hi, the TODO message sounded like it's planned to add a launcher script

udalov:
TODOs do not usually mean that the task is planned to be done soon :). At this point, we have no intention to add this launcher script since it's a pretty visible change for the user which might suggest that we're adding proper support for multiplatform projects in the base CLI, which we aren't. Thanks anyway!

PR #2711, opened 2019-10-21
[JVM_IR] Support suspend functions that can be overriden.

madsager:
The challenge for overridable suspend functions is that the calling
the state machine method to resume after suspension would be
virtually dispatced to the wrong method. To avoid that a static
suspend implementation method is generated which becomes the
state machine method used to resume.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
```diff
@@ -66,7 +67,7 @@ internal fun generateStateMachineForNamedFunction(
         shouldPreserveClassInitialization = state.constructorCallNormalizationMode.shouldPreserveClassInitialization,
         containingClassInternalName = classCodegen.visitor.thisName,
         isForNamedFunction = true,
-        needDispatchReceiver = irFunction.dispatchReceiverParameter != null,
+        needDispatchReceiver = irFunction.dispatchReceiverParameter != null || irFunction.name.asString().endsWith(SUSPEND_IMPL_NAME_SUFFIX),
```
madsager: A non-name based way of determining this would be nice. I'm open to suggestions! :)

ilmirus: You can use different origin, something like `SUSPEND_IMPL_STATIC_FUNCTION`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -127,7 +127,9 @@ open class ClassCodegen protected constructor(
             val shortName = File(fileEntry.name).name
             visitor.visitSource(shortName, null)
         }
-
+        if (irClass in context.suspendFunctionContinuations.values) {
```
madsager: This is a left-over from a previous hack. It can be removed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
madsager: Done, I added it to IrDeclarationOrigin.kt. However, it looks like it is JVM_IR only, so it might be better to move it somewhere else?

udalov: There's `JvmLoweredDeclarationOrigin` for JVM-specific origins

madsager: Thanks @udalov, that is what I was looking for. I found JvmDeclarationOrigin which wasn't exactly what I was looking for. Done!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
```diff
@@ -43,4 +43,5 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object SYNTHETIC_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("SYNTHETIC_INLINE_CLASS_MEMBER", isSynthetic = true)
     object INLINE_CLASS_GENERATED_IMPL_METHOD : IrDeclarationOriginImpl("INLINE_CLASS_GENERATED_IMPL_METHOD")
     object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD", isSynthetic = true)
+    object SUSPEND_IMPL_STATIC_FUNCTION :  IrDeclarationOriginImpl("SUSPEND_IMPL_STATIC_FUNCTION", isSynthetic = true)
```
ilmirus: Since `SUSPEND_IMPL_STATIC_FUNCTION` is used in only one file, could you, please, move it there and make private, like `CONTINUATION_CLASS`?

madsager: Unfortunately, it is used in two files: AddContinuationLowering (which sets it) and CoroutineCodegen (which reads it).

ilmirus: Ah, my bad.

ilmirus:
Thanks!

PR #2712, opened 2019-10-21
Fix test to actually fail if expected exception is not thrown.

madsager:


udalov:
Seems a bit roundabout :) How about just
```
try {
    f()
    return "Fail: f() should have thrown NPE"
} catch (e: Exception) {
}
```
?

madsager:
Thanks Alex, that was indeed silly. Thanks! :-)

PR #2713, opened 2019-10-21
Kotlin/Native performance gradle plugin

LepilkinaElena:
…le time and code size from binaries built with MPP

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/build.gradle.kts
+++ b/libraries/tools/kotlin-gradle-plugin/build.gradle.kts
@@ -210,4 +210,9 @@ pluginBundle {
         id = "org.jetbrains.kotlin.native.cocoapods",
         display = "Kotlin Native plugin for CocoaPods integration"
     )
+    create(
+        name = "kotlinNativePerformancePlugin",
+        id = "org.jetbrains.kotlin.native.performance",
+        display = "Kotlin Native plugin for tracking performance of building binaries"
+    )
 }
--- a/libraries/tools/kotlin-gradle-plugin/build.gradle.kts
+++ b/libraries/tools/kotlin-gradle-plugin/build.gradle.kts
@@ -210,9 +210,4 @@ pluginBundle {
         id = "org.jetbrains.kotlin.native.cocoapods",
         display = "Kotlin Native plugin for CocoaPods integration"
     )
-    create(
-        name = "kotlinNativePerformancePlugin",
-        id = "org.jetbrains.kotlin.native.performance",
-        display = "Kotlin Native plugin for tracking performance of building binaries"
-    )
 }
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
@@ -9,30 +9,44 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.execution.TaskExecutionListener
+import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.tasks.TaskState
 import org.jetbrains.kotlin.gradle.dsl.multiplatformExtension
-import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.addExtension
+import org.jetbrains.kotlin.gradle.plugin.mpp.disambiguateName
+import org.jetbrains.kotlin.gradle.targets.native.tasks.NativePerformanceReport
+import org.jetbrains.kotlin.gradle.utils.lowerCamelCaseName
 import java.io.File
+import java.util.concurrent.ConcurrentHashMap
 
-class TaskTimerListener : TaskExecutionListener {
-    companion object {
-        val tasksTimes = mutableMapOf<String, Double>()
-        fun getTime(taskName: String) = tasksTimes[taskName] ?: 0.0
-    }
-
-    private var startTime = System.nanoTime()
+class TaskTimerListener(project: Project) : TaskExecutionListener {
+    val tasksTimes = getOrRegisterStorage<String, Double>(project, "tasksTimes")
+    val tasksStartedTime = getOrRegisterStorage<String, Long>(project, "tasksStartedTime")
+    fun getTime(taskName: String) = tasksTimes[taskName] ?: 0.0
 
     override fun beforeExecute(task: Task) {
-        startTime = System.nanoTime()
+        tasksStartedTime[task.path] = System.nanoTime()
     }
 
     override fun afterExecute(task: Task, taskState: TaskState) {
-        tasksTimes[task.name] = (System.nanoTime() - startTime) / 1000.0
+        tasksTimes[task.path] = (System.nanoTime() - tasksStartedTime[task.path]!!) / 1000.0
     }
+
+    companion object {
+        @Suppress("UNCHECKED_CAST")
+        private fun <K, V> getOrRegisterStorage(project: Project, propertyName: String): ConcurrentHashMap<K, V> =
+            project.rootProject.extensions.getByType(ExtraPropertiesExtension::class.java).run {
+                if (!has(propertyName)) {
+                    set(propertyName, ConcurrentHashMap<K, V>())
+                }
+                get(propertyName)
+            } as ConcurrentHashMap<K, V>
+    }
+
 }
 
 open class KotlinPerformancePlugin : Plugin<Project> {
+    private lateinit var reportDirectory: File
     private fun checkSettings(project: Project, performanceExtension: PerformanceExtension): Boolean {
         var result = true
         if (performanceExtension.metrics.isEmpty()) {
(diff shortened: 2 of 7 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.gradle.api.execution.TaskExecutionListener
+import org.gradle.api.tasks.TaskState
+import org.jetbrains.kotlin.gradle.dsl.multiplatformExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
+import org.jetbrains.kotlin.gradle.plugin.addExtension
+import java.io.File
+
+class TaskTimerListener : TaskExecutionListener {
+    companion object {
+        val tasksTimes = mutableMapOf<String, Double>()
```
ilmat192: I would avoid using static data because of Gradle daemon. It doesn't reload classes if a classpath of a build isn't change so this property may contain obsolete data.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+        return result
+    }
+
+    private fun getCompilationResults(tasksNames: Iterable<String>, success: Boolean): String {
+        var time = 0.0
+        var status = true
+        tasksNames.forEach {
+            time += TaskTimerListener.getTime(it)
+            status = tasksNames.map { TaskTimerListener.tasksTimes.containsKey(it) }.reduce { a, b -> a && b }
```
ilmat192:
```kotlin
var status = success
....
status &= tasksNames.all { TaskTimerListener.tasksTimes.containsKey(it) }
```
?

ilmat192: Do we really need to take into account compile time of all associated compilations?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
```diff
@@ -0,0 +1,134 @@
...
+                    var compileTime: String? = null
+                    if (TrackableMetric.COMPILE_TIME in performanceExtension.metrics) {
+                        compileTime = getCompilationResults(
+                            listOf(binary.linkTaskName) + compileTasks.map { it.name },
+                            successStatus
+                        )
+                    }
+
+                    // Create report.
+                    val reportFile = File(project.buildDir, "${it.name}.txt")
```
ilmat192: I think it's better to create some subdirectory for these reports.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/PerformanceExtension.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Input
+    var metrics = listOf(TrackableMetric.COMPILE_TIME, TrackableMetric.CODE_SIZE)
+
+    @Input
+    var trackedBinaries: List<NativeBinary> = emptyList()
+
+    @Optional
+    @Input
+    var binaryNamesForReport = mapOf<NativeBinary, String>()
+}
```
ilmat192: May be add some DSL methods allowing one to add a binary along with its name for report? Say:
```
// Or `forBinary` or some other name, it's a subject for discussion.
fun binary(binary: NativeBinary, nameForReport: String = binary.name)
```

libraries/tools/kotlin-gradle-plugin/build.gradle.kts
```diff
@@ -210,4 +210,9 @@ pluginBundle {
         id = "org.jetbrains.kotlin.native.cocoapods",
         display = "Kotlin Native plugin for CocoaPods integration"
     )
+    create(
```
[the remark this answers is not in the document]
LepilkinaElena: Decided that not.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
```diff
@@ -0,0 +1,93 @@
...
+        return tasks
+    }
+
+    @TaskAction
+    fun generate() {
+        val compileTasks = if (settings.includeAssociatedTasks)
+            getAllExecutedTasks(binary.linkTask.compilation)
+        else
+            listOf(binary.linkTask.compilation.compileKotlinTask)
+        val allExecutedTasks = listOf(binary.linkTask) + compileTasks
```
ilmat192: (minor) May be move this logic into the `getAllExecutedTasks` function?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
```diff
@@ -0,0 +1,97 @@
...
+        val upToDateTasks = allExecutedTasks.filter { it.state.upToDate }.map { it.name }.joinToString("\n", "- ")
+        if (upToDateTasks.isEmpty()) {
```
ilmat192: `upToDateTasks.isEmpty()` -> `upToDateTasks.isNotEmpty()`?

Also may be make `upToDateTasks` a list instead of a string? And transform it to a string only when the warning is generated?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
```diff
@@ -0,0 +1,97 @@
...
+            if (reportDirectory.exists()) {
+                project.delete(reportDirectory.absolutePath)
```
ilmat192: Not sure if we need to delete the whole report directory here.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+            project.logger.warn("Next compile tasks which are needed for time measurement are upToDate and weren't executed:\n${upToDateTasks}")
+            return
+        }
+        val successStatus = allExecutedTasks.map { it.state.failure == null }.reduce { acc, it -> acc && it }
```
ilmat192:
```suggestion
        val successStatus = allExecutedTasks.all { it.state.failure == null }
```

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
LepilkinaElena: It's recursive

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/tasks/KotlinNativePerformanceTask.kt
LepilkinaElena: I think, we need, because then user can skip this warning and use wrong report. If user already run new build he's ready that this directory inside build will be rewritten.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
```diff
@@ -0,0 +1,102 @@
...
+class TaskTimerListener(project: Project) : TaskExecutionListener {
+    val tasksTimes = getOrRegisterStorage<String, TaskTime>(project, "org.jetbrains.kotlin.native.taskTimes")
+    fun getTime(taskName: String) = tasksTimes[taskName]?.duration ?: 0.0
+
+    override fun beforeExecute(task: Task) {
+        tasksTimes[task.path] = TaskTime(System.nanoTime())
+    }
+
+    override fun afterExecute(task: Task, taskState: TaskState) {
+        tasksTimes[task.path]!!.duration = (System.nanoTime() - tasksTimes[task.path]!!.startTime) / 1000.0
```
ilmat192:
```suggestion
        val taskTime = taskTimes.getValue(task.path)
        taskTime.duration = (System.nanoTime() - taskTime.startTime) / 1000.0
```

PR #2714, opened 2019-10-21
[Backend] Reorganize <1.3 Codegen Tests

kandersen:
- Extract all codegen tests that specifically target behaviour in to-be deprecated functionality from language versions < 1.3
- Remove those same tests from the JVM IR test suite.


kandersen:
@udalov I took a stab at moving all codegen tests targetting language version <1.3, as we discussed in the parameterless main thread.

The PR is horrible as git swaps the entire [BlackBoxCodegen|BytecodeText]TestGenerated.java files, which I don't know how to keep it from doing.

Suggestions on all aspects of this PR is very welcome!

udalov:
Yeah, I think it'd be slightly more convenient to keep the old language version tests in the corresponding test data directories. It's not a lot more difficult to remove several directories instead of one, when needed :)

kandersen:
@udalov There! Diff is looking _a lot_ better, too :)

udalov:
I actually meant the single directory for each test (see the message above), but this works well too :). [Merged](https://github.com/JetBrains/kotlin/commit/8af3b3e51e9084a0d541c49ec295892efc36cef9) with minor modifications, thank you!

PR #2715, opened 2019-10-21
JVM_IR inliner fixes: anonymous objects as anonymous object constructor parameters

pyos:
The SAM case was spotted while trying to compile RxKotlin; the object super constructor case is currently broken in the non-IR backend (see KT-21781).

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
@@ -6,17 +6,21 @@
 package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
+import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.common.lower.irBlock
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.ir.builders.declarations.addValueParameter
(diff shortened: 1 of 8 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingleAbstractMethodLowering.kt
```diff
@@ -120,6 +121,14 @@ class SingleAbstractMethodLowering(val context: CommonBackendContext) : FileLowe
                     }
                     +irIfNull(superType, irGet(invokableVariable), irNull(), instance)
                 }
+            } else if (invokable !is IrGetValue) {
+                // Hack for the JVM inliner: since the SAM wrappers might be regenerated, avoid putting complex logic
```
max-kammerer: Please mention KT-21781 here

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
```diff
@@ -100,14 +105,17 @@ private class AnonymousObjectSuperConstructorLowering(val context: JvmBackendCon
...
-            for (i in 0 until objectConstructorCall.valueArgumentsCount)
-                putValueArgument(i, objectConstructorCall.getValueArgument(i))
-            for ((i, argument) in newArguments.withIndex())
-                putValueArgument(i + objectConstructorCall.valueArgumentsCount, argument)
+        context.createIrBuilder(currentScope!!.scope.scopeOwnerSymbol).run {
+            expression.statements[expression.statements.size - 1] = irBlock(objectConstructorCall) {
+                +irCallConstructor(objectConstructor.symbol, listOf()).apply {
+                    for (i in 0 until objectConstructorCall.valueArgumentsCount)
+                        putValueArgument(i, objectConstructorCall.getValueArgument(i))
+                    // Avoid complex expressions between `new` and `<init>`, as the inliner gets confused if
```
max-kammerer: ...and here

max-kammerer:
Thank you!

PR #2716, opened 2019-10-22
JVM_IR inliner fixes: remove code that does nothing

pyos:
other than randomly throw assertion errors on inline calls in synthetic functions, which have no line number attached. (E.g. it makes compilation of `@JvmStatic inline fun` and `@JvmOverloads inline fun` fail *only when using the daemon*, as for some reason it runs with assertions enabled while gradle in-process compilation happens with assertions disabled. Arguably, nothing should be inlined into `@JvmStatic` and `@JvmOverloads` bridges as those are only for Java use anyway, but that's a separate issue.)

`lazySourceMapper.callSiteMarker` is set in `InlineCodegen.inlineCall`; `IrSourceCompilerForInline.doCreateMethodNodeFromSource` does not need to touch it.


ilmirus:
Thanks!

PR #2717, opened 2019-10-22
docs: improved descriptions of compiler options

p7nov:


compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
```diff
@@ -25,11 +25,11 @@ class K2JSCompilerArguments : CommonCompilerArguments() {
     }
 
     @GradleOption(DefaultValues.StringNullDefault::class)
-    @Argument(value = "-output", valueDescription = "<path>", description = "Output file path")
+    @Argument(value = "-output", valueDescription = "<filepath>", description = "Destination *.js file for the compilation result")
     var outputFile: String? by NullableStringFreezableVar(null)
 
     @GradleOption(DefaultValues.BooleanTrueDefault::class)
-    @Argument(value = "-no-stdlib", description = "Don't use bundled Kotlin stdlib")
+    @Argument(value = "-no-stdlib", description = "Exclude kotlin-stdlib-js.jar from the compilation dependencies")
```
bashor: I think it would be better to avoid using file names here.

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
```diff
@@ -78,26 +78,29 @@ class K2JSCompilerArguments : CommonCompilerArguments() {
...
-            description = "Path to file which will be added to the beginning of output file"
+            description = "Add the content of the specified file to the beginning of output file"
     )
     var outputPrefix: String? by NullableStringFreezableVar(null)
 
     @Argument(
             value = "-output-postfix",
             valueDescription = "<path>",
-            description = "Path to file which will be added to the end of output file"
+            description = "Add the content of the specified file to the end of output file."
```
bashor: In most cases we don't have a dot at the end.

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt
```diff
@@ -18,31 +18,35 @@ class K2JVMCompilerArguments : CommonCompilerArguments() {
     @Argument(value = "-d", valueDescription = "<directory|jar>", description = "Destination for generated class files")
     var destination: String? by NullableStringFreezableVar(null)
 
-    @Argument(value = "-classpath", shortName = "-cp", valueDescription = "<path>", description = "Paths where to find user class files")
+    @Argument(
+        value = "-classpath",
+        shortName = "-cp",
+        valueDescription = "<path>",
+        description = "List of directories and JAR/ZIP archives to search for user class files")
```
bashor: Why it isn't just "class files"? (instead of "user class files")

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
bashor: I think it isn't correct to say "exclude" here. It's the way to say the compiler "Don't include default/bundled stdlib".

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt
udalov: Java and Scala use this terminology

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
udalov: We use file names in the JVM's desription of `-no-stdlib`, and IMHO it adds clarity there. Maybe just `kotlin-stdlib-js` here for JS would be enough

bashor: Other questions, unrelated to this PR:
1. Do we really need this option in "public API"?
2. Does anybody except us and (our) build tools use it?
3. Could/should we add stdlib to deps by default in build tools too and have a way to provide another version.

In JS IR BE we are going to stop supporting this option.

udalov: Yes, we do need some option like this to be able to control the compilation classpath precisely. This is needed for external build tools and plugins, not all of which are under our control. E.g. Bazel, Buck, Kobalt, etc.

There can't be a default way to add stdlib to dependencies in build tools, because at least some build tools do not use our compiler distribution where all libraries are alongside the compiler. So there's no standard way to find `kotlin-stdlib.jar` from the compiler.

BTW this option could've been instead something like `-with-stdlib=<path>`. But I don't see any reason to change it, the current way already works for all use cases ¯\\_(ツ)_/¯

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
```diff
@@ -25,11 +25,11 @@ class K2JSCompilerArguments : CommonCompilerArguments() {
     }
 
     @GradleOption(DefaultValues.StringNullDefault::class)
-    @Argument(value = "-output", valueDescription = "<path>", description = "Output file path")
+    @Argument(value = "-output", valueDescription = "<filepath>", description = "Destination *.js file for the compilation result")
     var outputFile: String? by NullableStringFreezableVar(null)
 
     @GradleOption(DefaultValues.BooleanTrueDefault::class)
-    @Argument(value = "-no-stdlib", description = "Don't use bundled Kotlin stdlib")
+    @Argument(value = "-no-stdlib", description = "Don't use the default Kotlin/JS stdlib as a compilation dependency")
```
bashor: It would be better to use the same wording for both JS and JVM.

bashor:
LGTM

bashor:
@udalov as far as I understand you are going to merge it manualy

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt
bashor: cc @udalov

udalov: Can you clarify which wording do you think works better and how to apply it to the other platform?

I personally like the version with "Exclude ..." and file paths on JVM more

p7nov: I've asked our proofreader David Watson, he suggests using "Don't automatically include.." for such cases.

udalov: :okay:

bashor: LGTM

udalov:
Merged manually in https://github.com/JetBrains/kotlin/commit/c6f54b6412e385ab3b96c7f69d06aa5e346ad43a with subsequent fixes https://github.com/JetBrains/kotlin/commit/f248e4a2eee2585e555259a5adeafd3b6409b491, https://github.com/JetBrains/kotlin/commit/e7fe144f5cacacc0f2077af190ee35ad71410c6a

PR #2723, opened 2019-10-23
KT-33050: Pass source version to KAPT for JDK12+

gavra0:
In JDK 12+ source version should be passed using --source option,
instead of -source. This commit uses correct option based on the
JDK current build is using.

Fixes KT-33050


gavra0:
@yanex Can you please review this PR?

yanex:
Merged manually (https://github.com/JetBrains/kotlin/commit/3a7deffe2e964f886e8ffe765d8fadfcae59aac5). Thank you for your contribution!

PR #2724, opened 2019-10-23
Implementation of UastAnalysisPlugin for Kotlin

zinoviy23:
In IDEA we create UastAnalysisPlugin and it already implemented for Java


xiexed:
PR is obsolete. `KotlinUastAnalysisPlugin` is already available in the intellij-community repo

PR #2725, opened 2019-10-24
KT-31523 ReplaceWith introduces additional argument name for lambda when named argument is used on call-site

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31523


klunnii:
Thank you! Merged.

PR #2726, opened 2019-10-24
[JVM_IR] Support default interface suspend functions.

madsager:
Do this by reordering the lowerings so that the lowering that
adds continuations happen after default methods have been
eliminated if needed.

Mark more functions as known to be tail-calls: bridges and
delegated members.

Preserve source info and annotations when creating replacement
functions.


ilmirus:
Thanks!

PR #2727, opened 2019-10-24
JVM_IR: add accessors for protected members in divergent hierarchies

pyos:
Class B : A cannot access a protected field of A through a reference to an instance of C : A.


udalov:
Looks good, thanks!

PR #2729, opened 2019-10-25
KT-34575 False positive "Call chain may be simplified" for `sortedBy{}.firstOrNull()`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34575

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
@@ -7,11 +7,10 @@ package org.jetbrains.kotlin.idea.inspections.collections
 
 import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.codeInspection.ProblemsHolder
-import com.jetbrains.rd.util.firstOrNull
 import org.jetbrains.kotlin.backend.common.descriptors.isSuspend
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
+import org.jetbrains.kotlin.builtins.isFunctionType
 import org.jetbrains.kotlin.idea.KotlinBundle
-import org.jetbrains.kotlin.idea.quickfix.createFromUsage.callableBuilder.getReturnTypeReference
 import org.jetbrains.kotlin.js.resolve.JsPlatformAnalyzerServices
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.*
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -36,6 +42,15 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
                         }
                     ) return@check false
                 }
+                if (conversion.replacement == "maxBy" || conversion.replacement == "minBy") {
+                    val argument = firstResolvedCall.valueArguments.firstOrNull()?.value?.arguments?.firstOrNull()
+                    val returnType = when (val argumentExpression = argument?.getArgumentExpression()) {
+                        is KtLambdaExpression -> argumentExpression.bodyExpression?.statements?.lastOrNull()?.getType(context)
```
nikitabobko, line 48: I don't like this `statements?.lastOrNull()`. This code isn't supposed to work for such cases:
```
fun main() {
    listOf(1, null, 2).sortedBy {
        if (true) return@sortedBy it
        it!!
    }.first()
}
```

Even though test works in general because lambdas with return statement [are discarded](https://github.com/JetBrains/kotlin/blob/76ac4ffb96c45953ea11c51f9aa7f249a94cfee8/idea/src/org/jetbrains/kotlin/idea/inspections/collections/AbstractCallChainChecker.kt#L46)

But you relay on behavior which is placed somewhere else inside `qualifiedExpressionVisitor` function => code locality is broken.

Please, use not latest statement `statements?.lastOrNull()` but actual return type of lambda

t-kameyama: `argumentExpression.bodyExpression?.statements?.lastOrNull()?.getType(context)` always returns nullable type.
How do we get actual return type of lambda without using last statement?

yanex: @t-kameyama Why not just check the resulting descriptor parameter type?
`firstResolvedCall.resultingDescriptor.valueParameters`

t-kameyama: @yanex Fixed in 5e8d69e, PTAL.

yanex (approved):

yanex:
Hm, seems that the fix disables the inspection with not-null types as well:
```
fun main() {
    listOf(1, 3, 2).sortedBy { it }.first()
}
```

nikitabobko:
Merged to new repo https://github.com/jetbrains/intellij-kotlin/commit/7fc378826c9414ef7940a9013c2027a2664eefb3

PR #2730, opened 2019-10-25
Add debug information test for local variable types and naming check

neetopia:
This test checks local variable names and types at every step of execution. IR tests are disabled for line number mismatch between old backend and IR backend.

Slot number for local variable is not covered in this test, tried to access slot number but it appears slot() method for local variable is package private and is not part of interface.


max-kammerer:
Seems https://github.com/JetBrains/kotlin/pull/2739/files#diff-fcba868fc30b96f704497f295f16fa5e also could be transformed to to new scheme

ilmirus:
Thanks!

PR #2732, opened 2019-10-25
JVM_IR: Do not generate null checks for suspend function views.

madsager:
The arguments are null whenever the suspend function is resumed
and takes its state from the continuation parameter.


madsager:
@ilmirus

ilmirus:
Thanks!

However, I do not think, that just removing the assertions is the right way to go, because we can call suspend functions from Java. The old BE also has the bug. This is why I created https://youtrack.jetbrains.com/issue/KT-34619.

madsager:
Thanks Ilmir! I agree that there should be null checks here for the direct call path. I was surprised to see that the current backend didn't have them.

PR #2733, opened 2019-10-25
JVM: Use the correct line number for suspend calls.

madsager:
We compute the actual line number of the suspend call and then
throw it away instead of using it. Therefore, we had to generate
more line number instructions in order to pick up the right one.
With this change we don't have to do that.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
@@ -146,7 +146,8 @@ class CoroutineTransformerMethodVisitor(
         val suspensionPointLineNumbers = suspensionPoints.map { findSuspensionPointLineNumber(it) }
 
         val continuationLabels = suspensionPoints.withIndex().map {
-            transformCallAndReturnContinuationLabel(it.index + 1, it.value, methodNode, suspendMarkerVarIndex)
+            transformCallAndReturnContinuationLabel(
+                it.index + 1, it.value, methodNode, suspendMarkerVarIndex, suspensionPointLineNumbers[it.index])
         }
 
         methodNode.instructions.apply {
@@ -702,11 +703,12 @@ class CoroutineTransformerMethodVisitor(
         id: Int,
         suspension: SuspensionPoint,
         methodNode: MethodNode,
-        suspendMarkerVarIndex: Int
+        suspendMarkerVarIndex: Int,
+        suspendPointLineNumber: LineNumberNode?
     ): LabelNode {
         val continuationLabel = LabelNode()
         val continuationLabelAfterLoadedResult = LabelNode()
-        val suspendElementLineNumber = lineNumber
+        val suspendElementLineNumber = suspendPointLineNumber?.line ?: lineNumber
         var nextLineNumberNode = suspension.suspensionCallEnd.findNextOrNull { it is LineNumberNode } as? LineNumberNode
         with(methodNode.instructions) {
             // Save state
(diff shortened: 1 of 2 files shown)


madsager:
@ilmirus

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
```diff
@@ -702,11 +703,12 @@ class CoroutineTransformerMethodVisitor(
...
         methodNode: MethodNode,
-        suspendMarkerVarIndex: Int
+        suspendMarkerVarIndex: Int,
+        suspendPointLineNumber: LineNumberNode?
     ): LabelNode {
         val continuationLabel = LabelNode()
         val continuationLabelAfterLoadedResult = LabelNode()
-        val suspendElementLineNumber = lineNumber
+        val suspendElementLineNumber = suspendPointLineNumber?.line ?: lineNumber
         var nextLineNumberNode = suspension.suspensionCallEnd.findNextOrNull { it is LineNumberNode } as? LineNumberNode
```
madsager: This looks shaky to me. Scanning forward ignores control flow. Also, I would guess that for code on one line you might end up having one branch look like it is on the next line. So I think there is a bug here where you can get the wrong line number when you resume. I'll see if I can generate a test case and attempt a fix.

ilmirus:
Sorry, but I cannot accept this change.

We arrange LINENUMBERs this way to support 'step over' debugger action. The support is not ideal, but at least, there is some.

When one steps over suspending call, the debugger puts a breakpoint at the start of the function (hence the 'lineNumber' variable). Upon hitting the breakpoint, it removes it and places another breakpoint at the line after the call. So, the user should press 'step over' twice.

To better understand the issue, let's look at the example
```
1 suspend fun dummy() {}
2 
3 suspend fun foo() {
4   dummy()
5   dummy()
6 }
```
The generated code should have the following structure:
```
...
TABLESWITCH
...
LINENUMBER 4
INVOKESTATIC dummy
// if call was suspended
LINENUMBER 3
// return COROUTINE_SUSPENDED
// LINENUMBER 5
INVOKESTATIC dummy
// if call was suspended
LINENUMBER 3
// return COROUTINE_SUSPENDED
// LINENUMBER 6
```

To be completely honest, no one, who touched the coded, likes the hack and we have some solutions, but they require some functionality from the debugger, which it currently lacks. And adding the functionality requires changing IDEA debugger as well, which is not a quick process.

I cannot think of a change to a PR, which will make it follow the restriction. By the way, the failing tests are `KotlinSteppingTestGenerated$StepOver.testSoSuspendableCall*`. If you can make them pass, I will be happy to accept the PR.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
ilmirus: There is a bug. In fact, there are plenty of them. Every time a suspending call is placed at the same line with other suspending calls/throw expression, there will be a bug.
Some of them are fixed `suspendMe() ?: error("")`, some are not `suspendMe() + suspendMe()` etc.

madsager: Yes, after writing this comment I created an example that just had two throws on the same line:

```
if (b) throw A() else throw A()
```

and one of them indeed gets the wrong line number.

madsager:
Thanks for the details Ilmir. Will close this pull request and keep looking at it when time allows. :)

PR #2734, opened 2019-10-25
JVM_IR: treat interface methods in DefaultImpls as valid bridge targets

pyos:


udalov:
Do I understand correctly that this only fixes the problem of inconsistent hierarchy as seen from the BridgeLowering? In this case I believe it's better to fix the root cause, as for example is done in https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3431

pyos:
Yeah, I suppose that is more reliable.

PR #2735, opened 2019-10-25
JVM IR: Minor coroutine fixes

sfs:
Fixes to IrTypes in `AddContinuationLowering`

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
@@ -194,7 +194,7 @@ private class AddContinuationLowering(private val context: JvmBackendContext) : 
 
                     override fun visitReturn(expression: IrReturn): IrExpression {
                         val ret = super.visitReturn(expression) as IrReturn
-                        return IrReturnImpl(ret.startOffset, ret.endOffset, context.irBuiltIns.anyType, function.symbol, ret.value)
+                        return IrReturnImpl(ret.startOffset, ret.endOffset, ret.type, function.symbol, ret.value)
                     }
                 })
                 (irFunction.parent as IrDeclarationContainer).declarations.remove(irFunction)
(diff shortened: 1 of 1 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
```diff
@@ -194,7 +194,7 @@ private class AddContinuationLowering(private val context: JvmBackendContext) :
 
                     override fun visitReturn(expression: IrReturn): IrExpression {
                         val ret = super.visitReturn(expression) as IrReturn
-                        return IrReturnImpl(ret.startOffset, ret.endOffset, context.irBuiltIns.anyType, function.symbol, ret.value)
+                        return IrReturnImpl(ret.startOffset, ret.endOffset, ret.type, function.symbol, ret.value)
```
ilmirus: Thank you for the catch! This, in fact, should be `anyNType`, since `invokeSuspend` returns `ret.type | COROUTINE_SUSPENDED` and `ret.type` can be nullable. Thus, `ret.type` is also incorrect.

zhaolongqing:
I have a question, I have no way to write mdoels[modelClass] in models[] and I don't seem to be able to convert modelClass into Kclass. Big God, is there a way to solve this problem (turning modelClass into KClass)? Thank you for satisfying my curiosity!!
```
class ViewModelsFactory @Inject constructor(private val models: Map<KClass<out ViewModel>, Provider<ViewModel>>) :
    ViewModelProvider.Factory {

    override fun <T : ViewModel?> create(modelClass: Class<T>) =
        models[]?.get() as T

}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
sfs: The type here actually refers to the type of the `IrReturn` as an expression and is always `Nothing` (compare with `IrBuilderWithScope.irReturn`). I was confused about this as well, but it's probably a parameter since even builtin types are deserialized and not available as constants.

ilmirus:
Thanks!

PR #2737, opened 2019-10-28
Improvements in DependencyResolvers

niktolik:
1. Single string artifact coordinates in resolvers
2. ResultWithDiagnostics for artifact resolution result
3. Break compilation if dependency resolution failed
4. Tests for dependency resolvers


ligee:
Merged manually

PR #2738, opened 2019-10-28
JVM_IR: Fix Inline CallableReferences with Varargs

kandersen:
- Added tests to demonstrate broken behaviour: the interaction of inline
  functions and callable references with varargs and defaults in various
  combinations.
- Refactored InlineCallableReferencesToLambdaPhase to look like and use
  some of the infrastructure from CallableReferenceLowering.
- Lifted some of this infrastructure out to be broadly reusable.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
@@ -7,18 +7,16 @@ package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
-import org.jetbrains.kotlin.backend.common.ScopeWithIr
-import org.jetbrains.kotlin.backend.common.ir.copyTo
 import org.jetbrains.kotlin.backend.common.ir.copyTypeParametersFrom
-import org.jetbrains.kotlin.backend.common.ir.copyValueParametersToStatic
-import org.jetbrains.kotlin.backend.jvm.ir.isInlineParameter
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irBlock
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
-import org.jetbrains.kotlin.backend.jvm.codegen.isInlineFunctionCall
-import org.jetbrains.kotlin.backend.jvm.codegen.isInlineIrExpression
+import org.jetbrains.kotlin.backend.jvm.ir.IrInlineReferenceLocator
+import org.jetbrains.kotlin.backend.jvm.ir.createJvmIrBuilder
+import org.jetbrains.kotlin.backend.jvm.ir.irArray
+import org.jetbrains.kotlin.backend.jvm.ir.isLambda
 import org.jetbrains.kotlin.codegen.AsmUtil.BOUND_REFERENCE_RECEIVER
 import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.ir.builders.*
@@ -28,9 +26,7 @@ import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrFunctionReferenceImpl
 import org.jetbrains.kotlin.ir.types.IrSimpleType
-import org.jetbrains.kotlin.ir.types.classOrNull
-import org.jetbrains.kotlin.ir.types.isSubtypeOf
-import org.jetbrains.kotlin.ir.types.isSubtypeOfClass
+import org.jetbrains.kotlin.ir.types.IrTypeProjection
 import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 import org.jetbrains.kotlin.name.Name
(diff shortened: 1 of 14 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
```diff
@@ -49,191 +45,213 @@ internal val inlineCallableReferenceToLambdaPhase = makeIrFilePhase(
...
-                            expression.getValueArgument(valueParameter.index)?.let { argument ->
-                                if (argument is IrCallableReference && isInlineIrExpression(argument)) {
-                                    inlinableCR.add(argument)
-                                }
-                            }
-                        }
-                    }
-                }
+        IrInlineReferenceLocator.scan(context, irFile).apply {
+            inlinableFunctionReferences.addAll(inlineFunctionReferences.filter { !lambdaToCallSite.containsKey(it.symbol.owner) })
```
max-kammerer: @kandersen Could you clarify this filtering?

I also can't catch the difference with inlinablePropertyReferences.addAll(inlinePropertyReferences), actualy property references could be passed to crossinline argument

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
```diff
@@ -49,191 +45,213 @@ internal val inlineCallableReferenceToLambdaPhase = makeIrFilePhase(
...
                 name = Name.identifier("stub_for_inline")
                 visibility = Visibilities.LOCAL
-                returnType = referencedFunction.returnType
+                returnType = getter.returnType
                 isSuspend = false
             }.apply {
-                if (referencedFunction is IrConstructor) {
-                    copyTypeParametersFrom(referencedFunction.parentAsClass)
+
+                if (getter is IrConstructor) {
```
max-kammerer: Look like this piece of code should be in 'expandInlineFunctionReferenceToLambda' (and test is required for missed case)
Why did you make separation between `expandInlineGetterReferenceToLambda` and 'expandInlineFunctionReferenceToLambda'?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
kandersen: I can try! 

This pass should not "eta-expand" the function reference in a lambda in inline position. The IrInlineReferenceLocator tracks such lambdas in that map, and this was a convenient way to reuse that tracking. I can perhaps introduce a property on the Locator class that does the lookup in the keyset with a more meaningful name?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
kandersen: Thank you, will add a test.

Their bodies seemed different enough (varargs, defaults etc) that it seemed clearer to have two separate code paths for the two cases. I am open to suggestions.

kandersen: @max-kammerer Wait, got it. Admittedly it _was_ a long day yesterday :)

Hold for a revised PR.

kandersen:
@max-kammerer There! Take two :)

Looks much better, thanks for giving me that last little push.

max-kammerer:
@kandersen Please also look for failed tests https://teamcity.jetbrains.com/viewLog.html?buildId=2609920&buildTypeId=Kotlin_KotlinPublic_CompilerWithTests

@ilmat192 There is smth strange with BlackBoxCodegenTestGenerated$Coroutines$Bridges.testLambdaWithMultipleParameters_ir_1_2 
Do we still need Coroutines 1_2 tests?

kandersen:
Am I reading TeamCity right that it's only "complaining" that I can unmute that JS test?

max-kammerer:
Yes and seems lambdaWithMultipleParameters.kt can also be unmuted //cc @ilmirus

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
```diff
@@ -49,188 +45,147 @@ internal val inlineCallableReferenceToLambdaPhase = makeIrFilePhase(
...
+            val function = buildFun {
+                setSourceRange(expression)
+                origin = JvmLoweredDeclarationOrigin.FUNCTION_REFERENCE_IMPL
+                name = Name.identifier("stub_for_inlining")
+                visibility = Visibilities.LOCAL
+                returnType = referencedFunction.returnType
+                isSuspend = false
+            }.apply {
+                if (referencedFunction is IrConstructor) {
+                    copyTypeParametersFrom(referencedFunction.parentAsClass)
```
max-kammerer: Sorry, but actually this line is redundant cause 'expression.getTypeSubstitutionMap(referencedFunction)' abouve already look int parent class

compiler/testData/codegen/box/callableReference/function/inlineConstructor.kt
```diff
@@ -0,0 +1,7 @@
+class A<T> (val result: T)
```
max-kammerer: ...and test is redundant. there is more correct innerGenericConstuctor.kt with inners

max-kammerer:
Indeed, code looks much better and clearer!

max-kammerer:
Could you also swap positions of expandInlineFunctionReferenceToLambda and expandInlineFieldReferenceToLambda functions - it's much easier to read diff

kandersen:
Done, done, thanks and done! :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InlineCallableReferenceToLambda.kt
max-kammerer: Sorry again, I mislead myself and you mixing declaration site here and call site below. Actually all type parameters should be specified for correctness, but still they don't actually used because of further inlining. I can't image how it could be tested. So lets just revert type coping:
```
                if (referencedFunction is IrConstructor) {
                    copyTypeParametersFrom(referencedFunction.parentAsClass)
                }
                copyTypeParametersFrom(referencedFunction)
```

max-kammerer: And seems they should be used for call specification (so all parameters from parent are redundant) below

kandersen: No need to apologize! I am leaning a lot on the test suite for understanding (or the illusion thereof :) and yeah, there are no tests constraining this.

Reverting the lines.

kandersen: > And seems they should be used for call specification (so all parameters from parent are redundant) below

I don't understand what you mean here. They are copied as type parameters of the lambda, and the typeSubstitution map is used to introduce type arguements to the actual call.

kandersen:
```
...
                ).run {
                    irExprBody(irCall(referencedFunction).apply {
                        this@apply.descriptor.typeParameters.forEach {
                            putTypeArgument(it.index, expression.getTypeArgument(it.index))
                        }

                        var unboundIndex = 0
...
```
You mean something like this at the call site, driven by the copied parameters instead of the substitution map?

max-kammerer: 'typeArgumentsMap' now is unused

kandersen:
Cleared the unusued argument. Let me know if you find anything else.

Otherwise, I understand it's done?

max-kammerer:
Thank you! Merged manually with additional assertion message

PR #2740, opened 2019-10-29
KT-28605 "Convert to sealed class" for enum doesn't handle use of `values()` method

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-28605

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertEnumToSealedClassIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertEnumToSealedClassIntention.kt
@@ -9,8 +9,10 @@ import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiWhiteSpace
 import com.intellij.psi.codeStyle.CodeStyleManager
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
+import org.jetbrains.kotlin.idea.core.getOrCreateCompanionObject
 import org.jetbrains.kotlin.idea.refactoring.withExpectedActuals
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.*
@@ -18,6 +20,7 @@ import org.jetbrains.kotlin.psi.psiUtil.allChildren
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
 import org.jetbrains.kotlin.psi.psiUtil.siblings
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
 
 class ConvertEnumToSealedClassIntention : SelfTargetingRangeIntention<KtClass>(
     KtClass::class.java,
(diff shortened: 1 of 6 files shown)


mglukhikh:
I think it's Ok generally. However, I think the addition of `values()` and `valueOf()` functions to the generated sealed class (in case they were in use for original enum class) may be better solution here. First, it allows us to use the intention in more number of cases, and second, it removes the expensive reference search from `isApplicable` method.

yanex:
> `is KtQualifiedExpression -> parent.callExpression?.calleeExpression?.text in ENUM_SYNTHETIC_METHODS`

It's not a good idea to check references by comparing text chunks as they can be aliased. Please resolve the reference first.

yanex:
`ReferencesSearch.search()` is a potentially long operation, it should not be inside applicability checks (see `applicabilityRange()`).

yanex:
> `syntheticMethodReferences.any { "values" in it.element.parent.text }`

Oh wow. Let's then introduce an intermediate class that will handle all supported scenarios (`values()`/`valueOf()` and built-in/synthetic functions) and use it instead of comparing references by text.

t-kameyama:
> `ReferencesSearch.search()` is a potentially long operation, it should not be inside applicability checks (see `applicabilityRange()`).

Should we also generate enumValues/enumValues functions?

yanex:
> Should we also generate enumValues/enumValues functions?
I think the ideal would be to ask user if they want to adapt enum-specific usages throughout all code base after applying the intention. If the user says "yes", we can run reference search under a ProgressIndicator.

However, even simple unconditional generation of `values()` and `valueOf()` alone would be a nice compromise.

t-kameyama:
Fixed to generate only 'values/valueOf' in 53758f3, please take another look.

klunnii:
> Fixed to generate only 'values/valueOf' in [53758f3](https://github.com/JetBrains/kotlin/commit/53758f38a2f26e91112bcc72b6a2ebf318d0f318), please take another look.

Please, correct me if I'm wrong. As far as I understood @yanex [recommendation](https://github.com/JetBrains/kotlin/pull/2740#issuecomment-704169485), we shouldn't make unconditional references search prior to generating `values()` and `valueOf()`. From what I see in `org/jetbrains/kotlin/idea/intentions/ConvertEnumToSealedClassIntention.kt:108` (`init` block ) the search is still there and its scope includes the entire project:
```
init {
            val useScope: SearchScope = PsiSearchHelper.getInstance(enumClass.project).getUseScope(enumClass)
            val query = ReferencesSearch.search(enumClass, useScope)
            isUsedValues = query.any { it.hasEnumValuesMethodReference(enumClass) }
            isUsedValueOf = query.any { it.hasEnumValueOfMethodReference(enumClass) }
}
```

t-kameyama:
> However, even simple unconditional generation of values() and valueOf() alone would be a nice compromise.

I fixed it to generate 'values' and 'valueOf' functions without conditions. 
08da053

klunnii:
Please, pay attention to failing tests:
- QuickFixMultiModuleTestGenerated$Other.testConvertActualEnumToSealedClass
- QuickFixMultiModuleTestGenerated$Other.testConvertExpectEnumToSealedClass

t-kameyama:
> Please, pay attention to failing tests:
> 
> * QuickFixMultiModuleTestGenerated$Other.testConvertActualEnumToSealedClass
> * QuickFixMultiModuleTestGenerated$Other.testConvertExpectEnumToSealedClass

Fixed

t-kameyama:
> Please, pay attention to failing tests:
> 
> * QuickFixMultiModuleTestGenerated$Other.testConvertActualEnumToSealedClass
> * QuickFixMultiModuleTestGenerated$Other.testConvertExpectEnumToSealedClass

Fixed

t-kameyama:
New PR: https://github.com/JetBrains/intellij-community/pull/1603

PR #2741, opened 2019-10-29
JVM: Improve line number handling for suspend calls.

madsager:
Take branching and method calls into account when finding the line
number of the continuation. If there is no line number before
branching instructions or method calls, the following code is
still on the line of the suspend call itself.

This fixes a couple of issues with incorrect line numbers for
multiple throws on the same line or multipe suspend calls on
the same line.

In addition, it avoids the need to spam the method node with
repeated line number instructions in the IR backend.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt
@@ -11,7 +11,6 @@ import org.jetbrains.kotlin.codegen.ClassBuilder
 import org.jetbrains.kotlin.codegen.StackValue
 import org.jetbrains.kotlin.codegen.TransformationMethodVisitor
 import org.jetbrains.kotlin.codegen.inline.*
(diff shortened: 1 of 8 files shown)


madsager:
@ilmirus second attempt, this one passes the idea debugger tests and fixes at least a couple of issues with line numbers as illustrated by the new tests. Let me know what you think. :)

compiler/testData/codegen/box/coroutines/debug/elvisLineNumber.kt
```diff
@@ -28,14 +27,14 @@ fun box(): String {
         withLineBreak()
         return "FAIL 0"
     } catch (e: RuntimeException) {
-        if (e.stackTrace[0].lineNumber != 19) return "FAIL 1 ${e.stackTrace[0].lineNumber}"
```
max-kammerer: I would prefer to avoid unecessery test data changes: so it's best to add new line after // WITH_COROUTINES

madsager: Thanks Mike! Done.

ilmirus:
Thanks!

PR #2742, opened 2019-10-29
[JS] [KT-22228] [KT-19813] Fix testing with asynchronous @BeforeTest and @AfterTest

tokou:
## Description

This PR fixes the current way of running asynchronous tests in JS.
Currently, as described in [KT-22228](https://youtrack.jetbrains.com/issue/KT-22228) and [KT-19813](https://youtrack.jetbrains.com/issue/KT-19813) there's a way to run asynchronous test with Mocha by returning a promise in test functions.

However, as reported in [this comment](https://youtrack.jetbrains.com/issue/KT-19813#focus=streamItem-27-2684677.0-0), this does not work for `@BeforeTest` and `@AfterTest` annotated methods as the runner does not wait for them to finish before running the test.

This is explained when looking at the generated js code as the before and after method are actually run sequentially inside the generated test method (cf. the *Before* section below) and do not call the framework's (Mocha) `beforeEach` and `afterEach` methods.

This PR fixes this by:
- Adding `beforeEach` and `afterEach` to the `FrameworkAdapter` interface
- Updating `JSTestGenerator` to generate appropriate calls to those methods (cf. *After* section below)
- Updating Mocha from 3.4.2 to 6.2.2 because support for asynchronous `beforeEach` and `afterEach` (with promises) was not present in 3.4.2

**Note:** I noticed the presence of `TestGenerator` in `:compiler:ir:backend.js` which seems to be doing something similar to `JSTestGenerator` (but to IR instead of JS code). I'm not sure of its role and if it needs to be modified as well. But as the integration test was passing I did not investigate further.

## Tests

This was tested by:
- Making the `before` and `after` method asynchronous in the `MainTest` integration test
- Passing `./gradlew :kotlin-test:kotlin-test-js:kotlin-test-js-it:testMocha`

(Tests with Jasmine, Qunit, Jest and Tape all fail on async tests as before)
(I did not find any unit tests using JSTestGenerator class)

### Before

```javascript
    suite('SimpleTest', false, function () {
      test('testFoo', false, function () {
        var tmp$;
        tmp$ = new SimpleTest();
        tmp$.beforeFun();
        try {
          return tmp$.testFoo();
        }
        finally {
          tmp$.afterFun();
        }
      });
      test('testBar', false, function () {
        var tmp$;
        tmp$ = new SimpleTest();
        tmp$.beforeFun();
        try {
          return tmp$.testBar();
        }
        finally {
          tmp$.afterFun();
        }
      });
      test('testFooWrong', true, function () {
        var tmp$;
        tmp$ = new SimpleTest();
        tmp$.beforeFun();
        try {
          return tmp$.testFooWrong();
        }
        finally {
          tmp$.afterFun();
        }
      });
    });
```

### After

```javascript
    suite('SimpleTest', false, function () {
      var instanceSimpleTest = new SimpleTest();
      beforeEach('beforeFun', function () {
        return instanceSimpleTest.beforeFun();
      });
      afterEach('afterFun', function () {
        return instanceSimpleTest.afterFun();
      });
      test('testFoo', false, function () {
        return instanceSimpleTest.testFoo();
      });
      test('testBar', false, function () {
        return instanceSimpleTest.testBar();
      });
      test('testFooWrong', true, function () {
        return instanceSimpleTest.testFooWrong();
      });
    });
```

### DIFF
--- a/js/js.translator/src/org/jetbrains/kotlin/js/translate/test/JSTestGenerator.kt
+++ b/js/js.translator/src/org/jetbrains/kotlin/js/translate/test/JSTestGenerator.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlin.js.backend.ast.*
 import org.jetbrains.kotlin.js.translate.callTranslator.CallTranslator
 import org.jetbrains.kotlin.js.translate.context.TranslationContext
 import org.jetbrains.kotlin.js.translate.reference.ReferenceTranslator
+import org.jetbrains.kotlin.js.translate.utils.name
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.FqNameUnsafe
 import org.jetbrains.kotlin.name.Name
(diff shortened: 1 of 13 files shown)


ilgonmic:
Thank you! Looks Good to Me
**One small note:** Just discovered difference between Adapter's contract in `kotlin-test` and `kotlin-test-js-runner`. In `kotlin-test-js-runner` we have `fn: () => void` but in `kotlin-test` `fn: () -> Any?` for `beforeEach` and `afterEach`. Correct variant is in `kotlin-test`.
Anyway this defect was in `suite` and `test` earlier, so it seems that minor issue

js/js.translator/src/org/jetbrains/kotlin/js/translate/test/JSTestGenerator.kt
```diff
@@ -72,62 +73,94 @@ class JSTestGenerator(val context: TranslationContext) {
...
         val descriptors = classDescriptor.unsubstitutedMemberScope
                 .getContributedDescriptors(DescriptorKindFilter.FUNCTIONS, MemberScope.ALL_NAME_FILTER)
 
-        val beforeFunctions = descriptors.filterIsInstance<FunctionDescriptor>().filter { it.isBefore }
-        val afterFunctions = descriptors.filterIsInstance<FunctionDescriptor>().filter { it.isAfter }
+        val innerContext = context.contextWithScope(suiteFunction)
+        val classVal = classDescriptor.instance(innerContext)
+        val testInstanceName = JsName("instance${classDescriptor.name.toString()}")
+        val testInstanceRef = testInstanceName.makeRef()
+        val classVar = JsVars.JsVar(testInstanceName, classVal)
```
anton-bannykh: This changes the way suites are instantiated. Previously a new instance was created for each test, now the suite instance is shared between tests.

I'm not sure this change is desirable (in my opinion it is not, as it reduces isolation). At the very least it requires motivation and discussion.

tokou:
Thanks for your review @anton-bannykh 

I did not foresee all these issues because of my lack of understanding of the codebase.

I'll rework it and try to implement the alternative approach that you mentioned (handling Promise's inside `kotlin.test` library itself) 👍

anton-bannykh:
@tokou Thank you your efforts. Feel free to ping me on kotlinlang.slack.com if you need any help. =)

PR #2745, opened 2019-10-30
KT-33297 Improve parameter name in `Add parameter to constructor` quick fix

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33297


mglukhikh:
Merged 👍

PR #2746, opened 2019-10-31
KT-33131 No indent before get/set method while editing extension var declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33131


vladimirdolzhenko:
LGTM.
Thanks.

PR #2747, opened 2019-10-31
Support inlining functions with KT-28064 style objects

pyos:
Namely, anonymous objects defined in lambdas that have all captured variables as loose fields instead of a single reference to the parent.

The question is, when a lambda inside an inline function defines an anonymous object, and that object is not regenerated during codegen for the inline function itself, but then has to be regenerated at call site anyway, do we use an outer `this` or loose capture fields? For example, before KT-28064:

    inline fun f1(g: () -> Unit) = object { g() }
    // -> f1$1 { $g: () -> Unit }
    inline fun f2(g: () -> Unit) = f1 { object { g() } }
    // -> f2$$inlined$f1$1 { $g: () -> Unit }
    //    f2$$inlined$f1$1$lambda$1 { this$0: f2$$inlined$f1$1 }
    inline fun f3(g: () -> Unit) = f2 { object { g() } }
    // -> f3$$inlined$f2$1 { $g: () -> Unit }
    //    f3$$inlined$f2$1$1 { this$0: f3$$inlined$f2$1 }
    //    f3$$inlined$f2$1$1$lambda$1 { this$0: f3$$inlined$f2$1$1 }

After KT-28064:

    inline fun f2(g: () -> Unit) = f1 { object { g() } }
    // -> f2$$inlined$f1$1 { $g: () -> Unit }
    //    f2$1$1 { $g: () -> Unit }
    inline fun f3(g: () -> Unit) = f2 { object { g() } }
    // -> f3$$inlined$f2$1 { $g: () -> Unit }
    //    f3$$inlined$f2$2 { ??? }
    //    f3$1$1 { $g: () -> Unit }

Should `???` be `this$0: f3$$inlined$f2$1` or `$g: () -> Unit`? This commit chooses the latter.

Part of #2643. Unlike the previous version of the commit, this one makes the backends compatible; you can take [this test](https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/boxInline/anonymousObject/defineClass.kt), compile 1.kt with -Xuse-ir and 2.kt without, and it will still work.


max-kammerer:
Many thanks! Great works! Sorry for excessive meticulousness!

PR #2748, opened 2019-11-01
Fix type of IllegalArgumentException call for illegal step in StepHandler

punzki:
This broke the K/N build which validates IR types (reported by @vvlevchenko).

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
@@ -15,16 +15,18 @@ import org.jetbrains.kotlin.ir.expressions.impl.IrCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
 import org.jetbrains.kotlin.ir.types.IrType
 import org.jetbrains.kotlin.ir.types.getClass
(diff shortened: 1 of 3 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/Utils.kt
```diff
@@ -23,9 +23,14 @@ internal fun IrExpression.castIfNecessary(targetType: IrType, numberCastFunction
     return if (type == targetType) {
         this
     } else {
-        val function = type.getClass()!!.functions.first { it.name == numberCastFunctionName }
-        IrCallImpl(startOffset, endOffset, function.returnType, function.symbol)
-            .apply { dispatchReceiver = this@castIfNecessary }
+        // This expression's type could be Nothing from an exception throw, in which case the number cast function will not exist.
+        val castFun = type.getClass()!!.functions.firstOrNull { it.name == numberCastFunctionName }
+        return if (castFun != null) {
```
max-kammerer: @punzki Is it possible to switch null check to Nothing type check that mentioned in comment?

punzki: Good idea. Done!

max-kammerer:
Many thanks!

PR #2749, opened 2019-11-01
Frontend: Avoid retypechecking the right hand side of assignment operators

sfs:
Right now, we typecheck the right hand side of assignment operators  `x += 1` twice if the left hand side is an assignable lvalue. This can lead to exponential slowdowns in some cases, i.e.,
```kotlin
fun main() {
  var x = 0
  x += try {
    x += try {
      x += ...
      x
    } finally {}
    x
  } finally {}
}
```

In simple cases we can avoid this by typechecking the right hand side before resolution. Note that typechecking `+=` is still slow, though, since there are so many resolution candidates for `plusAssign` in the standard library.

### DIFF
--- a/compiler/fir/resolve/tests/org/jetbrains/kotlin/fir/FirDiagnosticsSmokeTestGenerated.java
+++ b/compiler/fir/resolve/tests/org/jetbrains/kotlin/fir/FirDiagnosticsSmokeTestGenerated.java
@@ -9763,6 +9763,11 @@ public class FirDiagnosticsSmokeTestGenerated extends AbstractFirDiagnosticsSmok
             runTest("compiler/testData/diagnostics/tests/inference/functionPlaceholderError.kt");
         }
 
+        @TestMetadata("genericAssignmentOperator.kt")
+        public void testGenericAssignmentOperator() throws Exception {
+            runTest("compiler/testData/diagnostics/tests/inference/genericAssignmentOperator.kt");
+        }
+
         @TestMetadata("hasErrorInConstrainingTypes.kt")
         public void testHasErrorInConstrainingTypes() throws Exception {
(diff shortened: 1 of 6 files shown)


mglukhikh:
Looks good, @demiurg906 and me both think that we should accept this PR after remote run.

pyos:
This breaks the following case:

```kotlin
class R<T>

class A {
    operator fun plusAssign(x: R<Int>) {}
}

fun <T> f(): R<T> = R<T>()

fun main() {
    val x = A()
    x += f()
}
```

since the call that caches the result for `f()` does not resolve `plusAssign` and thus does not have enough dataflow information to infer `T`.

mglukhikh:
This PR changes behavior of three `DiagnosticsTestSpecGenerated`: `.../NotLinked/Dfa/Neg/test13, test26, test39`. Investigation is needed what's happening there. cc @dsavvinov @zarechenskiy

sfs:
Thanks Denis and @mglukhikh, I'll look into it!

sfs:
There were two problems. I used the wrong dataflow info, causing `test13`, `test26`, and `test39` to fail and I didn't type check the right hand side with `ContextDependency.DEPENDENT` (needed, since the rhs. corresponds to a value argument), which caused @pyos examples to fail.

I've updated the PR with fixes for this and added @pyos' example as a DiagnosticTest.

@mglukhikh could you please rerun the tests? Thanks!

---

One more thing though: In the code we replace the context scope in lines 233 and 248 before resolving the plusAssign and plus calls. I would be really grateful if someone could explain to me why this is needed - as far as I can tell this doesn't do anything in any of our tests.

mglukhikh:
@sfs I sent it to remote run

mglukhikh:
All tests passed without failures

mglukhikh:
@sfs Merged, thank you.

PR #2750, opened 2019-11-01
Always use '\n' as line separator in GenerateSteppedRangesCodegenTestData

punzki:
@max-kammerer - I believe this fixes the line separator issue on Windows. One thing I'm not sure of is if the other generators in the same directory are giving the same problem -- they don't explicitly force a line separator either.


max-kammerer:
@punzki Yes, really, seems I misunderstood @mglukhikh because he originally pointed me to this problem on Windows. Maybe @mglukhikh could clarify with which tests there is a problem on Windows and which problem.

mglukhikh:
To be precise, 12 original changed files are `illegalStepZero.kt` (three times), `illegalStepNotConst.kt` (x3), `illegalStepNegative.kt` (x3), `collectionIndices.kt`, `charSequenceIndices.kt`, `arrayIndices.kt`.

mglukhikh:
I tried to fix the problem just by commiting 12 changed files. However, Git failed to do this, and after it the problem disappeared. Looks like it was something local.

max-kammerer:
@punzki Sorry for disturbance!

punzki:
@max-kammerer @mglukhikh Not a problem. Let me know if it pops up again.

PR #2751, opened 2019-11-01
[JVM_IR] Reintroduce non-IEEE comparisons

kandersen:
- Ieee754Equality and PrimitiveComparisons intrinsics
  respects absence of -Xproper-ieee754-comparisons with
  unmuted tests to show.


kandersen:
@udalov Have a look, I think it turned out quite neat.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -138,3 +137,12 @@ fun JvmBackendContext.createJvmIrBuilder(
 
 fun IrDeclaration.isInCurrentModule(): Boolean =
     getPackageFragment() is IrFile
+
+// Determine if the IrExpression is smartcast, and if so, if it is cast from higher than nullable target types.
+// This is needed to pinpoint exceptional treatment of IEEE754 floating point comparisons, where proper IEEE
+// comparisons are used "if values are statically known to be of primitive numeric types", taken to mean as
+// "not learned through smartcasting".
+fun IrExpression.isSmartcastFromHigherThanNullable(context: JvmBackendContext) =
+    (this as? IrTypeOperatorCall)?.let {
```
udalov: A bit cleaner would be
```
    this is IrTypeOperatorCall &&
            operator == IrTypeOperator.IMPLICIT_CAST &&
            !argument.type.isSubtypeOf(type.makeNullable(), context.irBuiltIns)
```

Also, I'd suggest expanding the last condition into:
```
            !argument.type.isNullable() && !argument.type.isSubtypeOf(type, context.irBuiltIns)
```
since it seems a bit easier to understand and checks the same (please correct me if I'm wrong)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/CompareTo.kt
```diff
@@ -84,6 +86,29 @@ class BooleanComparison(val op: IElementType, val a: MaterialValue, val b: Mater
...
+        when (a.type) {
+            Type.FLOAT_TYPE -> mv.invokestatic("java/lang/Float", "compare", "(FF)I", false)
+            Type.DOUBLE_TYPE -> mv.invokestatic("java/lang/Double", "compare", "(DD)I", false)
+            else -> throw UnsupportedOperationException()
+        }
+        mv.visitJumpInsn(numberCompareOpcode, target)
+    }
+
+    override fun jumpIfTrue(target: Label) {
+        when (a.type) {
```
udalov: Please extract the common part with `jumpIfFalse`

udalov:
@dnpetrov Could you please also take a look? I'm a bit unsure about the smart cast detection logic here, in particular if there's any way to observe a smart cast other than `IMPLICIT_CAST` type operator (which I actually suggested to Kristoffer in the first place).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
kandersen: If we're casting from Any? to Float?,  that is indeed "too high" a smart-cast: Any? is _not_ a subtype of Float?.makeNullable().

But the new definition OK's it as Any? is nullable.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/CompareTo.kt
kandersen: Done, thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
udalov: You're right, sorry. I intended the following:
```
            (!argument.type.isNullable() || !argument.type.isSubtypeOf(type, context.irBuiltIns))
```

But then it's not much easier than the original anymore, so never mind. :)

udalov:
@kandersen Thank you!

kandersen:
> @kandersen Thank you!

Right back at you! :)

PR #2753, opened 2019-11-01
Fir & IrTextTests: Produce IR_TEMPORARY_VARIABLE origin

sfs:
The JVM IR backend relies on the IR_TEMPORARY_VARIABLE origin to identify compiler introduced variables. Such variables can be freely optimized away and don't get an entry in the local variables table so as not to appear in the debugger.

This PR changes Fir2Ir to produce temporary variables for all variables with a special name (i.e., names wrapped in angle brackets).

Additionally, I'm changing IrTextTests to normalize the names of temporary variables in order to better compare the outputs of psi2ir and fir2ir.

There's also a small change to actually regenerate missing files in IrTextTests with the correct contents, instead of producing a file containing an error message. That may have been there on purpose though, so I'd be really grateful if someone could confirm this change...


dnpetrov:
Looks ok to me.
Actually, it'd be fine to drop name hints from temporarily variables.

sfs:
Couple of small changes:
- Rebased on master
- Squashed the IrJsTextTest fixes into the first commit
- Changed the naming scheme for temporary variables to never reuse names (based the name on the variable symbol instead of the variable name)

mglukhikh:
@sfs Thank you!

PR #2755, opened 2019-11-02
KT-31967 Typo in inspection name: "'+=' create new list under the hood"

tommykw:
https://youtrack.jetbrains.com/issue/KT-31967


mglukhikh:
Merged 👍

PR #2756, opened 2019-11-04
[JVM_IR] Fix interface method resolution

kandersen:
The JVM IR was too agressive in delegating statically to `$DefaultImpls`
class.

Consider the following library, compiled separately, and then client,
compiled against that library:

```
//library
interface A { @JvmDefault fun foo() = "A" }

interface Left : A { }
interface Right : A { @JvmDefault override fun foo() = "Right" }

//client
interface C : Left, Right {}

fun main() {
  val x = object : C {}
  println(x.foo())
}
```

Previously, the IR backend generates an overriding bridge in `C`, which
calls `C$DefaultImpls` which calls statically into `Right$DefaultImpls`.

When then library is recompilerd with an override of foo in Left, the
existing binary of the client continues to run against the recompiled
library erroneously running `Right.foo()`.

The old backend throws an IncompatibleClassChange exception due to
ambiguity of multiple inherited default implementations. kotlinc indeed
rejects the client when recompiled against the new library.

We have no tests constraining this behaviour, but we could conceivably
strap something together in CustomBinaries tests.


kandersen:
@max-kammerer Continuation of #2612, I suppose. I investigated the issues of binary incompatibility that you pointed out resulted in the IR backends' design of going without accessors, and this is a prelude to switching the delegation around to be in line with the old backend.

max-kammerer:
Seems it's possible to add test similar to compiler/testData/codegen/bytecodeText/jvm8/jvmDefault/compatibility/simpleFunction.kt with inherited class instead of interface

max-kammerer:
Looks like that `.+` and '.*' could be simply avoided in old scheme, so changes in AbstractWriteFlagsTest looks redundant

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -191,48 +192,52 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
...
+        // Finally, we need to produce accessors for calls to protected static methods coming from Java,
+        // which we put in the closest enclosing class which has access to the method in question.
+        val dispatchReceiverType = expression.dispatchReceiver?.type
+        val parent = source.accessorParent(dispatchReceiverType?.classOrNull?.owner ?: source.parent)
+
         return buildFun {
             origin = JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR
             name = source.accessorName()
             visibility = Visibilities.PUBLIC
+            modality = if (parent is IrClass && parent.isJvmInterface) Modality.OPEN else Modality.FINAL
```
max-kammerer: Why Open modality is required? Accessor is static anyway

kandersen: It's going on an interface, where you can't place `final` things.

kandersen:
> Looks like that `.+` and `.*` could be simply avoided in old scheme (or replaced with `.`), so changes in AbstractWriteFlagsTest looks redundant

I'm afraid I don't follow -- the old scheme used exact matches: `name.equals(..)`. The alternative is templates?

max-kammerer:
My point that all templates in new scheme could be silently removed in old one, or I miss smth?

kandersen:
So, in e.g. `compiler/testData/writeFlags/jvm8/defaults/compatibility/propertyAccessors.kt`, the test is affirming the presence of the accessors. The old backend generated a particular set of names for these, so the test simply asserted that a method of name ` access$getTest$jd` should be present.

The new backend numbers these, so it _does_ generate the accessor, but names it `access$getTest$0`.

The AbstractWriteFlagsTest would match the name found in the bytecode with the name in the test using `String.equals`.

I.e. the test fails on the new backend. 

I generalized the test to use regexps in order to express the presence of the accessor on both backends, but without heed to the actual name of the accessor.

I feel like _I_ am missing something :smile:

max-kammerer:
Seems we can just cut old string to `access$getTest$`

kandersen:
> Seems we can just cut old string to `access$getTest$`
The old backend then fails the test (`"access$getTest$".equals("access$getTest$jd") == false`).

Are you perhaps conflating the bytecode text tests with writeflags? Because _they_ are indeed already reg-exp based.

max-kammerer:
> Are you perhaps conflating the bytecode text tests with writeflags?

Yes, sorry

kandersen:
> > Are you perhaps conflating the bytecode text tests with writeflags?
> 
> Yes, sorry

No worries! You _do_ seem to be right in that we can drop the `.+` as reg-exp matching by default matches the prefix. Will push a revision in a sec once I confirm this.

max-kammerer:
But actually it's a bit dangerous to use templates here, at least i would prefer to have exactly same prefixes here for accessors for compatibility mode or to have //JVM  and //JVM_IR sections like in bytecode text tests.
Seems we can tune `accessorName` function a bit to produce same bytecode

compiler/testData/codegen/bytecodeText/jvm8/jvmDefault/compatibility/simpleFunctionWithAbstractOverride.kt
```diff
@@ -18,7 +17,7 @@ interface KInterface2 : KInterface  {
 // +
 // 0 INVOKESTATIC KInterface2.access\$test2\$jd
 // =
-// 1 INVOKESTATIC
+// 1 INVOKESTATIC KInterface
```
kandersen: The test was over-specified: the IR backend produces a nullcheck on `this` in `KInterface$DefaultImpls$test2` which is a static method.

kandersen:
@max-kammerer Got the `$s` and `$jd` suffixes working. The (companion) property suffix seems within reach too, perhaps by looking at the origin of the parent classes of the function being accessed.

max-kammerer:
@kandersen Is it possible to change order in too last commits to avoid unnecessery (and temporary within pr) test data change in "IR: Align Interface Defaults with Old Backend"?

kandersen:
> @kandersen Is it possible to change order in too last commits to avoid unnecessery (and temporary within pr) test data change in "IR: Align Interface Defaults with Old Backend"?

I am sure! I will get to it!

kandersen:
@max-kammerer OK, all done for now, I think, please take a look again. I think this is a good place to end this PR - we can have a look at the remaining accessor kinds (from `syntheticAccessorUtil.kt`):

```
    FIELD_FROM_LOCAL("lp")
    LATEINIT_INTRINSIC("li")
```

Discovered a few bugs in old and new backend that I will file ASAP, as well.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -443,21 +445,60 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
...
+        nameCounterMap[function] = next + 1
+        return next
+    }
 
-    private fun IrFunction.accessorName(): Name {
+    private fun IrFunction.accessorName(superQualifier: IrClassSymbol?): Name {
         val jvmName = context.methodSignatureMapper.mapFunctionName(this)
-        return Name.identifier("access\$$jvmName\$${nameCounter++}")
+        val suffix = when {
+            // The only function accessors placed on interfaces are for JvmDefault implementations
```
max-kammerer: Actually there could be private JvmDefault function than we need accessor for it. But we still can use $jd suffix cause there would no any accessor for it in compatibility mode. Please mention it in comment

max-kammerer: Or we can just check private visibility here and use default suffix in such case

max-kammerer: Please also add test for such case

kandersen: You are thinking something like the following?
```
interface I {
  @JvmDefault
  private fun foo() = 4

  fun bar() = {foo() + 5}()
}
```
The old and new backend actually differ on where to place the accessor. The old backend accesses `foo` via `I`, so the accessor goes there. IR goes via `I$DefaultImpls`. Might it be due to inlining of lambdas?

> Or we can just check private visibility here and use default suffix in such case

When you say "default" you mean the sequence number, "s+hashcode" or do you mean "jd"?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -108,15 +115,18 @@ internal class InterfaceLowering(val context: JvmBackendContext) : IrElementTran
                 Visibilities.isPrivate(function.visibility)
```
max-kammerer: It's true for private non-JvmDefault, private JvmDefault should be keeped in interface

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
max-kammerer: It should be same accessor as in 
```
class I {

  private fun foo() = 4

  fun bar() = {foo() + 5}()
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
kandersen: Ah, but now we're evolving the feature, right? Private interface members require Java 9, AFAIK, so we should be guarding it with target JVM.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -443,21 +445,60 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
...
+            // Accessor for _s_uper-qualified call
+            superQualifier != null -> "s" + superQualifier.descriptor.name.asString().hashCode()
+
+            // Access to static members that need an accessor must be because they are inherited,
+            // hence accessed on a _s_upertype.
+            isStatic -> "s" + hashForAccessorDisambiguation()
+
+            else -> drawNextNameSuffix(this)
+        }
+        return Name.identifier("access\$$jvmName\$$suffix")
```
max-kammerer: Is it possible to avoid `$0` generation? So simple cases would be similar to old backend

kandersen: The test suite says "yes" :) Let me push the revision.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
max-kammerer: JvmDefault requires target 8. Java 9 supports private interace members, but already JVM 8 allows to use private members in bytecode: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.6

kandersen:
@max-kammerer Pushed the fix for private JvmDefaults. Also, I think kt11969.kt, now unmuted, sufficiently covers that exact case?

max-kammerer:
Yes, except accessor suffix check for private JvmDefault) https://github.com/JetBrains/kotlin/pull/2756#discussion_r352025790

Could you also rebase on master?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -442,24 +445,55 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
...
-    private var nameCounter = 0
-
-    private fun IrFunction.accessorName(): Name {
+    private fun IrFunction.accessorName(superQualifier: IrClassSymbol?): Name {
         val jvmName = context.methodSignatureMapper.mapFunctionName(this)
-        return Name.identifier("access\$$jvmName\$${nameCounter++}")
+        val suffix = when {
+            // The only function accessors placed on interfaces are for private functions and JvmDefault implementations.
+            // The two cannot clash.
+            parentAsClass.isJvmInterface && Visibilities.isPrivate(visibility) -> ""
```
max-kammerer: Not sure, but maybe just check visibility otherwise it would be default case anyway
```parentAsClass.isJvmInterface && !isPrivate(visibility) -> "\jd"``` ?

Comment (kandersen) [reply compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt]: Then it will hit the `isStatic` case, though. I am not too certain on any of these, but they are either right or the test-suite does not have sufficient coverage ;)

**Comment by kandersen** [comment]
Rebased!

File: compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt (hunk 6 above)
Comment (max-kammerer) [reply compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt]: Yes, indeed )

**Comment by max-kammerer** [comment]
Thanks you for pr!  

Actually there is still small problem with delegation between DefaultImpls:
```
                function.origin == IrDeclarationOrigin.FAKE_OVERRIDE -> {
                    val implementation = function.resolveFakeOverride()!!
                         .....  
                        !implementation.hasJvmDefault() -> {
                            val defaultImpl = createDefaultImpl(function)
                            context.declarationFactory.getDefaultImplsFunction(implementation).also {
                                defaultImpl.bridgeToStatic(it)
                            }
                        }
                        // else -> Do nothing.
                    }
                }
```

In such case C has wrong delegated call to A, but should has delegated call to B$DefaultImpls
```
interface A {
    fun foo() {}
}

interface B : A  //INVOKESTATIC zzz/A$DefaultImpls.foo (Lzzz/A;)V 

interface C : B  //INVOKESTATIC zzz/A$DefaultImpls.foo (Lzzz/A;)V
```

PR #2757, opened 2019-11-04
JVM IR: Avoid getTopLevelClass for performance and Fir

sfs:
We recently saw a performance regression after merging #2657 on a small benchmark where deserializing `java.lang.String` ended up being a bottleneck. There are currently several symbols in JvmSymbols which we hardcode and some which are deserialized from disk instead (and in Fir mode we always have to hardcode all dependencies).

This PR makes JvmSymbols consistent by hardcoding all dependencies for the IR backend, which results in slightly better performance and allows us to remove the special cases for the Fir backend.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendContext.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendContext.kt
@@ -107,13 +107,6 @@ class JvmBackendContext(
 
     val inlineClassReplacements = MemoizedInlineClassReplacements()
 
-    internal fun getTopLevelClass(fqName: FqName): IrClassSymbol {
-        val descriptor = state.module.getPackage(fqName.parent()).memberScope.getContributedClassifier(
-            fqName.shortName(), NoLookupLocation.FROM_BACKEND
-        ) as ClassDescriptor? ?: error("Class is not found: $fqName")
-        return referenceClass(descriptor)
-    }
-
     internal fun referenceClass(descriptor: ClassDescriptor): IrClassSymbol =
         symbolTable.referenceClass(descriptor)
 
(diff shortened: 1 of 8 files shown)


udalov:
Do you maybe have specific numbers, how the execution time changed on this benchmark (and what this benchmark consists of)?

sfs:
Sure, the benchmark numbers are [on Mako](https://mako.dev/benchmark?benchmark_key=5336056912150528&xspread=1&~Kotlinc=mean&~KotlincUseIr=mean&~KotlincUseFir=mean&order=bid), revision number 54373. The change in question broke fir for a while, which is why the next data-point for fir is for revision 54412 - note that there is no performance degradation in fir mode.

The benchmark itself consists of a simple abstract class with 500 subclasses which each override the same method and a main file which instantiates all classes and calls the overridden method.

max-kammerer:
@sfs I'm curious, is this change affects test execution time?

sfs:
@max-kammerer I think so. I'm seeing the IR tests (`gradlew :compiler:test --tests '*.ir.*'`, excluding jar creation) run in a total time of about 1h13min without the PR and in 1h4m with the PR (but that amounts to 3 minutes real time compared to ~2:45 minutes real time, so take these numbers with a pinch of salt).

max-kammerer:
@sfs could you run test via ```gradlew :compiler:test --tests '*.ir.*' -Pkotlin.test.maxParallelForks=X``` where X is 2, 4 and 8. Looks like I will get similar amount of real time with much less CPU time

sfs:
@max-kammerer Thanks for the pointers! The sweet spot over here seems to be 14 threads, bringing the test time down to 2:22 minutes (compared to 2:41 on master).

max-kammerer:
@sfs And how many cores do you have (are they hyperthreaded or not?)? Seems we should change default for parallelization

udalov:
@sfs Thanks for the contribution!

mglukhikh:
Looks like https://youtrack.jetbrains.com/issue/KT-34496 can be closed now. Thank you again!

PR #2759, opened 2019-11-05
JVM_IR: fix NPE in interface companion initializers

pyos:
* When referencing the companion itself, they should use the $$INSTANCE field, not the (null until \<clinit\> returns) Companion field of the interface.

* Interface companion init blocks should be made static.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
@@ -45,5 +45,6 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD", isSynthetic = true)
     object GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD : IrDeclarationOriginImpl("GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD", isSynthetic = true)
     object SUSPEND_FUNCTION_VIEW : IrDeclarationOriginImpl("SUSPEND_FUNCTION_VIEW")
-    object SUSPEND_IMPL_STATIC_FUNCTION :  IrDeclarationOriginImpl("SUSPEND_IMPL_STATIC_FUNCTION", isSynthetic = true)
+    object SUSPEND_IMPL_STATIC_FUNCTION : IrDeclarationOriginImpl("SUSPEND_IMPL_STATIC_FUNCTION", isSynthetic = true)
+    object INTERFACE_COMPANION_PRIVATE_INSTANCE : IrDeclarationOriginImpl("INTERFACE_COMPANION_PRIVATE_INSTANCE", isSynthetic = true)
 }
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
@@ -34,6 +34,7 @@ class JvmDeclarationFactory(
     private val methodSignatureMapper: MethodSignatureMapper
 ) : DeclarationFactory {
     private val singletonFieldDeclarations = HashMap<IrSymbolOwner, IrField>()
+    private val interfaceCompanionFieldDeclarations = HashMap<IrSymbolOwner, IrField>()
     private val outerThisDeclarations = HashMap<IrClass, IrField>()
     private val innerClassConstructors = HashMap<IrConstructor, IrConstructor>()
 
@@ -133,6 +134,23 @@ class JvmDeclarationFactory(
                 parent = if (isNotMappedCompanion) singleton.parent else singleton
             }
         }
+
+    fun getPrivateFieldForObjectInstance(singleton: IrClass): IrField =
+        if (singleton.isCompanion && singleton.parentAsClass.isJvmInterface)
+            interfaceCompanionFieldDeclarations.getOrPut(singleton) {
+                buildField {
+                    name = Name.identifier("\$\$INSTANCE")
+                    type = singleton.defaultType
+                    origin = JvmLoweredDeclarationOrigin.INTERFACE_COMPANION_PRIVATE_INSTANCE
+                    isFinal = true
+                    isStatic = true
+                    visibility = JavaVisibilities.PACKAGE_VISIBILITY
+                }.apply {
+                    parent = singleton
+                }
+            }
+        else
+            getFieldForObjectInstance(singleton)
 
     fun getDefaultImplsFunction(interfaceFun: IrSimpleFunction): IrSimpleFunction {
         val parent = interfaceFun.parentAsClass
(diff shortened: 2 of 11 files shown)


max-kammerer:
`getPrivateFieldForObjectInstance` is used only from jvm specific lowers, so it's not neccessery to add it to common interface

max-kammerer:
Sorry, missed `LowerUtils.kt` change

max-kammerer:
But actually 'ReplaceThisByStaticReference' is used only in JVM backend

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -133,6 +135,23 @@ class JvmDeclarationFactory(
             }
         }
 
+    override fun getPrivateFieldForObjectInstance(singleton: IrClass): IrField =
+        if (singleton.isCompanion && singleton.parentAsClass.isJvmInterface)
+            interfaceCompanionFieldDeclarations.getOrPut(singleton) {
+                buildField {
+                    name = Name.identifier("\$\$INSTANCE")
+                    type = singleton.defaultType
+                    origin = IrDeclarationOrigin.FIELD_FOR_OBJECT_INSTANCE
```
max-kammerer: It should be `synthetic`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SingletonReferencesLowering.kt
```diff
@@ -52,7 +52,10 @@ private class SingletonReferencesLowering(val context: JvmBackendContext) : File
     }
 
     override fun visitGetObjectValue(expression: IrGetObjectValue): IrExpression {
-        val instanceField = context.declarationFactory.getFieldForObjectInstance(expression.symbol.owner)
+        val instanceField = if (allScopes.any { it.irElement == expression.symbol.owner })
```
max-kammerer: Looks like `$$INSTANCE` field should be used only in companion <clinit>, otherwise it could be exposed by inline through synthetic accessor:
```
interface Test {
    companion object {
        fun ok() = "OK"
        
        inline fun test() = run {
            Test.ok()
        }
    }
}
```

pyos: Nice catch, in this case it should in fact be the dispatch receiver of `test`, so neither the original nor the changed code is entirely correct.

pyos: ...though it actually works with `$$INSTANCE` as well because an accessor for it is created.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: Added a new origin.

pyos:
>But actually 'ReplaceThisByStaticReference' is used only in JVM backend

Yeah, looks like it's not referenced by K/N either, so I moved it to JVM IrUtils.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
```diff
@@ -45,5 +45,6 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD", isSynthetic = true)
     object GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD : IrDeclarationOriginImpl("GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD", isSynthetic = true)
     object SUSPEND_FUNCTION_VIEW : IrDeclarationOriginImpl("SUSPEND_FUNCTION_VIEW")
-    object SUSPEND_IMPL_STATIC_FUNCTION :  IrDeclarationOriginImpl("SUSPEND_IMPL_STATIC_FUNCTION", isSynthetic = true)
+    object SUSPEND_IMPL_STATIC_FUNCTION : IrDeclarationOriginImpl("SUSPEND_IMPL_STATIC_FUNCTION", isSynthetic = true)
+    object INTERFACE_COMPANION_INSTANCE : IrDeclarationOriginImpl("INTERFACE_COMPANION_INSTANCE", isSynthetic = true)
```
max-kammerer: INTERFACE_COMPANION_INTERMEDIATE/PRIVATE?_INSTANCE  - can't suggest smth better, any ideas?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -133,6 +135,23 @@ class JvmDeclarationFactory(
...
+    fun getPrivateFieldForObjectInstance(singleton: IrClass): IrField =
+        if (singleton.isCompanion && singleton.parentAsClass.isJvmInterface)
+            interfaceCompanionFieldDeclarations.getOrPut(singleton) {
+                buildField {
+                    name = Name.identifier("\$\$INSTANCE")
+                    type = singleton.defaultType
+                    origin = JvmLoweredDeclarationOrigin.INTERFACE_COMPANION_INSTANCE
+                    isFinal = true
+                    isStatic = true
+                    visibility = JavaVisibilities.PROTECTED_AND_PACKAGE
```
max-kammerer: Please check, but seems old backend uses PACKAGE visibility

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
pyos: Yeah, same problem here. ..._PRIVATE_... has at least some meaning here and is consistent with the JvmDeclarationFactory method name, so I'll use that one.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: Yes, there should be no access flag. For some reason I thought PROTECTED_AND_PACKAGE translates to that, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -153,57 +160,12 @@ private class MoveOrCopyCompanionObjectFieldsLowering(val context: CommonBackend
...
             ).apply {
                 parent = newParent
-                body = oldInitializer.body.transferToNewParent(oldParent, newParent)
+                body = oldInitializer.body
+                    .replaceThisByStaticReference(context, oldParent, oldParent.thisReceiver!!)
+                    .patchDeclarationParents(newParent) as IrBlockBody
             }
         }
 
-    private fun IrBlockBody.transferToNewParent(oldParent: IrClass, newParent: IrClass): IrBlockBody {
```
pyos: This also fixes a different bug, as the transformed was missing `override fun visitSetVariable`.

pyos:
@max-kammerer ping?

max-kammerer:
@pyos Sorry, will look soon

max-kammerer:
Sorry for delay. 
Thanks for PR!

PR #2760, opened 2019-11-05
JVM IR: Fix types in LateInitLowering

sfs:
On the JVM backend the `throwUninitializedPropertyAccessException` function is implemented in Java and has a return type of `Unit` (as far as Kotlin is concerned) instead of `Nothing`. In order to make the types match up we have to wrap its call in a block with an additional `IrThrow`.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LateinitLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LateinitLowering.kt
@@ -21,7 +21,6 @@ import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.descriptors.WrappedVariableDescriptor
 import org.jetbrains.kotlin.backend.common.ir.Symbols
 import org.jetbrains.kotlin.ir.IrStatement
-import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.builders.*
 import org.jetbrains.kotlin.ir.builders.declarations.buildField
 import org.jetbrains.kotlin.ir.declarations.*
@@ -36,7 +35,7 @@ import org.jetbrains.kotlin.ir.util.resolveFakeOverride
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 
-class LateinitLowering(val context: CommonBackendContext) : FileLoweringPass {
+open class LateinitLowering(val context: CommonBackendContext) : FileLoweringPass {
 
     private val nullableFields = context.lateinitNullableFields
     private fun buildOrGetNullableField(originalField: IrField): IrField {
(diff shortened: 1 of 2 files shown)


max-kammerer:
Is it possible to add some test? How did you catch this problem?

sfs:
I'm not sure how to test this. I was refactoring `ExpressionCodegen` and got some compilation errors, but the current `ExpressionCodegen` is loose enough that it will quietly ignore errors like this.

We could check `IrTypes` using `IrValidator`, which we might want to turn on during (some) tests. What do you think?

max-kammerer:
Just some thoughts: seems same should be applied for `checkNotNullExpressionValue` and `checkExpressionValueIsNotNull.` As they are hardcoded declaration could we switch type in declaration to `Nothing` then codegen will insert `throw` by itself?
And yes we need some validation: I asked in internal channel about best approach.

max-kammerer:
As I see with new approach there would be additional throw instruction in bytecode compare to old backend. Could you clarify what exactly problem  do you have?

sfs:
Currently, the IR backend compiles something like `lateinit var v: String` to code such as this:
```java
  public final getV()Ljava/lang/String;
    ALOAD 0
    GETFIELD A.v : Ljava/lang/String;
    ASTORE 1
    ALOAD 1
    IFNULL L1
    ALOAD 1
    ARETURN
   L1
    LDC "v"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.throwUninitializedPropertyAccessException (Ljava/lang/String;)V
    GETSTATIC kotlin/Unit.INSTANCE : Lkotlin/Unit;
    CHECKCAST java/lang/Void
    POP
    ACONST_NULL
    ARETURN
```
With these changes we generate the following code:
```java
  public final getV()Ljava/lang/String;
    ALOAD 0
    GETFIELD A.v : Ljava/lang/String;
    ASTORE 1
    ALOAD 1
    IFNULL L1
    ALOAD 1
    ARETURN
   L1
    LDC "v"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.throwUninitializedPropertyAccessException (Ljava/lang/String;)V
    ACONST_NULL
    ATHROW
```
The problem was that in the Ir, we have an `if` statement with return type `Void`, where one of the branches instead returns `Unit`, by calling `throwUninitializedPropertyAccessException`. At that point we ask `StackValue.coerce` to first materialize the unit value and then to cast it to `Nothing`, which then gets discarded because the `If` doesn't return a value. Next we see that we have reached the end of a non-void returning function which doesn't end in a return and so we add the default value for the return type and return it.

There's a lot wrong with this sequence of steps. We generate broken bytecode (the malformed cast) and hide errors (the return default value clause), but as is we cannot fix this because we actually did generate broken Ir in the first place. This commit is mainly there to make headway towards fixing `ExpressionCodegen`.

sfs:
By the way, if the additional throw is an issue we could also generate Ir such as this for access to a latinit var `v : String?`):
```kotlin
{
  if (v == null) { throwUninitializedPropertyAccessException(...)  }
  v as String
}
```
where the cast is implicit. That more or less matches what the old backend generates. Do you want me to go ahead and make this change?

max-kammerer:
@sfs Thanks for detailed clarification! For now I would prefer to keep old scheme with originally suggested fix.

JVM-IR is validated before and after all lowerings. Seems it's a bug in validators. It could be fixed as separate pr so I could merge this one or wait for fix

sfs:
@max-kammerer You're right, we are running the validator, but apparently [without checking types](https://github.com/JetBrains/kotlin/blob/4164b620e8778c2e0f7877753b527e714659d076/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/phaser/DumperVerifier.kt#L130). There are probably more type errors in the ir...

max-kammerer:
Thank you!

PR #2761, opened 2019-11-05
JVM IR: Implement CHECK_NOT_NULL as a lowering

sfs:
Currently, we generate calls to `CHECK_NOT_NULL` for `!!` expressions and generate the actual code for the check in an intrinsic. This PR instead removes `CHECK_NOT_NULL` in an early lowering, turning it into ordinary Ir constructs and moving special cases outside of codegen.

Since the `RedundantNullCheckMethodTransformer` checks that the resulting bytecode has a specific shape (and this optimization needs to run on the bytecode because of inlining) I had to extend it to handle the simplified form generated from the Ir instead of the hardcoded bytecode generated by the intrinsic.


udalov:
Thanks! I've moved the comment on the class into the phase description, and merged manually in https://github.com/JetBrains/kotlin/commit/dcd72b06d85b25b78b4885445f106a7d82574e03

PR #2762, opened 2019-11-05
KT-16886 Run configuration name should not contain FQN

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16886


t-kameyama:
8477c1935c2a09039f0d668706f865cb7521235c

PR #2763, opened 2019-11-06
JVM: Always invoke get()/charAt() in optimized for-loop over CharSequence.withIndex()

punzki:
...even if element variable is unused in destructuring declaration.


dnpetrov:
Thank you for your contribution! I'll run tests and merge manually.

PR #2765, opened 2019-11-06
JVM_IR: move lambda captures to end of signature when inlining

pyos:
For example, a lambda `{ param -> captured }` of type `E.(T) -> U` will be transformed by LocalDeclarationsLowering into a private static method

    fun f$lambda-0($this: E, $captured: U, param: T) = $captured

The reason for such an ordering is that a lambda looks the same as a local function, and local function can have default arguments, and those arguments can reference captured variables; thus, captured variables must come before actual declared arguments.

However, this is not the order that the inliner wants. Moreover, since it was written to handle lambdas represented as `invoke` methods of anonymous objects, it does not expect the actual callable method to have any parameters corresponding to captured variables at all. This results in it attempting to generate a temporary node with descriptor

    (LE;LU;LT;LU;)LU;

while still using locals 1 and 2 as `param` and `$captured` respectively. In the example above, this is not critical, as they both have reference type and the lambda will eventually be pasted into a different node anyway; however, if it happens that one of them is a primitive, or both are primitives of different types, the bytecode will use incorrect instructions, causing verification errors. The correct descriptor is

    (LE;LT;LU;)LU;

Part of #2643, changed not to use a flag since `LambdaInfo.invokeMethod` behaves the same way in both backends.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
@@ -413,9 +413,19 @@ class MethodInliner(
 
         val capturedParamsSize = parameters.capturedParametersSizeOnStack
         val realParametersSize = parameters.realParametersSizeOnStack
+        val reorderIrLambdaParameters = inliningContext.isInliningLambda &&
+                inliningContext.parent?.isInliningLambda == false &&
+                inliningContext.lambdaInfo is IrExpressionLambda
+        val newArgumentList = if (reorderIrLambdaParameters) {
+            // In IR lambdas, captured variables come before real parameters, but after the extension receiver.
+            // Move them to the end of the descriptor instead.
+            Type.getArgumentTypes(inliningContext.lambdaInfo!!.invokeMethod.descriptor) + parameters.capturedTypes
+        } else {
+            Type.getArgumentTypes(node.desc) + parameters.capturedTypes
+        }
         val transformedNode = MethodNode(
             Opcodes.API_VERSION, node.access, node.name,
-            Type.getMethodDescriptor(Type.getReturnType(node.desc), *(Type.getArgumentTypes(node.desc) + parameters.capturedTypes)),
+            Type.getMethodDescriptor(Type.getReturnType(node.desc), *newArgumentList),
             node.signature, node.exceptions?.toTypedArray()
         )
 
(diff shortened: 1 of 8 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
```diff
@@ -413,9 +413,18 @@ class MethodInliner(
 
         val capturedParamsSize = parameters.capturedParametersSizeOnStack
         val realParametersSize = parameters.realParametersSizeOnStack
+        val newArgumentList = if (node != inliningContext.lambdaInfo?.node?.node) {
```
max-kammerer: I would prefer to make smth like this:
```
        val isInsideInliningLambda = inliningContext.isInliningLambda
        val reorderParametersForRootIrLambdaInvoke =
            isInsideInliningLambda &&
                    inliningContext.parent?.isInliningLambda == false &&
                    inliningContext.lambdaInfo is IrExpressionLambda
        val newArgumentList = if (reorderParametersForRootIrLambdaInvoke) {
            // Inlining a lambda. Depending on the representation, the node may or may not have the captures as
            // parameters already (`invoke` of an object, for example, does not, as captures are passed to the
            // constructor; private static methods like ones generated by the IR backend do). `invokeMethod`,
            // on the other hand, is known to never have them.
            Type.getArgumentTypes(inliningContext.lambdaInfo!!.invokeMethod.descriptor) + parameters.capturedTypes
        } else {
            Type.getArgumentTypes(node.desc) + parameters.capturedTypes
        }
```
and then...

max-kammerer: .... change 
```
 if (inliningContext.isInliningLambda && lambdaInfo is IrExpressionLambda) {
```
below to
```
if (reorderParametersForRootIrLambdaInvoke && lambdaInfo is IrExpressionLambda)
```
and substitute to

```private val isInliningLambda = nodeRemapper.isInsideInliningLambda```  with isInsideInliningLambda

max-kammerer:
> thus, captured variables must come before actual declared arguments.
Actually I don't see problem in switching order (captures can't have defaults)

pyos:
>I don't see any problem in switching order (captures can't have defaults)

Captures can't, real parameters can, which is why real parameters must be *after* captures.

It is true, however, that parameters of *lambdas* can't have default values. However, this means either a special case in the common LocalDeclarationsLowering, or a separate JVM-only lowering for moving the parameters of lambdas at the IR level. And even then, changing the inliner would still be required to avoid duplicating the captures.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
pyos: >and substitute `private val isInliningLambda = nodeRemapper.isInsideInliningLambda` with isInsideInliningLambda

Not sure what you mean here. Also, this variable doesn't seem relevant to the PR, so I'd prefer to avoid touching it.

max-kammerer:
Yes, it would be enough to have changes only in inliner.
But in any way, I actually don't see any problem in next construction from bytecode point of view:
fun foo(s: String = z, z: String) 
On initializing `s` with default value proper initialized value for z already would be exist on stack

pyos:
The JVM bytecode for this would be correct, yes, but I don't know if this sort of construct is valid IR in general, and would be handled well by JS/Native in particular. Since the IR is technically more of an AST, I prefer to assume that anything disallowed by the language could lead to weird consequences.

max-kammerer:
>> Since the IR is technically more of an AST, I prefer to assume that anything disallowed by the language could lead to weird consequences

Yes, actually it should be discussed with other backend teams.
Thank you for pr!

PR #2767, opened 2019-11-06
JVM, JVM IR: Don't optimize null-checks based on nullability information

sfs:
I was a bit overeager in optimizing null checks as part of #2539. This PR backpedals a bit and only optimizes null checks where required for correctness.

In general we cannot remove a check of the form `x == null` just because we know that `x` has a non-nullable type, since Java interop allows us to circumvent the type system. And of course this happens in practice through dependency injection.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
@@ -3871,12 +3871,11 @@ public class ExpressionCodegen extends KtVisitor<StackValue, StackValue> impleme
         Type type = expressionType(exp);
         StackValue argument = pregeneratedExpr != null ? pregeneratedExpr : gen(exp);
 
-        if (kotlinType == null || TypeUtils.isNullableType(kotlinType)) {
+        if (kotlinType == null ||
+            !AsmUtil.isPrimitive(type) && (TypeUtils.isNullableType(kotlinType) || !InlineClassesUtilsKt.isInlineClassType(kotlinType))) {
             return StackValue.compareWithNull(argument, (KtTokens.EQEQ == opToken || KtTokens.EQEQEQ == opToken) ? IFNONNULL : IFNULL);
         } else {
-            // If exp has a non-nullable type, the comparison is vacuous.
-            // For inline classes we cannot necessarily compare with null at all.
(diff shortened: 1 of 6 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -3871,12 +3871,11 @@ private StackValue genCmpWithNull(KtExpression exp, IElementType opToken, @Nulla
         Type type = expressionType(exp);
         StackValue argument = pregeneratedExpr != null ? pregeneratedExpr : gen(exp);
 
-        if (kotlinType == null || TypeUtils.isNullableType(kotlinType)) {
+        if (kotlinType == null ||
+            !AsmUtil.isPrimitive(type) && (TypeUtils.isNullableType(kotlinType) || !InlineClassesUtilsKt.isInlineClassType(kotlinType))) {
```
udalov: I've checked the behavior on a similar test, but with `String` replaced to `Int`, and kotlinc 1.3.50 returns "OK" for that too. So I think the `!AsmUtil.isPrimitive(type)` condition is not needed here. And similarly in the IR backend

udalov:
@sfs Thanks! Since this is a regression in 1.3.60 which is about to be released, we'd like to have it in master as soon as possible. Therefore I've [merged](https://github.com/JetBrains/kotlin/commit/b80e157381853ad79b60c27aedf516ffbe8c5cd3) your changes with the suggestion above. However, I didn't succeed in fixing the IR behavior for primitive types right away, so I've added a muted test for now (not even sure that it needs fixing since it's questionable behavior, maybe will need discussion in the Language Committee): https://github.com/JetBrains/kotlin/commit/228fdb4436bbf7b0d772bfa4acdcdb7dcb1bdccb

Also I've logged this regression as [KT-34829](https://youtrack.jetbrains.com/issue/KT-34829). Thanks again!

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
sfs: That's possible, but note that `StackValue.compareWithNull` would produce incorrect bytecode for primitive types and the same issue exists on the IR backend.

Your test shows that there is also an issue with this check, though... I'll take another look!

PR #2768, opened 2019-11-06
Ensure the correct properties and functions are used in ForLoopsLowering

punzki:
Similar issue to #2560.


max-kammerer:
Thank you!

PR #2769, opened 2019-11-07
KT-34784 "Indent raw string" intention: do not suggest in const

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34784


mglukhikh:
Merged 👍

PR #2770, opened 2019-11-07
KT-34785 Enter handler: do not add 'trimIndent()' in const

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34785


goodwinnk:
Thank you for the cotribution! This is your commit in master: https://github.com/JetBrains/kotlin/commit/58fb1dede316e25acad8630adb923fa18b2d64d4.

PR #2771, opened 2019-11-07
klib expect serialization

alexander-gorshenev:


compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/metadata/KlibMetadataSerializer.kt
```diff
@@ -116,11 +122,32 @@ abstract class KlibMetadataSerializer(
...
+            this.filterNot { it.isExpectMember && !it.isSerializableExpectClass }
+        else
+            this.filterOutExpectsWithActuals()
+
+    private fun serializeClasses(packageName: FqName,
                                  //builder: ProtoBuf.PackageFragment.Builder,
                                  descriptors: Collection<DeclarationDescriptor>): List<Pair<ProtoBuf.Class, Int>> {
 
-        return descriptors.filterIsInstance<ClassDescriptor>().flatMap {
+        return descriptors.filterIsInstance<ClassDescriptor>()/*.filterOutExpectsWithActuals()*/.flatMap {
```
homuroll: What's up with that comment?

PR #2772, opened 2019-11-08
KT-21811 Convert string concatenation into multiline string

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21811


nj2k/nj2k-services/src/org/jetbrains/kotlin/nj2k/postProcessing/J2kPostProcessor.kt
```diff
@@ -220,7 +220,15 @@ private val inspectionLikePostProcessingGroup =
         MayBeConstantInspectionBasedProcessing(),
         RemoveForExpressionLoopParameterTypeProcessing(),
         intentionBasedProcessing(ReplaceMapGetOrDefaultIntention()),
-        inspectionBasedProcessing(ReplaceGuardClauseWithFunctionCallInspection())
+        inspectionBasedProcessing(ReplaceGuardClauseWithFunctionCallInspection()),
+        intentionBasedProcessing(ConvertToRawStringTemplateIntention()) { element ->
+            element.parents.none {
```
darthorimar: Why don't use `ConvertToStringTemplateIntention.shouldSuggestToConvert` here instead of custom check?

t-kameyama: `ConvertToStringTemplateIntention.shouldSuggestToConvert` returns false for:
```
var s = "asdf\n" +
        "nadfadsf\n" +
        "asdfasdf"
```

https://github.com/t-kameyama/kotlin/blob/fd53fe7fd1e5830ac7fcbe8127ce662c8c7d2b60/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToStringTemplateIntention.kt#L60
```entries.none { it is KtBlockStringTemplateEntry }``` returns false.

t-kameyama:
> In the `testFormattedString` please consider adding `trimIndent` call and saving formatting
We cannot add `trimIndent` to` const`.

darthorimar:
Thank you for your contribution ^_^

PR #2774, opened 2019-11-08
JVM_IR: remember facade fqnames of imported classes

pyos:
Necessary to support importing file classes annotated @JvmPackageName, since the actual package fragment they are a part of has the name from the `package` declaration.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendContext.kt
```diff
@@ -87,6 +87,10 @@ class JvmBackendContext(
 
     internal val localDelegatedProperties = mutableMapOf<IrClass, List<IrLocalDelegatedPropertySymbol>>()
 
+    // If the JVM fqname of a class differs from what is implied by its parent, e.g. if it's a file class
+    // annotated with @JvmPackageName, the correct name is recorded here.
+    internal lateinit var classNameOverride: MutableMap<IrClass, FqName>
```
udalov: I think it'd be a bit more convenient to store `JvmClassName` (`/`-separated) as values. Then the call site at `FileClassLowering` can use this:
```
context.classNameOverride[this] = JvmClassName.byInternalName(partClassType.internalName)
```
and the call site at `IrTypeMapper` can use this:
```
context.classNameOverride[irClass]?.let { return it.internalName }
```
eliminating the manual `.`-`/` conversions

pyos: Done.

udalov:
@pyos Thank you!

The missing piece in the `@JvmPackageName` support is the `packageName` field in `kotlin.Metadata`. I'll implement that shortly!

PR #2775, opened 2019-11-08
Ensure ForLoopsLowering only handles the CharSequence.iterator() stdlib extension function

punzki:
We only want to handle the known extension function for `CharSequence` in the standard library (top level `kotlin.text.iterator`). The behavior of this iterator is well-defined and can be lowered. `CharSequences` can have their own iterators, either as a member or extension function, and the behavior of those custom iterators is unknown.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
@@ -7,7 +7,6 @@ package org.jetbrains.kotlin.backend.common.lower.loops
 
 import org.jetbrains.kotlin.backend.common.CommonBackendContext
 import org.jetbrains.kotlin.backend.common.ir.Symbols
-import org.jetbrains.kotlin.backend.common.ir.isTopLevel
 import org.jetbrains.kotlin.backend.common.lower.DeclarationIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irIfThen
@@ -24,8 +23,6 @@ import org.jetbrains.kotlin.ir.symbols.IrSymbol
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.deepCopyWithSymbols
 import org.jetbrains.kotlin.ir.util.functions
-import org.jetbrains.kotlin.ir.util.getPackageFragment
-import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.util.OperatorNameConventions
 
 /**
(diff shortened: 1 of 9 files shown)


max-kammerer:
What about next test
'''
fun consume(cs: CharSequence): String {
    val result = StringBuilder()
    for (c in cs) {
        result.append(c)
    }
    return result.toString()
}

fun box(): String {
    val cs2 = CharSequenceWithMemberIterator("1234")

    val result2 = StringBuilder()
    for (c in cs2) {
        result2.append(c)
    }
    assertEquals("4321",consume(cs2) )

    return "OK"
}

open class CharSequenceWithExtensionIterator(val s: String) : CharSequence {
    fun get(foo: String): Char = TODO("shouldn't be called!")
    override val length = s.length
    override fun subSequence(startIndex: Int, endIndex: Int) = s.subSequence(startIndex, endIndex)
    override fun get(index: Int) = s.get(index)
}

// Returns no characters
operator fun CharSequenceWithExtensionIterator.iterator() = object : CharIterator() {
    public override fun nextChar() = TODO()
    public override fun hasNext() = false
}

class CharSequenceWithMemberIterator(s: String) : CharSequenceWithExtensionIterator(s) {
    // Returns characters in reverse
    operator fun iterator() = object : CharIterator() {
        private var index = 0
        public override fun nextChar() = get(length - ++index)
        public override fun hasNext() = index < length
    }
}
```
Looks like this optimization is totally unsafe

**Comment by max-kammerer** [comment]
Looks like this optimization changes behaviour on binary library update: 
e.g. CharSequenceWithMemberIterator is in a library class with custom member iterator.
if we compile smth against it and then remove custrom iterator declaration - it would be same behaviour.
But on opposite change it would be another one.
@udalov do we have any guarantees on binary libraries updates?

File: compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt [hunk 1]
```diff
@@ -265,30 +262,17 @@ internal class HeaderProcessor(
...
-        val iterable = (variable.initializer as? IrCall)?.let {
-            val extensionReceiver = it.extensionReceiver
+        val iteratorCall = variable.initializer as? IrCall
+        val iterable = iteratorCall?.run {
             if (extensionReceiver != null) {
-                val function = it.symbol.owner
-                if (it.valueArgumentsCount == 0
-                    && function.isTopLevel
-                    && function.getPackageFragment()?.fqName == FqName("kotlin.text")
-                    && function.name == OperatorNameConventions.ITERATOR
```
max-kammerer: Is this check is only actual for CharSequence?

punzki: Yes, it's only for CharSequence. This is what we're trying to match: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/iterator.html

punzki: I'm not sure I understood your question correctly, but the reason I moved this into `CharSequenceIterationHandler` is because the handlers are only try to match the iterable (the subject of the for-loop). In the previous code, if there is a `CharSequence object` with a member `iterator()`, the value of `iterable` in this block of code would be the dispatch receiver (the `CharSequence`). Later on it would match in `CharSequenceIterationHandler` and attempt to lower the loop, which is wrong because that iterator's behavior is unknown.

By moving the matching of the `iterator()` call to the handlers, we are able to match both `iterator()` call and the iterable at the same time.

punzki: I updated the test to cover more cases. I hope that helps clarify. Please let me know if I wasn't able to answer your question!

max-kammerer: Thank you for detailed explanation!

PR #2776, opened 2019-11-10
KT-34677 False positive "Collection count can be converted to size" with `Iterable`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34677


vladimirdolzhenko:
LGTM, Thank you for the contribution

PR #2777, opened 2019-11-11
KT-34593 Invert 'if' condition: Invert `String.isNotEmpty` should be `String.isEmpty`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34593


vladimirdolzhenko:
LGTM

PR #2778, opened 2019-11-11
JVM_IR: fold inline lambdas when computing OUTERCLASS

pyos:
so that the enclosing method of objects defined inside lambdas is the one they are declared in.

Note that this does not fix *all* enclosingInfo tests because JVM_IR currently follows the KT-28064 proposal, i.e. does not regenerate objects defined inside lambdas under any circumstances. For example, this causes test boxInline/enclosingInfo/inlineChain2.kt to fail because the enclosing method of objects is _2Kt.box instead of (non-existent in source code) `_2Kt$box$inlined$call$1.invoke` or whatever. What's more important is that OUTERCLASS no longer points to a non-existent `box$lambda-N` and therefore `.enclosingMethod` no longer throws.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -303,8 +304,8 @@ open class ClassCodegen protected constructor(
         }
     }
 
-    fun generateLocalClass(klass: IrClass, withinInline: Boolean): ReifiedTypeParametersUsages {
-        return ClassCodegen(klass, context, this, withinInline = withinInline || this.withinInline).generate()
+    fun generateLocalClass(klass: IrClass, parentFunction: IrFunction): ReifiedTypeParametersUsages {
+        return ClassCodegen(klass, context, this, parentFunction, withinInline = withinInline || parentFunction.isInline).generate()
```
max-kammerer: `withinInline` is unused after this change

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -58,11 +58,11 @@ class IrSourceCompilerForInline(
 
     override val inlineCallSiteInfo: InlineCallSiteInfo
         get() {
-            //TODO: support nested inline calls
+            val root = generateSequence(codegen) { it.inlinedInto }.last()
```
max-kammerer: If I understand  correctly this change should affect only `smap` related tests but no one of them is unmuted(

pyos: `InlineCallSiteInfo` is also used when adding OUTERCLASS to regenerated classes. For example, in `inlineChain.kt` we have

```
val res2 = call {
    call {
        test { "OK" }
    }
}
```

Here `test` is an inline function that contains an anonymous object that calls an inline lambda, so that object is regenerated. Since `test` is inlined into a lambda defined in `_2Kt.box`, we want it to have `OUTERCLASS _2Kt box()`. `codegen` points to `ExpressionCodegen` for the inner lambda, then its `inlinedInto` is the outer lambda, and finally that one's `inlinedInto` is `box`, which has `inlinedInto == null`.

As for SMAP tests, I think quite a lot of the muted ones depend on objects in lambdas being regenerated...

max-kammerer: Thank you for clarifications!

max-kammerer:
Many thanks!

PR #2779, opened 2019-11-11
IR type fixes

sfs:
This is yet another step towards enabling type checking in the jvm ir backend (this came up in the discussion for #2760).

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
@@ -235,7 +235,7 @@ class JvmSymbols(
     }
 
     val suspendLambdaClass: IrClassSymbol = createClass(FqName("kotlin.coroutines.jvm.internal.SuspendLambda")) { klass ->
-        klass.superTypes += suspendFunctionInterface.typeWith()
+        klass.superTypes += suspendFunctionInterface.owner.defaultType
         klass.addConstructor().apply {
             addValueParameter("arity", irBuiltIns.intType)
             addValueParameter("completion", continuationClass.typeWith(irBuiltIns.anyNType).makeNullable())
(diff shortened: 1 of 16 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -235,7 +235,7 @@ class JvmSymbols(
     }
 
     val suspendLambdaClass: IrClassSymbol = createClass(FqName("kotlin.coroutines.jvm.internal.SuspendLambda")) { klass ->
-        klass.superTypes += suspendFunctionInterface.typeWith()
+        klass.superTypes += suspendFunctionInterface.owner.defaultType
```
udalov: Maybe we could introduce an extension `val IrClassSymbol.defaultType: IrType` beside `starProjectedType`, to remove the `.owner.defaultType` pattern?

sfs: Good idea, I've added it as an extension on `IrClassifierSymbol`, since the pattern appears in several other places.

udalov:
Thanks!

max-kammerer:
Hi Steven,
actually I'm observing next bad bytecode in default functions:

```
   ...mask processsing...
    ....
    ALOAD 0
    LDC "inlineLambda"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull (Ljava/lang/Object;Ljava/lang/String;)V
    ALOAD 1
    LDC "noInlineLambda"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull (Ljava/lang/Object;Ljava/lang/String;)V
   L5
    LINENUMBER 8 L5
    ALOAD 1
    INVOKEINTERFACE kotlin/jvm/functions/Function0.invoke ()Ljava/lang/Object; (itf)
```
This also causes problem with default lambda inlining so i'm going to revert this change

PR #2780, opened 2019-11-11
[JVM IR] Fix @JvmOverloads + Parameterless Main

kandersen:
Resolves the interaction of @JvmOverloads annotations and
parameterless main methods.

In the following code, both mechanisms generate methods that
ultimately produce the signature `public static void main(String[] args)`
of which there can be only one (true in general of any signature).

```
fun main() { }

@JvmOverloads
fun main(Array<String> args, x: Int = 42) { }
```

This PR simply shuffles the lowerings around, letting parameterless
main methods detect the presence of the default overload produced by
the annotation.

Additionally, this PR improves the testing of parameterless main
methods by actual bytecode patterns, and not simple check for
successful compilation (as @sfs and I discovered, there are issues in
flagging an error on duplicate signatures on the IR backend).

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
@@ -184,15 +184,8 @@ private val syntheticAccessorPhase = makeIrFilePhase(
     prerequisite = setOf(objectClassPhase, staticDefaultFunctionPhase, interfacePhase)
 )
 
-private val mainMethodGenerationPhase = makeIrFilePhase(
-    ::MainMethodGenerationLowering,
-    name = "MainMethodGeneration",
-    description = "Identify parameterless main methods and generate bridge main-methods"
-)
-
 @Suppress("Reformat")
 private val jvmFilePhases =
-        mainMethodGenerationPhase then
         typeAliasAnnotationMethodsPhase then
         stripTypeAliasDeclarationsPhase then
         provisionalFunctionExpressionPhase then
@@ -271,6 +264,8 @@ private val jvmFilePhases =
         replaceKFunctionInvokeWithFunctionInvokePhase then
 
         checkLocalNamesWithOldBackendPhase then
+
+        mainMethodGenerationPhase then
 
         // should be last transformation
         removeDeclarationsThatWouldBeInlined then
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
@@ -8,6 +8,7 @@ package org.jetbrains.kotlin.backend.jvm.lower
 import org.jetbrains.kotlin.backend.common.ClassLoweringPass
 import org.jetbrains.kotlin.backend.common.ir.allParameters
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.backend.jvm.ir.getJvmNameFromAnnotation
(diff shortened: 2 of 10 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -23,14 +24,18 @@ import org.jetbrains.kotlin.ir.builders.irReturn
...
-import org.jetbrains.kotlin.resolve.DescriptorUtils
 import org.jetbrains.kotlin.types.Variance
 
+internal val mainMethodGenerationPhase = makeIrFilePhase(
+    ::MainMethodGenerationLowering,
+    name = "MainMethodGeneration",
+    description = "Identify parameterless main methods and generate bridge main-methods",
+    prerequisite = setOf(jvmOverloadsAnnotationPhase)
+)
+
```
kandersen: Introduced the prerequisite, and I felt that should be physically closer to the implementation of the pass -- it's a central assumption for correct behavior, after all.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -184,15 +184,11 @@ private val syntheticAccessorPhase = makeIrFilePhase(
...
 
-private val mainMethodGenerationPhase = makeIrFilePhase(
-    ::MainMethodGenerationLowering,
-    name = "MainMethodGeneration",
-    description = "Identify parameterless main methods and generate bridge main-methods"
-)
-
 @Suppress("Reformat")
 private val jvmFilePhases =
+        jvmOverloadsAnnotationPhase then
```
udalov: Can we move `mainMethodGenerationPhase` at the end instead? We would probably like to have a common prefix of lowerings across all backends some day in the future, so it seems useful to minimize the number of platform-specific lowerings in the beginning

kandersen: Indeed we can! O.o

I can't recall what prompted me to place it at the front, but I do recall discussing that it would be simpler to do during code-gen when things like `@JvmName` have been dealt with. Essentially parameterless-main is just a patch on the bytecode: do we have a signature `public static void main ()V` but no `public static void main ([LString)V`, add the latter, without regards for annotations, projects, type parameters etc.

udalov: Ideally we'd have lowerings that remove all these specifics and then operate on the IR that is as close to the bytecode as possible. Modifying the IR is far easier than modifying the bytecode anyway. But I don't know exactly why we don't for example handle `@JvmName` annotations in a lowering. I think this could get rid of JvmName checks in the `mainMethodGenerationPhase`, as well as in `MethodSignatureMapper`. But probably this could introduce a whole set of new problems, so I don't encourage this approach either.

kandersen: `JvmNamePhase` sounds like a small experiment for a rainy day :) I'll add it to my list regardless. Do you at the moment have a sense of where the "common prefix" ends in `JvmLower.kt`?

kandersen:
@udalov I shuffled the phases back around. Let me know if you find anything else you would like me to investigate.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
udalov: You can get a rough idea at where the common prefix is supposed to end in `JsLoweringPhases.kt` (see the "Common prefix ends" comment), but that's also not set in stone. Note that we're not looking into extracting and maintaining the common prefix in JVM IR right now, it's more of a goal for the future.

LaoTsing:
sorry! but not working in Java 13! in varargInAnnotationParameter.kt hard coded old versions - may be to remove this hard code?

udalov:
@LaoTsing Please use [discussion forums](https://discuss.kotlinlang.org/) and [YouTrack](https://youtrack.jetbrains.com/issues/KT) for raising issues like these

udalov:
@kandersen Looks good, thanks!

BTW feel free to look into supporting `suspend fun main(args: Array<String>)` and `suspend fun main()` in the IR backend as well, if you're interested :). These seem like the most relevant missing features related to your work on `main` method generation.

kandersen:
Thanks for the tip, will look into it.

PR #2782, opened 2019-11-11
KT-34450 `Convert function to property` intention should be also displayed on `fun` keyword

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34450


vladimirdolzhenko:
LGTM

PR #2562, opened 2019-09-04
KT-3262 Inspection "Inner class could be nested"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-3262

PR #2783, opened 2019-11-11
KT-34819 Inspection: report unnecessary "?: return null"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34819


vladimirdolzhenko:
LGTM. 

Same as #2562: Actual code base requires some extra changes (in plugin-common.xml bunch files). I've made them and successfully merged.

Thank you.

PR #2784, opened 2019-11-11
JVM_IR: resolve inline fake overrides before codegen

pyos:
See KT-33054.

### DIFF
--- a/compiler/testData/codegen/box/jvm8/kt29242.kt
+++ b/compiler/testData/codegen/box/jvm8/kt29242.kt
@@ -1,0 +1,24 @@
+// TARGET_BACKEND: JVM
+// IGNORE_BACKEND: JVM
+// JVM_TARGET: 1.8
+// See also kt33054.kt
+
+fun causesVerifyErrorSample(): Sample<Boolean> = Sample
+    .Success(true)
+    .flatMap { Sample.Failure(RuntimeException()) }
+
+sealed class Sample<out T> {
+    inline fun <R> flatMap(f: (T) -> Sample<R>): Sample<R> =
+        when (this) {
+            is Failure -> this
+            is Success -> f(this.value)
+        }
+
+    data class Failure(val exception: Throwable): Sample<Nothing>()
+    data class Success<out T>(val value: T): Sample<T>()
+}
+
+fun box(): String {
+    causesVerifyErrorSample()
+    return "OK"
+}
--- a/compiler/testData/codegen/box/jvm8/kt33054.kt
+++ b/compiler/testData/codegen/box/jvm8/kt33054.kt
@@ -1,0 +1,19 @@
+// TARGET_BACKEND: JVM
+// IGNORE_BACKEND: JVM
+// JVM_TARGET: 1.8
+
+open class A(val x: String) {
+    inline fun f() = if (this is C) this else A("O")
+
+    val y
+        inline get() = if (this is C) this else A("K")
+}
+
+class B : A("unused")
+class C : A("unused")
+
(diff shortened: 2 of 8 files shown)


compiler/testData/codegen/box/jvm8/kt33054.kt
```diff
@@ -0,0 +1,16 @@
+// TARGET_BACKEND: JVM
+// IGNORE_BACKEND: JVM
+// JVM_TARGET: 1.8
+
+open class A(val x: String) {
+    inline fun f() = if (this is C) this else A(x)
```
max-kammerer: Could you also add test with inline getter?

pyos: Something like this? https://github.com/JetBrains/kotlin/compare/08afa065ed01c48604f3620e3a2fa76437aab4fe..62fa1f00f6ecabd181b1f78ff1fdcf18fa005212

max-kammerer: Yes!

max-kammerer:
Looks like this commit also affects https://youtrack.jetbrains.com/issue/KT-29242, could you also add test from this issue?

pyos:
Isn't that exactly the same code, i.e. `this` in `flatMap` has type `Success` after inlining while the other branch is `Sample`, so from asm's viewpoint the join of those is `Object`?

max-kammerer:
Yes, looks like it's one problem. Actually I want to have test similar to original one reported in issue KT-29242 or KT-33054

max-kammerer:
..in additional to existing one

pyos:
Added copy of the code from KT-29242 as a test.

max-kammerer:
Thank you!

compiler/testData/codegen/box/jvm8/kt29242.kt
```diff
@@ -0,0 +1,24 @@
+// TARGET_BACKEND: JVM
+// IGNORE_BACKEND: JVM
+// JVM_TARGET: 1.8
```
max-kammerer: `JVM_TARGET ` is redundant. We have special test suite that set target to 1.8 in codegenTarget8Jvm8Test. And run configuration for it in https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinPublic_CodegenTestsOnDifferentJDKs?branch=

`TARGET_BACKEND` could be also removed

max-kammerer: So it's best to convert these tests to boxInline

pyos: Are you sure this is true for local runs? Without JVM_TARGET, these tests pass on the non-IR backend. (I'm running them as `./gradlew cleanTest :compiler:test --parallel`). The rest of the files in this directory seem to have both JVM_TARGET and TARGET_BACKEND as well.

pyos: Moved to boxInline,

max-kammerer: All of these tests are require //JVM_TARGET: 1.8 cause they wouldn't work on default target 1.6. And as result it's implies JVM target restriction.

And yes, you are right. This test would be passed on default target and failed on custom one. 
OK, let's keep tests as is. And after fixing problem in current backend I will convert this tests into boxInline.

pyos: Moved back to box/jvm8.

max-kammerer:
Thank you!

PR #2785, opened 2019-11-12
KT-33258 "Merge 'if's" intention drops comment before nested if

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33258
https://youtrack.jetbrains.com/issue/KT-39552


igoriakovlev:
Thank you for your effort!

PR #2787, opened 2019-11-12
JVM IR: Fix inlining of inline class properties in external modules

sfs:
Since the inliner uses descriptors to identify external declarations we need to preserve PropertyGetter/SetterDescriptors in the IR backend.

This hit the inline class lowering, which replaces getters/setters with new static functions inside of inline classes. I've not found any other places where this is broken so far, but there is a comment in the IR `MethodSignatureMapper` that we sometimes lose `correspondingPropertySymbols` in the IR backend. Should this happen for external inline properties we'll run into a `NullPointerException` in the inliner (which has nothing to do with this PR, though).

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
@@ -115,12 +115,13 @@ inline fun IrProperty.addSetter(builder: IrFunctionBuilder.() -> Unit = {}): IrS
     }
 
 fun IrFunctionBuilder.buildFun(originalDescriptor: FunctionDescriptor? = null): IrFunctionImpl {
-    val wrappedDescriptor = if (originalDescriptor is DescriptorWithContainerSource)
-        WrappedFunctionDescriptorWithContainerSource(originalDescriptor.containerSource)
-    else if (originalDescriptor != null)
-        WrappedSimpleFunctionDescriptor(originalDescriptor)
-    else
-        WrappedSimpleFunctionDescriptor()
+    val wrappedDescriptor = when(originalDescriptor) {
+        is DescriptorWithContainerSource -> WrappedFunctionDescriptorWithContainerSource(originalDescriptor.containerSource)
+        is PropertyGetterDescriptor -> WrappedPropertyGetterDescriptor(originalDescriptor.annotations, originalDescriptor.source)
+        is PropertySetterDescriptor -> WrappedPropertySetterDescriptor(originalDescriptor.annotations, originalDescriptor.source)
+        null -> WrappedSimpleFunctionDescriptor()
+        else -> WrappedSimpleFunctionDescriptor(originalDescriptor)
+    }
     return IrFunctionImpl(
         startOffset, endOffset, origin,
         IrSimpleFunctionSymbolImpl(wrappedDescriptor),
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
@@ -74,7 +74,7 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
         if (nameForSpecialFunction != null) return nameForSpecialFunction
 
         val property = (function as? IrSimpleFunction)?.correspondingPropertySymbol?.owner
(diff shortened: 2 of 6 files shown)


sfs:
Fixed a stupid copy-paste error. The first line in the `inlineClassInlineProperty` test was supposed to be `// !LANGUAGE: +InlineClasses` and not `TARGET_BACKEND: JVM`...

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
```diff
@@ -115,12 +115,13 @@ inline fun IrProperty.addSetter(builder: IrFunctionBuilder.() -> Unit = {}): IrS
     }
 
 fun IrFunctionBuilder.buildFun(originalDescriptor: FunctionDescriptor? = null): IrFunctionImpl {
-    val wrappedDescriptor = if (originalDescriptor is DescriptorWithContainerSource)
-        WrappedFunctionDescriptorWithContainerSource(originalDescriptor.containerSource)
-    else if (originalDescriptor != null)
-        WrappedSimpleFunctionDescriptor(originalDescriptor)
-    else
-        WrappedSimpleFunctionDescriptor()
+    val wrappedDescriptor = when(originalDescriptor) {
```
udalov: More of a side note, but I wonder if we can make this more general and implement `DescriptorsToIrRemapper.remapDeclaredSimpleFunction` this way and use that method everywhere `WrappedFunctionDescriptorWithContainerSource` is currently instantiated? Because I suspect other call sites might also need special descriptor instances for property accessors.

Also, maybe having `WrappedProperty{G,S}etterDescriptorWithContainerSource` would be useful.

But since such change might break something else, it's not directly related to this one.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
```diff
@@ -74,7 +74,7 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
         if (nameForSpecialFunction != null) return nameForSpecialFunction
 
         val property = (function as? IrSimpleFunction)?.correspondingPropertySymbol?.owner
-        if (property != null) {
+        if (property != null && function.name.isSpecial) {
```
udalov: Could you clarify why this is needed? It would seem that property accessor's name is always `isSpecial`

sfs: For properties of inline class type (or in an inline class) the name is mangled in the inline class lowering and no longer special at this point. This wasn't an issue before, since we treated such accessors as ordinary functions.

I'm not actually sure why we have special names for properties in the first place. Would it make sense to put in the JvmAbi names after psi2ir?

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
sfs: I wasn't aware of this code, but yes, we should probably produce `WrappedFunctionDescriptorWithContainerSource` here as well. This doesn't seem to be needed for property accessors, though, since these are already a special case in the inliner. I'm honestly not sure *why* the inliner has several different ways of identifying external declarations, so I don't know if we wouldn't end up breaking something else by introducing a new `...DescriptorWithContainerSource` variant.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
udalov: Thanks. That would make sense, yes. But since we don't lower lazy stub IR for external dependencies right now, this would help only for properties from the same module.

sfs: In that case could we do it as part of a generator extension in psi2ir?

udalov: Yes, an extension which is then passed to `ExternalDependenciesGenerator` much like `context.extensions.externalDeclarationOrigin` or `facadeClassGenerator` seems like a good idea!

udalov:
@sfs I suppose the improvements discussed above can be safely postponed, so I'm merging your change now as is. Thank you!

PR #2788, opened 2019-11-13
Handle `withIndex()` calls in ForLoopsLowering

punzki:
Handles all for-loops over `withIndex()` calls in the stdlib (arrays, `CharSequence`, `Iterable`, `Sequence`) except for `Iterator<T>.withIndex()` (which returns an `Iterator`). The for-loop must use a destructuring declaration to be lowered.

The loop transformer had to be restructured to be able to determine if there are destructuring declarations in the for-loop. There was also some refactoring in the lowered loop builder code.

TODO: Check for overflow for `Iterable` and `Sequence` by adding a call to `checkIndexOverflow()`. This is not working for current JVM backend: https://youtrack.jetbrains.com/issue/KT-34665


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -37,39 +39,123 @@ internal data class LoopReplacement(
...
+                    irType = it
+                )
+            }
+        }
+    }
+
+    // This cannot be declared and initialized directly in the constructor. At the time of the base class (ForLoopHeader) constructor
+    // execution, the `headerInfo` property overridden in the derived class (e.g., NumericForLoopHeader) is not yet initialized.
+    // Therefore, we must use the `headerInfo` constructor parameter in the "init" block above instead of using the property.
+    protected open val headerInfo: NumericHeaderInfo = headerInfo
```
max-kammerer: Would class generalization solve this problem (e.g. add generic parameters)?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
```diff
@@ -117,24 +117,53 @@ private class RangeLoopTransformer(
...
+        //
+        // We primarily need to determine HOW to optimize the for loop from the iterable expression in the header (e.g., if it's a
+        // `withIndex()` call, a progression such as `10 downTo 1`). However in some cases (e.g., for `withIndex()`), we also need to
+        // examine the while loop to determine if we CAN optimize the loop.
+        if (expression.origin != IrStatementOrigin.FOR_LOOP) {
+            return super.visitBlock(expression)  // Not a for-loop block.
+        }
+
+        with(expression.statements) {
+            assert(size == 2)
```
max-kammerer: Please add assertion message

max-kammerer: ...and maybe element `dump `

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
```diff
@@ -144,96 +173,140 @@ private class RangeLoopTransformer(
...
+            loopBodyStatements
+        )
 
-    private fun getLoopHeader(expression: IrExpression): ForLoopHeader? {
-        if (expression !is IrCall
-            || (expression.origin != IrStatementOrigin.FOR_LOOP_HAS_NEXT
-                    && expression.origin != IrStatementOrigin.FOR_LOOP_NEXT)
-        ) {
+        if (loopHeader.consumesLoopVariableComponents && mainLoopVariable.origin != IrDeclarationOrigin.IR_TEMPORARY_VARIABLE) {
+            // We determine if there is a destructuring declaration by checking if the main loop variable is temporary.
```
max-kammerer: Looks like we need more origins here to make such cases more explicit //cc @dnpetrov

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
punzki: @max-kammerer Done. Also added for other asserts.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
punzki: @max-kammerer That works! Fixed.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
punzki: We could put it in a composite with `origin=DESTRUCTURING_DECLARATION`, that's how it works for regular destructuring statements (see `compiler/testData/ir/irText/expressions/destructuring1.txt`)

max-kammerer:
@punzki Could you rebase on master and generate new Fir tests?

punzki:
> @punzki Could you rebase on master and generate new Fir tests?

@max-kammerer Done!

max-kammerer:
@punzki Many thanks!

PR #2789, opened 2019-11-13
IR: refactor the DefaultArgumentStubGenerator lowering

pyos:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -109,15 +105,11 @@ open class DefaultArgumentStubGenerator(
...
-                    endOffset = irFunction.endOffset,
-                    type = context.irBuiltIns.unitType,
-                    symbol = irFunction.symbol,
-                    typeArgumentsCount = newIrFunction.parentAsClass.typeParameters.size + newIrFunction.typeParameters.size
-                ).apply {
+                is IrConstructor -> +irDelegatingConstructorCall(irFunction).apply {
                     passTypeArgumentsFrom(newIrFunction.parentAsClass)
-                    passTypeArgumentsFrom(newIrFunction)
+                    // This is for Kotlin/Native, which differs from the other backends in that constructors
+                    // apparently do have type parameters and dispatch receivers:
```
gogabr: Constructors can have their own type parameters in Java, but this is irrelevant in this case, since there are no default parameters in Java.
Have you seen constructors with type parameters in Native code? I never heard about those.

pyos: Not sure about type parameters, actually, I've left this line in just in case. (There are definitely constructors with receivers, though.)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -240,106 +228,73 @@ open class DefaultParameterInjector(
...
+
+        val stubOverride = declaration.generateDefaultsFunction(context, skipInline, skipExternalMethods) ?: return null
+        // We *have* to resolve the fake override here since on the JVM, a default stub for a function implemented
+        // in an interface does not leave an abstract method after being moved to DefaultImpls (see InterfaceLowering).
+        // Calling the fake override on an implementation of that interface would then result in a call to a method
+        // that does not actually exist as DefaultImpls is not part of the inheritance hierarchy.
+        val stubFunction = if (stubOverride is IrSimpleFunction) stubOverride.resolveFakeOverride()!! else stubOverride
+        log { "$declaration -> $stubFunction" }
+
+        val realArguments = declaration.valueParameters.size
```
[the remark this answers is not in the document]
pyos: done

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
gogabr: Then I suggest leaving this line, as it makes no harm anyway, but removing the comment, because it seems misleading.

pyos: Edited the comment to mention that there are probably no type parameters.

gogabr:
Will merge as soon as the big test run completes. Thank you!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -415,181 +308,78 @@ class DefaultParameterCleaner constructor(val context: CommonBackendContext) : F
...
+        generateDefaultsFunctionImpl(context, IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER).also {
+            context.ir.defaultParameterDeclarationsCache[this] = it
+        }
+    this is IrSimpleFunction -> {
+        // If this is an override of a function with default arguments, produce a fake override of a default stub.
+        val overriddenStubs = overriddenSymbols.mapNotNull {
+            it.owner.generateDefaultsFunction(context, skipInlineMethods, skipExternalMethods)?.symbol as IrSimpleFunctionSymbol?
         }
+        if (overriddenStubs.isNotEmpty())
+            generateDefaultsFunctionImpl(context, IrDeclarationOrigin.FAKE_OVERRIDE).also {
```
pyos: I just realized that both the original version and the refactored one would generate fake overrides with all arguments non-nullable, as only the implementation had `defaultValue`s on the original function's arguments. This does not actually affect anything because `parametersForCall` resolves fake overrides, but it may be something to keep in mind.

gogabr: Thanks, indeed, it makes sense to remember this.

PR #2790, opened 2019-11-14
KT-33384 Intention to switch between single-line/multi-line lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33384


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaLineIntention.kt
```diff
@@ -0,0 +1,43 @@
...
+sealed class ConvertLambdaLineIntention(private val toMultiLine: Boolean) : SelfTargetingIntention<KtLambdaExpression>(
+    KtLambdaExpression::class.java, "Convert to ${if (toMultiLine) "multi" else "single"}-line"
+) {
+    override fun isApplicableTo(element: KtLambdaExpression, caretOffset: Int): Boolean {
+        val functionLiteral = element.functionLiteral
+        val body = functionLiteral.bodyBlockExpression ?: return false
+        val startLine = functionLiteral.getLineNumber(start = true)
+        val endLine = functionLiteral.getLineNumber(start = false)
+        return toMultiLine && startLine == endLine
+                || !toMultiLine && startLine != endLine && body.statements.size <= 1 && body.allChildren.none { it is PsiComment }
```
darthorimar: Why skipping all lambdas with comments? Just skipping single-line comments is enough

idea/testData/intentions/convertLambdaToMultiLine/simple4.kt.after
```diff
@@ -0,0 +1,6 @@
+// WITH_RUNTIME
+fun test(list: List<String>) {
+    list.forEach { item ->
+        /* aaa */ println(item); println(item) /* bbb */
```
darthorimar: Let's move every statement inside lambda to its own line in such cases as we already have multi-line lambda

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaLineIntention.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.psi.PsiComment
+import org.jetbrains.kotlin.idea.refactoring.getLineNumber
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.psiUtil.allChildren
+
+sealed class ConvertLambdaLineIntention(private val toMultiLine: Boolean) : SelfTargetingIntention<KtLambdaExpression>(
+    KtLambdaExpression::class.java, "Convert to ${if (toMultiLine) "multi" else "single"}-line"
```
darthorimar: Better add "lambda" (e.g,  "Convert to multi-line lambda") to the end of the title as it's not quite obvious for the user what should that intension do in case when many intentions are suggested to the user

darthorimar:
Thank you!

PR #2791, opened 2019-11-14
JVM_IR: keep the $default suffix for stubs for @JvmName functions

pyos:
Technically a backwards compatibility problem, as the new backend *consistently* renamed `f$default` on `f` with `@JvmName("g")` to `g` instead of `g$default`, so it all worked out. However, this breaks when encountering libraries compiled with the non-IR backend.


udalov:
@pyos Thank you!

Lowering `@JvmName` indeed could be useful. This was also brought up in https://github.com/JetBrains/kotlin/pull/2787#discussion_r346743142.

PR #2795, opened 2019-11-15
PSI2IR: generate IrEnumConstructorCalls with correct return type

pyos:
Normally, the fact that is was Unit was not visible as enum constructors are lowered to normal class constructors anyway. The exception is when the arguments are reordered, causing the incorrect return type to leak into the block that holds temporary variables.


dnpetrov:
Sorry for a delay. I'll run tests and push manually.

dnpetrov:
Pushed manually.

PR #2796, opened 2019-11-15
IR: don't attempt to move defaults to `actual` in another module

pyos:
May happen when a function in an `expect` class is aliased through an `actual typealias`; the matching declaration is filtered out in `ExpectedActualResolver.findActualForExpected` as it has no source.

### DIFF
--- a/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/ExpectDeclarationRemover.kt
+++ b/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/ExpectDeclarationRemover.kt
@@ -63,7 +63,12 @@ class ExpectDeclarationRemover(val symbolTable: ReferenceSymbolTable, private va
                     return
                 }
 
-                val actualParameter = function.findActualForExpected().valueParameters[index]
+                // If the containing declaration is an `expect class` that matches an `actual typealias`,
+                // the `actual fun` or `actual constructor` for this may be in a different module.
+                // Nothing we can do with those.
+                // TODO they may not actually have the defaults though -- may be a frontend bug.
+                val actualFunction = function.findActualForExpected() ?: return
+                val actualParameter = actualFunction.valueParameters[index]
 
                 // Keep actual default value if present. They are generally not allowed but can be suppressed with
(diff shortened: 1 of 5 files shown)


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/ExpectDeclarationRemover.kt
```diff
@@ -63,7 +63,12 @@ class ExpectDeclarationRemover(val symbolTable: ReferenceSymbolTable, private va
                     return
                 }
 
-                val actualParameter = function.findActualForExpected().valueParameters[index]
+                // If the containing declaration is an `expect class` that matches an `actual typealias`,
+                // the `actual fun` or `actual constructor` for this may be in a different module.
+                // Nothing we can do with those; besides, all default values should already exist, or
+                // the frontend would've complained.
```
pyos: Correction: turns out, the frontend does *not* complain. Instead, incorrect code is silently generated by the non-IR backend (attempts to call a nonexistent default stub in the imported class) while the IR backend fails with an assertion error in ExpressionCodegen (null value parameter).

skuzmich:
Please clarify behavior we expect when having default values both in `expect` and corresponding `actual` declarations. I assumed it to be a compiler error.

pyos:
`expect` and `actual` must be in the same module, and only the `expect` is allowed to have default parameters. (There's a comment right below the lines that I changed that mentions this.)

The problematic part is that an `expect class` apparently can be matched to an `actual typealias` and the alias can point to a different module, in which case *both* must have default parameters, or the typealias target must have an equivalent structure (e.g. a `constructor(x: String = ...)` can be matched to `constructor(x: String)` + `constructor()`). This is in fact the reason for this PR: the original code in [okio](https://github.com/square/okio) matches `expect class IOException(message: String? = null)` to `actual typealias IOException = java.io.IOException`.

IMO it'd be best to forbid `actual typealias` if any part of the `expect class` has default arguments since it's completely impossible to prove equivalency - who's to say `IOException(null) == IOException()`? That said, here I'm merely replicating existing behavior to keep a project working.

skuzmich:
Thank you! LGTM.

udalov:
@pyos Merged, thank you!

PR #2797, opened 2019-11-16
KT-34894 Action "Add not-null asserted (!!) call" doesn't fix error for properties with omitted `this`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34894


mglukhikh:
Merged 👍

PR #2798, opened 2019-11-17
KT-20357: Add samples for functions related to associate

Dattish:

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/arrays.kt
+++ b/libraries/stdlib/samples/test/samples/collections/arrays.kt
@@ -211,20 +211,27 @@ class Arrays {
 
         @Sample
         fun associateArrayOfPrimitivesTo() {
-            fun fib(n: Int, memo: MutableMap<Int, Int>): Int {
-                val fib = memo[n] ?: fib(n - 1, memo) + fib(n - 2, memo)
-                memo[n] = fib
-                return fib
+            fun fib(n: Int): Int {
+                var current = 0
+                var next = 1
+                var sum: Int
+                if (n == 0)
+                    return current
+                for (i in 2..n) {
+                    sum = current + next
+                    current = next
+                    next = sum
+                }
+                return next
             }
 
-            val numbers = intArrayOf(22, 23, 24, 25, 26)
+            val numbers = intArrayOf(1, 2, 3, 4, 5)
 
-            val fibonacciNumbers = mutableMapOf(0 to 0, 1 to 1)
+            val fibonacciNumbers = mutableMapOf<Int, Int>()
 
-            numbers.associateTo(fibonacciNumbers) { it to fib(it, fibonacciNumbers) }
+            numbers.associateTo(fibonacciNumbers) { it to fib(it) }
 
-            assertPrints(fibonacciNumbers[25], "75025")
-            assertPrints(fibonacciNumbers[26], "121393")
+            assertPrints(fibonacciNumbers, "{1=1, 2=1, 3=2, 4=3, 5=5}")
         }
 
         @Sample
--- a/libraries/stdlib/samples/test/samples/collections/arrays.kt
+++ b/libraries/stdlib/samples/test/samples/collections/arrays.kt
@@ -68,34 +68,22 @@ class Arrays {
         fun associate() {
             data class Person(val firstName: String, val lastName: String)
 
-            val computerScientists = arrayOf(Person("Grace", "Hopper"), Person("Edsger", "Dijkstra"), Person("Alan", "Turing"))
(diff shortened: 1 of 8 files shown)


libraries/stdlib/samples/test/samples/collections/arrays.kt
```diff
@@ -64,6 +64,176 @@ class Arrays {
...
+                var sum: Int
+                if (n == 0)
+                    return current
+                for (i in 2..n) {
+                    sum = current + next
+                    current = next
+                    next = sum
+                }
+                return next
+            }
```
TillmannBerg: This implementation of the Fibonacci numbers has three vars and three branches. How about the following?
```
fun fib(n: Int): Int =
    when (n) {
        0 -> 0
        1 -> 1
        else -> fib(n - 1) + fib(n - 2)
    }
```
I know it is less efficient, but its much clearer.

libraries/stdlib/samples/test/samples/collections/arrays.kt
```diff
@@ -64,6 +64,176 @@ class Arrays {
...
+            data class Person(val firstName: String, val lastName: String) {
+                override fun toString(): String = "$firstName $lastName"
+            }
+
+            val computerScientists = arrayOf(Person("Grace", "Hopper"), Person("Edsger", "Dijkstra"), Person("Alan", "Turing"))
+
+            val byLastName = computerScientists.associateBy { it.lastName }
+
+            assertPrints(byLastName.keys, "[Hopper, Dijkstra, Turing]")
+            assertPrints(byLastName.values, "[Grace Hopper, Edsger Dijkstra, Alan Turing]")
```
TillmannBerg: Printing keys and values separately made it harder (for me) to see that a map falls out of the `associateBy`. Also, I would change the selection of scientists a bit to include the edge case of repeated keys in the example:
```
val scientists = arrayOf(Person("Grace", "Hopper"), Person("Daniel", "Bernoulli"), Person("Jacob", "Bernoulli"))
val byLastName = scientists.associateBy { it.lastName }
assertPrints(byLastName, "{Hopper=Grace Hopper, Bernoulli=Jacob Bernoulli}")
```

Dattish: Sounds good, will add.

libraries/stdlib/samples/test/samples/collections/arrays.kt
Dattish: I went with a simpler example on these instead.

libraries/stdlib/samples/test/samples/collections/arrays.kt
```diff
@@ -64,6 +64,159 @@ class Arrays {
 
     class Transformations {
 
+        @Sample
+        fun associate() {
+            data class Person(val firstName: String, val lastName: String)
+
+            val scientists = arrayOf(Person("Grace", "Hopper"), Person("Jacob", "Bernoulli"), Person("Johann", "Bernoulli"))
```
ilya-g: Provided that [`associate` doc page](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate.html) shows overloads for both collections and all arrays, I suggest leaving only one sample for all arrays: the one for primitive arrays `associateArrayOfPrimitives`.

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -337,6 +337,94 @@ class Collections {
 
     class Transformations {
 
+        @Sample
+        fun associate() {
+            data class Person(val firstName: String, val lastName: String)
+
+            val scientists = listOf(Person("Grace", "Hopper"), Person("Jacob", "Bernoulli"), Person("Johann", "Bernoulli"))
+
+            val byLastName = scientists.associate { it.lastName to it.firstName }
```
ilya-g: While this reads good, it can be replaced with more efficient, though little less pretty  `associate({ it.lastName }, { it.firstName })`. The real strength of _this_ overload is that it can map the item somehow and then use the properties of that result as key and value.
For example, suppose we have a list of names as `listOf("Grace Hopper", "Jacob Bernoulli", ...)`. Then we can build the same map as in this sample as following:
```
names.associate { it.split(" ").let { (firstName, lastName) -> lastName to firstName } }
```
Without this overload we would have to split the name string twice or use an intermediate `map` operation.

libraries/stdlib/samples/test/samples/collections/arrays.kt
Dattish: Sounds good, I'll replace the array samples for objects.

libraries/stdlib/samples/test/samples/collections/collections.kt
Dattish: Alright, I'll use that String example.

ilya-g:
Merged. Thanks for such comprehensive sample coverage.

PR #2799, opened 2019-11-18
KT-10790 "Move statement up" for @file-targeted annotation moves package declaration to bottom of file

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-10790


vladimirdolzhenko:
Sorry for a long delay. 

I can't reproduce initial problem. Is this PR is still applicable ?

t-kameyama:
I think this PR is still applicable because we cannot move up file annotation in master (974dfbc9aba).

```kotlin
package foo
@file:JvmName("Utils")/*caret*/ // cannot move up
fun g(x: Int) {
}
```

vladimirdolzhenko:
LGTM, Thanks

PR #2800, opened 2019-11-18
[FIR] Enable blackbox codegen tests for FIR

punzki:
Split it up into 2 commits for easier review. The 2nd commit (4000+ files) is simply adding the disabling directive for FIR tests.

A total of 4776 FIR box tests were added, 757 of which pass. On a local machine (high-end spec), it takes 45 seconds to run them in parallel, and ~2 mins when _not_ in parallel.


compiler/tests-common/tests/org/jetbrains/kotlin/codegen/GenerationUtils.kt
```diff
@@ -65,22 +88,69 @@ object GenerationUtils {
         packagePartProvider: (GlobalSearchScope) -> PackagePartProvider,
         trace: BindingTrace = NoScopeRecordCliBindingTrace()
     ): GenerationState {
-        val analysisResult =
-            JvmResolveUtil.analyzeAndCheckForErrors(files.first().project, files, configuration, packagePartProvider, trace)
-        analysisResult.throwIfError()
+        val project = files.first().project
+        val state = if (configuration.getBoolean(CommonConfigurationKeys.USE_FIR)) {
```
mglukhikh: I think two branches of this `if` are worth extracting to separate functions (will do myself).

mglukhikh:
I will remote run and accept it. Thousands of thanks :)

mglukhikh:
Merged manually

punzki:
> Merged manually

Thanks for doing the follow up fixes too!

PR #2801, opened 2019-11-19
Several line number fix for JVM_IR backend

neetopia:
Several minor changes to fix line number mismatch between current backend and JVM_IR backend for existing test cases.


ilmirus:
Thanks!

PR #2803, opened 2019-11-19
JVM_IR: remap imported @JvmField accesses to parent class

pyos:
The proper fix should come in the form of the promised infrastructure for lazily applying lowerings to referenced files, but for now, this hack should suffice.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
@@ -10,6 +10,7 @@ import org.jetbrains.kotlin.backend.common.ir.*
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.backend.jvm.codegen.MethodSignatureMapper
 import org.jetbrains.kotlin.backend.jvm.codegen.isJvmInterface
+import org.jetbrains.kotlin.backend.jvm.ir.replaceThisByStaticReference
 import org.jetbrains.kotlin.builtins.CompanionObjectMapping.isMappedIntrinsicCompanionObject
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
@@ -22,6 +23,7 @@ import org.jetbrains.kotlin.ir.declarations.impl.*
 import org.jetbrains.kotlin.ir.descriptors.WrappedClassConstructorDescriptor
 import org.jetbrains.kotlin.ir.descriptors.WrappedClassDescriptor
 import org.jetbrains.kotlin.ir.descriptors.WrappedValueParameterDescriptor
+import org.jetbrains.kotlin.ir.expressions.IrExpressionBody
 import org.jetbrains.kotlin.ir.symbols.impl.*
 import org.jetbrains.kotlin.ir.types.IrType
 import org.jetbrains.kotlin.ir.util.*
@@ -37,6 +39,7 @@ class JvmDeclarationFactory(
     private val interfaceCompanionFieldDeclarations = HashMap<IrSymbolOwner, IrField>()
     private val outerThisDeclarations = HashMap<IrClass, IrField>()
     private val innerClassConstructors = HashMap<IrConstructor, IrConstructor>()
+    private val staticBackingFields = HashMap<IrProperty, IrField>()
 
     private val defaultImplsMethods = HashMap<IrSimpleFunction, IrSimpleFunction>()
     private val defaultImplsClasses = HashMap<IrClass, IrClass>()
@@ -151,6 +154,36 @@ class JvmDeclarationFactory(
             }
         else
             getFieldForObjectInstance(singleton)
+
+    fun getStaticBackingField(irProperty: IrProperty): IrField? {
+        // Only fields defined directly in objects should be made static.
+        // Fake overrides never point to those, as objects are final.
+        if (irProperty.origin == IrDeclarationOrigin.FAKE_OVERRIDE) return null
+        val oldField = irProperty.backingField ?: return null
+        val oldParent = irProperty.parent as? IrClass ?: return null
+        if (!oldParent.isObject) return null
+        return staticBackingFields.getOrPut(irProperty) {
+            buildField {
+                updateFrom(oldField)
+                name = oldField.name
+                isStatic = true
+            }.apply {
+                // We don't move fields to interfaces unless all fields are annotated with @JvmField.
+                // It is an error to annotate only some of the fields of an interface companion with
+                // @JvmField, so checking the current field only should be enough.
+                val hasJvmField = oldField.hasAnnotation(JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME)
+                parent = if (oldParent.isCompanion && (!oldParent.parentAsClass.isJvmInterface || hasJvmField))
+                    oldParent.parentAsClass
+                else
+                    oldParent
+                annotations += oldField.annotations
+                initializer = oldField.initializer
+                    ?.replaceThisByStaticReference(this@JvmDeclarationFactory, oldParent, oldParent.thisReceiver!!)
+                    ?.deepCopyWithSymbols(this) as IrExpressionBody?
+                (this as IrFieldImpl).metadata = oldField.metadata
+            }
+        }
+    }
 
     fun getDefaultImplsFunction(interfaceFun: IrSimpleFunction): IrSimpleFunction {
         val parent = interfaceFun.parentAsClass
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
@@ -6,34 +6,20 @@
 package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.ClassLoweringPass
+import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
-import org.jetbrains.kotlin.backend.jvm.codegen.isJvmInterface
 import org.jetbrains.kotlin.backend.jvm.ir.replaceThisByStaticReference
 import org.jetbrains.kotlin.descriptors.Visibilities
-import org.jetbrains.kotlin.ir.IrElement
-import org.jetbrains.kotlin.ir.IrStatement
+import org.jetbrains.kotlin.ir.builders.declarations.addField
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.declarations.impl.IrAnonymousInitializerImpl
-import org.jetbrains.kotlin.ir.declarations.impl.IrFieldImpl
-import org.jetbrains.kotlin.ir.declarations.impl.IrVariableImpl
-import org.jetbrains.kotlin.ir.descriptors.WrappedFieldDescriptor
-import org.jetbrains.kotlin.ir.descriptors.WrappedVariableDescriptor
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrGetFieldImpl
-import org.jetbrains.kotlin.ir.expressions.impl.IrGetValueImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrSetFieldImpl
-import org.jetbrains.kotlin.ir.symbols.IrFieldSymbol
 import org.jetbrains.kotlin.ir.symbols.impl.IrAnonymousInitializerSymbolImpl
-import org.jetbrains.kotlin.ir.symbols.impl.IrFieldSymbolImpl
-import org.jetbrains.kotlin.ir.symbols.impl.IrVariableSymbolImpl
-import org.jetbrains.kotlin.ir.util.deepCopyWithSymbols
-import org.jetbrains.kotlin.ir.util.hasAnnotation
-import org.jetbrains.kotlin.ir.util.isObject
-import org.jetbrains.kotlin.ir.util.patchDeclarationParents
+import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
-import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
-import org.jetbrains.kotlin.load.java.JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME
 
 internal val moveOrCopyCompanionObjectFieldsPhase = makeIrFilePhase(
     ::MoveOrCopyCompanionObjectFieldsLowering,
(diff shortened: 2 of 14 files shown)


compiler/frontend/src/org/jetbrains/kotlin/resolve/resolveUtil.kt
```diff
@@ -13,7 +13,7 @@ fun PropertyDescriptor.hasBackingField(bindingContext: BindingContext?): Boolean
     kind == CallableMemberDescriptor.Kind.FAKE_OVERRIDE -> overriddenDescriptors.any { it.hasBackingField(bindingContext) }
     source is KotlinSourceElement && bindingContext != null -> bindingContext.get(BindingContext.BACKING_FIELD_REQUIRED, this) ?: false
     compileTimeInitializer != null -> true
-//    backingField != null -> true // TODO: in case of DeserializedPropetyDescriptor this function returns incorrect result for FO field
+    backingField != null -> true
```
udalov: We can't know if the deserialized property has a backing field or not (by design), so we always create `backingField` for deserialized properties (see the last lines in `MemberDeserializer.loadProperty`), to be able to load annotations from those fields in the frontend if needed. This change makes `hasBackingField` return true on all deserialized properties, which in turn makes every `IrLazyProperty` have a backing field. In theory this could break some assumptions in the codegen (such as that interfaces can't have fields), but also somewhat worsen the debugging experience, etc.

I propose to only add a backing field if the property is annotated with `@JvmField` instead. Actually I've just done that yesterday in my branch in https://github.com/JetBrains/kotlin/commit/b09452477c14ff173876cc91aab9451d5c9a5b81 (with a preliminary refactoring in https://github.com/JetBrains/kotlin/commit/3be5c81c393a98955e4c08b27b80888d5e330a5f). Would you mind cherry-picking these two commits instead of this change in `resolveUtil.kt`?

pyos: These commits seem to change some origins in IrText tests, which looks suspicious to me - is there a call where the new StubGeneratorExtensions are not passed for some reason? Perhaps it's best to merge that refactoring separately then?

pyos: Right, AbstractIrGeneratorTestCase calls Psi2IrTranslator.generateModule which was not modified to take an additional StubGeneratorExtensions parameter.

udalov: You're right, indeed they do. I'll investigate and after fixing, will push that refactoring first, and then get back to reviewing your changes. Thanks.

udalov: [KOTLIN-CR-3537](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3537)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
```diff
@@ -90,7 +115,7 @@ class PropertiesToFieldsLowering(val context: CommonBackendContext) : IrElementT
     }
 
     private fun substituteGetter(irProperty: IrProperty, expression: IrCall): IrExpression {
-        val backingField = irProperty.backingField!!
+        val backingField = irProperty.backingField!!.resolveFakeOverride()!!.getRealBackingField()
```
udalov: Generally, we do not resolve fake override at compilation time, and I prefer if we didn't do it here as well. The reason is that if a member is moved to a subclass and the clients are not recompiled, in some circumstances the call site can end up calling the incorrect member.

It's not directly applicable to the case of `@JvmField` properties though, because they're always final and moving them is a binary incompatible change anyway. But I think it's still better to keep the old backend's behavior here, at least for some possible obscure migration cases.

pyos: Hm, I didn't know that JVM allows referencing a field through a subclass (or maybe I had simply forgotten...) Resolving the fake override only for the check should be sufficient then (and safe, since objects are final).

pyos: Speaking of fake overrides: the FIR team seems reluctant to add them, and we're also having trouble figuring out exactly what their purpose is. So far it seems like they're only really necessary when the type of a member changes, e.g. when inheriting from an instantiation of a generic type, as then some bridges might be required. For the case you've mentioned it seems sufficient to always reference members through the dispatch receiver's type (when non-static and not a `super` access) and repurpose `superQualifierSymbol` to mean "statically resolved target" (so it would point to some subtype of the field/method's owner if it's static). Is there something we're missing?

udalov: You're not missing anything, and I also think it'd be better to get rid of fake overrides in the IR. It's not immediately obvious how to do this though, since they're serialized to the klib format (see `KotlinIr.proto`) and the deserializer relies on their existence to properly link call sites. We're discussing this issue right now and I hope we can come up with a plan, but it's not an easy refactoring.

compiler/frontend/src/org/jetbrains/kotlin/resolve/resolveUtil.kt
udalov: I've pushed my changes to master (https://github.com/JetBrains/kotlin/commit/608449f41d8725190960b80a27afd54dc632c9ad, https://github.com/JetBrains/kotlin/commit/a485a5ffd6607e6a59fef2e09180508a0bf84bea).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
```diff
@@ -72,8 +76,29 @@ class PropertiesToFieldsLowering(val context: CommonBackendContext) : IrElementT
         return accessor.origin == IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR && Visibilities.isPrivate(accessor.visibility)
     }
 
+    private val remappedFieldCache = mutableMapOf<IrField, IrField>()
+
+    private fun IrField.getRealBackingField(): IrField {
+        val irClass = resolveFakeOverride()!!.parent as? IrClass ?: return this
+        if ((irClass.isCompanion || (irClass.isObject && irClass.visibility != Visibilities.LOCAL)) && !isStatic) {
```
udalov: Why is the visibility check necessary? I don't think local objects are possible. (Anonymous objects should have class kind `CLASS`.)

pyos: I simply copied the condition from MoveOrCopyCompanionObjectFieldsLowering. You're right, it's redundant.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
```diff
@@ -72,8 +76,29 @@ class PropertiesToFieldsLowering(val context: CommonBackendContext) : IrElementT
...
+            // This specific combination of circumstances means it's an imported @JvmField-annotated field.
```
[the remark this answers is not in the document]
pyos: I wouldn't really call this a fix, since the hack effectively creates a second orphaned `IrField` for `result`, so the one you arrive at by going through Z's `declarations` member is not the one that is the `symbol.owner` of the `IrGetField`. Removing `fieldReplacementMap` thus would probably be OK, but would increase the number of cases where this anomaly is visible.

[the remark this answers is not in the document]
udalov: Right, I've tried to do this, but hadn't finished yet (some tests are still failing). Overall I agree that just removing `fieldReplacementMap` is undesirable at this point. Let me look into the `JvmDeclarationFactory` approach once again after merging your changes. Thanks!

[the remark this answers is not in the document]
pyos: If you're interested, I've made my own version of the refactoring: https://github.com/pyos/kotlin/commit/db298f163f6141d5a577e9273999b8d1c76c7f46
The only failing tests are 3 FIR ones because FIR does not set correspondingPropertySymbol for some reason.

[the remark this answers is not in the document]
udalov: Yeah, your version looks very good and it's exactly what I had in mind. I suggest you to attach it to this PR then, since it looks very promising (and I'll have some minor comments).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
pyos: Simlified in both places in the follow-up commit

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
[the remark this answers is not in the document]
pyos: Thank you, added the commit here.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -152,6 +155,37 @@ class JvmDeclarationFactory(
         else
             getFieldForObjectInstance(singleton)
 
+    private fun IrClass.allFieldsAreJvmField() =
+        declarations.all { it !is IrProperty || it.backingField?.hasAnnotation(JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME) == true }
+
+    // We don't move fields to interfaces unless all fields are annotated with @JvmField.
+    // It is an error to annotate only some of the fields of an interface companion with @JvmField.
+    fun getStaticBackingFieldParent(irClass: IrClass): IrClass =
+        if (irClass.isCompanion && (!irClass.parentAsClass.isJvmInterface || irClass.allFieldsAreJvmField()))
```
udalov: I wonder if we can get rid of the linear scan needed in the `allFieldsAreJvmField` criterion, precisely because either all or none of the properties are `@JvmField`. I.e. only look at the JvmField-ness of the property whose field we're moving, not at all of them.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -152,6 +155,37 @@ class JvmDeclarationFactory(
...
+            irClass.parentAsClass
+        else
+            irClass
+
+    fun getStaticBackingField(irProperty: IrProperty, cachedNewParent: IrClass? = null): IrField? =
+        staticBackingFields.getOrPut(irProperty) {
+            if (irProperty.origin == IrDeclarationOrigin.FAKE_OVERRIDE) return null
+            val oldField = irProperty.backingField ?: return null
+            val oldParent = irProperty.parent as? IrClass ?: return null
+            if (!oldParent.isCompanion && !oldParent.isObject) return null
```
udalov: I'd prefer if these simple conditions were checked outside the `getOrPut` call, to avoid recording nulls in this map for every property in the program

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -6,221 +6,127 @@
...
-                copyConsts(irClass)
+            (irClass.declarations.singleOrNull { it is IrClass && it.isCompanion } as IrClass?)?.handle()
         }
-        irClass.replaceFieldReferences(fieldReplacementMap)
     }
 
-    private fun handleObject(irObject: IrClass, fieldReplacementMap: MutableMap<IrFieldSymbol, IrFieldSymbol>) {
-        irObject.declarations.replaceAll {
+    private fun IrClass.handle() {
+        val newParent = context.declarationFactory.getStaticBackingFieldParent(this)
```
udalov: Minor, but you can probably make `getStaticBackingFieldParent` private in `JvmDeclarationFactory` by:
```
val newParent = newDeclarations.firstOrNull()?.parentAsClass ?: return
```

pyos: Yes, this requires making it constant time first though. I'll try that tomorrow.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: Note the `return` - nulls are not actually stored. Yeah, this is indeed confusing and I should not have done that.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -6,221 +6,133 @@
...
-                copyConsts(irClass)
+            (irClass.declarations.singleOrNull { it is IrClass && it.isCompanion } as IrClass?)?.handle()
         }
-        irClass.replaceFieldReferences(fieldReplacementMap)
     }
 
-    private fun handleObject(irObject: IrClass, fieldReplacementMap: MutableMap<IrFieldSymbol, IrFieldSymbol>) {
-        irObject.declarations.replaceAll {
+    private fun IrClass.handle() {
+        val newDeclarations = declarations.map {
```
udalov: `mapNotNull`?

pyos: This would lose the position of these fields relative to other declarations, most importantly anonymous initializers (see the loop in the else branch below).

udalov: Oh, right.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MoveCompanionObjectFieldsLowering.kt
```diff
@@ -6,221 +6,132 @@
...
-                replacementMap.getValue(expression.symbol),
-                /* receiver = */ null,
-                visitExpression(expression.value),
-                expression.type,
-                expression.origin,
-                expression.superQualifierSymbol
-            )
+        expression.symbol.owner.remap()?.let {
+            with(expression) {
+                val newValue = value.transform(this@RemapObjectFieldAccesses, null)
```
pyos: @udalov there was also another problem here due to the use of `visitExpression(expression.value)`, which only transforms the *children* of `expression.value`. This breaks for an obvious reason if `value` is `IrGetField` of a field made static, but to get `PropertiesToFieldsLowering` to generate such a read, the target of it needs to be in a yet-unlowered file, like so:

```kotline
// FILE: A.kt
object A {
    @JvmField
    val result: String

    init { result = Z.result } // -> IrSetField(A.result, IrGetField(Z.result))
}

// FILE: z.kt
object Z {
    @JvmField
    val result: String = "OK"
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertiesToFieldsLowering.kt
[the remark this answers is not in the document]
pyos: The FIR problem is fixed now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -152,6 +155,36 @@ class JvmDeclarationFactory(
         else
             getFieldForObjectInstance(singleton)
 
+    fun getStaticBackingField(irProperty: IrProperty): IrField? {
+        // Only fields defined directly in objects should be made static.
+        // Fake overrides never point to those, as objects are final.
+        if (irProperty.origin == IrDeclarationOrigin.FAKE_OVERRIDE) return null
+        val oldField = irProperty.backingField ?: return null
+        val oldParent = irProperty.parent as? IrClass ?: return null
+        if (!oldParent.isCompanion && !oldParent.isObject) return null
```
udalov: Looks like `!oldParent.isObject` here is unnecessary

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
```diff
@@ -152,6 +155,36 @@ class JvmDeclarationFactory(
...
+            buildField {
+                updateFrom(oldField)
+                name = oldField.name
+                isStatic = true
+            }.apply {
+                // We don't move fields to interfaces unless all fields are annotated with @JvmField.
+                // It is an error to annotate only some of the fields of an interface companion with
+                // @JvmField, so checking the current field only should be enough.
+                val hasJvmField = oldField.hasAnnotation(JvmAbi.JVM_FIELD_ANNOTATION_FQ_NAME)
+                parent = if (oldParent.isCompanion && (!oldParent.parentAsClass.isJvmInterface || hasJvmField))
```
udalov: Looks like `oldParent.isCompanion` here is always true, since it's checked earlier?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: The condition is the other way around, so it's actually `!oldParent.isCompanion` that's unnecessary.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/descriptors/JvmDeclarationFactory.kt
pyos: No, non-companion objects' fields should also be made static (but not moved); the conditions are correct.

udalov:
@pyos Thank you! Merged in https://github.com/JetBrains/kotlin/commit/d221c1f2424e1e3957288ca68cca2c16c1092893.

PR #2804, opened 2019-11-19
[JVM IR] Support Suspend Main

kandersen:
This adds supports for (parameterless) suspend main entry points for
the JVM IR backend.

In case main is a suspend function, it gains a continuation during
lowering, so we simply generate a plain old `public static void
main(String[] args)`. The number of arguements disambiguate them.
This entry point invokes `suspend main` via
`kotlin.coroutines.jvm.internal.RunSuspendKt#runSuspend`.

This PR introduces `runSuspend` as a built-in, and generates the
following `main`, passing `args` as appropriate:

```
  runSuspend { main(args) }
}
```

Some dead code was discovered in AddContinuationLowering


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -223,6 +223,10 @@ private val jvmFilePhases =
         typeAliasAnnotationMethodsPhase then
         stripTypeAliasDeclarationsPhase then
         provisionalFunctionExpressionPhase then
+
+        jvmOverloadsAnnotationPhase then
+        mainMethodGenerationPhase then
```
udalov: Please add a `inventNamesForLocalClassesPhase` -> `mainMethodGenerationPhase` phase prerequisite, maybe with a comment.

udalov:
We need to test the execution somehow. There are some tests on suspend and parameterless main in `CompilerSmokeTest`, maybe they could be duplicated for the JVM IR backend (with `-Xuse-ir`)?

udalov:
Please remove my mention from the commit message :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
kandersen: Done.

kandersen:
> Please remove a mention of me from the commit message :)

Done :)

kandersen:
> We need to test the execution somehow. There are some tests on suspend and parameterless main in `CompilerSmokeTest`, maybe they could be duplicated for the JVM IR backend (with `-Xuse-ir`)?

Yep, looks doable. Will string something together. Should I go ahead and add to CompilerSmokeTest.java? It will be the first IR tests in there, it seems.

udalov:
> Done :)

The commit message still has a mention (https://github.com/kandersen/kotlin/commit/e095c98216cf57a34b02d173985c7a3c1b53ecd6)

> Should I go ahead and add to CompilerSmokeTest.java? It will be the first IR tests in there, it seems.

Yes, please do.

kandersen:
> The commit message still has a mention ([kandersen@e095c98](https://github.com/kandersen/kotlin/commit/e095c98216cf57a34b02d173985c7a3c1b53ecd6))

Apologies, should be fixed now.

kandersen:
@udalov ported the simplest smoketest, and it's not passing. But, I don't think it's on the part of the codegen for the `suspend main` function.

kandersen:
> @udalov ported the simplest smoketest, and it's not passing. But, I don't think it's on the part of the codegen for the `suspend main` function.

I changed my mind, perhaps a simple test showing a case that _does_ work is more useful. At least, it's evidence that the `runSuspend` wrapper is working.

kandersen:
@udalov We determined on friday that this feature is not complete due to interactions with suspend lambdas, which I believe is being revised with the incoming changes to coroutines. I ported integration tests of main that _do_ pass, awaiting the coroutine changes before adding the others.

udalov:
@kandersen Could you please rebase on latest master?

kandersen:
Done. In anticipation of merging, or just to make sure it doesn't go stale? As far as I can tell, the same issue persists: `testHelloAppSuspendMain` doesn't work on the IR backend.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
```diff
@@ -32,7 +32,8 @@ interface JvmLoweredDeclarationOrigin : IrDeclarationOrigin {
     object SYNTHETIC_INLINE_CLASS_MEMBER : IrDeclarationOriginImpl("SYNTHETIC_INLINE_CLASS_MEMBER", isSynthetic = true)
     object INLINE_CLASS_GENERATED_IMPL_METHOD : IrDeclarationOriginImpl("INLINE_CLASS_GENERATED_IMPL_METHOD")
     object GENERATED_ASSERTION_ENABLED_FIELD : IrDeclarationOriginImpl("GENERATED_ASSERTION_ENABLED_FIELD", isSynthetic = true)
-    object GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD : IrDeclarationOriginImpl("GENERATED_MAIN_FOR_PARAMETERLESS_MAIN_METHOD", isSynthetic = true)
+    object GENERATED_EXTENDED_MAIN : IrDeclarationOriginImpl("GENERATED_EXTENDED_MAIN", isSynthetic = true)
+    object SUSPEND_FUNCTION_VIEW : IrDeclarationOriginImpl("SUSPEND_FUNCTION_VIEW")
```
ilmirus: Looks like unused.

kandersen:
Yes, good catch! Rebasing error :p

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/DeclarationOrigins.kt
kandersen: Yes, good catch! Rebasing error :p

compiler/testData/integration/smoke/helloAppSuspendParameterlessMainIR/hello.run.expected
```diff
@@ -0,0 +1,4 @@
+OUT:
+OK
```
ilmirus: This line is not printed in both CI when I run locally. The same happens in `helloAppSuspendMainInMultifileIr`. `helloAppSuspendMainIR`, however, runs OK when I run as a test, but does not print "${args[0]$args[1]}" when I compile the file and run it from CLI.

kandersen: Yeah, this is but a WIP to include the tests in the suite - I haven't worked out what's wrong yet. I think it's the interaction of the testing setup for suspend and the new coroutine backend, but not sure.

Ideas?

ilmirus:
Manually rebased and pushed to master.

PR #2805, opened 2019-11-19
JVM_IR: Deal with special remove method overrides.

madsager:
When generating calls from `remove` bridges, make sure to target the
actual implementation method signatures. This is done by explicitly dealing
with types when generating the call in the bridge and never
mapping special signatures when called from bridges during code gen.


madsager:
This needs more investigation. We are getting into the orphanedCopy method for more methods than I thought. Closing for now and will continue investigating. Sorry for the noise.

gogabr:
Actually, I was just looking at this PR, and my impression is that your fix in `MethodSignatureMapper` might let us get rid of one of the `orphanedCopy`'s uses altogether. (The other use is still needed, unfortunately.)
See https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3542

madsager:
Cool, thanks @gogabr! That aligns with what I'm thinking now as well. Instead of hacking away on the orphanedCopy hack, we should attempt to get rid of it instead. Your change looks good to me.

There are still a number of cases left. I'll continue investigations on my side as well. :)

PR #2806, opened 2019-11-19
KT-32551 New J2K: Non-canonical modifiers order inspection is not applied during convertion of inner super class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32551


idea/src/org/jetbrains/kotlin/idea/inspections/SortModifiersInspection.kt
```diff
@@ -82,5 +55,22 @@ private class SortModifiersFix(private val modifiers: List<KtModifierKeywordToke
             .let { it.second + it.first }
             .forEach { owner.addModifier(it) }
     }
-}
 
+    private fun KtModifierList.modifierKeywordTokens(): List<KtModifierKeywordToken> {
+        return allChildren.toList().asSequence().mapNotNull { it.node.elementType as? KtModifierKeywordToken }.toList()
```
darthorimar: Redundant conversion from sequence to list and back :)

darthorimar:
Thanks!

PR #2807, opened 2019-11-19
Fix nodejs setup task check

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/nodejs/NodeJsSetupTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/nodejs/NodeJsSetupTask.kt
@@ -24,7 +24,7 @@ open class NodeJsSetupTask : DefaultTask() {
     init {
         @Suppress("LeakingThis")
         onlyIf {
-            settings.download && !env.nodeBinDir.isDirectory
+            settings.download && !File(env.nodeExecutable).isFile
         }
     }
 
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/nodejs/NodeJsSetupTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/nodejs/NodeJsSetupTask.kt
@@ -2,6 +2,7 @@ package org.jetbrains.kotlin.gradle.targets.js.nodejs
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.artifacts.repositories.IvyPatternRepositoryLayout
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
@@ -11,6 +12,7 @@ import org.jetbrains.kotlin.gradle.utils.patternLayoutCompatible
 import java.io.File
 import java.net.URI
 
+@CacheableTask
 open class NodeJsSetupTask : DefaultTask() {
     private val settings get() = NodeJsRootPlugin.apply(project.rootProject)
     private val env by lazy { settings.environment }
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/yarn/YarnSetupTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/yarn/YarnSetupTask.kt
@@ -7,12 +7,14 @@ package org.jetbrains.kotlin.gradle.targets.js.yarn
 
 import de.undercouch.gradle.tasks.download.DownloadAction
 import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
 import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin
 import java.io.File
 
+@CacheableTask
 open class YarnSetupTask : DefaultTask() {
     private val settings = project.yarn
     private val env by lazy { settings.environment }


ilgonmic:
@sschuberth  Thank you for your participating!
This part is necessary to not download NodeJS multiple times, if it iexists in gradle home. 
Gradle can perform `UP-TO-DATE` checks on this folder, that's true, but state of `:kotlinNodeJsSetup` is not shareable between projects, so without `onlyIf` check if we build on project, we download NodeJS in one project (task executed), then we switch on another project (and this project has its own `UP-TO-DATE` caches, where `:kotlinNodeJsSetup` is not `UP-TO-DATE`), so when we build another project, NodeJs is downloaded again. Moreover, we can reproduce this issue with only one project, if we build project with `--rerun-tasks` and then check `node` binary modified time. You can check, if we remove this `onlyIf` part, we face with changing of this modified time, but with this check modified time should not be changed.

But we want to download it only once and share between projects to reduce build time
P.S.: In fact, in Yarn there isn't this check, and now it is downloaded every time, when task is not `UP-TO-DATE`, and we plan to fix it (in particular add similar `onlyIf` check)

sschuberth:
> that's true, but state of `:kotlinNodeJsSetup` is not shareable between projects

That' actually not correct. Gradle offers a [build cache](https://docs.gradle.org/current/userguide/build_cache.html) (which is disabled by default, though) "to avoid work when a previous local build has already produced a set of task outputs. But instead of being limited to the previous build in the same workspace, task output caching allows Gradle to reuse task outputs from any earlier build in any location on the local machine. When using a shared build cache for task output caching this even works across developer machines and build agents."

So the right fix to not download NodeJs / Yarn again when building another project would be to globally / per project set `org.gradle.caching=true` in `gradle.properties`.

> In fact, in Yarn there isn't this check

That because I had removed it in https://github.com/JetBrains/kotlin/pull/2564 :wink:

ilgonmic:
Yes, sorry, I meant that by default this option is unavailable
And when someone doesn't enable it (or even doesn't know about it), node will be downloaded every time, when status is not `up-to-date` (and without `build-cache` it is often state)
And IMO it is strange, if user want to start new project (w/o build cache), node is downloaded in gradle home, everything is ok, and when he build new project, he needs to download fully copy of node, that he already have
Seems that we can to lose performance in major of cases

> That because I had removed it in #2564 😉

Yes, and thank you for that. There was deeper problem, that you told about, that we couldn't change yarn version.

ilgonmic:
@sschuberth I thought, that it can be useful to move this check on `upToDateWhen` block, but not to remove
It helps to use `--rerun-tasks` if we want to force redownload (in case of broken `node` home folder)
And it saves performance in basic case without explicit cache enabling

sschuberth:
So, to repeat your use case in my own words: You want `kotlinNodeJsSetup` to not download Node.js again if it has been downloaded before by *another* Gradle build than the current (multi-)project build, but you do *not* want use Gradle's built-in solution for that by leveraging the build cache as it's explicitly required to turn this feature on, and people might forget about that. Does this sum up your use-case correctly?

ilgonmic:
I would like to add to your summary one important point: make configuration as easy as possible. I would be happy to use Gradle built-in solution (it is really more beautiful), but it requires explicit action of user, so I think, that we shouldn't use it in this form.

`gradle.properties` file can be commited to repo, but there is necessary some additional action to add this line and some additional knowledge about this option. Some of Kotlin/JS users can know nothing about gradle

About `@CacheableTask` it make sense, I think

gradle.properties
```diff
@@ -1,3 +1,4 @@
+org.gradle.caching=true
```
ilgonmic: Does this option in our repository affects our user?
It seems that it can affect different Kotlin builds, so I don't think, that it is a good idea to make this change

sschuberth: You're right, this affect the build of the Kotlin plugin itself, not users of the Kotlin plugin. I'll drop it from the commit, although it might make sense for you to enable this anyway, to speed up the build of the Kotlin plugin itself.

sschuberth:
BTW, I've [asked](https://discuss.kotlinlang.org/t/programmatically-enable-the-build-cache/15107) whether it's possible to programmatically enable the build cache.

sschuberth:
So, looks like people from the Gradle forum share my view that [caching should not be enforced by the plugin](https://discuss.kotlinlang.org/t/programmatically-enable-the-build-cache/15107/4) if the user has not set `org.gradle.caching=true`. However, the idea to output a hint in this case to enable the build cache sounds like a good idea to me, what do you think @ilgonmic?

ilgonmic:
If I understand correctly, `build-cache` isn't fine-grained, and if something (every other plugin or task) works incorrectly with it, user can disable `build-cache` and we get problem with every time downloading.
I like beautiful built-in solutions. But now I don't understand problems, when custom plugins and tasks could cache something without built-in gradle cache service. Now I think it gives some advantages for us. Because in basic case we download node only once without any configurations.
What kind of real problem or case do we want to solve, except do it more Gradle-idiomatic (it is important too, but it seems that I don't understand real problem, but can see some disadvantages)?

sschuberth:
> If I understand correctly, `build-cache` isn't fine-grained, and if something (every other plugin or task) works incorrectly with it, user can disable `build-cache` and we get problem with every time downloading.

Disabling the whole build cache in such a case would be the sledge hammer method. The build cache is still task-based, so if you encounter problems with a specific task, the first step should be to call Gradle's auto-generated `clean...` task for that task.

> What kind of real problem or case do we want to solve, except do it more Gradle-idiomatic

Well, as mentioned in the commit message, the current "caching logic" does not work properly if only the `node` binary gets (accidentally) deleted. In that case the `nodeBinDir` still exists and Node.js would not be downloaded again, although the binary is missing.

While you could change the check to look for the `node` binary instead, then the same problem arises if any of the other directory's contents (like something from `lib`) is missing or corrupt. So, to do the caching properly (and also check against corruption), you'd need to compare hashes of *all* files in `nodeDir`. Which is essentially re-implementing Gradle's task snapshot mechanism, so my belief is that it makes more sense to use that mechanism right away.

ilgonmic:
@sschuberth We can add some kind of new flag for `gradle.properties`, which disable check in `onlyIf` block and delegate it to Gradle's `up-to-date` checks. It seems that good solution, because for major cases we don't change behaviour and not lose performance, but in some case of these problems, we can explicit change our behaviour via flag.

Additionally we think about other options of distribution NodeJS and Yarn. For example we investigate option with distribution of nodejs and yarn via maven dependencies (https://youtrack.jetbrains.com/issue/KT-34732). And if we will implement it, the problem with these checks will be irrelevant.

sschuberth:
Actually, if you don't want to go for the fix to rely on the Gradle build cache, then my second-best fix would be to simply make the check more specialized and not check for the existence of `nodeBinDir`, but of the `node` executable. How do you think about that?

sschuberth:
> For example we investigate option with distribution of nodejs and yarn via maven dependencies

BTW, the downside of this solution clearly is that you'd need to package and publish each Node.js and Yarn release you want to support by yourself, instead of relying on the upstream host.

ilgonmic:
> Actually, if you don't want to go for the fix to rely on the Gradle build cache, then my second-best fix would be to simply make the check more specialized and not check for the existence of nodeBinDir, but of the node executable. How do you think about that?

It can make sense, yes!

> BTW, the downside of this solution clearly is that you'd need to package and publish each Node.js and Yarn release you want to support by yourself, instead of relying on the upstream host.

Yes, you're right, it carries some inconvenience for us, and we investigate, is it ok for us
But in issue, there is valid use case and right arguments, why it can be convenient for our users. And for example Protobuf does it as well (although [puppeteer](https://github.com/puppeteer/puppeteer), which also use 3rd party resource (Chrome) download it via HTTP not via dependency). 
In general, it is not final, we just analyze other options.

sschuberth:
> It can make sense, yes!

I've updated the branch accordingly, please have a look.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/nodejs/NodeJsSetupTask.kt
```diff
@@ -24,7 +26,7 @@ open class NodeJsSetupTask : DefaultTask() {
     init {
         @Suppress("LeakingThis")
         onlyIf {
-            settings.download && !env.nodeBinDir.isDirectory
+            settings.download && !env.nodeExecutable.isFile
```
ilgonmic: `nodeExecutable` now is only String, not File
Maybe we need to wrap `env.nodeExecutable` in constructor call of File?

sschuberth: Sorry for the delay, done now.

ilgonmic:
Thank you!
I checked this with `kotlinNodeJsSetup` and `kotlinYarnSetup` and it works.

PR #2808, opened 2019-11-20
JVM_IR: do not set ACC_VARARGS if vararg is not the last parameter

pyos:
Fun fact: this is not actually validated when loading the class; and even when compiling a java file against a class file that does this, only javac 9 has a check (1.8 and before simply crashes with NullPointerException somewhere deep inside the compiler).


max-kammerer:
Thank you!

PR #2809, opened 2019-11-20
Fix java version parsing in KAPT

gavra0:
Handle case when "java.version" returns
"11" or "12". Previously, we expected it to
contain at least one dot in the version string.


yanex:
Merged manually (https://github.com/JetBrains/kotlin/commit/d269c194c771c2b82e42ad30933b017cec9dc75a). Ivan, thank you for your contribution!

PR #2811, opened 2019-11-20
Register kotlin mpp source sets in Android Gradle plugin

sellmair:
Related issue: 
https://youtrack.jetbrains.com/issue/KT-35016

It seems that the Android Gradle Plugin requires the kotlin MPP source set to be registered in order to execute instrumented Android tests. 

I also want to question KotlinPlugin.kt (742) and KotlinPlugin.kt (743)  in this pull request. It seems like the original intent was to still support the "main" sourceSet alongside with the new "androidMain". This might become a problem though because those lines might wrongly add "src/androidTest/kotlin" to the "androidAndroidTest" kotlin mpp source set which is conflicting with the "androidTest" mpp source set (intended for unit tests)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -739,8 +739,13 @@ abstract class AbstractAndroidProjectHandler(private val kotlinConfigurationTool
             val kotlinSourceSet = project.kotlinExtension.sourceSets.maybeCreate(
                 lowerCamelCaseName(kotlinAndroidTarget.disambiguationClassifier, sourceSet.name)
             ).apply {
-                kotlin.srcDir(project.file(project.file("src/${sourceSet.name}/kotlin")))
-                kotlin.srcDirs(sourceSet.java.srcDirs)
+                // TODO resolve androidTest and androidAndroidTest conflict!
```
sellmair: Maybe one could go with something like:

```kotlin
    fun configureTarget(kotlinAndroidTarget: KotlinAndroidTarget) {
        val project = kotlinAndroidTarget.project
        val ext = project.extensions.getByName("android") as BaseExtension

        fun kotlinSourceSetName(sourceSet: AndroidSourceSet): String {
            return lowerCamelCaseName(kotlinAndroidTarget.disambiguationClassifier, sourceSet.name)
        }

        val kotlinSourceSetNames = ext.sourceSets.map(::kotlinSourceSetName)

        ext.sourceSets.all { sourceSet ->
            logger.kotlinDebug("Creating KotlinBaseSourceSet for source set $sourceSet")
            val kotlinSourceSet = project.kotlinExtension.sourceSets.maybeCreate(kotlinSourceSetName(sourceSet)).apply {
                /* 
                Support old sourceSets like "main" or "test" alongside with e.g. "androidMain" and "androidTest"
                while avoid conflicts like "androidTest" and "androidAndroidTest"
                */
                if (sourceSet.name !in kotlinSourceSetNames) {
                    kotlin.srcDir(project.file(project.file("src/${sourceSet.name}/kotlin")))
                    kotlin.srcDirs(sourceSet.java.srcDirs)
                }

                sourceSet.java.srcDir(project.file(project.file("src/${name}/kotlin")))
            }
            sourceSet.addConvention(KOTLIN_DSL_NAME, kotlinSourceSet)

            ifKaptEnabled(project) {
                Kapt3KotlinGradleSubplugin.createAptConfigurationIfNeeded(project, sourceSet.name)
            }
        }

```

sellmair:
I checked my theory regarding original line 742 and 743 by adding simple log statements like
```kotlin
println("Adding ${project.file("src/${sourceSet.name}/kotlin").path} to ${project.path}/${kotlin.name}")
kotlin.srcDir(project.file(project.file("src/${sourceSet.name}/kotlin")))

println("Adding ${sourceSet.java.srcDirs} to ${project.path}/${kotlin.name}")
kotlin.srcDirs(sourceSet.java.srcDirs)
```

This lead to the following output in my demo project: 
```
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/androidTest/kotlin to :mpp-lib/androidAndroidTest Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/androidTest/java] to :mpp-lib/androidAndroidTest Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/androidTestDebug/kotlin to :mpp-lib/androidAndroidTestDebug Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/androidTestDebug/java] to :mpp-lib/androidAndroidTestDebug Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/debug/kotlin to :mpp-lib/androidDebug Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/debug/java] to :mpp-lib/androidDebug Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/main/kotlin to :mpp-lib/androidMain Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/main/java] to :mpp-lib/androidMain Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/release/kotlin to :mpp-lib/androidRelease Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/release/java] to :mpp-lib/androidRelease Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/test/kotlin to :mpp-lib/androidTest Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/test/java] to :mpp-lib/androidTest Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/testDebug/kotlin to :mpp-lib/androidTestDebug Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/testDebug/java] to :mpp-lib/androidTestDebug Kotlin source
Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/testRelease/kotlin to :mpp-lib/androidTestRelease Kotlin source
Adding [/Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/testRelease/java] to :mpp-lib/androidTestRelease Kotlin source

```

The very first line of the output shows that the "androidTest/kotlin" source folder is added to our "androidAndroidTest" source set, which is conflicting with the actual "androidTest" sourceSet

h0tk3y:
There's no direct file system paths conflict between the Kotlin source dirs and the Android Java source dirs, as there are:
* `src/test/java` (Android source set `test` Java sources)
* (!!!) `src/androidTest/kotlin` (Kotlin source set `androidTest` connected to the Android `test` source set)
* (!!!) `src/androidTest/java` (Android source set `androidTest` Java sources)
* `src/androidAndroidTest/kotlin` (Kotlin source set `androidAndroidTest` tied to the Android `androidTest` source set)

The presence of the two source directories marked with *(!!!)* is still very much confusing and should be adressed in some way (we still couldn't find out the proper way and will revisit this as we are reviewing the whole multiplatform project model).

However, we need to have both Kotlin source roots, for both the unit tests and Android tests, so we can't just leave one out due to ambigous naming I described above.

@sellmair Let me prepare a list of the options we consider as possible fixes for the source directory naming ambiguity issue and share the list with you so that maybe you could try to implement and evaluate them.

sellmair:
Regarding the "conflict": 

`Adding /Users/sellmair/Sources/sellmair/mpp-playground/mpp-lib/src/androidTest/kotlin to :mpp-lib/androidAndroidTest Kotlin source`

Doesn't this log mean, that our "androidAndroidTest" source set got the "src/androidTest/kotlin" source folder added? 

The conflict here would then be, that this "src/androidTest/kotlin" source folder is also the default folder for the "androidTest" kotlin source set. The */java source folders are harmless but can be confusing!

But we should decouple the conflict from the actual purpose of the PR: 
Adding the kotlin source set to the Android Gradle Plugin! 
Should I remove the code addressing the "conflict" from this pull request?

sellmair:
@h0tk3y 
I based this PR now onto master and removed the fix for the sourceSet conflicts (I would like to open another, new PR for this in the next days, as mentioned on slack) 

Adding the sourceSet to the Android Gradle Plugin also solved issues for running tests for me and should not be delayed by the rather harmless issue with the conflicting source sets 👍

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -741,6 +741,7 @@ abstract class AbstractAndroidProjectHandler(private val kotlinConfigurationTool
             ).apply {
                 kotlin.srcDir(project.file(project.file("src/${sourceSet.name}/kotlin")))
                 kotlin.srcDirs(sourceSet.java.srcDirs)
+                sourceSet.java.srcDir(project.file("src/$name/kotlin"))
```
h0tk3y: Having this path declared here might cause inconsistency if the path changes in the other places where it's used.
Kotlin source sets have their `srcDirs` initialized here: https://github.com/JetBrains/kotlin/blob/0c9d962f9f474479fd6328df9e13e539e25a31b5/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/sources/KotlinSourceSetFactory.kt#L36
I think it would be safer not to duplicate the path pattern here and reuse the existing content of the `srcDirs`. For example, this could be done as follows:
```
sourceSet.java.srcDirs(*kotlin.srcDirs.minus(sourceSet.java.srcDirs).toTypedArray())
```
(the `.minus(...)` part is a safety measure to avoid duplicating the Java source directories, for instance, if some plugin or a build script uses both collections to pack a sources JAR)

h0tk3y:
@sellmair I confirm that this change fixes the Gradle `connected*AndroidTest` tasks not running the tests if they are only placed under the Kotlin source directory.
I've pushed the branch to the Kotlin repo for integration tests remote run.

h0tk3y:
@sellmair You were right about the conflict of the source directories added to the Kotlin `androidAndroidTest` source set:
* `src/androidAndroidTest/kotlin` added [here](https://github.com/JetBrains/kotlin/blob/0c9d962f9f474479fd6328df9e13e539e25a31b5/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/sources/KotlinSourceSetFactory.kt#L36), and
* `src/androidTest/kotlin` added [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt#L742)
The second case seems to be a mistake. In single-platform projects where a Kotlin source set has the same name as the corresponding Android source set, the second addition just does nothing. In multiplatform projects, it leads to a conflict between `test` and `androidTest`, which, as far as I can see, might even cause redeclaration errors at compile time, if equivalent declarations are present in the two source directories.

It looks like this should be simply removed, so that only the first addition of the two above remains. However, this might be a breaking change for projects that already had instrumented tests placed under that path. Some research is needed to determine whether it is common to have any tests under `src/androidTest/kotlin` in open-source Kotlin multiplatform projects.

h0tk3y:
I have pushed the branch to the Kotlin repo for Gradle integration tests remote run, https://teamcity.jetbrains.com/viewLog.html?buildId=2649230&buildTypeId=Kotlin_dev_GradleIntegrationTests

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
sellmair: Oh, this is pretty clever, didn't think about that! The src folder was just a given in my head! 
I am just a little bit afraid, that this alone (without solving the test source folder conflict) might introduce some strange behavior. With only that line we will add 
- "src/androidTest/kotlin" to Android's "test" sourceSet
- "src/androidTest/kotlin" to Android's "androidTest" sourceSet

However: If we move that line before adding the kotlin source folder" 
`kotlin.srcDir(project.file(project.file("src/${sourceSet.name}/kotlin")))`
Then I expect it to be correct!

sellmair:
@h0tk3y 
I am pretty sure that we can second addition inside the KotlinPlugin [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt#L742)

My gut feeling tells me, that no one is relying on this so far, but let's see! I think it still makes sense to solve this conflict in either https://github.com/JetBrains/kotlin/pull/2829 or in some future PR. 

Otherwise: I already proposed a fix in one comment that should behave exactly as you proposed. So if you think something like this makes sense, then let's go for it!

sellmair:
@h0tk3y: I pushed a conflict resolution that behaves as you proposed. We could throw this commit away if you think it does not make sense to include it right now. 👍

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
sellmair: Pushed this change in the commit that also handles the conflict as you proposed. I now think it only makes sense to ship this with conflict resolution.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -734,18 +734,27 @@ abstract class AbstractAndroidProjectHandler(private val kotlinConfigurationTool
...
+
+        val kotlinSourceSetNames = ext.sourceSets.map(::kotlinSourceSetName)
+
+        ext.sourceSets.all { androidSourceSet ->
+            logger.kotlinDebug("Creating KotlinBaseSourceSet for source set $androidSourceSet")
+            val kotlinSourceSet = project.kotlinExtension.sourceSets.maybeCreate(kotlinSourceSetName(androidSourceSet)).apply {
+                /* Avoiding conflicts like the "androidTest" kotlin source set vs "androidTest" android source set. */
+                if (androidSourceSet.name !in kotlinSourceSetNames) {
+                    kotlin.srcDir(project.file(project.file("src/${androidSourceSet.name}/kotlin")))
+                    kotlin.srcDirs(androidSourceSet.java.srcDirs)
```
h0tk3y: Consider a single-platform project, where the target's `disambiguationClassifier` is null and `kotlinSourceSetName` will thus be always equal to the Android source set name. The condition `androidSourceSet.name !in kotlinSourceSetNames` will always be false, and, in particular, this line won't be executed for any source set:
```
kotlin.srcDirs(androidSourceSet.java.srcDirs)
```
I haven't checked this locally, but, according to my understanding, this will break the ability for a user to place Kotlin sources under the Android source set's `src/*/java` srcDir, which is commonplace in existing projects.

If I understand the intention correctly, this check should only be applied when the Kotlin source set names differ from the Android source set names.

h0tk3y: > this will break the ability for a user to place Kotlin sources under the Android source set's src/*/java srcDir

Sorry, my mistake, this is not the case, as the Java sources are directly added to the task [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt#L996).

sellmair: Oh yes! You are right: I did not think about the fact, that this `disambiguationClassifier` can be null! But I think simply checking for this case as well would do the trick there?

Other than that: 
Yes, this will not register the 
`src/androidTest/java` (android source dir) into the `androidAndroidTest` kotlin src dir, but this is a good thing IMO ☺️ 👍

sellmair: I updated the fix now. With this, all ambiguous java source folders are also disabled 
- `src/androidTest/kotlin` ✅ 
- `src/androidTest/java` :x:

Bildschirmfoto 2019-11-26 um 19 19 08

Sample project can be viewed at https://github.com/sellmair/mpp-playground/tree/connectedAndroidTest-not-executed-fix

I think supporting the  `src/androidTest/java` folder is just confusing since it would be actual instrumented tests!

sellmair: I also tested a single platform Android module. Seems to be working as well:
Bildschirmfoto 2019-11-26 um 21 53 59

sellmair: I'll try to push Gradle Integration tests in the following days 👍

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/AbstractKotlinAndroidGradleTests.kt
```diff
@@ -20,6 +21,31 @@ open class KotlinAndroid32GradleIT : KotlinAndroid3GradleIT() {
...
+            assertContains("AndroidTestKotlin > fail FAILED")
+            assertContains("TestJava > fail FAILED")
+        }
+
+        build("assemble") {
+            assertSuccessful()
+        }
+
+        getModels(KotlinProject::class.java).getModel(":lib")?.run {
+            // TODO: Why is this always null :(
```
sellmair: @h0tk3y : 
Do you know how I could access the source sets to build assertions on them?

sellmair:
@h0tk3y: 
Maybe this PR should be completely dropped in favor of https://github.com/JetBrains/kotlin/pull/2829 ? 
This makes only sense if either #2829 is discarded, or postponed for later releases.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -734,18 +734,32 @@ abstract class AbstractAndroidProjectHandler(private val kotlinConfigurationTool
...
+
+        val kotlinSourceSetNames = ext.sourceSets.map(::kotlinSourceSetName)
+
+        ext.sourceSets.all { androidSourceSet ->
+            logger.kotlinDebug("Creating KotlinBaseSourceSet for source set $androidSourceSet")
+            val kotlinSourceSet = project.kotlinExtension.sourceSets.maybeCreate(kotlinSourceSetName(androidSourceSet)).apply {
+
+                // Avoid overlapping source directories like src/androidTest/kotlin which would otherwise be present  in the
+                // androidTest and androidAndroidTest kotlin source sets
+                if (kotlinAndroidTarget.disambiguationClassifier == null || androidSourceSet.name !in kotlinSourceSetNames) {
```
sellmair: @h0tk3y Is this considered a breaking change?

sellmair: Because I think it is not. The source directory being present in both kotlin source sets already enforced that the sources in the directory were compatible for both source sets. Removing the overlap will fix the unwanted behavior while it should not lead to code not compiling (that did compile beforehand)

sellmair:
Continued in 
https://github.com/JetBrains/kotlin/pull/3918/files

PR #2813, opened 2019-11-21
FIR: set anonymous initializers' parents

pyos:


pyos:
@demiurg906 I was going to bundle this with #2812 since the fix is trivial, but you merged that PR faster than I could update it :)

PR #2814, opened 2019-11-22
JVM_IR: serialize anonymous functions' descriptors

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -239,6 +240,14 @@ open class ClassCodegen protected constructor(
                     }
                 }
             }
+            is MetadataSource.Function -> {
+                val functionProto = serializer!!.functionProto(metadata.descriptor)?.build()
```
udalov: I suppose you can call `createFreeFakeLambdaDescriptor` on this descriptor (as in `ClosureCodegen.generateKotlinMetadataAnnotation`) to also support `functionNtoStringGeneric.kt`.

udalov:
Looks like MetadataSource is not kept for lambdas which are copied in lowerings. For example, if the lambda is in the top level property's body, no metadata is generated, presumably because it's copied in `StaticInitializersLowering`. Also there's a test `IrIncrementalJvmCompilerRunnerTestGenerated.WithJava.Other.testInnerClassNotGeneratedWhenRebuilding`, one of the problems in which is the fact that the generated lambda has no metadata (it has an unexpected kind SYNTHETIC_CLASS).

However, to fix this we might need to introduce a new parameter to `deepCopyWithSymbols`, because some call sites of deep copy in JVM IR seemingly do not need to copy metadata. So I'm OK if we'll just add a bunch of ignored tests to be investigated later.

pyos:
>Looks like MetadataSource is not kept for lambdas which are copied in lowerings. For example, if the lambda is in the top level property's body, no metadata is generated, presumably because it's copied in `StaticInitializersLowering`.

An interesting case, but I wonder why copying the initializers is necessary at all, since either the field is `const val` and does not need any initialization in `<clinit>`, or its initializer should not matter after the lowering and might as well be erased. (Speaking of redundant copying, another example is `getStaticBackingField` which I recently added: the original code was used for both making fields static and copying `const val`s into interfaces, while the new function is only for the former case and `deepCopyWithSymbols(this)` can be replaced with `patchDeclarationParents(this)`.)

udalov:
@pyos Thank you!

PR #2815, opened 2019-11-22
KT-35022 Quickfix "change to var" doesn't remove `const` modifier

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35022


vladimirdolzhenko:
Thanks, LGTM

PR #2817, opened 2019-11-23
KT-34713 "Condition is always 'false'": quickfix "Delete expression" doesn't remove `else` keyword (may break control flow)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34713


vladimirdolzhenko:
LGTM

PR #2818, opened 2019-11-23
IR: refactor InitializersLoweringBase and descendants

pyos:
Also, in JVM_IR specifically do not add `const val` initializers to `<clinit>`, since they are implicitly generated by the JVM from ConstantValue attributes.


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InitializersLowering.kt
```diff
@@ -7,84 +7,72 @@ package org.jetbrains.kotlin.backend.common.lower
...
     }
 }
 
-abstract class InitializersLoweringBase(val context: CommonBackendContext) : ClassLoweringPass {
-    protected fun handleDeclaration(irClass: IrClass, declaration: IrDeclaration): IrStatement? = when (declaration) {
-        is IrField -> handleField(irClass, declaration)
-        is IrAnonymousInitializer -> handleAnonymousInitializer(declaration)
-        else -> null
+abstract class InitializersLoweringBase(open val context: CommonBackendContext) : ClassLoweringPass {
+    protected fun extractInitializers(irClass: IrClass, filter: (IrDeclaration) -> Boolean): List<IrStatement> {
```
udalov: I'd prefer to make it more readable, something like this, what do you think?
```
    protected fun extractInitializers(irClass: IrClass, filter: (IrDeclaration) -> Boolean): List<IrStatement> =
        irClass.declarations.filter(filter).mapNotNull {
            when (it) {
                is IrField -> handleField(irClass, it)
                is IrAnonymousInitializer -> handleAnonymousInitializer(it)
                else -> null
            }
        }.also { 
            irClass.declarations.removeAll { it is IrAnonymousInitializer }
        }
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InitializersLowering.kt
```diff
@@ -7,84 +7,72 @@ package org.jetbrains.kotlin.backend.common.lower
...
 
-    private fun transformInstanceInitializerCallsInConstructors(irClass: IrClass, instanceInitializerStatements: List<IrStatement>) {
-        irClass.transformChildrenVoid(object : IrElementTransformerVoid() {
-            override fun visitInstanceInitializerCall(expression: IrInstanceInitializerCall): IrExpression {
-                val copiedBlock =
-                    IrBlockImpl(UNDEFINED_OFFSET, UNDEFINED_OFFSET, context.irBuiltIns.unitType, null, instanceInitializerStatements)
-                        .copy(irClass) as IrBlock
-                return IrBlockImpl(irClass.startOffset, irClass.endOffset, context.irBuiltIns.unitType, null, copiedBlock.statements)
-            }
+            override fun visitSimpleFunction(declaration: IrSimpleFunction) = declaration
```
udalov: Could you explain why this line is needed?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InitializersLowering.kt
pyos: Sure. My intention was to not have any list copies, but I guess that's my C++ background making itself visible. (Though note that your snippet is slightly incorrect - `removeAll` should also check `filter(it)`.)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/InitializersLowering.kt
pyos: Same reason as `visitClassNew`: to not visit the parts of the tree that don't have any instance initializer calls anyway. The comment above applies to both of them, which would be clearer if there was no empty line, but then IntelliJ would complain about a style guide violation...

pyos:
@udalov I've also added a change in ConstLowering to this PR (the last commit) which is not relevant to InitializersLowering, but is just 3 lines and would otherwise have a merge conflict.

udalov:
@pyos Looks good, merged manually, thanks!

PR #2819, opened 2019-11-24
[FIR] FIR2IR: Populate calls with type arguments and function type parameters with bounds/supertypes.

punzki:


mglukhikh:
Merged manually. Thank you 👍

PR #2820, opened 2019-11-24
Stubs don't break KtTypeAlias::isTopLevel

matklad:
`foo ?: bar is baz` is parsed as `(foo ?: bar) is baz`. So, if stub is non-null, this always returns `False`, because `Boolean` is definitely not a `KtFile`. See https://pl.kotl.in/6DtggcImf 

This Kotlin's "most vexing parse" badly needs a diagnostics.


abelkov:
@matklad Thanks, added your example here: https://youtrack.jetbrains.com/issue/KT-24739#focus=streamItem-27-3818376.0-0

dnpetrov:
Pushed manually. Thank you for your contribution!

PR #2824, opened 2019-11-25
JVM_IR: More special bridge rewriting.

madsager:
If there is a concrete method that will have its argument types
remapped to boxed types, make sure to reflect that in the IR so
that code will be generated for a boxed value instead of a
primitive value.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.backend.jvm.ir.eraseTypeParameters
 import org.jetbrains.kotlin.backend.jvm.ir.hasJvmDefault
+import org.jetbrains.kotlin.codegen.AsmUtil
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.ReceiverParameterDescriptor
 import org.jetbrains.kotlin.descriptors.Visibilities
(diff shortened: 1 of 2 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
```diff
@@ -277,43 +278,63 @@ private class BridgeLowering(val context: JvmBackendContext) : ClassLoweringPass
...
-                    } else {
-                        (body as IrBlockBody).statements.forEach { +it }
                     }
                 }
             }
+        } else {
+            // If the signature of this method will be changed in the output to take a boxed argument instead of a primitive,
+            // rewrite the argument so that code will be generated for a boxed argument and not a primitive.
+            for ((i, p) in valueParameters.withIndex()) {
+                if (p.type.isPrimitiveType() && ourSignature.argumentTypes[i].descriptor.startsWith("L")) {
```
udalov: Maybe `AsmUtil.isPrimitive`?

udalov: Maybe `.sort == Type.OBJECT` instead of `.descriptor.startsWith("L")`?

madsager: Done, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
```diff
@@ -277,43 +279,63 @@ private class BridgeLowering(val context: JvmBackendContext) : ClassLoweringPass
...
-                    if (body is IrExpressionBody) {
-                        +(body as IrExpressionBody).expression
-                    } else {
-                        (body as IrBlockBody).statements.forEach { +it }
                     }
                 }
             }
+        } else {
+            // If the signature of this method will be changed in the output to take a boxed argument instead of a primitive,
+            // rewrite the argument so that code will be generated for a boxed argument and not a primitive.
```
gogabr: I would prefer a more explicit way to indicate that an argument should not be realized as a primitive type, but I guess `makeNullable()` will do for now.

gogabr:
Thank you!

PR #2825, opened 2019-11-25
JVM_IR: Implement data class array member intrinsics.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/IrDataClassArrayMemberHashCode.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.jetbrains.kotlin.backend.jvm.codegen.ExpressionCodegen
+import org.jetbrains.kotlin.codegen.StackValue
+import org.jetbrains.kotlin.ir.expressions.IrFunctionAccessExpression
+import org.jetbrains.kotlin.ir.util.isPrimitiveArray
+import org.jetbrains.kotlin.resolve.jvm.jvmSignature.JvmMethodSignature
+import org.jetbrains.org.objectweb.asm.Type
+import org.jetbrains.org.objectweb.asm.commons.InstructionAdapter
+
+object IrDataClassArrayMemberHashCode : IntrinsicMethod() {
+    override fun toCallable(
```
pyos: This is an old way of implementing intrinsics, it's generally better to override `fun invoke(expression: IrFunctionAccessExpression, codegen: ExpressionCodegen, data: BlockInfo): PromisedValue?`

madsager: Thanks @pyos ! Done. :)

udalov:
Thank you!

PR #2826, opened 2019-11-25
PSI2IR: copy isPropertyWithPlatformField to GeneratorExtensions

pyos:
@udalov your addition of `StubGeneratorExtensions` fixed *direct* accesses to `@JvmField` from other modules, but the fake overrides of these properties still have no fake overrides of fields. When such a fake override is used, a nonexistent accessor is called. (Surprisingly, the two compileKotlinAgainstKotlin tests that have non-static `@JvmField` properties only use them through `super`, managing to avoid the issue by luck.)

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
@@ -65,6 +65,9 @@ object JvmGeneratorExtensions : GeneratorExtensions() {
             descriptor.visibility
         else
             null
+
+    override fun isPropertyWithPlatformField(descriptor: PropertyDescriptor): Boolean =
+        descriptor.hasJvmFieldAnnotation()
 }
 
 class JvmStubGeneratorExtensions : StubGeneratorExtensions() {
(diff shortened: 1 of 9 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
```diff
@@ -65,6 +65,9 @@ object JvmGeneratorExtensions : GeneratorExtensions() {
             descriptor.visibility
         else
             null
+
+    override fun isPropertyWithPlatformField(descriptor: PropertyDescriptor): Boolean =
+        descriptor.hasJvmFieldAnnotation()
```
pyos: I don't really like the idea of copying this, though. Maybe it's best to merge the two extension classes instead?

udalov: Yeah, it'd make sense to merge them. I've separated them initially after thinking that they might have different lifecycles, but it doesn't seem to be the case.

pyos: There were some problems with module boundaries, so instead of moving the entire GeneratorExtensions to org.jetbrains.kotlin.ir.util org.jetbrains.kotlin.ir.util I simply made them inherit from StubGeneratorExtensions.

udalov:
Thanks @pyos! Great catch. Merged manually with minor style fixes.

PR #2827, opened 2019-11-25
JVM_IR: Fix field initialization in secondary enum constructors.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
```diff
@@ -68,13 +68,42 @@ private class EnumClassLowering(val context: JvmBackendContext) : ClassLoweringP
...
+            insertInstanceInitializer()
             assignOrdinalsToEnumEntries()
             lowerEnumConstructors(irClass)
             lowerEnumEntries()
             setupSynthesizedEnumClassMembers()
             lowerEnumClassBody()
         }
 
+        /* Make sure that an instance initializer call exists, so that field initialization code will be generated. */
+        private fun insertInstanceInitializer() {
```
madsager: This code is shared with the JS lowering. Most of the code is different for the lowerings, but we could create a base class and share it between the two.

udalov: Yeah, maybe this could be left as a TODO

madsager: Done.

udalov:
Thanks!

PR #2830, opened 2019-11-26
KT-31912 QF “Convert to anonymous object” do nothing on SAM-interfaces

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31912


fedochet:
@t-kameyama Thanks for your contribution!

PR #2832, opened 2019-11-26
KT-20357: Add sample for Collections distinct and distinctBy methods

TillmannBerg:


libraries/tools/kotlin-stdlib-gen/src/templates/Sets.kt
```diff
@@ -64,6 +64,7 @@ object SetOps : TemplateGroupBase() {
             returns("Sequence<T>")
             body { "return this.distinctBy { it }" }
         }
+        sample("samples.collections.Collections.Sets.distinctAndDistinctBy")
```
ilya-g: It'd be better to place the sample into `Collections.Transformations` class: `distinct` is a general collection operation, the fact that it uses a set to filter out duplicates is an implementation detail.

TillmannBerg: Are you sure? The function definitions are in a `SetOps` object in a `Sets.kt` file. To me the `Sets` class would be the place I'd expect the sample. But if you could confirm that you would prefer the sample in the `Transformations` section, I will move it.

ilya-g: Yes. The result functions generated from this template are placed in `_Arrays.kt`, `_Collections.kt`, `_Sequences.kt` and so on, therefore `distinct(By)` is a generic collection operation.

TillmannBerg: Done, I moved the sample to `collections.Transformations`.

PR #2835, opened 2019-11-27
ForLoopsLowering: Handle Iterable.withIndex() where the type is a bounded type parameter.

punzki:
TODO: Handle Sequences by extending DefaultIterableHandler.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
@@ -731,23 +731,26 @@ internal class WithIndexHandler(context: CommonBackendContext, private val visit
     }
 }
 
-/** Builds a [HeaderInfo] for iterables not handled by more specialized handlers. */
-internal class DefaultIterableHandler(private val context: CommonBackendContext) : ExpressionHandler {
+/** Builds a [HeaderInfo] for Iterables not handled by more specialized handlers. */
+internal open class DefaultIterableHandler(private val context: CommonBackendContext) : ExpressionHandler {
 
-    override fun matchIterable(expression: IrExpression) = true
+    protected open val iterableClassSymbol = context.ir.symbols.iterable
+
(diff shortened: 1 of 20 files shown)


max-kammerer:
Thank you!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -731,23 +731,26 @@ internal class WithIndexHandler(context: CommonBackendContext, private val visit
...
-                            it.valueParameters.isEmpty()
-                }).apply {
-                    dispatchReceiver = expression
-                }
+            // The lowering operates on subtypes of Iterable. Therefore, the IrType could be
+            // a type parameter bounded by Iterable. When that is the case, we cannot get
+            // the class from the type and instead uses the Iterable.iterator() function.
+            val iteratorFun = expression.type.getClass()?.functions?.single {
+                it.name == OperatorNameConventions.ITERATOR &&
+                        it.valueParameters.isEmpty()
```
udalov: Shouldn't we also check that there's no extension receiver, or is it not relevant in this case?

punzki: Member function should always "win out" over any extension function, and there's always an `iterator()` member function on `Iterable`s.

udalov: I meant that a subclass of `Iterable` can have its own member extension iterator, which could lead to `single` throwing an exception here. Something like this:
```
class MyIterable<T> : Iterable<T> {
    ....
    override fun iterator(): Iterator<T> = ...

    // Not an iterator, although has the name "iterator" and no value parameters!
    fun SomeUnrelatedType.iterator(): Double = ...
}
```

udalov: @punzki I've been investigating a related problem in our project bootstrap with JVM IR, and fixed it along with this one here, please take a look: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3586

PR #2836, opened 2019-11-27
JVM/JVM_IR: fix mapping of KClass in annotation classes

pyos:
* JVM incorrectly mapped T<KClass<...>> to T<Class<...>> because the annotation-ness of the type mapping mode was inherited one level down into a generic signature independent of T

 * JVM_IR was even worse as it did not use VALUE_FOR_ANNOTATION at all, mapping T<T<KClass<...>> to T<T<Class<...>> as well.

The correct behavior is to map KClass to Class only at top level or as an argument of Array.


udalov:
Could you please [report an issue](https://kotl.in/issue) about the bug in the current JVM backend? Since it's a breaking change (some usages from Java may stop working) I'm afraid we need to pass it through the Language Committee.

Otherwise this looks very good.

pyos:
Reported as [KT-35207](https://youtrack.jetbrains.com/issue/KT-35207).

udalov:
Thank you!

We'll run the Language Committee approval process asynchronously.

PR #2838, opened 2019-11-27
JVM_IR: Generate fake inlining local variables for debugging.

madsager:


madsager:
@pyos @ilmirus @max-kammerer

max-kammerer:
Looks like same could be implemented as simple lowering

madsager:
Thanks! That is indeed more direct. Please take another look @max-kammerer

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.jetbrains.kotlin.load.java.JvmAbi
+
+internal val fakeInliningLocalVariablesLowering = makeIrFilePhase(
+    ::FakeInliningLocalVariablesLowering,
+    name = "FakeInliningLocalVariablesLowering",
+    description = "Add fake locals to identify the range of inlined functions and lambdas"
+)
+
+internal class FakeInliningLocalVariablesLowering(val context: JvmBackendContext) : IrElementVisitorVoid, FileLoweringPass {
+    private val localFunForLambdaToCallee = mutableMapOf<IrFunction, IrFunction>()
```
pyos: This lowering looks like it could be done in a single pass with no state.

madsager: Yes, of course. :) Thanks @pyos !

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    override fun visitElement(element: IrElement) {
+        element.acceptChildrenVoid(this)
+    }
+
+    override fun visitCall(expression: IrCall) {
+        expression.acceptChildrenVoid(this)
+        val callee = expression.symbol.owner
+        if (callee.isInline) {
```
max-kammerer: Could we reuse IrInlineReferenceLocator here (maybe be with expanding `inlineReferences = mutableSetOf<IrCallableReference>()` to inlineReferences = mutableSetOf<IrCallableReference, IrFunction (called function)>())

madsager: We can (@pyos suggested that as well offline). I uploaded a version that does that. I'm adding methods for handling callable reference inline function arguments and for handling lamdba inline function arguments. In the fake variable lowering I do not record them in the collections as they will not be used. I could avoid doing that and just record the callee as well for lambda params and iterate them after the visit. I prefer to not build the collections when not needed, but what do you think?

max-kammerer: Your solution looks much better! Thanks for pr! Will merge it after remote run

max-kammerer:
Thank you!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private fun IrFunction.addFakeLocalVariable(name: String) {
+        context.createIrBuilder(symbol).run {
+            body = irBlockBody {
+                // Create temporary variable, but make sure it's origin is `DEFINED` so that
+                // it will materialize in the code.
+                createTmpVariable(irInt(0), name, origin = IrDeclarationOrigin.DEFINED)
+                if (body is IrExpressionBody) {
+                    +(body as IrExpressionBody).expression
```
max-kammerer: @madsager Last monent  thought: seems this branch is not executed at all cause there should be additional return

madsager: Nice catch, I have the same error in bridge lowering. I'll see if I can generate test cases hitting this.

PR #2843, opened 2019-11-28
KT-35175 Clarify documentation for XorWowRandom

Kerooker:
This commit changes XorWowRandom.kt documentation to include a HTTPS link instead of a HTTP link, for security purposes.

This commit also changes the same documentation to reflect the real cycles from the `xorwow` algorithm:

    Simple and very fast (125 million/sec), the elements in its cycle of 2^192 − 2^32 easily pass all the tests in Diehard.

 Marsaglia, G. 2003. Xorshift RNGs. J. Statis. Soft. 8, 14, p. 5

 At last, this commit explicits which part of the code is the implementation of the mentioned algorithm.

 Closes KT-35175

### DIFF
--- a/libraries/stdlib/src/kotlin/random/XorWowRandom.kt
+++ b/libraries/stdlib/src/kotlin/random/XorWowRandom.kt
@@ -6,11 +6,14 @@
 package kotlin.random
 
 /**
- * Random number generator, algorithm "xorwow" from p. 5 of Marsaglia, "Xorshift RNGs".
+ * Random number generator, using Marsaglia's "xorwow" algorithm
  *
- * Cycles after 2^160 * (2^32-1) repetitions.
+ * Cycles after 2^192 - 2^32 repetitions.
  *
- * See http://www.jstatsoft.org/v08/i14/paper for details.
+ * For more details, see Marsaglia, George (July 2003). "Xorshift RNGs". Journal of Statistical Software. 8 (14). doi:10.18637/jss.v008.i14
+ *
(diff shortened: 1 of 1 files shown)


libraries/stdlib/src/kotlin/random/XorWowRandom.kt
```diff
@@ -6,11 +6,13 @@
...
- * Random number generator, algorithm "xorwow" from p. 5 of Marsaglia, "Xorshift RNGs".
+ * Random number generator, using Marsaglia's "xorwow" algorithm
  *
- * Cycles after 2^160 * (2^32-1) repetitions.
+ * For more details, see Marsaglia, George (July 2003). "Xorshift RNGs". Journal of Statistical Software. 8 (14). doi:10.18637/jss.v008.i14
  *
- * See http://www.jstatsoft.org/v08/i14/paper for details.
+ * Available at https://www.jstatsoft.org/v08/i14/paper
+ *
+ * Cycles after 2^192 - 2^32 repetitions.
```
ilya-g: Minor: this sentence would look better in the beginning of the details, i.e. right after the summary line.

Kerooker: See if it's better now

Kerooker:
@ilya-g as an addendum: I re-read the whole implementation and examined similar codes around the web. The implementation does what xorwow does.

This isn't exactly visible in a commit, but I took the chance to review this code once more.

ilya-g:
Merged. 
Thanks for another review of the implementation, it never hurts.

PR #2845, opened 2019-11-29
KT-34511 kotlin.KotlinNullPointerException after using intention Replace Java Map.forEach with Kotlin's forEach for Map with Pairs as keys

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34511


mglukhikh:
Merged 👍

PR #2846, opened 2019-11-29
JVM_IR: implement -Xpolymorphic-signature

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
```diff
@@ -50,7 +50,8 @@ class MemoizedInlineClassReplacements {
                 !it.hasMangledParameters ||
                         it.isSyntheticInlineClassMember ||
                         it.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA ||
-                        it.origin == IrDeclarationOrigin.DELEGATED_PROPERTY_ACCESSOR -> null
+                        it.origin == IrDeclarationOrigin.DELEGATED_PROPERTY_ACCESSOR ||
+                        it.origin == JvmLoweredDeclarationOrigin.POLYMORPHIC_SIGNATURE_INSTANTIATION -> null
```
pyos: Note: in general, both backends generate garbage when a Java method takes an instance of an inline class as an argument.

```java
public class A {
    public int f(Z z) { return z.getX(); }
}
```
```kotlin
inline class Z(val x: Int)

fun main() {
    println(A().f(Z(1)))
}
```

-> call to `A."f-X5_cmG0":(LZ;)I` in JVM_IR and `Method A."f-IQRRRT4":(LZ;)I` in JVM. cc @sfs

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (!function.hasAnnotation(polymorphicSignatureAnnotation))
+            return null
+        assert(function.valueParameters.singleOrNull()?.varargElementType != null) {
+            "@PolymorphicSignature methods should only have a single vararg argument: ${dump()}"
+        }
+
+        val values = (getValueArgument(0) as IrVararg?)?.elements?.map {
+            when (it) {
+                is IrExpression -> it
+                is IrSpreadElement -> it.expression // `*xs` acts as `xs` (for compatibility?)
```
udalov: > (for compatibility?)

More like there's no other option here. We don't know what's inside the array under the spread, so we can't specialize the signature. But maybe spread operator should be disallowed on polymorphic method calls, I suppose we can think about it later.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
```diff
@@ -0,0 +1,87 @@
...
+    // If the return type is Any?, then it is also polymorphic (e.g. MethodHandle.invokeExact
+    // has polymorphic return type, while VarHandle.compareAndSet does not).
+    override fun visitTypeOperator(expression: IrTypeOperatorCall): IrExpression =
+        (expression.argument as? IrCall)?.takeIf { it.type.isNullableAny() }?.transform(expression.typeOperand)
+            ?: super.visitTypeOperator(expression)
+
+    override fun visitCall(expression: IrCall): IrExpression =
+        expression.transform(null) ?: super.visitCall(expression)
+
+    private fun IrCall.transform(castedReturnType: IrType?): IrCall? {
```
udalov: casted -> cast

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
```diff
@@ -0,0 +1,87 @@
...
+            "@PolymorphicSignature methods should only have a single vararg argument: ${dump()}"
+        }
+
+        val values = (getValueArgument(0) as IrVararg?)?.elements?.map {
+            when (it) {
+                is IrExpression -> it
+                is IrSpreadElement -> it.expression // `*xs` acts as `xs` (for compatibility?)
+                else -> throw AssertionError("unknown IrVarargElement: $it")
+            }
+        } ?: listOf()
```
udalov: I'd throw exception here instead of generating incorrect bytecode (in case they add a polymorphic method with something other than a vararg parameter to a future JDK)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
```diff
@@ -0,0 +1,87 @@
...
+class PolymorphicSignatureLowering(val context: JvmBackendContext) : IrElementTransformerVoid(), FileLoweringPass {
+    override fun lower(irFile: IrFile) {
+        if (context.state.languageVersionSettings.supportsFeature(LanguageFeature.PolymorphicSignature))
+            irFile.transformChildrenVoid()
+    }
+
+    // If the return type is Any?, then it is also polymorphic (e.g. MethodHandle.invokeExact
+    // has polymorphic return type, while VarHandle.compareAndSet does not).
+    override fun visitTypeOperator(expression: IrTypeOperatorCall): IrExpression =
+        (expression.argument as? IrCall)?.takeIf { it.type.isNullableAny() }?.transform(expression.typeOperand)
```
udalov: Do we need to check `expression.operator` here for `as`/`as?`? Otherwise looks like `invokeExact(...) is X` would cause the return type to be incorrectly specialized to `X`.

pyos: Yes, looks like we do.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
pyos: >But maybe spread operator should be disallowed on polymorphic method calls

That's what I initially expected to happen. The behavior of spreads is extremely weird right now, for example a spread and positional arguments after it are not concatenated into a single array.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
pyos: There's already an `assert` that checks that the only parameter is vararg above, and a cast that would fail with a non-IrVararg argument. The `null` case happens when the call has no arguments at all, in which case the generated descriptor `()...` is correct.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
udalov: > a spread and positional arguments after it are not concatenated into a single array

:o

Please report an [issue](https://kotl.in/issue), so that we don't forget to address it before 1.4.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
udalov: Oops. You're right, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PolymorphicSignatureLowering.kt
pyos: Created [KT-35226](https://youtrack.jetbrains.com/issue/KT-35226)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
udalov: Please report an [issue](https://kotl.in/issue).

pyos: [KT-35229](https://youtrack.jetbrains.com/issue/KT-35229)

udalov:
Thank you!

PR #2847, opened 2019-11-29
Add explicit irReturn when transforming expression body to block body.

madsager:


madsager:
@max-kammerer I cannot generate a test case hitting this. However, let's put in the explicit irReturn anyway. We do generate IrExpressionBody in lowerings, so at some point they might end up here.

It seems like ExpressionCodegen actually papers over this at this point and would generate a return of the value of the last thing in the IrBlockBody if there is no return. Have a look at ExpressionCodegen.kt line 197. But, we shouldn't rely on this (and that is not what is happening now - I have not been able to create an example hitting these lowerings with IrExpressionBody reaching these rewritings.)

max-kammerer:
Thank you!

PR #2853, opened 2019-12-03
JVM_IR: generate more correct parameter metadata

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -249,13 +257,52 @@ fun generateParameterAnnotations(
...
+            irParameter.origin == BOUND_RECEIVER_PARAMETER -> Opcodes.ACC_SYNTHETIC
+            irParameter.origin.isSynthetic -> Opcodes.ACC_SYNTHETIC
+            else -> 0
+        }
+        mv.visitParameter(name, access)
+    }
+}
+
+private fun getNameForReceiverParameter(irFunction: IrFunction, state: GenerationState): String {
+    if (!state.languageVersionSettings.supportsFeature(LanguageFeature.NewCapturedReceiverFieldNamingConvention)) {
```
max-kammerer: It's also worth to support asmLike/receiverMangling test in IR (as separate pr)

max-kammerer:
Thank you!

max-kammerer:
NB: Actually `AsmUtil.IS_BUILT_WITH_ASM6` is always false. It was actual for Idea 183.2

max-kammerer:
NB: just removed obsolete logic https://github.com/JetBrains/kotlin/tree/rr/mb/asm6

PR #2855, opened 2019-12-04
Add KCallable.parameterCount()

stsypanov:
Hello, originally the idea of those changes was born while improving `BeanUtils.instantiateClass` in Spring Framework here: https://github.com/spring-projects/spring-framework/pull/24104

It turned out that current implementation of `kotlin.reflect.KCallable` doesn't provide a separate method to get parameter count, this is why `KCallable.parameters()` is used. This method allocates and stores list of `KParameter` which is not always necessary for the cases when we need only to get param count (e.g. when it is 0 we can call no-args constructor).

In Java `java.lang.reflect.Method` originally provided only `getParameterTypes()` method, which cloned underlying array. Later `Method.getParameterCount()` was introduced for the sake of performance (it is allocation-free), see https://bugs.openjdk.java.net/browse/JDK-8004729

Usage of `KCallable.parameterCount()` allows to postpone or even rid unnecessary object allocation.

### DIFF
--- a/core/builtins/src/kotlin/reflect/KCallable.kt
+++ b/core/builtins/src/kotlin/reflect/KCallable.kt
@@ -29,6 +29,14 @@ public actual interface KCallable<out R> : KAnnotatedElement {
     public val parameters: List<KParameter>
 
     /**
+     * Parameters required to make a call to this callable.
+     * If this callable requires a `this` instance or an extension receiver parameter,
+     * they come first in the list in that order.
+     */
+    public val parameterCount: Int
+
+
+    /**
      * The type of values returned by this callable.
      */
     public val returnType: KType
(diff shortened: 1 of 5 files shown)


udalov:
Hi! I don't think such method will help in the current implementation of kotlin-reflect, since we still need to fully load the underlying model to determine the correct number of parameters, which takes a lot more time than allocating the list. In this change, it's happening in the first line of `parameterCount` (`val descriptor = descriptor`). You can verify that this is the case by measuring the time (by [JMH](https://openjdk.java.net/projects/code-tools/jmh/)) used for `parameterCount` against the time for `parameters.size`. I don't see a way to compute the parameter count without using the internal model right now.

vladimirdolzhenko:
Is there any update on it ? It's half year since last commit and this PR looks abandon - I tend to decline this PR if there is no any update for another week.

udalov:
Agreed. I'll close this because of the reason highlighted above. Feel free to get back with another solution for this issue.

PR #2856, opened 2019-12-04
JVM_IR: cast field receiver to owner type

pyos:
This is necessary because the descriptors for fake overrides incorrectly (?) have dispatch receiver type equal to the substituted type of the real implementation's owner, not the actual owner. Since psi2ir uses descriptors to generate IR fake overrides and insert implicit casts, the resulting IR contains incorrect casts, e.g. in

    open class A { fun f() = 1 }
    class B : A { fun g(x: Any) = if (x is B) x.f() else null }

the call to `f` points to the fake override `B::f`, but `x` is casted to `A` instead of `B`. While for functions, like in the example above, this problem is masked in ExpressionCodegen by casting the receiver (the bytecode contains two casts in a row), this was not done for fields.


max-kammerer:
@dnpetrov Seems you should be already aware about this problem. Could we fix it on IR side?

dnpetrov:
@max-kammerer @pyos it actually looks like a psi2ir problem to me, I'll investigate.

dnpetrov:
Yes, it can be fixed in psi2ir, although it uses a hack, because "old" FE doesn't provide full type information here. Fix pending.

PR #2857, opened 2019-12-04
KT-23322: Document 'reduce' operation behavior on empty collections

TillmannBerg:
At present the behavior of the `reduce` method when called on an empty argument is undocumented. I added the info that an `UnsupportedOperationException` is thrown.

### DIFF
--- a/libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
+++ b/libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
@@ -644,6 +644,7 @@ object Aggregates : TemplateGroupBase() {
             to current accumulator value and each ${f.element} with its index in the original ${f.collection}.
             @param [operation] function that takes the index of ${f.element.prefixWithArticle()}, current accumulator value
             and the ${f.element} itself and calculates the next accumulator value.
+            @throws UnsupportedOperationException if the ${f.collection} is empty.
             """
         }
         returns("T")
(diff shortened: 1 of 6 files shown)


libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
```diff
@@ -846,7 +860,12 @@ object Aggregates : TemplateGroupBase() {
         inline()
         specialFor(ArraysOfUnsigned) { inlineOnly() }
 
-        doc { "Accumulates value starting with last ${f.element} and applying [operation] from right to left to each ${f.element} and current accumulator value." }
+        doc {
+            """
+            Accumulates value starting with last ${f.element} and applying [operation] from right to left to each ${f.element} and current accumulator value.
+            @throws UnsupportedOperationException if the ${f.collection} is empty.
```
ilya-g: As per the migration plan of https://youtrack.jetbrains.com/issue/KT-33874 we're going to document in 1.4 only that an exception is thrown in that case, without specifying its concrete class.

We can also advise not to catch it, but use `reduceOrNull` alternative instead which is going to be introduced soon (https://youtrack.jetbrains.com/issue/KT-33761).

ilya-g:
Thank you for the PR, we've managed to document it ourselves, see 4d7597fa5aadb0093b7b90fa76823271a9336cdc

PR #2858, opened 2019-12-04
[FIR] Add FirCheckNotNullCall converted to CHECK_NOT_NULL intrinsic function call

punzki:
https://youtrack.jetbrains.com/issue/KT-31212

@mglukhikh @demiurg906 - Here's my pass at implementing a node for the `!!` operator. I mostly followed your suggestion to use a synthetic call to resolve the expression type.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -833,7 +844,7 @@ class FirDataFlowAnalyzer(private val components: FirAbstractBodyResolveTransfor
         if (fir is FirThisReceiverExpressionImpl) {
             return variableStorage.getOrCreateNewThisRealVariable(fir.calleeReference.boundSymbol ?: return null)
         }
-        val symbol: AbstractFirBasedSymbol<*> = fir.resolvedSymbol ?: return null
+        val symbol = fir.resolvedSymbol as? FirVariableSymbol<*> ?: return null
```
punzki: I made this change because of an issue I found with `codegen/box/unaryOp/callNullable.kt` (snippet below):

```kt
    val a1: Byte? = 1.unaryMinus()
    val a2: Short? = 1.unaryMinus()
    val a3: Int? = 1.unaryMinus()
// ...
    if (a1!! != (-1).toByte()) return "fail 1"
    if (a2!! != (-1).toShort()) return "fail 2"
    if (a3!! != -1) return "fail 3"
```

`exitVariableDeclaration` would create a "real variable" for the initializer with the `unaryMinus` function as the symbol, and create an alias variable for `a1` pointing to the `unaryMinus` variable. It would also create similar alias variables for `a2`, `a3` etc. Each `!!` call would add the not-nullable type to the type info for the `unaryMinus` function, which does not make sense (it builds an intersection type, e.g., `Byte & Short & Int` after `a3!!`).

FWIW, this issue also exists with the `as` type operator. If you changed all the `!!` to `as` (e.g., `a1 as Byte` instead of `a1!!`, etc.), you would see some failure: `Caused by: java.lang.ClassCastException: java.lang.Short cannot be cast to java.lang.Byte` (JVM tries to cast `a2` to `Byte`).

This change makes sure that only put real variable symbols in storage. Let me know if this seems right (or not)!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirCallCompletionResultsWriterTransformer.kt
```diff
@@ -223,49 +221,29 @@ class FirCallCompletionResultsWriterTransformer(
...
-        val calleeReference = whenExpression.calleeReference as? FirNamedReferenceWithCandidate ?: return whenExpression.compose()
+    // Transformations for synthetic calls generated by FirSyntheticCallGenerator
 
-        val whenExpression = whenExpression.transformChildren(this, data) as FirWhenExpression
+    override fun transformWhenExpression(whenExpression: FirWhenExpression, data: Nothing?) = transformSyntheticCall(whenExpression)
+    override fun transformTryExpression(tryExpression: FirTryExpression, data: Nothing?) = transformSyntheticCall(tryExpression)
+    override fun transformCheckNotNullCall(checkNotNullCall: FirCheckNotNullCall, data: Nothing?) = transformSyntheticCall(checkNotNullCall)
 
-        val declaration = whenExpression.candidate()?.symbol?.fir as? FirMemberFunction<*> ?: return whenExpression.compose()
+    private inline fun <reified D> transformSyntheticCall(syntheticCall: D): CompositeTransformResult<FirStatement>
```
punzki: Refactored duplicate code here. Can we also de-dupe code in `transformFunctionCall` as well?

compiler/fir/tree/src/org/jetbrains/kotlin/fir/FirRenderer.kt
```diff
@@ -923,6 +923,12 @@ class FirRenderer(builder: StringBuilder) : FirVisitorVoid() {
         }
     }
 
+    override fun visitCheckNotNullCall(checkNotNullCall: FirCheckNotNullCall) {
+        checkNotNullCall.annotations.renderAnnotations()
+        print("!!")
```
demiurg906: I think, it's better to swap this and next lines, because of `expression!!` is more readable than `!!expression`

punzki:
> I've not found any mistakes. However, I've run our "modularized test" and number of erroneously resolved types has increased for some reason. We will compare error dumps and either fix problems ourselves or write here what particular places are broken.
> 
> Good news: this PR saves us about 2% of total resolve time.

Is there any way I can run the modularized test myself?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -280,6 +280,31 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
...
+            return checkNotNullCall.compose()
+        }
+
+        checkNotNullCall.transformArguments(transformer, ResolutionMode.ContextDependent)
+
+        val result = components.syntheticCallGenerator.generateCalleeForCheckNotNullCall(checkNotNullCall)?.let {
+            callCompleter.completeCall(it, data.expectedType)
+        } ?: run {
+            checkNotNullCall.resultType =
+                FirErrorTypeRefImpl(null, FirSimpleDiagnostic("Can't resolve checkNotNull call", DiagnosticKind.InferenceError))
```
demiurg906: Replace error message with `Can't resolve !! operator call`. It prevent possible confuse if there is a regular function named `checkNotNull` in user's project

compiler/fir/resolve/testData/resolve/smartcasts/bangbang.txt
```diff
@@ -4,28 +4,12 @@ FILE: bangbang.kt
...
-        when (lval <bangbang>: R|A?| = R|<local>/a|) {
-            ==($subj$, Null(null)) ->  {
-                throw R|kotlin/KotlinNullPointerException.KotlinNullPointerException|()
-            }
-            else ->  {
-                R|<local>/<bangbang>|
-            }
-        }
-        .R|/A.foo|()
+        !!(R|<local>/a|).R|/A.foo|()
```
demiurg906: Add test case with bangbang over type parameter to this test please

```
fun <X : A?> test_6(x: X) {
    x!!.foo()
}
```

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
demiurg906: Yeah, it seems good

Actually, there is a bunch of bugs with aliased variables and I am planning to rewrite all code that is relative to them (and, probably, get rid of alias variables at all)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -338,6 +339,16 @@ class FirDataFlowAnalyzer(private val components: FirAbstractBodyResolveTransfor
         graphBuilder.exitJump(jump).mergeIncomingFlow()
     }
 
+    // ----------------------------------- Check not null call -----------------------------------
+
+    fun exitCheckNotNullCall(checkNotNullCall: FirCheckNotNullCall) {
+        // Note: For DFA purposes, `arg!!` is equivalent to `(arg as TypeOfArg)`
+        val node = graphBuilder.exitCheckNotNullCall(checkNotNullCall).mergeIncomingFlow()
+        val type = checkNotNullCall.typeRef.coneTypeSafe<ConeKotlinType>()?.withNullability(ConeNullability.NOT_NULL) ?: return
```
demiurg906: It's enough to add `Any` to set of possible types (`FirDataFlowInfo(setOf(session.builtinTypes.anyType.coneTypeUnsafe()), emptySet())`)

After smartcast there will be intersection type `T? & Any` that will be reduced to `T`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirCallCompletionResultsWriterTransformer.kt
demiurg906: We want refactor that place too

demiurg906: > Can we also de-dupe code in transformFunctionCall as well?

We also want refactor this place, but don't decide what is the better way to do

punzki: > We also want refactor this place, but don't decide what is the better way to do

Sorry, I don't understand... Can you clarify what you mean by "don't decide"? Would you like `transformFunctionCall` refactored as well, to remove the duplicate code?

demiurg906: Sorry for my english
I meant that we want unify code in code resolver, but didn't decide yet, how to change tree nodes for that
There is an open question about annotation calls that, probably, should be resolved via FirCallResolver, so we don't refactor resolver and tree before we design annotations

punzki: Okay, I will leave this alone then. It makes sense to unify in the call resolver.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
punzki: Makes sense. Done.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/FirRenderer.kt
punzki: I agree. I also eliminated the surrounding parentheses from calling `visitCall`, by directly rendering the argument.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
punzki: Ah good to know! Done. Now I understand why that's done elsewhere (e.g., for `null` equality expressions).

compiler/fir/resolve/testData/resolve/smartcasts/bangbang.txt
punzki: This was already covered in some other test (`compiler/testData/ir/irText/expressions/bangbang.kt`) but anyway I will add it here too.

punzki: I guess it's not quite the same, so I am adding both tests here.

demiurg906: It's good to have such testcase in diagnostics tests because they are they run much more often than fir2ir tests

punzki: Please ignore what I said as they don't make sense; the test in the other file does not call a function and is just using `Any`.

demiurg906:
@punzki Please, can you commit fixes to PR as separate commit in future please? It's easier to review than force push already reviewed commits. After review will done they can be squashed

punzki:
> @punzki Please, can you commit fixes to PR as separate commit in future please? It's easier to review than force push already reviewed commits. After review will done they can be squashed

My apologies! I will do so next time on FIR changes. When we've worked with the JVM backend team (cc @udalov) on PRs, typically been asked to squash before updating the PR. Probably because jt makes merging more straightforward.

punzki:
> > @punzki Please, can you commit fixes to PR as separate commit in future please? It's easier to review than force push already reviewed commits. After review will done they can be squashed
> 
> My apologies! I will do so next time on FIR changes. When we've worked with the JVM backend team (cc @udalov) on PRs, typically been asked to squash before updating the PR. Probably because jt makes merging more straightforward.

@demiurg906 For the follow-up commits, are you looking for `fixup!` commits or just regular commits? Or either?

demiurg906:
> @demiurg906 For the follow-up commits, are you looking for fixup! commits or just regular commits? Or either?

It's fine to `fixup!` and rebase before merging from your or our side, thanks

punzki:
@demiurg906 @mglukhikh - I investigated the differences in the modularized test resolved types as shared to me by @demiurg906:

1. Some of the additional errors are due to `FirCheckNotNullCall` not supported in the HTML dump. Fixed in the latest commit. There were other unsupported FIR nodes as well but I did not fix those at this time.
2. The rest of the additional errors are due to smartcasts of `this`. When there are multiple smartcasts of `this` (example below), all qualified access to `this` (e.g., `FirThisReceiverExpressionImpl`) after the first one get transformed to `FirExpressionWithSmartcastImpl` because there is now smartcast info for it. However  in `FirDataFlowAnalyzer.getOrCreateRealVariable()`, we do not "unwrap" the `FirExpressionWithSmartcastImpl` to get the original expression, so we cannot retrieve the variable for `this`. The previous code in `getOrCreateRealVariable()` was masking this issue. Example below:
```kt
fun Any.test_1(): Int = when {
    this is List<*> -> this.size
    this is String -> this.length
    else -> 0
}
```
3. There were some errors that should be fixed in this PR. The errors were due to the issues I [previously described](https://github.com/JetBrains/kotlin/pull/2858#discussion_r354051716). So overall, with this PR there should be fewer issues. I also believe that the 2% gain you observed is for the same reason. The issue I described causes erroneous intersection types, which take longer to calculate. Example below:
```kt
interface Y {
    fun foo(): String
}
interface X {
    fun foo(): Int
}
interface Z {
    fun any(): Any
}
fun test_6(z: Z) {
    val a1 = z.any()
    val a2 = z.any()
    if (a1 is X && a2 is Y) {
        // The bug would cause a1 and 2 to be of type (X & Y) making the calls below ambiguous
        val foo1: String = a1.foo()
        val foo2: Int = a2.foo()
    }
}
```

I'll update the PR with tests to catch (2) and (3).

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -818,23 +836,17 @@ class FirDataFlowAnalyzer(private val components: FirAbstractBodyResolveTransfor
...
-        val fir = fir.unwrapWhenSubjectExpression()
-        if (fir is FirThisReceiverExpressionImpl) {
-            return variableStorage.getOrCreateNewThisRealVariable(fir.calleeReference.boundSymbol ?: return null)
+        val fir = fir.unwrapWhenSubjectExpression().unwrapSmartcast()
+        val symbol = fir.resolvedSymbol ?: return null
+        return when {
+            fir is FirThisReceiverExpressionImpl -> variableStorage.getOrCreateNewThisRealVariable(symbol)
+            symbol is FirVariableSymbol<*> && symbol.fir.run {
+                this is FirProperty && isLocal || this is FirValueParameter
+            } -> variableStorage.getOrCreateNewRealVariable(symbol).variableUnderAlias
```
punzki: @demiurg906 - I made this change after I noticed an issue with the following code (this is also in `testData/resolve/smartcasts/notBoundSmartcasts.kt`):
```kt
fun test_3(d1: D, d2: D) {
    val a = d1?.any
    val b = d2?.any
    a as A
    a.foo()
    b as B
    b.foo()
}
```

The issue was we would be storing a `RealDataFlowVariable` for the non-local property symbol, and therefore erroneously smartcasting local variables that access the same property, even for different objects.

This change then broke some tests in `testData/resolve/smartcasts/nullability.kt`. After analyzing those failures, I noticed some incorrectness in `getRealVariablesForSafeCallChain()` -- see the comment there. It doesn't seem like a simple fix. I verified against the current frontend as well. The current frontend would reject certain smartcasts that FIR frontend would accept -- see the changes in the `nullability.kt`.

Let me know if my understanding of this is correct!

demiurg906: It's not correct change, because we can smartcast final member properties:
```
class A {
    val x: Any = ""
}

fun test(a: A) {
    if (a.x is String) {
        a.x.length // OK in old frontend
    }
}
```

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -863,6 +865,11 @@ class FirDataFlowAnalyzer(private val components: FirAbstractBodyResolveTransfor
             return variableStorage[symbol]
         }
 
+    // TODO: Fix this -- see broken test_11 in compiler/fir/resolve/testData/resolve/smartcasts/nullability.kt
+    // We currently do not create RealDataFlowVariables for FirQualifiedAccess of symbols of non-local properties.
+    // If we have multiple local variables (could be value parameter) of the same type, there should be separate DataFlowVariables for
+    // accesses to a property for each distinct local variable, i.e., DataFlowVariables in storage for properties should be keyed by
+    // property symbol AND the "owning" local variable. We also need to check that the property is not mutable and has no custom getter.
```
demiurg906: Symbol is unique identifier for some declaration (function, class, property, value parameter, etc). They are needed because in current design fir may be changed (I mean that we can create new `FirFunction` instead of old one), but symbol will be saved

So, it's incorrect use some fir as identifier for declaration

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
demiurg906: Please, remove this change (but keep tests, they are good)

demiurg906: I planned to start rewrite `DataFlowAnalyzer` tomorrow, so you tests and exploration will be helpful, thanks

punzki: I understand that it's possible to smartcast final member properties, but the current implementation is broken for the case I described above (type info is associated with the symbol). This change _does_ break it even further though, which would explain the increase in errors you saw.

Okay, I will revert this part of the changes for now.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
punzki: I'm not proposing to use `FirProperty` or `FirFunction` as identifier (for example)... using the symbol is correct. However, I think we should store the entire "chain" of `FirVariableSymbols` instead of just the last one in the access. In my example:

```
fun foo(a1: A?, a2: A?) {
  a1!!.x!!.y!!
  a2.x.y  
}
```

We should store "not null type info" for the chains `a1`, `a1 -> x`, `a1 -> x -> y`. This way, type info for `a2`, `a2 -> x`, `a2 -> x -> y` are not affected. The current bug is such that the `x` property symbol gets the "not null type info" which affects both `a1` and `a2` in the above example.

punzki:
@demiurg906 Updated PR and removed the change regarding non-local properties. Also rebased to resolve conflicts.

demiurg906:
@punzki Thank you for your work. I've merged PR manually

PR #2861, opened 2019-12-05
Fix/workaround for KT-34781 

clydebarrow:
Navigating to declaration of parent class doesn't work for class declared in the library and cannot be decompiled. The effect of this bug is endless re-indexing in Idea and lack of error flagging, syntax highlighting etc all of which are dependent on completed indexing.

### DIFF
--- a/core/descriptors/src/org/jetbrains/kotlin/types/checker/TypeCheckingProcedure.java
+++ b/core/descriptors/src/org/jetbrains/kotlin/types/checker/TypeCheckingProcedure.java
@@ -102,14 +102,16 @@ public class TypeCheckingProcedure {
             if (typeProjection1.isStarProjection() && typeProjection2.isStarProjection()) {
                 continue;
             }
-            TypeParameterDescriptor typeParameter1 = constructor1.getParameters().get(i);
-            TypeParameterDescriptor typeParameter2 = constructor2.getParameters().get(i);
+            if(constructor1.getParameters().size() > i) {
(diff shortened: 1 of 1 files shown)


clydebarrow:
Also relates to KT-34890.

I have tested this with Idea CE 2019.2 and Ultimate 2019.3, and the symptoms of those two issues are resolved in both.

clydebarrow:
Can we get this merged? I am currently working with patched binaries, which is a nuisance when Idea updates come along as I have to skip those files.

zarechenskiy:
Hi,

Sorry for the long delay. I debugged this problem a bit and found that the actual problem is that for some reason class `com.controlj.view.ListPresenter<T>.Section` is considered as a not found one during type deserialization. This leads to an incorrect mock class construction and only then the exception is thrown in the type-checking procedure.

I believe this problem requires more investigation and I cannot accept your PR for now as this fix hides the actual problem, while code in the type-checking procedure is correct. At the same time, I understand that this problem is a blocker due to endless re-indexing, so we're going to find a workaround for this problem first

clydebarrow:
I understand that hiding the problem is not good - maybe the type-checking should detect and flag the problem, while not throwing a fatal exception.

clydebarrow:
Any progress on this? It's still costing me considerable time with the need to patch each plugin update with the fix - and that may break at any time of course. Without the patch the IDE is unusable.

zarechenskiy:
I'm really sorry this issue takes a lot of time to process, currently, we're busy with our major release (1.4).
So, this issue is targeted for our next release - 1.4

PR #2862, opened 2019-12-05
KT-33761 Add reduceOrNull

adellibovi:
Hello everyone,

This is supposed to solve https://youtrack.jetbrains.com/issue/KT-33761
It adds a reduceOrNull method which would return null if the collection or sequence is empty.

This is my first PR, please let me know if there is anything missing :)

### DIFF
--- a/libraries/stdlib/common/src/generated/_Arrays.kt
+++ b/libraries/stdlib/common/src/generated/_Arrays.kt
@@ -12696,6 +12696,141 @@ public inline fun CharArray.reduceIndexed(operation: (index: Int, acc: Char, Cha
 }
 
 /**
+ * Accumulates value starting with the first element and applying [operation] from left to right to current accumulator value and each element. Returns null if the array is empty.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public inline fun <S, T : S> Array<out T>.reduceOrNull(operation: (acc: S, T) -> S): S? {
+    if (isEmpty())
+        return null
+    var accumulator: S = this[0]
+    for (index in 1..lastIndex) {
+        accumulator = operation(accumulator, this[index])
+    }
+    return accumulator
+}
+
+/**
+ * Accumulates value starting with the first element and applying [operation] from left to right to current accumulator value and each element. Returns null if the array is empty.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public inline fun ByteArray.reduceOrNull(operation: (acc: Byte, Byte) -> Byte): Byte? {
+    if (isEmpty())
+        return null
+    var accumulator = this[0]
+    for (index in 1..lastIndex) {
+        accumulator = operation(accumulator, this[index])
+    }
+    return accumulator
+}
+
+/**
+ * Accumulates value starting with the first element and applying [operation] from left to right to current accumulator value and each element. Returns null if the array is empty.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public inline fun ShortArray.reduceOrNull(operation: (acc: Short, Short) -> Short): Short? {
+    if (isEmpty())
+        return null
+    var accumulator = this[0]
+    for (index in 1..lastIndex) {
+        accumulator = operation(accumulator, this[index])
+    }
+    return accumulator
+}
+
+/**
+ * Accumulates value starting with the first element and applying [operation] from left to right to current accumulator value and each element. Returns null if the array is empty.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public inline fun IntArray.reduceOrNull(operation: (acc: Int, Int) -> Int): Int? {
+    if (isEmpty())
+        return null
+    var accumulator = this[0]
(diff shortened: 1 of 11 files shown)


libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
```diff
@@ -840,6 +840,63 @@ object Aggregates : TemplateGroupBase() {
         }
     }
 
+    val f_reduceOrNull = fn("reduceOrNull(operation: (acc: T, T) -> T)") {
+        include(ArraysOfPrimitives, ArraysOfUnsigned, CharSequences)
+    } builder {
+        inline()
+        specialFor(ArraysOfUnsigned) { inlineOnly() }
+
+        doc { "Accumulates value starting with the first ${f.element} and applying [operation] from left to right to current accumulator value and each ${f.element}. Returns null if the first ${f.element} does not exist." }
```
ilya-g: Use multiline doc string here and describe return value with the `@return` kdoc tag, e.g.
```
@return the resulting accumulator value or `null` if ...
```

`"... if the first element does not exists"` sounds strange, as it might be that the first element does not exist, but the second does.  Better to say just `"if the ${f.collection} is empty."`

libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
```diff
@@ -840,6 +840,63 @@ object Aggregates : TemplateGroupBase() {
         }
     }
 
+    val f_reduceOrNull = fn("reduceOrNull(operation: (acc: T, T) -> T)") {
+        include(ArraysOfPrimitives, ArraysOfUnsigned, CharSequences)
+    } builder {
+        inline()
+        specialFor(ArraysOfUnsigned) { inlineOnly() }
```
ilya-g: We can introduce this function in 1.3.70 as experimental: then we need to annotate it with `@SinceKotlin("1.3")` and `@ExperimentalStdlibApi`. Otherwise we need to set SinceKotlin to 1.4.

libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
adellibovi: Good point, changed!

libraries/tools/kotlin-stdlib-gen/src/templates/Aggregates.kt
adellibovi: Having this in the 1.3.70 would be great yeah! Added both annotation :) and pushed the changes, thanks for the review @ilya-g !

adellibovi: One question, you meant both? f_reduceOrNull and f_reduceOrNullSuper right?

ilya-g: Yes, both of them

ilya-g:
Thanks for PR. It would be also nice to add here:
- tests for Sequences and Strings
- `reduceRightOrNull` for completeness and the corresponding tests
- samples are optional, but won't hurt.

adellibovi:
@ilya-g thanks for the feedback, I pushed new changes that cover the missing tests and the `reduceRightOrNull`. Also added a sample for both new methods. What do you think? :)

libraries/stdlib/common/src/generated/_Arrays.kt
```diff
@@ -12710,6 +12865,8 @@ public inline fun <S, T : S> Array<out T>.reduceRight(operation: (T, acc: S) ->
 
 /**
  * Accumulates value starting with last element and applying [operation] from right to left to each element and current accumulator value.
+ * 
+ * @sample samples.collections.Collections.Aggregates.reduceRightOrNull
```
ilya-g: This sample is for `reduceRightOrNull` and doesn't relate to `reduceRight`. Note that the samples for `reduceRight` have been already submitted in the PR #2867.

ilya-g:
I think we should provide `...OrNull` alternatives for `reduceIndexed` and `reduceRightIndexed` as well, but I propose to make it a separate follow-up PR.

libraries/stdlib/common/src/generated/_Arrays.kt
adellibovi: Oops, yes, you are right, misplaced! Updated the PR :)

adellibovi:
Sure, @ilya-g, I would love to take the`reduceIndexedOrNull` and `reduceRightIndexedOrNull` in a separate PR 👍

ilya-g:
Thanks for the PR, I've merged it manually: https://github.com/JetBrains/kotlin/compare/6d092b5f71416421c32fb361b56500495160851c~4...6d092b5f71416421c32fb361b56500495160851c

adellibovi:
Thanks for the review @ilya-g, one last question, do you know if this will this go into 1.3.70 finally or 1.4?

ilya-g:
Yes, these functions will be available in 1.3.70 as experimental.

PR #2863, opened 2019-12-05
KT-14119 Change String.toBoolean(): Boolean to String?.toBoolean()

adellibovi:
Hello,

This is suppose to apply the changes requested in https://youtrack.jetbrains.com/issue/KT-14119
From my understanding, this is an old topic which got recently decided to support the nullable receiver.

Please let me know if I missed something :)


ilya-g:
Thanks for the PR, but we decided to make it a slightly different way because of previous versions compatibility reasons. Anyway the fix for KT-14119 is already in master.

PR #2864, opened 2019-12-06
KT-35288 False positive "Remove braces from 'when' entry" in 'when' expression which returns lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35288


mglukhikh:
Merged 👍

PR #2865, opened 2019-12-06
KT-35320 False positive "Replace explicit parameter 'x' with 'it'" in 'when' expression which returns lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35320


vladimirdolzhenko:
LGTM.

This fix actually fixes incorrect intention application and it relates [KT-35320](https://youtrack.jetbrains.com/issue/KT-35320) while it does not fix the root cause of problem itself.

vladimirdolzhenko:
adjusted commit message and merged manually

Thanks

PR #2866, opened 2019-12-06
KT-35329 Replace 'when' with 'if' intention: do not suggest if 'when' is used as expression and it has no 'else' branch

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35329


vladimirdolzhenko:
LGTM

PR #2868, opened 2019-12-08
Add "kapt.provide.generated.option" property to control injection of generated option

ZacSweers:
With proper Filer support available now, there should be very little reason to use the kapt generated option. It's still inserted into options on every usage however. This breaks remote build caching due to the full path being inserted into the options however, so this adds a new "kapt.provide.generated.option" gradle property that can be set to "false" (true by default). This way projects that have no processors relying on this can also not break remote build caching.


abelkov:
If there is no existing issue in YouTrack, I think it's worth to create one: http://kotl.in/issue

ZacSweers:
Actually it looks like this option is explicitly avoiding being considered a task input via https://github.com/JetBrains/kotlin/commit/68deedb589c7328c1c7bab1e3a96562ec3f0db2d. So I'll close this for now!

PR #2869, opened 2019-12-08
KT-33109 "Add constructor parameters" quick fix should add default parameters from super class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33109


vladimirdolzhenko:
Thanks @t-kameyama for the PR, looks good - maybe one small note regarding `varargs`

```
open class A(vararg strings: String = arrayOf("a", "b")) {

}
```

so far quickfix will generate from `B: A` following code
```
class B(vararg strings: String) : A(*strings)
```

vladimirdolzhenko:
LGTM

PR #2870, opened 2019-12-09
KT-33212 False positive "map.put() should be converted to assignment" inspection when class inherited from MutableMap has "set" method

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33212


yanex:
Review OK.

yanex:
Merged to master manually (https://github.com/JetBrains/kotlin/commit/653e20dcbab14a5e07694df23b44cfca991879d6). Thank you for your contribution!

PR #2871, opened 2019-12-10
KT-35346 False positive 'Make internal' suggestion for function inside interface

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35346


klunnii:
Merged. Thank you!

PR #2872, opened 2019-12-10
KT-21799 Quickfix "Change function signature" for receiver type doesn't change it

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-21799


klunnii:
Thank you! Merged.

PR #2873, opened 2019-12-10
JVM_IR: do not do invokeinterface on Object methods

pyos:
This works on HotSpot, but might confuse other VMs.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceDelegationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceDelegationLowering.kt
@@ -127,10 +127,10 @@ private class InterfaceDelegationLowering(val context: JvmBackendContext) : IrEl
             super.visitSimpleFunction(declaration)
         }
     }
-
-    private fun IrSimpleFunction.hasInterfaceParent() =
-        (parent as? IrClass)?.isInterface == true
 }
+
+private fun IrSimpleFunction.hasInterfaceParent() =
+    (parent as? IrClass)?.isInterface == true
 
 internal val interfaceSuperCallsPhase = makeIrFilePhase(
     lowering = ::InterfaceSuperCallsLowering,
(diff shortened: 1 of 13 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceDelegationLowering.kt
```diff
@@ -201,3 +201,25 @@ private fun IrSimpleFunction.isDefinitelyNotDefaultImplsMethod() =
...
+    lowering = ::InterfaceObjectCallsLowering,
+    name = "InterfaceObjectCalls",
+    description = "Resolve calls to Object methods on interface types to virtual methods"
+)
+
+private class InterfaceObjectCallsLowering(val context: JvmBackendContext) : IrElementTransformerVoid(), FileLoweringPass {
+    override fun lower(irFile: IrFile) = irFile.transformChildrenVoid(this)
+
+    override fun visitCall(expression: IrCall): IrExpression {
+        if (expression.superQualifierSymbol != null)
```
max-kammerer: Seems this cases also should be supported:
```
interface Base {

}

interface Derived: Base {
    fun foo() {
        super<Base>.hashCode()
    }
}
```
Actually compilation is failed in `InterfaceSuperCallsLowering`, but it could be simply fixed by smth like:
```
override fun visitCall(expression: IrCall): IrExpression {
        if (expression.superQualifierSymbol?.owner?.isInterface != true || expression.isSuperToAny()) {
            return super.visitCall(expression)
        }
```

pyos: I've added code to support the case where `Derived` is a class, but I'm pretty sure it's not possible to make your example work. The problem is that to call `Object.hashCode` precisely `foo` needs an INVOKESPECIAL, but `foo` is moved to `Derived$DefaultImpls` which cannot do INVOKESPECIAL on instances of `Derived`, and it's not possible to put an accessor into `Derived` because it's an interface. If you actually try to load `Derived` from your example after compiling it with the non-IR backend, you'll get a validation error.

max-kammerer: Yes, really such cases seems should be prohibited on both targets (except calls within @JvmDefault's methods)

max-kammerer: https://youtrack.jetbrains.com/issue/KT-35417

max-kammerer:
Thank you for pr!

PR #2874, opened 2019-12-10
Refactoring GradleMultiplatformWizardTests

useriq:
1) made tests more detalized
 2) fixed current method of tests running in tests
 Now test class runs atomized.
 If test not ran or test not found - throws exception
 3) changed type of running tests from methods to gradle tasks
 4) added FileChecker to checking exists of files
 5) added checking generated files in tests, e.g.
   checkSource("src") {
                    sourceSetsSize(6)
                    common("$commonMain/$kotlin/$sample/Sample.kt")
                    test("$commonTest/$kotlin/$sample/SampleTests.kt")
				  }
    Checks that:
        -"src" has 6 folders
		- $commonMain/$kotlin/$sample/Sample.kt is exists and contains "expect" in content    
		- $commonTest/$kotlin/$sample/SampleTests.kt is exists and contains "@Test"
 6) added AssertK library to detalize asserts
 7) switched on "testWeb" and added checks generated files 
 8) added checks of test-result (check count running tests and successful done tests)

### DIFF
--- a/idea/idea-gradle/build.gradle.kts
+++ b/idea/idea-gradle/build.gradle.kts
@@ -35,6 +35,8 @@ dependencies {
     testCompile(projectTests(":idea"))
     testCompile(projectTests(":idea:idea-test-framework"))
 
+    testCompile("com.willowtreeapps.assertk:assertk-jvm:0.20")
+
     testCompile(intellijPluginDep("gradle"))
     Platform[193].orHigher {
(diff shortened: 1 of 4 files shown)


dsavvinov:
This is a test task PR from interview, so there's no need to specifically review and process it. I'll handle it further. Sorry for bothering everyone!

PR #2875, opened 2019-12-10
KT-35165 "Replace 'if' with elvis operator": don't suggest if val initializer is a complex expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35165


yanex:
Merged to master manually (https://github.com/JetBrains/kotlin/commit/86ac44c23e22c76a35e37ff2b41fb70188811564). Thank you!

PR #2876, opened 2019-12-12
KT-34959 False positive "Redundant overriding method" with different implemented/overridden signatures

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34959


klunnii:
Merged, thanks!

PR #2877, opened 2019-12-12
KT-34690 Support intention `Convert lambda to reference` for qualified references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34690


vladimirdolzhenko:
Thanks a lot @t-kameyama for your PR. 

LGTM

PR #2879, opened 2019-12-14
KT-31749 "Surround with null check" produces incorrect check for 'in' expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31749


darthorimar:
Thank you @t-kameyama

PR #2880, opened 2019-12-14
KT-33096 Turn 'MapGetWithNotNullAssertionOperator' into an intention

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33096


klunnii:
Merged. Thank you!

PR #2881, opened 2019-12-14
KT-34026 Add "Remove argument" quick fix for redundant argument in constructor call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34026
https://youtrack.jetbrains.com/issue/KT-34332


vladimirdolzhenko:
Minor thing: could you please cover negative scenario, i.e. `Remove argument` quick fix does not appear if all arguments are matched

Thanks

vladimirdolzhenko:
LGTM. Some other tests fail due to missed `ACTION: Remove argument` in them, have fixed those tests and merged manually.

Thanks for your contribution

knutwannheden:
As commented on https://youtrack.jetbrains.com/issue/KT-34332 I can't get this to work in my IDEA Ultimate 2020.1. Is there maybe a regression?

I tested with the example from the issue:
```kt
fun foo(s: String) {
}
fun bar() {
    foo("", 1) // Add remove argument quick fix
}
```

vladimirdolzhenko:
Comment is [resolved](https://youtrack.jetbrains.com/issue/KT-34332#focus=streamItem-27-4100590.0-0) in YouTrack

PR #2883, opened 2019-12-15
KT-22665 "Create object" quick fix produce wrong code for enum

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22665


klunnii:
1. `Create object` quick fix is no longer available. Please, consider the context:
``` kotlin
enum class SampleEnum {}

fun usage(sample: SampleEnum) {
    SampleEnum.RED // suggesting object creation is still correct here
}
```
2. Provided test unfortunatly fails. Would you please rebase your commit on recent master and check it?

t-kameyama:
Fixed in cd961a1

t-kameyama:
Fixed in 86c7c58

klunnii:
Merged. Thank you!

PR #2885, opened 2019-12-15
Remove Redundant Project

stephanenicolas:
Just some syntax cleanup when visting the file.
Note: I was trying to understand why the output folders of kapt are creating so eagerly in a build, during configuration time indeed...(they should be created later, during execution).


yanex:
Thanks for your contribution! Merged to master manually (https://github.com/JetBrains/kotlin/commit/7b227f31138ff3a5f1ed7508b511d83121408afa). I will cherry-pick the change into the 1.3.70 branch after the EAP 1.

PR #2886, opened 2019-12-16
KT-29264 Add Quick Fix for 'JAVA_CLASS_ON_COMPANION' compiler warning

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29264


yanex:
Please note that your quick fix changes semantics.
I think we should suggest changing it to either `T.Companion::class.java` and to `T::class.java`, and the first option should have higher priority.

idea/src/org/jetbrains/kotlin/idea/quickfix/ReplaceJavaClassWithClassJavaFix.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.jetbrains.kotlin.psi.createExpressionByPattern
+
+class ReplaceJavaClassWithClassJavaFix(expression: KtDotQualifiedExpression) : KotlinQuickFixAction<KtDotQualifiedExpression>(expression) {
+    override fun getText(): String = KotlinBundle.message("replace.with.0", "::class.java")
+
+    override fun getFamilyName(): String = text
+
+    override fun invoke(project: Project, editor: Editor?, file: KtFile) {
+        val expression = element ?: return
+        element?.replace(KtPsiFactory(expression).createExpressionByPattern("$0::class.java", expression.receiverExpression))
```
yanex, line 24: IMO it makes sense to use `expression` here as you've already got the not-null value.

t-kameyama:
Added to quick fix to replace with `T.Companion::class.java` in 1fdc164, PTAL

yanex:
Note that companions can be named:

```
class Foo {
    companion object Bar
}

fun test() {
    Foo.javaClass<caret>
}
```

t-kameyama:
> Note that companions can be named:

Fixed in 09d5c27, PTAL.

yanex:
Review: OK, I applied an additional patch:

```
Index: idea/src/org/jetbrains/kotlin/idea/quickfix/JavaClassOnCompanionFixes.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- idea/src/org/jetbrains/kotlin/idea/quickfix/JavaClassOnCompanionFixes.kt	(revision add75704170a4f40846ac1dd9979f8db3e069e98)
+++ idea/src/org/jetbrains/kotlin/idea/quickfix/JavaClassOnCompanionFixes.kt	(revision ecf02de00d6c16f821aa24178e48cba94d708f7d)
@@ -12,13 +12,16 @@
 import org.jetbrains.kotlin.diagnostics.Diagnostic
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.references.mainReference
+import org.jetbrains.kotlin.name.SpecialNames
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 object JavaClassOnCompanionFixes : KotlinIntentionActionsFactory() {
     override fun doCreateActions(diagnostic: Diagnostic): List<IntentionAction> {
         val expression = diagnostic.psiElement.parent as? KtDotQualifiedExpression ?: return emptyList()
-        val companionName = expression.receiverExpression.mainReference?.resolve()?.safeAs<KtObjectDeclaration>()?.name ?: "Companion"
+        val companionName = expression.receiverExpression.mainReference?.resolve()?.safeAs<KtObjectDeclaration>()?.name
+            ?: SpecialNames.DEFAULT_NAME_FOR_COMPANION_OBJECT.identifier
+
         return listOf(
             ReplaceWithCompanionClassJavaFix(expression, companionName),
             ReplaceWithClassJavaFix(expression)
```

t-kameyama:
New PR: https://github.com/JetBrains/intellij-community/pull/1604

PR #2888, opened 2019-12-16
JVM_IR: Backwards compatible handling of default tailrec params.

madsager:
Before 1.4 tailrec function default arguments were evaluated
right-to-left instead of left-to-right. This is controlled
by a compile-time flag. This change adds support for the
right-to-left evaluation order when that flag is not set.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
@@ -16,10 +16,15 @@
 
 package org.jetbrains.kotlin.backend.common.lower
 
-import org.jetbrains.kotlin.backend.common.*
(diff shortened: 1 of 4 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/TailrecLowering.kt
```diff
@@ -133,28 +136,31 @@ private class BodyTransformer(
...
-                            variable.symbol, expression.origin
-                        )
-                    }
-                }, data = null)
-
-            +irSetVar(parameterToVariable[parameter]!!.symbol, defaultValue)
-        }
+        parameters
+            .filter { it !in specifiedParameters }
+            .let { if (properComputationOrderOfTailrecDefaultParameters) it else it.asReversed() }
```
madsager: The addition of this extra let is the only change here, the rest is just formatting.

PR #2889, opened 2019-12-16
PSI2IR: ignore enum annotation arguments of error type

pyos:
Similar to references to error type, this may happen if library A uses an entity from library B annotated with an annotation from C, but A is compiled without C on the classpath. (Note the comment about interfaces, though.)

Found while compiling [MvRx](https://github.com/airbnb/MvRx) (A = tests, B = library itself, C = androidx.annotations).


dnpetrov:

dnpetrov:
Pushed manually.

PR #2890, opened 2019-12-16
JVM IR: Fix names of fake local variables for inline arguments

sfs:
The debugger [looks for marker variables](https://github.com/JetBrains/kotlin/blob/0b7c11c96e844488dff632763cd6d64377274b99/idea/jvm-debugger/jvm-debugger-core/src/org/jetbrains/kotlin/idea/debugger/debuggerUtil.kt#L60) of the form `$i$a$-functionName-lambdaClassName` to determine whether we are inside of an inline argument, where `lambdaClassName` is the name which would have been generated for the class implementing the argument lambda.

This PR implements the same naming convention in the IR backend (which is simple, since the work is already done in `InventNamesForLocalClasses`).

Note that the PR doesn't fix any new tests, it just makes some existing tests more strict.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
@@ -9,7 +9,6 @@ import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
-import org.jetbrains.kotlin.backend.jvm.codegen.mapClass
 import org.jetbrains.kotlin.backend.jvm.ir.IrInlineReferenceLocator
 import org.jetbrains.kotlin.ir.builders.createTmpVariable
 import org.jetbrains.kotlin.ir.builders.irBlockBody
@@ -22,7 +21,7 @@ import org.jetbrains.kotlin.ir.declarations.IrFunction
 import org.jetbrains.kotlin.ir.expressions.IrBlockBody
 import org.jetbrains.kotlin.ir.expressions.IrCallableReference
 import org.jetbrains.kotlin.ir.expressions.IrExpressionBody
-import org.jetbrains.kotlin.ir.util.parentAsClass
+import org.jetbrains.kotlin.ir.expressions.IrFunctionReference
 import org.jetbrains.kotlin.ir.visitors.acceptChildrenVoid
 import org.jetbrains.kotlin.ir.visitors.acceptVoid
 import org.jetbrains.kotlin.load.java.JvmAbi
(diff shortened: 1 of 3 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
```diff
@@ -49,10 +48,14 @@ internal class FakeInliningLocalVariablesLowering(val context: JvmBackendContext
...
     }
 
-    override fun handleInlineFunctionLambdaParam(lambda: IrFunction, callee: IrFunction, callSite: IrDeclaration?) {
+    override fun handleInlineFunctionLambdaParam(lambdaReference: IrFunctionReference, callee: IrFunction, callSite: IrDeclaration?) {
         // Do not record lambda parameters. Instead deal with them now.
+        val lambda = lambdaReference.symbol.owner
         if (lambda.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA) {
-            lambda.addFakeInliningLocalVariablesForArguments(callee)
+            val argumentToFunctionName = context.methodSignatureMapper.mapFunctionName(callee)
+            val lambdaReferenceName = context.getLocalClassType(lambdaReference)!!.internalName.substringAfterLast("/")
```
madsager: Does this mean that the names used for the fake local variables are different from the names in the actual class files? Or was that the case before? Should there be a corresponding change in the way we name the actual elements in the output? I would find it very confusing if the names referred to in the fake local variables were different from what is in the class file. I can't figure out if this change is fixing that or making it a problem?

sfs: The names for local classes are also of the form `enclosingName$n`, where the enclosing name is of the form `enclosingType$enclosingFunction` for lambdas defined inside of a function. The main difference to the previous behavior is the `$n` suffix, which distinguishes the lambda from all other anonymous classes in the current file.

madsager: Makes sense, thanks Steven! So the point here is that all the references to named elements that are in the output class files have the right names. Only the numbering of the lambdas which are no in the actual output is changed here. And for full context these names seem to be used by the step-in handler in the Intellij debugger.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FakeInliningLocalVariablesLowering.kt
```diff
@@ -49,10 +48,14 @@ internal class FakeInliningLocalVariablesLowering(val context: JvmBackendContext
         // Do not record inline function callable reference parameters. They will not be used.
     }
 
-    override fun handleInlineFunctionLambdaParam(lambda: IrFunction, callee: IrFunction, callSite: IrDeclaration?) {
+    override fun handleInlineFunctionLambdaParam(lambdaReference: IrFunctionReference, callee: IrFunction, callSite: IrDeclaration?) {
         // Do not record lambda parameters. Instead deal with them now.
+        val lambda = lambdaReference.symbol.owner
         if (lambda.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA) {
-            lambda.addFakeInliningLocalVariablesForArguments(callee)
+            val argumentToFunctionName = context.methodSignatureMapper.mapFunctionName(callee)
```
udalov: Unrelated to your changes, but it's a bit sad that we have dependencies of IR lowerings on `MethodSignatureMapper`, which I initially hoped to be codegen-only logic (and maybe even enforcing it by separating lowerings and codegen into different modules). The reason why I wanted that was to avoid hacks in `MethodSignatureMapper` to support the state of IR at different stages between lowerings. But looks like there's not a lot of hacks anyway there, so never mind. :)

udalov:
Thank you!

PR #2893, opened 2019-12-18
Embed proguard/R8 rules in kotlin-reflect artifact jar

ZacSweers:
The Android build pipeline can extract embedded proguard configurations from dependencies and merge them automatically. This adds a conservative proguard configuration to the kotlin-reflect JVM artifact in support of that. This focuses mostly on just retaining what's necessary for kotlin-reflect's own functionality to operate, but could be expanded if community feedback discovers other good candidate rules.

With this in place - most Android projects using R8 or Proguard should Just Work™️ with kotlin-reflect.


ZacSweers:
Following conversation from https://kotlinlang.slack.com/archives/C0AVAB92L/p1576528186001900?thread_ts=1516315390.000641&cid=C0AVAB92L

CC @udalov

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
```diff
@@ -0,0 +1,14 @@
+##---------------Begin: Kotlin-reflect ----------
+# Keep Metadata annotations so they can be parsed at runtime.
+-keep class kotlin.Metadata { *; }
```
jsjeon: I wonder this is not actually needed. Recently, I started working on synthesizing @Metadata from scratch (for R8-processed Kotlin library), and kept adding more tests. During that, I found that, even without this rule, R8 will still keep metadata (annotation) associated to classes.

More conversion/details are here: https://r8-review.googlesource.com/c/r8/+/46240/1/src/test/java/com/android/tools/r8/kotlin/metadata/MetadataRenameInParametertypeTest.java#70

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
```diff
@@ -0,0 +1,14 @@
...
+
+# Keep kotlin-reflect internals.
+-keep class kotlin.reflect.jvm.** { *; }
+
+# Keep generic signatures and annotations at runtime.
+# R8 requires InnerClasses and EnclosingMethod if you keepattributes Signature.
+-keepattributes InnerClasses,Signature,*Annotation*,EnclosingMethod
```
jsjeon: Do we need broad *Annotation* or just RuntimeVisibleAnnotation (because metadata is runtime visible) ?

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
```diff
@@ -0,0 +1,14 @@
+##---------------Begin: Kotlin-reflect ----------
+# Keep Metadata annotations so they can be parsed at runtime.
+-keep class kotlin.Metadata { *; }
+
+# Keep kotlin-reflect internals.
+-keep class kotlin.reflect.jvm.** { *; }
```
jsjeon: This seems way too broad. At least, we can try _public_ APIs only.

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
jsjeon: Ah, one caveat is, though, R8 with this fix (https://r8-review.googlesource.com/c/r8/+/44220) will keep metadata annotation without definition. For prior versions of R8, it may be the case that we still need this rule.

ZacSweers: Yeah I think it's best to keep this as a conservative rule. We've had [reports in Moshi ](https://github.com/square/moshi/pull/1038) that Metadata is stripped in android builds without a custom keep rule for some users, though unsure what version they're using or if they're just using proguard.

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: I didn't know that `RuntimeVisibleAnnotations` was a thing! Historically any project reading annotations at runtime has always used `*Annotation*`, but maybe that's just a bespoke pattern that isn't needed. Let me try in our codebase

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: Let me try it in our project. To be totally honest I'm not convinced this is truly needed, but it's frequently suggested as a fix in the community. The last comprehensive list we were using internally was based on this:

https://github.com/square/moshi/pull/883

Though all of those are public classes as well, so it would work. @udalov any advice on what scope should be kept?

udalov: The list from the issue mentioned above is actually the combination of all services and implementations used in kotlin-reflect -- see `META-INF/services` in `kotlin-reflect.jar`. Shouldn't they be handled correctly in some other way by Proguard/R8? E.g. is it possible to "keep" the directory `META-INF/services`, and everything it references, to avoid enumerating classes used there?

Other than that, we do need to fully keep the contents of packages `kotlin.reflect.full` and `kotlin.reflect.jvm`, but not their subpackages (i.e. `kotlin.reflect.jvm.internal`).

JakeWharton: Types declared inside `META-INF/services/` are handled automatically provided that the code which calls into `ServiceLoader` with that service type is not being removed. You shouldn't need to do anything for them.

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
udalov: There are also `RuntimeVisibleParameterAnnotations` and `RuntimeVisibleTypeAnnotations`, so maybe the pattern should be `RuntimeVisible*Annotations`, if that works...

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
```diff
@@ -0,0 +1,14 @@
+##---------------Begin: Kotlin-reflect ----------
```
udalov: Are these Begin/End tags required?

Also, please consider renaming the file itself to `kotlin-reflect.pro`

JakeWharton: > Are these Begin/End tags required?

Not needed. They will be added automatically by R8 now for blaming purposes.

https://issuetracker.google.com/issues/134372926

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: They're handled automatically by R8, but not proguard iirc. Can we have rules that are only picked up by one or the other?

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: By R8 yes. Do we care about it in a proguard context? They're not strictly required but the blaming purpose is nice

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
JakeWharton: Yes. See kotlinx.coroutines for an example that targets R8 1.6 and newer only.

ZacSweers: Done in 08616fb

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: Removed in c6f922b

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: Also tweaked to just keep service-loaded APIs, only in proguard or R8 <1.6 c15aac4

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: That seems to work b843558

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
ZacSweers: @jsjeon what version of R8 would that fix be in? I'm not sure how to check the latest release history or what's in them off-hand there

core/reflection.jvm/resources/META-INF/com.android.tools/proguard/kotlin-reflect.pro
```diff
@@ -0,0 +1,23 @@
+# When editing this file, update the following files as well:
+# - META-INF/com.android.tools/r8-from-1.6.0/kotlin-reflect.pro
+# - META-INF/com.android.tools/r8-upto-1.6.0/kotlin-reflect.pro
+# - META-INF/proguard/kotlin-reflect.pro
+# Keep Metadata annotations so they can be parsed at runtime.
+-keep class kotlin.Metadata { *; }
+
+# Keep implementations of service loaded interfaces
+# R8 will automatically handle these these in 1.6+
```
jsjeon: Nit: this file is not quite related to `R8`. :) (and below too)

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
jsjeon: Thanks for taking care of this thoroughly! Those specific rules that aim for only interface and services look great!

core/reflection.jvm/resources/META-INF/proguard/kotlinreflect.pro
jsjeon: Ah, it's even difficult for me to check which version includes it. :\ Based on the timeline, I guess AGP 3.6 will include that.

However, that fix I mentioned was about keeping annotations associated with classes/members _if_ the definition of annotations is missing. Therefore, if annotation class `Metadata` is given (via kotlin reflect artifact jar of course), i.e., not missing, it is still the case that R8 will take it away if it thinks it's not used.

------

I was about to mention the case like https://issuetracker.google.com/issues/146534384 where `kotlin.Metadata` itself is retrieved at runtime. In that case, yeah, this rule is necessary.

ZacSweers: Sounds like it's safest to just keep this for now

core/reflection.jvm/resources/META-INF/com.android.tools/proguard/kotlin-reflect.pro
ZacSweers: D'oh, fixed in 5091302

ZacSweers:
@udalov anything else needed for this PR?

core/reflection.jvm/resources/META-INF/com.android.tools/proguard/kotlin-reflect.pro
```diff
@@ -0,0 +1,22 @@
...
+-keep class kotlin.Metadata { *; }
+
+# Keep implementations of service loaded interfaces
+-keep interface kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoader
+-keep class * implements kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoader { public protected *; }
+-keep interface kotlin.reflect.jvm.internal.impl.resolve.ExternalOverridabilityCondition
+-keep class * implements kotlin.reflect.jvm.internal.impl.resolve.ExternalOverridabilityCondition { public protected *; }
+
+# Keep generic signatures and annotations at runtime.
+# R8 requires InnerClasses and EnclosingMethod if you keepattributes Signature.
```
udalov: Minor: this file is for Proguard specifically, but you mention R8 here

ZacSweers: Ah good catch, fixed a7a9f2b

udalov: Also in `META-INF/proguard/kotlin-reflect.pro`

ZacSweers: That one is intended to be a generic fallback with configuration that could work with both at a minimal level, whereas these nested ones are more intended for tighter scopes where available

udalov:
@ZacSweers Thank you for the contribution!

Merged manually in 482874fdc1f. The changes will be available in 1.4-M1.

ZacSweers:
Thanks!

ColtonIdle:
@udalov any chance something changed here in 1.4.32?

We noticed this issue happening but it wasn't happening in the past (we've been using proguard since kotlin 1.4) but with 1.4.32 we need to manually add

```
-keep class kotlin.reflect.jvm.internal.** { *; }
```

for our app not to crash.

ZacSweers:
File an issue with repro steps, commenting on an old PR isn't really the best venue

PR #2897, opened 2019-12-18
KT-35525 False positive intention for 'run': "Convert to 'let'" when invoked without receiver

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35525


darthorimar:
@t-kameyama thanks :)

PR #2898, opened 2019-12-18
KT-35526 Intention "Eliminate argument of 'when'" is broken for 'when' expression without 'else' branch

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35526


darthorimar:
Thank you!

PR #2899, opened 2019-12-18
[JVM IR] ForLoopsLowering: Keep IMPLICIT_NOTNULL type-casts from `next()` and `componentN()` calls

punzki:
There were issues when we have iterables from Java where the element type has "not null" type information (i.e., `@EnhancedNullability`).

We discovered this issue when compiling OkHttp with JVM_IR, specifically here is where it breaks: https://github.com/square/okhttp/blob/e1d9cd2d905b6209253a868429eb4b9b49ee7a99/okhttp/src/main/java/okhttp3/internal/cache/DiskLruCache.kt#L637

@dnpetrov I could use a look here since you recently did some work around `@EnhancedNullability` in JVM_IR. If you search for `// IGNORE_BACKEND: JVM` you will see some behavioral differences between JVM IR and non-IR. (For example, see`compiler/testData/codegen/box/ranges/javaInterop/javaCollectionOfExplicitNotNullFailFast.kt`.) The not-null assertion is not generated when assigning to the loop variable (could be in a destructuring declaration or not). The root cause seems to be that the loop variable is a `KtParameter` and `CodegenAnnotatingVisitor/RuntimeAssertionsOnDeclarationBodyChecker` do not analyze the need for not-null assertions on KtParameters (or `KtDestructuringDeclaration/Entry`).

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ForLoopsLowering.kt
@@ -10,6 +10,7 @@ import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
+import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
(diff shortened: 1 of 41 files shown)


dnpetrov:

punzki:
@dnpetrov What are your thoughts on the behavioral difference I noted between JVM and JVM_IR in the previous comment? https://github.com/JetBrains/kotlin/pull/2899#issue-354827250

Sounds like a bug in JVM (non-IR)?

dnpetrov:
@punzki Yes, I expect to see a bunch of such discrepancies since we start to generate nullability assertions more systematically. We should record such cases as we find them, and get them through language design committee before 1.5.

dnpetrov:
Pushed manually.
Recorded corner case as https://youtrack.jetbrains.com/issue/KT-35698.

dnpetrov:
Attach files to your comment
Attach files by dragging & dropping, selecting or pasting them.

PR #2900, opened 2019-12-19
PSI2IR: generate field writes for all val property assignments

pyos:
Assuming the frontend is correct, and it hopefully is, all of them are initializations even if not done in the owner class directly.


compiler/testData/ir/irText/classes/initVal.kt
```diff
@@ -12,3 +12,11 @@ class TestInitValInInitBlock {
     }
 }
 
+class TestInitValInLambdaCalledOnce {
+    val x: Int
+    init {
+        1.run {
```
pyos: `init { 1.run { } }` causes fir to fail with an error at `ConeClassifierLookupTag.toSymbol` because the tag is of unexpected type `ConeTypeVariableTypeConstructor`. Is there a way to mute this test for fir?

dnpetrov: You can implement a corresponding directive in `AbstractFir2IrTextTest` (something like `FIR_IGNORE`). 
@mglukhikh is this a known issue in FIR? Could you please investigate it, or forward it to someone else?

mglukhikh: Let me look at it...

dnpetrov:
Everything else (except for FIR2IR issue mentioned above) looks ok to me.
I'd rather implement `FIR_IGNORE` directive in `AbstractFir2IrTextTest` (NB: test should fail if it actually passes; probably you can reuse code in `KotlinTestUtils#runTestImpl`, or use it for inspiration). However, that might be an overkill for this single case, so if the issue can be fixed in FIR2IR rather quickly (CC @mglukhikh ), we can wait for FIR2IR fix and then merge.

compiler/testData/ir/irText/classes/initVal.kt
mglukhikh: We investigated problem with @demiurg906. It's more or less clear what's happened, but it's non-trivial to fix it. I'd prefer to accept PR and mute test on teamcity, I'll fix it in a few days. If you do not like this way, we can return `null` from this `toSymbol` function temporarily, it will hide the problem.

Also it would be great to create separate IR test for this case (if possible) and not to add this case to existing test.

mglukhikh: cc @dnpetrov ^^^

compiler/testData/ir/irText/classes/initValInLambda.kt
```diff
@@ -0,0 +1,9 @@
+// IGNORE_BACKEND: ANY_FIR
```
mglukhikh: Great, thank you!

dnpetrov: Yes, that's ok.

dnpetrov:
Ok, since this can affect test runs in Native, I'll do the remote run, and then push manually if everything is fine.

compiler/testData/ir/irText/classes/initValInLambda.kt
punzki: Would you consider using the same custom ignore directive which we used for the FIR black box tests ([change](https://github.com/JetBrains/kotlin/pull/2800/commits/84d77d126c8c5c2bea0aa62c795a797f223786d4#diff-3e9a3103a63a111fbcef46ce4a8033eeR488)), `// IGNORE_BACKEND_FIR`? That is, you would put `// IGNORE_BACKEND_FIR: ANY` here. For that change, I drew inspiration from the custom ignore directive for multi-module.

One advantage would be it'd be slightly easier to find disabled tests for FIR using the same directive. Another is that you don't need to add a new backend in the enum which is not really a backend. 🙂

dnpetrov:
NB last change causes non-trivial merge due to changes in test generation. Please, rebase on `master` and re-generate tests.

pyos:
Yeah, I've amended a commit to implement @punzki's suggestion while you merged the old version, hence the conflict.

In any case, @mglukhikh already fixed the bug, so https://github.com/JetBrains/kotlin/commit/2adcb5dec474d532fbd6387bad4b6a8df7116499 can in fact be reverted completely.

PR #2901, opened 2019-12-19
JVM IR: Fix line numbers in callable reference classes

sfs:
In particular, there should be no line numbers in the constructor of a callable reference, since otherwise the debugger (smart) steps into the constructor, which is pretty jarring.


max-kammerer:
Thank you!

PR #2904, opened 2019-12-19
JVM_IR: Fix check for whether classes are from Java.

madsager:
This allows us to not generate redundant immutable collection
stubs. The code to generate the immutable collection stubs does
not deal well with thinking that all external declarations
come from Java.


pyos:
This also fixes the following code when compiled as separate modules:

```kotlin
// FILE: a.kt
open class A {
    @JvmName("g")
    fun <T> f(x: T, y: Int = 1) = x
}
```
```kotlin
// FILE: b.kt
class B : A()

fun box() = B().f("OK")
```

max-kammerer:
@madsager Could you also add test mentioned by @pyos  ?
To use module directive see for example `inlineWithJava.kt` test

madsager:
@max-kammerer absolutely! I added it to the existing CompilerKotlinAgainstKotlin test for JvmNames.

max-kammerer:
Thank you for PR!

PR #2905, opened 2019-12-20
KT-35216 <T : AutoCloseable?, R> T.use and <T : Closeable?, R> T.use should have contracts

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35216


ilya-g:
Perhaps it would be easier to test this aspect of `use` function as a plain stdlib test, rather than a compiler diagnostics test, provided that it doesn't render any diagnostics we want to assert.

ilya-g:
Merged as https://github.com/JetBrains/kotlin/commit/941de655c4686cae53136f4b95794f09d80230f6. Thank you for PR.

PR #2906, opened 2019-12-20
KT-35558 "Convert reference to lambda" intention generates wrong code for a High Order Function which expects an extension lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35558


igoriakovlev:
Thank you for you work!

PR #2907, opened 2019-12-20
Minor: IGNORE_BACKEND: ANY_FIR -> IGNORE_BACKEND_FIR: ANY

pyos:
for consistency with blackbox tests; see https://github.com/JetBrains/kotlin/pull/2900#discussion_r360024713

Also, unmute a test that was fixed in https://github.com/JetBrains/kotlin/commit/ada7f3da6ca9b9c5fc6b2cbb2ebf23db7d3aef22 while the linked PR was being reviewed.


dnpetrov:
Pushed manually. Should be in sync now.

PR #2909, opened 2019-12-20
 JVM_IR: do not copy defaults in functions with inline class parameters

pyos:
cc @sfs

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
@@ -151,12 +151,15 @@ class MemoizedInlineClassReplacements {
                 val name = if (parameter == function.extensionReceiverParameter) Name.identifier("\$receiver") else parameter.name
                 val newParameter: IrValueParameter
                 if (parameter == function.dispatchReceiverParameter) {
-                    newParameter = parameter.copyTo(this, index = -1, name = name)
+                    newParameter = parameter.copyTo(this, index = -1, name = name, defaultValue = null)
                     dispatchReceiverParameter = newParameter
                 } else {
-                    newParameter = parameter.copyTo(this, index = index - 1, name = name)
+                    newParameter = parameter.copyTo(this, index = index - 1, name = name, defaultValue = null)
                     valueParameters.add(newParameter)
                 }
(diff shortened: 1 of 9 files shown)


sfs:
Thanks @pyos! This is yet another bug due to unlowered IR constructs in different files in the same module. Specifically, we were copying expressions for default values of function parameters which were not lowered yet.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -89,6 +90,12 @@ private class JvmInlineClassLowering(private val context: JvmBackendContext) : F
         return declaration
     }
 
+    private fun copyDefaultArguments(replacement: IrReplacementFunction) {
+        for ((oldParameter, newParameter) in replacement.valueParameterMap) {
+            newParameter.defaultValue = oldParameter.owner.defaultValue?.deepCopyWithSymbols(replacement.function)
```
udalov: Since `deepCopyWithSymbols` doesn't copy attributes, wouldn't it cause problems if an anonymous class is used in such a default argument, since we wouldn't have a name for it in the codegen?
E.g.
```
inline class A(val i: Int)

fun A.f(x: String = object { val v = "OK" }.v): String = x
```

pyos: It does copy *attributes*, so there are no problems with names. You're probably thinking of metadata, and yeah, the object is missing some stuff compared to one in a similar `fun Any?.f`. Though note that `parameter.copyTo` calls `deepCopyWithSymbols` on the default value anyway, so that's not the fault of this PR specifically.

pyos: Looks like the copy is redundant, though: `transformSimpleFunctionFlat` only returns two functions if the function is an override, but then it cannot have default arguments.

pyos: Removed the copy and added a test.

udalov: Oh you're right. But indeed, the copy is not necessary. Thanks!

gogabr:
Thank you!

PR #2911, opened 2019-12-20
Field in KonanTarget to show if it has support of optimized allocator

LepilkinaElena:


sbogolepov:
Why it should property instead of extension?

PR #2912, opened 2019-12-20
Improve caching for failed class lookups

gharrma:
This slightly mitigates KT-35614, although it does
not fix the issue entirely.

Note: this speeds up findClass(), but not findClasses().


dzharkov:
Sorry for the late reply, but I'm not sure about this change because this cache should not be used frequently when the requested class doesn't exist (as @udalov has mentioned in the [comment](https://youtrack.jetbrains.com/issue/KT-35614#focus=streamItem-27-3891078.0-0)).

I would appreciate it if you'd share some insights or pieces of evidence that it may help.

PR #2916, opened 2019-12-23
KT-29934 No quickfix on primary constructor override val parameter when it has wrong type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29934


vladimirdolzhenko:
Nice! Thank you

PR #2918, opened 2019-12-23
JVM_IR: do not remap locals in contexts nested inside lambdas

pyos:
The only case where this code is reachable at all is objects in lambdas that use reified type parameters of the outer inline function. Since the type parameters are declared outside the inlining root, regenerating the object is actually pointless, and remapping its captures even more so (not to mention that the code under the condition uses the captures of the lambda, not of the method it's currently transforming).

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
@@ -911,7 +911,7 @@ class MethodInliner(
             return
         }
 
-        if (inliningContext.isInliningLambda && inliningContext.lambdaInfo is IrExpressionLambda) {
+        if (inliningContext.isInliningLambda && inliningContext.lambdaInfo is IrExpressionLambda && !inliningContext.parent!!.isInliningLambda) {
             val capturedVars = inliningContext.lambdaInfo.capturedVars
             var offset = parameters.realParametersSizeOnStack
             val map = capturedVars.map {
--- a/compiler/testData/codegen/boxInline/reified/nonCapturingObjectInLambda.kt
+++ b/compiler/testData/codegen/boxInline/reified/nonCapturingObjectInLambda.kt
@@ -1,0 +1,19 @@
+// FILE: 1.kt
+package test
+
+inline fun <R> myRun(g: () -> R) = g()
+
+// FILE: 2.kt
+import test.*
+
+inline fun <reified A> Any?.complicatedCast() =
+    myRun {                     // 1. Inline function call with a lambda
+        object {                // 2. Anonymous object inside a lambda
+            fun f(x: Any?): A { // 3. Anonymous object uses a reified type parameter
+                val y = x       // 4. A method in that object uses locals
+                return y as A
+            }
+        }.f(this)               // 5. The lambda captures more values than the object
+    }
+
+fun box() = "OK".complicatedCast<String>()
--- a/compiler/tests/org/jetbrains/kotlin/codegen/BlackBoxInlineCodegenTestGenerated.java
+++ b/compiler/tests/org/jetbrains/kotlin/codegen/BlackBoxInlineCodegenTestGenerated.java
@@ -2880,6 +2880,11 @@ public class BlackBoxInlineCodegenTestGenerated extends AbstractBlackBoxInlineCo
             runTest("compiler/testData/codegen/boxInline/reified/kt9637_2.kt");
         }
 
+        @TestMetadata("nonCapturingObjectInLambda.kt")
+        public void testNonCapturingObjectInLambda() throws Exception {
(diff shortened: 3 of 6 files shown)


pyos:
@max-kammerer

pyos:
ping?

max-kammerer:
@pyos Could you clarify a bit: there are identical number of captured parameters in lambda and in inline function (by definition/construction of captured parameters in inline function) and as I see in debugger the only difference is owner type of captured parameters so i can't understand if old code is wrong why it's not reproduced in more boxInline test cases?

pyos:
Maybe the comment is unclear: it's the *object* that captures fewer parameters than the lambda. More specifically, the lambda captures `this` in order to pass it to `f`, while the object captures nothing.

>Why in other cases this owner type calculation is right?

The bug is only triggered when using a reified type parameter in an anonymous object defined inside a lambda, which is already rare; the requirement that the object only captures a subset of the lambda's captures makes it even harder.

max-kammerer:
Sorry for delay, I need more time review this change

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
```diff
@@ -909,7 +909,7 @@ class MethodInliner(
         }
 
         if (inliningContext.isInliningLambda && inliningContext.lambdaInfo is IrExpressionLambda) {
-            val capturedVars = inliningContext.lambdaInfo.capturedVars
+            val capturedVars = parameters.captured.map { it.desc }
```
max-kammerer: @pyos My apologies for delay in review, actually I can't catch root cause in this change. But after fresh view there are some results:

I'm observing next exception:
```
Caused by: org.jetbrains.kotlin.codegen.inline.InlineException: Transformer for _2Kt$complicatedCast$1$1: couldn't inline method call
Cause: f (Ljava/lang/Object;)Ljava/lang/Object;:
    @Lorg/jetbrains/annotations/Nullable;() // invisible, parameter 0
   L0
    LINENUMBER 8 L0
    ALOAD 1
    ASTORE 2
   L1
    LINENUMBER 9 L1
    GETSTATIC _2Kt$complicatedCast$1.$$$$this : Ljava/lang/Object;
    ICONST_1
    LDC "A"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.reifiedOperationMarker (ILjava/lang/String;)V
    CHECKCAST java/lang/Object
    ARETURN
```
And there is very strange     `_2Kt$complicatedCast$1.$$$$this : Ljava/lang/Object;` access inside 'f' function. That shouldn't be at all. This code should be executed only for root inline lambdas, `isInliningLambda` is inherited by inner context of anonymoys object transformation (yes, bad naming). Switching condition (using  original code for this line) makes test pass but add some new failings)

pyos: >Switching condition (using original code for this line) makes test pass but add some new failings

What condition did you try? Simply adding ` && !inliningContext.parent!!.isInliningLambda` to filter out contexts inside the lambda does seem to work.

In general, like I said in the PR description, the object does not actually need to be regenerated, since it's an IR-generated object inside a lambda and thus KT-28064 applies. The only reason this code is reachable with a context nested inside a lambda at all is that `AnonymousObjectTransformationInfo.shouldRegenerate` returns true if the object uses a reified type parameter, which is does here, even though the type parameter is defined outside the root inlining context and thus does not need to be reified yet. (The "needClassReification" marker unfortunately does not carry any information about the type parameters themselves, so it's impossible to find that out without first scanning the entire class.) Thus, literally any change that prevents the code from modifying anonymous objects inside lambdas (like the PR itself, or changing the condition to not do anything in nested contexts, or even having `MethodInliner.isAlreadyRegenerated` return true if `inliningContext.lambdaInfo is IrExpressionLambda`) will fix the issue.

max-kammerer: Yes, I've tried mentioned condition. I will prefer solution with patched condition in 'transformCaptured' cause it will allow to remove dangerous code (that could affect another regenerations triggered by any (right or wrong) reasons)

pyos: Done, though the regeneration condition should still be eventually modified to comply with KT-28064

max-kammerer: Yes, indeed. It's best to create separate pr for this

max-kammerer:
> Switching condition (using original code for this line) makes test pass but add some new failings)

Looks like locally failed tests are not related to pr

max-kammerer:
Thank you for pr!

PR #2922, opened 2019-12-24
KT-31020 Confusing "Put arguments on separate lines"/"Put arguments on one line" at the same time

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31020


yanex:
Two intentions are still available in these caret positions:

```
fun foo() {
    f (
        1,
        2,
        g<caret>(3, 4, 5)<caret>,
    3
    )
}
```

t-kameyama:
PTAL

yanex:
`val offset = element.findExistingEditor()?.caretModel?.offset ?: return true`

I would not use `findExistingEditor()` in an applicability check as there might not be an open editor.

t-kameyama:
> `val offset = element.findExistingEditor()?.caretModel?.offset ?: return true`
> 
> I would not use `findExistingEditor()` in an applicability check as there might not be an open editor.

Fixed in c82b7f20f88, please take another look.

yanex:
Review: OK.

yanex:
Merged to the intelli-kotlin repository:
- https://github.com/JetBrains/intellij-kotlin/commit/43312ec6f5ba01ddf736405eef7a807c6cd61c58

PR #2924, opened 2019-12-25
Allow parallel access to klib zip filesystem

alexander-gorshenev:


compiler/util-io/src/org/jetbrains/kotlin/konan/file/ZipUtil.kt
```diff
@@ -46,7 +53,13 @@ fun <T> File.withZipFileSystem(mutable: Boolean = false, action: (FileSystem) ->
     return try {
         action(zipFileSystem)
     } finally {
-        zipFileSystem.close()
+        val counter = counters[zipFileSystem]!!
+        if (counter == 0) {
+            zipFileSystem.close()
+            counters.remove(zipFileSystem)
+        } else {
+            counters.computeIfPresent(zipFileSystem, { t, u -> u + 1})
```
homuroll: `u - 1`

compiler/util-io/src/org/jetbrains/kotlin/konan/file/ZipUtil.kt
```diff
@@ -46,7 +53,13 @@ fun <T> File.withZipFileSystem(mutable: Boolean = false, action: (FileSystem) ->
     return try {
         action(zipFileSystem)
     } finally {
-        zipFileSystem.close()
+        val counter = counters[zipFileSystem]!!
+        if (counter == 0) {
```
homuroll: Why 0? We're decrementing after this check, so probably here we need to compare with 1.

alexander-gorshenev: Uh, seems to be a stale cherry-pick. We've already fixed it all somewhere. Let me force push.

alexander-gorshenev: There you go.

compiler/util-io/src/org/jetbrains/kotlin/konan/file/ZipUtil.kt
```diff
@@ -7,17 +7,24 @@ package org.jetbrains.kotlin.konan.file
...
     val attributes = hashMapOf("create" to mutable.toString())
     return try {
-        FileSystems.newFileSystem(zipUri, attributes, null)
+        FileSystems.newFileSystem(zipUri, attributes, null).also {
+            counters[it] = 1
+        }
     } catch (e: FileSystemAlreadyExistsException) {
-        FileSystems.getFileSystem(zipUri)
+        FileSystems.getFileSystem(zipUri).also {
+            counters.computeIfPresent(it, { t, u -> u + 1})
```
olonho: _, value -> value + 1

compiler/util-io/src/org/jetbrains/kotlin/konan/file/ZipUtil.kt
```diff
@@ -46,7 +53,13 @@ fun <T> File.withZipFileSystem(mutable: Boolean = false, action: (FileSystem) ->
     return try {
         action(zipFileSystem)
     } finally {
-        zipFileSystem.close()
+        val counter = counters[zipFileSystem]!!
+        if (counter == 1) {
+            zipFileSystem.close()
+            counters.remove(zipFileSystem)
+        } else {
+            counters.computeIfPresent(zipFileSystem, { t, u -> u - 1})
```
olonho: `_, value -> value - 1`

PR #2925, opened 2019-12-25
[KT-15363] Container Builders

Fleshgrinder:
Added container builders for lists, sets, and maps. I decided to use names without the build prefix because [sequence](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/sequence.html) has no prefix either. There is [buildString](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/build-string.html) but I think that the containers are closer to sequence than they are to strings and not having the prefix also ensures that these functions show up together with their non-builder counterparts (e.g. [listOf](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/list-of.html), [setOf](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/set-of.html)).

The new experimental type inference only works for the simple builders with a single generic type. The versions with nullability and the maps still require explicit specification of the types. Obviously explicit specification is required for all users who are not using the new experimental inference. Improving the type inference is something that should be done separately and many things – including these builders – will benefit from it, however, this is not a blocker for these builders in my opinion.

https://youtrack.jetbrains.com/issue/KT-15363


ilya-g:
Let's rebase this branch on master

libraries/stdlib/src/kotlin/collections/Sets.kt
```diff
@@ -93,3 +94,30 @@ internal fun <T> Set<T>.optimizeReadOnlySet() = when (size) {
...
 }
+
+/**
+ * Build a new read-only [Set] with the [elements][E] from the [builderAction].
+ *
+ * @sample samples.collections.Builders.Sets.buildSetSample
+ */
+@kotlin.internal.InlineOnly
+@SinceKotlin("1.4")
+@Suppress("NEWER_VERSION_IN_SINCE_KOTLIN", "API_NOT_AVAILABLE" /* See KT-30129 */) // TODO: remove this in 1.4
```
ilya-g: You can use `@SinceKotlin("1.3")` + `@ExperimentalStdlibApi`. This way new functions could be published in 1.3.x release

libraries/stdlib/src/kotlin/collections/Collections.kt
```diff
@@ -388,10 +389,12 @@ private fun rangeCheck(size: Int, fromIndex: Int, toIndex: Int) {
 
 @PublishedApi
 @SinceKotlin("1.3")
+@Suppress("NO_ACTUAL_FOR_EXPECT") // JS IR
```
ilya-g: For JS_IR we copy JS sources during build, so this suppress should not be needed.

libraries/stdlib/samples/test/samples/collections/builders.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    class Sets {
+        @Sample
+        fun buildSetSample() {
+            val x = buildSet {
+                add('a')
+                add('b')
+            }
+            assertPrints(x.sorted(), "[a, b]")
```
ilya-g: `buildSet` and `buildMap` preserve the insertion order, so sorting is not needed here.

libraries/stdlib/src/kotlin/collections/Maps.kt
```diff
@@ -123,22 +124,11 @@ public inline fun <K, V> linkedMapOf(): LinkedHashMap<K, V> = LinkedHashMap<K, V
...
-    }
-    if (expectedSize < INT_MAX_POWER_OF_TWO) {
-        return expectedSize + expectedSize / 3
-    }
-    return Int.MAX_VALUE // any large value
-}
-
-private const val INT_MAX_POWER_OF_TWO: Int = Int.MAX_VALUE / 2 + 1
+@Suppress("NO_ACTUAL_FOR_EXPECT") // JS IR
+internal expect fun mapCapacity(expectedSize: Int): Int
```
ilya-g: Actual functions must be marked with `@PublishedApi` too.

Fleshgrinder:
@ilya-g all done. ☺️

ilya-g:
Thank you, I've merged the commits manually.

PR #2927, opened 2019-12-25
KT-20357: Samples on linkedSetOf

mcserra:


ilya-g:
Thanks for the sample. 
I've added an explicit type specification for `set` variable to emphasize the difference of `linkedSetOf` from  `mutableSetOf` function, and then merged it manually: https://github.com/JetBrains/kotlin/commit/4fa5e2d85c6cb49ba25d0fc8b92d0b962ba91db8

PR #2930, opened 2019-12-26
KT-20357: Samples for Map functions "contains" and "isNotEmpty"

mcserra:


libraries/stdlib/samples/test/samples/collections/maps.kt
```diff
@@ -169,6 +169,28 @@ class Maps {
 
             // map.containsValue("string") // cannot call extension when the argument type and the map value type are unrelated at all
         }
+
+        @Sample
+        fun mapIsNotEmpty() {
```
ilya-g: Perhaps you could think up an example that is closer to a real use case, like [`stringIsEmpty`](https://github.com/JetBrains/kotlin/blob/555d503ff914705209817d326f01d668271957cd/libraries/stdlib/samples/test/samples/text/strings.kt#L287)?

mcserra: Would something like this be better?

```
        fun mapIsNotEmpty() {
            fun totalValue(valueMap: Map<String, Int>): String =
                if (valueMap.isNotEmpty())
                    "Value: [%s]".format(valueMap.map { entry -> entry.value }.sum())
                else "No values."

            val emptyMap: Map<String, Int> = mapOf()
            assertPrints(totalValue(emptyMap),"No values.")

            val map: Map<String, Int> = mapOf("x" to 1, "y" to 2)
            assertPrints(totalValue(map), "Value: [3]")
        }
```
I wouldn't say it's a real use case, but maybe it's good as a sample. Thanks

ilya-g: Yes, that `totalValue` function looks more like a some real function in code. :+1:

ilya-g:
Thanks for the sample, I've improved it a bit and merged manually: 61ad32f012789ef95376ba30812c32e779c4806b

PR #2931, opened 2019-12-27
FIR: Add bindings for dispatch receiver parameters

juan-chen:
Before this commit, such descriptors have null owners, which causes problems when the getter of the owner property is called.


mglukhikh:
Looks great

mglukhikh:
Merged manually 👍

PR #2932, opened 2019-12-28
KT-35648 False negative intention "Remove argument name" on named positional arguments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35648

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveArgumentNameIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveArgumentNameIntention.kt
@@ -18,7 +18,9 @@ package org.jetbrains.kotlin.idea.intentions
 
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.config.LanguageFeature
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
+import org.jetbrains.kotlin.idea.project.languageVersionSettings
 import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 import org.jetbrains.kotlin.resolve.calls.components.isVararg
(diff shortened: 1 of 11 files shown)


idea/testData/intentions/removeArgumentName/mixedNamedArgumentsInTheirOwnPosition.kt
```diff
@@ -0,0 +1,7 @@
+// COMPILER_ARGUMENTS: -XXLanguage:+MixedNamedArgumentsInTheirOwnPosition
+
+fun foo(name1: Int, name2: Int, name3: Int) {}
+
+fun usage() {
```
fedochet: It seems like there are missing test when `MixedNamedArgumentsInTheirOwnPosition` is enabled, but the arguments are not in their own positions, and the intention should be disabled

With your code, this case will look like `foo(1, <caret>name3 = 3, name2 = 2)`

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveArgumentNameIntention.kt
```diff
@@ -30,7 +32,9 @@ class RemoveArgumentNameIntention : SelfTargetingRangeIntention<KtValueArgument>
 
         val argumentList = element.parent as? KtValueArgumentList ?: return null
         val arguments = argumentList.arguments
-        if (arguments.asSequence().takeWhile { it != element }.any { it.isNamed() }) return null
+        if (!element.languageVersionSettings.supportsFeature(LanguageFeature.MixedNamedArgumentsInTheirOwnPosition)
+            && arguments.asSequence().takeWhile { it != element }.any { it.isNamed() }
```
fedochet: When `MixedNamedArgumentsInTheirOwnPosition` is enabled, you cannot skip this check. Actually you have to check that all of the previous arguments are not named, or are named but in their own positions

You can take a look at `NamedArgumentCompletion::isOnlyNamedArgumentExpected` to see how this check looks exactly

To check that your implementation is correct, try this test: 

```kt
// COMPILER_ARGUMENTS: -XXLanguage:+MixedNamedArgumentsInTheirOwnPosition
// IS_APPLICABLE: false

fun foo(name1: Int, name2: Int, name3: Int) {}

fun usage() {
    foo(name2 = 2, name1 = 1, <caret>name3 = 3)
}
```

The intention should not be applicable, because first two arguments are not in their own positions

idea/src/org/jetbrains/kotlin/idea/intentions/RemoveArgumentNameIntention.kt
```diff
@@ -30,7 +32,9 @@ class RemoveArgumentNameIntention : SelfTargetingRangeIntention<KtValueArgument>
...
+        ) return null
 
         val callExpr = argumentList.parent as? KtCallElement ?: return null
         val argumentMatch = callExpr.resolveToArgumentMatch(element) ?: return null
```
fedochet: You are using your own `resolveToArgumentMatch` function here

Please, replace it with `org.jetbrains.kotlin.resolve.calls.callUtil.getParameterForArgument`, it is designed specifically for this case

fedochet:
I've squashed and merged your changes in master in 2daedc98b5c66fa0b92ad1a3856df73f6948b1b8

Also fixed formatting of test cases a little and added ticket id to the commit message

Thanks @t-kameyama!

PR #2933, opened 2019-12-29
KT-34371 "Surround with lambda" quickfix is not available for suspend lambda parameters.

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34371


darthorimar:
Thanks!

PR #2934, opened 2019-12-30
IR: unify 3 copies of function body remapping

pyos:
Also,

  1. remove some redundant copies;

  2. fix remapping of non-local returns in lambdas if the body is moved after LocalDeclarationsLowering (the lambda is no longer inside the body, but must still be visited)


pyos:
ping?

gogabr:
Thank you!

PR #2935, opened 2019-12-31
KT-33771 False positive "Redundant Companion reference" with Java synthetic property and same-named object property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-33771#focus=streamItem-27-3887109.0-0


darthorimar:
LGTM, thanks!

PR #2936, opened 2020-01-02
IR: work around a bug in interface delegation descriptors

pyos:
interface I {
        fun f(x: Int = 1)
    }
    class C(val y: I) : I by y {
        // implicit `override fun f(x: Int) = y.f(x)` has a default value for `x`
    }

-- the only case where a function with overridden symbols has defaults.


udalov:
Thanks! I've reported the metadata issue as [KT-35798](https://youtrack.jetbrains.com/issue/KT-35798).

PR #2937, opened 2020-01-02
JVM_IR: fix a couple of inline class reflection issues.

madsager:
1. Postpone the computation of the signature for property
   reference getters for extension properties until codegen time.

2. Generate metadata for static replacement functions instead
   of the original functions.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/SignatureString.kt
```diff
@@ -20,15 +22,9 @@ import org.jetbrains.kotlin.resolve.jvm.AsmTypes
...
     override fun invoke(expression: IrFunctionAccessExpression, codegen: ExpressionCodegen, data: BlockInfo): PromisedValue? {
         val function = (expression.getValueArgument(0) as IrFunctionReference).symbol.owner
-
-        // TODO do not use descriptors
-        val declaration = codegen.context.referenceFunction(
-            DescriptorUtils.unwrapFakeOverride(function.descriptor).original
-        ).owner
-
-        val method = codegen.context.methodSignatureMapper.mapAsmMethod(declaration)
+        val resolved = if (function is IrFunctionImpl) function.collectRealOverrides().first() else function
```
udalov: Maybe use `is IrSimpleFunction` to avoid checking against particular implementation?

madsager: Thanks! Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -95,6 +99,33 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
             calculateOwner(expression.propertyContainer, this@PropertyReferenceLowering.context)
         }
 
+    private fun IrBuilderWithScope.computeSignatureString(expression: IrCallableReference): IrExpression {
+        return expression.getter?.let {
+            localPropertyIndices[it]?.let { irString("<v#$it>") }
```
udalov: Please rename the outer `it` here to something like `getter` to fix the inspection

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -95,6 +99,33 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
+                ?: if (it.owner.origin == IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR) {
+                    // Default property accessor. Compute the signature now, so that we will not get into trouble
+                    // if the getter is transformed to a static method by inline classes lowering.
+                    irString(expression.signature)
```
udalov: I suppose it'd be useful to inline the `signature` extension property here, since it also contains the logic about local variable references (`<v#...>`), which is not needed anymore here since by this point it's proven to be null

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -95,6 +99,33 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
+                } else {
+                    // Delay the computation of the signature until after inline classes lowering to make sure
+                    // we mangle the function names correctly for things like extension methods on inline classes.
+                    irCall(signatureStringIntrinsic).apply {
+                        putValueArgument(
```
udalov: Minor, but could you please join this into a single line for easier reading :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
madsager: Good idea, the squiggly was so small that I didn't even notice it. Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
madsager: Yes, that is better. Inlined.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
madsager: Of course! Done. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
madsager: And now it is actually inlined and simplified properly. Thanks.

udalov:
Thank you!

PR #2940, opened 2020-01-03
JVM_IR: Fix default argument bit mask for methods made static.

madsager:
When called by reflection the bit mask will be generated
discounting dispatch/extension receivers. Make sure that the
interpretation of the bit mask is consistent for direct and
reflective calls.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
@@ -582,13 +582,13 @@ fun createStaticFunctionWithReceivers(
             name = Name.identifier("this"),
             index = offset++,
             type = dispatchReceiverType!!,
-            origin = IrDeclarationOrigin.MOVED_RECEIVER_PARAMETER
+            origin = IrDeclarationOrigin.MOVED_DISPATCH_RECEIVER
         )
         val extensionReceiver = oldFunction.extensionReceiverParameter?.copyTo(
             this,
             name = Name.identifier("receiver"),
             index = offset++,
-            origin = IrDeclarationOrigin.MOVED_RECEIVER_PARAMETER
+            origin = IrDeclarationOrigin.MOVED_EXTENSION_RECEIVER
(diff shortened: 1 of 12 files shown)


madsager:
@max-kammerer Friendly ping. Could you have a look when time permits? :)

max-kammerer:
@madsager Sorry for delay.
Looks like if possible to switch `jvmInlineClassPhase` and 'defaultArgumentStubPhase' order new origin would be redundant, but it's not clear how many work it would require

madsager:
We should mark these receivers with a different origin in order to generate correct parameter metadata. However, you are right that using MOVE_RECEIVER_PARAMETER here doesn't generate the same parameter names metadata for the parameters as in the current backend. Seems that in the current backend, the dispatch receiver is marked as synthetic and the extension receiver is marked as mandated. With this code, they will both be synthetic. So, we need to change the origins of these anyway to get the right parameters metadata. I have updated the pull request to have two origins for that to match the current backend.

Changing the order of those two lowerings leads to other issues. One example is that you would then get names such as method$default-af3sz instead of the expected method-af3sz$default. So, I don't think that changing the order of the lowerings will make this easier. What do you think?

max-kammerer:
Than lets keep current approach

max-kammerer:
As I remember there should be some tests on synthetic/mandated but nothing is unmuted that's strange

max-kammerer:
Oh they are not for inline classes and this pr doesn't affects non-inline classes cases

max-kammerer:
But after some thoughts I'm not sure that this behaviour in old backend was intentionally done, maybe best to reset all flags in case of inline class methods. //cc @udalov @dnpetrov

madsager:
Yes, I'm not sure if this was intentional or not either. The code that does this is in the old backend is a combination of KotlinTypeMapper at around line 815 and generateJava8ParameterNames that explicitly maps (extension) RECEIVER to mandated and THIS to synthetic.

max-kammerer:
NB: It was intended for general class methods, but I'm not sure that it have any sence for inline class methods.

max-kammerer:
Thank you!

PR #2942, opened 2020-01-03
JVM_IR: Remove accessor for internal inline class properties.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt
```diff
@@ -78,8 +78,10 @@ private class JvmInlineClassLowering(private val context: JvmBackendContext) : F
 
         if (declaration.isInline) {
             val irConstructor = declaration.primaryConstructor!!
-            // The field getter is used by reflection and cannot be removed here.
-            declaration.declarations.remove(irConstructor)
+            // The field getter is used by reflection and cannot be removed here unless it is internal.
+            declaration.declarations.removeIf {
+                it == irConstructor || (it is IrFunction && it.isInlineClassFieldGetter && it.visibility == Visibilities.INTERNAL)
```
udalov: I'd suggest to use `!it.visibility.isPublicAPI` instead, which is effectively probably the same, but reads a bit nicer, and is precisely replicating the old backend's behavior in `PropertyCodegen.areAccessorsNeededForPrimaryConstructorProperty`

madsager: Good idea, done. :)

udalov:
@madsager Thanks!

PR #2943, opened 2020-01-03
JVM_IR: support non-local break/continue in the inliner 

pyos:
Not in the frontend or psi2ir, though, so this not a complete implementation of KT-1436, but rather a part of it that is currently useful to make other code compile. In particular, lambdas passed to array constructors and JVM-style `assert` are inlined as IR returnable blocks, which are then converted into `do { ... } while (false)` loops, so non-local returns from them become non-local `break`s.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -725,35 +724,35 @@ class ExpressionCodegen(
...
-            loop.body?.accept(this, data)?.discard()
-        }
-        mv.goTo(continueLabel)
-        mv.mark(endLabel)
-        return immaterialUnitValue
-    }
-
-    override fun visitDoWhileLoop(loop: IrDoWhileLoop, data: BlockInfo): PromisedValue {
+    override fun visitLoop(loop: IrLoop, data: BlockInfo): PromisedValue {
+        // Spill the stack in case the loop contains inline functions that break/continue
```
max-kammerer: I would prefer to keep original separation: visitWhileLoop/visitDoWhileLoop

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -778,9 +777,13 @@ class ExpressionCodegen(
         jump.markLineNumber(startOffset = true)
         val endLabel = Label()
         val stackElement = unwindBlockStack(endLabel, data) { it is LoopInfo && it.loop == jump.loop } as LoopInfo?
-            ?: throw AssertionError("Target label for break/continue not found")
-        mv.fixStackAndJump(if (jump is IrBreak) stackElement.breakLabel else stackElement.continueLabel)
-        mv.mark(endLabel)
+        if (stackElement == null) {
```
max-kammerer: Is it possible to search proper loop outside (in inlinedInto codegens)?. And if not present throw exception, otherwise generate global return flag?

pyos: This data is stored not in the codegen itself, but in `BlockInfo`, which we don't have here. Though I suppose ExpressionCodegen could easily be changed to store the block stack in a field instead of passing it around as a parameter.

max-kammerer: Or pass it to ExpressionCodegen constructor

max-kammerer: Actually this is minor, because there should be some IR validation and this structure should be consistent. This request could be simply skipped

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -825,9 +833,13 @@ class ExpressionCodegen(
...
+            generateGlobalReturnFlagIfPossible(jump.loop.nonLocalReturnLabel(jump is IrBreak))
+            mv.areturn(Type.VOID_TYPE)
+        } else {
+            mv.fixStackAndJump(if (jump is IrBreak) stackElement.breakLabel else stackElement.continueLabel)
```
max-kammerer: Looks like `fixStackAndJump` should be also called in `then` branch in general case:  e.g. foo(param1, param2 ?: return@xxx), where return@xxx transformed to non-local break/continue

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -274,6 +272,7 @@ private val jvmFilePhases =
         singleAbstractMethodPhase then
         assertionPhase then
         returnableBlocksPhase then
+        sharedVariablesPhase then
```
max-kammerer: Why this move is required?

compiler/testData/codegen/boxInline/nonLocalReturns/fromArrayGenerator.kt
```diff
@@ -0,0 +1,5 @@
+// FILE: 1.kt
+inline fun runReturning(f: () -> Nothing): Nothing = f()
+
+// FILE: 2.kt
+fun box() = Array<String>(1) { runReturning { return@Array "OK" } }[0]
```
max-kammerer: @pyos Could you also add test with several nested inline calls?
e.g.:
` fun box() = Array<String>(1) { run { runReturning { return@Array "OK" } }}[0] `

and nested calls with capturing
` fun box() { 
val result = "OK"
Array<String>(1) { run { run { runReturning { return@Array result } }}}[0] 
}`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: This is true, but fixStackAndJump can only directly precede a jump instruction, so we can't generate it here yet. Instead, see the changes I made in MethodInliner: if the jump target is not the end of the inlined method, it will insert a fixStackAndJump and put any finally blocks before it.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
pyos: If a returnable block has multiple exit points, returnableBlocksPhase will create a variable and replace each return with a store to it followed by a break. If the return is inside an inline lambda, the variable might need to be wrapped into a ref for the store to work.

compiler/testData/codegen/boxInline/nonLocalReturns/fromArrayGenerator.kt
pyos: Done

pyos:
@max-kammerer so what to you think?

max-kammerer:
@pyos  Could you also rebase this one? 
Sorry for huge delay in response

pyos:
Rebased.

pyos:
@max-kammerer ping?

madsager:
Drive-by: what happened to this change @max-kammerer @pyos? Comments on https://youtrack.jetbrains.com/issue/KT-42932 seems to indicate that it fixes that issue. Should we rebase this, double check and get it landed?

pyos:
Rebased.

max-kammerer:
Sorry, I've tottaly forgot about this pr

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/CapturedVarsOptimizationMethodTransformer.kt
```diff
@@ -79,26 +62,29 @@ class CapturedVarsOptimizationMethodTransformer : MethodTransformer() {
...
+
+            methodNode.removeEmptyCatchBlocks()
+            methodNode.removeUnusedLocalVariables()
         }
 
         private fun AbstractInsnNode.getIndex() = methodNode.instructions.indexOf(this)
 
         private fun createRefValues() {
-            for (insn in insns) {
+            for (insn in methodNode.instructions.asSequence()) {
```
max-kammerer, line 87: Why not `for (insn in methodNode.instructions) `

max-kammerer:
@dnpetrov Could you also look into `CapturedVarsOptimizationMethodTransformer.kt` ?

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/CapturedVarsOptimizationMethodTransformer.kt
max-kammerer: and similar below

pyos: I think `InsnList` didn't implement `Iterable` in asm 6.

dnpetrov:
> @dnpetrov Could you also look into `CapturedVarsOptimizationMethodTransformer.kt` ?

`CapturedVarsOptimizationMethodTransformer` - LGTM.

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/CapturedVarsOptimizationMethodTransformer.kt
max-kammerer: Yes, it's true. But compiler  support now only 201+ platform and it should be asm 7+ everywhere

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/LambdaInfo.kt
```diff
@@ -256,7 +254,7 @@ class PsiExpressionLambda(
 
     val functionWithBodyOrCallableReference: KtExpression = (expression as? KtLambdaExpression)?.functionLiteral ?: expression
 
-    private val labels: Set<String>
+    override val returnLabels: Map<String, Label?>
```
max-kammerer, line 257: Maybe best to keep it private and just add new method here  to fill labels map required for IrSourceCompilerForInline?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -161,3 +166,6 @@ private tailrec fun IrDeclaration.isInlineOrInsideInline(): Boolean {
     if (parent !is IrDeclaration) return false
     return parent.isInlineOrInsideInline()
 }
+
+// TODO generate better labels; this is unique (includes the object's address), but not very descriptive
+internal fun IrLoop.nonLocalReturnLabel(forBreak: Boolean): String = "$this\$${if (forBreak) "break" else "continue"}"
```
max-kammerer, line 171: Is it possible to use another approach? I'm worry about hash inside `toString` call (in some rare cases there can be collisions). Maybe some kind of counter?

max-kammerer: ...and more debugable solution

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/CapturedVarsOptimizationMethodTransformer.kt
pyos: Done

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/LambdaInfo.kt
pyos: TBH that seems like an unnecessarily verbose way to achieve the same result, since both the field and the map are immutable anyway

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
pyos: Apparently `IrLoop` has a `label` field. Not sure what the point is since loops are referenced directly and not by label after the frontend, but I've reused it to give each loop a name that at least includes the name of the enclosing declaration and a unique counter.

dnpetrov: > Apparently `IrLoop` has a `label` field. Not sure what the point is since loops are referenced directly and not by label after the frontend, but I've reused it to give each loop a name that at least includes the name of the enclosing declaration and a unique counter.

It's preserved there for the sake of transpiler-like backends such as Kotlin/JS. Not quite sure that Kotlin/JS actually uses it, but they specifically requested it.

max-kammerer:
@pyos, sorry for delay in review. Everything looks fine. Could you also add more test...

compiler/testData/codegen/boxInline/nonLocalReturns/fromArrayGeneratorWithFinally.kt
```diff
@@ -0,0 +1,19 @@
+// FILE: 1.kt
+inline fun runReturning(f: () -> Nothing): Nothing = f()
+
+// FILE: 2.kt
+fun box(): String {
+    var r = ""
+    val x = try {
+        Array<String>(1) {
```
max-kammerer, line 8: 1. ...test with additional external layer of Array with try/finallies:
 a. where return jumps to external one 
 b. and to internal one

2. And some test with catch/exception throwing

pyos: Sorry, I don't understand what you mean...

max-kammerer: for 1 I meant smth like this
```
// FILE: 2.kt
fun box(): String {
    var r = ""
    val x = try {
        Array<String>(1) ext@{
            try {
                Array<String>(1) {
                    try {
                        runReturning { return@ext/Array "OK" }
                    } finally {
                        r += "O"
                    }
                }[0]
            } finally {
                r += ...
            }

        } )
    } finally {
        r += "K"
    }
    return r
}
```

max-kammerer: And for second one  original test with catch clause that is executed

pyos: Done, I guess?

max-kammerer: Almost, last two suggestion: It's good to have test similar to `fromArrayGeneratorWithFinallyX2.kt`  with return@Array  to inner lambda  (that I meant by second `Array` part in `return@ext/Array`). 

Catch test is also good. Could you add it's copy with catch inside  lambda?

pyos: I don't really get what's the point in adding tests with `catch` clauses, since no non-local returns could be executed in those.

max-kammerer:
@pyos Many thanks!

compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/CapturedVarsOptimizationMethodTransformer.kt
max-kammerer: I was wrog here ( 
`InsnList now implements Iterable interface` from 7.2, but 201 uses 7.0. Will fix

PR #2944, opened 2020-01-03
[FIR]: add missed bindings for WrappedReceiverParameterDescriptor.

juan-chen:


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -325,15 +329,19 @@ class Fir2IrDeclarationStorage(
             }
             if (containingClass != null && !isStatic) {
                 val thisType = containingClass.thisReceiver!!.type
+                val descriptor = WrappedReceiverParameterDescriptor()
```
punzki: If you make similar changes in `declareDefaultSetterParameter`, this also fixes `testData/codegen/box/traits/inheritedVar.kt` 🙂 (sorry GitHub won't let me comment on line 281)

juan-chen: Done. Thanks!

mglukhikh:
Merged manually 👍

PR #2948, opened 2020-01-05
KT-35805 FoldInitializerAndIfToElvis: should not add new line for multiline initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35805


idea/testData/inspectionsLocal/foldInitializerAndIfToElvis/MultiLineInitializer.kt
```diff
@@ -0,0 +1,14 @@
+fun foo(vararg args: String): String? = null
+
+fun test(): Int {
```
igoriakovlev, line 3: It seems that here we have wrong test data with redundant tab's.
This test does not green.

t-kameyama:
@igoriakovlev
I've fixed the test, please review.

igoriakovlev:
It seems that everything is great. Thank you!

PR #2951, opened 2020-01-06
JVM_IR: fix default argument stub visibility.

madsager:
Match the rules from the old backend:

1. Private and inline only methods get package private default
   argument stubs.

2. Everything else is public.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -165,6 +165,9 @@ fun IrDeclarationWithVisibility.getVisibilityAccessFlag(kind: OwnerKind? = null)
         ?: visibilityToAccessFlag[visibility]
         ?: throw IllegalStateException("$visibility is not a valid visibility in backend for ${ir2string(this)}")
 
+fun IrFunction.getVisibilityForDefaultArgumentStub(): Int =
```
udalov: I'd implement it like this, and probably move it to `FunctionCodegen`:
```
when (visibility) {
    Visibilities.PUBLIC -> Opcodes.ACC_PUBLIC
    JavaVisibilities.PACKAGE_VISIBILITY -> AsmUtil.NO_FLAG_PACKAGE_PRIVATE
    else -> error("Default argument stub should be either public or package private: ${ir2string(this)}")
}
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -16,9 +16,11 @@ import org.jetbrains.kotlin.codegen.state.GenerationState
 import org.jetbrains.kotlin.codegen.visitAnnotableParameterCount
 import org.jetbrains.kotlin.config.LanguageFeature
 import org.jetbrains.kotlin.descriptors.Modality
+import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.util.*
+import org.jetbrains.kotlin.load.java.JavaVisibilities
```
udalov: These imports seem unused

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
madsager: I agree. Thanks. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
madsager: They were, thanks. Now they are not anymore as I moved the code here. :)

madsager:
The change to only accept public and package private has made a number on constructor tests fail, so don't merge in this state. I'm looking into it. :)

madsager:
Fixed now. The issue was that enum class lowering did not preserve the visibility of the constructor. Therefore, the default argument constructor did not become package-private. It should and it is now.

udalov:
Thank you!

PR #2952, opened 2020-01-06
KT-35757 Inspection: convert initialized val to non-null type

kvirolainen:
https://youtrack.jetbrains.com/issue/KT-35757

### DIFF
--- a/idea/resources/META-INF/plugin-common.xml
+++ b/idea/resources/META-INF/plugin-common.xml
@@ -3471,6 +3471,15 @@
                      language="kotlin"
     />
 
+    <localInspection implementationClass="org.jetbrains.kotlin.idea.inspections.ConvertInitializedValToNonNullTypeInspection"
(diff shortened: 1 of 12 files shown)


t-kameyama:
False positive:
```
fun foo() {
    val s1: String? = bar()
    val s2: String? = null
}

fun bar(): String? = ""
```

kvirolainen:
Fixed

vladimirdolzhenko:

Sorry for the delay, LGTM

Resolved small conflict in `plugin-common.xml` (inspections have been migrated to `inspections.xml`) and merge PR manually.

vladimirdolzhenko:
My bas I missed some case: it produces false positive for vars

vladimirdolzhenko:
Reverted back as duplicate of https://youtrack.jetbrains.com/issue/KT-19321 / https://github.com/JetBrains/kotlin/pull/3657

PR #2953, opened 2020-01-06
[JVM IR] Support target templates in AbstractCheckLocalVariablesTableTest and fix some tests

punzki:
I also have a version of this fix similar to what @max-kammerer did with https://github.com/JetBrains/kotlin/commit/496765f41e26ea05a203a92e3ac39fea08e6604e where you can use `// JVM_TEMPLATES` and `// JVM_IR_TEMPLATES` to subdivide the expectations; let me know if that fix is preferable. I figured this way is simpler since there was only one test, but that other way would be better if there were many exceptions (which there aren't at the moment).


udalov:
Since you're saying that you already have a fix with the separate template approach, could you show it please? :)

Ideally, we'd like to minimize the number of duplicated tests, and if you already have that done, that change would be preferable. (Also, looks like it'd be useful for `itInReturnedLambda.kt` too, which will fail as soon as we revert fbe66c34, as per the internal discussion in #ext-google-compiler.)

punzki:
@udalov Here's that version: https://github.com/google/kotlin/pull/new/prr/punzki/lvt-localFun-fix-2 (sorry for delay, had to clean it up a little) Looking at it, I personally prefer that, and I will change the PR to use that version (and fix `itInReturnedLambda.kt` too), unless you have objections?

Ideally I would like to refactor the 2 tests (bytecode and LVT) and add a filtering directive, something like:

```
...common lines here...

// !APPLIES_TO_BACKEND: JVM
// METHOD : LocalFunKt$foo$1.invoke()V
// VARIABLE : NAME=y TYPE=I INDEX=1
// VARIABLE : NAME=this TYPE=LLocalFunKt$foo$1; INDEX=0

// !APPLIES_TO_BACKEND: JVM_IR
// METHOD : LocalFunKt.foo$bar(I)V
// VARIABLE : NAME=y TYPE=I INDEX=1
// VARIABLE : NAME=$x TYPE=I INDEX=0

// Optionally you can add more common lines:
// !APPLIES_TO_BACKEND: ANY
...more common lines here...
```

...I would prefer this but I can do that separately after all these "different in IR vs non-IR" LVT test failures are fixed.

udalov:
Thanks! Yes, let's use the version with `JVM_TEMPLATES/JVM_IR_TEMPLATES` right now in this PR :)

Not sure if the `APPLIED_TO_BACKEND` approach is that much better though, since these tests are only relevant for JVM & JVM_IR anyway...

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
```diff
@@ -33,43 +35,86 @@ abstract class AbstractCheckLocalVariablesTableTest : CodegenTestCase() {
...
+            } else if (currentBackend == TargetBackend.ANY || currentBackend == backend) {
+                filteredLines.add(line)
+            }
+        }
+        return filteredLines
+    }
+
+    private fun doCompare(
+        testFileLines: List<String>,
         text: String,
```
udalov: The `text` parameter is not used anymore

punzki:
> Thanks! Yes, let's use the version with `JVM_TEMPLATES/JVM_IR_TEMPLATES` right now in this PR :)

Updated the PR and fixed `box/reflection/classes/localClassSimpleName.kt`.

> Not sure if the `APPLIED_TO_BACKEND` approach is that much better though, since these tests are only relevant for JVM & JVM_IR anyway...

I was thinking I could add some common logic in `CodegenTestCase` to do the filtering of lines in the file, before doing the test. That way all the subclasses of tests could benefit.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/AbstractCheckLocalVariablesTableTest.kt
punzki: Fixed

udalov:
Merged, thank you!

Regarding extracting the common logic into `CodegenTestCase`: I don't have anything against it, but the test code is a bit dated and contrived, so it might take a bit more time than appears :). I'd maybe wait until there are more use cases, just to avoid sinking into refactoring everything there. But it's just my opinion :).

udalov:
@punzki Looks like this change causes `IrCompileKotlinAgainstInlineKotlinTestGenerated.AnonymousObject.testKt15751` to fail on Windows because apparently we now generate a file with the name `A$foo$<no name provided>$1$1.class`, and `<`/`>` are not welcome in Windows file names:
```
java.io.FileNotFoundException: C:\BuildAgent\temp\buildTmp\compilerProject_test_4395430721349375947\testRootGW33\unitTest_kt15751\AnonymousObjecttestKt15751\a\test\A$foo$<no name provided>$1$1.class (The filename, directory name, or volume label syntax is incorrect)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at com.intellij.openapi.util.io.FileUtil.writeToFile(FileUtil.java:1057)
```

Could you take a look please?

punzki:
> @punzki Looks like this change causes `IrCompileKotlinAgainstInlineKotlinTestGenerated.AnonymousObject.testKt15751` to fail on Windows because apparently we now generate a file with the name `A$foo$<no name provided>$1$1.class`, and `<`/`>` are not welcome in Windows file names:

@udalov I will look into it, thanks for letting me know.

PR #2954, opened 2020-01-07
KT-24522 Suggest to move typealias outside the class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24522


vladimirdolzhenko:
LGTM, merged manually (to resolve a small conflict)

PR #2956, opened 2020-01-07
JVM IR: Get rid of FAKE_OWNER in MethodSignatureMapper

sfs:
Here's my understanding of why we have `FAKE_OWNER` in the first place. Please let my know if I missed something!

---

The JVM IR backend handles most intrinsics in lowerings or as part of `IrIntrinsicMethods`. However, there are nine intrinsics which the JVM backend only handles during inlining. These are the inline array constructors (`arrayOf`, `emptyArray`), three special coroutine intrinsics (`intercepted`, `coroutineContext`, `suspendCoroutineUninterceptedOrReturn`), the `typeOf` intrinsic, two generic enum methods (`enumValues`, `enumValueOf`), and the synthetic `alwaysEnabledAssert`.

Four of these look like regular methods before we enter the inliner (the coroutine intrinsics and `typeOf`). No special handling is necessary in `ExpressionCodegen`. Similarly, no special handling is necessary for `alwaysEnabledAssert`, since we only create and process calls to this method in the inliner itself.

The remaining four inline intrinsics really are intrinsics, however. There is no code for them in the standard library and *they do not have class parents* in the IR.

This means that `ExpressionCodegen` has to be careful when computing the owner for a function call, even after `IrIntrinsicMethods` have been resolved. In particular, `MethodSignatureMapper` puts in a `FAKE_OWNER` when it cannot resolve a call.

This PR lowers or implements the remaining inline intrinsics as regular `IrIntrinsicMethods` and adds more stringent error checking in `ExpressionCodegen`.

I think this is important, because right now there are cases where compilation succeeds while quietly producing incorrect bytecode. Case in point: There are 35 FIR tests which *passed* even though the resulting code contained calls to non-existent methods (which were never executed as part of the test).


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/EnumIntrinsics.kt
```diff
@@ -0,0 +1,64 @@
...
+            // <REIFIED-OPERATIONS-MARKER>
+            // ICONST_0
+            // ANEWARRAY Ljava/lang/Enum;
+            putReifiedOperationMarkerIfTypeIsReifiedParameter(type, ReifiedTypeInliner.OperationKind.ENUM_REIFIED)
+            mv.iconst(0)
+            mv.newarray(AsmTypes.ENUM_TYPE)
+            MaterialValue(codegen, ENUM_ARRAY_TYPE, expression.type)
+        } else {
+            val enumType = typeMapper.mapType(type)
+            val enumArrayType = Type.getObjectType("[${enumType.descriptor}")
```
udalov: Minor suggestion to use `AsmUtil.getArrayType` here and below

compiler/testData/cli/jvm/firHello.args
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: ANY
```
udalov: Maybe comment out the `kotlin.io.println` call instead, since the rest of the test works?

cc @demiurg906

udalov:
Thank you! Your understanding is correct and this change looks great. Will merge after a remote run.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/EnumIntrinsics.kt
sfs: Done!

compiler/testData/cli/jvm/firHello.args
sfs: You're right, that makes more sense! Done.

udalov:
Thanks, merged!

PR #2959, opened 2020-01-08
[JVM IR] Fix issue with destructuring declaration in parameter for suspend lambda

punzki:
The Name for the special destructuring declaration parameter was incorrectly turned into a regular/non-special Name when the parameter was moved to a field.


ilmirus:
Thanks!

PR #2960, opened 2020-01-08
KT-26248 "Refactor -> Inline variable" breaks callable references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-26248


mglukhikh:
Merged 👍

PR #2961, opened 2020-01-08
[JVM IR] Mangle variable names for anonymous parameters in lambdas

punzki:
This also fixes the same 3 tests fixed by https://github.com/JetBrains/kotlin/pull/2959. However, it's coincidental simply because the Name for a destructuring declaration parameter is no longer "special". The changes in https://github.com/JetBrains/kotlin/pull/2959 are still correct and valid.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
@@ -7,6 +7,7 @@ package org.jetbrains.kotlin.backend.jvm
 
 import org.jetbrains.kotlin.backend.common.ir.createParameterDeclarations
 import org.jetbrains.kotlin.codegen.SamType
+import org.jetbrains.kotlin.codegen.getNameForDestructuredParameterOrNull
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.descriptors.annotations.FilteredAnnotations
 import org.jetbrains.kotlin.ir.builders.declarations.buildClass
@@ -20,7 +21,11 @@ import org.jetbrains.kotlin.load.java.sam.SamConstructorDescriptor
 import org.jetbrains.kotlin.load.java.sam.SingleAbstractMethodUtils
 import org.jetbrains.kotlin.load.java.typeEnhancement.hasEnhancedNullability
 import org.jetbrains.kotlin.load.kotlin.JvmPackagePartSource
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtParameter
 import org.jetbrains.kotlin.psi2ir.generators.GeneratorExtensions
+import org.jetbrains.kotlin.resolve.DescriptorToSourceUtils
+import org.jetbrains.kotlin.resolve.calls.util.isSingleUnderscore
 import org.jetbrains.kotlin.resolve.jvm.JvmClassName
 import org.jetbrains.kotlin.resolve.jvm.annotations.hasJvmFieldAnnotation
 import org.jetbrains.kotlin.serialization.deserialization.descriptors.DeserializedContainerSource
@@ -29,6 +34,7 @@ import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.TypeSubstitutor
 import org.jetbrains.kotlin.types.Variance
 import org.jetbrains.kotlin.types.typeUtil.replaceAnnotations
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 class JvmGeneratorExtensions(private val generateFacades: Boolean = true) : GeneratorExtensions() {
     val classNameOverride = mutableMapOf<IrClass, JvmClassName>()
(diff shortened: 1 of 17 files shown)


punzki:
Here's a version that renames the anonymous parameters in a JVM-specific lowering: https://github.com/google/kotlin/pull/new/prr/punzki/lvt-anonymous-fix-2 (no change required in `testData/ir/irText/lambdas/destructuringInLambda.txt` since it's in lowering and not in PSI2IR). 
Please let me know which version is preferable.

compiler/testData/ir/irText/lambdas/destructuringInLambda.txt
```diff
@@ -135,13 +135,13 @@ FILE fqName:<root> fileName:/destructuringInLambda.kt
     FIELD PROPERTY_BACKING_FIELD name:fn type:kotlin.Function1<<root>.A, kotlin.Int> visibility:private [static]
       EXPRESSION_BODY
         FUN_EXPR type=kotlin.Function1<<root>.A, kotlin.Int> origin=LAMBDA
-          FUN LOCAL_FUNCTION_FOR_LAMBDA name:<anonymous> visibility:local modality:FINAL <> (<name for destructuring parameter 0>:<root>.A) returnType:kotlin.Int
-            VALUE_PARAMETER name:<name for destructuring parameter 0> index:0 type:<root>.A
+          FUN LOCAL_FUNCTION_FOR_LAMBDA name:<anonymous> visibility:local modality:FINAL <> ($dstr$_u24__u24$y:<root>.A) returnType:kotlin.Int
```
madsager: Are these tests only enabled for the JVM_IR backend? Otherwise, I guess this will fail for js and native since the mangling happens in a JVM-specific generator extension.

punzki: AFAICT yes. There are IR text tests for JS_IR, but the test data for that is under `testData/ir/irJsText` (base class is `AbstractIrJsTextTestCase`). Also AFAICT, no specific tests for Native.

punzki:
Added fix for tests in `checkLocalVariablesTableTest/parametersInSuspendLambda` (fields are first stored in local variables for non-IR, and read directly in IR). There still seems to be issues in inlining suspend lambdas, so I left a TODO for `checkLocalVariablesTableTest/parametersInSuspendLambda/inline.kt`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmGeneratorExtensions.kt
```diff
@@ -74,6 +80,17 @@ class JvmGeneratorExtensions(private val generateFacades: Boolean = true) : Gene
         else
             null
 
+    override fun computeParameterName(descriptor: ParameterDescriptor): Name {
```
dnpetrov: Minor, just curious: why is it JVM-specific?

punzki: Honestly I'm not sure if it should be specific to JVM or not. The mangling only happens in codegen for the old JVM backend, so I took that as a cue that it should be specific to JVM.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt
```diff
@@ -33,6 +33,7 @@ import org.jetbrains.kotlin.ir.types.impl.IrUninitializedType
 import org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid
 import org.jetbrains.kotlin.ir.visitors.acceptChildrenVoid
 import org.jetbrains.kotlin.ir.visitors.acceptVoid
+import org.jetbrains.kotlin.name.Name
```
dnpetrov: Minor: unused import. I'll remove it before pushing.

dnpetrov:
Overall it looks ok to me. I'll run tests and push manually.
Also, I'll discuss anonymous value parameter mangling with my colleagues from Kotlin/JS and Kotlin/Native.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -221,7 +221,7 @@ class JvmSymbols(
                 addValueParameter("completion", continuationType.makeNullable())
             }
             klass.addFunction("invokeSuspend", irBuiltIns.anyNType, Modality.ABSTRACT).apply {
-                addValueParameter("result", resultClassStub.typeWith(irBuiltIns.anyNType))
+                addValueParameter("\$result", resultClassStub.typeWith(irBuiltIns.anyNType))
```
ilmirus: I somehow missed it. Thanks!
Minor: there is `SUSPEND_CALL_RESULT_NAME`, which, in fact, I should've used.

dnpetrov:
Discussed it with colleagues. JVM-specific lovering looks like a better approach here. 
@ilmirus please, take a look at https://github.com/JetBrains/kotlin/pull/2966

punzki:
@dnpetrov @ilmirus Thanks for discussing this and approving the other PR!

PR #2962, opened 2020-01-08
JVM_IR: skip synthetic enum parameters in signatures

pyos:


max-kammerer:
Thank you for pr!
I will look other ones late

PR #2964, opened 2020-01-08
[FIR] Fix Fir2Ir translation of library function parameters with defa…

juan-chen:
…ult values.

In the existing implementation, the information that a library function parameter has a default value is lost during Fir2Ir translation, and thus later on, calls to such functions are not converted to the corresponding stubs.

Test cases such as kt5844, which calls kotlin.test.assertEquals, fail with messages like "java.lang.IllegalArgumentException: Null argument in ExpressionCodegen for parameter VALUE_PARAMETER name:message index:2 type:kotlin.String?", due to nonnull checks in ExpressionCodegen.visitFunctionAccess(ExpressionCodegen.kt:421)

Functions defined in the code to be compiled don't have this problem, only those from deserializing jars.

This commit sets default values in IrValueParameterImpl for such parameters during Fir2Ir translation. Now kt5844 passes the nonnull check in ExpressionCodegen.visitFunctionAccess but still fails in a later stage (java.lang.ClassNotFoundException: kotlin.internal.ir.Intrinsic).


mglukhikh:
Merged manually 👍

PR #2966, opened 2020-01-09
[JVM IR] Mangle variable names for anonymous parameters in lambdas.

dnpetrov:
See https://github.com/JetBrains/kotlin/pull/2961


ilmirus:
Thanks!

PR #2967, opened 2020-01-09
JVM IR: Refactor EnumClassLowering

sfs:
Some refactoring around `EnumClassLowering`. Specifically:
- There is an unused argument in `DeclarationFactory.getFieldForEnumEntry`
- The `IrEnumValueOf` intrinsic is unecessary since we can now produce java class references in a lowering
- There is a bug in psi2ir, which causes it to omit initializer initializer calls in some enum constructors. There are workarounds for this in both the JS IR and JVM IR backends, but it's better to just fix it in psi2ir.
- Finally, a lot of the code in `EnumClassLowering` predates the nice ir builder interfaces and can be a lot more concise now. :)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/DeclarationFactory.kt
```diff
@@ -11,7 +11,7 @@ import org.jetbrains.kotlin.ir.types.IrType
 interface DeclarationFactory {
     object FIELD_FOR_OUTER_THIS : IrDeclarationOriginImpl("FIELD_FOR_OUTER_THIS", isSynthetic = true)
 
-    fun getFieldForEnumEntry(enumEntry: IrEnumEntry, entryType: IrType): IrField
+    fun getFieldForEnumEntry(enumEntry: IrEnumEntry): IrField
```
max-kammerer: Is it also unused in K/N?

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/KotlinUtils.kt
```diff
@@ -54,11 +54,15 @@ fun ResolvedCall<*>.isValueArgumentReorderingRequired(): Boolean {
 }
 
 fun KtSecondaryConstructor.isConstructorDelegatingToSuper(bindingContext: BindingContext): Boolean {
-    val delegatingResolvedCall = getDelegationCall().getResolvedCall(bindingContext) ?: return false
     val constructorDescriptor = bindingContext.get(BindingContext.CONSTRUCTOR, this) ?: return false
-    val ownerClassDescriptor = constructorDescriptor.containingDeclaration
-    val targetClassDescriptor = delegatingResolvedCall.resultingDescriptor.containingDeclaration
-    return targetClassDescriptor != ownerClassDescriptor
+    val delegatingResolvedCall = getDelegationCall().getResolvedCall(bindingContext)
```
max-kammerer: Looks like K/N also could be affected

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/DeclarationFactory.kt
sfs: Yes, K/N doesn't implement the `DeclarationFactory` interface.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/KotlinUtils.kt
sfs: Oh, you're right. Could you please run the K/N tests on this branch? I still haven't managed to set this up properly.

max-kammerer: "Composite Kotlin/Native sanity test" build is successful

max-kammerer: @sfs Looks like this change also requires irText test  (@dnpetrov please take a look)

compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/lower/EnumClassLowering.kt
```diff
@@ -125,9 +125,6 @@ class EnumClassConstructorTransformer(val context: CommonBackendContext, private
     private val loweredEnumConstructors = HashMap<IrConstructorSymbol, IrConstructor>()
 
     fun transform(): List<IrDeclaration> {
-        // Make sure InstanceInitializer exists
```
max-kammerer: @skuzmich Please take a look for js/lower/EnumClassLowering.kt  changes

skuzmich: @sfs thanks a lot for dealing with this workaround. LGTM.

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/KotlinUtils.kt
dnpetrov: @sfs Looks ok to me.
Please, add corresponding psi2ir test (test data located at compiler/testData/ir/irText).

sfs: I've added a test based on `secondaryConstructors/enums.kt`

max-kammerer:
@sfs Thank you for pr!

PR #2968, opened 2020-01-09
JVM_IR: handle suspend functions for signatures in callable references.

madsager:


madsager:
@ilmirus Looks like a couple of the tests fixed by this are also fixed on your state-machine branch, so I hope this doesn't conflict with what you are doing. I checked out your branch, but it is a bit behind master at this point. Let me know if you would rather have me wait until you are ready to land your branch. I can keep working without landing this.

PR #2969, opened 2020-01-09
KT-34572 Convert to block body action improperly works with suppress annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34572

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToBlockBodyIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToBlockBodyIntention.kt
@@ -13,9 +13,13 @@ import org.jetbrains.kotlin.idea.caches.resolve.analyze
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
 import org.jetbrains.kotlin.idea.core.setType
 import org.jetbrains.kotlin.idea.formatter.adjustLineIndent
+import org.jetbrains.kotlin.idea.refactoring.fqName.fqName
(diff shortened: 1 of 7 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertToBlockBodyIntention.kt
```diff
@@ -56,10 +60,24 @@ class ConvertToBlockBodyIntention : SelfTargetingIntention<KtDeclarationWithBody
...
-                val needReturn = returnsValue &&
-                        (bodyType == null || (!bodyType.isUnit() && !bodyType.isNothing()))
-                val statement = if (needReturn || unitWhenAsResult) factory.createExpressionByPattern("return $0", body) else body
-                return factory.createSingleStatementBlock(statement)
+                val needReturn = returnsValue && (bodyType == null || (!bodyType.isUnit() && !bodyType.isNothing()))
+                return if (needReturn || unitWhenAsResult) {
+                    val annotatedExpr = body as? KtAnnotatedExpression
+                    val suppressAnnotation = annotatedExpr?.annotationEntries?.singleOrNull()?.takeIf {
+                        val context = annotatedExpr.analyze(BodyResolveMode.PARTIAL)
+                        context[BindingContext.ANNOTATION, it]?.type?.fqName == FqName("kotlin.Suppress")
```
darthorimar, line 68: The problem happens not only for `suppress` annotation but for other ones too, so there is no need to filter them out

t-kameyama:
Fixed in 304d666

darthorimar:
Thanks!

PR #2970, opened 2020-01-09
[FIR]: fix Fir2Ir translation of this references in instance methods.

juan-chen:
Currently FirThisReceiverExpression of instance methods are translated to references of the class' thisReceiver, not the method's dispatch receiver, which causes problems with IrFrameMap::typeOf, as the class' thisReceiver is not in the typeMap.

This commit translates non-qualified "this" references of instance methods to references of the methods' dispatch receiver.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -867,6 +869,20 @@ class Fir2IrVisitor(
     }
 
     override fun visitThisReceiverExpression(thisReceiverExpression: FirThisReceiverExpression, data: Any?): IrElement {
+        val calleeReference = thisReceiverExpression.calleeReference
+        if (calleeReference.labelName == null && calleeReference.boundSymbol is FirRegularClassSymbol) {
+            val dispatchReceiver = this.functionStack.lastOrNull()?.dispatchReceiverParameter
+            if (dispatchReceiver != null) {
+                // Use the dispatch receiver of the containing function
+                return IrGetValueImpl(
```
mglukhikh: I will change it to `thisReceiverExpression.convertWithOffset` usage

mglukhikh:
Also, you forgot to update `Fir2IrTextTestGenerated` results, I will do it.

mglukhikh:
Merged manually. See https://github.com/JetBrains/kotlin/commit/573188bdc4ab1ad92f00d065a03d887106dcf855#diff-a565eac134035fb58271b76c8a11d4a1R885.

PR #2971, opened 2020-01-10
JVM_IR: Put continuation parameter before default mask and handler.

madsager:
This is important for calls using reflection as the reflect
library assumes this ordering of arguments.

It would be nice if this could be handled in the lowerings.
Currently AddContinuationLowering is after
DefaultArgumentStubGenerator. If we could add the continuation
first and then do default stub generation maybe we could avoid
the reshuffling introduced in coroutine codegen in this change.


madsager:
@ilmirus This is the most local fix that I can come up with but I'm not happy with the reshuffling taking place here. We should see if we can do this more directly in the lowerings before coroutine codegen. However, that will be a larger change which I don't think I should do at this point where you are making larger changes in this part of the code.

PR #2972, opened 2020-01-10
JVM IR: Generate bridges for fake override implementation methods

sfs:


udalov:
I've tried to compile the module `ir.tree` with this change, and it seemed to help with missing bridges for `IrSymbolImpl.descriptor`, but didn't help with `IrElementImpl.transform`. Let me extract another small sample from that, maybe it's a completely different problem.

udalov:
Here's a simplified test case for the `IrElementImpl.transform` problem:
```
// IGNORE_BACKEND_FIR: JVM_IR

interface A {
    val result: Any get() = "Fail"
}

interface B : A {
    override val result: String get() = "OK"
}

abstract class AImpl : A

class BImpl : AImpl(), B

fun box(): String =
    (BImpl() as A).result.toString()
```

Looks quite related to the problem fixed in your changes, so maybe it's possible to handle this case as well?

sfs:
This looks like a different bug. `InterfaceDelegationLowering` fails to put in the default impl delegation in class `BImpl`. Let me see if it's a simple fix.

sfs:
There was an imprecise check in `InterfaceDelegationLowering`. The phase has to generate a delegation unless one already exists in a superclass and it did not check whether the delegation in the superclass redirects to the same method.

In your example, `AImpl.getResult` redirects to `A.DefaultImpls.getResult` and we should still generate a method `BImpl.getResult` which redirects to `B.DefaultImpls.getResult`.

sfs:
Rebased on master with a small fix in `InheritedDefaultMethodsOnClassesLowering`.

gogabr:
Thank you!

PR #2976, opened 2020-01-13
JVM_IR: Fix bug in suspend function with many default arguments.

madsager:


madsager:
@pyos pointed out that the fix that puts the continuation before default argument mask and handler did not account for the fact that there can be more masks. It does now. :)

@ilmirus can you have a look?

PR #2977, opened 2020-01-13
KT-35829 "Redundant suspend modifier" is missed in recursive call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35829


darthorimar:
Thank you!

PR #2978, opened 2020-01-13
Add a basic IR extension to the Android plugin

pyos:
Only supports uncached findViewById/findFragmentById for now; other features are experimental and the cache only affects performance, so this should probably be fine for testing Android apps with -Xuse-ir.


yanex:
Thanks for your contribution! Merged into master manually.

PR #2979, opened 2020-01-13
JVM IR: Fix generation of parameterless default constructors

sfs:
The JVM backend generates parameterless default constructors in some cases when we have a primary constructor with default values for all arguments. This PR (mostly) implements the same rules 
for the IR backend and fixes the visibility of parameterless default constructors. It also adds a couple of tests for all the special cases.

One difference between the JVM and JVM IR backends is in the handling of parameterless main methods for constructors taking inline class arguments (see `compiler/testData/codegen/bytecodeText/constructors/inlineArgumentPrimaryDefaults.kt`). I think the behavior of the JVM backend is wrong here and I've reported this as [KT-35902](https://youtrack.jetbrains.com/issue/KT-35902).

Another small difference is in the flags used for static replacement constructors of inline classes (`compiler/testData/codegen/bytecodeText/constructors/inlinePrimaryDefaults.kt`). Again, I'm not sure if we shouldn't change this in the JVM backend instead.

For now I've added tests with `JVM(_IR)_TEMPLATES` to document the current behavior.

### DIFF
--- a/compiler/testData/codegen/bytecodeText/constructors/innerPrimaryDefaults.kt
+++ b/compiler/testData/codegen/bytecodeText/constructors/innerPrimaryDefaults.kt
@@ -1,0 +1,11 @@
+class A(val s: String) {
+    inner class B(val x: Int = 0)
+}
+
+// @A.class
+// 1 public <init>\(Ljava/lang/String;\)V
+
+// @A$B.class
+// 0 <init>\(\)V
+// 1 public <init>\(LA;I\)V
+// 1 public synthetic <init>\(LA;IILkotlin/jvm/internal/DefaultConstructorMarker;\)V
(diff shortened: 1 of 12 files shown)


compiler/testData/codegen/bytecodeText/constructors/innerPrimaryDefaults.kt
```diff
@@ -0,0 +1,11 @@
+class A(val s: String) {
+    inner class B(val x: Int = 0)
+}
+
+// @A.class
+// 1 public <init>\(Ljava/lang/String;\)V
+
+// @A$B.class
+// 0 <init>\(\)V
```
udalov: I'll amend your changes with adding `// 0 <init>\(LA;\)V` here. This is needed to check the `irClass.isInner` condition in `JvmDefaultConstructorLowering`, since the lowering is run before `InnerClassesLowering` where an additional parameter for the outer class is added.

udalov:
Merged manually in 8e074828625. Thank you!

compiler/testData/codegen/bytecodeText/constructors/innerPrimaryDefaults.kt
sfs: Thank you! I was sloppy in writing these tests and missed this part.

PR #2982, opened 2020-01-14
JVM_IR: actually forward the parameters to $suspendImpl

pyos:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
```diff
@@ -529,15 +529,10 @@ private class AddContinuationLowering(private val context: JvmBackendContext) :
                     it.putValueArgument(i++, irGet(irFunction.dispatchReceiverParameter!!))
                 }
                 if (irFunction.extensionReceiverParameter != null) {
-                    val defaultValueForParameter = irFunction.extensionReceiverParameter!!.type.defaultValue(
-                        UNDEFINED_OFFSET, UNDEFINED_OFFSET, backendContext
```
madsager: Total nit, but these are probably the only usages of backendContext defined on line 512, so that can be removed.

pyos: You're right, thanks!

PR #2984, opened 2020-01-15
KT-22368 "Convert to block body" intention incorrectly formats closing brace

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22368


darthorimar:
Thank you!

PR #2985, opened 2020-01-15
Initial implementation of klib components

alexander-gorshenev:


alexander-gorshenev:
The native counterpart is https://github.com/JetBrains/kotlin-native/pull/3758

PR #2986, opened 2020-01-15
[JVM IR] Fix `lambdaInLocalFunction.kt`

punzki:
@udalov Rather than checking if the enclosing class/method is consistent with at least one representation, I decided to use a helper method that returns the target backend under test that we can check against.

### DIFF
--- a/compiler/tests-common/tests/org/jetbrains/kotlin/TestHelperGenerator.kt
+++ b/compiler/tests-common/tests/org/jetbrains/kotlin/TestHelperGenerator.kt
@@ -6,8 +6,10 @@
 package org.jetbrains.kotlin
 
 import org.jetbrains.kotlin.resolve.DescriptorUtils
+import org.jetbrains.kotlin.test.TargetBackend
 
-fun createTextForHelpers(isReleaseCoroutines: Boolean, checkStateMachine: Boolean, checkTailCallOptimization: Boolean): String {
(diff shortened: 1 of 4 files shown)


compiler/tests-common/tests/org/jetbrains/kotlin/TestHelperGenerator.kt
```diff
@@ -188,3 +190,11 @@ fun createTextForHelpers(isReleaseCoroutines: Boolean, checkStateMachine: Boolea
             |${if (checkTailCallOptimization) checkTailCallOptimizationString else ""}
         """.trimMargin()
 }
+
+// Add the directive `// WITH_HELPERS` to use these helpers in codegen tests (see CodegenTestCase.java).
+fun createTextForCodegenTestHelpers(backend: TargetBackend) =
+    """
+        |package helpers
+        |
+        |fun getTargetBackend() = "${backend.name}"
```
udalov: Maybe `fun isIR(): Boolean = "${backend.isIR}"` instead? Seems less error-prone than to compare strings, where you can have a typo...

punzki: Ok, changed. I suppose if we need to check for specific backends in the future, we can add the method back.

udalov:
Thank you! Merged manually in 7329a1641a2.

PR #2987, opened 2020-01-15
JVM_IR: mark the exception local as live only after the store

pyos:
Otherwise, should the local coincide with another one, the incorrect range causes D8 to generate invalid bytecode in debug mode.


max-kammerer:
Thank you!

PR #2988, opened 2020-01-15
JVM_IR: Support suspend lambdas with many arguments.

madsager:


madsager:
This is becoming more and more hackish. This is basically replicating a little bit of FunctionNVarargBridgeLowering in CoroutineCodegen. Seems like another case where maybe over time we should attempt to do more in AddContinuationLowering (add the continuation parameter and rewrite calls) and do it earlier in the pipeline. But this seems like the simplest and most local fix right now.

ilmirus:
Well, this was the idea from the beginning: AddContinuaitonLowering was going to not only add continuation classes, but also continuation parameters. However, this the lowerings are run file by file, when we encounter an inline suspend function from another file, it will be unlowered and we would have link to unlowered suspend function view in suspendFunctionToView map in the context stored. Thus, I moved the continuation parameter adding to the end of the pipeline - codegen, which is run, when all files in a module are lowered.
But, I still want to move the continuation addition to AddContinuationLowering. One way to do this is to turn in into a module lowering, i.e. run all file lowerings file by file before it, then run AddContinuationLowering on the module, then resume file lowerings until codegen. This way, there would be no unlowered inline function views. It would require changing the pipeline and the lowerings would look as pretty as they look now. However, changing some other lowering into module lowerings would simplify the code of them a bit. In addition, in would allow us to move some mappings from the context to the lowering itself, further simplifying the code.  
Fortunately, there is another solution, which I came up with today. You see, when we call an inline function from another file, we do not need its body. Thus, we can create a stub without a body and when we lower the function, we fix the symbol to point to lowered function instead of stub. This way, we do not need to change the scheme and still move view creation from codegen to the lowering. BTW, @udalov, I wonder, do we really the stub in this case? Cannot we just create a symbol and fix it when we lower inline function. In any case, I will try moving the view creation to AddContinuationLowering soon. After this, I think, we can remove the hacks. Until then, I postpone the review. If I encounter some issues, I will try changing the scheme. Looks like this is finally the time to do this.

sfs:
@ilmirus Thanks for the explanation!

The second option sounds very similar to what we are already doing for inline classes in [MemoizedInlineClassReplacements](https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt). One thing to be aware of is that even the stubs may contain unlowered expressions in the form of default arguments. The easiest way around this is to not copy default arguments, but rather to point to the originals first and fix the value parameter references / parents when you lower the body.

Another thing to be aware of is that you need to use a different version of `WrappedDescriptor` for inline functions. In particular, you have to use `buildFunWithDescriptorForInlining` when creating the stub. Apart from that you just have to make sure to copy everything else over to the stub, including correspondingPropertySymbols, annotations, and metadata.

ilmirus:
@madsager Can you, please, rebase the PR to the master, since I finished with moving views creating to the lowering?

madsager:
Rebased. :)

ilmirus:
I manually rebased and pushed to master.

PR #2989, opened 2020-01-15
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,5 +1,5 @@
 name: "Validate Gradle Wrapper"
-on: [push]
+on: [push, pull_request]
 
 jobs:
   validation:
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,12 @@
+name: "Validate Gradle Wrapper"
+on: [push]
+
+jobs:
+  validation:
+    name: "Validation"
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,12 @@
+name: "Validate Gradle Wrapper"
+on: [push]
```
ZacSweers: Without `pull_request`, this won't actually run for external contributors

JLLeitschuh: Are you sure this is accurate?

ZacSweers: Pretty sure. Push only applies to branches on the project, and your PR is not a branch on the project. We had to fix this in our autodispose project as well when enabling actions

JLLeitschuh: You were right. Should be fixed.
Thanks!

4u7:
Thank you!

4u7:
Looks like actions are disabled for our repo. Will check if we can enable it.

JLLeitschuh:
😢 https://github.com/JetBrains/kotlin/commit/9b7e904390ce4eff2a96cc6b529e793fa4440b98

4u7:
@JLLeitschuh  Unfortunately we can't enable workflows for JetBrains organisation and without it enabled every push fails validation and sends tons of emails. Sorry!

JLLeitschuh:
No worries. I ran into this with Spring (Pivotal) and Square already.

PR #2991, opened 2020-01-16
IR: close over the set of captured type parameters

pyos:


pyos:
@gogabr BTW, the implementation still seems incomplete because `seeType` is only called for expression return types. This means for this code:
```
fun <T, V : T> f(x: V): Int {
    fun g(y: V) = 2
    fun h(y: V) = y
    return g(x)
}
```
these methods are generated:
```
private static final int f$g(V);
private static final <V extends T, T extends java.lang.Object> V f$h(V);
```
The first one is obviously incorrect.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -319,36 +313,23 @@ fun IrType.remapTypeParameters(
...
-                                    it.type.remapTypeParameters(source, target, srcToDstParameterMap),
-                                    it.variance
-                                )
-                                else -> it
-                            }
-                        },
-                        annotations
+            val oldClassifier = classifier
+            val newClassifier = when {
+                oldClassifier is IrTypeParameterSymbol && oldClassifier.owner.parent == source ->
```
gogabr: This condition is wrong; see the test `typeParametersInLocalFunction.kt`, where type parameter `C` is captured by `inner2()`, but will not be remapped by your condition.
(Unfortunately, this does not lead to a visible test failure).

gogabr:
> @gogabr BTW, the implementation still seems incomplete because `seeType` is only called for expression return types. 
Yes, I have a fix ready, hope to merge today.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
pyos: I'll revert this change for now, this code should probably be refactored separately since such errors are not clearly visible.

gogabr:
Thank you!

PR #2992, opened 2020-01-16
[JVM_IR] Propagate interface type parameters to DefaultImpls

kandersen:
This PR ensures ultimately correct generation of generic signatures for interface methods that are moved to DefaultImpls during compilation. The IR backend had until now not handled them, but they were otherwise not used during compilation, causing no errors in their absence.

Here we copy the parameters from the interface and adjust all rewritten call site that now must take into account these added parameters. Here it becomes less straightforward. To illustrate, suppose we have the following:

```
interface A<T,U,V> {
  fun foo(t: T, u: U): V? = null
}

interface B<T, V> : A<T, Int, V>

class C: B<String, Boolean> {
  override fun foo(t: String, u: Int): Boolean? {
    return super.foo(t, u)
  }
}
```
`A.foo` is moved to `A$DefaultImpls`, and the parameters to it are pushed to the front of foo:
```
class A$DefaultImpls {
  fun <T,U,V> foo(t: T, u: U): V? = null
}
```
and the calll to `super.foo` in `C` is rewritten to call `A$DefaultImpl.foo` (this is a mistake, however, orthogonal to the issue at hand):
```
class C: B<String, Boolean> {
  override fun foo(t: String, u: Int): Boolean? {
    return A$DefaultImpls.foo<?X, ?Y, ?Z>(t, u)
  }
}
```
The question is, how should we instantiate the type arguments to foo?

@sfs pointed out it that problem essentially reduces to posing and solving the type constaint `C <: A<?X, ?Y, ?Z>`, but that machinery is not immediately at hand.


kandersen:
@madsager @sfs I wanted to break this PR in half and get some eyes on it as it, at the moment, resolves the outstanding IR test (produce valid signatures on DefaultImpls) but reveals a good piece of work that is as of yet untested in both old and new compiler.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InheritedDefaultMethodsOnClassesLowering.kt
```diff
@@ -180,7 +212,28 @@ private class InterfaceDefaultCallsLowering(val context: JvmBackendContext) : Ir
...
+                expression,
+                expression.symbol.owner,
+                this.symbol.owner,
+                receiversAsArguments = true,
+                argumentsAsReceivers = false
+            )
+            //TODO: See Note [Interface Parameters]
+            var offset = 0
+            expression.symbol.owner.parentAsClass.typeParameters.forEach {
+                putTypeArgument(offset++, context.irBuiltIns.anyNType)
```
sfs: Technically, `Any?` might not be a valid instantiation for a given type parameter, since the parameter could have bounds. Could you instantiate it with `it.defaultType` instead?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -243,7 +250,31 @@ internal class InterfaceLowering(val context: JvmBackendContext) : IrElementTran
         val newFunction = removedFunctions[expression.symbol]?.owner
         return super.visitCall(
             if (newFunction != null) {
-                irCall(expression, newFunction, receiversAsArguments = true)
+                IrCallImpl(
+                    expression.startOffset,
+                    expression.endOffset,
+                    expression.type, // TODO: Here we copy the old type, but really it should be [newFunction.returnType] under an appropriate substitution
```
sfs: Apart from potential references to the type parameters of the interface instead of the default method, wouldn't these be the same type?

kandersen: You are right! And the frontend would have done the substitution as we saw in #3009, I guess :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InheritedDefaultMethodsOnClassesLowering.kt
kandersen: You are right, I actually think two of these cases can be done correctly! Rewriting, pushing a revision.

kandersen:
@udalov I can't cook up an example where the made-up type arguments filled in for the context parameters copied to DefaultImpls can show up in or otherwise influence the resulting byte code. Perhaps you can think of something? Otherwise I propose we just park this for now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InheritedDefaultMethodsOnClassesLowering.kt
```diff
@@ -181,7 +215,32 @@ private class InterfaceDefaultCallsLowering(val context: JvmBackendContext) : Ir
...
+            origin = expression.origin
+        ).apply {
+            copyValueArgumentsFrom(
+                expression,
+                expression.symbol.owner,
+                this.symbol.owner,
+                receiversAsArguments = true,
+                argumentsAsReceivers = false
+            )
+            //TODO: See Note [Interface Parameters]
```
udalov: Why not just move the comment here?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InheritedDefaultMethodsOnClassesLowering.kt
```diff
@@ -111,8 +112,13 @@ private class InheritedDefaultMethodsOnClassesLowering(val context: JvmBackendCo
             irFunction.body = irBlockBody {
                 +irReturn(
                     irCall(defaultImplFun.symbol, irFunction.returnType).apply {
+                        interfaceImplementation.parentAsClass.typeParameters.forEachIndexed { index, irTypeParameter ->
+                            //TODO: See Note [Interface Parameters]
```
udalov: Minor, but could you extract a simple method that would return `context.irBuiltIns.anyNType` and put the comment there instead, so that we have only one TODO instead of three?

kandersen: Yes, sure! Will rebase, as well.

kandersen: Done. Thanks!

kandersen:
@udalov Missed TODOs in `InterfaceLowering.kt`. Should be good now :)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -589,22 +608,74 @@ fun createStaticFunctionWithReceivers(
...
+    contextParameters: List<IrTypeParameter>,
+    existingParameters: Collection<IrTypeParameter>
+): List<IrTypeParameter> {
+    val newParameters = mutableListOf<IrTypeParameter>()
+
+    val existingNames =
+        (contextParameters.map { it.name.asString() } + existingParameters.map { it.name.asString() }).toMutableSet()
+
+    contextParameters.forEach { contextType ->
+        val newName = if (existingParameters.any { it.name.asString() == contextType.name.asString() }) {
```
gogabr: Maybe `new Name = contextType.name.asString() + "$" + (contextType.parent as IrDeclarationWithName).name.asString()`?
This does leave room for conflicts, but in most cases reads better IMO.

kandersen: Possibly better, but would require a renaming in the old backend as well to keep them in agreement. Possibly this is not an issue...?

kandersen: @gogabr I take the resolution to mean we stay with the current approach of matching the old backend?

gogabr: Yes. Sorry, I should have indicated this explicitly.

kandersen: No problem! Thank you :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InheritedDefaultMethodsOnClassesLowering.kt
```diff
@@ -118,11 +126,38 @@ private class InterfaceSuperCallsLowering(val context: JvmBackendContext) : IrEl
...
+        ).apply {
+            copyValueArgumentsFrom(
+                expression,
+                expression.symbol.owner,
+                this.symbol.owner,
+                receiversAsArguments = true,
+                argumentsAsReceivers = false
+            )
+
+            var offset = 0
```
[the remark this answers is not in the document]
kandersen: Done, thank you!

gogabr:
Thank you!

PR #2994, opened 2020-01-16
IR: Use name \$\$delegate_n for the field for interface delegation.

madsager:
Without this change, these fields could have names such as
`<no name provided>$SuperClass$delegate` for something like
`object SuperClass by I { }`.


compiler/testData/codegen/box/delegation/delegationToVal.kt
```diff
@@ -1,4 +1,5 @@
 // IGNORE_BACKEND_FIR: JVM_IR
+// IGNORE_BACKEND: JVM_IR
```
madsager: This test should not have passed. It is checking that no delegate field is created for interface delegation. It is doing so based on the name of the field, and therefore it now correctly fails because we do not have that optimization.

madsager:
I should relabel this. Have done too many JVM_IR changes recently. This is changing psi2ir so it is really an IR change.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrDelegateDescriptor.kt
```diff
@@ -104,25 +104,19 @@ class IrPropertyDelegateDescriptorImpl(
...
         delegateType
     ),
     IrImplementingDelegateDescriptor
 
 internal fun getDelegateName(name: Name): Name =
     Name.identifier(name.asString() + "\$delegate")
 
-internal fun getDelegateName(classDescriptor: ClassDescriptor, superType: KotlinType): Name =
-    Name.identifier(
-        classDescriptor.name.asString() + "\$" +
```
punzki: When I was working on this, I simply remove the containing class name, which could be unnamed. It was providing no useful information anyway since it's always the same class. The supertype name at least provided some information. Adding the index is still safer though since you could have multiple implemented interfaces with the same name.

madsager:
Friendly ping. Does this make sense to you @dnpetrov?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrDelegateDescriptor.kt
madsager: Thanks @punzki. Yes, I definitely want to add the number to disambiguate. I'm not sure we gain much from using different names than the current backend for these. The current backend uses $$delegate_n which is what this change ends up doing as well. @dnpetrov would you prefer that we keep the supertype name here if it exists?

dnpetrov:
Merged manually.

PR #2995, opened 2020-01-16
JVM_IR: Avoid using parent name in name of captures.

madsager:
If parent is an anonymous object that could lead to the use
of names such as `$this$no name provided` which is invalid
in dex files.


udalov:
Ref: #2655

max-kammerer:
Thank you!

udalov:
@madsager @max-kammerer Looks like the test `bytecodeText/mangling/parentheses.kt` fails now on the old backend, since it uses `this` as a local variable name instead of `$this`:
https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_AggregateBranch/2724178?buildTab=tests&pager.currentPage=1&filterText=testParentheses

I'm not entirely sure why do we use the name `$this`, and not `this`, in the new backend though

madsager:
Oh shoot, I ran the wrong tests when checking this. I ran blackbox and not bytecode tests. Will upload a change to add the templates per backend back. Then we can look at $this vs this.

PR #2996, opened 2020-01-16
Gradle plugin configuration avoidance improvements

LouisCAD:
In addition to further refactoring for configuration avoidance, I replaced a few deprecated usages in the way.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
@@ -240,7 +240,7 @@ internal open class GradleCompilerRunner(protected val task: Task) {
                 ?: return null
             val sourceSet = javaConvention.sourceSets.findByName(sourceSetName) ?: return null
             val jarTask = project.tasks.findByName(sourceSet.jarTaskName) as? Jar
-            return jarTask?.archivePath
+            return jarTask?.archiveFile?.get()?.asFile
         }
 
         @Synchronized
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
@@ -307,7 +307,7 @@ internal class Kotlin2JsSourceSetProcessor(
 
     private fun registerCleanSourceMapTask() {
         val taskName = kotlinCompilation.composeName("clean", "sourceMap")
-        val registerTask = registerTask(project, taskName, Delete::class.java) {
+        val registerTask = project.registerTask<Delete>(taskName) {
             it.onlyIf { kotlinTask.get().kotlinOptions.sourceMap }
             it.delete(object : Closure<String>(this) {
                 override fun call(): String? = (kotlinTask.get().property("outputFile") as File).canonicalPath + ".map"
@@ -445,7 +445,7 @@ internal abstract class AbstractKotlinPlugin(
                 return
             }
             val inspectTask =
-                registerTask(project, "inspectClassesForKotlinIC", InspectClassesForMultiModuleIC::class.java) {
+                project.registerTask<InspectClassesForMultiModuleIC>("inspectClassesForKotlinIC") {
                     it.sourceSetName = SourceSet.MAIN_SOURCE_SET_NAME
                     it.archivePath.set(project.provider { jarTask.archivePathCompatible.canonicalPath })
                     it.archiveName.set(project.provider { jarTask.archiveNameCompatible })
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
@@ -307,13 +307,15 @@ internal class Kotlin2JsSourceSetProcessor(
 
     private fun registerCleanSourceMapTask() {
         val taskName = kotlinCompilation.composeName("clean", "sourceMap")
-        val registerTask = project.registerTask<Delete>(taskName) {
+        project.registerTask<Delete>(taskName) {
             it.onlyIf { kotlinTask.get().kotlinOptions.sourceMap }
             it.delete(object : Closure<String>(this) {
                 override fun call(): String? = (kotlinTask.get().property("outputFile") as File).canonicalPath + ".map"
             })
         }
-        project.tasks.findByName("clean")?.dependsOn(taskName)
+        project.locateTaskByName("clean")?.configure { cleanTask ->
+            cleanTask.dependsOn(taskName)
+        }
     }
 }
 
@@ -327,10 +329,12 @@ internal class KotlinCommonSourceSetProcessor(
     kotlinCompilation = compilation
 ) {
     override fun doTargetSpecificProcessing() {
-        project.tasks.findByName(kotlinCompilation.compileAllTaskName)!!.dependsOn(kotlinTask)
-        // can be missing (e.g. in case of tests)
+        project.tasks.named(kotlinCompilation.compileAllTaskName).configure { it.dependsOn(kotlinTask) }
         if (kotlinCompilation.compilationName == KotlinCompilation.MAIN_COMPILATION_NAME) {
-            project.tasks.findByName(kotlinCompilation.target.artifactsTaskName)?.dependsOn(kotlinTask)
+            // can be missing (e.g. in case of tests)
+            project.locateTaskByName(kotlinCompilation.target.artifactsTaskName)?.configure {
+                it.dependsOn(kotlinTask)
+            }
         }
 
         project.whenEvaluated {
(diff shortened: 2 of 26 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
```diff
@@ -240,7 +240,7 @@ internal open class GradleCompilerRunner(protected val task: Task) {
                 ?: return null
             val sourceSet = javaConvention.sourceSets.findByName(sourceSetName) ?: return null
             val jarTask = project.tasks.findByName(sourceSet.jarTaskName) as? Jar
```
stephanenicolas: This will cause the tasks to be realized. You can use named, and also precise the type if the lowest gradle version is bumped to 5.1.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -307,13 +307,15 @@ internal class Kotlin2JsSourceSetProcessor(
 
     private fun registerCleanSourceMapTask() {
         val taskName = kotlinCompilation.composeName("clean", "sourceMap")
-        val registerTask = registerTask(project, taskName, Delete::class.java) {
+        project.registerTask<Delete>(taskName) {
             it.onlyIf { kotlinTask.get().kotlinOptions.sourceMap }
```
stephanenicolas: I am wondering here when the `onlyIf` statement is gonna be evaluated as calling `get()` will realize the task.
TBH, it would be nice to have someone from Gradle having a look at it.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/TasksProvider.kt
```diff
@@ -55,6 +53,12 @@ internal fun <T : Task> Project.registerTask(
 }
 
 
+internal fun Project.locateTaskByName(name: String): TaskProvider<*>? = try {
```
stephanenicolas: nice, but maybe `namedOrNull` would be a better name. It would show clearly that we use the TCA API.

stephanenicolas: Can we also rename `locateTask` on line 61-66 to `namedOrNull`. We loose the notion of overload by changing the method name.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -433,25 +437,24 @@ internal abstract class AbstractKotlinPlugin(
         }
 
         private fun configureClassInspectionForIC(project: Project) {
-            val classesTask = project.tasks.findByName(JavaPlugin.CLASSES_TASK_NAME)
-            val jarTask = project.tasks.findByName(JavaPlugin.JAR_TASK_NAME)
+            val classesTask = project.locateTaskByName(JavaPlugin.CLASSES_TASK_NAME)
+            val jarTask = project.locateTask<Jar>(JavaPlugin.JAR_TASK_NAME)
```
stephanenicolas: Can we change all the variable/property names that refer to providers to use the suffix `Provider` (e.g. `jarTaskProvider`). It will help maintenance and the API will be more clear.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -433,25 +437,24 @@ internal abstract class AbstractKotlinPlugin(
...
 
-            if (classesTask == null || jarTask !is Jar) {
+            if (classesTask == null || jarTask == null) {
                 project.logger.info(
-                    "Could not configure class inspection task " +
-                            "(classes task = ${classesTask?.javaClass?.canonicalName}, " +
-                            "jar task = ${classesTask?.javaClass?.canonicalName}"
+                    "Could not configure class inspection task because at least one of the 2 following tasks " +
+                            "have not been found: ${JavaPlugin.CLASSES_TASK_NAME} & ${JavaPlugin.JAR_TASK_NAME}."
```
stephanenicolas: I  am wondering if we should not invest a little code here to be more precise about which task(s) is/are null.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -433,25 +437,24 @@ internal abstract class AbstractKotlinPlugin(
...
                 )
                 return
             }
             val inspectTask =
```
stephanenicolas: rename to `inspectTaskProvider`.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -433,25 +437,24 @@ internal abstract class AbstractKotlinPlugin(
...
-                registerTask(project, "inspectClassesForKotlinIC", InspectClassesForMultiModuleIC::class.java) {
+                project.registerTask<InspectClassesForMultiModuleIC>("inspectClassesForKotlinIC") {
                     it.sourceSetName = SourceSet.MAIN_SOURCE_SET_NAME
-                    it.archivePath.set(project.provider { jarTask.archivePathCompatible.canonicalPath })
-                    it.archiveName.set(project.provider { jarTask.archiveNameCompatible })
+                    it.archivePath.set(project.provider { jarTask.get().archivePathCompatible.canonicalPath })
+                    it.archiveName.set(project.provider { jarTask.get().archiveNameCompatible })
```
stephanenicolas: I know it might be too big a change but here the configuration of `inspectTask` will trigger the creation of `jarTask`, but the dependency between the 2 tasks is the other way around (see line 457). Ideally it should be the `inspectTask` that would contain the data and `jarTask` can realize it during its configuration because it will know for sure that `inspectTask` will be realized. An alternative is to have the data being hold independently of tasks.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt
```diff
@@ -948,7 +951,7 @@ abstract class AbstractAndroidProjectHandler(private val kotlinConfigurationTool
 
     private val BaseVariant.javaTaskProvider: TaskProvider<out AbstractCompile>
         get() = getJavaTaskProvider()
-            // for older versions, get the task instance and wrap it into a TaskProvider
+        // for older versions, get the task instance and wrap it into a TaskProvider
             ?: getJavaTask(this).let { it.project.tasks.withType(AbstractCompile::class.java).named(it.name) }
```
stephanenicolas: if we use gradle 5.1 apis, this can be shortened to `named(Class, String)`. Or we need to keep using this "old" api because this case is a backward compatibility case ?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinJvmTarget.kt
```diff
@@ -61,11 +61,11 @@ open class KotlinJvmTarget @Inject constructor(
 
         javaPluginConvention.sourceSets.all { javaSourceSet ->
             val compilation = compilations.getByName(javaSourceSet.name)
-            val compileJavaTask = project.tasks.getByName(javaSourceSet.compileJavaTaskName) as AbstractCompile
+            val compileJavaTask = project.locateTask<AbstractCompile>(javaSourceSet.compileJavaTaskName)!!
```
stephanenicolas: sometimes we use `named`, sometimes we use `locateTask!!`. It would be better to stick to one method.

stephanenicolas: Please rename to `compileJavaTaskProvider`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinJvmTargetConfigurator.kt
```diff
@@ -49,7 +50,9 @@ class KotlinJvmTargetConfigurator(kotlinPluginVersion: String) :
     ): KotlinJvmTestRun = KotlinJvmTestRun(name, target).apply {
         val testTaskOrProvider = target.project.registerTask<KotlinJvmTest>(testTaskName) { testTask ->
```
stephanenicolas: `testTaskProvider`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/KotlinNativeTargetConfigurator.kt
```diff
@@ -134,21 +134,21 @@ open class KotlinNativeTargetConfigurator<T : KotlinNativeTarget>(
...
-            onlyIf { binary.outputFile.exists() }
-            dependsOn(binary.linkTaskName)
+            task.onlyIf { binary.outputFile.exists() }
+            task.dependsOn(binary.linkTaskName)
         }
     }
 
-    private fun Project.createKlibCompilationTask(compilation: KotlinNativeCompilation) {
+    private fun Project.registerKlibCompilationTask(compilation: KotlinNativeCompilation) {
         val compileTask = registerTask<KotlinNativeCompile>(
```
stephanenicolas: `compileTaskProvider`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/tasks/TasksProvider.kt
stephanenicolas: We could even define here a extensions for `project.tasks.named` to `Project.named()` so that the API used for tasks all point to this file and make it easier to read.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/testing/internal/KotlinTestsRegistry.kt
```diff
@@ -95,12 +96,12 @@ class KotlinTestsRegistry(val project: Project, val allTestsTaskName: String = "
         return "clean" + taskName.capitalize()
     }
 
-    private val Project.cleanAllTestTask: Task
+    private val Project.cleanAllTestTask: TaskProvider<*>
```
stephanenicolas: `cleanAllTest(s)TaskProvider`, I am not sure about the `s`.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/deprecations.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+package org.jetbrains.kotlin.gradle.utils
+
+import org.gradle.api.logging.Logger
+
+const val deprecatedBecauseNoConfigAvoidanceUseProvider = "Using this brings performance issues. " +
+        "Use the provider instead to benefit from configuration avoidance."
```
stephanenicolas: Use the equivalent provider method to benefit from gradle task configuration avoidance.

stephanenicolas:
Thanks @LouisCAD , this will help. I would recommend to have someone from Gradle checking the PR too. 
Also, there are no integration tests, I believe we could verify that no tasks are created, for instance, we can call the `help` task on a kotlin project and check that the build output doesn't mention any task related to kapt or kotlin compilation being created / realized.

stephanenicolas:
And definitely, I would like to test a build when the PRs are merged. The real problem with TCA, and especially with this plugin is to make sure that there is not a single call to the obsolete APIs to trigger the task realization/creation. By running builds and analyzing build scans, we can make sure the goal is achieved.

PR #2997, opened 2020-01-17
JVM_IR: add more origins for suspend views

pyos:
Used for correctly inlining $default stubs, avoiding declaration clashes of $$forInline versions, and not crashing when trying to handle $suspendImpl of functions without source code.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
```diff
@@ -116,25 +116,31 @@ internal fun IrFunction.isInvokeOfSuspendCallableReference(): Boolean = isSuspen
...
     }
 
 internal fun IrFunction.shouldNotContainSuspendMarkers(): Boolean =
     isInvokeSuspendOfContinuation() || isKnownToBeTailCall()
 
+internal val IrDeclarationOrigin.isDefaultStub: Boolean
+    get() = this == JvmLoweredDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER_SUSPEND_VIEW ||
+            this == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER
+
+internal val IrDeclarationOrigin.isSuspendView: Boolean
```
pyos: This should also be used in KOTLIN-CR-3674 in MethodSignatureMapper.

PR #2998, opened 2020-01-17
JVM_IR: mark $suspendImpl of known-tail-call suspend functions

pyos:


pyos:
Merged into #2997

ilmirus:
Thank you for pointing out the issue with interface delegation of suspend functions! However, I would rather not add yet another origin, there are already a lot of them. After moving suspend function view creating to AddContinuationLowering (https://github.com/JetBrains/kotlin/pull/2988#issuecomment-574817145) I will remove view-related origins. Additionally, current tail-call optimization in old BE is capable of handling this case. Thus, there is only one issue remaining: absent PSI element of suspend function. Old BE uses PSI element of containing class, so, in this particular case a simpler change would be to retrieve PSI element of the class.
To summarize, I think, there is a simpler fix: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3744

pyos:
@ilmirus while I agree that there's a lot of origins, it seems somewhat weird to me to rely on bytecode-based tail call detection (and I also assume you generate a continuation class for the function only to throw it away later?) just to avoid adding another one. If you believe this is a better approach, though, then so be it.

ilmirus:
You are right, currently, we generate a continuation class to throw it away. However, with addition of TailCallOptimizationLowering this is no longer necessary. I am going to merge TailCallOptimizationLowering with AddContinuationLowering (or pass information about tail-call-ness between them somehow) in the near future and use the absence of continuation class as an indication, that we do not need to call CoroutineTransfromerMethodVisitor (which, to be honest, should be renamed to StateMachineBuilder).

PR #2999, opened 2020-01-17
KT-34640 Replace 'if' with 'when' leads to copy comment line above when from another if

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34640


vladimirdolzhenko:
LGTM. Thanks @t-kameyama for your contribution.

PR #3000, opened 2020-01-18
KT-35994 "Remove braces from 'if' statement": do not suggest when lambda is directly under 'if'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35994


darthorimar:
LGTM :+1:

PR #3001, opened 2020-01-18
KT-35995 "Remove single lambda parameter declaration": do not suggest when lambda is directly under 'if/when'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35995


darthorimar:
Merged manually in cc0cb30b534d816e13ae88032a1535b54b2d118b, thank you!

PR #3005, opened 2020-01-21
Fix test expectations.

madsager:
Do not check for occurrences of "this" on the current backend.
I accidentally unified the checking for the two backends
without checking that it worked (used the wrong test suite to
test).


madsager:
Not checking on the old backend since there are local variables named "this" as well and therefore I'm not sure the count of occurrences of "this" makes sense (currently 11).

udalov:
@madsager Thank you!

PR #3006, opened 2020-01-21
[JVM IR] Add tests to verify correct parameter reflection metadata for suspend functions

punzki:


compiler/testData/codegen/box/parametersMetadata/suspendFunction.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+}
+
+fun box(): String {
+    val clazz = A::class.java
+    val method = clazz.getDeclaredMethod("test", String::class.java, Continuation::class.java)
+    val parameters = method.getParameters()
+
+    if (parameters[0].modifiers != 0) return "wrong modifier on value parameter: ${parameters[0].modifiers}"
+    if (parameters[1].modifiers != 0) return "wrong modifier on Continuation parameter: ${parameters[1].modifiers}"
```
punzki: I only checked the modifier and not the name because it's `$completion` in IR and `continuation` in the old backend/non-IR. In the old backend, the name in the parameter descriptor (used in the metadata) is `continuation` yet in the local variable table it's `$completion`. In IR, it's consistently `$completion`.

Should we change the name in the parameter descriptor for the old backend?

max-kammerer: cc @ilmirus

punzki: @ilmirus @max-kammerer ping?

punzki: @ilmirus @max-kammerer Gentle ping again!

punzki:
@ilmirus The changes to `FunctionCodegen.kt` were no longer needed with https://github.com/JetBrains/kotlin/commit/fc70455877aa4b21696d84b41635c3eee2f55ddc, so I rebased and kept the tests.

compiler/testData/codegen/box/parametersMetadata/suspendFunction.kt
ilmirus: There is an issue with renaming continuation parameter in old BE with inliner, where it cannot inline non-suspend lambdas properly, will investigate and notify you, when this is done. I totally agree, that the parameter should be renamed to `$competion`, since it is `$completion` in LVT.
For now, I accept the PR.

PR #3007, opened 2020-01-22
KT-36051 IfThenToSafeAccessInspection: do not report if condition is SENSELESS_COMPARISON

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36051


darthorimar:
Thanks!

PR #3008, opened 2020-01-22
KT-34686 False positive "Constructor parameter is never used as a property" if property is used as a reference

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34686


mglukhikh:
Merged 👍

PR #3009, opened 2020-01-22
JVM IR: Remove type substitutions from ExpressionCodegen

sfs:
Instead, determine whether the return type of a call is Nothing, by looking at the type of the call expression.

CC @kandersen


max-kammerer:
Thank you for pr!

PR #3010, opened 2020-01-22
JVM IR: Support -Xno-call-assertions

sfs:
CC @kandersen

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmArgumentNullabilityAssertionsLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmArgumentNullabilityAssertionsLowering.kt
@@ -6,249 +6,98 @@
 package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.lower.SpecialBridgeMethods
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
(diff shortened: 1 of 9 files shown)


dnpetrov:
@sfs Please, take a look at `JvmArgumentNullabilityAssertionsLowering`.
It was supposed to remove assertions that are not required (including those that are not required because of call assertions disabled).
If it doesn't remove some particular assertions even with the corresponding flag enabled, it should most likely be fixed there.

sfs:
@dnpetrov Thanks for the pointer, I completely missed this lowering. Seems like the rules for null checks are more complicated than just checking a flag. I'll have a look at the lowering!

sfs:
@dnpetrov I've had a look at `JvmArgumentNullabilityAssertionsLowering`. The problem was that it did not remove all `NOT_NULL` assertions even if the `-Xno-call-assertions` flag was set.

I've refactored the lowering to ensure that we visit all type casts in the IR. The way I understand it, there has to be some special handling for dispatch and extension receiver parameters as well as calls to type erased bridges. Based on the code this special handling only extends to type casts which are immediately in the argument position.

I've also removed the descriptor usage from the code, since there is an equivalent facility for recognizing type adjusted special bridges in the IR.

dnpetrov:
@sfs Am I missing something, or you don't process results of `IrGetField`?

sfs:
@dnpetrov You're right, I missed the special case for receiver parameters on `IrGetField` and `IrSetField`! I've added these and also changed `testNullabilityAssertionOnDispatchReceiver` to properly check for this optimization. I've also rebased on master.

dnpetrov:
Ok, I'll run tests and merge manually.

dnpetrov:
Merged manually (cf3e4608f3b605711ebe88c73aa78d0604c2734b).

PR #3012, opened 2020-01-23
KT-36094 "Add missing component" intention adds extra comma to an empty destructuring declaration

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36094


darthorimar:
Thank you!

PR #3013, opened 2020-01-24
[JVM IR] Copy metadata in IrFieldBuilder

punzki:
This fixes an issue with `lateinit` properties where the metadata from the original field was not copied to the nullable field in `LateinitLowering`. I discovered this issue while building some project that uses Dagger with the JVM IR backend, and the issue with the missing metadata surfaced.

Special thanks to @sfs for observing the property in question was `lateinit` and for pointing me exactly which lowering was the culprit. 👍


udalov:
Looks good! I suppose we can now delete the similar assignment with a cast to `IrFieldImpl` in `JvmDeclarationFactory.getStaticBackingField`?

punzki:
> Looks good! I suppose we can now delete the similar assignment with a cast to `IrFieldImpl` in `JvmDeclarationFactory.getStaticBackingField`?

Yup, good catch! Deleted the line.

udalov:
Thank you!

PR #3015, opened 2020-01-24
Fix file components order

davidschreiber:
The path of the webpacked JavaScript file that should be included in the server JAR is constructed by taking the webpack tasks output directory, and appending the filename of the JavaScript module. Previously, the constructor arguments were flipped, leading to an invalid path.

Furthermore, I migrated away from the deprecated `outputPath` property, to the recommended `destinationDirectory` property.

I verified that the code was not working previously, but is working as expected with the suggested changes.

### DIFF
--- a/idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleWebMultiplatformModuleBuilder.kt
+++ b/idea/idea-gradle/src/org/jetbrains/kotlin/idea/configuration/KotlinGradleWebMultiplatformModuleBuilder.kt
@@ -264,7 +264,7 @@ class KotlinGradleWebMultiplatformModuleBuilder : KotlinGradleAbstractMultiplatf
 
             jvmJar {
                 dependsOn(jsBrowserProductionWebpack)
-                from(new File(jsBrowserProductionWebpack.entry.name, jsBrowserProductionWebpack.outputPath))
+                from(new File(jsBrowserProductionWebpack.destinationDirectory, jsBrowserProductionWebpack.entry.name))
             }
             
             task run(type: JavaExec, dependsOn: [jvmJar]) {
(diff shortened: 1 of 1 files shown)


ilgonmic:
Hi,

Thank you!
Did you use Groovy DSL?
Because for me, previous way works too (in Groovy DSL)
I guess, it is related with existence private constructor of File with signature `private File(String child, File parent)`, but access modifier doesn't matter for Groovy  (https://issues.apache.org/jira/browse/GROOVY-1875)

Anyway, I think these changes make code more consistent and straight-forward (w/o groovy specific)

But now I can see some other defect of this code, that is unrelated with your changes, but if it's easy to you, I would like to ask fix it too (if you can't, please say, I merge this PR and will fix defect on my own).

Now we use `jsBrowserProductionWebpack.entry.name` as name of output file, but it is incorrect in case of multi-module build, where our entry name includes root project prefix, but output of Webpack doesn't.
Usage of `outputFileName` property instead of `entry.name` is more extensible and more correct.

davidschreiber:
That makes a lot of sense! I was only noticing the mistake after migrating the script file from Groovy to the Kotlin DSL. I'll have another look at using `outputFileName` instead of `entry.name` 👍

ilgonmic:
Ohh, I see, this is actual for Kotlin DSL, but for Groovy DSL everything is OK
And we really need to swap arguments to help with migration from groovy to Kotlin DSL
You're right

davidschreiber:
@ilgonmic I opted for directly using the `KotlinWebpack`'s `outputFile` property, instead of constructing the file path myself. Since the property is public and encodes destination folder and filename, it should be good to use: https://github.com/JetBrains/kotlin/blob/075eea5a968119a784261883e7f8e988ef8aa9d9/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/js/webpack/KotlinWebpack.kt#L94-L96

Unless you see any issue with that change, this PR is ready from my side.

ilgonmic:
Thank you!

PR #3017, opened 2020-01-26
KT-32801 False positive "Call on collection type may be reduced" with mapNotNull, generic lambda block and new inference

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32801


t-kameyama:
4638a97bbcf402c01eb1ebdc96cb567697fb38f9

PR #3018, opened 2020-01-27
KT-34696 Wrong 'Redundant qualifier name' for 'MyEnum.values' usage

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34696
https://youtrack.jetbrains.com/issue/KT-37181


vladimirdolzhenko:
I have couple notes here:

-  using test data `notApplicableEnumCompanion.kt`: `value1?.let { <caret>QueryOperation.Companion.from("AA") }` still suggests `Remove redundant qualified name`, while ticket describes avoid dropping qualified name for ref method - this case looks similar 
- please use neutral package names rather `com.*`

t-kameyama:
PTAL

vladimirdolzhenko:
LGTM, Thank you

vladimirdolzhenko:
Merged

PR #3019, opened 2020-01-27
KT-16907 "Convert to lambda reference" intention is erroneously shown for suspending lambda parameters, producing bad code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-16907


mglukhikh:
Merged 👍

PR #3020, opened 2020-01-27
KT-27651 'Condition is always true' inspection should not be triggered when the condition has references to a named constant

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-27651


t-kameyama:
6ee4b5e39380f96308d328e66ccd63d4f6469817

PR #3021, opened 2020-01-27
KT-22878 Empty argument list at the call site of custom function named "suspend" shouldn't be reported as unnecessary 

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22878


t-kameyama:
581f83421ad228b8789af83055c54be6396af33f

PR #3023, opened 2020-01-27
JVM_IR: Use direct field access instead of calling certain accessors.

madsager:
Final default properties accessors that access a backing field
on the same class can be replaced by direct field use.

Perform the optimization late in the pipeline to allow lowerings
to expose more opportunities for optimizations.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
@@ -6,15 +6,17 @@
 package org.jetbrains.kotlin.backend.jvm.lower
 
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
+import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.common.lower.irBlock
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.lower.inlineclasses.unboxInlineClass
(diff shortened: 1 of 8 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
```diff
@@ -71,9 +72,38 @@ class JvmBuiltinOptimizationLowering(val context: JvmBackendContext) : FileLower
...
+
+            // For some functions, we clear the current class field since the code could end up
+            // in another class then the one it is nested under in the IR.
+            // TODO: Loosen this up for local functions for lambdas passed as an inline lambda
+            // argument to an inline function. In that case the code does end up in the current class.
+            override fun visitFunction(declaration: IrFunction, currentClass: IrClass?): IrStatement {
+                val codeMightBeGeneratedInDifferentClass = declaration.isSuspend ||
+                        declaration.isInline ||
+                        declaration.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA
+                declaration.transformChildren(this, if (codeMightBeGeneratedInDifferentClass) null else currentClass)
```
kandersen: Nit: good place for `takeIf`: `currentClass.takeIf { codeMightBeGeneratedInDifferentClass }`

udalov: `takeUnless` probably

kandersen: Very likely! :) Thank you!

madsager: Done. Thank you. :)

max-kammerer:
Thank you!

PR #3024, opened 2020-01-27
JVM: Add D8 check that class files can be dexed with D8.

madsager:
This change adds D8 in addition to Dx which is being deprecated.
At some point after the official Dx deprecation, we should remove
the Dx checker.

Moving to D8 has the additional benefit that D8 can dex code
using java 8 features without using a separate desugaring tool.


udalov:
Thank you!

PR #3029, opened 2020-01-28
KT-36194 "Add braces to 'for' statement" inserts extra line break and moves the following single-line comment

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36194


t-kameyama:
777995a3a064ffc10de9ca5b4407151d9b17a0c8

PR #3030, opened 2020-01-28
JVM_IR: further refine synthetic accessor generation

pyos:
References to protected members from crossinline lambdas in the same package do not need accessors.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
@@ -641,17 +641,18 @@ private class AddContinuationLowering(private val context: JvmBackendContext) : 
     private fun markSuspendLambdas(irElement: IrElement): Pair<List<SuspendLambdaInfo>, List<IrFunction>> {
         val suspendLambdas = arrayListOf<SuspendLambdaInfo>()
         val capturesCrossinline = mutableSetOf<IrCallableReference>()
-        val visitor = object : IrInlineReferenceLocator(context) {
+        val inlineReferences = mutableSetOf<IrCallableReference>()
+        irElement.acceptChildrenVoid(object : IrInlineReferenceLocator(context) {
             override fun visitElement(element: IrElement) {
                 element.acceptChildrenVoid(this)
             }
 
-            override fun handleInlineFunctionCallableReferenceParam(valueArgument: IrCallableReference) {
-                super.handleInlineFunctionCallableReferenceParam(valueArgument)
-                val getValue = (valueArgument as? IrGetValue) ?: return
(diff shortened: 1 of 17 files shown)


compiler/testData/codegen/boxAgainstJava/visibility/protectedAndPackage/protectedPropertyInPackageFromCrossinline.kt
```diff
@@ -0,0 +1,19 @@
...
+
+// FILE: 1.kt
+
+package protectedPack
+
+inline fun foo(crossinline bar: () -> String) = object {
+    fun baz() = bar()
+}.baz()
+
+fun box(): String {
```
max-kammerer: Could you add more test with protected, e.g. when protected methods are called from crossinline lambda, but in case when it's inlined again in inline function, e.g.:
class A {
  protected fun smth() {}
}

class B: A() {
   inline fun test(l: () -> ?) = inlineCrossInlineCall{ foo(); l() }
}

and test call from another package

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AddContinuationLowering.kt
```diff
@@ -641,17 +641,18 @@ private class AddContinuationLowering(private val context: JvmBackendContext) :
...
+        val inlineReferences = mutableSetOf<IrCallableReference>()
+        irElement.acceptChildrenVoid(object : IrInlineReferenceLocator(context) {
             override fun visitElement(element: IrElement) {
                 element.acceptChildrenVoid(this)
             }
 
-            override fun handleInlineFunctionCallableReferenceParam(valueArgument: IrCallableReference) {
-                super.handleInlineFunctionCallableReferenceParam(valueArgument)
-                val getValue = (valueArgument as? IrGetValue) ?: return
+            override fun visitInlineReference(argument: IrCallableReference) {
```
max-kammerer: @pyos @ilmirus Looks like that this code now is called more times then before. is it make any sense (should it affect any test?)

max-kammerer: Oh, looks like that after 'inlineCallableReferenceToLambdaPhase' there are no such cases

compiler/testData/codegen/boxAgainstJava/visibility/protectedAndPackage/protectedPropertyInPackageFromCrossinline.kt
pyos: Done. This is actually an interesting example, as it fails on JVM (while JVM_IR correctly traces the call site to an inline function and adds an accessor).

max-kammerer:
Thank you!

PR #3032, opened 2020-01-28
[JVM IR] Maintain KT-36188 bug compatibility between non-IR and IR backends

punzki:
The tests fail on JS (non-IR). Currently running tests on Native; not sure if this util function is used there.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
@@ -96,12 +96,16 @@ internal class InterfaceLowering(val context: JvmBackendContext) : IrElementTran
                  *    ```
                  */
                 function.origin == IrDeclarationOrigin.FAKE_OVERRIDE -> {
-                    val implementation = function.resolveFakeOverride()!!
+                    // We check to see if this is a default stub function BEFORE finding the implementation because of a front-end bug
+                    // (KT-36188) where there could be multiple implementations. (resolveFakeOverride() only returns the implementation if
+                    // there's only one.)
+                    if (function.name.asString().endsWith("\$default")) {
+                        continue@loop
+                    }
+                    val implementation = function.resolveFakeOverride() ?: error("No single implementation found for: ${function.render()}")
 
                     when {
-                        Visibilities.isPrivate(implementation.visibility) ||
-                                implementation.isMethodOfAny() ||
-                                implementation.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER ->
+                        Visibilities.isPrivate(implementation.visibility) || implementation.isMethodOfAny() ->
                             continue@loop
                         !implementation.hasJvmDefault() -> {
                             val defaultImpl = createDefaultImpl(function)
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -8535,6 +8535,16 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
                 runTest("compiler/testData/codegen/box/defaultArguments/function/innerExtentionFunctionManyArgs.kt");
             }
 
+            @TestMetadata("kt36188.kt")
+            public void testKt36188() throws Exception {
+                runTest("compiler/testData/codegen/box/defaultArguments/function/kt36188.kt");
+            }
+
+            @TestMetadata("kt36188_2.kt")
+            public void testKt36188_2() throws Exception {
+                runTest("compiler/testData/codegen/box/defaultArguments/function/kt36188_2.kt");
+            }
+
             @TestMetadata("kt5232.kt")
             public void testKt5232() throws Exception {
                 runTest("compiler/testData/codegen/box/defaultArguments/function/kt5232.kt");
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -8535,6 +8535,21 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
                 runTest("compiler/testData/codegen/box/defaultArguments/function/innerExtentionFunctionManyArgs.kt");
             }
 
+            @TestMetadata("kt15971.kt")
+            public void testKt15971() throws Exception {
+                runTest("compiler/testData/codegen/box/defaultArguments/function/kt15971.kt");
+            }
+
+            @TestMetadata("kt15971_2.kt")
+            public void testKt15971_2() throws Exception {
+                runTest("compiler/testData/codegen/box/defaultArguments/function/kt15971_2.kt");
+            }
+
+            @TestMetadata("kt15971_3.kt")
+            public void testKt15971_3() throws Exception {
+                runTest("compiler/testData/codegen/box/defaultArguments/function/kt15971_3.kt");
+            }
+
             @TestMetadata("kt36188.kt")
             public void testKt36188() throws Exception {
                 runTest("compiler/testData/codegen/box/defaultArguments/function/kt36188.kt");
--- a/compiler/testData/codegen/box/defaultArguments/function/kt15971.kt
+++ b/compiler/testData/codegen/box/defaultArguments/function/kt15971.kt
@@ -1,0 +1,19 @@
+// IGNORE_BACKEND_FIR: JVM_IR
+// IGNORE_BACKEND: JVM
+interface Q {
+    fun foo(a: Double): Double
+}
+
+interface Z {
+    fun foo(a: Double = 1.0): Double
+}
+
+class R : Q, Z {
+    override fun foo(a: Double) = a
+}
+
+fun box(): String {
+    if (R().foo() != 1.0) return "fail 1"
+    if (R().foo(2.0) != 2.0) return "fail 2"
+    return "OK"
+}
--- a/compiler/testData/codegen/box/defaultArguments/function/kt15971_2.kt
+++ b/compiler/testData/codegen/box/defaultArguments/function/kt15971_2.kt
@@ -1,0 +1,19 @@
+// IGNORE_BACKEND_FIR: JVM_IR
+// IGNORE_BACKEND: JVM
+interface Q {
+    fun foo(a: Double) = 0.0
+}
+
+interface Z {
+    fun foo(a: Double = 1.0): Double
+}
+
+class R : Q, Z {
+    override fun foo(a: Double) = a
+}
+
+fun box(): String {
+    if (R().foo() != 1.0) return "fail 1"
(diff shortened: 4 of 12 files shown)


udalov:
`resolveFakeOverride` is used in multiple places where this change might mask some other problems.

For example, just recently I was debugging an incremental compilation issue where the same class was present both in sources and in binaries, and `resolveFakeOverride` on a member of its subclass could not select a declaration among those two superclasses. Of course, this is an irrelevant problem, but my point is that changing `singleOrNull` to `firstOrNull` here would have masked the problem, or made it more difficult to debug.

Can we only change the callsite at `DefaultArgumentStubGenerator.parametersForCall` instead?

udalov:
Looks like this has actually been fixed by @anton-bannykh in `rr/abannykh/pir-rebase3`:
https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3792?commentId=01471b0f-0246-46c9-85f8-653186884812&filePath=/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt

Maybe we could just wait for that change to land then.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -286,7 +286,7 @@ fun IrSimpleFunction.collectRealOverrides(): Set<IrSimpleFunction> {
 // This implementation is from kotlin-native
 // TODO: use this implementation instead of any other
 fun IrSimpleFunction.resolveFakeOverride(): IrSimpleFunction? {
-    return collectRealOverrides().singleOrNull { it.modality != Modality.ABSTRACT }
+    return collectRealOverrides().firstOrNull { it.modality != Modality.ABSTRACT }
```
bashor: I think there is an assumption that we could have only one real implementation for a fake override and it would be nice to keep it. I'd suggest adding another implementation with Deprecated annotation instead of modifying this one.

punzki: @udalov @bashor Thanks for the review. The original fix I had was more localized (similar to what @bashor is recommending) but I went this route because the comment above `resolveFakeOverride` mentioned it was from K/N, and the implementation for K/N was using `first`: https://github.com/JetBrains/kotlin-native/blob/bf3d5965df8df0999dfd6d3439c407090c4c5de9/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/descriptors/DescriptorUtils.kt#L78 

Aside from `DefaultArgumentStubGenerator.parametersForCall`, there was another problematic usage of `resolveFakeOverride` (that assumes only one real implementation) that I am not sure is fixed by KOTLIN-CR-3792. It's this one in `InterfaceLowering`: https://github.com/JetBrains/kotlin/blob/c9df17f2f18700702db3d844047ac5556edea911/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt#L96

This is in fact why I had two tests. `compiler/testData/codegen/box/defaultArguments/function/kt36188_2.kt` covers the problematic assumption in `InterfaceLowering`. If this is also fixed in KOTLIN-CR-3792, then I'm ok to just wait for that to merge with these tests added. I didn't go through that CR thoroughly but I _don't_ think it's fixed, in which case I can just make the localized fix: Add a new implementation that uses `firstOrNull` and apply it in the 2 problematic call sites.

udalov: @punzki You're right, I've missed the usage in `InterfaceLowering`.

In this case it seems reasonable to add a new function which returns `firstOrNull` and use it in these two call sites (I don’t think annotating it with Deprecated, as proposed by @bashor, is correct though). Thanks!

While you're at it, consider also improving the exception message when no fake override is found. Right now it's just NPE because of `!!` at call sites, but we could make the function itself check if no overrides are found and render the function to ease debugging:
```
fun IrSimpleFunction.resolveFakeOverrideToFirst(): IrSimpleFunction {
    return collectRealOverrides().firstOrNull { it.modality != Modality.ABSTRACT }
        ?: error("No override is found: ${render()}")
}
```

bashor: My point is that for fake overrides we shouldn't have more than one overridden declaration with an implementation modulo bugs. 
But in the case of default arguments, it's slightly different -- we have to find first with default arguments independent to modality, just finding the first implementation is not enough. And in this case, it's "firstOrNull" but not "singleOrNull" only because of the issue in FE.

Probably we have to rename these functions to avoid feature misuses.

bashor: BTW, current JVM BE generates wrong code for these valid cases:
```kotlin
interface Q {
    fun foo(a: Int) {}
}

interface Z {
    fun foo(a: Int = 1)
}

class R : Q, Z {
    override fun foo(a: Int) {
        println(a)
    }
}

fun main() {
    R().foo()
}
```

```kotlin
interface Q {
    fun foo(a: Int = 1) {}
}

interface Z {
    fun foo(a: Int)
}

class R : Z, Q {
    override fun foo(a: Int) {
        println(a)
    }
}

fun main() {
    R().foo()
}
```

bashor: I didn't check them yet for IR BEs and didn't find related issues in YT

bashor: Some related issues
https://youtrack.jetbrains.com/issue/KT-15971
https://youtrack.jetbrains.com/issue/KT-9474
https://youtrack.jetbrains.com/issue/KT-34866

punzki: @bashor Can you clarify what you mean by "But in the case of default arguments, it's slightly different -- we have to find first with default arguments independent to modality, just finding the first implementation is not enough."?

You mean we should just find the first overridden method, even if it _is_ abstract? Why is that?

bashor: Sorry, if I confused you.

>  first with default arguments
I meant the first declaration where the default value is defined and it could have any modality (see examples above).
It's about how we should find declaration if we need default values (specifically in `DefaultArgumentStubGenerator#parametersForCall`)

In `InterfaceLowering` we have to find the first `*$default` with implementation.
Another point was that we could be more strict here and use "single" if we didn't have a bug in FE.

bashor: BTW, examples above look like work in JVM IR BE

punzki: @bashor Sorry for the delayed reply. In `DefaultArgumentStubGenerator` we are trying to resolve fake overrides for `*$default` and not the original function. All "real" (_not_ fake overrides) `*$default` functions will _never_ have "abstract" modality; they are always "final" since there's always some implementation eventually (the implementation that calls the original function). So while it's okay to find the first "real" function regardless of modality, it should also be okay to keep the `{ it.modality != Modality.ABSTRACT }` filter predicate.

I will also add those examples. Those are KT-15971, right?

punzki: @bashor @udalov Rebased and made the requested changes, now in `BridgeLowering` instead of `DefaultArgumentStubGenerator`. Please take a look!

I also added your examples as tests. I named the new function `resolveFakeOverrideWithBugCompatibility`; hopefully this makes it very clear that it's there only for bug-compatibility purposes, without having to mark it `@Deprecated`.

udalov:
Can we localize it even more, e.g. only use the new implementation if the function name ends with `$default`, and otherwise use `resolveFakeOverride`?

Also, please note the changes in 79060e7f40d89c009294652cadf7ded4402add87 ([KOTLIN-CR-3825](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3825)). It's probably not related to the implementation of `resolveFakeOverrideWithBugCompatibility`, because implementations of `foo$default` cannot be found in Java interfaces, so this is just an FYI.

punzki:
> Can we localize it even more, e.g. only use the new implementation if the function name ends with `$default`, and otherwise use `resolveFakeOverride`?
> 
> Also, please note the changes in [79060e7](https://github.com/JetBrains/kotlin/commit/79060e7f40d89c009294652cadf7ded4402add87) ([KOTLIN-CR-3825](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3825)). It's probably not related to the implementation of `resolveFakeOverrideWithBugCompatibility`, because implementations of `foo$default` cannot be found in Java interfaces, so this is just an FYI.

@udalov Done, updated PR. Let me know if I understood you correctly (or not!)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
```diff
@@ -174,7 +174,10 @@ private class BridgeLowering(val context: JvmBackendContext) : FileLoweringPass,
             // (MutableMap.remove and getOrDefault). Trying to produce (special) bridges for these methods could
             // result in incorrect bytecode on older JVM versions. However, all such methods are declared
             // in interfaces and thus we don't need a separate check for them.
-            return !irFunction.isFakeOverride || !irFunction.resolveFakeOverride()!!.parentAsClass.isJvmInterface
+            //
+            // We use `resolveFakeOverrideWithBugCompatibility` due to a front-end bug (KT-36188) where there could be multiple
+            // fake-overridden implementations of a default stub function.
+            return !irFunction.isFakeOverride || !irFunction.resolveFakeOverrideWithBugCompatibility()!!.parentAsClass.isJvmInterface
```
udalov: FYI [KOTLIN-CR-3895](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3895) gets rid of this `resolveFakeOverride`.

punzki: Ok. I'll rebase once that's land.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/InterfaceLowering.kt
```diff
@@ -96,7 +96,10 @@ internal class InterfaceLowering(val context: JvmBackendContext) : IrElementTran
                  *    ```
                  */
                 function.origin == IrDeclarationOrigin.FAKE_OVERRIDE -> {
-                    val implementation = function.resolveFakeOverride()!!
+                    // We use `resolveFakeOverrideWithBugCompatibility` due to a front-end bug (KT-36188) where there could be multiple
+                    // fake-overridden implementations of a default stub function.
+                    val implementation =
+                        function.resolveFakeOverrideWithBugCompatibility() ?: error("No implementation found for: ${function.render()}")
```
udalov: Now that we have this workaround localized... :) Observe that a few lines below, we actually ignore functions for default parameters. So it seems that we could factor the name checking logic out of the new function and just `continue` the loop here if the fake override's name ends with `$default`.

This would make `resolveFakeOverrideWithBugCompatibility` not needed, at least for now. What do you think?

punzki: Makes sense to me! Updated. (This PR has shrunk quite a bit from the original.) 🙂

udalov:
Looks good now, thank you for the contribution!

PR #3034, opened 2020-01-29
KT-23934 IntelliJ suggest "merge map to joinToString" even when such action is impossible due to suspending actions in map

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23934

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
@@ -40,10 +40,12 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
                         }
                     ) return@check false
                 }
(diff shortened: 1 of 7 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/collections/SimplifiableCallChainInspection.kt
```diff
@@ -35,6 +39,10 @@ class SimplifiableCallChainInspection : AbstractCallChainChecker() {
                         }
                     ) return@check false
                 }
+                if (!conversion.enableSuspendFunctionCall && firstResolvedCall.call.callElement.anyDescendantOfType<KtCallExpression> {
```
yanex: I would do something with the formatting here :)
Maybe, extract the `firstResolvedCall.call.callElement.anyDescendantOfType()` call so the whole `if` condition fit into a single line?

t-kameyama:
@yanex PTAL

yanex:
Pushed into master manually (https://github.com/JetBrains/kotlin/commit/b008e6c9be057f8a1bdeee960059add6435a216a).
Thank you for your contribution, as always!

PR #3035, opened 2020-01-29
KT-17368 Don't highlight members annotated with @JsName as unused

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17368


t-kameyama:
905d0c10167c8711d4db16a7ecfdefaf64b4c615

PR #3036, opened 2020-01-29
[JVM IR] ProperVisibilityForCompanionObjectInstanceField

kandersen:
This PR causes the IR backend to respect the visibility modifier of companion objects.

This causes a slight complication in synthetic accessor lowering: if a protected/private companion object implements a class unrelated to the containing class, accessors for members of superclass must be placed on the companion object, not on the containing class.

 An exception is made for interfaces as this functionality is implemented across JVM targets and <1.8 does not allow implementations (hence, accessors) on interfaces - the static companion object instance field must remain public to avoid the accessor.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
@@ -38,6 +38,7 @@ import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
 import org.jetbrains.kotlin.load.java.JavaVisibilities
 import org.jetbrains.kotlin.load.java.JvmAbi
 import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.synthetic.isVisibleOutside
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrElementTransformerVoidWithContext(), FileLoweringPass {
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -167,7 +173,11 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
     private fun IrDeclarationWithVisibility.accessorParent(parent: IrDeclarationParent = this.parent) =
         if (visibility == JavaVisibilities.PROTECTED_STATIC_VISIBILITY) {
             val classes = allScopes.map { it.irElement }.filterIsInstance<IrClass>()
-            classes.lastOrNull { parent is IrClass && it.isSubclassOf(parent) } ?: classes.last()
+            val companions = classes.mapNotNull { it.companionObject() }.filterIsInstance<IrClass>()
+            val objectsInScope =
+                classes.flatMap { it.declarations.filter(IrDeclaration::isAnonymousObject).filterIsInstance<IrClass>() }
+            val candidates = objectsInScope + companions + classes
+            candidates.lastOrNull { parent is IrClass && it.isSubclassOf(parent) } ?: classes.last()
```
kandersen: Feedback on these list gymnastics is more than welcome. I feel this could be done better.

dnpetrov:

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
dnpetrov: I think it's ok. I'm somewhat unsure about `?: classes.last()` clause, but if the tests like those in testData\codegen\box\objects\companionObjectAccess\multipleCompanionsWithAccessors pass, that's fine.

kandersen: Yeah, `classes.last()` is the fallback left in from previous version. You know, I never actually tried removing it, seeing what instances arent covered by the others.

I guess in principle the backend shouldn't error out?

dnpetrov: I actually prefer to error out in such cases, so that we identify problems fast. 
It _seems_ to me that `classes.last()` is ok because of how scopes are organized, although I'd better write some more tests with nested/inner classes.

kandersen: OK! Do you want me to do anything more for this PR?

dnpetrov: I think it's ok for now. Since the tests passed, I'll close the PR and push manually.

dnpetrov:
Pushed manually ( f782ea075ba6590911573a0301154be435b2198f )

PR #3037, opened 2020-01-29
[JVM + IR] Fix Generic Signature Test Expectations

kandersen:


udalov:
Makes sense. Thanks!

kandersen:
Personal record for smallest PR, I think :)

PR #3038, opened 2020-01-29
[JVM IR] Interface Delegation by Property should use existing field

kandersen:
When a property backed by a final field is the delegate for interface implementation by delegation, the backend should reuse the backing field rather than introduce a new, extraneous `$$delegate_n` field.


dnpetrov:
Looks good to me.
I'll run tests and merge manually.

dnpetrov:
Pushed manually (61e6d346aa02baab0ccdcca096d1366c6ec51603).

PR #3040, opened 2020-01-30
ForLoopsLowering: Fix issue with break/continue in loop over withIndex

punzki:
@max-kammerer Thanks for notifying me of this issue, here is the fix.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
@@ -15,7 +15,6 @@ import org.jetbrains.kotlin.ir.builders.*
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.IrCall
-import org.jetbrains.kotlin.ir.expressions.IrContainerExpression
 import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrLoop
 import org.jetbrains.kotlin.ir.expressions.impl.IrDoWhileLoopImpl
@@ -374,6 +373,7 @@ internal class WithIndexLoopHeader(
     private val nestedLoopHeader: ForLoopHeader
     private val indexVariable: IrVariable
     private val ownsIndexVariable: Boolean
+    private val incrementIndexStatement: IrStatement?
 
     init {
         with(builder) {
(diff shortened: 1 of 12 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -396,18 +396,34 @@ internal class WithIndexLoopHeader(
...
             } else {
                 indexVariable = scope.createTemporaryVariable(
                     irInt(0),
                     nameHint = "index",
                     isMutable = true
                 )
                 ownsIndexVariable = true
+                // `index++` during iteration initialization
+                // TODO: MUSTDO: Check for overflow for Iterable and Sequence (call to checkIndexOverflow()).
+                val plusFun = indexVariable.type.getClass()!!.functions.first {
```
max-kammerer: What problem with `checkIndexOverflow` support>

max-kammerer: Looks like `checkIndexOverflow` should be check in another place, otherwise exception would be thrown too early

punzki: I don't follow, where should the check be? When I looked at the stdlib code, this looks like the right place to do it. It's always `checkIndexOverflow(index++)`

punzki: The reason I didn't rush to add the check in is because it's also currently broken current JVM backend. https://youtrack.jetbrains.com/issue/KT-34665

It's not hard to add it though.

max-kammerer: Did you mean that loop optimization is broken in old JVM backend in same case?  Then is should be fixed in 1.4 (maybe with LC ticket cc @zarechenskiy @dnpetrov )
Look like that proper one behaviour should be supported for K/N and K/JS in any way. cc @homuroll 

My point was that such check should be performed after loop body execution for last unoverflow value (e.g. MAX_INT). But if it not presented it doesn't make any sence except maybe mention in comment

punzki: Yes it's also broken in old JVM backend, that's what KT-34665 is for. I personally don't think it should require LC approval, because the addition of the check was already approved (https://youtrack.jetbrains.com/issue/KT-16097) and the implementation was simply missed for the optimized for-loop case.

If you check the index value (`if (index < 0) throw...`) _before_ incrementing then that would be equivalent to `checkIndexOverflow(index++)`. The implementation of `checkIndexOverflow` checks if the index is negative.

max-kammerer:
Thank you for pr and clarifications! Assertion related things could be postponed to another pr

PR #3042, opened 2020-01-30
JVM_IR: Use direct field access isntead of calling certain accessors.

madsager:
Final default properties accessors that access a backing field
on the same class can be replaced by direct field use.

Perform the optimization late in the pipeline to allow lowerings
to expose more opportunities for optimizations.

This change broke a couple of FIR tests. In order to fix that,
this changes the translation from FIR to IR to deal with
smartcasts differently.


madsager:
@max-kammerer The first commit is the optimization as you already reviewed it.

@demiurg906 Can you have a look at the Fir2Ir change?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
```diff
@@ -24,16 +26,15 @@ import org.jetbrains.kotlin.ir.types.isNullableAny
...
 import org.jetbrains.kotlin.ir.util.*
-import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
-import org.jetbrains.kotlin.ir.visitors.transformChildrenVoid
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformer
 
-internal val jvmBuiltinOptimizationLoweringPhase = makeIrFilePhase(
-    ::JvmBuiltinOptimizationLowering,
+internal val jvmOptimizationLoweringPhase = makeIrFilePhase(
+    ::JvmOptimizationLowering,
     name = "JvmBuiltinOptimizationLowering",
```
udalov: Also this could be renamed

madsager: Whoops! Thanks Alex, done. :)

max-kammerer:
Thank you!

PR #3043, opened 2020-01-30
KT-36160 False positive "Constructor has non-null self reference parameter" with vararg parameter of class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36160


mglukhikh:
Merged 👍

PR #3044, opened 2020-01-31
KT-34317 False positive "Redundant 'suspend' modifier" inspection when function reference is used as parameter with declared 'suspend' type

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-34317


mglukhikh:
I think we should not fix it this way at all. The right way to fix is to provide an opportunity to use non-suspend function reference when suspend function is needed. It's planned and will be done. See https://youtrack.jetbrains.com/issue/KT-15917.

PR #3045, opened 2020-01-31
KT-25050 False-positive inspection "Call replaceable with binary operator" for 'equals'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-25050


t-kameyama:
43b106fc727e8e9bfe54122e13f12bed9d7b7584

PR #3046, opened 2020-01-31
Android/IR: handle nulls from getView/getContainerView

pyos:
Turns out it's possible to create an IR builder without a backend context


pyos:
@yanex ping?

yanex:
Sorry for the delay.
Merged to master. Thanks!

PR #3048, opened 2020-02-01
KT-32565 False positive "Variable is the same as 'credentials' and should be inlined" with object declared and returned from lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32565


yanex:
The bug also reproduces with ordinary classes:

```
fun findCredentials() = 1

fun println(i: Int) {}

fun test() {
    val data = run {
        val credentials = findCredentials()

        class Foo {
            val foundCredentials = credentials
        }

        Foo()
    }

    val foundCredentials = data.foundCredentials
}
```

yanex:
Review: 👍

t-kameyama:
7c0af78b087e0c68936bded3e25d20a74df7f37c

PR #3049, opened 2020-02-02
Prr/juanch/invoke

juan-chen:
* fixed NoSuchMethod for lambda "invoke" methods due to signature mismatch
* added a wrapper for the returned expression in lambdas.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -692,6 +691,8 @@ class Fir2IrVisitor(
             is FirPropertyFromParameterResolvedNamedReference -> IrStatementOrigin.INITIALIZE_PROPERTY_FROM_PARAMETER
             is FirResolvedNamedReference -> when (resolvedSymbol) {
                 is AccessorSymbol, is SyntheticPropertySymbol -> IrStatementOrigin.GET_PROPERTY
+                is FirNamedFunctionSymbol -> if (resolvedSymbol.callableId.isInvoke()) IrStatementOrigin.INVOKE else null
+                is FirVariableSymbol<*> -> IrStatementOrigin.VARIABLE_AS_FUNCTION
```
mglukhikh: Why it's true if `statementOrigin` is called for `IrValueSymbol` (see line 720 below)?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
```diff
@@ -428,3 +429,8 @@ fun BodyResolveComponents.transformQualifiedAccessUsingSmartcastInfo(qualifiedAc
     }
     return FirExpressionWithSmartcastImpl(qualifiedAccessExpression, intersectedTypeRef, typesFromSmartCast)
 }
+
+fun CallableId.isInvoke() =
+    packageName == KotlinBuiltIns.BUILT_INS_PACKAGE_FQ_NAME
+            && className?.asString()?.startsWith("Function") == true
```
mglukhikh: How does it work if `className` is e.g. `KFunction*`?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -162,6 +169,22 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
         return completeInference.compose()
     }
 
+    // Use the generic invoke method in Function classes, to match bridges generated by the backend.
+    private fun checkIfInvoke(functionCall: FirFunctionCall) : FirFunctionCall {
```
semoro: I'd think frontend shouldn't rewrite declarations in sake of knowledge of backend behavior, so this logic should rather be in fir2ir or even in backend itself

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -271,12 +274,17 @@ class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransformer)
...
-                result
-            } else {
-                result
             }
+            val statement = body?.statements?.lastOrNull()
+            if (statement is FirExpression && statement !is FirReturnExpression && !body.resultType.isNothing) {
+                val returnExpression = FirReturnExpressionImpl(statement.source, statement)
+                returnExpression.target = FirFunctionTarget(null)
+                returnExpression.target.bind(result)
+                (body.statements as? MutableList<FirStatement>)?.set(body.statements.size - 1, returnExpression)
```
mglukhikh: This is very hacky way, moreover it will work incorrectly for anonymous functions with expression bodies. It's better to do it via transformer (`LastBodyStatementTransformer`?) instead.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
juan-chen: Not sure I understand the question. statementOrigin is called on FirReference, right?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
juan-chen: the className needs to start with "Function", so "KFunction*" will not match.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
juan-chen: moved to fir2ir

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
juan-chen: Moved to Fir2Ir.

Also would you please elaborate on "work incorrectly for anonymous functions with expression bodies"? I tried an example `val f = {"K"}`: currently the frontend generates a function body with a single FirConstExpression, and thus inserting a return wrapper would work. Are there other cases you'd expect not to insert return?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
mglukhikh: This function has two use-sites: one for `FirReference` of a function call (line 711), and another one for `FirReference` of a variable access (line 724). Your code is correct for a function call, but probably not correct for a variable access.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
mglukhikh: Yes, but `KFunction*` also have `invoke` call, and (may be, I don't know exactly) something same should be done for it. I think it's the reason for a test of this case.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
mglukhikh: No, the example should look like `callOfFunctionWithLambda(x, y, fun(n: Int) = 42 * n)`.

mglukhikh:
Meanwhile, I've fixed problem with `invoke` call on lambda. Fix is already in master, see this commit: https://github.com/JetBrains/kotlin/commit/64c7ab13023e56d6d7d3d51515dc657bf2952867. In your PR, this provides additional 18 fixed black box tests, see here: https://github.com/JetBrains/kotlin/commit/b48eaa8c5aa0ef1b9d084fbe0cc837d580bb38ec.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
mglukhikh: As discussed today, this should be kept in FIR itself, but utility transformer like `StoreType` should be used.

juan-chen: Moved back to Fir, using a transformer

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
juan-chen: Tried an example `foo(Int::dec)`, KFunction is generated for arguments, not for parameters. I've added two test cases for KFunction into functions/invoke/invoke.kt.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
juan-chen: Yeah, removed the case for FirVariableSymbol. The Invoke case seems ok, no matter which call sites.

mglukhikh:
Merged manually. Thank you!

PR #3051, opened 2020-02-03
KT-35587 Avoid uses of plain "kotlin" in metadata deserialization.

jsjeon:


core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/ClassMapperLite.kt
```diff
@@ -13,6 +13,8 @@ package org.jetbrains.kotlin.metadata.jvm.deserialization
 // WARNING: improving the behavior of this class MAY BREAK BINARY COMPATIBILITY of code compiled by Kotlin, because it may make
 // the new compiler skip writing the signatures it now thinks are trivial, and the old compiler would recreate them incorrectly.
 object ClassMapperLite {
+    // Simply "kotlin", but to avoid being renamed by namespace relocation (e.g., Shadow.relocate gradle plugin)
+    private val kotlin = listOf('k', 'o', 't', 'l', 'i', 'n').joinToString(separator = "")
```
udalov: Have you tried just `private val kotlin = "kotlin"` here?

Note that we had a change in compilation strategy and since 1.4, non-const vals should _not_ be inlined at call sites, and thus I don't think Shadow will rename the class names below. Unless of course it just renames any string literal "kotlin" found in the class file, which would be mind-boggling to say the least.

jsjeon: I can't link an internal bug, but hopefully this code review has some clues: https://r8-review.googlesource.com/c/r8/+/23820

Anyhow, I'm pretty much sure Shadow renames any string literal "kotlin" if the relocation task is asked to do so: https://r8.googlesource.com/r8/+/refs/heads/master/build.gradle#701 and indeed, that was a surprise to us too. Actually, not just "kotlin", but literally any appearances, like "kotlin/Unit" -> "relocated_kotlin/Unit".

udalov: Thanks, I've tried a small project with Shadow and it seems to be the case. Which looks like a bug in Shadow. At least, there could be a flag to disable relocation for string literals which are not used for reflection immediately.

udalov:
@jsjeon I don't have anything against this change in particular, but I'm not sure it completely fixes the underlying problem. We have other places where we use the package name `kotlin` inside a string literal. As you mentioned in [KT-35587](https://youtrack.jetbrains.com/issue/KT-35587), `ClassMapperLite`, or for example `ModuleMapping` (where "kotlin" in `"kotlin_module"` is also relocated), or `JvmProtoBufUtil.PLATFORM_TYPE_ID`.

Moreover, there's no barrier to reintroduce this problem by refactoring our core metadata/metadata.jvm code. Maybe we could have some kind of an integration test on `kotlinx-metadata-jvm` specifically, that would check that there's no "kotlin" inside string literals in the whole resulting .jar file?

jsjeon:
Thanks for the pointers to string literals that still have `kotlin` in it. The type conversion issue arose only for string literals in JVM binary name format, so those two should be okay.

As per the suggestion, I added an integration test that walks through `.class` file in the resulting `.jar` file and
1) makes sure that kotlin types that are intentionally built with programmatically added `kotlin` prefix indeed don't appear as string constants; and
2) also double-checks that we don't accidentally add string literals that start with `kotlin/` prefix, which is prone to package relocation.

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
+import org.junit.Test
+import java.io.File
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.util.zip.ZipFile
+import kotlin.text.Charsets.UTF_8
+
+class JarContentTest {
```
udalov: Could you please move this test to `libraries/kotlinx-metadata/jvm/test`? This way you'll be testing the actual `kotlinx-metadata-jvm` artifact.

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.metadata.jvm
+
+import jdk.internal.org.objectweb.asm.ClassReader
```
udalov: Please use `org.jetbrains.org.objectweb.asm` instead

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                assertNull("$constant found at ${it.name}", findPattern(constant))
+                // Implicitly checking none of string constants starts with "kotlin/" prefix, just in case.
+                assertFalse("$constant found at ${it.name}", constant.startsWith("kotlin/"))
+            }
+        }
+    }
+
+    private fun findPattern(constant: String): String? {
+        for (pattern in PREDEFINED_STRINGS) {
+           if (constant.contains(pattern)) {
```
udalov: Please reformat

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                // Explicitly checking types that are programmatically built don't appear as string constants.
+                assertNull("$constant found at ${it.name}", findPattern(constant))
+                // Implicitly checking none of string constants starts with "kotlin/" prefix, just in case.
+                assertFalse("$constant found at ${it.name}", constant.startsWith("kotlin/"))
+            }
+        }
+    }
+
+    private fun findPattern(constant: String): String? {
+        for (pattern in PREDEFINED_STRINGS) {
```
udalov:
```
return PREDEFINED_STRINGS.find { it in constant }
```

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
jsjeon: Since `ClassMapperLite` is under `core/metadata.jvm/` I somehow thought that was the right place. Migrated.

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
jsjeon: I wasn't aware of differences, thanks!

core/metadata.jvm/test/org/jetbrains/kotlin/metadata/jvm/JarContentTest.kt
jsjeon: Changed.

core/metadata.jvm/src/org/jetbrains/kotlin/metadata/jvm/deserialization/ClassMapperLite.kt
jsjeon: @madsager and I discussed this a couple times. Since other shrinkers also suffer from the same issue, it'd be good to avoid such plain prefix in general. However, it's also silly that that relocating tool is modifying literally all appearances of target string literals in the string pool. Maybe, `R8` (or actually `L8`, a library desugaring variant) has enough utils to do a better job, like reading `.jar`, building hierarchy, relocating target package (as it does for, e.g., `java.lang`), and outputting relocated `.jar`. I'll file a feature request after discussing other folks in `R8`. Meanwhile, hope this could be a reasonable workaround.

udalov:
Thank you! Merged with minor [corrections](https://github.com/JetBrains/kotlin/commit/3a4e540d440).

jsjeon:
Looks not minor :) I'm still learning. Thanks for corrections on behalf of me!

PR #3052, opened 2020-02-03
KT-24869 False positive inspection "Redundant 'suspend' modifier"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24869

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/highlighter/KotlinSuspendCallLineMarkerProvider.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/highlighter/KotlinSuspendCallLineMarkerProvider.kt
@@ -76,10 +76,11 @@ class KotlinSuspendCallLineMarkerProvider : LineMarkerProvider {
 }
 
 private fun KtExpression.isValidCandidateExpression(): Boolean {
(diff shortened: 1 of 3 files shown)


idea/src/org/jetbrains/kotlin/idea/highlighter/KotlinSuspendCallLineMarkerProvider.kt
```diff
@@ -108,10 +109,11 @@ fun KtExpression.hasSuspendCalls(bindingContext: BindingContext = analyze(BodyRe
             }
         }
         else -> {
-            val resolvedCall = getResolvedCall(bindingContext)
-            if ((resolvedCall?.resultingDescriptor as? FunctionDescriptor)?.isSuspend == true) true
+            val target = (this as? KtParenthesizedExpression)?.getStrictParentOfType<KtCallExpression>() ?: this
```
mglukhikh: This is a very hacky way to solve the problem. The root of problem here is the fact that the considered call expression is not a valid candidate (see check at line 88), so I believe this place should be fixed instead.

mglukhikh:
Merged 👍

PR #3053, opened 2020-02-03
JVM_IR: copy annotations to nullary bridges for default constructors

pyos:


udalov:
Thank you!

Looks like I've missed the `compileJavaAgainstKotlin` tests in the document. But it's nice that all of them pass now.

PR #3056, opened 2020-02-04
JVM_IR: reuse MethodNodes for inline functions in same module

pyos:
This fixes the weird cases when a class gets overwritten by an imperfect copy, reduces the number of classes in the output if an inline function contains an inline call that causes it to have regenerated anonymous objects, and makes inlining of same module functions a bit faster in general. On the other hand, this may increase memory footprint a bit because classes cannot be flushed to the output jar, as the inliner would not be able to locate classes for anonymous objects if they have already been unloaded from memory.


pyos:
@udalov are you sure you assigned the right person? :)

pyos:
@max-kammerer ping?

compiler/testData/diagnostics/testsWithJvmBackend/inlineCycle.kt
```diff
@@ -1,11 +1,11 @@
 // !RENDER_DIAGNOSTICS_FULL_TEXT
-
+// IGNORE_BACKEND: JVM_OLD
```
udalov: Looks like this was the only test on `INLINE_CALL_CYCLE` on the old JVM backend, so could you please keep a copy with target JVM_OLD?

pyos: Done. (BTW, why is it JVM_OLD and not JVM?)

udalov: What I meant was to use `TARGET_BACKEND: JVM_OLD` instead of `IGNORE_BACKEND: JVM_IR`, so that we wouldn't need to come back to this test and wonder why it's still failing for JVM_IR, since the former is a bit more explicit :). The other tests in `diagnostics/testsWithJvmBackend` are handled in this way exactly for this reason.

AFAIK `JVM` and `JVM_OLD` are handled differently only in `TARGET_BACKEND`. `IGNORE_BACKEND: JVM` and `IGNORE_BACKED: JVM_OLD` should be equivalent.

pyos: Oh, that makes sense, thanks :)

compiler/testData/diagnostics/testsWithJvmBackend/inlineCycle.kt
```diff
@@ -1,5 +1,5 @@
 // !RENDER_DIAGNOSTICS_FULL_TEXT
-
+// TARGET_BACKEND: JVM_OLD
```
max-kammerer: Please rebase on master, as I remember @dnpetrov supported cycle check in IR

pyos: This is rebased on master. The change in this test is that both functions used to be compiled twice, thus the INLINE_CALL_CYCLE is also generated twice on each call that is part of the cycle. With the patch, there is only one INLINE_CALL_CYCLE per call. This only affects the test and not user-visible diagnostics (inlineCycle.txt and inlineCycle_ir.txt are the same) because they are deduplicated before printing.

pyos: By the way, I had to reimplement inline cycle checking without using the old backend code because it's delayed by one level of nesting. Consider the test:
```
inline fun f() { g() }
inline fun g() { f() }
```
What used to happen here is that `f` pushed `g` to a stack in `enterIntoInlining`, then `g` pushed `f` to the same stack, and finally `f` tried to push `g` again and noticed that `g` is already on the stack. This, however, requires starting bytecode generation for `f` *twice*, since it's the second pass that detects the cycle. With my patch, this is not possible: when `g` tries to inline `f`, it gets an incomplete method node from the class codegen, and the diagnostics have not been generated yet. This is why there is now a new stack in `JvmBackendContext` which codegen for `g` can look into to emit the diagnostics when it sees the incomplete node.

pyos:
@max-kammerer ping?

max-kammerer:
@pyos Sorry for long delay, there were many different circumstances that postpones this review.

Maybe I'm wrong but looks like there are 2 separate issue/problem:
1. Redundant (and wrong) class generation in result output (that could be observed by updated test data in PR where some classes were removed)
2. Optimization itself
By unknown reason Optimization fixes 1. But looks like it would be much better to localize original problem in 1. and separate fixes for 1 and 2

If it's possible to split this pr in more smaller pieces it would be helpful

compiler/testData/diagnostics/testsWithJvmBackend/inlineCycle.kt
max-kammerer: Could we make common solution for both backends with remembering original compilation context (function `f`) in same stack so delay would be avoided?

pyos:
The redundant class generation is a consequence of generating code inline methods for each call site: if the inline function has inline calls itself, the anonymous classes from these calls are regenerated multiple times and the name generator gives them new names every time. Sure, this can be fixed separately by giving the same regenerations of the same objects same names, but it's not trivial to tell this situation apart from the one where multiple functions have equal names (the reason name generators are cached in the first place), and not generating inline methods more than once will override this fix anyway.

max-kammerer:
@pyos I'm correct that problem exists only in IR bakend?

max-kammerer:
Is it coroutine related on not? (Looks like that only coroutines related tests are affected, but i'm not sure that we have listing test for non coroutines ones)

pyos:
No, the old backend also does that. With this PR, the `_ir.txt` test data actually contains fewer classes.

>(Looks like that only coroutines related tests are affected, but i'm not sure that we have listing test for non coroutines ones)

Those are just the ones that happen to dump the classes, yeah.

compiler/testData/diagnostics/testsWithJvmBackend/inlineCycle.kt
pyos: Maybe. I'll try and submit that as a separate PR if possible.

pyos: Done in #3194

pyos:
Rebased on master.

@max-kammerer @jsjeon please note that the first commit reimplements #3180 in a different way: instead of creating a new ClassCodegen, only a new DefaultSourceMapper is passed directly to FunctionCodegen, then to ExpressionCodegen, and finally to nested IrSourceCompilerForInline. This addresses my comments about inner classes in that PR as well as removes the need for subclassing ClassCodegen.

pyos:
>(Looks like that only coroutines related tests are affected, but i'm not sure that we have listing test for non coroutines ones with inline)

@max-kammerer I've added CHECK_BYTECODE_LISTING to a test that shows the same change on a smaller scale and without coroutines. In `regeneratedLambdaName2.txt` (old backend), you can see `_2Kt$sameName$$inlined$call$3` and `_2Kt$sameName$$inlined$call$4`: those are actually copies of `_2Kt$sameName$$inlined$call$1` and `_2Kt$sameName$$inlined$call$2`, respectively, which were generated when these functions were compiled a second time while being inlined into `box`. Without the PR, `regeneratedLambdaName2_ir.txt` looks pretty much the same; with the PR, those classes are no longer present and `$1` and `$2` are used by `box` instead.

This illustrates the "bad case", where the objects are not regenerated so *all copies are referenced*, resulting in bloat. If the objects were to capture a crossinline lambda, for example, `_2Kt$sameName$$inlined$call$3` would be regenerated as `_2Kt$box$$inlined$sameName$1` and become unused, so a minifier would theoretically remove it.

jsjeon:
> please note that the first commit reimplements #3180 in a different way: instead of creating a new ClassCodegen, only a new DefaultSourceMapper is passed directly to FunctionCodegen, then to ExpressionCodegen, and finally to nested IrSourceCompilerForInline. This addresses my comments about inner classes in that PR as well as removes the need for subclassing ClassCodegen.

This sounds like a better way, thanks!

compiler/testData/codegen/bytecodeText/kt10259.kt
```diff
@@ -28,7 +27,8 @@ inline fun test(s: () -> Unit) {
 // this behavior is equivalent to javac and seems to be correct.
 
 // JVM_IR_TEMPLATES
-// 2 INNERCLASS Kt10259Kt\$box\$1\$1\s
+// 5 INNERCLASS
```
max-kammerer: Am I correct that old backend misses one entry? Or it opposite situation?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -161,33 +164,46 @@ open class ClassCodegen protected constructor(
...
+                val method = enclosingFunction?.let(context.methodSignatureMapper::mapAsmMethod)
+                visitor.visitOuterClass(parentClassCodegen.type.internalName, method?.name, method?.descriptor)
+            }
         }
 
-        // Generate nested classes at the end, to ensure that codegen for companion object will have the necessary JVM signatures in its
-        // trace for properties moved to the outer class
-        for (codegen in nestedClasses) {
-            codegen.generate()
+        if (sourceMapShouldBeWritten) {
```
max-kammerer: Could you explaint a bit intention of moving this part from done?

compiler/testData/codegen/bytecodeText/kt10259.kt
pyos: Yes, the old backed regenerates the class and to be 100% correct the regenerated copy should be added to inner classes of Kt10359Kt.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
pyos: A separate done() is simply no longer necessary as coroutine-related code in FunctionCodegen calls generate() instead after the method transformer finishes.

max-kammerer: Actually `generate` method is too big, maybe extract inner class related logic to separate method?

pyos: Sure.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -95,31 +87,14 @@ class IrSourceCompilerForInline(
...
-                    Opcodes.API_VERSION,
-                    flags,
-                    asmMethod.name.removeSuffix(FOR_INLINE_SUFFIX),
-                    asmMethod.descriptor,
-                    signature.genericsSignature,
-                    null
-                )
-                return wrapWithMaxLocalCalc(node!!)
-            }
+        get() = codegen.smapOverride ?: codegen.classCodegen.sourceMapper.also {
```
max-kammerer: Wouldn't be here some kind of instability in case for example of IC?

pyos: You mean something like
```
// a.kt
fun f() = h()
// b.kt
fun g() = otherInlineFun() // written to source map first if b.kt is compiled separately
inline fun h() = otherInlineFun() // written to source map first if a.kt and b.kt are compiled together
// c.kt
inline fun otherInlineFun() {}
```
? It actually would, yes. Ideally, writes to the source map should be done in source order, just like method nodes are written into methods by ClassCodegen and not whenever IrSourceCompilerForInline requests them.

Of course, the current state isn't free from instability either:
```
// a.kt
fun f() = h()
// b.kt
// These two methods reference different classes if compiled separately:
fun g() = h()
inline fun h() = myRun { "OK" }
// c.kt
inline fun <T> myRun(crossinline x: () -> T) = { x() }()
```

pyos: One way to solve this that I see is to create separate source maps for each method and cache them together with method nodes, then in `generateMethod` merge each cached SMAP into the class and use a method visitor to remap the line numbers.

max-kammerer: @pyos Could you also add IC test (@altsvetkov Please help with best form for it)

max-kammerer: > One way to solve this that I see is to create separate source maps for each method and cache them together with method nodes, then in generateMethod merge each cached SMAP into the class and use a method visitor to remap the line numbers.

If there independent part in this pr it could be extracted in separate one

pyos: Extracted various refactorings into #3205 - the rest of the changes are not as big: https://github.com/pyos/kotlin/compare/ir/refactor-class-codegen..ir/reuse-inline-methods
I have a general idea on how to make source maps stable, but this will require modifying SMAP-related classes, so I'll submit that part separately as well.

max-kammerer: Last moment thought: @altsvetkov Is it correct that if 'b.kt' is changed somehow but `h` and `myRun` untouched then 'a.kt' wouldn't recompiled by IC. Or it's would be recompiled in anyway cause of inline dependency?

pyos: That's actually a good question - if the line numbers in b.kt shift, then the source maps in AKt need to be updated even if `h` stayed the same otherwise...

max-kammerer: NB: just understand that changed smap in 'b' implies different linenumbers in 'h' (cause there would be another fake line numbers). So 'a' also would be affected cause of changed bytecode of inline function (looks like that fake linenumbers in 'a' would be same except also changed smap mapping). So recompilation order in IC would looks like b+c, then a (@altsvetkov am I correct? Would be any additional IC phase triggered?)

altsvetkov: > Is it correct that if 'b.kt' is changed somehow but `h` and `myRun` untouched then 'a.kt' wouldn't recompiled by IC. Or it's would be recompiled in anyway cause of inline dependency?

It will be recompiled, if bytecode is changed. We estimate changes in bytecode by calculating MD5 hash of a JVM method corresponding to an inline declaration. Currently usages of modified inline function are recompiled separately during the next round, but we may change that behaviour in future (currently it is possible to incrementally define a cycle of inline functions (which would be impossible to compile together, but can be compiled separately)). Here is the code calculating hashes https://github.com/JetBrains/kotlin/blob/master/build-common/src/org/jetbrains/kotlin/incremental/IncrementalJvmCache.kt#L491

altsvetkov: > NB: just understand that changed smap in 'b' implies different linenumbers in 'h' (cause there would be another fake line numbers). So 'a' also would be affected cause of changed bytecode of inline function (looks like that fake linenumbers in 'a' would be same except also changed smap mapping. Question what would be with 'a'when 'h' bytecode is totatlly same are still open). So recompilation order in IC would looks like b+c, then a (@altsvetkov am I correct? Would be any additional IC phase triggered?)

First lets check, if I understand dependencies correctly: a.kt depends on b.kt, which in turn depends on c.kt (depends == by inlining). 
If c.kt is changed, then IC will do the following:
1. Compile c.kt, compare proto + bytecode of `otherInlineFun`.
2. If bytecode of `otherInlineFun` is changed, then compile and compare `b.kt`.
3. If bytecode of `h` is changed, then compile and compare `a.kt`.

The algorithm is the same:
0. All source files modified by user are considered dirty.
1. Dirty files are compiled, resulting proto and bytecode for inline declarations/constants are compared with cached values.
2. Dirty set is cleared.
3. Usages of modified FQ-names are added to dirty set (by searching lookups).
4. If new dirty set is not empty, go to 1.

altsvetkov: For tests check test data in https://github.com/JetBrains/kotlin/tree/master/jps-plugin/testData/incremental/incrementalJvmCompilerOnly

All test are a bunch of source files and steps. Each step creates, modifies, deletes or "touches" some of source files (similar to bunch files `a.kt.new.1` means that on step 1 `a.kt` content will be replaced with the content of `a.kt.new.1`; instead of `new` you can use `delete` or touch where appropriate). Conceptually after each step we run IC and compare what files were compiled with expected files (in actuality we log results of each step, and compare all steps with build.log.txt in the end). If all steps performed IC as expected (i.e. only expected files were compiled on each step), then we rebuild all files in a test case, and compare incremental and non-incremental outputs.

pyos: @altsvetkov thanks for the explanation.

@max-kammerer I've fixed this problem and added two incremental compilation tests, both of which pass with the PR - a strict improvement over the current situation (one of them also fails on the non-IR backend). `inlineFunctionRegeneratedObjectStability` could still be modified to fail by having multiple inline functions with the same name, though. In general, it'd probably be best to refactor ClassCodegen to have an object wrapping the class visitor/smap/etc. with `generateField`/`generateMethod` as extension methods on that object, and avoid giving a reference to this object to `FunctionCodegen` - this would completely prevent this sort of error, but such a refactoring is not trivial due to the fact that local classes can be anywhere within function bodies.

max-kammerer:
@pyos is this pr dependes on 
JVM_IR: reuse same bodies for suspend funs and $$forInline versions pyos* 
JVM_IR: remove $$forInline checks from ExpressionCodegen pyos* 
commits or not?
I'm observing them in git log before pr ones

max-kammerer:
So class layout (method order) would also depend on order of inline function calls?
Than removing simple inline call could affect class file. And it will affect Kotlin IC if proto affected (not sure how deal Java/Scala/... IC with such changes)

pyos:
No, generating a method node for an inline call only inserts it into a map, then ClassCodegen will arrange them in the correct order (see generateMethod).

max-kammerer:
oh..indeed. Thank you!

max-kammerer:
Could you also add next test:
class A {
 inline fun a()  = b()
 inline fun c() = d()
}

class B {
   inline fun b() = c()

   inline fun d() = "OK"
}

}

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -82,21 +79,25 @@ open class ClassCodegen protected constructor(
...
+            state.classFileVersion,
+            irClass.flags,
+            signature.name,
+            signature.javaGenericSignature,
+            signature.superclassName,
+            signature.interfaces.toTypedArray()
+        )
     }
 
+    var writeSourceMap: Boolean = withinInline
```
max-kammerer, line 99: `internal`

max-kammerer:
@pyos Please take a look into `Bytecode tool window` (`Show Kotlin bytecode` action):
next simple cases is not working with this pr:
```
fun main() {
    inlineFunCall()
}
```

max-kammerer:
@pyos Oh sorry looks like it's unrelated problem with bytecode window

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -83,11 +83,10 @@ internal val DeclarationDescriptorWithSource.psiElement: PsiElement?
 fun JvmBackendContext.getSourceMapper(declaration: IrClass): DefaultSourceMapper {
     val sourceManager = this.psiSourceManager
     val fileEntry = sourceManager.getFileEntry(declaration.fileParent)
-    check(fileEntry != null) { "No PSI file entry found for class: ${declaration.dump()}" }
```
max-kammerer, line 86: Could you clarify this change?

pyos: Otherwise, FunctionCodegen cannot create per-function SMAPs if the class is in a synthetic file (a multifile facade).

I actually don't know what `getFileEntry` is for: synthetic files *do* have entries, which are accessible as `declaration.fileParent.fileEntry`, and I might be wrong, but I believe for non-synthetic files `getFileEntry(fileParent)` and `fileParent.fileEntry` should be the same.

max-kammerer: Looks like we need smap test with multifile facades (ideally with 2 parts each with inline method but I'm not sure that framework could support such thing).

pyos: Not sure what the point is, since at least right now nothing is inlined into multifile facades and thus they do not have SMAPs (IrSourceCompilerForInline never sets writeSourceMap to true).

max-kammerer: Yes, but i just want to test that part ones would be presented. 

Looks like there was nonsence in 'ideally with 2 parts each with inline method but I'm not sure that framework could support such thing'

max-kammerer: @pyos Maybe I'm a little overweight here, sorry.

max-kammerer:
@pyos Could you rebase on master?

max-kammerer:
Please also add smap check inside 'classCycle.kt'

pyos:
done

max-kammerer:
Thank you for pr! And sorry for very long review!

PR #3058, opened 2020-02-04
JVM IR: Handle external static declarations in companion objects

sfs:


sfs:

sfs:

The PR simplified nicely, after @gogabr's changes in 182e1c1 :)
I didn't even notice that this phase runs before `SyntheticAccessorLowering`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStaticAnnotationLowering.kt
```diff
@@ -60,70 +63,59 @@ private class CompanionObjectJvmStaticLowering(val context: JvmBackendContext) :
...
+            }
         }
-
     }
 
-    private fun createProxy(target: IrSimpleFunction, irClass: IrClass, companion: IrClass): IrSimpleFunction {
-        return buildFun {
-            updateFrom(target)
+    private fun IrClass.addProxy(target: IrSimpleFunction, companion: IrClass, isStatic: Boolean = true) =
+        addFunction {
```
udalov: Did you omit `updateFrom` here intentionally? It looks like that might cause some flags from the declaration not propagated to the proxy, such as `isOperator`, `isSuspend`, etc. I'm not totally sure what's the correct behavior here

sfs: Yes. The previous code manually set all of the other flags to `False`. Apart from `isOperator`, the two implementations end up doing the same thing and I honestly don't know what we should be doing about most of these flags.

As far as I can tell we write the metadata based on the class descriptors, so I'm not sure whether these flags even end up in the output.

udalov:
Thanks!

PR #3060, opened 2020-02-05
KT-36171 intention "Replace 'get' call with indexing operator" works incorrectly with spread operator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36171

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceGetOrSetInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceGetOrSetInspection.kt
@@ -48,8 +48,7 @@ class ReplaceGetOrSetInspection : AbstractApplicabilityBasedInspection<KtDotQual
         if (callExpression.typeArgumentList != null) return false
         val arguments = callExpression.valueArguments
         if (arguments.isEmpty()) return false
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/conventionNameCalls/ReplaceGetOrSetInspection.kt
```diff
@@ -45,19 +45,19 @@ class ReplaceGetOrSetInspection : AbstractApplicabilityBasedInspection<KtDotQual
         val callExpression = element.callExpression ?: return false
         val calleeName = (callExpression.calleeExpression as? KtSimpleNameExpression)?.getReferencedNameAsName()
         if (calleeName !in operatorNames) return false
+        if (callExpression.typeArgumentList != null) return false
+        val arguments = callExpression.valueArguments
+        if (arguments.isEmpty()) return false
+        if (arguments.any { it.isNamed() }) return false
+        if (calleeName == OperatorNameConventions.GET && arguments.first().isSpread) return false
```
mglukhikh: What if not the first but the second or third argument is spread?

mglukhikh:
Ok, I will recheck and merge it.

mglukhikh:
Merged manually 👍

PR #3061, opened 2020-02-05
KT-36357 "Lift assignment out of 'if'" breaks code for oneliner function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36357


mglukhikh:
Merged 👍

PR #3062, opened 2020-02-06
KT-36369 "To raw string literal" intention is not available if string content starts with newline symbol \n

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36369


mglukhikh:
Merged 👍

PR #3063, opened 2020-02-06
JVM_IR: Relax bailout condition for SAM conversion generation.

jsjeon:
Previously, resolved call is expected to have SAM converted argument map
if SamConversionPerArgument is enabled. However, if SAM argument is
followed by vararg parameter and an array _without_ a spread operator is
passed, New Inference left a type mismatch error on a resolved call, which
made that resolved candidate filtered out. Instead, another resolution
result wihtout SAM converted arugment map will be provided.

All other logics, such as adding SAM conversion type op and lowering SAM
conversion and/or call references, are already there for resolved calls
without SAM converted argument map. This change just relaxes the bailout
condition so that array passed to vararg after SAM argument can be
handled in JVM IR.


dnpetrov:
I'll add corresponding psi2ir test, run tests, and push manually.

In future, when you modify psi2ir, please, add corresponding tests under testData/ir/irText. There are other back-ends besides JVM_IR that depend on psi2ir, and since SAM conversion is a "pure Kotlin" feature in 1.4+, we'd like to make sure we have consistent behavior across the board.

dnpetrov:
Pushed to master ( 5a49ccac760d479fd3830d4dc0823448c9d952ea )

jsjeon:
I see. Thank you for the pointer (to the corresponding tests), and will keep that in mind when modifying upper layers.

PR #3065, opened 2020-02-06
JVM IR: Fix references to const properties in multi-file facade classes

sfs:
When we have a publicly visible const property in a multi-file part class we move it to the facade class. Unfortunately, this happens long after we have generated property reference classes, which now refer to a non-existing property and break.

There's really no good way of fixing this right now. Ideally we'd only generate the reflection information (`getSignature`, `getOwner`, `getName`) during codegen, when we know what the signatures ought to be, but that would be a much larger refactoring. So for now, I'm just fixing up the class references as part of `GenerateMultifileFacades`. This is messy, but it works...

The PR also contains a bugfix for the `getSignature` method in `PropertyReferenceLowering`, which was broken for multi-file classes.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/GenerateMultifileFacades.kt
```diff
@@ -262,3 +267,26 @@ private class UpdateFunctionCallSites(
         }
     }
 }
+
+private class UpdateConstantFacadePropertyReferences(private val context: JvmBackendContext) : ClassLoweringPass {
```
punzki: Do you need this as a ClassLoweringPass? You're creating it only to call lower(). It can just be a private method, or merge it with the anonymous IrElementTransformerVoid.

udalov:
Merged as a part of #3068.

PR #3066, opened 2020-02-06
KT-36406 "To ordinary string literal" intention adds unnecessary escapes to characters in template expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36406


t-kameyama:
81d01a8f8e70347f654439e5214a6e1ba0d3a77d

PR #3067, opened 2020-02-06
[JVM IR] Do not add nullability annotation for functions that return platform types

punzki:
This test breaks code I saw that uses Dagger looks like this:

```kt
@Module
object TestModule {
  // Returns platform type and is annotated with @Nullable in IR without this fix
  @Provides
  fun provideFoo() = Mockito.mock(Foo::class.java)  
}

// Broken during injection as it expects not-nullable Foo but provided Foo is nullable
class Bar @Inject constructor(val foo: Foo)
```

Also added IR equivalent of `AbstractBytecodeListingTest`. @udalov @max-kammerer Should we make it a priority to get these tests unmuted for JVM IR?

@gogabr You originally rewrote `AnnotationCodegen` for IR (451cda79debdd0d13a3bd3b849f49047d9fd0a68) but left out the the logic added in f16dcdd8a904c88eda90e9e5c620370dd1274298. That's the logic I added back in here to make it work. Is there a reason it was originally left out in the rewrite?


dnpetrov:

dnpetrov:
Pushed manually
06b6477d044d10779f68189909ed76346e97f78c
b782e8f0f012abb9c5e2be1de486e3c30cc84062

PR #3068, opened 2020-02-07
JVM IR: Fix flags on multi-file part classes

sfs:
In particular, they should be package-private synthetic. This PR adds a new origin and does not add new special cases in codegen (but it does need to touch a bit of code which used the `FILE_CLASS` origin directly). I have an alternative implementation which just adds a special case in `ClassCodegen` instead, but that seems more messy.

Commit 578c645 is from #3065, which is needed because otherwise we end up breaking `testConstPropertyReferenceFromMultifileClass` (which would access a reference to a package private class from a public class, even if it is not used in the test itself).


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt
```diff
@@ -37,6 +37,7 @@ interface IrDeclarationOrigin {
     object MOVED_EXTENSION_RECEIVER : IrDeclarationOriginImpl("MOVED_EXTENSION_RECEIVER")
 
     object FILE_CLASS : IrDeclarationOriginImpl("FILE_CLASS")
+    object MULTIFILE_PART_CLASS : IrDeclarationOriginImpl("MULTIFILE_PART_CLASS", isSynthetic = true)
```
udalov: We have another compilation mode of multifile class facades and parts, enabled by `-Xmultifile-parts-inherit` and used for example in kotlin-stdlib, where instead of generating delegating "bridges" in the facade, we just sort all the parts and inherit them from each other, and then inherit the facade from the last part.

In that mode I believe the parts should _not_ be synthetic, since otherwise methods declared there would not be visible from Java. I'm not sure why there are no tests on this (maybe because we use kotlin-stdlib itself as a test case).

sfs: You're right, I didn't think about this at all. I've added the special case for `-Xmultifile-parts-inherit` and also added a test for this.

I've also renamed the corresponding origin to `SYNTHETIC_FILE_CLASS`, since not all multifile parts will use this origin now.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/MainMethodGenerationLowering.kt
```diff
@@ -40,7 +41,7 @@ internal class MainMethodGenerationLowering(val context: JvmBackendContext) : Cl
 
     override fun lower(irClass: IrClass) {
         if (!context.configuration.languageVersionSettings.supportsFeature(LanguageFeature.ExtendedMainConvention)) return
-        if (irClass.origin != IrDeclarationOrigin.FILE_CLASS) return
+        if (!irClass.isFileClass) return
```
udalov: I wonder if parameterless main in JvmMultifileClass works in the old backend, and if there are tests on it?

sfs: Good question. There doesn't seem to be a test for it. I just tried it and it works normally (both the parameterless main and the version with arguments get delegated from the facade class) *unless* you compile with `-Xmultifile-parts-inherit`.

I've reported the latter as [KT-36492](https://youtrack.jetbrains.com/issue/KT-36492)

udalov:
Thank you!

PR #3069, opened 2020-02-07
JVM_IR: partially fix inline methods using captured crossinline lambdas

pyos:
The fields containing crossinline lambdas should be package-private to avoid generating synthetic accessors, which break object regeneration.

Note that the inline methods cannot actually be called, as call sites will attempt to read the captured lambda from a field through a *copy* of the local containing the object, so these reads will not be inlined, causing an exception at runtime:

    inline fun f(crossinline g: () -> Unit) = object : I {
        inline fun h() = g()
        // effectively `val tmp = this; return tmp.$g()`:
        override fun run() = h()
    }

    f {}.run() // NoSuchFieldError: $g

This particular example can be fixed by reusing locals for receiver parameters in IrInlineCodegen, but explicitly assigning `this` to another variable and calling an inline method on it will break it again. (This is only applicable to the JVM_IR backend, as the non-IR one fails to generate `f` at all for some other reason.)


max-kammerer:
Please rebase on master and regenerate compiler test. There were some changes in JS infrastructure around inline tests

pyos:
done

max-kammerer:
Thank you!

PR #3070, opened 2020-02-07
[FIR] fixed overridden symbols of "invoke" in KFunction

juan-chen:
IR expects overridden symbols of "invoke" in KFunction to be "invoke" in the corresponding Function classes. The current implementation doesn't set overriddenSymbol.


mglukhikh:
Merged manually, thank you.

PR #3071, opened 2020-02-08
KT-36414 "Convert to anonymous object" intention does not insert inferred parent type parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36414


t-kameyama:
https://github.com/JetBrains/intellij-community/pull/1614

PR #3072, opened 2020-02-08
[Fir] Fix array iteration

juan-chen:
* Fixed ArrayIterator
  The current implementation tries to get a type argument from the array class, which would fail. See
[ArrayIterator](https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ArrayIterator.kt#L32) and [IrTypeMapper](https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrTypeMapper.kt#L135)

* Fixed NoSuchMethod for Iterator::next, same issue with invoke


juan-chen:
Psi2Ir doesn't have this problem because ForLoopsLowering lowers `Array::iterator()` after pattern-matching the exact statements Psi2Ir generates for loops, and thus no need to call the intrinsic iterator function.

ForLoopsLowering doesn't do anything to code generated by the new frontend because the code is different from what ForLoopsLowering expects.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
```diff
@@ -444,3 +444,6 @@ fun CallableId.isInvoke() =
     packageName == KotlinBuiltIns.BUILT_INS_PACKAGE_FQ_NAME
             && className?.asString()?.startsWith("Function") == true
             && callableName.asString() == "invoke"
+
+fun CallableId.isArrayOperator() =
```
mglukhikh: Why this function is called so strange? As I can see, it has nothing relevant with arrays, as well as with operators. `isIteratorNext`?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ArrayIterator.kt
```diff
@@ -29,11 +30,14 @@ object ArrayIterator : IntrinsicMethod() {
         signature: JvmMethodSignature,
         context: JvmBackendContext
     ): IrIntrinsicFunction {
-        val owner = context.typeMapper.mapClass(expression.symbol.owner.parentAsClass)
+        val isPrimitive = ((expression.type as? IrSimpleType)?.arguments?.size ?: 0) == 0
```
mglukhikh: This change can break other back-ends, we should at least check it by remote run (I will do). In general, it's better to avoid changes in back-ends when fixing FIR2IR. I think @udalov can tell us more precise whether this change is acceptable or not.

udalov: I agree. It's a bit suspicious in general to adapt the JVM IR backend to two different frontend models. I'd strongly prefer to keep the frontend model (or at least the part that the backend works with) the same between psi2ir and fir2ir.

In this case, the change looks safe but I'm very curious why is it needed, because it may reveal deeper problems in fir2ir.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
juan-chen: done

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/ArrayIterator.kt
juan-chen: Updated the commit to change fir2ir, so that ArrayIterator can stay as it is.

It seems the backend has some baked-in knowledge of how psi2ir translated to IR, which is often not followed by fir2ir. I'll make fir2ir follow what's expected by the backend, to avoid changing the backend.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
MOHTOP100: شكراً لك اخي الكريم
shkraan lak 'akhi alkarim

mglukhikh:
Merged manually, thank you 👍

PR #3073, opened 2020-02-09
[Fix] Don't set the -noverify flag if JDK >= 13

itsmepetrov:
Fixes the following warning:

```
OpenJDK 64-Bit Server VM warning: Options -Xverify:none and -noverify were deprecated in JDK 13 and will likely be removed in a future release.
```

### DIFF
--- a/compiler/cli/bin/kotlinc
+++ b/compiler/cli/bin/kotlinc
@@ -66,7 +66,14 @@ then
     kotlin_app=("${KOTLIN_HOME}/lib/kotlin-runner.jar" "org.jetbrains.kotlin.runner.Main")
 else
     [ -n "$KOTLIN_COMPILER" ] || KOTLIN_COMPILER=org.jetbrains.kotlin.cli.jvm.K2JVMCompiler
-    java_args=("${java_args[@]}" "-noverify")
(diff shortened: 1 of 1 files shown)


udalov:
Hi! Thanks for the contribution.

I'd prefer if we read the property `JAVA_VERSION` from the file `$JAVA_HOME$/release`, and parse it with a regex, with the built-in shell tools instead of running external processes (`java`, `awk`). I believe it could be faster that way.

Also, it would be nice to do something similar in `kotlinc.bat` as well. But I'm not sure how difficult it is to do in batch scripts.

udalov:
On second thought, it seems that it's already difficult to find out the Java home directory if `JAVA_HOME` is not set, and moreover, the `release` file is not guaranteed to exist on every JDK. (As a matter of fact, on Zulu JDK it's named `release.txt`; and on Amazon Corretto, it's named `version.txt` and just contains one number, as opposed to key=value ini-style config in the other JDKs.)

In any case, we can use bash's built-in regular expressions to parse the output of `java -version`, which I'll work on next ([KT-44232](https://youtrack.jetbrains.com/issue/KT-44232)). I'll also try to invent something for `kotlinc.bat`.

Thanks again for your pull request!

PR #3074, opened 2020-02-10
KT-31800 False positive "never used" with function in private val object expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31800
https://youtrack.jetbrains.com/issue/KT-20868


vladimirdolzhenko:
LGTM

PR #3075, opened 2020-02-10
KT-20844 Re-enable stack spilling during inline.

jsjeon:
`dx` has entered the final stage of sunset:
https://android-developers.googleblog.com/2020/02/the-path-to-dx-deprecation.html

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
@@ -149,12 +149,7 @@ abstract class InlineCodegen<out T : BaseExpressionCodegen>(
         }
     }
 
-    @Suppress("UNREACHABLE_CODE")
     private fun canSkipStackSpillingOnInline(methodNode: MethodNode): Boolean {
-        // TODO: Temporary disable this optimization until
-        // https://issuetracker.google.com/issues/68796377 is fixed
-        // or until d8 substitute dex
-        return false
         // Stack spilling before inline function 'f' call is required if:
         //  - 'f' is a suspend function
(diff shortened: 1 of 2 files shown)


jsjeon:
I left some comments on local builds/tests:
* https://issuetracker.google.com/issues/68796377#comment5 (and/or https://youtrack.jetbrains.com/issue/KT-20844#focus=streamItem-27-3935226.0-0)
* https://issuetracker.google.com/issues/68796377#comment6 (and/or https://youtrack.jetbrains.com/issue/KT-20844#focus=streamItem-27-3937984.0-0)

max-kammerer:
@jsjeon Do you have any statistics about `dex` usage?  If we switch this behaviour we will break clients that use obsolete `dex` build.

max-kammerer:
@jsjeon Did you tried to reproduce issue with dx and Android Gradle Plugin 3.1 or 3.2?

jsjeon:
I should have tried that combination: re-enablement, `dx`, and AGP 3.1+. It's not reproduced as mentioned at https://issuetracker.google.com/issues/68796377#comment7

I also learned how to collect related data, which took me a little while. As of this week, among 7-day active projects that opt in for statics sharing (the opt-in ratio is 1/3 or 1/4), `dx`/(`dx` + `d8`) is 2.8% (the size of data set is over 600k). But, if I add one more constraint---Kotlin project, the ratio is 0.9% (over 200k data set).

Another metric to mention is the ratio of `android.enableD8=false`, the way users can still use `dx` instead of `d8` (and that option will be sunset together with `dx`). For all versions, it's below 0.8%, but when it comes to "release" version of AGP since 3.3, it's zero. For now, AGP 3.5 is the dominant version (over 77%).

max-kammerer:
@jsjeon Thank you for pr!

PR #3077, opened 2020-02-12
KT-36504 "Extract property" suggests potentially invalid name for new property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36504


vladimirdolzhenko:
Thanks a lot @t-kameyama for your PR, I found some pitfall:

It works fine with a test data of [kt24615.kt](https://github.com/JetBrains/kotlin/blob/master/idea/testData/refactoring/introduceProperty/kt24615.kt) but it takes into account only current class scope but conflict could be with a parent class: 

```
open class Base(protected val i: Int)

class Impl(z: Int) : Base(z) {
    fun foo(): Int {
        return <selection>2 + 3 + i</selection>
    }
}
```

Extract property still produces invalid name of a new property  (`i` in this case)

Could you please improve your PR regarding this ? Thank you in advance.

vladimirdolzhenko:
Thank you so much, PR is merged

PR #3078, opened 2020-02-12
[JVM IR] Maintain KT-36625 bug compatibility between non-IR and IR backends

punzki:
...by removing `IMPLICIT_NOTNULL` casts from `IrStringConcatenation` arguments.

Also fixed an issue where `IrStringConcatenation` can be lowered into a `null` String instead of a literal `"null"` String if the lone argument was a platform type String or String with enhanced nullability and the value was null (e.g., `"${FromJava.nullPlatformString()}"`).

### DIFF
--- a/compiler/testData/codegen/box/strings/stringFromJavaPlus.kt
+++ b/compiler/testData/codegen/box/strings/stringFromJavaPlus.kt
@@ -1,0 +1,138 @@
+// IGNORE_BACKEND_FIR: JVM_IR
+// TARGET_BACKEND: JVM
+// WITH_RUNTIME
+
+// FILE: J.java
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class J {
+    public static String platformStringIsNull() {
+        return null;
+    }
+
+    @NotNull
+    public static String notNullStringIsNull() {
+        return null;
+    }
+
+    @Nullable
+    public static String nullableStringIsNull() {
+        return null;
+    }
+
+    @NotNull
+    public static String notNullStringIsNotNull() {
+        return "foo";
+    }
+
+    public static String platformStringIsNotNull() {
+        return "foo";
+    }
+
+    @Nullable
+    public static String nullableStringIsNotNull() {
+        return "foo";
+    }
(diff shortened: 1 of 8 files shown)


compiler/testData/codegen/box/strings/stringFromJavaPlus.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+// FILE: main.kt
+import kotlin.test.assertEquals
+
+fun box(): String {
+    val n = 123
+
+    // Due to KT-36625, certain concatenation calls below (marked with a comment) SHOULD fail a nullability check but do not.
```
dnpetrov: Minor: not clear if it's really "SHOULD". Language design formalities are not settled in this case.
I'd change this comment to something like 
```
// Null check behavior in string concatenation might change depending on language design decision for KT-36625.
```

compiler/testData/codegen/box/strings/stringFromJavaPlus.kt
```diff
@@ -0,0 +1,138 @@
...
+        return "foo";
+    }
+
+    @Nullable
+    public static String nullableStringIsNotNull() {
+        return "foo";
+    }
+}
+
+// FILE: main.kt
```
dnpetrov: Minor: I usually give the "main" file in the multi-file test the same name as the test file (`stringFromJavaPlus.kt` in this case). Enables navigation from stack trace in IDEA.

dnpetrov:
Looks ok to me. I'll run tests and push manually.

dnpetrov:
Pushed manually ( ba606147c95edb9e3988654fb44bbe1ac4b24bfe ).

punzki:
@dnpetrov Thanks! I don't know if you saw this but added an extra commit on the branch to clarify the comments about KT-36625. Feel free to merge, ignore, or replace with your own.

dnpetrov:
Thank you! Yes, I see, I've missed it. I'll push it as a separate commit.

dnpetrov:
Done ( 89cf32eccc813b10b506799ae132f1fb5dbf03a9 )

PR #3079, opened 2020-02-12
Fix KT-36657

kovalpgoogle:
https://youtrack.jetbrains.com/issue/KT-36657

Don't stop process (don't return false) if options.isSearchForTextOccurrences is false

options.isSearchForTextOccurrences == false is not valid reason to stop whole find usages process.

Before fix:
options.isSearchForTextOccurrences is false ->
 KotlinFindUsagesHandler#searchTextOccurrences returns false ->
  KotlinFindUsagesHandler#processElementUsages returns false

It leads to problems in places where processElementUsages is called, e.g
code inside com.intellij.find.findUsages.FindUsagesManager#createUsageSearcher never
reaches CustomUsageSearcher.EP_NAME.

It means that you can't provide any custom Usages for Kotlin elements through CustomUsageSearcher.


vladimirdolzhenko:
@kovalpgoogle is it possible to add some tests ?

vladimirdolzhenko:
LGTM, Remote Run passed. 

Merge it manually: we have some rules - the 1st line is a short description of the problem problem, last line should be like `#KT-XXXX Fixed` to be picked up by YouTrack robot to change ticket state and mark build no.

PR #3081, opened 2020-02-12
[JVM IR] Use `append(Char)` for 1-length string literals in string templates and plus concatenations

punzki:
This is slightly more efficient and mirrors the behavior of the non-IR backend for templates (but not for plus concatenations).

Well, in theory it is more efficient but in practice, any performance benefit is likely negligible. I just noticed the difference in bytecode and I figured why not make the calls match up better? 🙂

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmStringConcatenationLowering.kt
@@ -15,13 +15,12 @@ import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
 import org.jetbrains.kotlin.backend.jvm.ir.JvmIrBuilder
 import org.jetbrains.kotlin.backend.jvm.ir.createJvmIrBuilder
 import org.jetbrains.kotlin.backend.jvm.lower.inlineclasses.unboxInlineClass
-import org.jetbrains.kotlin.ir.builders.IrBuilderWithScope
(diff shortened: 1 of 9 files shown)


punzki:
Hmm only after uploading this PR did I consider doing this in PSI2IR. Looking at the implementation of `append(Char)` in Kotlin/Native, this could benefit K/N more significantly.

dnpetrov:
Looks like it has conflicts with ba606147c95edb9e3988654fb44bbe1ac4b24bfe
( [JVM IR] Maintain KT-36625 bug compatibility between non-IR and IR... )
Please, rebase on master.

dnpetrov:
@punzki Please, rebase on master, resolve related conflicts, and update PR.

punzki:
@dnpetrov Thanks for the reminder. I am still looking into doing this in PSI2IR, or doing this in `FlattenStringConcatenationLowering` which is a common lowering, and I'll update the PR when done.

dnpetrov:
Ok, ping me when it's ready.

punzki:
@dnpetrov Sorry for the delay, updated the PR. I attempted to do this PSI2IR, and in `FlattenStringConcatenationLowering`, but on both cases I had to add these same changes in `JvmStringConcatenationLowering` because `FoldConstantLowering` was converting the Char concatenation arguments back to Strings.

punzki:
@dnpetrov Can you please take a look?

punzki:
@dnpetrov Ping, could you look at this?

dnpetrov:

dnpetrov:
Pushed manually.

4234fa79c0746e0aae8b2e9ca363fd2afe6cb81a

PR #3080, opened 2020-02-12
JVM IR: Avoid redundant coercions in ExpressionCodegen 

sfs:
CC @kandersen 

This PR changes `ExpressionCodegen` to only coerce `PromisedValues` right before materialization. This is closer to what the JVM backend did and fixes some tests. It's also cleaner. Here's what's changed:
- We now produce correct owners for field accesses. Previously we would sometimes produce owners referring to a superclass of the current class. The JVM was fine with this, *unless* you also introduce an unnecessary CHECKCAST...
- The java Class to KClass coercions are now properly restricted to annotation classes.
- We no longer produce redundant casts for constants (`bipush 8; i2l`, etc...)
- We generally produce fewer casts, which probably fixes more Java interoperability bugs. Since we cannot reliably identify platform types in Kotlin due to erased generics and the absence of platform types in the IR backend, we have to be careful about introducing casts. Ideally, we should only add casts when it is necessary to produce valid bytecode.

PR #3083, opened 2020-02-13
JVM IR: Don't optimize comparisons between boxed primitive types and null

sfs:
We currently optimize null checks on primitive types as part of `JvmOptimizationLowering` in the JVM IR backend. However, this is only sound if we are comparing an *unboxed* primitive against null, which we cannot easily determine in the lowering.

In particular, a comparison of the form `x.f() == Null` where `x.f()` returns a primitive type may not be vacuous if we are dealing with a boxed value produced in Java code. And as shown in `testUnsafeNullCheckWithPrimitive` we cannot reliably identify Java values due to type erasure.

This PR just postpones the optimization to codegen time. Funnily enough, this optimization already exists in the `Equals.kt` intrinsic. However, before #3080 this was not enough, since there were still cases where we would produce NPEs due to unnecessary casts.

The first three commits in this PR are from #3080, only 3277ef7 is new.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
@@ -10,7 +10,6 @@ import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irBlock
 import org.jetbrains.kotlin.backend.common.phaser.makeIrFilePhase
 import org.jetbrains.kotlin.backend.jvm.JvmBackendContext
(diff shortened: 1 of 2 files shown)


dnpetrov:
Looks like a duplicate of https://github.com/JetBrains/kotlin/pull/3080.
Or am I missing something important?

sfs:
Commit 3277ef7 is new, but depends on #3080, which is why the PR includes that branch.

dnpetrov:
Ok then, let's deal with #3080, and then maybe rebase 3277ef710f11e16de8f366cb6c1f87b57eca11a3 on master and run tests, is that ok for you?

sfs:
Yes, that would be great!

dnpetrov:
@sfs #3080 is in master. Please, rebase on master and update PR.

sfs:
@dnpetrov Thanks for the reminder! Rebased on master.

dnpetrov:

dnpetrov:
Pushed manually.
4b954c347a189dd61eb5ddb98f502f525c749a95

PR #3084, opened 2020-02-13
KT-20907 Secondary constructor is marked as unused by IDE when called by typealias 

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20907


vladimirdolzhenko:
LGTM 👍

PR #3085, opened 2020-02-14
KT-29870 Inline variable doesn't handle 'when' subject val correctly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-29870


vladimirdolzhenko:
LGTM

PR #3086, opened 2020-02-14
JVM, JVM IR: Fix assertion status for regenerated anonymous objects

sfs:
Right now, we're setting the assertion status of regenerated anonymous objects based on the regenerated class name in the JVM backend and based on the top-level enclosing class name at the declaration site in the JVM IR backend.

Both are wrong, as we should be determining the assertion status based on the top-level enclosing class at the *call-site*. This is what we already do for normal inline functions, at least in the absence of other errors.

This PR is an attempt to fix this behavior and while it passes some additional tests I'm not sure that it is completely correct yet. In particular, I'm using `inliningContext.root.callSiteInfo.ownerClassName` to get at the name of the enclosing class. I have no idea if this always refers to the *top-level* enclosing class though.

Although, if that's not the case then there are more bugs of this form in both backends...

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/AssertCodegenUtil.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/AssertCodegenUtil.kt
@@ -6,6 +6,8 @@
 package org.jetbrains.kotlin.codegen
 
 import org.jetbrains.kotlin.codegen.coroutines.createCustomCopy
+import org.jetbrains.kotlin.codegen.optimization.common.InsnSequence
+import org.jetbrains.kotlin.codegen.optimization.common.findPreviousOrNull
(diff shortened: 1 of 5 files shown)


sfs:
It looks like I managed to break something right before committing. Let me just fix this...

sfs:
Turns out it was the opposite.

I changed `testJvmCrossinlineLambdaDeclarationSite` to set the assertion status for the `Checker` classes instead of for the `test` package. That broke the test, or rather made it clear that the test was broken to begin with. Previously assertions were always disabled since the `Checker` classes were not in the `test` package at all. The test only passed since all the checks for `ShouldBeEnabled` also tested that assertions were *disabled*.

I've fixed the test.

sfs:
One more change, to make the test for the `$assertionsDisabled` initializer more precise.

ilmirus:
Manually rebased and pushed to master.

PR #3087, opened 2020-02-14
Added unreachable precondition

Fleshgrinder:


Fleshgrinder:
Closing, it does not seem as if there is interest in this and it probably isn't as useful as it is in Rust.

PR #3088, opened 2020-02-15
[JVM IR] Fix issue where fields are not being set to their default values within initializer blocks

punzki:
The issue occurs in code like this:
```kt
class C {
  var b = true
  init {
    b = false   // Missing PUTFIELD for this statement
  }
}
```

Added a new statement origin for field initialization (at declaration) instead of relying on `origin == null` in `ExpressionCodegen` to determine whether to generate the initializations.

This was unintentionally broken in d68a1898d0135e947ecdbebd39003b8ceaf8f57f. We could also preserve the statement origin in `JvmOptimizationLowering` when creating the `IrSetField` but I think it is cleaner and more robust to differentiate initialization-at-declaration. (cc @madsager)


dnpetrov:

dnpetrov:
Pushed manually.
64141b8b38bde486478aa9fb72c1bb671ad89782

PR #3089, opened 2020-02-15
KT-36707 False positive redundant companion object on calling companion object members

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36707


vladimirdolzhenko:
LGTM. Thanks a lot @t-kameyama

PR #3090, opened 2020-02-16
KT-36685 "Convert to a range check" transform hex range to int if it is compared with "Less" or "Greater"

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36685


vladimirdolzhenko:
Thanks @t-kameyama, LGTM

PR #3092, opened 2020-02-17
Added lastMaxBy function to stdlib

acefalobi:
This is a fix to the problem brought up in this [issue](https://youtrack.jetbrains.com/issue/KT-34966)

This commit adds a new standard library function `lastMaxBy` that replicates the functionality of `maxBy`, but returns the last maximum element instead of the last. It also adds new tests and samples for this function

### DIFF
--- a/idea/tests/org/jetbrains/kotlin/idea/slicer/SlicerLeafGroupingTestGenerated.java
+++ b/idea/tests/org/jetbrains/kotlin/idea/slicer/SlicerLeafGroupingTestGenerated.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
  */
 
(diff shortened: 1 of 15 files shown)


ilya-g:
We had a discussion about the similar problem here: https://discuss.kotlinlang.org/t/should-max-maxby-maxwith-return-the-last-largest-element/2139/11
However we haven't found so far a use case motivating enough to change `max` function behavior or introduce overloads like `lastMax(By)`.

acefalobi:
I think a motivating use case would be like the one mentioned in the issue where you have a list sorted by a predicate and wanted to get the maximum element of the list by another predicate.

If I had a list `[(C, 2), (A, 1), (B, 2)]` and then sorted by the first parameter, I would get `[(A, 1), (B, 2), (C, 2)]`. Running `maxBy` on the new sorted list with the second parameter would return `(B, 2)`, even though the desired result would be to keep the previous order and return the maximum element from both sorts which would be `(C, 2)`.

ilya-g:
Sorry, I don't get how `lastMaxBy` would help here. The first sort could have reordered elements in an arbitrary way, so if `(C, 2)` is the desired result, it could be either the first maximum, or the last one, or neither of those in that sorted list.

Also using the elements in the motivating example like `(A, 1)` doesn't help either. Is your real use case about sorting pairs of letters and numbers?

I suggest to move the discussion about use cases to that forum thread, since it isn't related to the code in this PR directly, and in order to get more opinions there.

acefalobi:
Moving to the discussion thread then

acefalobi:
@ilya-g seems like the discussion is inconclusive. Should I close the PR?

PR #3095, opened 2020-02-17
KT-36735 Inspection 'Replace 'toString' with string template' miss curly braces and generates wrong code for constructor calls

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36735


vladimirdolzhenko:
LGTM

PR #3098, opened 2020-02-18
JVM IR: Fix compilation of interface methods in inline classes

sfs:
The JVM backend always generates `-impl` stubs for interface methods in inline classes (except for a bug with `@JvmDefault` methods). This PR changes the JVM IR backend to adopt the same convention. For more details, see [KT-27077](https://youtrack.jetbrains.com/issue/KT-27077).

The upshot is that this is required when mixing the JVM and JVM IR backends, as the latter always tries to call an `-impl` stub. Apart from that, this is also necessary for binary compatibility. I.e., it should be possible to move an implementation in an inline class into a super interface without recompiling.


udalov:
Looks good, thanks!

PR #3099, opened 2020-02-18
[JVM IR] Refactor PromisedValue...

kandersen:
... for an _even leaner_ codegen!

This PR combines the notion of coercing and materializing `PromisedValue`s such that we only generate instructions when materializing the value, at the discretion of the value itself.

The interface of `PromisedValue` is now `materializeAt` and `discard`, with the same contract as before: you must call one or the other before generating more code.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
@@ -13,7 +13,6 @@ import org.jetbrains.kotlin.codegen.AsmUtil
 import org.jetbrains.kotlin.codegen.StackValue
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.parentAsClass
-import org.jetbrains.kotlin.resolve.jvm.AsmTypes
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.org.objectweb.asm.Label
 import org.jetbrains.org.objectweb.asm.Type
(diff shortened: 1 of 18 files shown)


kandersen:
CC @sfs

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -61,103 +94,73 @@ abstract class BooleanValue(codegen: ExpressionCodegen) : PromisedValue(codegen,
...
+}
+
+val IrType.unboxed: IrType
+    get() = InlineClassAbi.getUnderlyingType(erasedUpperBound)
+
+// A Non-materialized value of Unit type that is only materialized through coercion.
+val ExpressionCodegen.unitValue: PromisedValue
+    get() = object : PromisedValue(this, Type.VOID_TYPE, context.irBuiltIns.unitType) {
+        override fun discard() {}
+    }
```
udalov:
```suggestion
    get() = MaterialValue(this, Type.VOID_TYPE, context.irBuiltIns.unitType)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/AndAnd.kt
```diff
@@ -25,6 +25,13 @@ object AndAnd : IntrinsicMethod() {
             arg1.accept(codegen, data).coerceToBoolean().jumpIfTrue(target)
             mv.visitLabel(stayLabel)
         }
+
+        override fun discard() {
+            val end = Label()
+            arg0.accept(codegen, data).coerceToBoolean().jumpIfFalse(end)
+            arg1.accept(codegen, data).discard()
+            codegen.mv.visitLabel(end)
```
udalov:
```suggestion
            mv.visitLabel(end)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
udalov: I think it makes more sense to use `MaterialValue` here, since it can be considered a "materialized" value of `VOID_TYPE`, a type, any value of which takes zero space on the stack.
```suggestion
    get() = MaterialValue(this, Type.VOID_TYPE, context.irBuiltIns.unitType)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/AndAnd.kt
kandersen: Found the same in OrOr.kt

kandersen:
Revised and pushed, thanks for having a look!

udalov:
Merged in https://github.com/JetBrains/kotlin/commit/6e40117116eeaf43893dae54fbe1312a4b13bb95. Thank you!

PR #3100, opened 2020-02-18
JVM IR: Fix visibility for $assertionsDisabled fields

sfs:
It should be package private, as stated in `compiler/testData/codegen/bytecodeText/assert/jvmNestedClass.kt`.


sfs:
I just noticed that we don't move the `$assertionsDisabled` field to interface `DefaultImpls` classes, which becomes an issue with this PR. Let me just fix this...

sfs:
Rebased on master and moved `$assertionsDisabled` field in interfaces to `DefaultImpls`.

PR #3101, opened 2020-02-19
KT-15723 Add 'Convert to value' quickfix for property containing only getter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15723


vladimirdolzhenko:
LGTM, Thanks a lot @t-kameyama

PR #3104, opened 2020-02-19
JVM IR: Fix inline class constructors taking default values of inline class type

sfs:


sfs:
Rebased and updated ignore directive (the added test fails with Fir).

udalov:
Thanks!

PR #3106, opened 2020-02-19
[JVM IR] Improve debugging of inline $default methods

kandersen:
Improves the stepping behaviour around inline methods with default
arguments. In particular, we now accurately step through the
evaluation of default arguments, but do _not_ spuriously show the exit
from the $default handler.


kandersen:
Depends on #3099

kandersen:
Rebased now that #3099 landed.

max-kammerer:
@kandersen Please rebase on master, GH reports conflict

kandersen:
> @kandersen Please rebase on master, GH reports conflict

Done!

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
```diff
@@ -206,9 +207,9 @@ abstract class InlineCodegen<out T : BaseExpressionCodegen>(
                 ?: error("No stack value for continuation parameter of suspend function")
     }
 
-    protected fun inlineCall(nodeAndSmap: SMAPAndMethodNode, inlineDefaultLambda: Boolean): InlineResult {
+    protected fun inlineCall(nodeAndSmap: SMAPAndMethodNode, inlineDefaultLambda: Boolean, isSyntheticCall: Boolean): InlineResult {
         assert(delayedHiddenWriting == null) { "'putHiddenParamsIntoLocals' should be called after 'processAndPutHiddenParameters(true)'" }
-        defaultSourceMapper.callSiteMarker = CallSiteMarker(codegen.lastLineNumber)
+        if (!isSyntheticCall) defaultSourceMapper.callSiteMarker = CallSiteMarker(codegen.lastLineNumber)
```
max-kammerer: If I undestand correctly it should work in case of inlining original function to $default companion one? If so maybe check container origin in this case instead of `isSyntheticCall = expression.startOffset == UNDEFINED_OFFSET`?

max-kammerer:
RR fails with
```
          org.jetbrains.kotlin.codegen.ir.FirBlackBoxCodegenTestGenerated$Reified > testDefaultJavaClass FAILED
10:39:34
              java.lang.AssertionError: com.android.tools.r8.CompilationFailedException: Compilation failed to complete
10:39:34
                at com.android.tools.r8.utils.t.a(:55)
10:39:34
                at com.android.tools.r8.D8.run(:11)

```

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
kandersen: I believe this is when inlining the $default companion into the call site of the original. That is, in the source is a call `foo`, it gets redirected to a call to `foo$default` and this is then inlined.

max-kammerer: So on inlining `foo$default` we shouldn't reset markers, that looks strange. I thought that we do smth wrong on inliniing foo into foo$default

max-kammerer: Even in such case i think  it's possible  to check origin for FUNCTION_FOR_DEFAULT_PARAMETER. 
But I still confused in difference between general and $default function processing in ir-inliner

compiler/testData/debug/stepping/assertion.kt
```diff
@@ -0,0 +1,45 @@
...
+fun box(): String {
+    massert(true)
+    massert(true) {
+        "test"
+    }
+
+    return "OK"
+}
+
+// LINENUMBERS
```
max-kammerer: Test data of this test is affected by just merged #3157

jsjeon: I was about to say this! :)

kandersen: Huh, you are absolutely right, though I don't think it's an improvement :) Steps are as they should be, will update.

jsjeon: Well, for (inline) tests where multiple modules/files are involved, source names help you understand to which each line belongs.

max-kammerer:
@kandersen Please also take a look into https://github.com/JetBrains/kotlin/commit/186302d84e71faec6028ee2d67f091e3a1276689
from #3056 
Maybe it's possible just pass custom SourceMapper

max-kammerer:
...It was extracted to #3204

kandersen:
> @kandersen Please also take a look into [186302d](https://github.com/JetBrains/kotlin/commit/186302d84e71faec6028ee2d67f091e3a1276689)
> from #3056
> Maybe it's possible just pass custom SourceMapper

Will check it out, thanks!

kandersen:
Rebased on master, unmuted 3 more tests. We think there's a separate but related issue with source locations on lambdas. I'll start that work in a separate PR with some tests to illustrate the problem.

max-kammerer:
@pyos Could you take a look (there are a lot of upcoming and already merged changes in source mapper related logic from you). Could problem be fixed without additional flag but maybe by passing specially tuned source mapper?

pyos:
I don't think so. The fix here is not complete: it does remove incorrect call site markers, but it also discards everything in `f` when generating `f$default`. For a complete fix, `NestedSourceMapper` created in `inlineCall` should be told to keep the markers with the `sameFile` parameter, like in `InlineCodegenForDefaultBody`.

I wonder if adding a flag for that would make `InlineCodegenForDefaultBody` unnecessary, though.

max-kammerer:
> I wonder if adding a flag for that would make InlineCodegenForDefaultBody unnecessary, though.

Intresting note!

max-kammerer:
@kandersen Could you rebase on master?
And maybe squash commits to avoid changes in first one that rollbacked in second one?

kandersen:
> @kandersen Could you rebase on master?
> And maybe squash commits to avoid changes in first one that rollbacked in second one?

Certainly!

kandersen:
Done.

max-kammerer:
Many thanks!

PR #3109, opened 2020-02-19
KT-23510 "Remove parameter" quick fix keeps lambda argument when it's out of parentheses

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-23510


vladimirdolzhenko:
LGTM, Thanks!

PR #3110, opened 2020-02-20
[FIR]: fix library methods in packages

juan-chen:
Library methods such as listOf are resolved to have the package fragments as their parents, but JVM expects their containing file classes as parents. This fix generates those file classes and replaces as parents for such library methods.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmIrCodegenFactory.kt
```diff
@@ -45,6 +46,16 @@ class JvmIrCodegenFactory(private val phaseConfig: PhaseConfig) : CodegenFactory
             irModuleFragment.descriptor, irModuleFragment.irBuiltins, symbolTable, extensions = extensions
         )
         ExternalDependenciesGenerator(symbolTable, irProviders).generateUnboundSymbolsAsDependencies()
+
+        irProviders.map {
```
mglukhikh: This is a bit unclear way to express what you want. Call to `map` usually assumes you really need list mapping, but here you just use the providers so `forEach` is better alternative. Moreover, there is exactly one `DeclarationStubGenerator` in the list, so it's better to find it first and then use. I've rewritten this piece of code in the following way:

```
        val stubGenerator = irProviders.filterIsInstance<DeclarationStubGenerator>().first()
        for (descriptor in symbolTable.functionDescriptorsWithNonClassParent()) {
            val parentClass = stubGenerator.generateOrGetFacadeClass(descriptor)
            descriptor.owner.parent = parentClass ?: throw AssertionError("Facade class for ${descriptor.name} not found")
        }
```

mglukhikh:
Merged manually, thank you! 👍

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmIrCodegenFactory.kt
juan-chen: Sounds good, thanks!

PR #3111, opened 2020-02-20
KT-36834 Convert use-site targets and usages with convert property to fun intention

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36834


vladimirdolzhenko:
LGTM

PR #3115, opened 2020-02-20
[JVM IR] Add null-checks of handler for $default calls

kandersen:
To match the old backend, `$default` dispatch methods should, as appropriate, include a check that super-class handler is not supplied in cases where this does not make sense; i.e. methods on final classes and top-level declarations - places that will never see overrides.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmDefaultArgumentStubGenerator.kt
```diff
@@ -47,4 +47,24 @@ class JvmDefaultArgumentStubGenerator(override val context: JvmBackendContext) :
...
+        newIrFunction: IrFunction,
+        handlerDeclaration: IrValueDeclaration
+    ) {
+        +irIfThen(
+            context.irBuiltIns.unitType,
+            irNot(irEqualsNull(irGet(handlerDeclaration))),
+            irThrow(irCall(this@JvmDefaultArgumentStubGenerator.context.ir.symbols.ThrowUnsupportOperationExceptionClass).apply {
+                putValueArgument(
+                    0,
+                    irString("Super calls with default arguments not supported in this target, function: ${irFunction.name.identifier}")
```
udalov: Please use `asString()` instead of `identifier`

kandersen: Thanks, don't know what happened :) The toString is equivalent, is that fine by you?

udalov: We specifically prefer to use `Name.asString()` to differentiate between "meaningful" usages of the name contents vs "debug info" usages which traditionally use `toString`.

And `Name.identifier` is undesirable since it throws exception on special names, which IMHO is not needed here (even though I can't think of an example where this could fire right now).

kandersen: Got it, thanks! Updated.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -152,6 +153,11 @@ open class DefaultArgumentStubGenerator(
         newIrFunction: IrFunction,
         params: MutableList<IrValueDeclaration>
     ): IrExpression {
+        val handlerDeclaration = newIrFunction.valueParameters.last()
```
udalov: I'd move this logic before handling of default arguments, as it's done in the old backend. Otherwise side effects of default argument intiialization can be observed before we get to assert that supercalls to methods with default arguments are not supported. This is not that important until we actually support them, but let's fix it nevertheless, otherwise we'll easily forget about this difference later.

The difference with the old backend here can be observed on the following test:
```
// TARGET_BACKEND: JVM
// WITH_RUNTIME

open class O {
    open fun foo(s: String = throw Error("Fail: this expression should not be evaluated")) {}
}

fun box(): String = try {
    val f = O::class.java.declaredMethods.single { it.name == "foo\$default" }
    f(null, O(), "s", 1, "non-null")
    "Fail: exception should have been thrown"
} catch (e: Exception) {
    val cause = e.cause
    if (cause is UnsupportedOperationException) {
        "OK"
    } else {
        cause.toString()
    }
}
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -180,6 +185,15 @@ open class DefaultArgumentStubGenerator(
         } else dispatchCall
     }
 
+    protected open fun needsSuperCallHandlerCheck(irFunction: IrSimpleFunction) = false
```
udalov: I'd combine these two methods into one `generateSuperCallHandlerCheckIfNeeded` to avoid a stub implementation below

kandersen: Done!

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
kandersen: Done. Thanks! Reshuffled and added the test.

udalov:
Thank you!

PR #3117, opened 2020-02-21
KT-12329 "invert if" inserts unnecessary 'continue' for statement inside a loop with 'continue'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-12329


vladimirdolzhenko:
Thanks a lot @t-kameyama for the PR.

Indeed it works fine with an example specified in [KT-12329](https://youtrack.jetbrains.com/issue/KT-12329). I've spotted some problem:

let's say we have following snipplet:
```
fun test() {
    for (x in "abc") {
        <caret>if (x == 'a') continue
        if (x == 'b') continue
        println("else")
    }
}
```

and sequentially apply `Invert if condition`

after step 1:
```
fun test() {
    for (x in "abc") {
        if (x != 'a') {
            <caret>if (x == 'b') continue
            println("else")
        }
    }
}
```

after step 2:
```
fun test() {
    for (x in "abc") {
        if (x != 'a') {
            if (x != 'b') {
                println("else")
            }
            continue
        }
    }
}
```

Is it possible to improve your solution ? 

Thank you in advance.

vladimirdolzhenko:
LGTM, merged manually. 

Thanks @t-kameyama

PR #3119, opened 2020-02-21
[build][native] version with milestone clause

vvlevchenko:


compiler/util-io/src/org/jetbrains/kotlin/konan/CompilerVersion.kt
```diff
@@ -54,6 +57,10 @@ data class CompilerVersionImpl(
             append('.')
             append(maintenance)
         }
+        if (milestone != 0) {
```
ilmat192: `M0` seems to be a valid version. May be use -1 to indicate that the milestone version is empty?

PR #3122, opened 2020-02-21
[FIR]: fix translation of property accessors like array.indices.

juan-chen:
This commit addresses the following issues:

* accessors didn't take into account the receiver type, which caused NoSuchMethod due to signature mismatch. Now the property's receiver type is passed to Fir2Ir translation of accessors.

* property's parent was not class, e.g., kotlin.collections.indices. Now the symbol table collects WrappedPropertyDescriptorWithContainerSource besides WrappedFunctionDescriptorWithContainerSource, so that facade classes for such properties can be generated before codegen.

* accessor's parent was not class. Now the containerSource of the property descriptor is passed to accessor descriptor.


mglukhikh:
Merged manually, thank you. I will send you some interesting details around.

PR #3123, opened 2020-02-22
Adding sample for filterNotNull.

AdamMc331:
I noticed `filterNotNull` didn't have a sample on the website, so I've added it for this ticket: https://youtrack.jetbrains.com/issue/KT-20357

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -799,4 +799,14 @@ class Collections {
         }
 
     }
+
+    class Filtering {
+        @Sample
+        fun filterNotNull() {
+            val numbers: List<Int?> = listOf(1, 2, null, 4)
+            val nonNullNumbers = numbers.filterNotNull()
+
+            assertPrints(nonNullNumbers, "[1, 2, 4]")
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -799,4 +799,17 @@ class Collections {
...
 
     }
+
+    class Filtering {
+        @Sample
+        fun filterNotNull() {
+            val numbers: List<Int?> = listOf(1, 2, null, 4)
+            val nonNullNumbers = numbers.filterNotNull()
+
+            assertEquals(
```
ilya-g: Better to use `assertPrints` and keep it in a single line. Then it will be replaced with `println(...)` in the runnable sample.

AdamMc331: Thanks!

AdamMc331:
Hey @ilya-g Thanks for the review! Anything else that needs to happen to this before it's merged? Let me know if I need to rebase or anything similar.

ilya-g:
Thanks for the sample, I've merged it: 840f2228670ab37f9f3e000899f5a023bba61730

ilya-g:
Live on kotlinlang.org: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-not-null.html

PR #3124, opened 2020-02-22
Adding sample for filter

brianPlummer:
Hey there! I noticed Collections filter didn't have a sample on the website, so I've added it for this ticket: https://youtrack.jetbrains.com/issue/KT-20357

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -764,6 +764,16 @@ class Collections {
         }
     }
 
+    class Filtering {
+        @Sample
+        fun filter() {
+            val numbers: List<Int> = listOf(1, 2, 3, 4)
+            val evenNumbers = numbers.filter { number -> number.rem(2) == 0 }
+
+            assertPrints(evenNumbers, "[2, 4]")
+        }
+    }
+    
     class Sorting {
 
         @Sample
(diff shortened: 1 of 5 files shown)


libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -764,6 +764,16 @@ class Collections {
         }
     }
 
+    class Filtering {
+        @Sample
+        fun filter() {
+            val numbers: List<Int> = listOf(1, 2, 3, 4)
+            val evenNumbers = numbers.filter { number -> number.rem(2) == 0 }
```
ilya-g: Two notes:
- better to use `%` infix operator instead of its named form `rem`
- short lambdas usually do not introduce name for its parameter and use default `it` instead.

It may be beneficial, however, to show both variants, so we can do the following: 
- use `it` in `filter` lambda
- write another expression with `filterNot` that uses explicit parameter name
- now this sample can be used for `filterNot` as well :)

libraries/tools/kotlin-stdlib-gen/src/templates/Filtering.kt
```diff
@@ -548,6 +548,7 @@ object Filtering : TemplateGroupBase() {
         specialFor(ArraysOfUnsigned) { inlineOnly() }
 
         doc { "Returns a ${f.mapResult} containing only ${f.element.pluralize()} matching the given [predicate]." }
+        sample("samples.collections.Collections.Filtering.filter")
```
ilya-g: Have you run stdlib code generation after that? It should affect array `filter` extension too.

ilya-g:
Ok, I've made the required edits and then merged it manually: https://github.com/JetBrains/kotlin/compare/dd2aa3de2847cf4bd65dd9e46329ff32820b1a70~2...dd2aa3de2847cf4bd65dd9e46329ff32820b1a70
Thanks for the effort.

PR #3125, opened 2020-02-23
KT-36686 Implement members quickfix puts the implementation *before* the data class if it already has a body

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36686


vladimirdolzhenko:
LGTM.

It's very nice.

PR #3126, opened 2020-02-23
[build][native] version with milestone clause

vvlevchenko:
(cherry picked from commit 98ce49ba73dd972fc7a98f32642da4178698d095)


compiler/util-io/src/org/jetbrains/kotlin/konan/CompilerVersion.kt
```diff
@@ -43,6 +45,7 @@ data class CompilerVersionImpl(
     override val major: Int,
     override val minor: Int,
     override val maintenance: Int,
+    override val milestone: Int = 0,
```
udalov: `0` here looks strange, given that below we check for `-1`

udalov: ^ @vvlevchenko

PR #3130, opened 2020-02-25
KT-14021 Quickfix to add parameter to function gives strange name to parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-14021

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt
@@ -36,10 +36,7 @@ import org.jetbrains.kotlin.idea.refactoring.changeSignature.KotlinChangeSignatu
 import org.jetbrains.kotlin.idea.refactoring.changeSignature.KotlinMethodDescriptor
 import org.jetbrains.kotlin.idea.refactoring.changeSignature.modify
 import org.jetbrains.kotlin.idea.refactoring.changeSignature.runChangeSignature
-import org.jetbrains.kotlin.psi.KtCallElement
-import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.psi.KtNameReferenceExpression
-import org.jetbrains.kotlin.psi.ValueArgument
+import org.jetbrains.kotlin.psi.*
 import org.jetbrains.kotlin.psi.psiUtil.getStrictParentOfType
 import org.jetbrains.kotlin.resolve.calls.callUtil.getCall
 import org.jetbrains.kotlin.resolve.calls.callUtil.getType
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt
```diff
@@ -59,11 +60,12 @@ abstract class ChangeFunctionSignatureFix(
     }
 
     protected fun getNewArgumentName(argument: ValueArgument, validator: Function1<String, Boolean>): String {
-        val argumentName = argument.getArgumentName()
+        val argumentName = argument.getArgumentName()?.asName?.asString()
+            ?: (argument.getArgumentExpression() as? KtNameReferenceExpression)?.getReferencedName()?.takeIf { it != "it" }
```
yanex: I suppose, `field` is the another special case:

```
fun foo() {}

class Test {
    val x: String = ""
        get() {
            foo(field)
            return field
        }
}
```

Maybe we can even take the property name instead.

yanex: Maybe also deparenthesize the name expression?

yanex:
Review 👍 .

yanex:
By the way, I extracted the name comparison so the special names are in one place:

```
Index: idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt	(revision 3ce148df1934767af3c906a42ac840442256a1c4)
+++ idea/src/org/jetbrains/kotlin/idea/quickfix/ChangeFunctionSignatureFix.kt	(revision 20b9b74107602c93459e063d142335542e7e8c10)
@@ -59,14 +59,14 @@
     protected fun getNewArgumentName(argument: ValueArgument, validator: Function1<String, Boolean>): String {
         val expression = KtPsiUtil.deparenthesize(argument.getArgumentExpression())
         val argumentName = argument.getArgumentName()?.asName?.asString()
-            ?: (expression as? KtNameReferenceExpression)?.getReferencedName()?.takeIf { it != "it" && it != "field" }
+            ?: (expression as? KtNameReferenceExpression)?.getReferencedName()?.takeIf { !isSpecialName(it) }
 
         return when {
             argumentName != null -> KotlinNameSuggester.suggestNameByName(argumentName, validator)
             expression != null -> {
                 val bindingContext = expression.analyze(BodyResolveMode.PARTIAL)
                 val expressionText = expression.text
-                if (expressionText == "it" || expressionText == "field") {
+                if (isSpecialName(expressionText)) {
                     val type = expression.getType(bindingContext)
                     if (type != null) {
                         return KotlinNameSuggester.suggestNamesByType(type, validator, "param").first()
@@ -78,6 +78,10 @@
         }
     }
 
+    private fun isSpecialName(name: String): Boolean {
+        return name == "it" || name == "field"
+    }
+
     companion object : KotlinSingleIntentionActionFactoryWithDelegate<KtCallElement, CallableDescriptor>() {
         override fun getElementOfInterest(diagnostic: Diagnostic): KtCallElement? = diagnostic.psiElement.getStrictParentOfType()
 ```

t-kameyama:
Added test for https://youtrack.jetbrains.com/issue/KT-14021

t-kameyama:
19d1532dc768c01d66b956b5699d2de4d5a27e73

PR #3131, opened 2020-02-25
Reduce UI freezes by preparing affected Gradle project files lists in advance

nskvortsov:
isRelevant() method is a hotspot, as it is called for each file in bulk VFS event.
As there can be hundreds of files in each event, we should keep operations minimal.


nskvortsov:
@snrostov please take a look

nskvortsov:
update missing import for *.201 file

PR #3132, opened 2020-02-26
KT-17121 "Implement members" missed

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-17121


yanex:
The fix should behave like the normal "Implement members".
If there're several methods/properties to override, it should open the "Override members" dialogue, from which user can choose multiple entries at once. Now it suggests overriding methods one by one.

yanex:
Review 👍 .
Just in case: `menyImplMemberNotImplemented.kt.after` looked a bit strange at first, however, it's how our "implement members" tests work if all items are selected.

t-kameyama:
8b8059acdd2ec323c0e053ee395e9d76e929593b

PR #3134, opened 2020-02-26
KT-15262: Allow generate toString() to include properties with non-default getters

n-p-s:
This resolves [KT-15262](https://youtrack.jetbrains.com/issue/KT-15262), but also has the side effect of allowing properties with non-default getter (i.e delegated properties) to be included in the wizard when generating `toString()`, `equals()` and `hashCode()`. Happy to refine this further if this isn't ideal.

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/actions/generate/utils.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/actions/generate/utils.kt
@@ -40,8 +40,7 @@ fun getPropertiesToUseInGeneratedMember(classOrObject: KtClassOrObject): List<Kt
         classOrObject.declarations.asSequence().filterIsInstance<KtProperty>().filterTo(this) {
             val descriptor = it.unsafeResolveToDescriptor()
             when (descriptor) {
-                is ValueParameterDescriptor -> true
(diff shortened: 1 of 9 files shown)


idea/src/org/jetbrains/kotlin/idea/actions/generate/utils.kt
```diff
@@ -40,8 +40,7 @@ fun getPropertiesToUseInGeneratedMember(classOrObject: KtClassOrObject): List<Kt
         classOrObject.declarations.asSequence().filterIsInstance<KtProperty>().filterTo(this) {
             val descriptor = it.unsafeResolveToDescriptor()
             when (descriptor) {
-                is ValueParameterDescriptor -> true
-                is PropertyDescriptor -> descriptor.getter?.isDefault ?: true
+                is ValueParameterDescriptor, is PropertyDescriptor -> true
```
nikitabobko, line 43: This done intentionally. User custom getter may contain complex computations or side effects. It's not safe to suggest custom getters in toString/hashCode default generator.

You can suggest pull request where custom getters are suggested but are disabled by default in this dialog:
image

PR #3136, opened 2020-02-26
maven-plugin: config for force execution even without kotlin sources

alexey-anufriev:
In case if Kotlin needs to be added to the existing Java/Maven project there is a need to reconfigure `maven-compiler-plugin` to set the proper order of executions. In the end `kotlin-maven-plugin` will be executed first. This can be not an ideal case for projects that are using annotation processing quite heavily. For example project may contain multiple modules and `maven-compiler-plugin` is configured in the root POM to unify the setup once. But in case if configuration contained annotation processing it is recommended to let this be done by `kapt`. But `kapt` does not perform annotation processing for modules where there is only java source code. 

Detailed example:
```
- root-module
-- java-only-module
-- ... other modules
-- java-kotlin-module
```

> `root-module` contains `maven-compiler-plugin` configuration with annotation processors

> other modules are using this single config for build

After adding Kotlin to `java-kotlin-module` there is a need to reconfigure `maven-compiler-plugin`. But after this reconfiguration annotation processing will stop working in modules other than this one.

This is a limitation that can be solved by current PR. The idea is to allow force execution of `kapt` via a dedicated parameter - `forceExecutionWithoutKotlinSources`. The default is still `false` to be compatible.

### DIFF
--- a/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
+++ b/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
  */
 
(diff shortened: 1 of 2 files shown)


alexey-anufriev:
@cy6erGn0m, any information on when this can be merged?

cy6erGn0m:
1.4 may be, need to discuss

alexey-anufriev:
It would be awesome!
please let me know if there are any remarks so we do not delay it for longer

alexey-anufriev:
@cy6erGn0m, kotlin 1.4-M1 has been released recently. What about current PR, can we also proceed?

alexey-anufriev:
@abelkov, @cy6erGn0m, hello guys, do you have any information if this PR has any chance to get into 1.4? if not then what is missing and how can I help?

alexey-anufriev:
hello @cy6erGn0m, any updates here?

abelkov:
cc @CherepanovAleksei

cy6erGn0m:
Sorry, I don't work on maven plugin any longer

alexey-anufriev:
> Sorry, I don't work on maven plugin any longer

I see, no problem, but I would update the assignment.
And who is the new contact person for this issue?

PR #3137, opened 2020-02-27
KT-36508 False positive "Replace 'to' with infix form" when 'to' lambda generic type argument is specified explicitly

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36508


t-kameyama:
8d3e41ae602d87a1f653bc979bae6f1bfb59c1c6

PR #3138, opened 2020-02-27
KT-5271 Missing QuickFix for Multiple supertypes available

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-5271

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/quickfix/QuickFixRegistrar.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/quickfix/QuickFixRegistrar.kt
@@ -648,5 +648,7 @@ class QuickFixRegistrar : QuickFixContributor {
         TOPLEVEL_TYPEALIASES_ONLY.registerFactory(MoveTypeAliasToTopLevelFix)
 
         CONFLICTING_JVM_DECLARATIONS.registerFactory(AddJvmNameAnnotationFix)
+
+        AMBIGUOUS_SUPER.registerFactory(SpecifySuperTypeFix)
     }
 }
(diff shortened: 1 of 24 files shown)


yanex:
Case 1: `X` is erroneously suggested.

```
interface X {
    fun foo() {}
}

open class Y: X {
    override fun foo() {}
}

interface Z {
    fun foo() {}
}

class Test : X, Y(), Z {
    override fun foo() {
        <caret>super.foo()
    }
}
```

yanex:
Case 2: Type arguments should not be present in super<>.

```
interface X {}

open class Y<T> {
    open fun foo() {}
}

interface Z {
    fun foo() {}
}

class Test : X, Y<Int>(), Z {
    override fun foo() {
        <caret>super.foo()
    }
}
```

yanex:
Case 3: `Z` is not a valid option.

```
interface Z {
    fun foo(): CharSequence = ""
}

open class Y: Z {
    override fun foo(): String = ""
}

class Test : Y(), Z {
    override fun foo(): String {
        return super.foo()
    }
}
```

yanex:
Case 4: Labels are not treated properly.

```
class Test : Y(), Z {
    override fun foo() {}

    inner class Boo : Y(), Z {
        override fun foo() {
            super@Test.foo()
        }
    }
}
```

Should replace with `super<Y>@Test` instead of `super<Y>`.

yanex:
Case 5, similar to 2: annotations should be removed.

```
interface Z {
    fun foo() {}
}

open class X {
    open fun foo() {}
}

class Test<T : CharSequence> : (@Suppress("foo") X)(), Z {
    override fun foo() {
        super.foo()
    }
}
```

yanex:
Review 👍

t-kameyama:
f6a13df388540d71ccdbc99f69f6d60dd4648810

PR #3139, opened 2020-02-27
KT-36461 "Create enum constant" quick fix adds after semicolon, if the last entry has a comma

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-36461
https://youtrack.jetbrains.com/issue/KT-36462


yanex:
`IS_ADDED_TRAILING_COMMA` flag is never removed from the user data.
Also, I think the visitor can be rewritten without using user data at all.

yanex:
Review 👍

t-kameyama:
4973d3c359ecadb2a3cfe26de135c8a150413050

PR #3140, opened 2020-02-27
JVM_IR: cache signatures in BridgeLowering

pyos:
`mapAsmMethod` is somewhat slow, so this helps if there are a lot of overrides of same methods and/or deep class hierarchies in a single file (though it would probably help more if lowerings were not reconstructed for every file).


gogabr:
Thanks!

PR #3141, opened 2020-02-27
JVM, JVM IR: Sort declarations in bytecode listing tests

sfs:
The order differs between the JVM and JVM IR backends, which makes it difficult to compare the two.


dnpetrov:

dnpetrov:
There are some more bytecode listing tests in project, I'll update corresponding testData in a separate commit.

dnpetrov:
In fact, the idea of sorting declarations in testData by signature + annotations looks somewhat questionable, because nullability annotations also differ in JVM_IR. So, it might make comparing those declarations simple in one case (different set of declarations), but difficult in other cases (different nullability annotations on declarations).

sfs:
Do we want to ignore nullability annotations? Looking at the tests, I think that right now the JVM IR backend generates too many (there's no point in putting annotations on synthetic methods) while the JVM backend generates too few (there's several non-synthetic methods, e.g., special bridges, which don't carry annotations in the JVM backend).

Maybe we should just sort by flags + signature and ignore annotations for sorting and then go through the test data by hand to figure out which behavior we want on a case by case basis before fixing it in both the old and new BE. What do you think?

dnpetrov:
I agree that JVM_IR generates "too many" nullability annotations, and old back-end generates probably "too few", but we have to identify some proper middle ground. Or probably just ignore them on declarations that are not part of "public API" (but might be part of public ABI).

I actually think about rendering generated declarations to some more structured format for ease of comparison (because yes, differences in those one-liners are hard to read). Right now I think that they should be sorted by JVM descriptors (so that equivalent declarations are matched against each other in ideal case).

sfs:
I've changed `AbstractBytecodeListingTest` to sort by method/field descriptors and started fixing some of the errors in the tests.

There is a mix of problems in the JVM and the JVM IR backends. For now, I've focused on the few tests that look at generic signatures. The problems were as follows:
- The JVM backend didn't generate generic signatures for `$$delegate` fields.
- The JVM backend created incorrect generic signatures for special bridge methods (based on the *target* of the bridge, rather than the overridden method).
- The JVM IR backend created non-instantiated generic signatures for special bridge methods.
- The JVM IR backend created unnecessary annotations for bridge methods.

Fortunately, it turns out that we don't need generic signatures for bridges. That's a good thing, since we would have to go through the type system to determine the correct generic signatures to use based on the overriding class, which would be relatively expensive.

sfs:
I've split the PR so that this one only contains the changes to `AbstractBytecodeListingTest` and the remaining changes are in #3162 and #3163. Together these tests fix `testGenericClass`, `testNonGenericClass`, and `testPartiallySpecializedClass` in the IR bytecode listing tests.

sfs:

dnpetrov:
@sfs Hi Steven, I though we agreed that sorting declarations textually in bytecode listing tests is a questionable idea because of differences in generated nullability annotations (for some tests it might be ok, but in some cases it makes harder to compare test results against golden data).

If you still need it, please, let me know. Otherwise, I'm going to close this PR.

sfs:
@dnpetrov That's right, I changed the code to sort by JVM method signatures instead. I still think these changes are still needed in order to compare the JVM and JVM IR backends. Having completely separate test expectations seems like the wrong approach.

PR #3142, opened 2020-02-27
JVM_IR: don't remap current method signature when detecting clashes

pyos:
TODO: don't repeatedly map signatures of inherited functions either.


dnpetrov:

dnpetrov:
Pushed manually.

035cc57cf47054383a44e668612927dd793f9d22

PR #3143, opened 2020-02-27
JVM IR: Box arguments in abstract method stubs for "remove" bridges

sfs:
The problem is that in code such as:
```kotlin
abstract class C : MutableCollection<Int>
```
we need to generate an abstract method stub with signature `remove(Ljava/lang/Integer)Z` instead of `remove(I)Z`, in order to avoid clashes with a different method with the same signature in `java.util.List` which could be implemented by a subclass.

I didn't add a test, since this is already tested for in some bytecode listing tests, which are still failing because of differences in nullability annotations and `toArray` methods.


gogabr:
Thank you!

PR #3144, opened 2020-02-28
JVM_IR: preserve old backend's logic for special mutable collection c…

neetopia:
…lass stub methods, this also fix an issue with redundant stub generation for Kotlin class extending Java class of subtype of a collection class.


neetopia:
cc @sfs

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -136,12 +136,20 @@ private class CollectionStubMethodLowering(val context: JvmBackendContext) : Cla
...
+
     private inner class StubsForCollectionClass(
         val readOnlyClass: IrClassSymbol,
         val mutableClass: IrClassSymbol
     ) {
         val mutableOnlyMethods: Collection<IrSimpleFunction> by lazy {
-            val readOnlyMethodSignatures = readOnlyClass.functions.map { it.owner.toSignature() }.toHashSet()
+            val readOnlyMethodSignatures = readOnlyClass
+                .functions
+                .filter { it.owner.name.asString() !in specialNames }
```
udalov: Please use the full JVM signature (after the subsequent `.map { it.owner.toSignature() }` call) to filter out these members, instead of only name

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -172,13 +180,22 @@ private class CollectionStubMethodLowering(val context: JvmBackendContext) : Cla
 
     // Compute stubs that should be generated, compare based on signature
     private fun generateRelevantStubMethods(irClass: IrClass): Set<IrSimpleFunction> {
+
+        // For superTypes from java, we only generate stub if all further superTypes which are subtype of the immutable class are interfaces,
```
udalov: Could you please move this logic into `comesFromJava`, rename that method accordingly and make it private, and also update the comment on lines 195-196?

neetopia: Done, didn't notice I forgot to change comment.

udalov:
Related: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3931

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/CollectionStubMethodLowering.kt
```diff
@@ -136,12 +136,24 @@ private class CollectionStubMethodLowering(val context: JvmBackendContext) : Cla
         return mutableClass.typeParameters.map { it.symbol }.zip(readOnlyClassTypeArguments).toMap()
     }
 
+    // Preserve old backend's logic to generate stubs for special cases where a mutable method
+    // have same jvm signature with an immutable method.
+    private val specialSignatures = setOf(
```
udalov: Could you please name this variable something more specific and reflecting its purpose?

For example, `specialCaseStubSignaturesForOldBackend`.

And also please mention the issue [KT-36724](https://youtrack.jetbrains.com/issue/KT-36724) in the comment. We'd like to fix this issue in the old backend at some point.

udalov:
Regarding the change to avoid generating redundant methods in collections, I think @gogabr's solution in [KOTLIN-CR-3931](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3931) is a bit more robust. So let's wait until that one is merged.

The fix of the `subList`/`listIterator` problem looks good to me, will merge after that 👍.

neetopia:
Updated

udalov:
Sorry for the delay. I was trying to figure out if we can avoid hardcoding the signatures by reusing the (buggy) predicate from the old backend, but did not succeed.

Pushed manually in 03613d47087 with minor changes (grammar in the comment). Thank you!

PR #3145, opened 2020-02-28
JVM_IR: refactor ToArrayLowering and make matching more precise

pyos:
* If `toArray` is inherited from Java, it may take an argument or return a value of a flexible type, which looks nullable in IR;
  * The returned array may also have `out` variance of the type argument;
  * `Array<T>` is not the same as `Array<Any?>` if `T` is neither `Any` nor `in Something`, so presence of `toArray(): Array<T>` does not mean we don't need to generate a new `toArray`.


udalov:
Thanks! Merged in 20ea77d55d7.

PR #3146, opened 2020-02-28
KT-32511 Create class quick fix is not suggested in super type list in case of missing primary constructor

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-32511


yanex:
I think the logic should be quite different. Now the quick fix isn't shown if the class has a primary constructor, and it's not clear for me why.
The "Create class" option should be available for all classes (excluding `annotation`, `enum`, `inline`).

yanex:
Review 👍

t-kameyama:
21d6d5024137bfabcce9671d69d7918a503de103

PR #3147, opened 2020-03-01
KT-22014 Should "convert lambda to reference" intention be available in the following case

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-22014
https://youtrack.jetbrains.com/issue/KT-22015
https://youtrack.jetbrains.com/issue/KT-34378


t-kameyama:
9b6b25f069b13abbe4fd270bbff529dcc1b4212e
f7a69d4aa7e3a4a47f4e6280dd2afed3cfbd4949

PR #3148, opened 2020-03-02
JVM_IR: default-initialize variables visible in the LVT

pyos:
See KT-36812. Aside from the problem stated there, D8 will throw out the entire LVT if it sees a variable that has not been written to (and will generate incorrect SSA if the slot is reused with a different type).

Note: this only fixes a FIR test because it's missing an `else -> throw` branch, and default initialization satisfies the verifier and masks the incorrect control flow.


dnpetrov:
Looks good to me.

NB I had plans to "normalize" LVT entries based on reaching definitions, but that can be postponed.

dnpetrov:
Pushed manually.

7e6d0801235a76f3a74836ee1ac6432d1f4a0e3a

PR #3149, opened 2020-03-02
Optimize more redundant `kotlin/jvm/internal/Ref`s

pyos:
The number of initializations of the `value` field before the live range begins does not really matter so long as we insert a write of a default value to the local if there were none.


dnpetrov:
Looks good to me.

dnpetrov:
Merged and pushed manually.

ed83e3ccef2fcb114543da1c52f88f09ec590988

PR #3151, opened 2020-03-02
KT-37148 "Remove redundant `.let` call doesn't remove extra calls

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37148


t-kameyama:
b8ae13b3d92aa102d4a0cb38bdf5f01833fb4033

PR #3154, opened 2020-03-03
[FIR] diagnostics tests

anastasiiaSpaseeva:


compiler/tests-spec/tests/org/jetbrains/kotlin/spec/checkers/DiagnosticsTestSpecGenerated.java
```diff
@@ -3317,6 +3317,71 @@ public void testAllFilesPresentInOverriding() throws Exception {
             }
         }
 
+        @TestMetadata("compiler/tests-spec/testData/diagnostics/linked/overload-resolution")
```
mglukhikh: I don't quite understand why this file is changed here

anastasiiaSpaseeva: My fault. Reverted

mglukhikh:
Merged 👍

PR #3156, opened 2020-03-03
KT-37173 "Replace with string templates" intention for String.format produces uncompilable string template

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37173


t-kameyama:
bf3e739edffc24659f699dbaca9d1d7ccb9491f6

PR #3157, opened 2020-03-03
JVM: missed line number for return expression.

jsjeon:


compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1650,7 +1650,7 @@ public StackValue visitReturnExpression(@NotNull KtReturnExpression expression,
 
             Label afterReturnLabel = new Label();
             generateFinallyBlocksIfNeeded(returnType, returnKotlinType, afterReturnLabel);
-
+            markLineNumber(expression, false);
```
jsjeon: FYI, counterpart in JVM_IR is here: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt#L667

compiler/testData/debug/stepping/smap.kt
```diff
@@ -0,0 +1,63 @@
+
```
max-kammerer: Name of test a bit confusing. Could you rename it?

max-kammerer:
@yanex Please also take a look

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt
```diff
@@ -310,6 +311,22 @@ internal fun firstLabelInChain(node: LabelNode): LabelNode {
     return curNode
 }
 
+internal val MethodVisitor.bytecode: String?
```
max-kammerer: There is already smae function in FunctionCodegen.renderByteCodeIfAvailable

jsjeon: Exactly. During debugging (line number table difference between old and new backend), I somehow thought `FunctionCodegetn.renderByteCodeIfAvailable(anyMethodVisitir)` is not resolvable from new backend, but just tried it again and it works fine. Not sure what I saw, hm. Removed.

compiler/testData/debug/stepping/smap.kt
jsjeon: Sure, renamed. Also left a comment to indicate this is from `.../testData/codegen/boxInline/smap/smap.kt`

compiler/testData/debug/stepping/nestedInline.kt
```diff
@@ -0,0 +1,63 @@
+// This is same as kotlin/compiler/testData/codegen/boxInline/smap/smap.kt
```
yanex: I think it would be much easier to understand the LINENUMBERS data if this comment will be below the files.

yanex: I didn't get why we got these line numbers:
```
// TestKt.box():10
// TestKt.box():14
// TestKt.box():10 
```

The 14th line is empty. I initially thought the comment line before the `// FILE` breaks the numbering, and I have to decrement 1 from all line numbers. But then comes:

```
// TestKt.box():7 // becomes 6, sort of understandable
// TestKt.box():9 // becomes 8, an empty line
```

How to read these tests? Or, is it intentional to put the line number to the empty string?

jsjeon: I was also confused: initially thought either 1) both old and new backends are wrong to pick up proper `file.fun` or 2) the test setting is somewhat incorrect such that it made both backends pick up that wrong `file.fun` info because those lines you were confused are actually in `1.kt`, not `test.kt`. But, then, realized that even within `test.kt`, everything is at `box` and... there are several lines that seem not belong to `box` here and there. Maybe it's just how `inilne` works.

Anyhow, my understanding looks like this:
```
// TestKt.box(): 19
// TestKt.test(): 7   // but inlined to `box` so somewhat make sense
// TestKt.test(): 9   // calling `html` in 1.kt
// 1Kt.html(): 18     // Note that 1.kt starts at 24, so line 41 (inside html where it calls init) becomes 18
// 1Kt.init(): 6      // Again, from 24, line 29 (the body of init) becomes 6
// TestKt.test(): 10  // the parameter `init` of fun `init` in 1.kt is `head` in `test`
// 1Kt.head(): 14     // From 24, the body of `head` at line 37 becomes 14
// 1Kt.initTag2(): 10 // From 24, 1st line of `initTag2` at line 33 becomes 10
// 1Kt.initTag2(): 11 // next line
// TestKt.test(): 11  // param `init` of `initTag2` came from `head`, which is a lambda in `test` containing `res = "Ok"` at line 11
// TestKt.test(): 12  // end of the lambda `init` passed to `head` and then `initTag2`
// 1Kt.initTag2(): 12 // end of `initTag2` at line 35 becomes 12, starting from 24
// 1Kt.head(): 14     // From 24, the end of `head` at the same line 37 becomes 14
// TestKt.test(): 13  // the end of lambda `init` passed to `html` and then `init` in 1.kt
// 1Kt.init(): 7      // the end of `init` at line 30 becomes 7, starting from 24
// 1Kt.html(): 18     // the end of `html` at line 41 becomes 18, starting from 24 // this line is missing in old backend w/o this PR
// TestKt.test(): 15  // the end of `test`
// TestKt.box(): 19   // finished inlining `test()`
// TestKt.box(): 21   // the end of `box`
```

max-kammerer: Look like that test framework doesn't support source mapping information that cause wrong class names. cc @juan-chen

jsjeon: I guess you meant to cc @neetopia (Jiaxiang Chen) who initiated these stepping tests.  :) I'll directly ask him where to look.

jsjeon: Actually, `event` coming from JDI is mapped back to original lines according to SMAP. Here is snippet of bytecode of `box` (not this file, but `smap.kt` under `boxInline`:

```
  // access flags 0x19
  public final static box()Ljava/lang/String;
  @Lorg/jetbrains/annotations/NotNull;() // invisible
   L0
    LINENUMBER 20 L0
   L1
    ICONST_0
    ISTORE 1
   L2
    LINENUMBER 40 L2
    LDC "Fail"
    ASTORE 2
   L3
    LINENUMBER 42 L3
   L4
    ICONST_0
    ISTORE 3
   L5
    LINENUMBER 43 L5
    ICONST_0
    ISTORE 4
  ...
```

and the SMAP for that method is: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/boxInline/smap/smap.kt#L104
```
20#1,3:40
20#1:45
20#1,2:49
20#1:52
20#1:54
9#1:26
9#1,9:27
9#1,3:36
9#1:39
...
```

I switched positions of `1.kt` and `2.kt`, so lines are not exactly 20 or 9, but the point is we're not seeing non-sense line numbers after conducting nested inlines; line numbers are aligned with source and make sense (to me at least).

------

At this line:
https://github.com/JetBrains/kotlin/blob/master/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/debugInformation/AbstractSteppingTest.kt#L59 more info we can retrieve from `Location` are source name. With some tweak, I get:
```
test.kt:TestKt.box():19
test.kt:TestKt.box():7
test.kt:TestKt.box():9
1.kt:TestKt.box():18
1.kt:TestKt.box():6
test.kt:TestKt.box():10
1.kt:TestKt.box():14
1.kt:TestKt.box():10
1.kt:TestKt.box():11
test.kt:TestKt.box():11
test.kt:TestKt.box():12
1.kt:TestKt.box():12
1.kt:TestKt.box():14
test.kt:TestKt.box():13
1.kt:TestKt.box():7
1.kt:TestKt.box():18
test.kt:TestKt.box():15
test.kt:TestKt.box():19
test.kt:TestKt.box():21
```
where the format is `source-name:method:line` Since everything is inlined to `box`, no further info can be presented. Let me know how this looks like, then I can update all stepping tests accordingly in a follow-up commit.

jsjeon: Update all stepping tests at the follow-up commit: ac37bf50752218955dad62bba5f6f7b45f47ab96. It would look like redundant info for many cases where only single file, `test.kt`, is used, but for some cases, including `nestedInline.kt` added in this PR, it's helpful to understand where that specific line belongs.

max-kammerer:
@jsjeon Thank you, it much clearer. I wonder do we need method info at all (was original intention just check linenumbers or not @neetopia )? Confusing linenumbers could be observed even in simple case with 2 classes.
If keeping method info it's best to place source and linenumber info together

jsjeon:
> I wonder do we need method info at all (was original intention just check linenumbers or not @neetopia )?

Actually his first suggestion was to use source name instead of method info. Let me change it a bit so that only source name and linenumber are tested.

max-kammerer:
Thank you!

PR #3161, opened 2020-03-04
JVM IR: Respect toArray implementations coming from Java

sfs:
The JVM IR backend currently ignores existing `toArray` implementations containing platform types, which can lead to potentially wrong behavior (see `testToArrayJava`).


udalov:
Isn't this fixed in #3145?

(Sorry for the lag in reviews, will look at that one soon.)

sfs:
You're right, that's duplicated work. Since @pyos also refactored `ToArrayLowering` I would much prefer that his version be merged. Let me close this one.

PR #3162, opened 2020-03-04
JVM: Generate generic signatures for delegate fields

sfs:
The JVM IR backend already generates generic signatures on `$$delegate_n` fields and this PR changes the JVM backend to do the same.

There was a TODO comment from 2014 for this, so this is probably not very pressing. :)


udalov:
Thank you!

PR #3163, opened 2020-03-04
JVM, JVM IR: Do not generate annotations or generic signatures on bridge methods

sfs:
This seems to be more in line with what javac is doing, as javac does not produce generic signatures for bridge methods. Plus the current behavior of the JVM backend is to take the generic signature from the *bridge target* instead of the overridden method, which is clearly wrong.

For example, in
```kotlin
interface MS<T> : MutableSet<T>

abstract class AbstractStringStringMap : MutableMap<String, String> by HashMap<String, String>() {
    abstract override val keys: MS<String>
}
```
we produce a method
```java
public bridge final <()LMS<Ljava/lang/String;>;> method keySet(): java.util.Set
```
whose generic signature does not even match its descriptor.


udalov:
Please rebase on master, and also please investigate a failing IDE test `JavaAgainstKotlinBinariesCheckerTestGenerated.testUsingMutableInterfaces`:

```
com.intellij.rt.execution.junit.FileComparisonFailure: UsingMutableInterfaces.java: extra (12:13/30): 'E e1 = cMutableList.remove(3);' (Incompatible types. Found: 'java.lang.Object', required: 'E') [HighlightInfoTypeImpl[severity=ERROR, key=ERRORS_ATTRIBUTES]]
 expected:<...(elem);
            [E e1 = cMutableList.remove(3);]
            E e2 = ...> but was:<...(elem);
            [<error descr="Incompatible types. Found: 'java.lang.Object', required: 'E'">E e1 = cMutableList.remove(3);</error>]
            E e2 = ...>
```

sfs:
Rebased on master.

The test `JavaAgainstKotlinBinariesCheckerTestGenerated.testUsingMutableInterfaces` fails because of special bridges, which are non-synthetic and thus can be called as normal methods from Java. I honestly should have thought of that.

It's still fine to remove the generic signatures from normal bridges, but for special bridges we need to compute the correct generic signatures in both the JVM and JVM IR backends. I'll look into this some more.

sfs:
I'm closing this PR in favor of #3537 

This does not yet include the fixes to the generic signatures for `containsAll`, `removeAll`, and `retainAll`. I'll add that part again in a separate PR.

PR #3164, opened 2020-03-04
JVM_IR: reuse IR for $$forInline versions of suspend funs

pyos:
This should improve performance a bit, especially with nested hierarchies of anonymous objects/lambdas in suspend methods. Also, this fixes non-local returns from inline suspend functions, as currently they are remapped to `$$forInline` versions so the bytecode of the original is incorrect (this is only observable from Java or by tools like Proguard which complain about the nonexistent class `$$$$$NON_LOCAL_RETURN$$$$$`); the only alternative here is to worsen the performance even more by copying lambdas as well.


pyos:
@ilmirus since various refactorings of ClassCodegen were merged, I was able to remove the dependency of this PR on the code that caches method nodes. Could you take a look, please?

pyos:
@ilmirus ping?

pyos:
Another observable effect of copying method bodies:
```
inline suspend fun f(): String {
    val x = object {}
    return x::class.java.enclosingMethod
}

suspend fun main() {
    println(f())
}
```
Non-IR: `public static final java.lang.Object AKt.f(kotlin.coroutines.Continuation)`
IR: `private static final java.lang.Object AKt.f$$forInline(kotlin.coroutines.Continuation)`

PR #3166, opened 2020-03-05
[JVM IR] Ensure there is one accessor for each `super` access from a subclass when there are multiple subclasses in a file

punzki:
This was an interesting edge case where only one accessor was being generated for a target function within the same file, but it is possible to need multiple accessors for the same target function. See test data for examples.


udalov:
Looks good, thanks!

I'm merely curious, did you find this failure in some real code?

punzki:
@udalov Yes, in this case the base class and both subclasses were all in the same file. Kind of makes sense that code could be organized this way.

PR #3169, opened 2020-03-06
[JVM IR] Use reference equality when comparing enums

punzki:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/intrinsics/Equals.kt
```diff
@@ -80,10 +81,14 @@ class Equals(val operator: IElementType) : IntrinsicMethod() {
         val leftType = with(codegen) { a.asmType }
         val rightType = with(codegen) { b.asmType }
         val opToken = expression.origin
+        val aIsEnum = a.type.classOrNull?.owner?.isEnumClass == true
+        val bIsEnum = b.type.classOrNull?.owner?.isEnumClass != true
```
udalov: Not sure if this works, but I guess it should be `== true` here?

udalov:
Thanks! Merged manually in 368b0d9b0bc with a FIR test ignore

PR #3171, opened 2020-03-06
JVM IR: Generate fewer CHECKCAST instructions

sfs:
Plus some small codegen fixes. Generating redundant CHECKCAST instructions can end up confusing tools which look for specific bytecode patterns. That's what the `testNoCheckcastInClinit` is about.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
@@ -12,6 +12,7 @@ import org.jetbrains.kotlin.backend.jvm.lower.inlineclasses.InlineClassAbi
 import org.jetbrains.kotlin.codegen.AsmUtil
 import org.jetbrains.kotlin.codegen.StackValue
 import org.jetbrains.kotlin.ir.types.*
+import org.jetbrains.kotlin.ir.util.isSubclassOf
 import org.jetbrains.kotlin.ir.util.parentAsClass
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.org.objectweb.asm.Label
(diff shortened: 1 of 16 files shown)


sfs:
We just found a bug with this PR, where we don't generate enough CHECKCAST instructions for `when` expressions. Please hold off on merging this while I update it with a fix.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -53,7 +54,14 @@ abstract class PromisedValue(val codegen: ExpressionCodegen, val type: Type, val
             }
         }
 
-        if (type != target) {
+        if (type == target) return
+
+        if (
```
max-kammerer: This piece of code looks very overcomplicated

sfs: Yeah, it does, but it's the opposite. There are still corner cases where this is not enough.

I'll close this PR now, so let me just record why this was complicated:
- forceCast is necessary to ensure that we have exactly the same type in all join points, because we use asm to generate stack maps and asm is not aware of subtyping relationships. This is also the part which is still broken in some corner cases.
- the correct rules for casting primitive types and arrays are already implemented in `StackValue.coerce` and also contain some corner cases
- finally we just want to check whether the source type is assignable to the target type, but we do not have any subtype relationships between asm types. We need to approximate this using the Kotlin subclass relationship (between the underlying types to account for inline classes). However, this is only valid if the asm and Kotlin types are in sync (they go out of sync through boxing with generic types, dispatch receivers, etc.). That's what the `typeMapper.mapType(irType) != type` checks are for. There are only a few cases where this check is going to fail, and we could probably make the optimization less expensive by keeping track of this.

sfs:
Closing this PR in favor of #3327 (which is the only case that ended up breaking things for us).

PR #3175, opened 2020-03-08
KT-37217 Replace UseExperimental with OptIn intention removes target prefixes for annotations

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37217

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/codeInliner/ReplacementPerformer.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/codeInliner/ReplacementPerformer.kt
@@ -19,6 +19,7 @@ package org.jetbrains.kotlin.idea.codeInliner
 import com.intellij.openapi.util.Key
 import com.intellij.psi.PsiTreeChangeAdapter
 import com.intellij.psi.PsiTreeChangeEvent
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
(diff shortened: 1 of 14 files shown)


idea/src/org/jetbrains/kotlin/idea/codeInliner/ReplacementPerformer.kt
```diff
@@ -62,8 +67,9 @@ internal class AnnotationEntryReplacementPerformer(
         assert(range.first == range.last)
         assert(range.first is KtAnnotationEntry)
         val annotationEntry = range.first as KtAnnotationEntry
-        val text = annotationEntry.valueArguments.single().getArgumentExpression()!!.text
-        return annotationEntry.replaced(psiFactory.createAnnotationEntry("@$text"))
+        val text = "@${useSiteTargetText}${annotationEntry.valueArguments.single().getArgumentExpression()!!.text}"
+        val newAnnotationEntry = if (isFileUseSiteTarget) psiFactory.createFileAnnotation(text) else psiFactory.createAnnotationEntry(text)
+        return annotationEntry.replaced(newAnnotationEntry)
```
dimonchik0036, line 72:
```suggestion
        val text = annotationEntry.valueArguments.single().getArgumentExpression()!!.text
        val newAnnotationEntry = if (isFileUseSiteTarget)
            psiFactory.createFileAnnotation(text)
        else
            psiFactory.createAnnotationEntry("@$useSiteTargetText$text")
        return annotationEntry.replaced(newAnnotationEntry)
```

PR #3176, opened 2020-03-09
KT-37213 "Move to top level" intention does not update imports for extension functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37213

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/MoveMemberOutOfObjectIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/MoveMemberOutOfObjectIntention.kt
@@ -25,8 +25,10 @@ import org.jetbrains.kotlin.idea.refactoring.move.moveDeclarations.*
 import org.jetbrains.kotlin.idea.util.application.runWriteAction
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtElement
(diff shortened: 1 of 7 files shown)


vladimirdolzhenko:
looks good to me - the only question - is it possible to hide _Move to top level_ (or at least show some info popup like `Sorry, this intention could not be applied here`) if it is not applicable ?

t-kameyama:
Could you please provide examples that is not applicable?

t-kameyama:
Could you please provide a example that is not applicable?

vladimirdolzhenko:
Sure, I assumed it is quite visible from the code you've submitted:

```
object A {
    private fun foo() {}
} 
```

actually for `foo()` intention is still available but does nothing

t-kameyama:
When declaration is private, it is moved by following code:
https://github.com/t-kameyama/kotlin/blob/862789a23e464868c7c020c4e510c696e025c8dc/idea/src/org/jetbrains/kotlin/idea/intentions/MoveMemberOutOfObjectIntention.kt#L75-L81

Similar test case works correctly.
https://github.com/t-kameyama/kotlin/blob/31a1fb916af7a3de49c2b83c38118f7bcb68e1cc/idea/testData/intentions/moveMemberToTopLevel/function.kt

vladimirdolzhenko:
Sorry, it's my fault - you're right and thanks for the extra details 👍

vladimirdolzhenko:

PR #3178, opened 2020-03-09
JVM_IR: fix accesses from crossinline lambdas in other packages

pyos:
Not sure how many more attempts I'll need to finally get this right.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
@@ -21,7 +21,6 @@ import org.jetbrains.kotlin.codegen.syntheticAccessorToSuperSuffix
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.descriptors.Visibility
-import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.builders.declarations.addValueParameter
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SyntheticAccessorLowering.kt
```diff
@@ -578,18 +563,35 @@ internal class SyntheticAccessorLowering(val context: JvmBackendContext) : IrEle
...
+                context = context.parent as? IrDeclaration ?: return false
+            }
+        }
+
+        val samePackage = ownerClass.getPackageFragment()?.fqName == context.getPackageFragment()?.fqName
+        val fromSubclassOfReceiversClass = !throughCrossinlineLambda &&
+                context.isSubclassOf(ownerClass) && (thisObjReference == null || context.symbol.isSubtypeOfClass(thisObjReference))
         return when {
             // private suspend functions are generated as synthetic package private
             declaration is IrFunction && declaration.isSuspend && declaration.visibility.isPrivate && samePackage -> true
```
max-kammerer: Shouldn't `throughCrossinlineLambda` also be checked here? //cc @ilmirus

max-kammerer: In case smth like:
```
package zz;

// add suspend
private fun suspendFun() {

}

inline fun call(crossinline l: () -> Unit) {
    {
        l();
    }()
}

internal inline fun foo(crossinline s: () -> Unit) {
    call {
        s()
        suspendFun();
    }
}
```
```
package zz2

import zz.foo


fun box() {
    foo {
        println(123)
    }
}
```

pyos: In this case we'll reach `context = inline fun foo` and return `false`.

max-kammerer: Yes, indeed!  Thank you!

max-kammerer:
@pyos  Many thanks!

PR #3179, opened 2020-03-09
Don't assume default constructor in Java classes read from bytecode.  Fixes https://youtrack.jetbrains.com/issue/KT-37107

kevin1e100:
This change limits logic to assume presence of a default constructor to Java classes read from source, where these constructors are implicit.  When classes are read from bytecode or through reflection, default constructors are explicit, so if no constructors are available there really aren't any.


kevin1e100:
Friendly ping here.  Has this been merged?  I'd also prefer to include a test that compiles against a .class file without constructors to avoid regressions.  I'd love to get some guidance on how to go about that though.

udalov:
@kevin1e100 Sorry for the delay! The change looks good.

To add such test, I suppose you could transform a class file with ASM. I'd suggest taking a Kotlin file facade class, removing the `kotlin.Metadata` annotation on it with:
```
ClassReader(classFileBytes).accept(object : ClassVisitor(Opcodes.API_VERSION) {
    override fun visitAnnotation(desc: String, visible: Boolean): AnnotationVisitor? =
        if (desc == JvmAnnotationNames.METADATA_DESC) null else super.visitAnnotation(descriptor, visible)
}, 0)
```
and then compiling something against that class. It should't have constructors in the bytecode, and it will be loaded as a Java class (since there's no Kotlin metadata) by the Kotlin compiler.

The best test class for these manipulations would be `CompileKotlinAgainstCustomBinariesTest`. Check out for example `testInlineFunWithoutDebugInfo` where debug info is deleted and something is compiled against that class.

kevin1e100:
@udalov thank you, that worked like a charm.  I added the additional test to this PR, PTAL!

udalov:
Looks good, thank you! Merged manually in 7448761dfd07e34c95ec9bf024cfde6f38876d32.

PR #2494, opened 2019-07-30
JVM IR: Callable reference SMAP fixes

sfs:
This is a separate PR for the SMAP change from #2483. In particular, the first 8 commits are from #2483, only 743fbd2 is actually new.

PR #3180, opened 2020-03-10
JVM_IR: use fresh source map when generating lambda body for inline.

jsjeon:
Otherwise, file mappings in the source map would be accumulated, resulting in consistent shifting of line numbers (as shown at `smap.kt` test).

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
@@ -171,6 +171,14 @@ object IdenticalSourceMapper : SourceMapper {
         get() = null
 
     override fun mapLineNumber(lineNumber: Int) = lineNumber
+
+    override fun mapLineNumber(source: Int, sourceName: String, sourcePath: String): Int {
+        throw UnsupportedOperationException(
+            "IdenticalSourceMapper#mapLineNumber($source, $sourceName, $sourcePath)\n"
+                    + "This mapper should not encounter a line number out of range of the current file.\n"
+                    + "This indicates that SMAP generation is missed somewhere."
+        )
+    }
 }
 
 class CallSiteMarker(val lineNumber: Int)
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
@@ -117,8 +117,10 @@ class IrSourceCompilerForInline(
         return SMAPAndMethodNode(node!!, SMAP(classCodegen.getOrCreateSourceMapper().resultMappings))
     }
 
-    override fun generateLambdaBody(lambdaInfo: ExpressionLambda): SMAPAndMethodNode =
-        makeInlineNode((lambdaInfo as IrExpressionLambdaImpl).function, codegen.classCodegen, true)
+    override fun generateLambdaBody(lambdaInfo: ExpressionLambda): SMAPAndMethodNode {
+        val function = (lambdaInfo as IrExpressionLambdaImpl).function
+        return makeInlineNode(function, codegen.classCodegen.createClassCodegenForLambdaBody(codegen.classCodegen.irClass, function), true)
+    }
 
     override fun doCreateMethodNodeFromSource(
         callableDescriptor: FunctionDescriptor,
--- a/compiler/testData/codegen/bytecodeText/kt10259_3.kt
+++ b/compiler/testData/codegen/bytecodeText/kt10259_3.kt
@@ -19,15 +19,16 @@ inline fun test(crossinline s: () -> Unit) {
     }()
 }
 
+// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\s
+// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\$1\s
+// 13 INNERCLASS
+
 // JVM_TEMPLATES
 // 3 INNERCLASS Kt10259_3Kt\$test\$1 null
 // 2 INNERCLASS Kt10259_3Kt\$test\$1\$1
-// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\s
-// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\$1\s
 // inlined:
 // 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\$1\$lambda\$1\s
 // 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\$1\$lambda\$1\$1\s
-// 13 INNERCLASS
 
 // NB JVM_IR generates
 //  final static INNERCLASS Kt10259_3Kt$box$1$1 null null
@@ -37,10 +38,7 @@ inline fun test(crossinline s: () -> Unit) {
 // this behavior is equivalent to javac and seems to be correct.
 
 // JVM_IR_TEMPLATES
-// 3 INNERCLASS Kt10259_3Kt\$box\$1\$1\s
+// 2 INNERCLASS Kt10259_3Kt\$box\$1\$1\s
 // 2 INNERCLASS Kt10259_3Kt\$box\$1\$1\$1\s
 // 3 INNERCLASS Kt10259_3Kt\$test\$1\s
 // 2 INNERCLASS Kt10259_3Kt\$test\$1\$1\s
-// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\s
-// 2 INNERCLASS Kt10259_3Kt\$box\$\$inlined\$test\$1\$1\s
-// 14 INNERCLASS
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
@@ -343,7 +343,18 @@ open class ClassCodegen protected constructor(
     }
 
     fun createLocalClassCodegen(klass: IrClass, parentFunction: IrFunction): ClassCodegen =
-            ClassCodegen(klass, context, this, parentFunction, withinInline = withinInline || parentFunction.isInline)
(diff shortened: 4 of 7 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -171,6 +171,8 @@ object IdenticalSourceMapper : SourceMapper {
         get() = null
 
     override fun mapLineNumber(lineNumber: Int) = lineNumber
+
+    override fun mapLineNumber(source: Int, sourceName: String, sourcePath: String) = source
```
kandersen: This change appears unused... ?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
```diff
@@ -127,8 +127,12 @@ class IrSourceCompilerForInline(
         return SMAPAndMethodNode(node!!, SMAP(classCodegen.getOrCreateSourceMapper().resultMappings))
     }
 
-    override fun generateLambdaBody(lambdaInfo: ExpressionLambda): SMAPAndMethodNode =
-        makeInlineNode((lambdaInfo as IrExpressionLambdaImpl).function, codegen.classCodegen, true)
+    override fun generateLambdaBody(lambdaInfo: ExpressionLambda): SMAPAndMethodNode {
+        val function = (lambdaInfo as IrExpressionLambdaImpl).function
+        return makeInlineNode(
+            function, if (lambdaInfo.isCrossInline) codegen.classCodegen else FakeClassCodegen(function, codegen.classCodegen), true
```
pyos: This does not seem right.
1. Why the exception for crossinline lambdas?
2. This affects more than the source maps, e.g. anonymous objects defined in (non-crossinline) inline lambdas would no longer be listed in InnerClasses of the class where the lambda is declared.
Also note that #3056 removes FakeClassCodegen.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
pyos: As I understand, this method should only be called if you have line numbers out of range for the current file. Thus, it's not correct to call it on IdenticalSourceMapper, and if that happens, it means you're missing some SMAPs. (See #2494, for example.)

jsjeon: Thanks for the pointer! And your guess in the other comment was right; I saw some exceptions since a link to the parent function was missed in `FakeClassCodegen`; (re)storing it resolved the issue.

So, to prevent others from accidentally overriding it again, I wonder whether throwing an exception (yeah, super type already throws an exception but) with more detailed explanation makes sense.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
jsjeon: My strategy was simple: in a debug mode, stepping through source map create points side-by-side, i.e., comparing old and new backends. Then, spotted here: https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SourceCompilerForInline.kt#L153, an override of `generateLambdaBody` in old backend, which in turn goes to `generateMethodBody` where `FakeMemberCodegen` (a counterpart of `FakeClassCodegen`) is always used.

> 1. Why the exception for crossinline lambdas?

From here: https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SourceCompilerForInline.kt#L152 the control goes down to here: https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/context/InlineLambdaContext.kt#L29 where, for crossinline, the original context was used. Well, it has other conditions too, and I think those other conditions are handled in `IrSourceCompilerForInline#makeInlineNode` I think.

> 2. ... anonymous objects defined in (non-crossinline) inline lambdas would no longer be listed in InnerClasses of the class where the lambda is declared.

To be honest, I can't follow this fully. I wonder, though, we don't have good test coverage for that scenario (since no test is broken)?

> 3. FakeClassCodegen here loses the link to the parent function ... (this is probably why you needed the other change to avoid exceptions).

You're right. Delegating parent function (in FakeClassCodegen) resolves it. Also, not needed to override a map function in identical source mapper.

> Also note that #3056 removes FakeClassCodegen.

Thanks for the heads-up. I've tried to rewrite this change without using FakeClassCodegen, like that PR does, but not successful yet. :\

pyos: >To be honest, I can't follow this fully. I wonder, though, we don't have good test coverage for that scenario (since no test is broken)?

We *kind of* do. The IR backend generates objects in lambdas differently from the old backend: in the old backend, objects capture the lambda itself, then use that "outer this" to access all captured variables; in the new backend, objects capture everything they need directly. Because of this, the old backend needs to regenerate those objects after inlining the lambdas (since there's no longer a lambda object) while the new backend does not. This leads to visible differences in class names, which makes tests for InnerClass attributes fail on JVM_IR.

(This is also why it's fine to use a completely fake codegen in the old backend, where the classes are going to be removed anyway and replaced with regenerated copies, but not in the old backend, where you need to generate the original object correctly.)

jsjeon: Alright, I have to admit a special handling of `crossinline` at this point is quite awkward, but was not able to understand a new test failure. See my other comment inlined to that test. Perhaps, it's still related to InnerClasses...?

Also, my new change is... yeah it's still another variant of FakeClassCodegen, but in preparation of sunset of it at #3056 . Still can't grasp the ideas of that PR and rewrite things on top of it...

compiler/testData/codegen/boxInline/smap/crossroutines.kt
```diff
@@ -1,4 +1,4 @@
-// This test depends on line numbers
+// IGNORE_BACKEND_MULTI_MODULE: JVM_IR
```
jsjeon: At lines 128 and 142, source name (I guess?) became `_2Kt`, instead of `_2Kt$box$1`. @pyos can you spot what could go wrong here?

pyos: It's because you're using the wrong class when creating a fake `ClassCodegen`.

```kotlin
fun f() {
    someNonInlineFunction {
        someInlineFunction {
            println("Hello, World!")
        }
    }
}
```

is transformed into

```kotlin
class AKt {
    fun f() {
        someNonInlineFunction(`AKt$f$1`())
    }
    fun `f$lambda-1`() {
        println("Hello, World!")
    }
}

class `AKt$f$1` : ... {
    override fun invoke() {
        someInlineFunction(AKt::`f$lambda-1`)
    }
    // + other stuff
}
```
Note that when inlining `someInlineFunction`, `codegen.classCodegen.irClass` is `AKt$f$1`, but the parent class of `f$lambda-1` is `AKt`.

In any case, there's still the matter of inner classes, so this approach is not fully correct either.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -171,6 +171,14 @@ object IdenticalSourceMapper : SourceMapper {
         get() = null
 
     override fun mapLineNumber(lineNumber: Int) = lineNumber
+
+    override fun mapLineNumber(source: Int, sourceName: String, sourcePath: String): Int {
+        throw UnsupportedOperationException(
+            "IdenticalSourceMapper should not encounter a line number out of range of the current file."
+                    + System.lineSeparator()
+                    + "This indicates that SMAP generation is missed somewhere."
```
max-kammerer: Please add source, sourceName, sourcePath for exception message

max-kammerer: also pls use '\n' for line separator

jsjeon: Updated.

compiler/testData/codegen/boxInline/smap/crossroutines.kt
jsjeon: Thanks for the pointer. Using the right parent IrClass made this test see correct SMAP again.

I didn't close your other comment above, so let's move on to there.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrSourceCompilerForInline.kt
jsjeon: 1. special handling of `crossinline` is gone.
3. link(s) to parent function (as well as parent class) are handled.

Remaining one is 2. Not sure parent class is the right term here; using the current context's IR class is same as using the enclosing class or maybe not? Linking it to the new ClassCodegen is not enough to keep InnerClasses attributes for that enclosing class?

Anyhow, the latest commit is a bit different from the initial one, so PTAL.

compiler/testData/codegen/bytecodeText/kt10259_3.kt
```diff
@@ -37,10 +37,10 @@ inline fun test(crossinline s: () -> Unit) {
 // this behavior is equivalent to javac and seems to be correct.
 
 // JVM_IR_TEMPLATES
-// 3 INNERCLASS Kt10259_3Kt\$box\$1\$1\s
+// 2 INNERCLASS Kt10259_3Kt\$box\$1\$1\s
```
max-kammerer: Templates in JVM_IR_TEMPLATES and JVM_TEMPLATES mostly looks identical. Could you extract common part in header, e.g.
// COMMON_TEMPLATE_1
...

// JVM_TEMPLATES
// JVM_PATTERN_1
...
//JVM_IR_TEMPLATES
// JVM_IR_TEMPLATE_1
...

here and in other affected files

jsjeon: Didn't know that we can put common part ahead. Done in that way, but not that many due to the slightly different inner-class names.

max-kammerer:
Thank you!

compiler/testData/codegen/bytecodeText/kt10259_3.kt
max-kammerer: Oh..I expected more commons, sorry

PR #3182, opened 2020-03-10
KT-15846 'Change lambda expression return type' quick fix does nothing

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-15846


darthorimar:
Thank you :)

PR #3183, opened 2020-03-11
KT-37312 "Implement members" intention put function in the primary constructor if there are unused brackets in class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37312


darthorimar:
Thank you for your contribution :smiley:

PR #3184, opened 2020-03-11
JVM IR: Mangle interface implementation methods in inline classes

sfs:
Fixes [KT-37024](https://youtrack.jetbrains.com/issue/KT-37024).


sfs:
Moved things around a bit to make it clear that this is not relevant for fake overrides.

udalov:
Looks good, thanks!

PR #3186, opened 2020-03-12
Add IrProviders to IrPluginContext

jimgoog:
Needed to allow IR symbol binding in IR plugins after https://github.com/JetBrains/kotlin/commit/a2d8e2435fbd93e52ee047d8e0dbf61adf945dd9

This change becomes less necessary after https://youtrack.jetbrains.com/issue/KT-37255 gets fixed, but even after https://youtrack.jetbrains.com/issue/KT-37255, it seems likely that we will want to provide plugins with a way of binding symbols that were introduced by the plugin and thus not referenced when symbols were generated by Psi2Ir.

We are still in the process of rebasing, so I haven't yet been able to test this exact change (we tested a similar change applied against an older revision), but I think this gets us what we need here, and wanted to post it for your review.  We can merge now or wait until the rebase is completely done on our end.

cc @erokhins

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/extensions/IrGenerationExtension.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/extensions/IrGenerationExtension.kt
@@ -12,6 +12,7 @@ import org.jetbrains.kotlin.extensions.ProjectExtensionDescriptor
 import org.jetbrains.kotlin.ir.builders.IrGeneratorContext
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
 import org.jetbrains.kotlin.ir.descriptors.IrBuiltIns
+import org.jetbrains.kotlin.ir.util.IrProvider
 import org.jetbrains.kotlin.ir.util.SymbolTable
 import org.jetbrains.kotlin.ir.util.TypeTranslator
 import org.jetbrains.kotlin.resolve.BindingContext
@@ -23,7 +24,8 @@ class IrPluginContext(
     val symbolTable: SymbolTable,
     val typeTranslator: TypeTranslator,
     override val irBuiltIns: IrBuiltIns,
-    val symbols: BuiltinSymbolsBase = BuiltinSymbolsBase(irBuiltIns.builtIns, symbolTable)
+    val symbols: BuiltinSymbolsBase = BuiltinSymbolsBase(irBuiltIns.builtIns, symbolTable),
+    val irProviders: List<IrProvider>
 ) : IrGeneratorContext()
 
 interface IrGenerationExtension {
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendFacade.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendFacade.kt
@@ -34,6 +34,17 @@ object JvmBackendFacade {
         val psi2ir = Psi2IrTranslator(state.languageVersionSettings, signaturer = signaturer)
         val psi2irContext = psi2ir.createGeneratorContext(state.module, state.bindingContext, extensions = extensions)
 
+        val stubGenerator = DeclarationStubGenerator(
+            psi2irContext.moduleDescriptor, psi2irContext.symbolTable, psi2irContext.irBuiltIns.languageVersionSettings, extensions
+        )
+        val deserializer = JvmIrLinker(
+            EmptyLoggingContext, psi2irContext.irBuiltIns, psi2irContext.symbolTable
+        )
+        psi2irContext.moduleDescriptor.allDependencyModules.filter { it.getCapability(KlibModuleOrigin.CAPABILITY) != null }.forEach {
+            deserializer.deserializeIrModuleHeader(it)
+        }
+        val irProviders = listOf(deserializer, stubGenerator)
+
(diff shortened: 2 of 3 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/extensions/IrGenerationExtension.kt
```diff
@@ -23,7 +24,8 @@ class IrPluginContext(
     val symbolTable: SymbolTable,
     val typeTranslator: TypeTranslator,
     override val irBuiltIns: IrBuiltIns,
-    val symbols: BuiltinSymbolsBase = BuiltinSymbolsBase(irBuiltIns.builtIns, symbolTable)
+    val symbols: BuiltinSymbolsBase = BuiltinSymbolsBase(irBuiltIns.builtIns, symbolTable),
+    val irProviders: List<IrProvider>
```
bashor: I'd suggest to add an ability to provide only one `IrProvider` and pass `CompositeIrProvider` (unexisting yet).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmBackendFacade.kt
```diff
@@ -53,6 +65,7 @@ object JvmBackendFacade {
         val stubGenerator = DeclarationStubGenerator(
             psi2irContext.moduleDescriptor, psi2irContext.symbolTable, psi2irContext.irBuiltIns.languageVersionSettings, extensions
         )
+        stubGenerator.setIrProviders(irProviders)
         val deserializer = JvmIrLinker(
```
gogabr: Move to line 47?

gogabr: Do we need to recreate the `IrProvider` setup here? Why not reuse the one you built earlier?

gogabr:
As it is, the code does not even compile.
I fixed the compilation and created an Upsource review: https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-3984.

raniejade:
Is there still plans to work on this?

gogabr:
There is a rework underway of the way `IrProvider`s work. After that is merged, we'll need to decide whether something similar to this change is needed. 
See https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4039

raniejade:
@gogabr Any ETA on it? Do you know what version of Kotlin will this get fixed?

romanart:
Will try to make story a bit clear.
We have a plan to make some design and improve IR plugin API by M3. 
In short we are going to try to give following guarantees about symbols when IR plugin is being applied: 
1. Any declaration from **current** module has its IR-representation. In other word, any symbol from current compiling module is bound.
2. Any referenced external symbol is bound.
3. Any external class requested on plugin registration step and any its declaration is accessible via IR tree API and bound too.

What guarantees we cannot give right by M3:
1. Symbol referenced inside plugin which are not in scope of any declaration mentioned in previous part.
2. Symbol which is not accessible in compile time (no corresponding library provided).

Are those guarantees enough for you cases?

raniejade:
Yes, my plugin just scans classes in the current module if it's a certain type and based on that creates top level property which calls a function in my library to register that type.

jimgoog:
@romanart 

I think I am just not familiar with all the terminology you use.  When you say "Any referenced external symbol is bound", is that any symbol referenced prior to the plugin running, or any external symbol referenced during plugin execution (including symbols created within the plugin which reference external classes/functions)?  When you say "Any external class requested on plugin registration", what are you referring to here / how does a plugin request external classes on plugin registration / is this a new API being added?

just to clarify:

If a plugin adds a call to a class/function that wasn't previously used in the user's code, and isn't defined/used in that module, but is on the classpath, will that symbol be bound?  Specifically, suppose the user's code references `kotlin.jvm.functions.Function5`, and our plugin decides to add two additional parameters (making the call a `kotlin.jvm.functions.Function7`).  Obviously `Function5` will be bound, but I just want to ensure that when our plugin introduces the symbol referencing the external class `kotlin.jvm.functions.Function7` will that that symbol will become bound as well, as per https://youtrack.jetbrains.com/issue/KT-37255#focus=streamItem-27-4076393.0-0 ?

romanart:
@jimgoog 

>When you say "Any referenced external symbol is bound", is that any symbol referenced prior to the plugin running, or any external symbol referenced during plugin execution (including symbols created within the plugin which reference external classes/functions)?

I mean `prior to the plugin running`

>When you say "Any external class requested on plugin registration", what are you referring to here / how does a plugin request external classes on plugin registration / is this a new API being added?

Yes, a new API will be provided. 

>If a plugin adds a call to a class/function that wasn't previously used in the user's code, and isn't defined/used in that module, but is on the classpath, will that symbol be bound?

Yes, that is an idea. Plugin has to request such symbols and infra provides them. It's kind of temporary solution and I hope it will be deprecated some day once required infrastructure is implemented.

>Specifically, suppose the user's code references kotlin.jvm.functions.Function5, and our plugin decides to add two additional parameters (making the call a kotlin.jvm.functions.Function7)

For that specific case answer should be "yes". Moreover, seems like there is lake of API in `IrBuiltIns` where such symbols has (`*Function*`) to be accessed to make plugin platform independent. But as I said in first comment at the current moment we cannot give such guarantee for random symbol. It's not a final solution.

romanart:
I think it could be closed.

PR #3188, opened 2020-03-13
[FIR] add support for translating augmented assignments of array acce…

juan-chen:
…ss expressions.

Currently they're translated to FirArraySetCall, which are then translated to IrErrorCall.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
@@ -6,11 +6,13 @@
 package org.jetbrains.kotlin.fir.resolve.transformers.body.resolve
 
 import com.intellij.openapi.progress.ProcessCanceledException
+import org.jetbrains.kotlin.KtNodeTypes.ARRAY_ACCESS_EXPRESSION
 import org.jetbrains.kotlin.fir.*
 import org.jetbrains.kotlin.fir.declarations.FirTypeParametersOwner
 import org.jetbrains.kotlin.fir.diagnostics.DiagnosticKind
 import org.jetbrains.kotlin.fir.diagnostics.FirSimpleDiagnostic
 import org.jetbrains.kotlin.fir.expressions.*
+import org.jetbrains.kotlin.fir.expressions.builder.buildArgumentList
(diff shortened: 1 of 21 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -270,20 +275,33 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
             return when {
                 operatorCallReference == null || (!lhsIsVar && !assignIsError) -> resolvedAssignCall.compose()
                 assignCallReference == null -> {
-                    val assignment =
-                        buildVariableAssignment {
+                    if (leftArgument is FirFunctionCall && leftArgument.calleeReference.source?.psi is KtArrayAccessExpression) {
+                        val setCall = buildFunctionCall {
```
mglukhikh: I'm not quite sure here, but may be in this case we should create function call from the beginning (I mean, in raw FIR)?

demiurg906:
I think it's better to stay building fir for array assignments as is and implement transformation of array set call in `FirExpressionBodyResolveTransformer` (currently it has default implementation like `transformChildren(transformer, data)`)

demiurg906:
Also there is missing some fir diagnostic test that checks behavior of different combinations of `get`, `set`, `plus`, `plusAssign` like

```kotlin
class A<T> {
    operator fun get(index: Int): T = null!!
    operator fun set(index: Int, value: T) {}
}

class B {
    operator fun plusAssign(other: B) {}
}

class C {
    operator fun plus(other: C): C = this
}

class D {
    operator fun plusAssign(other: D) {}
    operator fun plus(other: D): D = this
}

fun test_1(a: A<B>) {
    a[0] += B() // get, plusAssign
}

fun test_2(a: A<C>) {
    a[0] += C() // get, set, plus
}

fun test_3(a: A<D>) {
    a[0] += D() // ambiguity
}
```

demiurg906:
@juan-chen Thank you for request. I take your commit and fix it by myself, if you don't mind. There are some corner cases that faster to implement than to describe them to you

mglukhikh:
Finally implemented in this commit: https://github.com/JetBrains/kotlin/commit/2b986194fb15f8044c40c66c23230c6f4caa3293 and this smaller one: https://github.com/JetBrains/kotlin/commit/f85022532d0f0819241439a6d7f43a5068222fe9.

PR #3189, opened 2020-03-14
KT-37496 False positive "Remove redundant backticks" for multiple underscores variable name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37496


darthorimar:
LGTM, thanx!

PR #3190, opened 2020-03-14
KT-29969 Making var arg kParameters default to empty array no argument given

nerdsupremacist:
Fixes: [KT-29969](https://youtrack.jetbrains.com/issue/KT-29969)

`KCallable.callBy` did not allow for empty `vararg` parameters

### DIFF
--- a/core/reflection.jvm/src/kotlin/reflect/jvm/internal/KCallableImpl.kt
+++ b/core/reflection.jvm/src/kotlin/reflect/jvm/internal/KCallableImpl.kt
@@ -135,7 +135,7 @@ internal abstract class KCallableImpl<out R> : KCallable<R> {
                     anyOptional = true
                 }
                 parameter.isVararg -> {
-                    arguments.add(emptyArray<Any>())
+                    arguments.add(parameter.type.asArrayType()?.emptyArray())
                 }
                 else -> {
                     throw IllegalArgumentException("No argument provided for a required parameter: $parameter")
@@ -177,7 +177,7 @@ internal abstract class KCallableImpl<out R> : KCallable<R> {
                     args[parameter] ?: throw IllegalArgumentException("Annotation argument value cannot be null ($parameter)")
                 }
                 parameter.isOptional -> null
-                parameter.isVararg -> emptyArray<Any>()
+                parameter.isVararg -> parameter.type.asArrayType()?.emptyArray()
                 else -> throw IllegalArgumentException("No argument provided for a required parameter: $parameter")
             }
         }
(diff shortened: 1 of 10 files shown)


udalov:
Tests on `callBy` are at `compiler/testData/codegen/box/reflection/callBy/`. Please add some and use the `GenerateCompilerTests` IntelliJ run configuration.

core/reflection.jvm/src/kotlin/reflect/jvm/internal/util.kt
```diff
@@ -195,3 +199,10 @@ internal fun <M : MessageLite, D : CallableDescriptor> deserializeToDescriptor(
     )
     return MemberDeserializer(context).createDescriptor(proto)
 }
+
+@Suppress("UNCHECKED_CAST")
+internal fun KType.asArrayType(): KClass<Array<*>>? = jvmErasure
```
udalov: I think this function signature is incorrect because the component type can be primitive, and then it would return a value of e.g. `KClass<ByteArray>?`, which is not a subtype of `KClass<Array<*>>`.

Also please add tests on primitive varargs in both cases

core/reflection.jvm/src/kotlin/reflect/jvm/internal/KCallableImpl.kt
```diff
@@ -134,6 +134,9 @@ internal abstract class KCallableImpl<out R> : KCallable<R> {
                     mask = mask or (1 shl (index % Integer.SIZE))
                     anyOptional = true
                 }
+                parameter.isVararg -> {
+                    arguments.add(parameter.type.asArrayType()?.emptyArray())
```
udalov: In core and compiler codebase, and especially in reflection, we prefer to fail fast in any unknown scenario. So here I'd rather propose to assert that the type's JVM erasure is array first (and throw `KotlinReflectionInternalError` with a meaningful message otherwise), and then take its component type and create an empty array. And naturally this could be extracted into a single function to avoid duplication.

core/reflection.jvm/src/kotlin/reflect/jvm/internal/util.kt
nerdsupremacist: This seems to work just fine. The Unchecked cast seems to still go through ¯\_(ツ)_/¯ 

But I can change it if you like

core/reflection.jvm/src/kotlin/reflect/jvm/internal/KCallableImpl.kt
nerdsupremacist: Alright so for the sake of making the message more specific I'll move this logic to `KCallableImpl` since it's the one that should know what error to throw

udalov:
Merged manually with minor corrections in 34a64d9171d.

Thank you!

PR #3194, opened 2020-03-17
Detect inline cycles faster

pyos:
E.g. in the following code

    fun x() {}
    inline fun f() { x(); g() }
    inline fun g() { x(); f() }

the old implementation of inline cycle detection bailed out after generating 3 calls of x() in each function, while the new one stops after 2. In other words, code generation for a single function is no longer reentered.


max-kammerer:
Previous changes made around cycle detection by @dnpetrov have aim to avoid descriptor usage in ir part. Please avoid descriptor usage in IR

pyos:
It's not possible to completely avoid descriptors here because they're necessary for diagnostics anyway. #3164 will eliminate the need for the `originalDescriptor` hack, so line 160 of IrInlineCodegen.kt will become

```
state.globalInlineContext.enterIntoInlining(codegen.irFunction.descriptor, expression.symbol.owner.descriptor, element)
```

which is about as little use of descriptors as possible here.

max-kammerer:
@pyos aftrer short private discussion with @dnpetrov we decided that descriptor usage here is ok

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -145,36 +144,22 @@ class IrInlineCodegen(
...
-                codegen.context.psiSourceManager.findPsiElement(irFunctionAccessExpression, function)
-                    ?: codegen.context.psiSourceManager.findPsiElement(function)
-
-        override val id: Any
-            get() = irFunctionAccessExpression
-
-        override fun toString(): String = irFunctionAccessExpression.render()
-    }
+    private fun IrFunction.originalDescriptor() =
+        if (isSuspend) suspendFunctionOriginal().descriptor else descriptor
```
max-kammerer: @pyos Please add test with suspends

max-kammerer:
I'm also observing different behaviour in next case:
```
inline fun test(crossinline h: () -> String) {
    object  {
        inline fun g() {
            foo()
        }
    }
}

inline fun foo() {
    test {"123"}
}
```
Now it's exception during compilation, before proper error reporting

Looks like cause of 
`
assert(context.original !in processing) { "entered inlining cycle on $context" }
`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
pyos: done

pyos:
Yeah, I wanted to do this without adding calls to push/pop the context around expression codegen, but I guess that's not possible. Fixed.

max-kammerer:
So looks like that `context` in enterIntoInlining is redundant or not?

pyos:
It would be redundant if LinkedHashMap provided a method to retrieve the last element, but for some reason it does not. So it's either the `context` parameter or additional memory in form of a stack of keys.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/GlobalInlineContext.kt
```diff
@@ -5,24 +5,44 @@
...
+            return false
         }
+        typesUsedInInlineFunctions.push(hashSetOf())
+        return true
+    }
 
-    fun exitFromInliningOf(call: InlineCall?) {
-        inlineCycleReporter.exitFromInliningOf(call)
+    fun exitFromInlining(context: CallableDescriptor) {
+        inlineCalls[context.original] = null
```
max-kammerer: Shoudn't it be 'remove'?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/GlobalInlineContext.kt
```diff
@@ -5,24 +5,44 @@
...
+    fun enterIntoInlining(context: CallableDescriptor, callee: CallableDescriptor?, element: PsiElement?): Boolean {
+        assert(context.original in inlineCalls) { "$context not in current declaration stack" }
+        inlineCalls[context.original] = element
+        if (callee?.original in inlineCalls) {
+            var currentElement = element
+            for ((callContainer, callElement) in inlineCalls.entries.dropWhile { it.key != callee?.original }) {
+                currentElement?.let { diagnostics.report(Errors.INLINE_CALL_CYCLE.on(it, callContainer)) }
+                currentElement = callElement // next container is the target of this call
+            }
+            inlineCalls[context.original] = null
```
max-kammerer: ...and here too

pyos:
Actually, never mind, I'll rewrite this with a normal map and a stack.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/GlobalInlineContext.kt
pyos: The original idea was that enterDeclaration adds the element, then enterIntoInlining sets the value, then exitFromInlining clears it, then steps 2-3 are repeated, and finally exitDeclaration removes the key. That's not relevant anymore.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/GlobalInlineContext.kt
```diff
@@ -5,24 +5,48 @@
...
+    // No call in it should point to a declaration that's before it in the stack.
+    private val inlineCallsAndDeclarations = LinkedList<Any? /* CallableDescriptor | PsiElement? */>()
+    private val inlineDeclarationSet = mutableSetOf<CallableDescriptor>()
 
     private val typesUsedInInlineFunctions = LinkedList<MutableSet<String>>()
 
-    fun enterIntoInlining(call: InlineCall?) =
-        inlineCycleReporter.enterIntoInlining(call).also {
-            if (it) typesUsedInInlineFunctions.push(hashSetOf())
+    fun enterDeclaration(descriptor: CallableDescriptor) {
```
max-kammerer: What about inline flag check?

pyos: It's not really necessary, all other declarations are simply not used for anything.

max-kammerer:
Thank you, now it's much clearer (at least for me).

PR #3196, opened 2020-03-17
JVM_IR: use non-nullable types as SAM super types

pyos:
The wrappers for conversions to nullable and non-nullable SAM types are the same, so differentiating those has no effect other than class name collisions.


udalov:
Thank you!

PR #3197, opened 2020-03-17
Change ArrayConstructorLowering to check the Array class symbol by Fq…

juan-chen:
…Name.

Currently it checks by equality comparisons with the builtin arrayClass symbol, which Fir2Ir doesn't use when translating array constructor calls. This causes array constructor calls from the new frontend not to be lowered.


mglukhikh:
Juan, in such situations it's generally better to change FIR2IR itself than change backend clients, like this lowering you changed here. Today I pushed a pack of commit with FIR2IR fixes. I included there an integration of IR built-in types into FIR2IR, so now FIR2IR does use them. See this commit: https://github.com/JetBrains/kotlin/commit/6739135d17323f3ede8a3843506267cf833115ea.

PR #3198, opened 2020-03-18
KT-37240 Deterministic order for webpack config patches

turansky:
Additional fix - only files with `js` extension processed (directories excluded)

`listFiles()` already contains 'is directory' check. Duplicated check removed


ilgonmic:
Thanks a lot!

PR #3202, opened 2020-03-18
Do fewer unnecessary anonymous object regenerations

pyos:
**Commit 1:**

All reified type parameters referenced by objects in non-default lambdas are from outside the inline call, thus they cannot be substituted yet. This means if the only reason for object regeneration is reification, then regeneration isn't actually necessary.

E.g. in this code:

    inline fun <reified T> f(x: () -> Unit) = x()
    inline fun <reified V> g() = f<V> { { println(V::class.simpleName) }() }

we first inline the lambda into f(), then inline f()-with-lambda into g(). Because the object inside the lambda captures nothing from outside g(), the original class can be used both times.

**Commit 2**:

Same thing, but instead of reified type parameters it's suspend lambdas: if the actual lambda to inline is not known yet (or will never be known because the inline root is not an inline function or the captured parameter is `noinline`), regenerating an object is pointless.

This brings the JVM_IR backend into complete compliance with KT-28064.

### DIFF
--- a/compiler/testData/codegen/bytecodeListing/coroutines/tcoContinuation_ir.txt
+++ b/compiler/testData/codegen/bytecodeListing/coroutines/tcoContinuation_ir.txt
@@ -213,28 +213,6 @@ public final class TcoContinuationKt$transform$$inlined$flow$1$1 {
 }
 
 @kotlin.Metadata
-public final class TcoContinuationKt$transform$$inlined$flow$1$lambda$1$1 {
-    field label: int
-    synthetic field result: java.lang.Object
-    synthetic final field this$0: TcoContinuationKt$transform$$inlined$flow$1$lambda$1
-    inner class TcoContinuationKt$transform$$inlined$flow$1$lambda$1
-    inner class TcoContinuationKt$transform$$inlined$flow$1$lambda$1$1
-    public method <init>(p0: TcoContinuationKt$transform$$inlined$flow$1$lambda$1, p1: kotlin.coroutines.Continuation): void
-    public final @org.jetbrains.annotations.Nullable method invokeSuspend(@org.jetbrains.annotations.NotNull p0: java.lang.Object): java.lang.Object
-}
-
-@kotlin.Metadata
-public final class TcoContinuationKt$transform$$inlined$flow$1$lambda$1 {
-    synthetic final field $this$inlined: FlowCollector
-    synthetic final field $transformer$inlined: kotlin.jvm.functions.Function3
-    inner class TcoContinuationKt$transform$$inlined$flow$1$lambda$1
-    inner class TcoContinuationKt$transform$$inlined$flow$1$lambda$1$1
-    public method <init>(p0: kotlin.jvm.functions.Function3, p1: FlowCollector): void
-    public @org.jetbrains.annotations.Nullable method emit$$forInline(p0: java.lang.Object, @org.jetbrains.annotations.NotNull p1: kotlin.coroutines.Continuation): java.lang.Object
-    public @org.jetbrains.annotations.Nullable method emit(p0: java.lang.Object, @org.jetbrains.annotations.NotNull p1: kotlin.coroutines.Continuation): java.lang.Object
-}
-
-@kotlin.Metadata
 public final class TcoContinuationKt$transform$$inlined$flow$1 {
     synthetic final field $this$inlined: Flow
     synthetic final field $transformer$inlined: kotlin.jvm.functions.Function3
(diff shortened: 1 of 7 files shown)


compiler/testData/codegen/bytecodeListing/coroutines/tcoContinuation_ir.txt
```diff
@@ -301,36 +257,80 @@ public final class TcoContinuationKt$transform$$inlined$flow$3$1 {
...
+
+@kotlin.Metadata
+public final class TcoContinuationKt$transform$lambda-1$$inlined$collect$2$1 {
     field label: int
     synthetic field result: java.lang.Object
-    synthetic final field this$0: TcoContinuationKt$transform$$inlined$flow$3$lambda$1
-    inner class TcoContinuationKt$transform$$inlined$flow$3$lambda$1
-    inner class TcoContinuationKt$transform$$inlined$flow$3$lambda$1$1
-    public method <init>(p0: TcoContinuationKt$transform$$inlined$flow$3$lambda$1, p1: kotlin.coroutines.Continuation): void
+    synthetic final field this$0: TcoContinuationKt$transform$lambda-1$$inlined$collect$2
```
max-kammerer: @ilmirus Please take a look.

I'm also confused a bit with `-1` in class names

max-kammerer: Could we switch `-` to '_' or smth else (e.g. '$')?

ilmirus: LGTM, the `-1` part is needed, when there are several lambda arguments. @pyos, can you, please, add a test with something like
```
interface FlowCollector<T> {
    suspend fun emit(value: T)
}
interface Flow<T> {
    suspend fun collect(collector: FlowCollector<T>)
}
public inline fun <T> flow(crossinline block: suspend FlowCollector<T>.() -> Unit) = object : Flow<T> {
    override suspend fun collect(collector: FlowCollector<T>) = collector.block()
}
suspend inline fun <T> Flow<T>.collect(crossinline action: suspend (T) -> Unit, crossinline action2: suspend (T) -> Unit): Unit =
    collect(object : FlowCollector<T> {
        override suspend fun emit(value: T) = action(value)
    })
public inline fun <T, R> Flow<T>.transform(crossinline transformer: suspend FlowCollector<R>.(value: T) -> Unit): Flow<R> {
    return flow {
        return@flow collect({ value ->
            return@collect transformer(value)
        }, {})
    }
}
public inline fun <T, R> Flow<T>.map(crossinline transformer: suspend (value: T) -> R): Flow<R> = transform { value -> return@transform emit(transformer(value)) }
suspend fun foo() {
    flow<Int> {
        emit(1)
    }.map { it + 1 }
        .collect ({
        }, {})
}
```

pyos: Sure, but could you please clarify what this test should check that's not covered by others?

pyos: Actually, the `-1` part is not really necessary. Note that all of these are regenerated objects: if the names of lambdas clash, they will simply get different indices because of 07bde889b4fa7b40bb7e3b8737093ef18713eb2d

pyos: >Could we switch - to '_' or smth else (e.g. '$')?

To answer this question, yes, `ClassCodegen.getRegeneratedObjectNameGenerator` can be changed to give these objects different names.

ilmirus: I see, this should already be covered by `crossinlineSuspendLambdaInsideCrossinlineSuspendLambda.kt`.

max-kammerer: > To answer this question, yes, ClassCodegen.getRegeneratedObjectNameGenerator can be changed to give these objects different names

Cool, it could be postponed to separate pr

max-kammerer:
@pyos  Could you also add explicit `inline` to lambdas in commit/source comments

max-kammerer:
Any non-suspend test  with bytecode listing would be appreciated

pyos:
Done. I've already encountered bugs with regeneration of objects with reified type parameters in 2f0f4e570f1ca2cdb5f879bcc0e7324e7c01e344, so I just added the CHECK_BYTECODE_LISTING directive to that test. You can see the effect of this PR because the dump has `_2Kt$complicatedCast$1$1` while without the PR it would be `_2Kt$complicatedCast$$inlined$myRun$lambda$2`.

pyos:
I just realized that I need to rebase on master because this bytecode listing is affected by #3180.

pyos:
Rebased.

max-kammerer:
Many thanks!

PR #3204, opened 2020-03-20
JVM_IR: restore InnerClasses for objects in lambdas

pyos:
First commit of #3056

PR #3205, opened 2020-03-20
JVM_IR: refactor ClassCodegen and FunctionCodegen

pyos:
Non-behavior-affecting parts of #3056 and #3164. Depends on #3204.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
@@ -354,7 +354,22 @@ open class ClassCodegen protected constructor(
         }
 
         val node = FunctionCodegen(method, this).generate()
-        node.accept(with(node) { visitor.newMethod(method.OtherOrigin, access, name, desc, signature, exceptions.toTypedArray()) })
+        val mv = with(node) { visitor.newMethod(method.OtherOrigin, access, name, desc, signature, exceptions.toTypedArray()) }
+        if (method.hasContinuation() || method.isInvokeSuspendOfLambda()) {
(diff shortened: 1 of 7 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -384,8 +353,26 @@ open class ClassCodegen protected constructor(
             return
         }
 
-        val signature = FunctionCodegen(method, this).generate().asmMethod
-        jvmSignatureClashDetector.trackMethod(method, RawSignature(signature.name, signature.descriptor, MemberKind.METHOD))
+        val node = FunctionCodegen(method, this).generate()
+        val mv = with(node) { visitor.newMethod(method.OtherOrigin, access, name, desc, signature, exceptions.toTypedArray()) }
```
max-kammerer: NB: cached bytecode in general case would be totally different from writen one: cause optimizations are not perfromed (visitor.newMethod adds a bunch of optimization wrappers)

max-kammerer: Oh, sorry it's not cached

max-kammerer:
By unknown reason `org.jetbrains.kotlin.codegen.ir.FirBlackBoxCodegenTestGenerated.Coroutines.VarSpilling#testRefinedIntTypesAnalysis` test is unmuted. @pyos Please unmute

pyos:
Done. The test no longer fails because the `suspend fun` in it is trivial so the only reason compilation fails is because it has no PSI element. With this refactoring, the check is not triggered unless a diagnostic on said element is actually needed.

max-kammerer:
Thank you!

PR #3206, opened 2020-03-20
FIR: regard property accessor without body as FirDefaultPropertyAccessor

jsjeon:
Otherwise, those will fall into function declaration without body, resulting in NPE when generating the body.


jsjeon:
As I mentioned at quip, running tests via gradle, like `./gradlew :compiler:cleanTest :compiler:test --parallel` can't catch unmutable tests, whereas I found more and more by running them in IDE's gradle tests. Not sure there are some known issues on FIR blackbox tests or just some issues on my gradle/IDE settings.

mglukhikh:
We support also FIR builder from so-called lighter AST. It works faster but has some additional problems with source identification. Everytime you change something in RawFirBuilder, you should change the same thing in `fir:lightTree` module. See implementation of your fix for lighter AST here: https://github.com/JetBrains/kotlin/commit/1036f41e584b44e10dcbbc2b48290305f5e9268d. 

Everything else looks good, I will do remote run to be sure and merge this.

jsjeon:
Thanks for the quick review! I wonder, though, you saw the second-to-last commit. I also noticed some failures on `fir:lightTree` and came up with a similar fix already: https://github.com/JetBrains/kotlin/pull/3206/commits/e182fec5f74860876232ed7c63153277f6a9191a#diff-325cd09e6103c1d9bb56e787db1924fc

mglukhikh:
I suspect I started to review it before you force-pushed this fix. However, it's not important. I've merged it, thank you 👍

PR #3207, opened 2020-03-20
FIR: Support diagnostic SUPER_IS_NOT_AN_EXPRESSION

lunakoly:
Added support for SUPER_IS_NOT_AN_EXPRESSION diagnostic and FirDiagnosticWithParameters1


mglukhikh:
There are things missed in this PR:

- all commits should be squashed into one
- FIR tests should be run and their test data should be fixed
- simultaneously with any `RawFirBuilder` change, we should change `fir:lightTree` module when another method of FIR building (via lighter AST nodes) is implemented

The ~ final version is here: https://github.com/JetBrains/kotlin/commit/98ed7719c070d81865d6dec7e5988ba040940105. See, in particular, changes in `ExpressionsConverter` and in `implicitInvokeOnSuper.kt` (test data). I will perform remote run and then merge your changes.

mglukhikh:
Merged manually. Congratulations 👍

PR #3209, opened 2020-03-23
[JVM_IR] Rebase test of string concatenation codegen.

madsager:
The JVM IR backend code seems saner to me. The string concatenation
lowering for JVM IR calls the stringPlus intrinsic if there are
only two arguments. That leads to a lot less code:

```
load string
load argument
box argument
call Intrinsics.stringPlus
```

instead of

```
allocate StringBuilder
call StringBuilder.<init>
load string
call StringBuilder.append
load argument
call StringBuilder.append
call StringBuilder.toString
```

This will lead to more boxing, but a lot smaller code. We still
use StringBuilders in JVM IR if there are more than two strings
being concatenated.


dnpetrov:
Ok, let's stick to that; probably we could also implement similar logic in old BE later.

dnpetrov:

dnpetrov:
Pushed manually

baf1f56859cf621210c521fec7d74300bddfe917

PR #3210, opened 2020-03-23
JVM: do not write trivial SMAPs to classes outside inline funs

pyos:
where trivial == those that map the file to itself.

Also, add synthetic line numbers to the SMAP. Otherwise, should the SMAP be present, IDEA will ignore the synthetic line numbers and thus hit KT-23064.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -83,11 +83,10 @@ internal val DeclarationDescriptorWithSource.psiElement: PsiElement?
 fun JvmBackendContext.getSourceMapper(declaration: IrClass): DefaultSourceMapper {
     val sourceManager = this.psiSourceManager
     val fileEntry = sourceManager.getFileEntry(declaration.fileParent)
-    check(fileEntry != null) { "No PSI file entry found for class: ${declaration.dump()}" }
```
max-kammerer: Could we check smap absent  for such synthetic things?

compiler/backend/src/org/jetbrains/kotlin/codegen/MemberCodegen.java
```diff
@@ -108,6 +109,8 @@ public MemberCodegen(
         this.propertyCodegen = new PropertyCodegen(context, v, functionCodegen, this);
         this.parentCodegen = parentCodegen;
         this.jvmAssertFieldGenerated = false;
+        // TODO fix the script runner (shouldn't require a SMAP) & don't write trivial SMAPs to script main classes.
+        this.alwaysWriteSourceMap = this instanceof ScriptCodegen;
```
max-kammerer: Could you clarify this a bit? What exception is failed?

pyos: See the example in the PR/commit description - if the SMAP for the file class does not exist, the stack trace will contain "Unknown Source" instead of actual line numbers. There's actually a test that checks the stack trace https://github.com/JetBrains/kotlin/blob/master/compiler/testData/integration/smoke/scriptException/script.kts but it happens to pass because `error("my error")` is an inline call which forces generation of a SMAP; replacing it with `throw ...` breaks the test.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
pyos: Maybe, but I'm not certain we should. Is it theoretically desirable to never inline functions into synthetic classes?

max-kammerer: OK

compiler/backend/src/org/jetbrains/kotlin/codegen/MemberCodegen.java
max-kammerer: `smaps` triggers `visitSource` call, and looks like it's not called in general case for scripts

max-kammerer: so debug information is missed

pyos: Oh, so it's not that it always tries to use the SMAP, it's just that ScriptCodegen.generateDeclaration should've called visitSource...that makes more sense, thanks.

max-kammerer:
@pyos Please look at filed tests
```
FileRankingTestGenerated.testTopLevelorg.jetbrains.kotlin.idea.debugger.test
KotlinSteppingTestGenerated$StepIntoOnly.testInlineOnlyorg.jetbrains.kotlin.idea.debugger.test
KotlinSteppingTestGenerated$StepIntoOnly.testSiSuspendFunorg.jetbrains.kotlin.idea.debugger.test
KotlinSteppingTestGenerated$StepOver.testKt14869org.jetbrains.kotlin.idea.debugger.test
```

pyos:
Wow, this is weird.

>KotlinSteppingTestGenerated$StepIntoOnly.testSiSuspendFun

This is another one of those "it happens to work because it does unrelated stuff that magically fixes it" tests, like the script one that I mentioned above. Try replacing line 20 of siSuspendFun.kt
```
suspend fun second(): Int = third()?.let { return it } ?: 0
```
with
```
suspend fun second(): Int = third() ?: 0
```
- you'd expect the test to behave the same, but nope, now that the SMAP is gone, "step into" stops at line 23 for some reason. And the SMAP is trivial, so this PR has the same effect.

>FileRankingTestGenerated.testTopLevel
>KotlinSteppingTestGenerated$StepIntoOnly.testInlineOnly
>KotlinSteppingTestGenerated$StepOver.testKt14869

These three appear to be related to https://github.com/JetBrains/kotlin/commit/f4337eede12e0f3709d88f6d22550928c2d3bac6: the trivial SMAP makes the IDEA plugin skip this "synthetic line number", so when this PR removes it,
```
inlineOnly.kt:7
```
becomes
```
inlineOnly.kt:7
inlineOnly.kt:65100
inlineOnly.kt:7
```
which actually looks more correct to me (isn't this the point of that commit in the first place?), but I don't know.

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
```diff
@@ -647,7 +647,19 @@ else if (isCompatibilityStubInDefaultImpls(functionDescriptor, context, jvmDefau
                     KtDeclarationWithBody declaration = ((ClosureGenerationStrategy) strategy).getDeclaration();
                     BindingContext bindingContext = typeMapper.getBindingContext();
                     if (declaration instanceof KtFunctionLiteral && isLambdaPassedToInlineOnly((KtFunction) declaration, bindingContext)) {
-                        lineNumber = LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER;
```
pyos: @yanex might be a weird question, but are you sure this ever worked? I tried debugging the following code in IDEA with the 1.4-M1 Kotlin plugin:
```
    forEach { print2("123") }
}
fun print2(s: String){
    println(s)
}
@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
@kotlin.internal.InlineOnly
inline fun forEach(s: () -> Unit) {
    for (i in 1..2) {
        s()
    }
}
```
and a breakpoint on the lambda is never triggered. Recompiling the file manually using the current master kotlinc does not help, but recompiling it with this patch does.

pyos:
@max-kammerer the tests should be fixed now, could you rerun the CI please?

max-kammerer:
@Pyos Could you add simple case for new behavior?
Because affected ones are too heavy.

@pyos @yanex 
Could we also check body of lambda and generate additional mapping only in case if it have any tatement on same line as inline call (so most of affected tests would be unaffected)?

compiler/testData/compileKotlinAgainstCustomBinaries/suspensionPointInMonitor/output.txt
```diff
@@ -4,7 +4,7 @@ compiler/testData/compileKotlinAgainstCustomBinaries/suspensionPointInMonitor/so
 compiler/testData/compileKotlinAgainstCustomBinaries/suspensionPointInMonitor/source.kt:14:13: error: a suspension point at SourceKt$test$2.invokeSuspend(source.kt:17) is inside a critical section
     builder {
             ^
-compiler/testData/compileKotlinAgainstCustomBinaries/suspensionPointInMonitor/source.kt:24:9: error: a suspension point at SourceKt$test$3$invokeSuspend$$inlined$withCrossinline$2.invokeSuspend(source.kt:62) is inside a critical section
+compiler/testData/compileKotlinAgainstCustomBinaries/suspensionPointInMonitor/source.kt:24:9: error: a suspension point at SourceKt$test$3$invokeSuspend$$inlined$withCrossinline$2.invokeSuspend(source.kt:63) is inside a critical section
```
max-kammerer: Is there any InlineOnly calls?

pyos: Yes, `synchronized` itself is `@InlineOnly`.

pyos:
>Could you add simple test for new behavior?

Done.

max-kammerer:
I've rechecked simple case in current master:
```
fun main() {    
    run  { "12"}
}
```
There is only one strata generate for such case. With patch there would be two but I'm not sure that second one would useful. In any case it would increase number of class files with invalid strata (addition *E #3213). So it's best to hide new logic under flag (that controls proper smap format) and remove old logic (as already done). Or don't emit such synthetic mappings into Debug strata

pyos:
I believe the reason there's a second stratum for this line number is actually KT-35006: when the lambda is inlined into `run`, the old (null) `callSiteMarker` in its line ranges is discarded and replaced with the one for `run` which points to line 2. Thus, fixing this bug will also remove the redundant stratum.

pyos:
>Could we also check body of lambda and generate additional mapping only in case if it have any statement on same line as inline call (so most of affected tests would be unaffected)?

This actually seems to work, yes. In fact, *all* tests except the new one have multi-line lambdas.

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
yanex: The last time I tried it, it worked. However, I don't have anything against your version. The main reasoning with the introduction of 65100 was to slightly reduce the generated lines count.

max-kammerer:
> This actually seems to work, yes. In fact, all tests except the new one have multi-line lambdas.
Cool!

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
```diff
@@ -647,7 +647,20 @@ else if (isCompatibilityStubInDefaultImpls(functionDescriptor, context, jvmDefau
...
+                        // We want the debugger to be able to break both on the inline call itself, plus on each
+                        // invocation of the inline lambda passed to it. For that to happen there needs to be at least
+                        // one different line number in between those breakpoints for the VM to emit a locatable event.
+                        // @InlineOnly functions, however, contain no line numbers, so if the lambda is single-line,
+                        // the entire call will "meld" into a single region. To break it up, we insert a different line
+                        // number that is remapped by the SMAP back to the lambda.
+                        Integer lambdaLine = CodegenUtil.getLineNumberForElement(declaration, false);
+                        Integer firstBodyLine = CodegenUtil.getLineNumberForElement(declaration.getBodyExpression(), false);
```
max-kammerer: Please note that 'declaration.getBodyExpression()' is Nullable

Could you also add test for next cases to test absent of additional mapping:
// IGNORE_BACKEND_MULTI_MODULE: JVM_IR
// FILE: 1.kt
package test

@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
@kotlin.internal.InlineOnly
inline fun <T, R> T.myLet(block: (T) -> R) = block(this)

// FILE: 2.kt
import test.*

fun box(): String {
    // in order to let the debugger break on it (KT-23064)
    return "O".myLet { param -> 
            it + "K" 
    }
}
and 
    return "O".myLet { fun foo(a: String) {} -> 
         a + "K" 
    }

compiler/testData/codegen/boxInline/smap/inlineOnly/stdlibInlineOnly.kt
```diff
@@ -0,0 +1,36 @@
...
+@kotlin.internal.InlineOnly
+inline fun <T, R> T.myLet(block: (T) -> R) = block(this)
+
+// FILE: 2.kt
+import test.*
+
+fun box(): String {
+    // should not have a line number for 1.kt:7; *should* have one for the lambda
+    // in order to let the debugger break on it (KT-23064)
+    return "O".myLet { it + "K" }
```
max-kammerer: Please test also 
     return "O".myLet(fun (it: String) : String {return it +  "K"})

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
pyos: Am I wrong in assuming that function literals should never have a null body?

max-kammerer: You are right, but there could be some suppressions. It best to keep some kind of robustness with explicit checks, e.g. see FunctionGenerationStrategy.FunctionDefault

pyos: Added both tests. Thanks for pointing out the anonymous `fun` case - looks like it was still broken because the checks around the code were too restrictive. Unfortunately, I can't figure a way to detect if a `KtNamedFunction` is single-line - do you have any pointers about this?

pyos: I just realized that checking for single-line functions is not correct anyway: even it they are not single-line, we still need to insert a different line number *in between two calls*. Consider this test:
```
    1.alsoTwice {
        println(it)
    }
}

@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
@kotlin.internal.InlineOnly
inline fun <T> T.alsoTwice(block: (T) -> Unit) {
    block(this)
    block(this)
}
```
If a breakpoint is placed on `println(it)` it will only be triggered once because there is no different line number in between two `println` invocations.

max-kammerer: As I remember `inlineOnly` functions in most cases are quite simple wrappers around java ones. But it's very intresting notice cc @yanex

pyos: Looks like this is actually broken even without `InlineOnly`, and adding a line number does not help.

compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
```diff
@@ -643,11 +643,24 @@ else if (isCompatibilityStubInDefaultImpls(functionDescriptor, context, jvmDefau
 
             if (context instanceof InlineLambdaContext) {
                 int lineNumber = -1;
-                if (strategy instanceof ClosureGenerationStrategy) {
-                    KtDeclarationWithBody declaration = ((ClosureGenerationStrategy) strategy).getDeclaration();
+                if (strategy instanceof FunctionGenerationStrategy.FunctionDefault) {
```
max-kammerer: Look a bug in `PsiSourceCompilerForInline` there should be ClosureGenerationStrategy if KtNamedFunction if isLambda flag passed

pyos: Probably doesn't matter since KtNamedFunction cannot have destructuring parameters though.

max-kammerer: It's mostly for consistency. Because non-inlined anonymous function Generated through ClosureGenerationStrategy

yanex:
>  Could we also check body of lambda and generate additional mapping only in case if it have any statement on same line as inline call (so most of affected tests would be unaffected)?

What do you mean? Is it something like:

```
run(
    { foo() }
)
```

or 

```
run { foo() }
```

? If the latter, then probably we have to generate an additional number as IDEA provides mixed (lambda & line) breakpoints.

max-kammerer:
@yanex I meant 
```run { 
    foo() 
}```
vs
```run { foo() }```

**Comment by yanex** [comment]
If the remaining number will lead to a line breakpoint, then I think it's OK.

***

By the way, `LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER` isn't the only case when synthetic line numbers might be required. I planned to add another one before coroutine suspends.

Normally, we can somehow delegate to the JVM-level stepping requests, however, with coroutines we are out of luck, as a coroutine can be suspended and resumed at any time on a different thread. So the debugger has to know it should place a one-time breakpoint and resume the process.

Unlike the case with @InlineOnly, a fixed marker is preferred, as a generated line number won't tell anything to the debugger. I suppose it might be a fixed line number (it is in my prototype), or some extra local variable (anything else?).

@max-kammerer, @pyos, what do you think?

**Comment by max-kammerer** [comment]
@yanex Could you describe a bit more detailed your case: where do you plan to insert a new synthetic linenumber and how it will help debugging coroutines?

Please also note additional changes around addition navigation (and maybe debugging) support for crossinline lambdas in #3248

**Comment by yanex** [comment]
@max-kammerer Now we add an additional line number right before the suspension `areturn`. My plan was to change this line number to a fixed one: https://github.com/JetBrains/kotlin/commit/da4d68aa0b31b82394ccb6bede03904430dc2574.
This will allow placing a stepping breakpoint whenever the debugger steps to this synthetic line number.

**Comment by max-kammerer** [comment]
@pyos Could you rebase on master?

> I believe the reason there's a second stratum for this line number is actually KT-35006: when the lambda is inlined into run, the old (null) callSiteMarker in its line ranges is discarded and replaced with the one for run which points to line 2. Thus, fixing this bug will also remove the redundant stratum.

Please also check that with KT-35006 fixed it's as expected

**Comment by max-kammerer** [comment]
> Even if compatibility with pre-release versions is not important

It's not important 

> That's the bootstrap compiler already, so the standard library contains references to that line number. BTW, is it considered OK that 1.3.71 will be unable to inline some code generated by 1.4?..

I'm not sure that it's good (I assume you speak about 65100 case  cc @udalov)

**Comment by pyos** [comment]
Rebased on master.

The new version maps line numbers to a nonexistent class in a fake file because mapping them to the lambda itself would always have to have some sort of hack to work around range folding. For example, consider this:

```
inline fun f() {
    1.let { println(it) }
}
fun g() = f()
```

For IDEA to break on `println(it)`, we need to do this in `f`:
```
LINENUMBER 2
// some bytecode from let
LINENUMBER X
ICONST_0
ISTORE ...
LINENUMBER 2
// println(it)
```
If `X` is mapped back to line 2, when inlining `f` `SourceMapCopier` will use the same line number for both 2 and X, removing the breakpoint again.

File: compiler/testData/codegen/boxInline/smap/inlineOnly/stdlibInlineOnlyOneLine.kt [hunk 8]
```diff
@@ -1,11 +1,30 @@
...
+Kotlin
+*S Kotlin
+*F
++ 1 2.kt
+_2Kt
++ 2 fake.kt
+kotlin/jvm/internal/FakeKt
+*L
+1#1,9:1
+65100#2:10
```
max-kammerer, line 29: cc @yanex Could this approach be used for coroutines?

Do we still need here to use 65100? "1" looks more intuitive and require less space

max-kammerer:
@pyos Looks cool!

compiler/testData/codegen/boxInline/smap/inlineOnly/stdlibInlineOnlyOneLine.kt
max-kammerer: I've implemented it in https://github.com/JetBrains/kotlin/commit/f5affff6a4add981c1ba87857c9780c1ee7a1efd

pyos: I did it a bit differently to avoid a magic constant 1 and to change existing instances of 65100 to 1 until the standard library is rebootstrapped: https://github.com/JetBrains/kotlin/pull/3210/commits/cf2ca80c4ba1c28c89374d5b7ed4354c3dd11c03

core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAbi.java
```diff
@@ -49,18 +49,12 @@
...
+        to @InlineOnly functions. All line information is stripped from the inlined implementation of such
+        functions, so if a lambda is on the same line with the call, the debugger can't stop on both 'line'
+        and 'lambda' breakpoints (KT-23064).
      */
-    public static final int LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER = 65100;
-
-    public static final List<Integer> SYNTHETIC_MARKER_LINE_NUMBERS = CollectionsKt.listOf(
-            LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER
-    );
+    public static final int LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER = 1;
```
max-kammerer, line 57: @pyos I would prefer to move out this constant from JvmAbi.
Client (debugger) could determine syntheticness by  ` kotlin/jvm/internal/FakeKt ` source

max-kammerer:
@pyos Pr also affects `https://github.com/JetBrains/kotlin/commit/d6f058e4c082216d3f5135d68220d8c074293265#diff-3449109ce8d2ef7a5aec2d1663917a50` test. Looks like debugger should skip new linenumber

compiler/testData/codegen/boxInline/smap/inlineOnly/stdlibInlineOnlyOneLine.kt
max-kammerer: @pyos Thank you, actually it's a bug in my implementation with missed remapping

core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAbi.java
max-kammerer: `inlineCodegenUtils.kt` looks as better place

pyos: But if later debugger needs to determine *which* line number it is (like for the step-over-suspend case mentioned above), wouldn't the constant need to be in JvmAbi?

pyos:
Yeah, looks like if the file for a class cannot be found, the debugger intentionally defaults to the current source file: https://github.com/JetBrains/kotlin/blob/257c4c560451c15d20edebc3926f6eba7d0f9f35/idea/jvm-debugger/jvm-debugger-core/src/org/jetbrains/kotlin/idea/debugger/KotlinPositionManager.kt#L104-L119
I don't know how to make it ignore synthetic line numbers though, since simply throwing NoDataException does not seem to do anything...

core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAbi.java
max-kammerer: Not necessery

max-kammerer: Now coruines related staff implemented via additional local variables. If it would be neccessery it could be rethinked later

max-kammerer:
@pyos Lets postpone debugger changes to another pr as we need current changes for M3 branching.

core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAbi.java
pyos: Ok, moved the constant to InlineOnlySmapSkipper.

pyos:
Added the new breakpoints to the test for now, since IMO the old data was not correct either: ideally, "step into" should stop on line 7 twice, highlighting the entire call the first time, and the lambda the second time.

max-kammerer:
Thanks you for PR! Your changes are more precise!

PR #3211, opened 2020-03-23
JVM: refactor some SMAP-related classes

pyos:
In particular, remove redundant copying from SMAPAndMethodNode (which also fails for empty methods).


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -253,20 +244,20 @@ open class DefaultSourceMapper(val sourceInfo: SourceInfo) : SourceMapper {
...
+    val sourceInfo: SourceInfo = run {
         assert(fileMappings.isNotEmpty()) { "File Mappings shouldn't be empty" }
+        val defaultFile = fileMappings.first()
+        val defaultRange = defaultFile.lineMappings.first()
+        SourceInfo(defaultFile.name, defaultFile.path, defaultRange.source + defaultRange.range - 1)
     }
 
-    val default: FileMapping
-        get() = fileMappings.first()
+    private val intervals = fileMappings.flatMap { it.lineMappings }.sortedWith(RangeMapping.Comparator)
```
max-kammerer: Looks like that sorting is redudant here cause smap data itself already sorted (as we writes it now). It had meaning after `createLineNumberSequence` because it searches intervals in `linenumbers` appeareance order (that could jump).
But it could be useful if we chnage smth in future (now sure how)

pyos: Are you sure that sorting is redundant here? It looks to me like it is useful when several files are interleaved, e.g.
```
// 1.kt
inline fun f() = 1
// 2.kt
inline fun g() = 2
// 3.kt
inline fun h() {
    f() // 3.kt:6 -> 1.kt:1
    g() // 3.kt:7 -> 2.kt:2
    f() // 3.kt:8 -> 1.kt:1
}
```
=> `fileMappings.flatMap { it.lineMappings }` returns something like [6 -> 1.kt:1, 8 -> 1.kt:1, 7 -> 2.kt:1] because mappings are grouped by file. So if we remove the sort here and then try to inline h() into a function in the same module, the binary search will fail.

max-kammerer: Oh, I've totatlly missed that they are grouped by file. Sorry, You are totally right!

max-kammerer: @pyos Could you add such test into CompileKotlinAgainstCustomBinariesTest (you could use `testInlineFunWithoutDebugInfo` as sample)

pyos: Does this really need a custom binary test? Looks like it could be a normal boxInline one (with 2.kt and 3.kt merged because multi-module tests require exactly 2 files).

max-kammerer: In this case all line entries would be sorted in smap

max-kammerer: Actually there should be also 4th file where h is inlined

pyos: Not sure why would they be.
```
// 1.kt
inline fun f() {}
// 2.kt
inline fun g() {}
inline fun h() {
    f() // N+1 -> 1.kt:1
    g() // N+2 -> 2.kt:1
    f() // N+3 -> 1.kt:1
}
fun box() {
    // N+4 -> 2.kt:3
    // N+5 -> N+1 -> 1.kt:1
    // N+6 -> 2.kt:4
    // N+7 -> N+2 -> 2.kt:1
    // N+8 -> 2.kt:5
    // N+9 -> N+3 -> 1.kt:1
    // N+10 -> 2.kt:6
    h()
    return "OK"
}
```
See https://github.com/JetBrains/kotlin/pull/3211/commits/ef982668eb2299c1413af58774880f22911f298c

max-kammerer: Cool!

max-kammerer:
Many thanks!

PR #3213, opened 2020-03-23
JVM: add a language feature to omit *E between SMAP strata

pyos:
Fixes #KT-37704

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
@@ -23,7 +23,8 @@ const val KOTLIN_DEBUG_STRATA_NAME = "KotlinDebug"
 class SMAPBuilder(
     val source: String,
     val path: String,
-    private val fileMappings: List<FileMapping>
+    private val fileMappings: List<FileMapping>,
+    private val backwardsCompatibleSyntax: Boolean
 ) {
     private val header = "SMAP\n$source\nKotlin"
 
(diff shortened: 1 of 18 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -39,25 +40,25 @@ class SMAPBuilder(
...
         val defaultStrata = generateDefaultStrata(realMappings)
         val debugStrata = generateDebugStrata(realMappings)
-
-        return "$header\n$defaultStrata$debugStrata"
+        if (backwardsCompatibleSyntax && defaultStrata.isNotEmpty() && debugStrata.isNotEmpty()) {
+            // Old versions of kotlinc might fail if there is no END between defaultStrata and debugStrata.
+            // This is not actually correct syntax according to JSR-045.
+            return "$header\n$defaultStrata$END\n$debugStrata$END\n"
+        }
+        return "$header\n$defaultStrata$debugStrata$END\n"
```
max-kammerer: @udalov Looks like it's also best to add message from future for classes with strata in new format

max-kammerer:
@pyos Please add box inline test with new flag

max-kammerer:
And with flag disabled (to test logic after switch default behaviour)

pyos:
done

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
udalov: I suppose we should make the new language feature's kind `UNSTABLE_FEATURE`, so that manually enabling it would produce pre-release binaries. But otherwise we need not do anything, since 1.5 binaries will only be readable by 1.4+.

compiler/testData/codegen/boxInline/smap/smapWithNewSyntax.kt
```diff
@@ -0,0 +1,35 @@
+// !LANGUAGE: +CorrectSourceMappingSyntax
+// FILE: 1.kt
+package test
+
+public inline fun inlineFun(): String {
```
max-kammerer: Please make some inline calls inside 1.kt file to generate smap for it (it will allow to check smap parsing during compilation against binaries)

Same for second test

pyos: ok

max-kammerer:
Many thanks!

max-kammerer:
@pyos could you also support new scheme in debugger utils: 
org/jetbrains/kotlin/idea/debugger/smapUtil.kt:70

pyos:
Already did it as part of #3248: with that PR, SMAPParser handles call site markers by itself, so the additional logic in smapUtil.kt is simply not necessary.

PR #3214, opened 2020-03-23
JVM: check inline call site line numbers in stepping tests

pyos:
Note: I've modified the test data to what *should* be generated, not the output right now: there are still some issues, such as KT-37704 and KT-35006.

### DIFF
--- a/compiler/testData/debug/stepping/inlineCallableReference.kt
+++ b/compiler/testData/debug/stepping/inlineCallableReference.kt
@@ -13,8 +13,11 @@ inline fun f(block: () -> Unit) {
 // LINENUMBERS
 // test.kt:3
 // test.kt:4
-// test.kt:10
+// test.kt:10 @ test.kt:4
 // test.kt:5
 // test.kt:6
-// test.kt:11
+// test.kt:11 @ test.kt:4
 // test.kt:7
+
+// line 11 has incorrect call site (:5)
+// IGNORE_BACKEND: JVM, JVM_IR
--- a/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/debugInformation/AbstractDebugTest.kt
+++ b/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/debugInformation/AbstractDebugTest.kt
@@ -15,9 +15,12 @@ import com.sun.jdi.request.EventRequest.SUSPEND_ALL
 import com.sun.jdi.request.StepRequest
 import com.sun.tools.jdi.SocketAttachingConnector
 import org.jetbrains.kotlin.backend.common.output.SimpleOutputFileCollection
+import org.jetbrains.kotlin.checkers.forceEnableFeatures
 import org.jetbrains.kotlin.cli.common.output.writeAllTo
 import org.jetbrains.kotlin.codegen.*
 import org.jetbrains.kotlin.codegen.forTestCompile.ForTestCompileRuntime
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.LanguageFeature
 import org.jetbrains.kotlin.test.ConfigurationKind
 import org.jetbrains.kotlin.test.KotlinTestUtils
 import org.jetbrains.kotlin.test.clientserver.TestProcessServer
@@ -118,6 +121,11 @@ abstract class AbstractDebugTest : CodegenTestCase() {
     @After
     public override fun tearDown() {
         super.tearDown()
+    }
+
+    override fun updateConfiguration(configuration: CompilerConfiguration) {
+        super.updateConfiguration(configuration)
+        configuration.forceEnableFeatures(LanguageFeature.CorrectSourceMappingSyntax)
     }
 
     internal fun invokeBoxInSeparateProcess(classLoader: URLClassLoader, aClass: Class<*>, port: Int): String {
--- a/compiler/testData/debug/stepping/for.kt
+++ b/compiler/testData/debug/stepping/for.kt
@@ -10,12 +10,12 @@ inline fun foo(n: Int) {}
 // LINENUMBERS
 // test.kt:3
 // test.kt:4
-// test.kt:8
+// test.kt:8 @ test.kt:4
 // test.kt:3
 // test.kt:4
-// test.kt:8
+// test.kt:8 @ test.kt:4
 // test.kt:3
 // test.kt:4
-// test.kt:8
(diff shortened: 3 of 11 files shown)


pyos:
Rebased on top of #3213. Note that some tests are still muted.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/debugInformation/AbstractDebugTest.kt
```diff
@@ -120,6 +123,11 @@ abstract class AbstractDebugTest : CodegenTestCase() {
         super.tearDown()
     }
 
+    override fun updateConfiguration(configuration: CompilerConfiguration) {
+        super.updateConfiguration(configuration)
+        configuration.forceEnableFeatures(LanguageFeature.CorrectSourceMappingSyntax)
```
max-kammerer: Why not enable this feauture in test data explicitly?

pyos: I thought it would be weird to require tests to opt into a bug fix in order to get the complete output.

max-kammerer: There are not so much test with inline. Otherwise we a getting additional complexity in test infrastructure, `forceEnableFeatures` and 'LanguageFeatureOverride' also looks a bit `weird`

pyos: On a related note, if I'm reading this code https://github.com/JetBrains/kotlin/blob/master/compiler/tests-common/tests/org/jetbrains/kotlin/codegen/CodegenTestCase.java#L199-L212 correctly, tests tagged as `COMMON_COROUTINES_TEST` ignore the `!LANGUAGE` directive when running in experimental coroutine mode.

max-kammerer: oh...it's fail // cc @ilmirus

pyos: LanguageFeatureOverride would fix this, just saying...

max-kammerer: @pyos thank you for point for this problem, let's keep LanguageFeatureOverride. Thank you!

ilmirus: (－‸ლ)

compiler/testData/debug/stepping/inlineCallableReference.kt
```diff
@@ -13,8 +13,11 @@ inline fun f(block: () -> Unit) {
...
 // test.kt:4
-// test.kt:10
+// test.kt:10 @ test.kt:4
 // test.kt:5
 // test.kt:6
-// test.kt:11
+// test.kt:11 @ test.kt:4
 // test.kt:7
+
+// line 11 has incorrect call site (:5)
```
pyos: @max-kammerer small problem here. In this code, for example
```
fun main() {
    f()
}

inline fun f() {
    println("...")
    throw RuntimeError("...")
}
```
the SMAP contains the following entry in the KotlinDebug stratum:
```
2#1,2:9
```
The problem is that according to JSR-045, this actually means "map line 9 to 2 and line 10 to 3", which is why these tests fail. However, if you paste this code into IDEA and click "navigate to inline function call site", it will map line 10 to 2, apparently interpreting this line to mean "map *both* 9 and 10 to 2" because it's in the call site stratum.

Unfortunately, it seems like the only way to emit JSR-045-correct data here is to map each line separately:
```
2#1:9
2#1:10
```
because something like
```
2#1,2:9,0
```
actually means "map line 9 to range [2,2) and line 10 to range [2,2)", i.e. it actually maps each line to precisely nothing.

Obviously, this will increase the size of SMAPs when inlining large functions. However, I assume that this increase in size will not be noticeable in jars because "2#1:" is repeated and thus deflate should be able to compress it. What do you think?

pyos: JDI doesn't know what `2#1:9,2` means either.

So much for following specifications...

max-kammerer: > which is the correct way to say "map lines 9 and 10 to 2" is not supported by IDEA 

Maybe it's lack in Kotlin Debugger support and could be simple fixed // cc  @yandex

pyos: You mean @yanex? :)

max-kammerer: > JDI doesn't know what 2#1:9,2 means either and simply ignores the ,2 part.

it's quite strange, it's definetly should be a issue. Maybe it's fixed in later versions of JDK

pyos: It does support this syntax after all. The problem is that so far I don't see an API to correctly retrieve the line number from KotlinDebug:
1. like the comment in this PR says, if an unmapped line number does not appear in the stratum, JDI will return "undefined garbage", or more specifically it will attempt to guess the closest line number by bytecode position instead of just saying that there's no line number: https://github.com/AdoptOpenJDK/openjdk-jdk/blob/master/src/jdk.jdi/share/classes/com/sun/tools/jdi/LocationImpl.java#L159
2. the `allLineLocations` hack I'm using to work around this intentionally deduplicates the line numbers: https://github.com/AdoptOpenJDK/openjdk-jdk/blob/master/src/jdk.jdi/share/classes/com/sun/tools/jdi/ConcreteMethodImpl.java#L308 but the equality comparison for LineStratum is "same SMAP stratum, same mapping, and maps to the same result" - so lines 9 and 10 that are both remapped by the same `2#1:9,2` to line 2 are considered equal, thus only 9 is returned, and the test no longer tries to print the call site for line 10.

max-kammerer:
Is this PR awaits other ones or not?

pyos:
No, but I don't think there's a way to dump the correct line numbers using JDI, and manually parsing the SMAP is probably pointless. Using exception filter tests seems like a better approach for now.

PR #3218, opened 2020-03-24
KT-37502 False positive "redundant lambda arrow" with inline generic function with reified type in object and anonymous parameter name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37502


darthorimar:
LGTM :+1: thank you

PR #3219, opened 2020-03-24
KT-34770 JS. Configurable properties

turansky:
Both backends supported


bashor:
@turansky could you please add some tests?

turansky:
Is there existed related tests for such cases?

bashor:
You can check it using `Object.getOwnPropertyDescriptor`, to get prperties' names you can use `Object.getOwnPropertyNames` or `Object.keys`.
The closest example what I can find is [js/js.translator/testData/box/propertyAccess/enumerable.kt](https://github.com/JetBrains/kotlin/blob/master/js/js.translator/testData/box/propertyAccess/enumerable.kt)

turansky:
Is there possibility to run tests only for `js` or `js/js.translator` subprojects?

turansky:
> @turansky could you please add some tests?

Done

bashor:
I've rebased and pushed changed manually. 

PR #3220, opened 2020-03-24
FIR: use dispatch receiver of the enclosing function if any.

jsjeon:
Main motivation was to dig into the failure of `classFieldInsideLambda`:

```
class My {
    val my: String = "O"
        get() = { field }() + "K"
}

fun box() = My().my
```

`field` symbol inside a lambda is translated to `GET_VALUE` IR with `this` pointer. In `psi2ir` path, that dispatch receiver refers to `My.<get-my>` whereas in `fir` path, it refers to `My`. Anonymous function doesn't have a dispatch receiver, and thus the current logic of `lastDispatchReceiverParameter` directly returns the `this` receiver of the containing class. Rather, it need to check if the dispatch receiver of the enclosing function is available.


jsjeon:
FYI, build/test results are here: https://teamcity.jetbrains.com/viewLog.html?buildId=2827400&tab=buildResultsDiv&buildTypeId=Kotlin_KotlinForGoogle_AggregateBranch

mglukhikh:
I merged it but I think this function can be made even better...

mglukhikh:
Thank you!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConversionScope.kt
```diff
@@ -77,15 +77,17 @@ class Fir2IrConversionScope {
...
-            // Use the dispatch receiver of the containing class
-            lastClassSymbol?.owner?.thisReceiver
-        }
+        var dispatchReceiver = functionStack.lastOrNull()?.dispatchReceiverParameter
+        // Use the dispatch receiver of the containing function
+        dispatchReceiver?.let { return it }
+
+        // Use the dispatch receiver of the enclosing function if any
+        dispatchReceiver = functionStack.elementAtOrNull(functionStack.size - 2)?.dispatchReceiverParameter
+        dispatchReceiver?.let { return it }
```
punzki: What if there is more than one nested anonymous function, like:
```
class My {
    val my: String = "O"
        get() = { { field }() }() + "K"
}
```
I'm just wondering if you should keep going up the stack to find a receiver (i.e., with a while loop).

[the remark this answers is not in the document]
jsjeon: Yeah, I initially tried like, `functionStack.asReversed().firstNotNullResult { it.dispatchReceiverParameter }`, and the results were same. As per Mark's suggestion below, I'll revisit that approach with an additional nested lambda.

jsjeon: Thanks for the suggestion! I'll try that one as well, and see if we actually need to look up function stack all the way up.

mglukhikh:
I've rewritten it this way

```
    fun lastDispatchReceiverParameter(): IrValueParameter? {
        for (function in functionStack.asReversed()) {
            // Use the dispatch receiver of the containing/enclosing function
            function.dispatchReceiverParameter?.let { return it }
        }

        // Use the dispatch receiver of the containing class
        val lastClass = classStack.lastOrNull()
        return lastClass?.thisReceiver
    }
```

but yet is does not change any test behavior. May be we should try writing a new one, e.g. like @punzki  suggested above.

jsjeon:
I just tested what @punzki suggested (plus one more nest as suggested in the internal chat), and it behaves differently. Yes, we need to iterate all functions' dispatch receiver in the stack. I'm running some necessary tests to make sure, and will send out another PR soon.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConversionScope.kt
jsjeon: Indeed, having more nested lambdas failed on FIR, but succeed on IR. Thanks for the suggestion again!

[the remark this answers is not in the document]
jsjeon: Yes, we should. I'll prepare a follow-up PR.

PR #3221, opened 2020-03-25
FIR: consider all functions in scope when computing dispatch receiver parameter.

jsjeon:
This is a follow-up of https://github.com/JetBrains/kotlin/pull/3220. Special thanks to @punzki


mglukhikh:
👍

PR #3224, opened 2020-03-25
Backport of Add IrProviders to IrPluginContext to 1.3.70

alexander-gorshenev:


romanart:
Seems like it is not longer needed so it could be closed.

PR #3225, opened 2020-03-25
KT-24750 Blank line after class header formatter should use user setting

kozaxinan:
Fixes; https://youtrack.jetbrains.com/issue/KT-24750

When `Minimum Blank Line` -> `After class header` formatter setting changed, code formatter needs to respect new setting.

| Before | Minimum 1 | Minimum after header 5 and Max line 2 |
| - | - | - |
| Screenshot 2020-03-25 at 15 10 07 | Screenshot 2020-03-25 at 15 10 12 | Screenshot 2020-03-25 at 15 10 31 |

### DIFF
--- a/idea/formatter/src/org/jetbrains/kotlin/idea/formatter/kotlinSpacingRules.kt
+++ b/idea/formatter/src/org/jetbrains/kotlin/idea/formatter/kotlinSpacingRules.kt
@@ -153,13 +153,16 @@ fun createSpacingBuilder(settings: CodeStyleSettings, builderUtil: KotlinSpacing
                     null
                 } else {
                     val minLineFeeds = if (right.requireNode().elementType == FUN || right.requireNode().elementType == PROPERTY)
-                        1
+                        kotlinCommonSettings.BLANK_LINES_AFTER_CLASS_HEADER + 1
                     else
                         0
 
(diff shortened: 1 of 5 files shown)


idea/formatter/src/org/jetbrains/kotlin/idea/formatter/kotlinSpacingRules.kt
```diff
@@ -153,13 +153,16 @@ fun createSpacingBuilder(settings: CodeStyleSettings, builderUtil: KotlinSpacing
...
                     else
                         0
 
                     builderUtil.createLineFeedDependentSpacing(
-                        1, 1, minLineFeeds,
-                        settings.KEEP_LINE_BREAKS, settings.KEEP_BLANK_LINES_IN_DECLARATIONS,
+                        1,
+                        1,
+                        minLineFeeds,
+                        commonCodeStyleSettings.KEEP_LINE_BREAKS,
```
kozaxinan: Is this right call to update this part? settings.* fields are deprecated.

dimonchik0036: Yes, thanks.

PR #3226, opened 2020-03-25
Produce deterministic klib files.

mpetrov:
Build systems like bazel rely on the output of a compilation being
deterministic. File checksums determine when an artifact is safe to
cache across builds. One can produce a klib deterministically by using a
fixed timestamp during klib compression.

Before this change:

```
$ kotlinc-native  -produce library -output /tmp/first/color.klib Color.kt
$ kotlinc-native  -produce library -output /tmp/second/color.klib Color.kt
$ shasum  /tmp/first/color.klib /tmp/second/color.klib
bc3f73678ff025cfbec9009f9a851a8ca74e1037  /tmp/first/color.klib
65aa37886fbd53285f2e449a4dab6a2ad02732e6  /tmp/second/color.klib
```

After this change:

```
$ kotlinc-native  -produce library -output /tmp/first/color.klib Color.kt
$ kotlinc-native  -produce library -output /tmp/second/color.klib Color.kt
$ shasum  /tmp/first/color.klib /tmp/second/color.klib
fcba304493916ae34d372188991f87b60a113cf3  /tmp/first/color.klib
fcba304493916ae34d372188991f87b60a113cf3  /tmp/second/color.klib
```

### DIFF
--- a/compiler/util-io/src/org/jetbrains/kotlin/konan/file/File.kt
+++ b/compiler/util-io/src/org/jetbrains/kotlin/konan/file/File.kt
@@ -14,7 +14,9 @@ import java.lang.Exception
 import java.nio.MappedByteBuffer
 import java.nio.channels.FileChannel
 import java.nio.file.*
+import java.nio.file.attribute.BasicFileAttributeView
 import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.attribute.FileTime
 
 data class File constructor(internal val javaPath: Path) {
     constructor(parent: Path, child: String): this(parent.resolve(child))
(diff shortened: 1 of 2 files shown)


mpetrov:
@alexander-gorshenev could you take a look?

Deterministic checksums are important for supporting our build caches.

romanart:
Thank you a lot for this fix and very sorry for long answer. Timestamp also affect out incremental compilation infrastructure. 

From compiler perspective everything seems to be OK. 

@ddolovov Could you please review from IDE side. I would like to merge it.

ddolovov:
I do not expect any issues from IDE side. LGTM.

PR #3227, opened 2020-03-25
KT-37776 JS. Remove @JsModule/@JsQualifier non-external declaration limitation

turansky:
Allow non-external declarations even if file is marked with `@JsModule` or `@JsQualifier` annotation

```Kotlin
@file:JsModule("yyy")

external class Node {}

// valid now
typealias YNode = Node
```


bashor:
Before we have decided to do it such a way to avoid confusion due to the fact that it actually affects only external declarations.
We need to reconsider it before applying any PR.

Thank you for your effort!

PR #3228, opened 2020-03-26
[FIR:] fix test module set up in GenerationUtils::compileFilesUsingFrontendIR

juan-chen:
Added creation and initialization of the test module and its dependencies.

When looking at the emptyProgression test, I found that the "getProgressionLastElement" method was not found, which should be resolved from the dependencies.


mglukhikh:
Great, thank you! However, most test files you have changed are auto-generated and should not be edited directly. See comments at the beginning of files. I will add commit that modifies auto-generation.

mglukhikh:
Merged manually 👍  See also https://github.com/JetBrains/kotlin/commit/547bb64b18eeba389cc7fddcf05b6cae8fc8de16 (commit that fixes auto-generation).

PR #3229, opened 2020-03-26
KT-37749 "Convert to anonymous object" intention is suggested for Java SAM conversion, but not for Kotlin

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37749

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/SamConversionToAnonymousObjectIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/SamConversionToAnonymousObjectIntention.kt
@@ -17,7 +17,6 @@ import org.jetbrains.kotlin.idea.caches.resolve.resolveToCall
 import org.jetbrains.kotlin.idea.core.replaced
 import org.jetbrains.kotlin.idea.search.usagesSearch.descriptor
 import org.jetbrains.kotlin.lexer.KtTokens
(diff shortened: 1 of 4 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/SamConversionToAnonymousObjectIntention.kt
```diff
@@ -39,12 +38,12 @@ class SamConversionToAnonymousObjectIntention : SelfTargetingRangeIntention<KtCa
         val callee = element.calleeExpression ?: return null
         val lambda = getLambdaExpression(element) ?: return null
         val functionLiteral = lambda.functionLiteral
-        val descriptor = (functionLiteral.descriptor as? FunctionDescriptor) ?: return null
         val bindingContext = functionLiteral.analyze()
         val sam = element.getSingleAbstractMethod(bindingContext) ?: return null
 
         val samValueParameters = sam.valueParameters
         val samValueParameterSize = samValueParameters.size
+        val descriptor = bindingContext[BindingContext.DECLARATION_TO_DESCRIPTOR, functionLiteral] as? FunctionDescriptor ?: return null
```
fedochet: You are reusing `bindingContext` here, but in the `applyTo` you are still using `descriptor` extension. What is the reason for that

fedochet:
I've squashed your commits into 6a1923a4777be955d0b7cb92aa2a740d1065d2e8 and pushed to the master

Thanks!

PR #3230, opened 2020-03-26
JVM: preserve call site markers when inlining lambdas

pyos:
Fixes #KT-35006

Also, reuse more mappings (apparently this resolves a TODO in some SMAP tests)


pyos:
@max-kammerer since trying to get JDI to output call sites correctly was a failure, I gave up and added a simple SMAP test with some comments.

max-kammerer:
I've noticed that smap data only optimized in facade classes. Could you add (if we don't have yet) test with explicit classes that call inline functions from another class or package in same file

pyos:
Do you mean something like this, to test the `isLastMapped` check and this PR's equivalent `mapping.maxDest == currentIndex`? https://github.com/JetBrains/kotlin/pull/3230/files#diff-c3d17038d014cb7effbe284f813fad91 Yes, even though theoretically a range for `D` could be preallocated if we completely scan the method node for line numbers before creating the actual mapping, this is currently not done so this case is not very optimal.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -267,72 +214,36 @@ class SMAP(val fileMappings: List<FileMapping>) {
...
-        val dest: Int
-        val rangeMapping: RangeMapping
-        if (rangeMappings.isNotEmpty() && isLastMapped && couldFoldInRange(lastMappedWithNewIndex, source)) {
-            rangeMapping = rangeMappings.last()
-            rangeMapping.range += source - lastMappedWithNewIndex
-            dest = rangeMapping.mapSourceToDest(source)
+    fun mapNewLineNumber(source: Int, currentIndex: Int, callSiteMarker: CallSiteMarker?): Int {
+        var mapping = lineMappings.lastOrNull()
+        if (mapping != null && mapping.callSiteMarker == callSiteMarker &&
+            (source - mapping.source) in 0 until mapping.range + (if (mapping.maxDest == currentIndex) 10 else 0)
```
max-kammerer: In `couldFoldInRange` there was <= 10

max-kammerer: So `until` should be `..`
But I'm a bit confused by `mapping.maxDest == currentIndex '  check

pyos: Yes, but `lastMappedWithNewIndex` was `mapping.source + mapping.range - 1`.

pyos: The `mapping.maxDest == currentIndex` is equivalent to checking that this is the `FileMapping` that `mapNewLineNumber` was last called on. If it was not, then we cannot *extend* the range (because we've already used the next line number for some other file) but we can still reuse it (thus `0 until mapping.range` is OK -- but not `0..mapping.range`, because `mapping.source + mapping.range` is out of bounds for this mapping).

max-kammerer: Thanks! I've cleared my doubts

compiler/testData/codegen/boxInline/smap/rangeFoldingInClass.kt
```diff
@@ -0,0 +1,100 @@
...
+        C.h()
+        B.g()
+    }
+}
+
+// FILE: 2.kt
+import test.*
+
+object X {
+    // Unlike `rangeFolding.kt`, the calls in `D.together` refer to different
```
max-kammerer: @pyos Yes, it's expected.
There additional difference with spec: spec expects source path but compiler write class fqn. 
Not sure but maybe this behaviour should be revised

pyos: Yeah, I was a bit confused by this part. Does IDEA use the class name for anything?

By the way, note that `KotlinDebug` could still be a single range, `generateDebugStrata` just has to sort ranges by `dest` and merge them separately from `generateDefaultStrata`.

max-kammerer: > Does IDEA use the class name for anything

No, it's Kotlin related artifact. Actually  there was a problem with determining source paths (not sure maybe smth was changed and it's possible to make some workaround to this lack now. cc @udalov)

> By the way, note that KotlinDebug could still be a single range, generateDebugStrata just has to sort ranges by dest and merge them separately from generateDefaultStrata.

Looks great. But best to postpone for implement it in another pr

It looks like a lack of debug/stepping test. Am right that it would be covered by #3214 ? It's best to have one or two related test in this pr too (if they are not present yet in master)

pyos: What kind of stepping test do you mean? I don't think there's currently any kind of test at all other than the SMAP dumping ones that can possibly check KotlinDebug.

max-kammerer: Doesn't matter, I've missed that kt35006.kt is added in PR

max-kammerer: @pyos Could you add navigation test (KotlinExceptionFilterTestGenerated) for KT-35006
@NataliaUkhorskaya Could help with this type of tests

pyos: Doesn't look like there are currently IDEA tests for this either (navigation tests are for clicking around in the source code), but I've added a NAVIGATE_TO_CALL_SITE directive to the exception filter tests (inlineFunCallSiteInInlineLambda.kt is KT-35006).

max-kammerer:
Thank you for this nice refactoring!

PR #3231, opened 2020-03-26
JVM IR: Don't use IrStringConcatenation for ordinary toString calls

sfs:
We can only use `IrStrinConcatentation` to represent calls to `Any?.toString` and `toString` calls on primitive types. Otherwise, `x.toString()` and `"$x"` are observably different when `x` is a non-null type with `null` value (e.g., an `@NotNull` value coming from Java).

Without this PR we are always producing `String.valueOf` calls, even for explicit calls to `toString`.


punzki:
Since you mentioned it in the PR, could you add a test where you call `toString()` on a `@NonNull` value from Java that is actually `null`? (Should throw NPE)

sfs:
Good idea, Mark, I've added a test for it.

dnpetrov:

dnpetrov:
Pushed manually.

58685be4e2fcae3e8463e21462b125146c544e6b

PR #3234, opened 2020-03-27
FIR: use dispatch receiver for field expression without selector.

jsjeon:
Motivation was to investigate the failure of `classFieldInsideNested`:

```
abstract class Your {
    abstract val your: String
    fun foo() = your
}

class My {
    val my: String = "O"
        get() = object : Your() {
            override val your = field
        }.foo() + "K"
}

fun box() = My().my
```

where `field` in overridden `your` initializer refers to the receiver of <get-your> in the anonymous object. On the other hand, the expected resolution is <get-my>.

When setting the receiver for field expression _without_ selector, we should consult the functions in the scope. The caveat is we should make sure to pick a function that has access to the field of interest.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallGenerator.kt
```diff
@@ -328,7 +331,18 @@ internal class CallGenerator(
             is IrFieldExpressionBase -> {
```
jsjeon: Hm, github doesn't allow me to leave comments outside of changes. :\

Anyhow, I saw that applying the same idea to `IrCallImpl` case above, in particular, `dispatchReceiver` for call expression without selector could also unmute 17 cases, but broke many others. :p It may require more careful accessibility check I guess. I'll dig into that as well, if this approach looks good.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallGenerator.kt
```diff
@@ -328,7 +331,18 @@ internal class CallGenerator(
...
                 val ownerField = symbol.owner
                 if (!ownerField.isStatic) {
-                    receiver = qualifiedAccess.findIrDispatchReceiver()
+                    // Qualified access with a selector: use the bound dispatch receiver.
+                    if (qualifiedAccess.explicitReceiver != null) {
+                        receiver = qualifiedAccess.findIrDispatchReceiver()
+                    } else {
+                        // Otherwise, find the dispatch receiver from the scope with a condition:
+                        //   the function in the scope has access to the field of interest.
+                        conversionScope.lastDispatchReceiverParameter { function ->
```
mglukhikh: IMHO this logic looks a bit hacky. I will think about alternative variants and write you.

mglukhikh:
I found the problem in `Fir2IrVisitor` code. Now we have:

```
    override fun visitThisReceiverExpression(thisReceiverExpression: FirThisReceiverExpression, data: Any?): IrElement {
        val calleeReference = thisReceiverExpression.calleeReference
        if (calleeReference.labelName == null && calleeReference.boundSymbol is FirRegularClassSymbol) {
            // Object case ...
            // ...
        }
    }
```

but should be

```
    override fun visitThisReceiverExpression(thisReceiverExpression: FirThisReceiverExpression, data: Any?): IrElement {
        val calleeReference = thisReceiverExpression.calleeReference
        if (calleeReference.labelName == null && calleeReference.boundSymbol is FirClassSymbol) {
            // Object case ...
            // ...
        }
    }
```

That's why anonymous object isn't handled correctly here.

mglukhikh:
So in Friday evening I have to drop this uncompleted, my current investigations are here: https://github.com/JetBrains/kotlin/commits/rr/mg-jeon/fir2ir-this (it fixes 15-20 BB tests but also breaks a bit). At this moment my hypotheses are:

- `findIrDispatchReceiver` is correctly called in this case but produces wrong results (due to several bugs)
- `visitThisReceiverExpression` has at least two problems to be fixed (see two commits in the branch)
- also, I noticed that FIR resolve sometimes produces incorrect `this` receiver for synthetic `field` variable; e.g. in `classFileInsideLocalInSetter.kt` it has `this@Local` receiver, but should have `this@My` receiver. It's FIR resolve problem and it should be fixed there, not in FIR2IR

You may continue from this point if you like. I will return to these questions on Monday (hope so).

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -73,7 +72,7 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
         when {
             local -> CallableId(name)
             context.className == FqName.ROOT -> CallableId(context.packageFqName, name)
-            context.className.shortName() === ANONYMOUS_OBJECT_NAME -> CallableId(FqName.ROOT, FqName("anonymous"), name)
+            context.className.shortName() === ANONYMOUS_OBJECT_NAME -> CallableId(ANONYMOUS_CLASS_ID, name)
             else -> CallableId(context.packageFqName, context.className, name)
```
jsjeon: I guess this default case could be potentially `CallableId(context.currentCallId, name)` but broke a couple cases. :\

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/TowerLevels.kt
```diff
@@ -190,8 +191,18 @@ class ScopeTowerLevel(
                 empty = false
                 if (candidate.hasConsistentReceivers(extensionReceiver)) {
                     val dispatchReceiverValue =
-                        if (candidate !is FirBackingFieldSymbol) null
-                        else bodyResolveComponents.implicitReceiverStack.lastDispatchReceiver()
+                        when {
+                            candidate !is FirBackingFieldSymbol -> null
+                            candidate.callableId.classId != null -> {
```
jsjeon: So, at this point, we know `candidate` is a backing field; `candidate`'s holder information is available. Since receivers on tack are "implicit", I guess we can do further filtering. In this case, since we know the exact holder (of the property), it seems to make sense to me that we should choose the receiver on the stack whose type matches.

mglukhikh: I think this code is Ok.

mglukhikh:
> This may look hacky again, but I still believe resolution should check accessibility somehow. 

I think code in FIR resolve is much less hacky. It's a task of FIR resolve to choose a correct dispatch receiver for everything.

I will look at all three commits again. There are a couple of tests which are still broken, I'd like to understand the reason (not sure I will finish today). Thank you for continuing our work :)

mglukhikh:
In `selfcreate.kt` here

```
class B () {}
open class A(val b : B) {
    fun a(): A = object: A(b /* here */) {}
}
```

we also have incorrect receiver after FIR resolve, `this@anonymous` instead of `this@A`.

mglukhikh:
I've accepted this branch. Thanks again 👍

jsjeon:
Thank you for reviews and continuing the work. 👍

PR #3235, opened 2020-03-27
[JVM_IR] Use iinc for incrementing Int variables.

madsager:
Fix line number generation for assignments where the right-hand
side of the assignment is not on the same line.

Fix line number generation for intrinsics functions where the
function is not on the same line as the last argument.

Be careful to not break stepping behavior with the iinc
optimizations.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -149,13 +149,15 @@ class ExpressionCodegen(
 
     private fun markNewLabel() = Label().apply { mv.visitLabel(this) }
 
+    private fun getLineNumberForOffset(offset: Int): Int = fileEntry?.getLineNumber(offset)?.plus(1) ?: -1
+
     private fun IrElement.markLineNumber(startOffset: Boolean) {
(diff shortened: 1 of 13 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -591,9 +595,73 @@ class ExpressionCodegen(
...
+                expression.markLineNumber(startOffset = true)
+                mv.iinc(findLocalIndex(expression.symbol), if (isMinus) -delta else delta)
+                return true
+            }
+        }
+        return false
+    }
+
+    private fun hasSameLineNumber(element0: IrElement, element1: IrElement): Boolean
+        = getLineNumberForOffset(element0.startOffset) == getLineNumberForOffset(element1.startOffset)
```
dnpetrov: Minor, style: put `=` on the same line as function definition ( https://kotlinlang.org/docs/reference/coding-conventions.html#expression-body-formatting )

madsager: Thanks. Done.

dnpetrov:
Ok, looks good to me.

dnpetrov:
Pushed manually.

5570a5fe74b12dcd74f99447f5511534ca6963f3

PR #3237, opened 2020-03-28
README Update IDEA plugin dist path

turansky:


ilya-g:
Thanks for the correction.

PR #3240, opened 2020-03-30
KT-37842 "Convert to anonymous function" creates broken code with suspend functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37842


yanex:
Review: OK.

yanex:
Merged into `master`. Thanks!

PR #3243, opened 2020-03-31
FIR: set proper visibility of backing fields.

jsjeon:


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
```diff
@@ -168,6 +168,13 @@ internal class ClassMemberGenerator(
         return irProperty
     }
 
+    private val FirProperty.fieldVisibility: Visibility
+        get() = when {
+            isLateInit -> setter?.visibility ?: status.visibility
+            isConst -> status.visibility
+            else -> Visibilities.PRIVATE
+        }
+
```
jsjeon: Counterpart I referred is: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt#L183 which includes computation of JVM field visibility, whereas here we don't deal with JVM fields, right?

mglukhikh: Well, yes, JVM fields is a separate questions.

mglukhikh:
Merged manually, thank you!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
jsjeon: Addressed at https://github.com/JetBrains/kotlin/pull/3570

PR #3244, opened 2020-03-31
[JVM] Improve debug step behavior around lambdas.

kandersen:
[JVM] Improve debug step behavior around lambdas. …

- Add tests to clarify problematic behavior
- Avoid line numbers on return instructions of lambdas without explicit returns


compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1786,6 +1785,14 @@ private static boolean isLambdaVoidBody(@NotNull KtElement bodyExpression, @NotN
         return false;
     }
 
+    private static boolean isLambdaBody(@NotNull KtElement bodyExpression) {
```
madsager, line 1788: Maybe rewrite isLambdaVoidBody about to:

```
return isLambdaBody(bodyExpression) && Type.VOID_TYPE.equals(returnType);
```

compiler/testData/debug/stepping/voidLambdaStepInline.kt
```diff
@@ -0,0 +1,20 @@
+
+
+
+
+// FILE: test.kt
+fun box(): String {
+    run { println("OK") }
```
madsager, line 7: I believe that we normally avoid having println calls in tests to not generate stdout output. Does this should the same if you call some other function?

compiler/testData/debug/stepping/lambdaStepInline.kt
```diff
@@ -0,0 +1,26 @@
...
+// FILE: test.kt
+inline fun foo(stringMaker: () -> String): String {
+    return stringMaker()
+}
+
+fun box(): String {
+    foo { "OK "}
+    foo {
+        "OK"
+        // This line is hit by the old backend after line 13
```
pyos, line 14: Not anymore though :)

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
```diff
@@ -1753,10 +1753,9 @@ public void returnExpression(@NotNull KtExpression expr) {
         if (!endsWithReturn(expr)) {
             if (isLambdaVoidBody(expr, typeForExpression)) {
                 markLineNumber((KtFunctionLiteral) expr.getParent(), true);
-            }
-            else {
+            } else if (!isLambdaBody(expr)){
                 markLineNumber(expr, true);
-            }
+            }g
```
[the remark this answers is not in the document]
kandersen: Indeed!

compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java
kandersen: Good idea!

compiler/testData/debug/stepping/voidLambdaStepInline.kt
kandersen: Not quite the same, but it still illustrates the issue that Denis pointed out in the chat last night - that there really should be two steps at 7.

compiler/testData/debug/stepping/lambdaStepInline.kt
```diff
@@ -0,0 +1,26 @@
+
```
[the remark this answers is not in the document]
kandersen: Cleaned it up a little. Will just have to eat it and adjust line numbers if needed :laughing:

compiler/testData/debug/stepping/lambdaStepInline.kt
kandersen: We can race PR's ;)

max-kammerer:
@yanex Please take a look

compiler/testData/debug/stepping/lambdaStepInline.kt
```diff
@@ -0,0 +1,23 @@
...
+// FILE: test.kt
+inline fun foo(stringMaker: () -> String): String {
+    return stringMaker()
+}
+
+fun box(): String {
+    foo { "OK "}
+    foo {
+        "OK"
+        // This line is hit by the old backend after line 13
```
max-kammerer, line 11: @kandersen I bit confused by meaning of "after line 13"

max-kammerer: It was wrong linenumber? Or 13 is correct and i didn't caught the meaning?

kandersen: Sorry for the premature "closure", I think I might have hit "Resolve" from my phone.

max-kammerer:
Thank you!

compiler/testData/debug/stepping/lambdaStepInline.kt
max-kammerer: Thanks for explanation!

PR #3247, opened 2020-04-01
KT-37908 "Convert to anonymous object" quickfix: false negative when interface has concrete functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37908


yanex:
Review: OK.

yanex:
Merged into `master`. Thanks!

PR #3248, opened 2020-04-01
JVM: keep call site markers when inlining lambdas into objects

pyos:
A follow-up for KT-35006:

    fun f() = foo {
        bar()
    }
    inline fun foo(crossinline x: () -> Unit) = { x() }()
    inline fun bar() = TODO()

does not provide the option to navigate to bar's call site at all.

~~The changes to SMAPs may look big, but they're basically because of the different order of files. In the example above, `f$$inlined$foo$1` used to have line number 4 pointing directly to line 4 and then an out-of-bounds line number 6 pointing to line 2; now line number 2 points directly to 2 and 6 is remapped to 4. Because only line numbers that are actually used are remapped, this also exposes some differences in line number generation between JVM and JVM_IR.~~


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -113,21 +114,14 @@ class AnonymousObjectTransformer(
...
+        // the original line numbers because SMAPParser does not parse call site markers.
+        if (inliningContext.isInliningLambda) {
+            sourceMap = SMAPParser.parseOrCreateDefault(null, sourceInfo, oldObjectType.internalName, 1, 65535)
+            sourceMapper = SourceMapper(sourceMap.sourceInfo)
         } else {
-            if (sourceInfo != null) {
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
-            sourceMapper = IdenticalSourceMapper
+            sourceMap = SMAPParser.parseOrCreateDefault(debugInfo, sourceInfo, oldObjectType.internalName, 1, 65535)
```
pyos, line 123: @max-kammerer note that this is more accepting than the old code, so generating trivial SMAPs in anonymous objects in inline function is no longer necessary. If the inliner hits an object that has no SMAP, it will create one and map all line numbers to the original file.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -113,21 +114,14 @@ class AnonymousObjectTransformer(
...
-                SourceMapper.createFromSmap(SMAPParser.parse(debugInfo!!))
-            } else {
-                //seems we can't do any clever mapping cause we don't know any about original class name
-                IdenticalSourceMapper
-            }
-            if (sourceInfo != null && !GENERATE_SMAP) {
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
+        // When regenerating objects in lambdas, we have to pass the SMAP straight through and keep
+        // the original line numbers because SMAPParser does not parse call site markers.
```
pyos, line 118: This also means call sites are currently lost during regeneration:

```
fun box() {
    foo<Any>()
}

inline fun <reified T> foo() = object {
    fun run() = bar<T>()
}.run()

inline fun <reified T> bar() {
    null as T
}

// NAVIGATE_TO_CALL_SITE
// FILE: inlineFunCallSiteInRegeneratedObject.kt
// LINE: 6
```
There's no option to navigate to `bar`'s call site.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
max-kammerer: > is no longer necessary

Do you mean just for compatibility with old compiler?

max-kammerer:
Is it correct that trough regeneration of anonymous object/lambda from inline function source info for them is switched from original one to call site one (and original mappings are kept through file `+X filename` where X > 1)

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
pyos: Well, yeah.

pyos:
Yes, that's exactly it.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
max-kammerer: Is it possible to parse them and make some wise remapping if needed (maybe with additional support in navigation) to support `bar` navigation? //cc @yanex

pyos: Of course, just might take some time to ensure the parser is as general as possible.

pyos: So I wrote the code for that, but then realized it's not that easy because in this case call site markers might need to point to several files. E.g. the KotlinDebug stratum of the regenerated object in
```
// FILE: 1.kt
fun box() {
    foo {
        bar()
    }
}
// FILE: 2.kt
inline fun foo(crossinline x: () -> Unit) = object {
    fun run() {
        bar()
        x()
    }
}.run()

inline fun <reified T> bar() = ...
```
should map some line numbers to 2.kt:4 and some to 1.kt:4. This requires *even more* refactoring of SourceMapper and SMAPBuilder.

pyos: Done.

pyos:
@max-kammerer ping?

max-kammerer:
@pyos Sorry for delay
I'm dumbed on first commit, it's affect test data could you clarify why it changed?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -217,54 +128,40 @@ data class SMAPAndMethodNode(val node: MethodNode, val classSMAP: SMAP)
...
-        }
+    fun toSourceInfo(): SourceInfo =
+        SourceInfo(name, path, lineMappings.fold(0) { result, mapping -> max(result, mapping.source + mapping.range - 1) })
+
+    fun mapNewLineNumber(source: Int, currentIndex: Int, callSite: SourcePosition?): Int {
+        // Save some space in the SMAP by reusing (or extending if it's the last one) the existing range.
+        // TODO some *other* range may already cover `source`; probably too slow to check them all though.
+        //   Maybe keep the list ordered by `source` and use binary search to locate the closest range on the left?
+        val mapping = lineMappings.lastOrNull()?.takeIf { it.canReuseFor(source, currentIndex, callSite) }
+            ?: lineMappings.firstOrNull()?.takeIf { it.canReuseFor(source, currentIndex, callSite) }
```
max-kammerer, line 139: As I understand here you try to reuse  mapping from original source. But if it not last (previous line) then smth is already mapped and `canReuseFor` should be false, shoudn't it?

Most majority of tests have only top level declaration so file mapping mapped to class file 1:1.
Could  you add some test that contains classes so actual source range in class would be wider that actual class declaration. Would it be any problem with such cases: when finding some existing mapping outside real declaration. Could you also add navigation test for such cases?

pyos:
>I'm dumbed on first commit, it's affect test data could you clarify why it changed?

It's because AnonymousObjectTransformationInfo now starts from an empty source map, then re-adds all line numbers in the order in which they appear in method nodes. Take `mappingInInlineFunLambda.kt`, for example: the original object, `_1Kt$test$1`, contains the following sequence of line numbers:
1. (1.kt, _1Kt$test$1, 13)
2. (1.kt, _1Kt$test$1, 14)
3. (1.kt, _1Kt, 7) <-- inlined myrun()
4. (1.kt, _1Kt, 8) <-- call of inline lambda s()
5. (1.kt, _1Kt, 9) <-- end of myrun()
6. (1.kt, _1Kt$test$1, 15)
Of these, 3-5 are out-of-bounds line numbers remapped by the SMAP. Thus, when s() is inlined, new line numbers are inserted between 4 and 5 and the range is broken up, and the regenerated copy of the object has a range for 1.kt:7-8 followed by a range for the lambda and a range for 1.kt:9, rather than a range for 1.kt:7-9.

I'm certain there are more complicated examples where the old behavior causes creation of redundant ranges (e.g. if inline calls are arranged so that the equivalent of 1.kt:7-9 here is not the last range for that file, so after inlining of the lambda a new range for 1.kt:9 will be created even though it's already mapped).

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
pyos: For every range except the last, `canReuseFor` returns true if this source line number is already within the range (and the call site marker is the same). Thus for the original file it's exactly equivalent to the old `if (sameFile && lineNumber <= smap.sourceInfo.linesInFile)` check, as `sameFile` is true if call site marker is null, and the first range is from 1 to smap.sourceInfo.linesInFile.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAPParser.kt
```diff
@@ -21,60 +21,64 @@ object SMAPParser {
...
+            val destMultiplierSeparator = line.indexOf(',', destSeparator).let { if (it < 0) line.length else it }
 
-            val fileIndex = lineMapping.substring(fileSeparator + 1, rangeSeparator).toInt()
-            val targetIndex = lineMapping.substring(targetSplit + 1).toInt()
-            fileMappings[fileIndex]!!.mapNewInterval(originalIndex, targetIndex, range)
+            val file = fileMappings[line.substring(fileSeparator + 1, sourceRangeSeparator).toInt()] ?: return null
+            val source = line.substring(0, fileSeparator).toInt()
+            val dest = line.substring(destSeparator + 1, destMultiplierSeparator).toInt()
+            val range = when {
+                // These two fields have a different meaning, but for compatibility we treat them the same. See `SMAPBuilder`.
```
max-kammerer, line 75: What is migration plan for proper meaning?

max-kammerer: Could you also add navigation test with `// !LANGUAGE: +CorrectSourceMappingSyntax` (Please check that debugger support such directives)?

pyos: Since CorrectSourceMappingSyntax is tagged as a 1.5 feature, the parser can be modified to handle the fields according to spec in 1.6, although I don't think this is particularly useful.

>Could you also add navigation tests with // !LANGUAGE: +CorrectSourceMappingSyntax (Please check that debugger support such directives)?

Done.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -72,14 +71,8 @@ class SourceMapCopier(val parent: SourceMapper, private val smap: SMAP, val call
             return mappedLineNumber
         }
 
-        val range = lastVisitedRange?.takeIf { lineNumber in it }
-            ?: smap.findRange(lineNumber)
-            ?: error("Can't find range to map line $lineNumber in ${smap.sourceInfo.source}: ${smap.sourceInfo.pathOrCleanFQN}")
```
max-kammerer: I'm not sure that this is major, but in anyway: why are you prefer to remove `error`?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -117,7 +117,8 @@ class AnonymousObjectTransformer(
         // When regenerating objects in lambdas, keep the old SMAP and don't remap the line numbers to save time.
         // The result is effectively the same anyway.
         val debugInfoToParse = debugInfo.takeIf { !inliningContext.isInliningLambda }
-        sourceMap = SMAPParser.parseOrCreateDefault(debugInfoToParse, sourceInfo, oldObjectType.internalName, 1, 65535)
+        val (firstLine, lastLine) = methodsToTransform.lineNumberRange()
```
max-kammerer, line 120: `constructor` is missed in `methodsToTransform`
Please add test if change wouldn't affect any test

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
pyos: IMO it's better to accept as much input as possible without throwing internal errors. In particular, this error is why the "synthetic" line numbers have to be under a flag right now even though compilers not aware of their meaning could've simply discarded them.

max-kammerer: Thank you! cc @yanex

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAPParser.kt
max-kammerer: NB: I'm mostly aware of debugger because he should distinguish old format from new one of proper ranging (but can't imagine any case where it would be major). Looks like class abi version could be used here

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
pyos: Fixed, thanks for pointing this out. I'm not surprised that no test broke: because objects inside inline functions always have SMAPs, hitting this bug would require having an init block in an object inside an inline lambda in a non-inline function while using the non-IR backend.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -113,22 +114,12 @@ class AnonymousObjectTransformer(
...
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
-        } else {
-            if (sourceInfo != null) {
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
-            sourceMapper = IdenticalSourceMapper
-        }
+        // When regenerating objects in lambdas, keep the old SMAP and don't remap the line numbers to save time.
+        // The result is effectively the same anyway.
```
max-kammerer, line 118: If object inlines smth inside itself will this mapping be lost with this optimization?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
```diff
@@ -113,22 +114,12 @@ class AnonymousObjectTransformer(
...
-            if (sourceInfo != null && !GENERATE_SMAP) {
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
-        } else {
-            if (sourceInfo != null) {
-                classBuilder.visitSource(sourceInfo!!, debugInfo)
-            }
-            sourceMapper = IdenticalSourceMapper
-        }
+        // When regenerating objects in lambdas, keep the old SMAP and don't remap the line numbers to save time.
```
max-kammerer, line 117: lambdas -> inline lambdas

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/AnonymousObjectTransformer.kt
pyos: What do you mean? There is nothing to inline into objects inside lambdas at this point - the purpose of regenerating them is only to replace `this$0`.

max-kammerer: I meant  next case:
```
package test

inline fun test(s: () -> Unit) {
    s()
}

inline fun test2(crossinline s: () -> Unit) {
    s()
}
```
```
import test.test
import test.test2


fun box(): String {
    var z = "fail"

    test2 {
        { // regenerated object in inline lambda
            test { //will mapping for this call is lost?
                z = "OK"
            }
        }()
    }
    return z;
}
```

pyos: No, the ranges for this call are already present in the original object and will be copied to the regenerated version.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
max-kammerer: I found that I've already asked you about such tests (https://github.com/JetBrains/kotlin/pull/3230#issuecomment-605899143), and they was added in #3230

max-kammerer:
Many thanks! Sorry for delay!

PR #3249, opened 2020-04-01
FIR: reset type reference for 1st arg of GetClassCall.

jsjeon:
Type qualifier for `KClass` has always been `Unit`, i.e., `KClass<Unit>`. During transformation of `FirGetClassCall`, we can reset the type reference of the 1st argument so that the result type of `A::class` can be `KClass<A>`, not `KClass<Unit>`.

Although 1 case (where `.java` access is not followed) is broken, this can unmute 84 cases.

Alternative I tried was to modify `...fir.resolve.ResolveUtils#typeForQualifierByDeclaration`, but that could unmute 40+ and broke 3 cases.


mglukhikh:
Merged manually 👍

PR #3252, opened 2020-04-01
FIR2IR: do not create fake overrides for private property accessors.

jsjeon:
Otherwise, IR validation failed: "Overrides private declaration ..."

An alternative would be to edit `...fir.scopes.impl.FirClassSubstitutionScope#createFakeOverride*`.

### DIFF
--- a/compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
+++ b/compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
@@ -6,6 +6,7 @@
 package org.jetbrains.kotlin.fir.declarations
 
 import org.jetbrains.kotlin.descriptors.ClassKind
+import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.fir.declarations.builder.AbstractFirRegularClassBuilder
 import org.jetbrains.kotlin.fir.declarations.builder.FirTypeParameterBuilder
 import org.jetbrains.kotlin.fir.symbols.impl.FirAnonymousObjectSymbol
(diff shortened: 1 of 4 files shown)


compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
```diff
@@ -55,6 +58,8 @@ inline val FirMemberDeclaration.isFromEnumClass: Boolean get() = status.isFromEn
 
 inline val FirPropertyAccessor.modality get() = status.modality
 inline val FirPropertyAccessor.visibility get() = status.visibility
+inline val FirPropertyAccessor.allowsToHaveFakeOverride: Boolean
+    get() = !Visibilities.isPrivate(visibility) && this != Visibilities.INVISIBLE_FAKE
```
jsjeon, line 62: Counterpart I referred is: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/PropertyGenerator.kt#L180

Technically, `INVISIBLE_FAKE` is not used in `fir` so far, but that visibility is self-explanatory, and having it here won't hurt...

mglukhikh:
Merged manually 👍

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
punzki: I think you mean `get() = !Visibilities.isPrivate(visibility) && visibility != Visibilities.INVISIBLE_FAKE` (`visibility` instead of `this`). Same for the code above.

jsjeon: Duh, you're right! Thanks for spotting this. Will be fixed with https://github.com/JetBrains/kotlin/pull/3259

PR #3255, opened 2020-04-02
IR: restore fixed-point logic in ExternalDependenciesGenerator.

jsjeon:
In 1.3.70 and 1.3.71, we're seeing an infinite loop during dependency calculation. It seems to be a regression from https://github.com/JetBrains/kotlin/commit/1ed23d7c54dbc3d8329ec66189f258a5da6504d6#diff-d99f8724f866b94e951c46087072cad9. Before that, `ExternalDependenciesGenerator#generateUnboundSymbolsAsDependencies` has checked if unbound symbols are stabilized, and if so, it stopped the iteration (even though there could be unbound symbols after deseriazliation attempt).


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/ExternalDependenciesGenerator.kt
```diff
@@ -42,9 +44,9 @@ class ExternalDependenciesGenerator(val symbolTable: SymbolTable, private val ir
                 if (!symbol.isBound) {
                     irProviders.getDeclaration(symbol)
                 }
-                assert(symbol.isBound) { "$symbol unbound even after deserialization attempt" }
+                assert(symbol.isBound) { "${symbol.descriptor} unbound even after deserialization attempt" }
             }
-        } while (unbound.isNotEmpty())
+        } while (haveNotStabilized(unbound, symbolTable.allUnbound))
```
punzki, line 49: The `haveNotStabilized` method seems very expensive computationally. I think all you have to do is to track in the do-while loop whether at least one previously-unbound symbol was bound, and the condition would be `while (unbound.isNotEmpty() || newSymbolBound)`

jsjeon: That sounds better indeed, thanks! Addressed at d6ec9fc40a5722f81791aec67227c5630950e09c (and intentionally being a separate commit)

romanart:
Could you please provide an example when you meet with infinite loop? The best would be adding a test.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/ExternalDependenciesGenerator.kt
```diff
@@ -34,19 +34,21 @@ class ExternalDependenciesGenerator(val symbolTable: SymbolTable, private val ir
...
+            notStabilized = false
 
             for (symbol in unbound) {
                 // Symbol could get bound as a side effect of deserializing other symbols.
                 if (!symbol.isBound) {
                     irProviders.getDeclaration(symbol)
                 }
+                // If any symbol became bound, the entire loop is not stabilized yet.
+                notStabilized = notStabilized || symbol.isBound
                 assert(symbol.isBound) { "${symbol.descriptor} unbound even after deserialization attempt" }
```
gogabr, line 49: Is there a legitimate case where `symbol` does not get bound?
It seems that the change only makes sense when the assertion is disabled.

jsjeon:
> Could you please provide an example when you meet with infinite loop? The best would be adding a test. For me fix looks like it hides some problem which has to be fixed.

I managed to reproduce the issue(s): https://youtrack.jetbrains.com/issue/KT-38026 I saw some tests about stubs, but not sure how to map those scenarios to tests...

gogabr:
Your example contains a resolution failure. The right way to fix it is to either avoid the failure or report it, instead of passing broken data down to the backend.
In the backend code, it is assumed that all symbols are bound, `symbol.owner` is called freely, so any unbound symbol can lead to crash. We should not pass unbound symbols to the backend.

romanart:
As I can see there are two cases with unbound symbols (`IrEnumEntry` and `IrSimpleFunction`) which seems like a linkage (probably about LazyIr) issue. Have you check it btw? 
```
FUN FAKE_OVERRIDE name:getOffset visibility:public modality:OPEN <> ($this:y.MyDrawable) returnType:kotlin.Int [fake_override]
  annotations:
    RestrictTo(value = [GET_ENUM 'UNBOUND IrEnumEntrySymbolImpl' type=IrErrorType]) <= (1)
  overridden:
    UNBOUND IrSimpleFunctionSymbolImpl <= (2)
  $this: VALUE_PARAMETER name:<this> type:y.MyDrawable
```
And I agree with @gogabr that allowing unbound symbols in backend is not good idea because it wasn't designed for it. I would ask you to take a look into `LazyIr` and `DeclarationStubGenerator`.

jsjeon:
Duh, sending PR is not a good way to cherry-pick a hot fix, or maybe I'm not good at setting origin, branch target, etc.

Anyhow, @udalov pointed out a relevant test in https://youtrack.jetbrains.com/issue/KT-38026, which has been introduced with a fix: e54ef3bdb84a1bbc3f2185e6c1cd1a12794f1705 I can confirm cherry-picking it into 1.3.70 resolved the internal issue as well.

PR #3256, opened 2020-04-02
Properly link properties and their getter/setters.

madsager:
This allows removing a hack in the MethodSignatureMapper.

When introducing non-getter methods that dispatch to a getter
we have to be careful to use the right name.

DeepCopyIrTreeWithSymbols should create the link between new
properties and their new getter/setter.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
@@ -90,22 +90,7 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
             return mangleMemberNameIfRequired(accessorName, function)
         }
 
-        val functionName = workaroundPropertyAccessorName(function) ?: function.name.asString()
-        return mangleMemberNameIfRequired(functionName, function)
-    }
-
-    // This is a dirty hack which is needed to correctly map names of property accessors whose correspondingPropertySymbol was lost.
-    // The reason it's needed is that it sometimes gets lost during DeepCopyIrTreeWithSymbols because that algorithm doesn't copy
-    // the link to the property when transforming a simple function. This can be fixed but might require a lot of changes in backends,
-    // and will likely be unneeded as soon as we have proper properties in the IR (IrProperty which is a declaration parent).
-    // TODO: remove this hack as soon as IrProperty is a declaration parent
-    private fun workaroundPropertyAccessorName(function: IrFunction): String? {
-        val name = function.name.asString()
-        val ifGetter = name.removeSurrounding("<get-", ">")
-        if (ifGetter != name) return JvmAbi.getterName(ifGetter)
-        val ifSetter = name.removeSurrounding("<set-", ">")
-        if (ifSetter != name) return JvmAbi.setterName(ifSetter)
-        return null
+        return mangleMemberNameIfRequired(function.name.asString(), function)
     }
 
     private fun mangleMemberNameIfRequired(name: String, function: IrFunction): String {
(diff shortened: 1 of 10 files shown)


madsager:
@udalov can you have a look at this. I opted for translating to the jvm name when lowerings generate non-getter methods that dispatch to a property getter. Do you see any issues that this could cause?

udalov:
I'm slightly worried that this hides the underlying problem even more. I mean, the fact that deep copy doesn't copy `correspondingPropertySymbol` is a bug which we're working around here. Your approach is still probably a bit better than the current hack, but all these `mapFunctionName` calls could at least be qualified with TODOs, because ideally we'd just rely on correct IR structure after all the lowerings and not need any JVM names until the codegen. What do you think?

madsager:
Thanks Alex! I don't think this change hides the underlying problem more. None of the lowering fixes here are related to `DeepCopyIrWithSymbols`. I think what is hiding the issues is actually that `PatchDeclarationParents` do not just fix parents, it also fixes corresponding property symbols. It is not hard to get out of that situation. I added a second commit that does that. Psi2Ir needs to setup `correspondingPropertySymbol` correctly and then deep copy needs to maintain it (which I fixed in commit 1). Let me know what you think.

As for the lowerings and renaming:

For bridge lowering, we are replacing a getter with another function and therefore, we can just properly link in the new function as the getter for the property. I updated commit 1 to do that.

For the other 2 lowerings, what we are generating are methods that call and accessor and that need to have the same name. For that, there is no property to link it up to. I'm wondering if we need a mechanism to link the name of a function to the name of another function? So that when we output, we do not try to generate the name based on the bridge, but we get the name for the bridge from the bridge target. That would avoid mapping the name in the lowerings.

udalov:
Oh, you're right, thanks for the explanation!

I was actually thinking about a different, slightly related problem: the fact that deep copy loses corresponding property symbol if it's applied _after_ the property flattening phase `JvmPropertiesLowering` (there's no corresponding property symbol handling in `DeepCopyIrTreeWithSymbols.visitSimpleFunction`). For some reason I thought that we still have such usages of deep copy in our lowerings. But that's not true, we don't anymore.

So even though deep copy still doesn't support flattened properties, it doesn't manifest right now, so everything is fine. And I hope I understand the Compose use case correctly in that they're only doing a deep copy on the initial IR which still has `IrProperty` intact. So your change indeed helps here and it looks good!

> I think what is hiding the issues is actually that `PatchDeclarationParents` do not just fix parents, it also fixes corresponding property symbols. It is not hard to get out of that situation. I added a second commit that does that. Psi2Ir needs to setup `correspondingPropertySymbol` correctly and then deep copy needs to maintain it (which I fixed in commit 1).

Yeah, I missed that and this makes a lot of sense.

> For bridge lowering, we are replacing a getter with another function and therefore, we can just properly link in the new function as the getter for the property. I updated commit 1 to do that.

👍 

> For the other 2 lowerings, what we are generating are methods that call and accessor and that need to have the same name. For that, there is no property to link it up to.

IMHO it would actually be a bit more "fair" to generate _new_ properties in cases where we create some sort of copies for property accessors. For example, a multifile delegate for a property getter could actually be a getter itself, of a new property created specifically for that purpose in the multifile facade class. Similarly for JvmStatic wrappers.

Conceptually such a delegate or wrapper is, I think, still a property accessor (but of a new copy of the property), and other lowerings should be able to observe that in case it's necessary. In other words, I don't see any reason why a copy of a property accessor would be an ordinary function, not a property accessor :). And doing that would also straightforwardly result in correct JVM name mapping in the codegen, without the need to invoke `mapFunctionName` in lowerings.

One catch here is that both `bridgePhase` and `generateMultifileFacadesPhase` are happening after properties are flattened. So new properties only need to appear in the `correspondingPropertySymbol` fields, and not the IR declaration tree. (A possible future improvement here would be to actually remove the property flattening phase.)

But in any case, sorry for the rant. The change as is looks good to me, so even if I didn't convince you about copying property accessors, I can merge it after the remote run :).

udalov:
@homuroll @romanart Please take a look from the Native & JS's point of view

romanart:
If fix doesn't broke any js test so LGTM.

udalov:
Thanks! :)

PR #3257, opened 2020-04-03
KT-37977 "Replace 'invoke' with direct call" intention: false positive when function is not operator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37977


vladimirdolzhenko:
LGTM, Thank you for your contribution

PR #3258, opened 2020-04-03
KT-37967 Replace 'invoke' with direct call intention adds unnecessary parenthesis

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37967


vladimirdolzhenko:
LGTM, Thanks

PR #3261, opened 2020-04-03
FIR2IR: set Kotlin.Metadata when declaring IR symbols.

jsjeon:
`SymbolTable#declare*` utils have default symbol factory which applies `Kotlin.Metadata` to the IR symbol being created, e.g., 
* Class: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt#L370
* Constructor: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt#L414
* Property: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt#L563
* and so on.

In `fir2ir`, we use those utils with our own symbol factory, which should have a proper step to set the metadata. Otherwise, e.g., reflection call via KClass, which often requires kclass-to-java-class conversion, complains that the target class is not a Kotlin class, due to the lack of the metadata.


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/WrappedDescriptors.kt
```diff
@@ -570,11 +672,11 @@ open class WrappedClassDescriptor(
 
     override fun getSource() = sourceElement
 
-    override fun getConstructors() = owner.declarations.asSequence().filterIsInstance<IrConstructor>().map { it.descriptor }.toList()
+    override fun getConstructors() =
+        owner.declarations.filterIsInstance<IrConstructor>().filter { !it.origin.isSynthetic }.map { it.descriptor }.toList()
```
jsjeon, line 676: There are 10+ cases that failed with visibility conversion error: package-private (which should be normalized in general) doesn't have a counter part. By comparing `psi2ir` path, I noticed that all those cases have private constructor with default values, which is lowered by adding a synthetic constructor with package-private visibility.

The difference is, during serialization, `LazyClassDescriptor#getConstructors` returns constructors that literally appear in the PSI, which means we need to filter out synthetic ones here.

compiler/testData/codegen/box/ir/kt25405.kt
```diff
@@ -1,3 +1,5 @@
+// IGNORE_BACKEND_FIR: JVM_IR
+
```
jsjeon, line 2: This is the only one regression, though. After having a correct metadata, proper serialization started, and found that, when serializing the extension receiver of `tableView` below, type parameter `T` for `Table<T>` didn't have a counter part, type argument in the converted IrType. This means that type arguments for generic types are missing in general. I left a TODO/comment about this on our doc.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/WrappedDescriptors.kt
```diff
@@ -443,7 +443,109 @@ open class WrappedSimpleFunctionDescriptor(
     override fun <V : Any?> getUserData(key: CallableDescriptor.UserDataKey<V>?): V? = null
 
     override fun newCopyBuilder(): FunctionDescriptor.CopyBuilder<out SimpleFunctionDescriptor> {
-        TODO("not implemented")
+        // Use original descriptor's copy builder if available.
+        if (originalDescriptor is SimpleFunctionDescriptor) {
```
romanart, line 447: Just in case Is that possible if `originalDescriptor === this`?

romanart:
LGMT

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/WrappedDescriptors.kt
jsjeon: If that's the case, usage of `newCopyBuilder` in JVM_IR lowering would fail with stack overflow, and I didn't see that at all. :) But, better safe than sorry, so added a check to make sure that original descriptor is not `this`.

romanart: Thank you

mglukhikh:
Merged manually. Thank you!

PR #3265, opened 2020-04-04
KT-31126 JS. Fix super constructor call. Old backend

turansky:
Primary ordinary -> secondary external

### DIFF
--- a/js/js.translator/testData/box/jsModule/externalConstructor.kt
+++ b/js/js.translator/testData/box/jsModule/externalConstructor.kt
@@ -1,0 +1,22 @@
+package foo
+
+@JsModule("lib")
+open external class A(data: String) {
+    constructor(data: Int)
+    constructor(data: Boolean)
+
+    val data: Any
+}
+
+class B(data: String) : A(data)
+
+class C(data: Int) : A(data)
+
+class D(data: Boolean) : A(data)
+
+fun box(): String {
+    assertEquals("13", B("13").data)
+    assertEquals(42, C(42).data)
+    assertEquals(true, D(true).data)
+    return "OK"
+}
--- a/js/js.translator/testData/box/jsModule/externalConstructor.kt
+++ b/js/js.translator/testData/box/jsModule/externalConstructor.kt
@@ -1,3 +1,5 @@
+// EXPECTED_REACHABLE_NODES: 1334
+// MODULE_KIND: AMD
 package foo
 
 @JsModule("lib")
--- a/js/js.tests/test/org/jetbrains/kotlin/js/test/ir/semantics/IrBoxJsTestGenerated.java
+++ b/js/js.tests/test/org/jetbrains/kotlin/js/test/ir/semantics/IrBoxJsTestGenerated.java
@@ -4956,6 +4956,11 @@ public class IrBoxJsTestGenerated extends AbstractIrBoxJsTest {
             runTest("js/js.translator/testData/box/jsModule/externalClassWithDefaults.kt");
         }
 
+        @TestMetadata("externalConstructor.kt")
+        public void testExternalConstructor() throws Exception {
+            runTest("js/js.translator/testData/box/jsModule/externalConstructor.kt");
+        }
+
         @TestMetadata("externalFunction.kt")
         public void testExternalFunction() throws Exception {
             runTest("js/js.translator/testData/box/jsModule/externalFunction.kt");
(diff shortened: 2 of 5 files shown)


romanart:
Thank you for PR. Could please add a test which your change fixes?

turansky:
Test for constructor require external `js` resource (as in my [test example](https://github.com/turansky/kotlin-battleground/tree/master/KT-31126))
* [Test](https://github.com/turansky/kotlin-battleground/blob/master/KT-31126/src/test/kotlin/com/test/DataProviderTest.kt)
* [External resource](https://github.com/turansky/kotlin-battleground/blob/master/KT-31126/src/main/resources/data-provider.js)

I don't see tests with external in `js.translator`.
Is it possible to use external resources in `js.translator`?
Is there better place for such tests?

romanart:
Hope this could help you
https://github.com/JetBrains/kotlin/blob/master/js/js.translator/testData/box/jsModule/externalClass.kt
https://github.com/JetBrains/kotlin/blob/master/js/js.translator/testData/box/jsModule/externalClass.js

Think `js/js.translator/testData/box/jsModule` is a good place for such test.

turansky:
> Could please add a test which your change fixes?
Done

romanart:
Pushed to CI. Once it done without failures will merge your fix.

Thank you :rocket:

turansky:
@romanart What about one more review #3219? Tests already added.

romanart:
@turansky Could you please add the following lines in the beginning of `externalConstructor.kt`?
```
// EXPECTED_REACHABLE_NODES: 1334
// MODULE_KIND: AMD
```
Those lines says test infra how to run this test. Otherwise it fails with 
```
CALL_TO_JS_MODULE_WITHOUT_MODULE_SYSTEM: Can't access constructor of 'A' marked with @JsModule annotation from non-modular project (11,25)
```

turansky:
@romanart Which task generate following code:
```
// EXPECTED_REACHABLE_NODES: 1334
```

romanart:
`gradle :js:js.tests:test --tests "org.jetbrains.kotlin.js.test.ir.semantics.IrBoxJsTestGenerated$JsModule.testExternalClass"`

turansky:
@romanart comments added :)

js/js.translator/testData/box/jsModule/externalConstructor.kt
```diff
@@ -0,0 +1,24 @@
+// EXPECTED_REACHABLE_NODES: 1334
+// MODULE_KIND: AMD
+package foo
+
+@JsModule("lib")
```
bashor, line 5: Isn't it reproduced without any module systems?

bashor:
It would be nice to check that we have tests for all cases? You can just create an issue for that.

|ctor|super ctor| super is external |
| :--- | :--- | :--- | 
|primary|primary| false |
|primary|secondary| false |
|secondary|primary| false |
|secondary|secondary| false |
|primary|primary| true |
|primary|secondary| true |
|secondary|primary| true |
|secondary|secondary| true |

cc @romanart

js/js.translator/testData/box/jsModule/externalConstructor.kt
turansky: Works in both cases.
Modular variant can be simply tested in [separate project](https://github.com/turansky/kotlin-battleground/blob/master/KT-31126/src/test/kotlin/foo/externalConstructor.kt)

turansky:
> You can just create an issue for that.

@bashor You = @me?

js/js.translator/testData/box/jsModule/externalConstructor.kt
bashor: My thought was that (without the fix) it should be reproducible without any module system. Isn't it?

bashor:
@turansky, yes, could you please create an issue for https://github.com/JetBrains/kotlin/pull/3265#issuecomment-609901039

js/js.translator/testData/box/jsModule/externalConstructor.kt
turansky: Yes, it is.

turansky: Modular system used like more typical.

turansky:
[KT-38114](https://youtrack.jetbrains.com/issue/KT-38114)

PR #3266, opened 2020-04-06
[FIR]: add backing fields to const properties during Fir2Ir translation.

juan-chen:


mglukhikh:
Merged manually 👍  NB: I also moved backing field creation into Fir2IrDeclarationStorage: https://github.com/JetBrains/kotlin/commit/4cb4aae7eeff98ce22ed0ec5e003039e4bf8ff97

PR #3268, opened 2020-04-07
FIR2IR: retrieve members from WrappedClassDescriptor.

jsjeon:
Metadata generated in FIR path doesn't have member information, which prevented some of reflection calls, such as KClass<*>#getDeclaredMembers, from literally finding (declared) members. From the `owner` of `WrappedClassDescriptor`, we can retrieve declared members by leveraging flags and origins of members.


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/WrappedDescriptors.kt
```diff
@@ -665,7 +672,36 @@ open class WrappedClassDescriptor(
...
+                                it.origin != IrDeclarationOrigin.ENUM_CLASS_SPECIAL_MEMBER &&
+                                // TODO: Better way to filter out <clinit>, e.g., JvmLoweredDeclarationOrigin.CLASS_STATIC_INITIALIZER?
+                                (!it.isStaticMethodOfClass || !it.nameForIrSerialization.isSpecial)
+                        is IrProperty -> !it.isFakeOverride && !it.origin.isSynthetic &&
+                                it.origin != IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR
+                        is IrField -> !it.isFakeOverride && !it.origin.isSynthetic &&
+                                it.origin != IrDeclarationOrigin.FIELD_FOR_OBJECT_INSTANCE &&
+                                it.origin != IrDeclarationOrigin.FIELD_FOR_ENUM_ENTRY
+                        else -> !it.isFakeOverride && !it.origin.isSynthetic
+                    }
```
jsjeon, line 702: I have to admit this is quite ad-hoc, but at the same time, tweaking `synthetic` in `IrDeclarationOrigin` seems risky and/or deserves a separate commit/PR.

E.g., I found a TODO for `<clinit>`: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/StaticInitializersLowering.kt#L43, since `<clinit>` is synthesized only if there are static jvm fields. But, marking it directly on `JvmLoweredDeclarationOrigin.CLASS_STATIC_INITIALIZER` makes `<clinit>` have `synthetic` modifier.

Similarly, special static fields and enum-specific members/entries are also somewhat synthetic, not user contributed/declared.

romanart:
Could you please explain a bit more which exact part of backend requires this API? I would strongly recommend  to fix it there instead of produce additional tech debt in wrapped descriptors.

udalov:
Yeah, it seems like the best approach here would be to allow `MetadataSource` to be backed by something other than descriptors, i.e. FIR elements, and then basically reimplementing all of the `serialization` module and `JvmSerializerExtension` for FIR elements as well.

I don't know if it's a good idea to do this for FIR in its current state though.

mglukhikh:
After internal discussion we decided to implement FIR elements serialization, probably I will do it next week, hope after it this reflection stuff will work. As for this PR, I have to close it, sorry.

jsjeon:
> Could you please explain a bit more which exact part of backend requires this API?

As mentioned by @udalov current descriptor-based `serialization` used those APIs.

> After internal discussion we decided to implement FIR elements serialization

Okay. Please note that, after FIR element serialization, you may be able to revert some changes in `WrappedClassDescriptor` introduced at https://github.com/JetBrains/kotlin/pull/3261

jsjeon:
One more note before I forgot (and I should admit wrapped descriptor is not good for metadata serialization because of cases like this): the owner of wrapped descriptor can be changed during IR lowering, and the very next error I saw was, to not allow getter/setter access for certain type of properties, those fields in `IrProperty` are reset to null in the middle, which bothers serializing getter/setter flags for that property at the end. So, as the name implies, wrapped descriptor is not a clear/immutable snapshot of declarations.

jsjeon:
> > After internal discussion we decided to implement FIR elements serialization
> 
> Okay. Please note that, after FIR element serialization, you may be able to revert some changes in `WrappedClassDescriptor` introduced at #3261

It looks like FIR element serialization is in (https://github.com/JetBrains/kotlin/tree/master/compiler/fir/fir-serialization), so I'm reverting those changes I mentioned before: https://github.com/JetBrains/kotlin/pull/3366

PR #3271, opened 2020-04-07
Parcelize: Implement support for the JVM IR backend

sfs:
This PR implements support for the JVM IR backend in the `@Parcelize` part of the Kotlin Android extensions. The PR also adds a number of tests and reveals several problem with the current implementation of `@Parcelize`. With the exception of inline class support, these should all be fixed in the JVM IR version and I'll add proper YouTrack issues later.

There are some things about the implementation which are probably a bit weird. So about that:

I first tried to do this based on the existing `IrGenerationExtension` extension point, but this turned out to be impractical. That's because the IR is only half constructed at this point and we essentially need to use descriptors for everything. This won't work with FIR, and in any case, we're trying to reduce descriptor usage in the IR backend. Based on some discussion on Slack, I understand that this situation is temporary, so I added a temporary extension point after IR construction and before any of the lowerings. This works beautifully.

The next problem is that we occasionally need to resolve symbols which won't be present in the dependencies we compile against (since we compile against some version of android-api.jar, which does not include hidden methods). This is not much of a problem, since we can just create stubs for the necessary symbols, in the same way as we are already doing in the JVM IR backend. The only problem here was that we needed rather more of these stubs compared to the compiler backend, which is the reason for the large (auto-generated) `AndroidSymbols` class. As an added advantage, this will fail much more gracefully if a user ends up compiling against a future version of the dependencies without the necessary symbols. By using fixed stubs, we could end up with a NoSuchMethodError at runtime, while using resolution we would end up with some new version of the method, which could break in many interesting new ways if, e.g., only the types in the signature change (since the compiler backend is much more permissive about type errors than the Kotlin language).

Finally, I had trouble getting the tests to run and just hard coded some versions into the build files and test harness. This is probably the wrong way to do this and I would be grateful if someone could point out a better way to handle this.


plugins/android-extensions/android-extensions-compiler/src/org/jetbrains/kotlin/android/parcel/ir/AndroidSymbols.kt
```diff
@@ -0,0 +1,477 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+// This file was autogenerated based on android.jar, do not edit it directly.
```
kandersen, line 5: Personal nit, but I feel it useful to have a reference to the script that generated this file, too.

sfs:
I filed YouTrack issues for the bugs I found here: [KT-38103](https://youtrack.jetbrains.com/issue/KT-38103), [KT-38104](https://youtrack.jetbrains.com/issue/KT-38104), [KT-38105](https://youtrack.jetbrains.com/issue/KT-38105), [KT-38106](https://youtrack.jetbrains.com/issue/KT-38106), [KT-38107](https://youtrack.jetbrains.com/issue/KT-38107). I've also added tests for several previously reported issues.

Additionally, while it's not a bug I found the lack of scoping for custom Parcelers surprising (see `testCustomParcelerScoping`). It might be better to use lexical scoping.

Furthermore, there seems to be a regression in `testEfficientParcelable`, which is currently using the generic parcelable serializer on the JVM backend.

plugins/android-extensions/android-extensions-compiler/src/org/jetbrains/kotlin/android/parcel/ir/AndroidSymbols.kt
sfs: I hear you. I'm just putting this off for later, since the script in question is more than a little bit messy and larger than it should be. That said, I would also like to include it at some point.

plugins/android-extensions/android-extensions-compiler/testData/parcel/box/customSerializerWriteWith.kt
```diff
@@ -1,3 +1,5 @@
+// IGNORE_BACKEND: JVM
+// The JVM backend is missing support for custom parcelers in List<String>
```
udalov, line 2: It'd be really nice to mention the specific YouTrack issues you've created in these ignore directives

plugins/android-extensions/android-extensions-compiler/test/org/jetbrains/kotlin/android/parcel/AbstractParcelBytecodeListingTest.kt
```diff
@@ -20,6 +20,6 @@ abstract class AbstractParcelBytecodeListingTest : AbstractAsmLikeInstructionLis
             ?: throw RuntimeException(
                 "Unable to get a valid path from 'ideaSdk.androidPlugin.path' property, please point it to the Idea android plugin location"
             )
-        environment.updateClasspath(listOf(JvmClasspathRoot(File(androidPluginPath, "layoutlib.jar"))))
+        environment.updateClasspath(listOf(JvmClasspathRoot(File(androidPluginPath, "layoutlib-26.5.0.2.jar"))))
```
udalov, line 23: I'd extract this file name as a constant somewhere

plugins/android-extensions/android-extensions-compiler/build.gradle.kts
```diff
@@ -33,7 +35,7 @@ dependencies {
 
     testRuntime(intellijPluginDep("junit"))
 
-    robolectricClasspath(commonDep("org.robolectric", "robolectric"))
+    robolectricClasspath("org.robolectric:robolectric:4.0")
```
udalov, line 38: Could you please try to change `versions.robolectric` in the root `build.gradle.kts` instead?

plugins/android-extensions/android-extensions-compiler/testData/parcel/box/customSerializerWriteWith.kt
sfs: Good idea. Done!

plugins/android-extensions/android-extensions-compiler/test/org/jetbrains/kotlin/android/parcel/AbstractParcelBytecodeListingTest.kt
sfs: Done. Moved the whole thing to the companion of `AbstractParcelBoxTest`, since the path is effectively static.

plugins/android-extensions/android-extensions-compiler/build.gradle.kts
sfs: Certainly, thanks for the pointer.

sfs:
Fixed two nasty bugs. There was missing metadata, since the generated declarations did not set the metadata field and psi2ir generates broken stubs for the synthetic parcelize descriptors. Both are fixed in commit bf96d01, but I hope that this is going to be a temporary arrangement until we have a better plugin API.

I wanted to use the `IrExtensionGenerator` extension point to create proper stubs for the synthetic descriptors, but unfortunately there is no way to make this work right now since the API has no access to symbol tables or type translators. Changing this would require a number of other changes to psi2ir and is probably not a good idea right now. So for now, I've added a work around (remap the broken symbols and set the correct metadata manually), but I really want to get back to this later.

udalov:
Thanks, Steven! Merged to master. I've also deprecated `PureIrGenerationExtension` in 7a369b3a6ae3f6cdb9152741adf47efa5282e863 since AFAIK @romanart is working on common IR changes that would make this extension point unnecessary.

PR #3273, opened 2020-04-08
KT-38068 Explicitly declare throw of IOException for `readLine`

Him188:
The following is copied from [KT-38068](https://youtrack.jetbrains.com/issue/KT-38068)

Current signature:
```kotlin
fun readLine(): String?
```
This function uses `InputStream.read()` from Java. 
It's very error-prone because `readLine` is blocking but can be used in a suspend function without the warning "BlockingMethodInNonBlockingContext" .

Should be modified to
```kotlin
@Throws(IOException::class)
fun readLine(): String?
```
to help create this warning:
image

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/io/Console.kt
+++ b/libraries/stdlib/jvm/src/kotlin/io/Console.kt
@@ -7,6 +7,7 @@
 
 package kotlin.io
 
+import java.io.IOException
 import java.io.InputStream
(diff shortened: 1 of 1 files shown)


Him188:
Shelved by [KT-38068](https://youtrack.jetbrains.com/issue/KT-38068#focus=streamItem-27-4111628.0-0)

PR #3274, opened 2020-04-08
ForLoopsLowering: Add support for unsigned progressions

punzki:
Symbols for unsigned numbers and progressions cannot be constructed from scratch; the functions used are inline. I initially attempted to add sources for the same classes to the minimal stdlib for JVM and JS but this required pulling in much of the full stdlib for JS. Instead, upon @skuzmich's suggestion, these symbols are now "optional" so we don't need to add the full stdlib to all tests.

The bytecode text tests I added were "ported" over from the existing tests for signed progressions. I picked a set that I thought gave enough test coverage of the optimizations.

I tested this on JVM, JS, and Native. Corresponding changes in K/N repo are required; see https://github.com/JetBrains/kotlin-native/pull/4077.


punzki:
@udalov Gentle ping!

udalov:
Sorry for the delay, will review soon!

BTW looks like newly added tests are failing for FIR, could you please add the corresponding directives?
https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/2852260

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/Ir.kt
```diff
@@ -64,6 +64,12 @@ open class BuiltinSymbolsBase(protected val builtIns: KotlinBuiltIns, private va
             builtInsPackage(*packageNameSegments).getContributedClassifier(name, NoLookupLocation.FROM_BACKEND) as ClassDescriptor
         )
 
+    private fun getClassOrNull(name: Name, vararg packageNameSegments: String = arrayOf("kotlin")): IrClassSymbol? =
```
udalov, line 67: I'd rewrite `getClass` above now with something like
```
getClassOrNull(name, *packageNameSegments) ?: error("Class $name is not found in ${packageNameSegments.toList()}")
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
```diff
@@ -262,13 +318,17 @@ internal abstract class HeaderInfoBuilder(context: CommonBackendContext, private
...
 
     // TODO: Include unsigned types
-    private val progressionElementTypes = listOf(
-        context.irBuiltIns.byteType,
-        context.irBuiltIns.shortType,
-        context.irBuiltIns.intType,
-        context.irBuiltIns.longType,
-        context.irBuiltIns.charType
-    )
+    private val progressionElementTypes = listOfNotNull(
```
udalov, line 321: I suppose the TODO above is fixed

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -160,60 +166,117 @@ internal abstract class NumericForLoopHeader<T : NumericHeaderInfo>(
...
+            // `compareTo` must be used for UInt/ULong; they don't have intrinsic comparison operators.
+            val intCompFun = if (isLastInclusive) {
+                builtIns.lessOrEqualFunByOperandType.getValue(builtIns.intClass)
+            } else {
+                builtIns.lessFunByOperandType.getValue(builtIns.intClass)
+            }
+            val elementCompareToFun = progressionElementType.getClass()!!.functions.single {
+                it.name == OperatorNameConventions.COMPARE_TO &&
+                        it.valueParameters.size == 1 &&
+                        it.valueParameters[0].type == progressionElementType
```
udalov, line 180:
```suggestion
                it.name == OperatorNameConventions.COMPARE_TO &&
                        it.dispatchReceiverParameter != null && it.extensionReceiverParameter == null &&
                        it.valueParameters.size == 1 &&
                        it.valueParameters[0].type == progressionElementType
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
```diff
@@ -13,40 +13,65 @@ import org.jetbrains.kotlin.ir.declarations.IrVariable
...
+
+    fun castStepIfNecessary(step: IrExpression, context: CommonBackendContext) =
+        step.castIfNecessary(stepType(context.irBuiltIns), stepCastFunctionName)
+
+    private fun IrExpression.castIfNecessary(targetType: IrType, numberCastFunctionName: Name) =
+        // This expression's type could be Nothing from an exception throw.
+        if (type == targetType || type.isNothing()) {
+            this
+        } else {
+            val castFun = type.getClass()!!.functions.single { it.name == numberCastFunctionName && it.valueParameters.isEmpty() }
```
udalov, line 63:
```suggestion
            val castFun = type.getClass()!!.functions.single {
                it.name == numberCastFunctionName &&
                        it.dispatchReceiverParameter != null && it.extensionReceiverParameter == null && it.valueParameters.isEmpty()
            }
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -160,60 +166,117 @@ internal abstract class NumericForLoopHeader<T : NumericHeaderInfo>(
...
-                            irCall(builtIns.greaterFunByOperandType[stepType.classifierOrFail]!!).apply {
+                            irCall(builtIns.greaterFunByOperandType.getValue(stepType.classifierOrFail)).apply {
                                 putValueArgument(0, irGet(stepVariable))
-                                putValueArgument(1, if (isLong) irLong(0) else irInt(0))
+                                putValueArgument(1, if (progressionType.isLong) irLong(0) else irInt(0))
                             },
-                            irCall(compFun).apply {
-                                putValueArgument(0, irGet(inductionVariable))
-                                putValueArgument(1, lastExpression)
+                            if (progressionType.isUnsigned) {
```
udalov, line 235: Looks like this part, and the one below, is exactly the same as in handling `DECREASING`/`INCREASING` progressions above. Maybe extract those into some functions and reuse them here?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/ProgressionHandlers.kt
```diff
@@ -458,29 +489,29 @@ internal class StepHandler(
...
 
-        // Call `getProgressionLastElement(first, last, step)`
-        val stepType = progressionType.stepType(context.irBuiltIns).toKotlinType()
-        val getProgressionLastElementFun = symbols.getProgressionLastElementByReturnType[stepType]
-            ?: throw IllegalArgumentException("No `getProgressionLastElement` for step type $stepType")
+        // Call `getProgressionLastElement(first, last, step)`. The following overloads are present in the stdlib:
+        //   - getProgressionLastElement(Int, Int, Int): Int         // Used by IntProgression and CharProgression (uses Int step)
+        //   - getProgressionLastElement(Long, Long, Long): Long     // Used by LongProgression
+        //   - getProgressionLastElement(UInt, UInt, Int): UInt      // Used by UIntProgression (uses Int step)
+        //   - getProgressionLastElement(ULong, ULong, Long): Long   // Used by ULongProgression (uses Int step)
```
udalov, line 496:
```suggestion
        //   - getProgressionLastElement(ULong, ULong, Long): ULong   // Used by ULongProgression
```

compiler/testData/codegen/bytecodeText/forLoop/forInUntil/forInUntilIntMinValue.kt
```diff
@@ -17,10 +17,10 @@ fun f(a: Int): Int {
 // 0 getFirst
 // 0 getLast
 // 0 getStep
-// 0 LINENUMBER 7
```
udalov, line 20: Why is this change needed? Looks like a regression...

punzki: Firstly, there is no code in line 7. It used to be correct, but it was not adjusted to line 6 in https://github.com/JetBrains/kotlin/commit/496765f41e26ea05a203a92e3ac39fea08e6604e#diff-eb501130ee852898f2efdc4789134b52

Also, the loop is only eliminated as dead code in JVM_IR. So `0 LINENUMBER 6` is only relevant for JVM_IR.

punzki:
> Sorry for the delay, will review soon!
> 
> BTW looks like newly added tests are failing for FIR, could you please add the corresponding directives?
> https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/2852260

Updated the PR, disabled the failing tests for FIR using the generator.

compiler/testData/codegen/bytecodeText/forLoop/forInUntil/forInUntilIntMinValue.kt
udalov: Thanks, makes total sense.

udalov:
I wonder if we can do something about the resulting bytecode before specifying its current structure so extenslively in the bytecode text tests?

For example, taking the `stepNonConstOnNonLiteralProgression.kt` test case, the old backend produces 45 lines of bytecode, where as JVM IR produces 98. Looks like most of the extra bytecode is either unsigned integer constructor invocations (`kotlin/UInt.constructor-impl`) which can probably be dealt with, or lots of stores followed immediately by loads, which I don't exactly know the nature of.

punzki:
> stepNonConstOnNonLiteralProgression

I chatted with @sfs sometime back, and he mentioned we might eliminate some of the constructor calls using `<unsafe-coerce>`s. But I'm not sure how many we can eliminate, maybe the increments.

Unsigned progressions are optimized mostly the same way as signed progressions in the non-IR backend ([code](https://github.com/JetBrains/kotlin/blob/1ffdc11d3101c1f98676ddcee3641cdda6697ff3/compiler/backend/src/org/jetbrains/kotlin/codegen/range/RangeValues.kt#L64-L67)), except for comparisons. However, `getProgressionLastElement` has a different version for `UInt/ULong` and the calculations will _probably_ not work correctly if we use the signed version; the non-IR backend doesn't optimize stepped progressions so it's irrelevant for non-IR. Comparisons are also different for unsigned so we'll have to manually add calls to `uintCompare`.

I'll look into this some more...

punzki:
@udalov I was able to figure out where the unnecessary store/loads were coming from. There were a few places where we were creating temporary variables where we didn't need to. I believe I eliminated most of them (see 1st additional commit).

I think I was also able to eliminate all unnecessary construction/boxing/unboxing of `UInt/ULong`. Basically I was using signed types for the induction variable and bounds. The two places where we do conversion are:
1. Bounds checking: This must be done with unsigned types so that codegen will insert calls to `uint/ulongCompare`.
2. When calling `getProgressionLastElement`: This function does calculations (e.g., modulo) which _must_ be done with unsigned numbers, or else the last element will be incorrect. The return value is unsigned and also must be converted back to signed.

For JVM, we are using the `<unsafe-coerce>` intrinsic to convert from unsigned to signed and vice-versa. For JS and other backends, where the intrinsic symbol is not available, we are falling back to calling `to(U)Int/Long`. Do you know if there's an equivalent on other backends? The tests pass on JS, and I'm currently running for Native.

I still need to clean up the code... I want to refactor `ProgressionType` and move more logic into it. I don't like how the code looks right now. :slightly_frowning_face: In the meantime, could you do a preliminary review?

punzki:
@udalov I finished the refactoring of `ProgressionType`, much more satisfied with the code now. :slightly_smiling_face: Could you review?

I also added some more tests and found an issue on the non-IR backend (https://youtrack.jetbrains.com/issue/KT-38833) which is handled correctly on the IR backend.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
```diff
@@ -106,12 +71,18 @@ internal class ProgressionHeaderInfo(
     additionalNotEmptyCondition = additionalNotEmptyCondition
 ) {
 
+    @ExperimentalUnsignedTypes
```
udalov, line 74: I feel like annotating all files which use unsigned API with `@file:OptIn(ExperimentalUnsingedTypes::class)` would be a bit less intrusive

udalov:
Thanks for the refactoring and taking care of the redundant boxing/unboxing! I still see lots of stores/loads though, on the same test case. For example, here's an excerpt right after checking step > 0:
```
        81: istore        5
        83: iload_3
        84: ifle          92
```
where it looks like we could reuse the same slot instead of writing into new one.

I haven't investigated it closely, but from the IR it seems that we're introducing a new variable `checkedStep` which is either equal to `stepArg` or leads to an exception. It seems that we could simply check the condition and throw exception if needed, without introducing a new temporary variable?

Similarly later, we introduce a variable `newStep` which is either `checkedStep` or `-checkedStep`. Probably we could make `checkedStep` a `var` and flip its sign if needed instead.

Maybe we could get rid of some other temporary variables using this technique too.

Would you mind taking a look at whether this is feasible?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
punzki: I do prefer it that way too. Although... I don't know if it's an IDE bug, but if I use `@file:OptIn(...)` in `ProgressionType.kt`, IntelliJ gives still gives me a warning on the usages in `UIntProgressionType` and `ULongProgressionType` on `override fun DeclarationIrBuilder.minValueExpression() = irInt(UInt.MIN_VALUE.toInt())`. Very weird because I use `UInt.MIN_VALUE` in the super-constructor call. That is the reason I annotated the members instead. Anyway I changed it to use `@file`, we can choose to ignore this for now.

punzki:
> ...
> Maybe we could get rid of some other temporary variables using this technique too.
> 
> Would you mind taking a look at whether this is feasible?

@udalov I believe you are spot on for the reasons for these extra variables. Thanks for looking into it. I'm glad to take a look at this further. However, can I do this in a separate change? It is beyond the scope of the original intent of this PR. If you don't mind, let's try to land this PR and I'll do a thorough scrubbing of the temp variables separately.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderInfo.kt
punzki: IDE issue looks like https://youtrack.jetbrains.com/issue/KT-30264

udalov:
Thanks! I agree, let's do it in a separate change. Could you please rebase this one on master?

punzki:
> Thanks! I agree, let's do it in a separate change. Could you please rebase this one on master?

@udalov Done!

udalov:
Thank you! Merged.

PR #3275, opened 2020-04-08
JVM IR: Box inline classes in covariant overrides (KT-30419)

sfs:


sfs:
Rebased on master and fixed issues in bridge lowering. There are some rather ad-hoc workarounds there (see comments), but these are currently unavoidable, since there is no way in the IR to force boxing of an inline class type wrapping a non-nullable reference types.

dnpetrov:
Sorry for a delay, we had some inline class-related design work that might invalidate thios PR. Thankfully, we decided to keep it as is.

Ok, looks good to me.

FYI:
There are some ABI issues (most important: introducing a covariant override might silently break code at runtime), we take some more time to think about it. We keep inline classes experimental.
Currently it's Ok if JVM_IR will not implement suspend fun return value boxing optimization.

dnpetrov:
Pushed manually (74791495d726999405fd8c96722421163ddd6233)

PR #3276, opened 2020-04-09
KT-38139 False negative "Add suspend modifier" quickfix when suspend function is called in inline lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38139


fedochet:
@t-kameyama Thanks for your contribution!

PR #3277, opened 2020-04-09
[FIR] fix body generation for data class's copy method.

jsjeon:


compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -722,8 +731,243 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
                         }
                     }
 
+                    // TODO: copy should call <init> with parameters and return the copied instance.
```
jsjeon, line 734: This seems like a coverage hole. A simple test, like copy() and test ===, equals, hashCode, and toString would be good enough. I can do that in a follow-up PR.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -640,26 +639,40 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
...
+
+        private val copyName = Name.identifier("copy")
+        private val hashCodeName = Name.identifier("hashCode")
+        private val toStringName = Name.identifier("toString")
 
         fun generateMembers() {
             generateComponentFunctions()
             generateCopyFunction()
-            // TODO: equals, hashCode, toString
+            // TODO: need to check if these are overridable, e.g., user-given function, `final` one in hierarchy.
```
jsjeon, line 651: Three regressions indicate that such condition should be checked, and synthetic ones should not be added if, e.g., user defines their own hashCode, or super type has its own with `final` modifier. Is this a right place to check and/or class builder has enough information to check that?

mglukhikh:
I think that re-use `DataClassMemberGenerator` from PSI2IR could be better idea. As I can see now, binding context and other old FE-specific stuff is used only inside copy & component functions, and we already have them in FIR.  Could you look is it possible in practice or not? Yet I will put this PR on hold.

jsjeon:
> I think that re-use `DataClassMemberGenerator` from PSI2IR could be better idea. As I can see now, binding context and other old FE-specific stuff is used only inside copy & component functions, and we already have them in FIR. Could you look is it possible in practice or not? Yet I will put this PR on hold.

Here you are: https://github.com/JetBrains/kotlin/pull/3286 If that one looks like a better design/direction, we can keep discussing it there. Meanwhile, I'll reuse this PR and branch to deal with the incorrect `copy` generation in `fir` raw builder.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
jsjeon: Now I found how `psi2ir` avoids creating synthetic members if not necessary (or, in other words, how it creates those synthetic members only if needed): https://github.com/JetBrains/kotlin/blob/master/compiler/backend-common/src/org/jetbrains/kotlin/backend/common/FunctionsFromAnyGenerator.kt#L38 which goes to here: https://github.com/JetBrains/kotlin/blob/master/compiler/backend-common/src/org/jetbrains/kotlin/backend/common/CodegenUtil.kt#L139

Again, member scope from class descriptor and then `getContributedFunctions`, which are not available in wrapped descriptor. @mglukhikh any suggestions to do a similar checking? Note that I left a similar TODO in https://github.com/JetBrains/kotlin/pull/3286, resulting in very similar regression cases.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
jsjeon: It turns out there are couple BB tests that actually cover data class copy, followed by other synthetic member calls.

jsjeon:
@mglukhikh This PR is ready again for review. This time, it's about fixing raw fir generation for body of data class's `copy` method.

mglukhikh:
Merged manually 👍

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
jsjeon: I found my own answer: https://github.com/JetBrains/kotlin/pull/3294/commits/97ac4031fefa2aafe2ffd0fda779fdc11bf2915c#diff-2ed519bfe85d8b96c7fbba4ba954e24c

PR #3278, opened 2020-04-10
KT-31397 NONE_APPLICABLE quick fix suggests to re-create function if its argument expression contains unresolved operator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31397


klunnii:
Thanks! Merged as https://github.com/JetBrains/intellij-kotlin/commit/6d5365c9376ee61ee6beea83a8f267b15bbfb11a.

PR #3280, opened 2020-04-10
[FIR] fix argument ordering of Fir2Ir.

juan-chen:
Add an IrBlock with statements to assign arguments to temp variables if needed.


mglukhikh:
Merged manually 👍 

I added https://github.com/JetBrains/kotlin/commit/1ba160b3dedf298057ff8997041fe92127017c80 and https://github.com/JetBrains/kotlin/commit/1500131eb0fb575ef474114c257afcb5ff2e563a to deal with delegated call argument mapping in more regular way. This also fixes four more BB tests.

PR #3282, opened 2020-04-11
[FIR] Support diagnostic ABSTRACT_SUPER_CALL

lunakoly:
That's it. One more diagnostic.

### DIFF
--- a/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/FirHelpers.kt
+++ b/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/FirHelpers.kt
@@ -7,10 +7,17 @@ package org.jetbrains.kotlin.fir.analysis.checkers
 
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.fir.FirSession
+import org.jetbrains.kotlin.fir.FirSymbolOwner
+import org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext
 import org.jetbrains.kotlin.fir.declarations.FirClass
+import org.jetbrains.kotlin.fir.declarations.FirClassLikeDeclaration
 import org.jetbrains.kotlin.fir.declarations.FirRegularClass
+import org.jetbrains.kotlin.fir.expressions.FirQualifiedAccessExpression
+import org.jetbrains.kotlin.fir.references.FirResolvedNamedReference
+import org.jetbrains.kotlin.fir.resolve.firSymbolProvider
 import org.jetbrains.kotlin.fir.resolve.toSymbol
 import org.jetbrains.kotlin.fir.resolve.transformers.firClassLike
+import org.jetbrains.kotlin.fir.symbols.impl.FirCallableSymbol
 import org.jetbrains.kotlin.fir.symbols.impl.FirRegularClassSymbol
 import org.jetbrains.kotlin.fir.types.ConeClassLikeType
 import org.jetbrains.kotlin.fir.types.ConeKotlinType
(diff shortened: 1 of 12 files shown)


compiler/testData/diagnostics/tests/thisAndSuper/unqualifiedSuper/unqualifiedSuperWithAbstractMembers.fir.kt
```diff
@@ -25,7 +25,7 @@ class B : A(), I {
     override val y: Int = super.y
 
     override fun foo(): Int {
-        super.foo()
+        super.<!ABSTRACT_SUPER_CALL!>foo()<!>
```
mglukhikh, line 28: Please note that old front-end reports diagnostic on `foo` only, and here we have diagnostic on `foo()` with parentheses. It's worse, imagine some call like
```
super.foo(
    longArgument1,
    longArgument2,
    longArgument3,
    // ...
)
```

In IDEA you will have all these six lines underscored with red, it's too noisy. 

You may take `source` not from the whole call but from one of its children (I hope) to fix this.

compiler/testData/diagnostics/tests/thisAndSuper/unqualifiedSuper/unqualifiedSuperWithAbstractMembers.fir.kt
```diff
@@ -25,7 +25,7 @@ class B : A(), I {
...
         return super.x
```
mglukhikh, line 29: Note that old front-end reports the same diagnostic here, and you are not. You should check also qualified access expressions to achieve this.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/FirHelpers.kt
```diff
@@ -105,4 +116,43 @@ fun ConeKotlinType.toRegularClass(session: FirSession): FirRegularClass? {
  */
 fun FirTypeRef.toRegularClass(session: FirSession): FirRegularClass? {
     return safeAs<FirResolvedTypeRef>()?.type?.toRegularClass(session)
+}
+
+/**
+ * Returns the ClassLikeDeclaration where the function has been defined
+ * or null if no proper declaration has been found.
+ */
+fun FirFunctionCall.getClassLikeDeclaration(context: CheckerContext): FirClassLikeDeclaration<*>? {
```
mglukhikh, line 125: Here I'd prefer name like `getCalledFunctionContainingClass`

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirAbstractSuperCallChecker.kt
```diff
@@ -0,0 +1,54 @@
...
+            val function = functionCall.getSimpleFunctionDeclaration(context)
+                ?: return
+
+            val declaration = functionCall.getClassLikeDeclaration(context)
+                .safeAs<FirRegularClass>()
+                ?: return
+
+            if (
+                declaration.markedAs(ABSTRACT_KEYWORD) &&
+                function.markedAs(ABSTRACT_KEYWORD)
```
mglukhikh, line 42: It's not enough, you may have an interface function instead of an abstract function in an abstract class. The correct way here is to check function `modality`, it should be `ABSTRACT`.

compiler/testData/diagnostics/tests/thisAndSuper/unqualifiedSuper/unqualifiedSuperWithAbstractMembers.fir.kt
lunakoly: Got it. I'll fix my code

compiler/testData/diagnostics/tests/thisAndSuper/unqualifiedSuper/unqualifiedSuperWithAbstractMembers.fir.kt
lunakoly: You're right. I should've paid attention to it.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/FirHelpers.kt
```diff
@@ -105,4 +116,43 @@ fun ConeKotlinType.toRegularClass(session: FirSession): FirRegularClass? {
...
+        return context.session.firSymbolProvider.getClassLikeSymbolByFqName(classId)?.fir
+    }
+
+    return null
+}
+
+/**
+ * Returns FirSimpleFunction based on the given FirFunctionCall
+ */
+fun FirFunctionCall.getSimpleFunctionDeclaration(context: CheckerContext): FirSimpleFunction? {
```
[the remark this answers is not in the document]
lunakoly: First I was going to name it `toFir()` but decided such name would look inconsistent with `getClassLikeDeclaration()`. If `getCalledFunction()` sounds all right, I'll go with it for the next commit.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/FirHelpers.kt
lunakoly: All right, I'll change it

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirAbstractSuperCallChecker.kt
lunakoly: Ok. I didn't know such a thing existed.

lunakoly:
I made the changes to the code

mglukhikh:
Merged manually. Thank you! 👍

PR #3283, opened 2020-04-11
KT-20357 Added sortedBy() sample

Dmitry-Borodin:
sample for standard library

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -800,6 +800,14 @@ class Collections {
             assertPrints(people.joinToString(), "Sweyn Forkbeard, Ragnar Lodbrok, Bjorn Ironside")
         }
 
+        @Sample
+        fun sortedBy() {
+            val list = listOf("bbb", "aa", "cccc")
+            val sorted = list.sortedBy { it.length }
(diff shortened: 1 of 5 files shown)


libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -800,6 +800,14 @@ class Collections {
             assertPrints(people.joinToString(), "Sweyn Forkbeard, Ragnar Lodbrok, Bjorn Ironside")
         }
 
+        @Sample
+        fun sortedBy() {
+            val list = listOf("bbb", "aa", "cccc")
+            val sorted = list.sortedBy { it.length }
+
+            assertPrints(list, "[bbb, aa, cccc]")
+            assertPrints(sorted, "[aa, bbb, cccc]")
```
ilya-g, line 809: Currently it's unclear whether the result is sorted by length or alphabetically, because either way produces the same order. Perhaps, it'd be better to change input to `"aaaa", "cc", "bbb"`, so that the sorting criteria is clearly visible in the result?

Dmitry-Borodin: updated

ilya-g:
Thanks for the sample, merged!

PR #3284, opened 2020-04-11
added flatMap() map sample

Dmitry-Borodin:


libraries/stdlib/samples/test/samples/collections/maps.kt
```diff
@@ -353,6 +353,11 @@ class Maps {
             assertPrints(peopleToAge.map { it.value }, "[20, 21]")
         }
 
+        @Sample
+        fun flatMap() {
+            val map = mapOf("122" to 2, "3455" to 3)
+            assertPrints(map.flatMap { entry -> entry.key.take(entry.value).toList() }, "[1, 2, 3, 4, 5]")
```
ilya-g, line 359: It would be more succinct to destructure the entry in lambda parameters: `flatMap { (key, value) -> ... }`

Dmitry-Borodin: done

ilya-g:
Thanks for the sample, I've merged it manually as dd46d5ca5af4a5703d2674595a52917b925194bf.
I've decided to make `sampleClass` a local function for now.

PR #3285, opened 2020-04-12
KT-37746 "Redundant suspend modifier" should not be reported for functions with actual keyword

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37746


klunnii:
Accepted. Thank you!

PR #3286, opened 2020-04-13
IR: consolidate generation of synthetic members for data class.

jsjeon:
As per the suggestion from https://github.com/JetBrains/kotlin/pull/3277#issuecomment-611505077 this PR consolidates logics to generate synthetic members for data class (in particular, `equals`, `hashCode`, and `toString`) so that both of `psi2ir` path and `fir2ir` path can share them (and thus generate the same IR body for those synthetic members).


mglukhikh:
Merged manually 👍 

I added https://github.com/JetBrains/kotlin/commit/6827280521c5170e08a0b4b7fe1dfa7da5cd2851 and https://github.com/JetBrains/kotlin/commit/b6cf08d09762d682d4a711c280633c63e4a192dd to make code a bit shorter and more regular.

jsjeon:
> I added [6827280](https://github.com/JetBrains/kotlin/commit/6827280521c5170e08a0b4b7fe1dfa7da5cd2851) and [b6cf08d](https://github.com/JetBrains/kotlin/commit/b6cf08d09762d682d4a711c280633c63e4a192dd) to make code a bit shorter and more regular.

Thanks, those indeed made code simpler and cleaner!

PR #3291, opened 2020-04-14
JVM: fix remapping of `new T` inside regenerated copies of `T`

pyos:
In general, `InliningContext.findAnonymousTransformationInfo` was not reliable because it mapped each type to *some* info for that type, preferring ones with `shouldRegenerate == true` if those exist. Thus, it returned incorrect results if one type was regenerated multiple times, e.g. in a nested inlining context or because of a `finally` (which duplicates anonymous objects). The solution is to avoid a global map and attach the current transformation info directly to the current inlining context.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
@@ -73,6 +73,9 @@ class MethodInliner(
         if (!inliningContext.isInliningLambda) {
             inliningContext.root.state.globalInlineContext.recordTypeFromInlineFunction(info.oldClassName)
         }
+        if (info.shouldRegenerate(isSameModule)) {
+            inliningContext.recordRegeneratedAnonymousObject(info.oldClassName)
+        }
         transformations.add(info)
     }
 
@@ -171,7 +174,8 @@ class MethodInliner(
                     val childInliningContext = inliningContext.subInlineWithClassRegeneration(
                         inliningContext.nameGenerator,
                         currentTypeMapping,
-                        inlineCallSiteInfo
+                        inlineCallSiteInfo,
+                        transformationInfo!!
                     )
                     val transformer = transformationInfo!!.createTransformer(
                         childInliningContext, isSameModule, fakeContinuationName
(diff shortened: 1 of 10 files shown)


compiler/testData/codegen/boxInline/suspend/twiceRegeneratedAnonymousObject.kt
```diff
@@ -0,0 +1,27 @@
+// WITH_RUNTIME
+// WITH_COROUTINES
```
max-kammerer, line 2: Does problem relate to coroutines only?

max-kammerer: Is it poossible to construct general test without coroutines?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
```diff
@@ -305,51 +309,33 @@ class MethodInliner(
                     inlineOnlySmapSkipper?.markCallSiteLineNumber(remappingMethodAdapter)
                 } else if (isAnonymousConstructorCall(owner, name)) { //TODO add method
                     //TODO add proper message
-                    assert(transformationInfo is AnonymousObjectTransformationInfo) {
+                    var info = transformationInfo as? AnonymousObjectTransformationInfo ?: throw AssertionError(
                         "<init> call doesn't correspond to object transformation info for '$owner.$name': $transformationInfo"
+                    )
+                    val ownerIsParentClass = inlineCallSiteInfo.ownerClassName == info.oldClassName
```
max-kammerer, line 315: maybe smth like `isRecreatesItself`?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
```diff
@@ -305,51 +309,33 @@ class MethodInliner(
...
+                    if (ownerIsParentClass) {
+                        // inline fun f() -> new f$1 -> fun something() in class f$1 -> new f$1
+                        //                   ^-- fetch the info that was created for this instruction
+                        info = inliningContext.parent?.transformationInfo as? AnonymousObjectTransformationInfo
```
max-kammerer, line 319: @pyos As you have fresh look: could you point to  logic where compiler checks that object is already in process of regeneration and don't try regenerate it on second turn?

compiler/testData/codegen/boxInline/suspend/twiceRegeneratedAnonymousObject.kt
max-kammerer: Looks like the coroutines is only case for now

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/MethodInliner.kt
pyos: In `handleAnonymousObjectRegeneration`, `remapper.addMapping` is done before invoking the transformer. Thus, the nested `MethodInliner`'s `isAlreadyRegenerated` will return `true` for this type, causing `handleAnonymousObjectRegeneration` to ignore it. I think that's the only case where `isAlreadyRegenerated` returns true, actually...

max-kammerer:
Many thanks!

PR #3293, opened 2020-04-14
Native specific frontend checker for `@ThreadLocal`

LepilkinaElena:


native/frontend/src/org/jetbrains/kotlin/resolve/konan/diagnostics/NativeThreadLocalChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun check(declaration: KtDeclaration, descriptor: DeclarationDescriptor, context: DeclarationCheckerContext) {
+        check(threadLocalFqName, declaration, descriptor, context, ErrorsNative.INAPPLICABLE_THREAD_LOCAL) {
+            val isVariable = descriptor is VariableDescriptor
+            val hasBackingField = descriptor is PropertyDescriptor && descriptor.hasBackingField(context.trace.bindingContext)
+            val hasDelegate = declaration is KtProperty && declaration.delegate != null
+            (isVariable && (hasBackingField || hasDelegate)) ||
+                    (descriptor is ClassDescriptor && descriptor.kind == ClassKind.OBJECT)
+        }
+        check(threadLocalFqName, declaration, descriptor, context, ErrorsNative.INAPPLICABLE_THREAD_LOCAL_TOP_LEVEL) {
+            DescriptorUtils.isTopLevelDeclaration(descriptor) || descriptor is ClassDescriptor && descriptor.kind == ClassKind.OBJECT
```
homuroll, line 45: Doesn't `DescriptorUtils.isTopLevelDeclaration` return true for a top level property without field?

LepilkinaElena: Without backing field? It should be checked with previous part, it should be another type of error.

native/frontend/src/org/jetbrains/kotlin/resolve/konan/diagnostics/NativeThreadLocalChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.jetbrains.kotlin.resolve.checkers.DeclarationChecker
+import org.jetbrains.kotlin.resolve.checkers.DeclarationCheckerContext
+import org.jetbrains.kotlin.resolve.hasBackingField
+
+internal fun DeclarationChecker.check(
+    annotationFqName: FqName, declaration: KtDeclaration, descriptor: DeclarationDescriptor, context: DeclarationCheckerContext,
+    error: DiagnosticFactory0<KtElement>, successCondition: (DeclarationDescriptor) -> Boolean
+) {
+    if (successCondition(descriptor)) return
+    val threadLocalAnnotation = descriptor.annotations.findAnnotation(annotationFqName)
```
KirpichenkovPavel, line 26: [minor] It looks like variable name has become outdated

PR #3294, opened 2020-04-14
FIR2IR: generate synthetic members for data class only if needed.

jsjeon:
This PR resolves regressions from https://github.com/JetBrains/kotlin/pull/3286, which unconditionally added synthetic members for data classes.

Inspired by what `psi2ir` path did---sort of hierarchical member lookup with the given function descriptor, here we introduce similar utils to collect callable or contributed functions/declarations with configurable condition checks. All those newly added utils will use default configuration that unconditionally adds found declarations. For data class's synthetic members, we'll check function signatures, visibility, and modality of owner/function.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DataClassMembersGenerator.kt
```diff
@@ -67,37 +71,83 @@ class DataClassMembersGenerator(val components: Fir2IrComponents) {
...
-            val equalsFunction = createSyntheticIrFunction(
-                Name.identifier("equals"),
-                components.irBuiltIns.booleanType,
-            ).apply {
-                valueParameters = listOf(
-                    createSyntheticIrParameter(this, Name.identifier("other"), components.irBuiltIns.anyNType)
-                )
+            val result = mutableListOf<Name>()
+
+            val superTypesContributedFunctionsThatMatchDataClassSyntheticMemberSignatures =
```
mglukhikh, line 100: There are *three* different cases here:

- `equals` in data class itself, then we should not generate synthetic
- non-final `equals` in super class, then we *should* generate synthetic (please check this case!)
- final `equals` in super class, then we again *should not* generate synthetic

jsjeon: > There are _three_ different cases here:

Thank you for the clear categorization. This helps me figure out why my initial attempt looked so complicated. I've tried to collect contributed members in one go, but as you described below, ones in data class itself and ones in super class should be treated differently!

> * `equals` in data class itself, then we should not generate synthetic

Yeah, modality doesn't matter in this case. Note that unmuted tests in this PR whose file name is `alreadyDeclared.kt` cover this case.

> * non-final `equals` in super class, then we _should_ generate synthetic (please check this case!)

That's what I meant KT-6206, also covered here: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/dataClasses/nonTrivialMemberInSuperClass.kt

because of this, I've added somewhat strange condition, but now, after separating contributed members in this type v.s. super types, the logic looks much clearer (to me, at least).

> * final `equals` in super class, then we again _should not_ generate synthetic

The unmuted test in this PR (`nonTrivialFinalMemberInSuperClass.kt`) covers this: https://github.com/JetBrains/kotlin/pull/3294/commits/6e5162916e9e6dc25aac728a0d8aa36336e1328c#diff-2bd0650f999b623ec18e3a6806e27516

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DataClassMembersGenerator.kt
```diff
@@ -67,37 +69,93 @@ class DataClassMembersGenerator(val components: Fir2IrComponents) {
                 valueParameterDescriptor.bind(this)
             }
 
+        private val equalsName = Name.identifier("equals")
+        private val hashCodeName = Name.identifier("hashCode")
+        private val toStringName = Name.identifier("toString")
```
jsjeon, line 74: Nit (to myself): It might be better to move these into companion object, like I did at https://github.com/JetBrains/kotlin/pull/3300/commits/8ea1b876f445ec11456d1e962ffa93e0e003f8e2#diff-ad4af4ebe1770279317802646a8911e7

jsjeon: Done while rebasing: https://github.com/JetBrains/kotlin/pull/3294/commits/1b740726ccc2bea286c7c74e53483abd8ea4e603

jsjeon:
Rebased on top of https://github.com/JetBrains/kotlin/pull/3300. PTAL.

mglukhikh:
Merged with minor simplifications: https://github.com/JetBrains/kotlin/commit/09f9e767ede52fc47e55415fe1f06c2c2e0cc39a, https://github.com/JetBrains/kotlin/commit/28051080611191adb7605d3bdc3eb00b162ab49c.

👍

PR #3295, opened 2020-04-15
KT-38240 False positive redundant semicolon with `as` cast and `not` unary operator on next line

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38240


yanex:
Merged into `master`. Thanks!

PR #3297, opened 2020-04-15
Psi2Ir: Bind unbound type parameters

sfs:
Without NewInference we sometimes generate unbound type parameters, which end up breaking at random points during lowering. This PR fixes the problem for the JVM IR backend. I'm not sure if the same thing can happen in the JS_IR or NATIVE backends, though.


gogabr:
Thank you!

PR #3298, opened 2020-04-15
JVM IR: Use original descriptors for lambdas in JvmDeclarationOrigin

sfs:
There is an extension point which allows plugins to supply a custom `ClassBuilder`, which has access to descriptors through `JvmDeclarationOrigin`. One application I've seen is to mark lambdas used in inline functions as part of the public ABI of a Kotlin module. This works in the JVM backend by checking if the descriptor `isInlineOrContainedInInline`. This PR ensures that this particular application continues working with the JVM IR backend.

Unfortunately there's no good way to test this change, short of enabling the `DumpDeclarationTests` for the JVM IR backend (which, of course, is also not really a solution, since the tests are broken in more places than this).


dnpetrov:
Sorry for a long delay, I was occupied with 1.4-related work.
PR looks ok to me. I'll run tests and merge manually.

dnpetrov:
Pushed manually ( 86cd03952625c0c52f43b7869085611b1aa5b176 )

PR #3300, opened 2020-04-16
IR: more consolidation of synthetic member generation for data class.

jsjeon:
https://github.com/JetBrains/kotlin/pull/3286 covered consolidation of `equals`, `hashCode`, and `toString` for data class. This one covers `componentN` and `copy`.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
@@ -120,11 +120,24 @@ internal class ClassMemberGenerator(
                     irFunction.body = body
                 }
             } else if (irFunction !is IrConstructor) {
-                if (irFunction.origin == IrDeclarationOrigin.ENUM_CLASS_SPECIAL_MEMBER) {
-                    val kind = Fir2IrDeclarationStorage.ENUM_SYNTHETIC_NAMES.getValue(irFunction.name)
-                    irFunction.body = IrSyntheticBodyImpl(startOffset, endOffset, kind)
-                } else {
-                    irFunction.body = firFunction?.body?.let { visitor.convertToIrBlockBody(it) }
+                when {
+                    irFunction.origin == IrDeclarationOrigin.ENUM_CLASS_SPECIAL_MEMBER -> {
+                        val kind = Fir2IrDeclarationStorage.ENUM_SYNTHETIC_NAMES.getValue(irFunction.name)
+                        irFunction.body = IrSyntheticBodyImpl(startOffset, endOffset, kind)
+                    }
+                    irFunction.parent is IrClass && irFunction.parentAsClass.isData -> {
+                        when {
+                            DataClassMembersGenerator.isComponentN(irFunction) ->
+                                DataClassMembersGenerator(components).generateDataClassComponentBody(irFunction)
+                            DataClassMembersGenerator.isCopy(irFunction) ->
+                                DataClassMembersGenerator(components).generateDataClassCopyBody(irFunction)
+                            else ->
+                                irFunction.body = firFunction?.body?.let { visitor.convertToIrBlockBody(it) }
+                        }
+                    }
+                    else -> {
+                        irFunction.body = firFunction?.body?.let { visitor.convertToIrBlockBody(it) }
+                    }
                 }
             }
             if (irFunction !is IrConstructor || !irFunction.isPrimary) {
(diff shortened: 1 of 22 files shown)


compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -662,6 +660,7 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
         private fun generateComponentAccess(parameterSource: FirSourceElement?, firProperty: FirProperty) =
             buildQualifiedAccessExpression {
                 source = parameterSource
+                typeRef = firProperty.returnTypeRef
```
jsjeon, line 663: @demiurg906 pointed out that here I missed to specify the type. Actually, it was a pure refactoring, and it's _not_ been set before.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -732,27 +722,7 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
                             isVararg = false
                         }
                     }
-
-                    // TODO: Handle generic types.
-                    val initCallExpression = buildFunctionCall {
-                        argumentList = buildArgumentList {
-                            for ((ktParameter, firProperty) in zippedParameters) {
-                                val parameterSource = ktParameter?.toFirSourceElement()
-                                arguments += generateComponentAccess(parameterSource, firProperty)
```
jsjeon, line 741: @demiurg906 also pointed out that this part generates `copy` that looks like:
```
data class Foo(val x: Int, val y: Int) {
    fun copy(x: Int = this.x, y: Int = this.y) = Foo(this.x, this.y)
}
```
which should have been using `Foo(x, y)` instead, since default values for parameters are already set.

Instead of fixing it directly here, I think this entire refactoring/consolidation would make more sense, because now both `psi2ir` and `fir2ir` generate the same body for all the synthetic members for data class.

jsjeon:
This one may be conflict with https://github.com/JetBrains/kotlin/pull/3294 of course. :) Feel free to merge anything first, and just let me know; I can deal with merge conflict for the remaining one.

demiurg906:
Nice change, thank you

I ran modularized test and number of untouched types came back to previous value

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
jsjeon: > I ran modularized test and number of untouched types came back to previous value

Good to hear this, and I assume this addition affects that.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
```diff
@@ -120,11 +120,19 @@ internal class ClassMemberGenerator(
...
-                    val kind = Fir2IrDeclarationStorage.ENUM_SYNTHETIC_NAMES.getValue(irFunction.name)
-                    irFunction.body = IrSyntheticBodyImpl(startOffset, endOffset, kind)
-                } else {
-                    irFunction.body = firFunction?.body?.let { visitor.convertToIrBlockBody(it) }
+                when {
+                    irFunction.origin == IrDeclarationOrigin.ENUM_CLASS_SPECIAL_MEMBER -> {
+                        val kind = Fir2IrDeclarationStorage.ENUM_SYNTHETIC_NAMES.getValue(irFunction.name)
+                        irFunction.body = IrSyntheticBodyImpl(startOffset, endOffset, kind)
+                    }
+                    irFunction.parent is IrClass && irFunction.parentAsClass.isData &&
```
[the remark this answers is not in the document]
jsjeon: Right. Simplified.

jsjeon:
> I ran modularized test and number of untouched types came back to previous value

I usually ran `:compiler:test` if I touch `ir` (assuming `psi2ir` could be affected, and this PR does), and all modules under `:compiler:fir` if I touch `fir`. Please let me know if there are something else that I've missed so far.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
jsjeon: I worried this might indicate the test coverage hole, but it's actually covered by an unmuted test in this PR: https://github.com/JetBrains/kotlin/pull/3300/commits/ccabcb277de2bb7b7afa0c5a1918f343e2204043#diff-a872e759042c15f66b453f6b59f64b7d

demiurg906:
Merged manually

PR #3302, opened 2020-04-16
JVM_IR: avoid copying when optimizing ?:/?.

pyos:
Copying is slow, especially when the subtrees are large (with lambdas, etc.). It is also not always correct (e.g. discards reflection metadata), but all cases in which it isn't are probably rejected by the size check afterwards.


dnpetrov:
There were some related changes made by @gogabr . Could you please rebase on master?

pyos:
Done. Seems like he fixed one of the problems as well.

dnpetrov:

dnpetrov:
Pushed manually ( 887165a23b5d0899e7c0bff3888aaf1d28d632b6 )

PR #3304, opened 2020-04-17
KT-38261 Redundant 'let' call removal leaves ?. operator and makes code uncompilable

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38261


t-kameyama:
I think this issue will be resolved by [KT-37148](https://youtrack.jetbrains.com/issue/KT-37148).

PR #3306, opened 2020-04-17
[FIR] Use ieee754equals for floating-point equality operations

punzki:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -216,10 +213,21 @@ fun IrDeclaration.isInCurrentModule(): Boolean =
...
-    this is IrTypeOperatorCall &&
-            operator == IrTypeOperator.IMPLICIT_CAST &&
-            !this.argument.type.isSubtypeOf(type.makeNullable(), context.irBuiltIns)
+fun IrExpression.isSmartcastFromHigherThanNullable(context: JvmBackendContext): Boolean {
+    return when (this) {
+        is IrTypeOperatorCall -> operator == IrTypeOperator.IMPLICIT_CAST && !argument.type.isSubtypeOf(
+            type.makeNullable(),
+            context.irBuiltIns
+        )
+        is IrGetValue -> {
```
punzki, line 222: @dnpetrov @udalov @kandersen Could you take a look at this expanded logic for smart cast detection? Adding you here since you reviewed this when the code was added in https://github.com/JetBrains/kotlin/pull/2751/files#diff-b2724d380aa7f929ef1fc06c26653277

punzki: For the following code:
```kt
if (plusZero is Double) {
  when (plusZero) {
    -0.0 -> return "fail"
// ...
```

In FIR, the `IMPLICIT_CAST` is added on initializing the subject variable:
```
                BLOCK type=kotlin.Unit origin=WHEN
                  VAR IR_TEMPORARY_VARIABLE name:tmp0_subject type:kotlin.Double [val]
                    TYPE_OP type=kotlin.Double origin=IMPLICIT_CAST typeOperand=kotlin.Double
                      GET_VAR 'val plusZero: kotlin.Any [val] declared in <root>.WhenKt.box' type=kotlin.Any origin=null
                  WHEN type=kotlin.Unit origin=WHEN
                    BRANCH
                      if: CALL 'public final fun ieee754equals (arg0: kotlin.Double?, arg1: kotlin.Double?): kotlin.Boolean declared in kotlin.internal.ir' type=kotlin.Boolean origin=EQEQ
                        arg0: GET_VAR 'val tmp0_subject: kotlin.Double [val] declared in <root>.WhenKt.box' type=kotlin.Double origin=null
                        arg1: CONST Double type=kotlin.Double value=-0.0
                      then: RETURN type=kotlin.Nothing from='public final fun box (): kotlin.String declared in <root>.WhenKt'
                        CONST String type=kotlin.String value="fail"
```

...whereas with the current front-end, the `IMPLICIT_CAST` is added in the call (parameter/target type is different from value type):

```
              BLOCK type=kotlin.Unit origin=WHEN
                VAR IR_TEMPORARY_VARIABLE name:tmp0_subject type:kotlin.Any [val]
                  GET_VAR 'val plusZero: kotlin.Any [val] declared in <root>.box' type=kotlin.Any origin=null
                WHEN type=kotlin.Unit origin=WHEN
                  BRANCH
                    if: CALL 'public final fun ieee754equals (arg0: kotlin.Double?, arg1: kotlin.Double?): kotlin.Boolean declared in kotlin.internal.ir' type=kotlin.Boolean origin=EQEQ
                      arg0: TYPE_OP type=kotlin.Double origin=IMPLICIT_CAST typeOperand=kotlin.Double
                        GET_VAR 'val tmp0_subject: kotlin.Any [val] declared in <root>.box' type=kotlin.Any origin=null
                      arg1: CONST Double type=kotlin.Double value=-0.0
                    then: BLOCK type=kotlin.Unit origin=null
                      RETURN type=kotlin.Nothing from='public final fun box (): kotlin.String declared in <root>'
                        CONST String type=kotlin.String value="fail"
```

It was simpler to expand the smart cast detection logic (to follow variables) rather than move the `IMPLICIT_CAST`, though I am slightly concerned I missed a possible case where it is incorrect. I do think the approach in FIR makes more sense and results in fewer `IMPLICIT_CAST`s.

kandersen: Looks good to me, Mark!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -216,10 +213,21 @@ fun IrDeclaration.isInCurrentModule(): Boolean =
 // This is needed to pinpoint exceptional treatment of IEEE754 floating point comparisons, where proper IEEE
 // comparisons are used "if values are statically known to be of primitive numeric types", taken to mean as
 // "not learned through smartcasting".
-fun IrExpression.isSmartcastFromHigherThanNullable(context: JvmBackendContext) =
-    this is IrTypeOperatorCall &&
-            operator == IrTypeOperator.IMPLICIT_CAST &&
-            !this.argument.type.isSubtypeOf(type.makeNullable(), context.irBuiltIns)
+fun IrExpression.isSmartcastFromHigherThanNullable(context: JvmBackendContext): Boolean {
+    return when (this) {
+        is IrTypeOperatorCall -> operator == IrTypeOperator.IMPLICIT_CAST && !argument.type.isSubtypeOf(
```
kandersen, line 218: Nit: I'd break this differently.
```kotlin
is IrTypeOperatorCall -> operator == IrTypeOperator.IMPLICIT_CAST &&
    !argument.type.isSubtypeOf(type.makeNullable(), context.irBuiltIns)
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
udalov: To be honest, I'd very much prefer to have the same IR generated by fir2ir and psi2ir, to avoid dealing with both of the possibilities in the backend (not only in this utility function, but in a broader sense). Maybe it makes more sense to correct the psi2ir behavior here, so that it only produces the implicit cast on the variable?

punzki: @udalov If we modified psi2ir behavior so that the implicit cast is on the variable, we'd still need to make this change to expand the logic for smart cast detection.

udalov: You're right. Actually I'd still prefer either that (i.e. both change psi2ir, and handle `IrGetField` here), or only changing fir2ir to the "suboptimal" behavior, just to minimize mismatch between frontends, which could potentially lead to problems elsewhere in the backend(s).

punzki: I investigated whether it is possible in psi2ir to add the implicit cast on the variable initializer, and it to me it seems the only way to achieve this is to get data flow information from the front end. Old backend does it that way (ref: [here](https://github.com/JetBrains/kotlin/blob/f42d9eefa1aab68b2d70faf6daad7874e8a09fc5/compiler/backend/src/org/jetbrains/kotlin/codegen/ExpressionCodegen.java#L4123) and [here](https://github.com/JetBrains/kotlin/blob/65244b4bea81f737466618927d4f3afe339cad0d/compiler/backend/src/org/jetbrains/kotlin/codegen/ieee754.kt#L37)). I can't think of another way to get the smartcast type at the variable declaration. Any other ideas? I don't think we want to go that route for the sake of matching the generated IR.

@mglukhikh What do you think of modifying fir2ir behavior? (see comments above) It will lead to more `IMPLICIT_CAST`s.

mglukhikh: At this moment FIR2IR converter creates IR IMPLICIT_CAST iff `FirExpressionWithSmartcast` is visited, so it just repeats structure of resolved frontend IR tree. It (probably) means that to achieve the same behavior with PSI2IR we will have to change FIR resolve logic, and I do not like this idea too much, especially when it leads to sub-optimal behavior as you said before. Better (probably) solution would be to leave FIR resolve as is and change converter behavior, but I'm not sure we can implement this easily.

cc @demiurg906

punzki: I tried over the last few days to essentially duplicate the logic in `InsertImplicitCasts` from PSI2IR, which is what we'd need to move the `IMPLICIT_CAST`s away from the variable declaration and instead at the receivers and value arguments. I haven't come close to getting all the tests to pass, and I'm not confident that I can fully implement this correctly. I know @dnpetrov has been working on this over the years and I have been trying for just a few days. :slightly_smiling_face: 

@udalov I understand your concern about minimizing mismatch between front-ends. Can we re-visit this if/when we do find the mismatch causes issues on the back-end? I believe that the current behavior in FIR (to add the `IMPLICIT_CAST` as soon as we know an expression is smartcast) is more correct anyway.

udalov: Yeah, makes sense, let's postpone it. Thanks for the research. I suppose we'll still encounter this mismatch in the future if we are going to chase the goal of 100% equality of IR constructed by psi2ir and fir2ir, for example on `irText` tests, and it could be fixed then.

mglukhikh:
Merged manually 👍

PR #3308, opened 2020-04-17
Fix "X has several compatible actual declarations" errors

altsvetkov:
See https://youtrack.jetbrains.com/issue/KT-35957


altsvetkov:
Merged

PR #3309, opened 2020-04-18
FIR: record 'suspend' function modifier and resolve suspend lambda properly.

jsjeon:
The motivating example is:
```
fun build(c: suspend () -> Unit) {
  c.createCoroutine(...)
}

fun box() {
  build {
    ... // lambda
  }
}
```

Without recording type modifier `suspend`, the type for the value parameter of `build` was `kotlin.Function0<Unit>`, not `kotlin.coroutine.SuspendFunction0<Unit>`. Since it makes sense to apply `suspend` to function type only (thanks to @punzki), in the 1st commit, which introduces `isSuspend` to `typeRef` itself and all subtypes, we set default value `false` for all cases, except for function type reference. With that, call sites that build function type references are only affected.

Besides fixing the value parameter type, we also need to adjust the type of lambda, an actual argument. After call resolution, from the expected argument type, we can learn that it should be a `suspend` lambda (of, again, `kotlin.coroutine.SuspendFunction0<Unit>` type), not regular lambda (of `kotlin.Function0<Unit>`). The 2nd commit deals with correcting lambda's type and corresponding `isSuspend` field in anonymous function.


jsjeon:
Hm, I'm not sure how to create another PR on top of an ongoing PR. :\ Anyhow, there is an additional commit: https://github.com/google/kotlin/pull/2/commits/f91ee87783b204eae9abfa100dfec871f32271c2 with some description: https://github.com/google/kotlin/pull/2 (Or I can create a PR later after this one is landed).

mglukhikh:
It's was a great shot. I added some commits fixing side behavior, e.g.

https://github.com/JetBrains/kotlin/commit/aada7bfe510bca0d0b274d47bc5a453b0799ba1e
https://github.com/JetBrains/kotlin/commit/c5dea02074d049c62c01dd287adf4b4ca2740dc2
https://github.com/JetBrains/kotlin/commit/fe02c2bab318878855859e4bd7024c522d6d22d9

BTW last commit about "return Unit" in lambda does not fix `overrideDefaultArgument.kt`. Coroutine synchronization does not work there for some reason. I had to hardly ignore this situation: https://github.com/JetBrains/kotlin/commit/4bb6410ce3ba5086ccdd23d4187261c0acba50a7. https://youtrack.jetbrains.com/issue/KT-38416 created about this problem, if you want to solve it -- welcome.

Thanks for this PR!

mglukhikh:
Could you please create separate PR about https://github.com/google/kotlin/commit/f91ee87783b204eae9abfa100dfec871f32271c2? It looks like not very related...

jsjeon:
Thank you for fixing missed parts on handling `suspend` functions in different places. Also, I was about to dig into an incorrect translation of an empty lambda, and glad to see that you already fixed it: https://github.com/JetBrains/kotlin/commit/fe02c2bab318878855859e4bd7024c522d6d22d9 :)

Also, thank you for filing an issue explicitly. Let me take a look at it.

> Could you please create separate PR about [google@f91ee87](https://github.com/google/kotlin/commit/f91ee87783b204eae9abfa100dfec871f32271c2)? It looks like not very related...

Will do, thanks.

PR #3310, opened 2020-04-19
KT-38349 Invalid suggestion to fold to elvis when having a var-variable

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38349


yanex:
The code itself seems OK for me.
However, I would rather think that changing the result type is closer to what the user wants:

```
fun test(foo: Int?, bar: Int): Int {
    var i = foo
    if (i == null) {
        return bar
    }
    return i
}
```

to

```
fun test(foo: Int?, bar: Int): Int {
    var i = foo ?: return bar
    return i
}
```

t-kameyama:
@yanex PTAL

t-kameyama:
915dc6ce

PR #3311, opened 2020-04-19
[FIR] Support diagnostic QUALIFIED_SUPERTYPE_EXTENDED_BY_OTHER_SUPERTYPE

lunakoly:
A possible implementation of a checker for `QUALIFIED_SUPERTYPE_EXTENDED_BY_OTHER_SUPERTYPE`.


compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.kotlin.fir.expressions.FirQualifiedAccessExpression
+import org.jetbrains.kotlin.fir.references.FirSuperReference
+import org.jetbrains.kotlin.fir.resolve.transformers.firClassLike
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+object FirQualifiedSupertypeExtendedByOtherSupertypeChecker : FirQualifiedAccessChecker() {
+    override fun check(functionCall: FirQualifiedAccessExpression, context: CheckerContext, reporter: DiagnosticReporter) {
+        // require to be called over a super reference
+        val superReference = functionCall.calleeReference.safeAs<FirSuperReference>()
+            ?: return
```
demiurg906, line 26: It's better to write `?: return` on same line as expression

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+
+        val surroundingType = context.findClosestClass()
+            ?: return
+
+        // how many supertypes of the `surroundingType`
+        // have `explicitType` as their supertype
+        var count = 0
+
+        for (it in surroundingType.superTypeRefs) {
+            val that = it.firClassLike(context.session).safeAs<FirClass<*>>()
```
demiurg906, line 39: There can be a typealias in llist of supertype ref that is not a `FirClass`

Consider following example:
```
interface IBase<T> {
    fun foo() {}
    fun bar() {}
}

typealias AliasedIBase = IBase<String>

interface IDerived<T> : IBase<T> {
    override fun foo() {}
    fun qux() {}
}

class Test : IDerived<String>, AliasedIBase {
    fun test() {
        super<IBase>.foo()
        super<IBase>.bar()
        super<AliasedIBase>.foo() // QUALIFIED_SUPERTYPE_EXTENDED_BY_OTHER_SUPERTYPE
        super<AliasedIBase>.bar() // QUALIFIED_SUPERTYPE_EXTENDED_BY_OTHER_SUPERTYPE
    }
}
```

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+                ?: return
+
+            if (
+                explicitType == that ||
```
demiurg906, line 43: This is short expression so it's better to write it in one line

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
```diff
@@ -0,0 +1,84 @@
...
+            reporter.report(superReference.superTypeRef.source)
+        }
+    }
+
+    /**
+     * Returns the closest to the end of context.containingDeclarations
+     * item like FirRegularClass or FirAnonymousObject
+     * or null if no such item could be found.
+     *
+     * TODO: This function repeats the one in FirNotASupertypeChecker
```
demiurg906, line 64: You can move this extension to some `ContextUtils.kt` to avoid duplicating code

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
lunakoly: I'm afraid I can't agree with you. 

First, due to the fact that Intellij IDEA doesn't warn me about bad formatting, I suppose that the style I use can't be considered unconventional. 

Second, the idea of this style is to allow the reader to easily see that this is a "try or give up" assignment. Writing the `?: return` on the next line helps to know that this assignment may fail even before the one who reads gets to the end of the line. So if you are not interested in the way the actual value is obtained you just see that it's some `superReference` that we deal with and if it fails we leave.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
lunakoly: You're right, thanks. I'll fix my code

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
lunakoly: Agreed. Personally for me it makes no difference but if someone finds single-line style in this case easier to read I'm ready to refactor my code.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
lunakoly: Thanks or the advice I'll definitely follow it. Although I'm not quite certain about the name of the function. I doubt `findClosestClassOrAnonymousObject()` is a good one (btw, there's already the `findClosest<T>()` inside of the context class).

lunakoly: I updated the code and put the function `findClosestClassOrObject()` into `FirHelpers.kt` - the file I use to store some helper functions for checkers. It may be worth splitting it into separate ones like `ContextUtils.kt` and other, but I want to do it later.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/call/FirQualifiedSupertypeExtendedByOtherSupertypeChecker.kt
lunakoly: I consulted play.kotlinlang.org and it sais that `super<IBase>.bar()` and `super<IBase>.foo()` should also throw `QUALIFIED_SUPERTYPE_EXTENDED_BY_OTHER_SUPERTYPE` so that's the way I made it in my code.

demiurg906:
Looks fine, thank you. I ran all tests on buildserver and when it's done I'll push your changes

demiurg906:
Merged manually

PR #3313, opened 2020-04-20
JVM IR: Fix correspondingPropertySymbols

sfs:
There are several work arounds in `MethodSignatureMapper` for functions with missing or broken `correspondingPropertySymbols`. This PR fixes `correspondingPropertySymbols` throughout the compiler and removes the work arounds.

I've also added a check for broken properties to `IrValidator` and enabled it on the JVM IR backend.


udalov:
It looks like your changes include the work done by @madsager in #3256. Would it be OK for you to rebase on top of that PR, so I could merge that one first, and continue with your changes?

sfs:
Wow. I wasn't aware of Mads' PR and we ended up duplicating work.

I'll go over it and rebase.

sfs:
Rebased on master

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -99,7 +99,7 @@ private val lateinitUsageLoweringPhase = makeIrFilePhase(
     description = "Insert checks for lateinit field references"
 )
 
-private val propertiesPhase = makeIrFilePhase(
+internal val propertiesPhase = makeIrFilePhase(
     ::JvmPropertiesLowering,
     name = "Properties",
     description = "Move fields and accessors for properties to their classes, and create synthetic methods for property annotations",
```
udalov, line 105: Please adapt this phase description to the changes

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
```diff
@@ -176,33 +176,41 @@ class MemoizedInlineClassReplacements {
...
+        copyTypeParameters(function.allTypeParameters)
+        metadata = function.metadata
+        function.safeAs<IrFunctionBase<*>>()?.metadata = null
+
+        if (function is IrSimpleFunction) {
+            val propertySymbol = function.correspondingPropertySymbol
+            correspondingPropertySymbol = propertySymbol
+            if (propertySymbol != null) {
+                when (function) {
+                    propertySymbol.owner.getter -> propertySymbol.owner.getter = this
```
udalov, line 203: Suppose we have a property accessor declaration with inline class types in the signature in `decl.kt`, and its usage in `use.kt`, and we're lowering `use.kt` first. Looks like we'll reach this place when we encounter the call site, change the `IrProperty`'s accessor and this would probably mess up the subsequent run of lowerings on `decl.kt`, in particular `JvmPropertiesLowering`, because we'll observe the already modified declaration there. Or at least it would make the lowerings' behavior non-trivially dependent on the file order.

I wonder if we can do something about it? In general it seems that modifying any declaration going through the symbol link (`correspondingPropertySymbol`, call's `symbol`, `overriddenSymbols` etc.) is risky because it can be declared in another file.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
```diff
@@ -311,3 +312,21 @@ fun firstSuperMethodFromKotlin(
...
     }
 }
+
+// MethodSignatureMapper uses the corresponding property of a function to determine correct names
+// for property accessors.
+fun IrSimpleFunction.copyCorrespondingPropertyFrom(source: IrSimpleFunction) {
+    val property = source.correspondingPropertySymbol?.owner ?: return
+    val target = this
+
+    correspondingPropertySymbol = buildProperty {
```
udalov, line 322: Can we also copy the attributes from the property, as the previous code did, such as visibility, modality, modifiers, etc.?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/CheckIrElementVisitor.kt
```diff
@@ -323,7 +331,7 @@ class CheckIrElementVisitor(
 
     override fun visitFunctionReference(expression: IrFunctionReference) {
         super.visitFunctionReference(expression)
-
+        expression.checkFunction(expression.symbol.owner)
```
udalov, line 334: Is this needed though? Looks like either a function is from sources, in which case it'll be validated in `visitFunction`, or from binaries, in which case it's not modified and needs no validation.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
udalov: Probably because of this issue, compilation of kotlin-stdlib with JVM IR fails:
```
  Caused by: java.lang.RuntimeException: Exception while generating code for:
  FUN STATIC_INLINE_CLASS_REPLACEMENT name:getSize-impl visibility:public modality:OPEN <> (arg0:kotlin.UByteArray) returnType:kotlin.Int
    correspondingProperty: PROPERTY name:size visibility:public modality:OPEN [val]
    overridden:
      public abstract fun <get-size> (): kotlin.Int declared in kotlin.collections.Collection
    VALUE_PARAMETER MOVED_DISPATCH_RECEIVER name:arg0 index:0 type:kotlin.UByteArray
    at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.generate(FunctionCodegen.kt:50)
    at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generateMethodNode(ClassCodegen.kt:261)
    at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generateMethod(ClassCodegen.kt:278)
    at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generate(ClassCodegen.kt:119)
    at org.jetbrains.kotlin.backend.jvm.JvmBackendFacade.doGenerateFilesInternal$backend_jvm(JvmBackendFacade.kt:120)
    ... 37 more
  Caused by: java.lang.NullPointerException
    at org.jetbrains.kotlin.backend.jvm.codegen.ExpressionCodegen.generate(ExpressionCodegen.kt:201)
    at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.doGenerate(FunctionCodegen.kt:105)
    at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.generate(FunctionCodegen.kt:48)
    ... 41 more
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
sfs: Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
sfs: You're right, that was a terrible idea. I've changed the code to create a new property instead and link it up properly with the replaced getter/setter. I've also added a test for the case you mentioned (which did indeed fail with the previous code).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrUtils.kt
sfs: Done, since this is a bit cleaner. But note that the property in this case is only a placeholder to mark a function as a getter/setter. We're not necessarily linking up the getters/setters correctly.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/CheckIrElementVisitor.kt
sfs: > or from binaries, in which case it's not modified and needs no validation.

Unfortunately there are cases where we have to modify functions coming from binaries, e.g., when changing the names of functions with inline class parameters, or when adding default argument stubs to binary classes. Similarly, psi2ir has a failure case where we end up adding a stub function without any type information (when symbol resolution fails with the default case). This is only caught if we verify every function call.

sfs:
Implemented the various fixes discussed above and also changed the code for property builders to take the original descriptor into account. For inline property getters/setters, the inliner looks at the descriptor of the corresponding property to determine if a declaration should be inlined from source or binary dependencies and so we have to preserve the `containerSource` when changing the `correspondingPropertySymbol`.

This also means that we could probably break inlining of property getters/setters from dependencies without this PR, in the cases where we loose the `correspondingPropertySymbol`.

sfs:
Rebased on master and changed `JvmInlineClassLowering` to take [0acaede](https://github.com/JetBrains/kotlin/commit/0acaedef92d6f90625cf608c11b2349fcf90a635) into account, see the comment at [JvmInlineClassLowering.kt#L179](https://github.com/JetBrains/kotlin/blob/c9be0fccb19318ac1d2295a5768dfc828c0fe8a0/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt#L179).

udalov:
Looks good, thank you!

PR #3315, opened 2020-04-20
FIR2IR: cache more type parameters from FirTypeParameterRefsOwner.

jsjeon:
In addition to literally `typeParameters` field in `FirTypeParameterRefsOwner`, there are more places to retrieve/cache type parameters in it: `returnTypeRef` if it's callable, `returnTypeRef` of value parameters if it's a function. Otherwise, we're seeing unknown type parameter error when visiting, e.g., `inline suspend fun <T, R> Flow<T>.fold(init: R, crossinline operation: suspend (acc: R, value: T) → B)`


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -89,6 +91,32 @@ class Fir2IrClassifierStorage(
                     ?: createIrTypeParameterWithoutBounds(original, index, context)
             }
         }
+        if (owner is FirCallableDeclaration<*>) {
+            val typeParameterInReturnType = owner.returnTypeRef.retrieveTypeParameter()
+            if (typeParameterInReturnType != null) {
+                getCachedIrTypeParameter(typeParameterInReturnType)
+                    ?: createIrTypeParameterWithoutBounds(typeParameterInReturnType, index = 0)
```
mglukhikh, line 98: This is strange place, as well as the next one. Normally we should not create type parameter when we don't know its index. This will anyway work incorrectly if a corresponding declaration has more than one type parameter.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -89,6 +91,32 @@ class Fir2IrClassifierStorage(
...
+            }
+        }
+        if (owner is FirFunction<*>)
+            owner.valueParameters.forEachIndexed { index, valueParameter ->
+                val typeParameterInValueParameter = valueParameter.returnTypeRef.retrieveTypeParameter()
+                if (typeParameterInValueParameter != null) {
+                    getCachedIrTypeParameter(typeParameterInValueParameter)
+                        ?: createIrTypeParameterWithoutBounds(typeParameterInValueParameter, index)
```
mglukhikh, line 106: `index` is incorrect here, it's value parameter index, but should be type parameter index.

mglukhikh:
After some analysis I fixed the problem this way: https://github.com/JetBrains/kotlin/commit/39e1635c411bc1e5e270d2925ec2fd1d5b51aaec. Fake overrides may re-use type parameters of original function, so we have to pre-cache them. We should always try to pre-cache type parameter before it can be referenced, and this thing can be performed correctly only from declaration header (it's the only place when type parameter index is known).

Thank you.

jsjeon:
Aha. Thank you for the detailed explanations!

PR #3317, opened 2020-04-21
KT-37486 Don't store instances from stream debugger plugin in companions

bibaev:


bibaev:
It looks like the fix has been merged manually by @yanex. Thank you! I'll close the pull request.

PR #3318, opened 2020-04-21
KT-38282 False positive "Remove redundant spread operator" inspection with array as class property or fun argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38282

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantSpreadOperatorInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantSpreadOperatorInspection.kt
@@ -13,11 +13,15 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElementVisitor
 import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.analysis.analyzeAsReplacement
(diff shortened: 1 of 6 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantSpreadOperatorInspection.kt
```diff
@@ -31,6 +35,16 @@ class RemoveRedundantSpreadOperatorInspection : AbstractKotlinInspection() {
...
                     is KtCallExpression -> {
                         if (!argumentExpression.isArrayOfMethod()) return
+                        if (argumentExpression.valueArguments.isEmpty()) {
+                            val call = argument.getStrictParentOfType<KtCallExpression>()
+                            if (call != null) {
+                                val index = call.valueArguments.indexOfFirst { it == argument }
+                                val newCall = call.copy() as KtCallExpression
+                                newCall.valueArgumentList?.removeArgument(index)
+                                val newContext = newCall.analyzeAsReplacement(call, call.analyze(BodyResolveMode.PARTIAL))
+                                if (newCall.getResolvedCall(newContext)?.resultingDescriptor == null) return
```
dimonchik0036, line 45: This check doesn't work for:
```kotlin
fun foo(vararg args: Int) = println(1)
fun foo() = println(2)
fun test() {
    foo(*intArrayOf())
} 
```
I think you should compare result descriptor with previous value.

PR #3323, opened 2020-04-23
JVM IR: Optimize field initializers in secondary constructors

sfs:
Don't generate code to initialize fields to their default values in secondary constructors as well as in the primary constructor. This optimization is done by the JVM backend and has some observable effects.

Note that this currently fails on both JS backends.

CC @pyos

### DIFF
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -26716,6 +26716,11 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
             runTest("compiler/testData/codegen/box/secondaryConstructors/enums.kt");
         }
 
+        @TestMetadata("fieldInitializerOptimization.kt")
+        public void testFieldInitializerOptimization() throws Exception {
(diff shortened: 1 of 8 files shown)


max-kammerer:
Not sure that it should work in same way in JS and Native. @bashor @SvyatoslavScherbina  it's interesting case of different semantics on different platforms

max-kammerer:
Thank you!

SvyatoslavScherbina:
> it's interesting case of different semantics on different platforms

IIUC, current Kotlin/JVM behaviour is just weird. It can provoke different initialization orders depending on whether default or non-default value is used.

sfs:
> IIUC, current Kotlin/JVM behaviour is just weird. It can provoke different initialization orders depending on whether default or non-default value is used.

Agreed. My understanding is that this is only necessary for J2K.

The problem is that in Java there is a difference between explicitly initialized and default initialized fields. E.g.,
```java
class A { int x = 0; int y; }
```
produces code to initialize `x` in the constructor, while no code is produced to initialize `y`.

Meanwhile, Kotlin only supports explicitly initialized fields.

However, default initialized fields are very common in Java, so in order to make J2K work in more cases we optimize away default initialized fields.

---

That said, if we want to avoid hard to debug corner cases in multiplatform projects we should settle on a single implementation strategy and changing the current behavior would be a non-backwards compatible change.

udalov:
Here's the related issue: https://youtrack.jetbrains.com/issue/KT-15642

PR #3324, opened 2020-04-23
[JVM_IR] Use ifne and ifeq for integer zero comparisons.

madsager:


compiler/testData/codegen/bytecodeText/when/whenZero.kt
```diff
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
 // TODO KT-36841 Generate integer comparison with 0 instructions (IFNE, etc) when comparing integers with 0 in JVM_IR
```
udalov, line 1:
```suggestion
```

madsager: Whoops, thanks Alex. :)

max-kammerer:
Thank you!

PR #3325, opened 2020-04-23
JVM: move synthetic line number generation to the inliner

pyos:
This makes the behavior of JVM and JVM_IR the same without having to deal with PSI nodes.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt
@@ -536,10 +536,24 @@ fun isFakeLocalVariableForInline(name: String): Boolean {
 internal fun isThis0(name: String): Boolean = AsmUtil.CAPTURED_THIS_FIELD == name
 
 class InlineOnlySmapSkipper(codegen: BaseExpressionCodegen) {
-
     private val callLineNumber = codegen.lastLineNumber
 
-    fun markCallSiteLineNumber(mv: MethodVisitor) {
+    fun onInlineLambdaStart(mv: MethodVisitor, info: LambdaInfo) {
+        val firstLine = info.node.node.instructions.asSequence().mapNotNull { it as? LineNumberNode }.firstOrNull()?.line ?: -1
+        if (callLineNumber >= 0 && firstLine == callLineNumber) {
+            // We want the debugger to be able to break both on the inline call itself, plus on each
+            // invocation of the inline lambda passed to it. For that to happen there needs to be at least
+            // one different line number in between those breakpoints for the VM to emit a locatable event.
+            // @InlineOnly functions, however, contain no line numbers, so if the lambda is single-line,
+            // the entire call will "meld" into a single region. To break it up, we insert a different line
+            // number that is remapped by the SMAP to a line that does not exist.
+            val label = Label()
+            mv.visitLabel(label)
+            mv.visitLineNumber(JvmAbi.LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER, label)
+        }
+    }
+
+    fun onInlineLambdaEnd(mv: MethodVisitor) {
         if (callLineNumber >= 0) {
             val label = Label()
             mv.visitLabel(label)
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java
@@ -78,7 +78,6 @@ import static org.jetbrains.kotlin.codegen.state.KotlinTypeMapper.isAccessor;
 import static org.jetbrains.kotlin.descriptors.CallableMemberDescriptor.Kind.DECLARATION;
 import static org.jetbrains.kotlin.descriptors.CallableMemberDescriptor.Kind.DELEGATION;
 import static org.jetbrains.kotlin.descriptors.ModalityKt.isOverridable;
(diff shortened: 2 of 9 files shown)


compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt
```diff
@@ -536,10 +536,24 @@ fun isFakeLocalVariableForInline(name: String): Boolean {
 internal fun isThis0(name: String): Boolean = AsmUtil.CAPTURED_THIS_FIELD == name
 
 class InlineOnlySmapSkipper(codegen: BaseExpressionCodegen) {
-
     private val callLineNumber = codegen.lastLineNumber
 
-    fun markCallSiteLineNumber(mv: MethodVisitor) {
+    fun onInlineLambdaStart(mv: MethodVisitor, info: LambdaInfo) {
+        val firstLine = info.node.node.instructions.asSequence().mapNotNull { it as? LineNumberNode }.firstOrNull()?.line ?: -1
+        if (callLineNumber >= 0 && firstLine == callLineNumber) {
```
max-kammerer, line 543: What would be in case when lambda is inline function call(maybe without arguments)?
E.g.
```inlineOnlyExt { anotherInlineCall { } }```
and 
```inlineOnlyExt { inlineOnlyInlineCall {...} }```

Comment (pyos) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: The first line number in those lambdas is for the inner call itself, so 65100 would be inserted.

Comment (max-kammerer) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: And if `anotherInlineCall` is not inlineOnly?

Comment (max-kammerer) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: > The first line number in those lambdas is for the inner call itself, so 65100 would be inserted.

Looks like it's should be mapped line number (am not sure of all possibilities), so 65100 shoudn't be inserted (at least for not inlineOnly inline call)

Comment (max-kammerer) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: Looks like i forgot about nop, that is generated before inline call with call site linenumber

Comment (pyos) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: The nop isn't even necessary because inline functions start with fake variables which have no line number.

```
fun f() = inlineOnly { inlineOnly { } }
```
```
         0: iconst_0
         1: istore_0
         2: iconst_0
         3: istore_1
         4: iconst_0
         5: istore_2
         6: iconst_0
         7: istore_3
         8: nop
         9: nop
        10: return
      LineNumberTable:
        line 5: 0 <-- outer call site
        line 65100: 2 <-- outer lambda call
        line 5: 4 <-- inner call site
        line 65100: 6 <-- inner lambda call
        line 5: 8 <-- inside inner lambda
        ...
```

If the inner call is not `@InlineOnly`:
```
         0: iconst_0
         1: istore_0
         2: iconst_0
         3: istore_1
         4: iconst_0
         5: istore_2
         6: iconst_0
         7: istore_3
         8: nop
         9: nop
        10: nop
        11: return
      LineNumberTable:
        line 5: 0 <-- outer call
        line 65100: 2 <-- outer lambda call
        line 5: 4 <-- inner call
        line 7: 6 <-- inside inline function
        line 5: 8 <-- inside inner lambda
        ...
```

Comment (max-kammerer) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: Yes, nop is removed by optimizers further.
Please unmute
FirBlackBoxCodegenTestGenerated$CallableReference$AdaptedReferences.testInlineDefault

Comment (pyos) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: That test doesn't look related to this PR, seems like it should've been unmuted in current master.

Comment (max-kammerer) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: Oh..do you mean  we can avoid it generation at all?

Comment (pyos) [reply compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt:543]: Yes, but that's super minor. As you said, it's removed later anyway.

File: compiler/backend/src/org/jetbrains/kotlin/codegen/inline/inlineCodegenUtils.kt [hunk 2]
```diff
@@ -536,10 +536,24 @@ fun isFakeLocalVariableForInline(name: String): Boolean {
...
+            // @InlineOnly functions, however, contain no line numbers, so if the lambda is single-line,
+            // the entire call will "meld" into a single region. To break it up, we insert a different line
+            // number that is remapped by the SMAP to a line that does not exist.
+            val label = Label()
+            mv.visitLabel(label)
+            mv.visitLineNumber(JvmAbi.LOCAL_VARIABLE_INLINE_ARGUMENT_SYNTHETIC_LINE_NUMBER, label)
+        }
+    }
+
+    fun onInlineLambdaEnd(mv: MethodVisitor) {
```
max-kammerer, line 556: Could you explain new logic added with `onInlineLambdaEnd`?
Is it tested?

max-kammerer: Oh sorry it's old one

max-kammerer:
Thank you!

PR #3326, opened 2020-04-23
Reflect: Handle methods with inline class default parameters

sfs:
Fix for [KT-38491](https://youtrack.jetbrains.com/issue/KT-38491).


udalov:
Thanks! Merged manually in 905a8ca0d07.

PR #3327, opened 2020-04-23
JVM IR: Avoid unnecessary CHECKCASTs in enum classes

sfs:
This is a small optimization to remove CHECKCAST instructions from Enum class initializers.

The background is that we had a bug with an internal tool which tries to parse the bytecode in an enum class initializer and looks for patterns such as 
```
LDC "C"
ICONST_2
ACONST_NULL
INVOKESPECIAL Foo$C.<init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
<...>
PUTSTATIC Foo.C : LFoo;
```
We currently generate an unnecessary CHECKCAST instruction before the PUTSTATIC, which ended up breaking the tool.

I originally tried to fix this by just generating fewer CHECKCAST instructions in general (that was what #3171 was about), but doing this correctly is fairly complicated and expensive.


max-kammerer:
Thank you!

PR #3329, opened 2020-04-23
KT-32640 KDoc. Support markdown link in '@see' block

turansky:
Related issue - [KT-32640](https://youtrack.jetbrains.com/issue/KT-32640)

Screenshot 2020-04-23 at 18 52 19

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/kdoc/KDocRenderer.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/kdoc/KDocRenderer.kt
@@ -133,7 +133,7 @@ object KDocRenderer {
                 if (subjectName != null) {
                     DocumentationManagerUtil.createHyperlink(this, subjectName, subjectName, false)
                 } else {
-                    append(tag.getContent())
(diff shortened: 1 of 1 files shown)


idea/src/org/jetbrains/kotlin/idea/kdoc/KDocRenderer.kt
```diff
@@ -199,6 +198,18 @@ object KDocRenderer {
         }
     }
 
+    private fun markdownToHyperLink(markdown: String): String? =
```
darthorimar, line 201: Do we to need to check the markdown structure here? Maybe just `parseMarkdown(markdown).toHtml()`?

vladimirdolzhenko:
Any update on this ?

idea/src/org/jetbrains/kotlin/idea/kdoc/KDocRenderer.kt
turansky: I think that `@see` block must support only links, other markdown looks like error

turansky:
Not soon. Feel free to use these changes for new PR (with tests)

PR #3332, opened 2020-04-24
FIR2IR: populate overriddenSymbols and superQualifierSymbol for overriding functions.

jsjeon:
It seems IR backend relies on `overriddenSymbols` to traverse/collect overriding members. In particular, bridge lowering and interface lowering use it to determine, e.g., what methods require bridges (for covariant return types or default interface methods).

So far, `fir2ir` populates it only for fake overrides, and this PR adds support for overriding functions during FIR-2-IR conversion. An alternative is to do so during FIR resolution (`AbstractFirUseSiteMemberScope#doProcessFunctions`), but it seems FIR resolution itself depends on overridden symbols to resolve call targets.

Other things that are open for design choices are what to use for hierarchy traversals (FIR elements v.s. IR elements) and how to determine overriding functions, in particular, type checking (FIR overriding checker v.s. IR type checker). To avoid redundant type conversion, this PR chose IR-based hierarchical search, along with IR type checker.


jsjeon:
Btw, this unmutes 66, yet breaks 6 cases. All those broken cases are stack overflow. I briefly mentioned the root cause in our quip doc, but the key is a wrong `final` modifier for overriding functions. Due to that, bridge lowering incorrectly introduces a bridge that causes mutual recursion (between the bridge and the overriding method, which calls `super`). I noticed `fir2ir` generally uses `final` more often than `psi2ir`, and already explored where to fix; so I can work on it as a follow-up.

jsjeon:
A quick fix about visibility didn't resolve the broken cases, sorry for the false alarm. It's rather about missing `superQualifierSymbol` in `IrCallImpl`:

`fir2ir`:
```
    FUN name:test visibility:public modality:OPEN <> ($this:<root>.Z) returnType:kotlin.String
      overridden:
        internal open fun test (): kotlin.String declared in <root>.A
      $this: VALUE_PARAMETER name:<this> type:<root>.Z
      BLOCK_BODY
        RETURN type=kotlin.Nothing from='public open fun test (): kotlin.String declared in <root>.Z'
          CALL 'internal open fun test (): kotlin.String declared in <root>.A' type=kotlin.String origin=null
            $this: GET_VAR '<this>: <root>.Z declared in <root>.Z.test' type=<root>.Z origin=null
```

v.s.

`psi2ir`:
```
    FUN name:test visibility:public modality:OPEN <> ($this:<root>.Z) returnType:kotlin.String
      overridden:
        internal open fun test (): kotlin.String declared in <root>.A
      $this: VALUE_PARAMETER name:<this> type:<root>.Z
      BLOCK_BODY
        RETURN type=kotlin.Nothing from='public open fun test (): kotlin.String declared in <root>.Z'
          CALL 'internal open fun test (): kotlin.String declared in <root>.A' superQualifier='CLASS CLASS name:A modality:OPEN visibility:public superTypes:[kotlin.Any]' type=kotlin.String origin=null
            $this: GET_VAR '<this>: <root>.Z declared in <root>.Z.test' type=<root>.Z origin=null
```

So, `super.foo()` points to the base type A's `foo` in `psi2ir`, whereas, in `fir2ir`, it points to the bridge method, resulting in a mutual recursion. Will keep digging into it.

jsjeon:
Here we go: https://github.com/JetBrains/kotlin/pull/3332/commits/d23ce29741ee545f52db6893716e160acf17dfdd (set `superQualifierSymbol`). This unmutes additional 36, including 5 out of 6 regressions from the 1st commit (https://github.com/JetBrains/kotlin/pull/3332/commits/352e91dd2d3edf56f390f1d4b699f865b8b583d9). So, there are still one or two corner cases...

mglukhikh:
Generally both fixes look good, I will look at them on Monday more attentively and probably merge. Thank you!

mglukhikh:
Just for you to know, correct macro for YouTrack looks this way: `#KT-38416 Fixed`, beginning from `#` and with capital `F`.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
```diff
@@ -279,6 +296,64 @@ internal tailrec fun FirCallableSymbol<*>.deepestMatchingOverriddenSymbol(root:
...
+    return findMatchingOverriddenSymbolsFromSupertypes(irBuiltIns, target, result)
+}
+
+fun isOverriding(
+    irBuiltIns: IrBuiltIns,
+    target: IrFunction,
+    superCandidate: IrFunction
+): Boolean =
+    target.name == superCandidate.name &&
+            // Not checking the receiver type since we are traversing super type hierarchy, i.e., the expected subtype relation between
```
mglukhikh, line 345: It's not true IMO, extension receivers can have different types

mglukhikh:
```
open class Base {
    fun String.foo() {}
}

class Derived : Base() {
    fun Int.foo() {}
}
```

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
```diff
@@ -279,6 +296,64 @@ internal tailrec fun FirCallableSymbol<*>.deepestMatchingOverriddenSymbol(root:
...
+            // receivers is implicitly guaranteed.
+            target.returnType.isSubtypeOf(superCandidate.returnType, irBuiltIns) &&
```
mglukhikh, line 347: This check isn't necessary
```
open class Base {
    fun bar(): String = ""
}

class Derived : Base() {
    // Error: 'bar' hides member of supertype 'Base' ...
    fun bar(): Int = 42
}
```

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
```diff
@@ -26,21 +28,27 @@ import org.jetbrains.kotlin.fir.types.*
...
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
 import org.jetbrains.kotlin.ir.symbols.*
-import org.jetbrains.kotlin.ir.types.IrErrorType
-import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.symbols.impl.IrClassSymbolImpl
+import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.types.impl.IrErrorTypeImpl
 import org.jetbrains.kotlin.ir.util.SymbolTable
+import org.jetbrains.kotlin.ir.util.isFakeOverride
+import org.jetbrains.kotlin.ir.util.isTypeParameter
```
mglukhikh, line 43: This function is no more accessible, so I inlined it

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
```diff
@@ -204,6 +212,15 @@ private fun FirConstKind<*>.toIrConstKind(): IrConstKind<*> = when (this) {
     FirConstKind.IntegerLiteral, FirConstKind.UnsignedIntegerLiteral -> throw IllegalArgumentException()
 }
 
+internal fun FirTypeRef.toFirClassSymbolOrNull(session: FirSession): FirClassSymbol<*>? {
+    if (this is FirResolvedTypeRef) {
+        if (type is ConeClassLikeType) {
+            return (type as ConeClassLikeType).lookupTag.toSymbol(session).safeAs()
```
mglukhikh, line 218: You can use `coneTypeSafe` for such transformation. We will think about better utility functions. At this moment I inlined this function.

mglukhikh:
Total list of corner cases (first two appeared here when I omitted return type check):

- `overridingMethodInGenericClass2.kt`
- `kt1397.kt`
- `removeSetInt.kt`
- `interfaceHashCode.kt`

mglukhikh:
Merged manually

jsjeon:
> Just for you to know, correct macro for YouTrack looks this way: `#KT-38416 Fixed`

Aha, good to know that.

> I will fix all problems myself. BTW omitting return type check fixes 28 more tests (mainly because of generic return types) and breaks 2 more tests.

Thank you for the comments and fixes!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
jsjeon: Ack. Thanks.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
jsjeon: Oh, totally missed extension receiver; thank you for the fix!

But, my comment was mainly about the dispatch receiver type, so I believe it still holds (i.e., we still don't check the dispatch receiver type since their subtype relations are obvious).

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
jsjeon: Good to learn that, thanks!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
jsjeon: Your example (with proper `open` and `override` modifiers) rather indicates that we need a subtype check for return type. In the current `psi`-based frontend and diagnostic, I got the following error message:
```
Return type is 'Int', which is not a subtype of overridden
public open fun bar(): String defined in Base 
```

Without such check, we're now setting `overriddenSymbol` for `Derived#bar` as `Base#bar`, aren't we? Actually, in your comment, you said: "(they should match each other...". It makes sense that the previous use of subtype didn't handle generic return types well, but I wonder we should at least have `equalTypes` check for return types, which should be extended to cover generic types if needed.

------

With a quick test of using `equalTypes`, though, I saw a number of failing cases (of course), and one of them makes me believe that we need that subtype check for return types due to covariant return types:
```
// In .../box/defaultArguments/function/covariantOverride.kt
open class Foo {
    open fun foo(x: CharSequence = "O"): CharSequence = x
}

class Bar(): Foo() {
    override fun foo(x: CharSequence): String {   // Note the covariant return type
        return x.toString() + "K"
    }
}
```

jsjeon: Discussed offline. The example case would be a compilation error anyway, so whether setting `overriddenSymbols` or not doesn't matter.

PR #3333, opened 2020-04-24
KT-38267 False positive "Call on collection type may be reduced" with Java platform types: suggested to reduce 'mapNotNull' call to 'map'

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38267

### DIFF
--- a/idea/testData/inspectionsLocal/collections/uselessCallOnCollection/MapNotNullWithLambda5.kt
+++ b/idea/testData/inspectionsLocal/collections/uselessCallOnCollection/MapNotNullWithLambda5.kt
@@ -1,0 +1,14 @@
+// WITH_RUNTIME
+// FIX: Change call to 'map'
+
+fun test(): List<String> {
+    return listOf(1, 2, 3).<caret>mapNotNull { i ->
+        foo {
+            bar(i)
+        }
+    }
+}
+
+fun <T> foo(f: () -> T): T = f()
+
+fun bar(i: Int): String = ""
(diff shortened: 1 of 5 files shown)


igoriakovlev:
Hi. Could you please check this FP bug for latest kotlin master. It seems that this bug is not reproducible for now. If so, please close MR.

idea/testData/inspectionsLocal/collections/uselessCallOnCollection/MapNotNullWithLambda5.kt
```diff
@@ -0,0 +1,14 @@
...
+    return listOf(1, 2, 3).<caret>mapNotNull { i ->
+        foo {
+            bar(i)
+        }
+    }
+}
+
+fun <T> foo(f: () -> T): T = f()
+
+fun bar(i: Int): String = ""
```
t-kameyama, line 14: @igoriakovlev 
`MapNotNullWithLambda5.kt` test fails In latest master(069adebf0102d5b9f6fe5929e902eda816c608f2).

igoriakovlev: @t-kameyama 
I mean that I do not see this FP in last master IDE. Did you checked it?

t-kameyama: I checked IDE behaviour, found false negative in latest master(5efbe6ae).

In lastest master:
```kotlin
fun test() {
    // Not suggested. It's OK
    val a: List<String> = listOf(1, 2, 3).mapNotNull { i ->
        foo {
            bar(i)
        }
    }

    // Not suggested. It's false negative.
    val b: List<String> = listOf(1, 2, 3).mapNotNull { i ->
        foo {
            baz(i)
        }
    }
}

fun <T> foo(f: () -> T): T = f()

fun bar(i: Int): String? = null

fun baz(i: Int): String = ""
```


In this branch:
```kotlin
fun test() {
    // Not suggested It's OK.
    val a: List<String> = listOf(1, 2, 3).mapNotNull { i ->
        foo {
            bar(i)
        }
    }

    // Suggested. It's OK.
    val b: List<String> = listOf(1, 2, 3).mapNotNull { i ->
        foo {
            baz(i)
        }
    }
}

fun <T> foo(f: () -> T): T = f()

fun bar(i: Int): String? = null

fun baz(i: Int): String = ""
```

Should I close this pull request and create separate issue?

igoriakovlev:
Merged. Thank you.

PR #3334, opened 2020-04-25
KT-38559 "Change JVM name" (@JvmName) quickfix: improve name suggester for generic functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38559


yanex:
Merged into `master`. Thanks!

PR #3337, opened 2020-04-25
KT-35411 Explicitly handle array annotation args in scripting pre-processing

efemoney:
https://youtrack.jetbrains.com/issue/KT-35411


efemoney:
I have a few concerns
- Should this be moved into `ConstantExpressionEvaluator` as a special case for scripting (this was my first approach)
- This handles only annotation array args, so constructing annotations is still broken for enums and `KClass` annotation args
- I don't know enough but I'm interested in a more permanent solution. How can I fix this once and for all? Is there something we arent invoking in the scripting compiler that isnt properly intializing the compiler FE?

ligee:
Merged manually along with #3467

PR #3339, opened 2020-04-27
Add samples for collections

elijahverdoorn:
Add a sample for functions in collections.

### DIFF
--- a/libraries/stdlib/samples/test/samples/collections/collections.kt
+++ b/libraries/stdlib/samples/test/samples/collections/collections.kt
@@ -773,6 +773,27 @@ class Collections {
             val emptyList = emptyList<Int>()
             assertPrints(emptyList.elementAtOrElse(0) { "no int" }, "no int")
         }
+
+        @Sample
+        fun find() {
+            val list = listOf(1, 2, 3, 1, 2, 3)
+            assertPrints(list.find { it == 1 }, "0")
+            assertPrints(list.find { it == 4 }, "null")
+        }
+
+        @Sample
+        fun findLast() {
+            val list = listOf(1, 2, 3, 1, 2, 3)
+            assertPrints(list.findLast { it == 1 }, "3")
+            assertPrints(list.findLast { it == 4 }, "null")
+        }
+
+        @Sample
+        fun getOrNull() {
+            val array = arrayOf(1, 2, 3)
+            assertPrints(array.getOrNull(1), "2")
+            assertPrints(array.getOrNull(4), "null")
+        }
     }
 
     class Sorting {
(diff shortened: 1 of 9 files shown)


libraries/stdlib/samples/test/samples/collections/maps.kt
```diff
@@ -318,6 +318,20 @@ class Maps {
             assertPrints(map2, "{beverage=2.7$, meal=12.4$}")
         }
 
+        @Sample
+        fun mapNotNull() {
+            val map1 = mapOf("foo" to 1, "bar" to 2, "baz" to 3)
+            val map2 = map1.mapNotNull {
```
ilya-g, line 324: There are several problems with this example:
- it doesn't seem to be referenced in code
- `component1` and `equals` are functions to satisfy operator conventions (destructuring and ==), they are rarely called directly in real code. So if you want to compare map entry key with "bar", it would be more idiomatic to write `it.key == "bar"`
- for `if (predicate(it)) it else null` we have `takeIf` and `takeUnless` functions: 
`it.takeUnless { it.key == "bar" }`
- the example overall looks like a contrived way to filter map based on keys, so `Map.filterKeys` would probably do fine in this case. On the other hand, the advantage of `mapNotNull` is that is can additionally transform an element, for example, with it we can filter by keys, but return the corresponding values:
`map1.mapNotNull { entry -> entry.takeUnless { it.key == "bar" }?.value }` 
or using the entry destructuring 
`map1.mapNotNull { (key, value) -> value.takeUnless { key == "bar" } }`

libraries/stdlib/samples/test/samples/collections/collections.kt
```diff
@@ -773,6 +773,27 @@ class Collections {
             val emptyList = emptyList<Int>()
             assertPrints(emptyList.elementAtOrElse(0) { "no int" }, "no int")
         }
+
+        @Sample
+        fun find() {
+            val list = listOf(1, 2, 3, 1, 2, 3)
+            assertPrints(list.find { it == 1 }, "0")
```
ilya-g, line 780: `find` return the found element and not its index in the list. Please, ensure that samples really pass by running `:kotlin-stdlib:samples:check` gradle task. 

Here, having the first and the last elements indistinguishable, it's hard to tell which one is actually returned.

So, I propose to use slightly less trivial predicate, that could return the same result for different elements, for example `it % 2 == 0` or `(it and 1) == 0` that checks parity.
```
val numbers = listOf(1, 2, 3, 4, 5, 6, 7)
val firstOdd = numbers.find { it % 2 != 0 }
val lastEven = numbers.findLast { it % 2 == 0 }
...
```

qurbonzoda:
The have fixed the review notes and merged the changes to `master`: https://github.com/JetBrains/kotlin/commit/41d56156086b7fd59f1b2523a1eadd77c92ba069
Thanks for the effort!

PR #3341, opened 2020-04-27
Add repeat(times: Long){} overload

BenWoodworth:
My use case is for use large files with `java.io.RandomAccessFile`. This class uses `Long` for most of its operations, and `repeat(long.toInt()) {...}` isn't sufficient because file pointers can be in the multi-gigabyte range, which `Int`s can't handle.


qwwdfsad:
Unfortunately, we cannot introduce such an overload for two reasons:

* It will be a source-incompatible change for every `val function = ::repeat` call in the while
* Due to how lambdas are resolved, `repeat(20) {}` is also ambiguous

PR #3344, opened 2020-04-28
KT-18242 KT-19192 Add java.nio.Path extensions to stdlib-jdk7

ajalt:
This PR adds most extensions on `java.io.File` in `kotlin.io` to `java.nio.Path`. This includes extensions from `FileReadWrite.kt`, `Utils.kt`, and `FileTreeWalk.kt`.

I attempted to keep the implementations, documentation, and tests as similar as possible to the existing implementations.

I am happy to add, remove, or move to separate PRs any of the functions of this PR.

##### `createTempDir`, `createTempFile`

These functions have no `File` parameters, so can't be overloaded. Equivalents exist as `Files.createTempFile()` and `Files.createTempDirectory()`.

##### `startsWith`, `endsWith`, `normalize`, `resolve`, `resolveSibling`

These exist as member functions on `Path`

##### `relativeTo`, `relativeToOrNull`, `relativeToOrSelf`, `toRelativeString`, `toRelativeStringOrNull`

This functionality exists as the `Path.relativize` member function, which is equivalent to `relativeTo`, but with the receiver and parameter flipped. `foo.relativeTo(bar)` is equivalent to `bar.relativize(foo)`. We could potentially add a `relativizeOrNull` extension to make that pattern simpler.

##### `isRooted`

`Path` has a `root` method, so `isRooted` is equivalent to `root != null`

### New extensions

All of the simple boolean attribute checks from `java.nio.Files` were added as extensions on `Path`. These extensions are used commonly enough that it seems worth supporting them. This functionality for `File` is implemented as member methods.

The following `Path` extensions were added:

- `exists`
- `isDirectory`
- `isExecutable`
- `isFile`
- `isHidden`
- `isReadable`
- `isSameFile`
- `isSymbolicLink`
- `isWritable`

Some of these extensions take options that are forwarded to their `Files` method, so all of the extensions were implemented as functions rather than properties for consistency.

Additionally, `Path.listFiles` was added to match the `File.listFiles` method. One motivation for its addition was that it's used several times in the implementation of other file extensions. The way to list directory contents with `java.nio` is via `Files.newDirectoryStream()`, which returns an iterable object that must be closed to avoid leaking resources. It's difficult to use correctly with functions like `map` and `filter`, so this extension was added as a simpler, less error-prone alternative.

### Other changes

I added overloads of several of the read-write that take `OpenOptions` to expose the greater control that `java.nio` introduces. For example, you can use `printWriter(APPEND)` to create a `PrintWriter` that doesn't delete the contents of an existing file.

All the new extensions throw exceptions (such as `NoSuchFileException`) from `java.nio` rather than the copies from `kotlin.io`. The `kotlin.io` copies take `File` objects as parameters, and so aren't compatible with `Path`s.


libraries/stdlib/jdk7/src/kotlin/io/PathReadWrite.kt
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * Appends [text] to the content of this file using UTF-8 or the specified [charset].
+ *
+ * @param text text to append to file.
+ * @param charset character set to use.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.appendText(text: String, charset: Charset = Charsets.UTF_8): Unit {
+    writeText(text, StandardOpenOption.APPEND, charset = charset)
```
JakeWharton, line 243: This impl shows why I always kept the options last. It could be
```kotlin
writeText(text, charset, APPEND)
```
if flipped

libraries/stdlib/jdk7/src/kotlin/io/PathTreeWalk.kt
```diff
@@ -0,0 +1,274 @@
...
+    /** Abstract class that encapsulates file visiting in some order, beginning from a given [root] */
+    private abstract class WalkState(val root: Path) {
+        /** Call of this function proceeds to a next file for visiting and returns it */
+        public abstract fun step(): Path?
+    }
+
+    /** Abstract class that encapsulates directory visiting in some order, beginning from a given [rootDir] */
+    private abstract class DirectoryState(rootDir: Path) : WalkState(rootDir) {
+        init {
+            if (_Assertions.ENABLED)
```
JakeWharton, line 54: Yikes. I guess they don't specify `-Xassertions=jvm` for stdlib builds?

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+        throw FileAlreadyExistsException(toString(), null, "The destination file already exists.")
+    }
+
+    if (this.isDirectory()) {
+        if (target.isDirectory() && Files.newDirectoryStream(target).use { it.firstOrNull() } != null) {
+            throw FileAlreadyExistsException(toString(), null, "The destination file already exists.")
+        }
+        try {
+            Files.createDirectories(target)
+        } catch (e: FileAlreadyExistsException) {
```
JakeWharton, line 114: nit:
```suggestion
        } catch (_: FileAlreadyExistsException) {
```

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+})
+
+/**
+ * Check if this file exists.
+ *
+ * @param options Options to control how symbolic links are handled.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.exists(vararg options: LinkOption): Boolean = Files.exists(this, *options)
```
JakeWharton, line 276: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path is a file.
+ *
+ * @param options Options to control how symbolic links are handled.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isFile(vararg options: LinkOption): Boolean = Files.isRegularFile(this, *options)
```
JakeWharton, line 285: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+/**
+ * Check if this path is a directory.
+ *
+ * By default, symbolic links in the path are followed.
+ *
+ * @param options Options to control how symbolic links are handled.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isDirectory(vararg options: LinkOption): Boolean = Files.isDirectory(this, *options)
```
JakeWharton, line 296: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path exists and is a symbolic link.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isSymbolicLink(): Boolean = Files.isSymbolicLink(this)
```
JakeWharton, line 303: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path exists and is executable.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isExecutable(): Boolean = Files.isExecutable(this)
```
JakeWharton, line 310: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path is considered hidden.
+ *
+ * This check is dependant on the current filesystem. For example, on UNIX-like operating systems, a
+ * path is considered hidden if its name begins with a dot. On Windows, file attributes are checked.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isHidden(): Boolean = Files.isHidden(this)
```
JakeWharton, line 320: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path exists and is readable.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isReadable(): Boolean = Files.isReadable(this)
```
JakeWharton, line 327: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check that this path exists and is writable.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isWritable(): Boolean = Files.isWritable(this)
```
JakeWharton, line 334: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Check if this path points to the same file or directory as [other].
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.isSameFile(other: Path): Boolean = Files.isSameFile(this, other)
```
JakeWharton, line 341: inline + @InlineOnly

libraries/stdlib/jdk7/src/kotlin/io/PathReadWrite.kt
```diff
@@ -0,0 +1,373 @@
...
+ *
+ * @param options options to determine how the file is opened
+ * @param charset character set to use.
+ * @param bufferSize necessary size of the buffer.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+@kotlin.internal.InlineOnly
+public inline fun Path.bufferedWriter(
+    vararg options: OpenOption,
```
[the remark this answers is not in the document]
ajalt: Yes, i went with this order because `charset` has a default value.

With the varargs first, you have to pass `charset` as a keyword argument, but you can omit it to use the defautl value.

With varargs last, you would have to specify an explicit value of `charset` in order to pass any vararg arguments.

I kept the overloads without vararg options so that you can pass charset without varargs.

Another possible design would be to instead have three overloads of each function, one with just a charset and default value, one with just varargs, and one with charset (with no default) followed by varargs. That would be allow you to pass both a charset and varargs without keyword arguments.

libraries/stdlib/jdk7/src/kotlin/io/PathReadWrite.kt
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/**
+ * Appends an [array] of bytes to the content of this file.
+ *
+ * @param array byte array to append to this file.
+ */
+@SinceKotlin("1.3")
+@ExperimentalStdlibApi
+public fun Path.appendBytes(array: ByteArray): Unit {
```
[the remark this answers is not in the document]
ajalt: The `File` versions of these extensions aren't inline, so I matched that here. It's possible that's just due to legacy reasons, so I'll go ahead and make these inline. I can revert the change if there's a reason not to inline them.

libraries/stdlib/jdk7/src/kotlin/io/PathReadWrite.kt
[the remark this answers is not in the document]
ajalt: I went ahead and moved the varargs to the last position and implemented all the extra overloads. The overloads are `@InlineOnly`, so it seems fine to add a few. 

I can always revert the change.

ilya-g:
Great job here!
Unfortunately, we won't be able to review this carefully by 1.4, so more realistically it could make its way to 1.4.x as an experimental API.

ilya-g:
@ajalt, thanks again for the PR. 
We've reviewed the proposed functions and compared the API provided in `java.nio.file` package suitable to use with `Path` with the `File` API consisting of `File` members and extensions from `kotlin.io`.
In general, we want to expose the most of static functions provided in `java.nio.file.Files` as extensions for `Path` and also enrich them with extensions already known from `kotlin.io`.

This will require an additional effort in this branch, so first I'd like to ask whether you could devote some more time to do that work in the coming weeks? If not, it's not a problem, we can take the PR in the current state and take care of the rest.

ajalt:
@ilya-g I'm happy to make changes, just let me know what you need.

ilya-g:
Great.

First, we want to postpone some controversial or not straightforward parts: 
- `PathTreeWalker` and `*recursive` functions dependent on it;
- `isRooted` and `forEachBlock` extensions, which are rarely used.

Then, for path relativizing we'd like to have the same functions as we have for `File` for those who already got used to them, with the exception of `toRelativeString`. I.e. `relativeTo`/`relativeToOrSelf`/`relativeToOrNull`. 

Regarding vararg overloads, it should be possible to call a function with both optional and vararg arguments using an explicit argument name: e.g. `reader(options = arrayOf(...))`. 
We expect options to be needed only in advanced use cases, so we'd like to avoid inflating the number of overloads just for sake of adding these options. Therefore, one vararg after all optional parameters would be enough for now.

ajalt:
I rebased the command, addressed the review comments and did a copy edit pass of all kdocs.

One thing to note is that I implemented `Path.relativeTo` in terms of `Path.relativize`. In some edge cases (namely, when `this` is an empty path), `Path.relativeTo` will return a slightly different result than `File.relativeTo` would for the same paths. I don't know if that's a problem or not.

libraries/stdlib/jdk7/test/PathExtensionsTest.kt
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Test
+    fun testBufferedReader() {
+        val file = Files.createTempFile(null, null)
+        val lines = listOf("line1", "line2")
+        Files.write(file, lines)
+
+        assertEquals(file.bufferedReader().use { it.readLines() }, lines)
+        assertEquals(file.bufferedReader(StandardOpenOption.READ).use { it.readLines() }, lines)
```
ilya-g, line 135: Could you check these overloads' usage in tests? 
I suppose they should be updated after we've removed all except one overload of each.

libraries/stdlib/jdk7/src/kotlin/io/PathUtils.kt
```diff
@@ -0,0 +1,294 @@
...
+        return if (separator != "/") toString().replace(separator, "/") else toString()
+    }
+
+/**
+ * Returns this path's [fileName][Path.getFileName] without an extension.
+ */
+@SinceKotlin("1.4")
+@ExperimentalStdlibApi
+public val Path.nameWithoutExtension: String
+    get() = fileName.toString().substringBeforeLast(".")
```
ilya-g, line 44: `fileName` could be `null`. I doubt that we want to have `nameWithoutExtension` return "null" string in this case.

libraries/stdlib/jdk7/test/PathExtensionsTest.kt
ajalt: I removed the lines testing the removed overloads and left tests for calls with the default argument and with all argument specified. Let me know if you want something else.

libraries/stdlib/jdk7/test/PathExtensionsTest.kt
```diff
@@ -0,0 +1,353 @@
...
+        val writer3 = file.printWriter(Charsets.UTF_8, StandardOpenOption.APPEND)
+        val str4 = "Hello one last time!"
+        writer3.println(str4)
+        writer3.close()
+
+        val reader = file.bufferedReader()
+        assertEquals(str1, reader.readLine())
+        assertEquals(str2, reader.readLine())
+        assertEquals(str3, reader.readLine())
+        assertEquals(str4, reader.readLine())
```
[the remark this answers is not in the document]
ajalt: Good catch. This test was copied from the existing `Files.helpers2` test, which also doesn't close it's reader. Would you like me to fix that test was well, or would you rather it be left alone?

libraries/stdlib/jdk7/test/PathExtensionsTest.kt
```diff
@@ -0,0 +1,353 @@
...
+    private fun compareFiles(src: Path, dst: Path, message: String? = null) {
+        assertTrue(dst.exists())
+        assertEquals(src.isRegularFile(), dst.isRegularFile(), message)
+        if (dst.isRegularFile()) {
+            assertTrue(src.readBytes().contentEquals(dst.readBytes()), message)
+        }
+    }
+
+    @Test
+    fun testBufferedReader() {
```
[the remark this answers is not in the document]
ajalt: That definitely makes more sense, I'll move them. I put them here originally because the equivalent tests for `File` live in the `File.kt` test rather than the `ReadWrite.kt` test.

libraries/stdlib/jdk7/test/PathExtensionsTest.kt
```diff
@@ -0,0 +1,353 @@
...
+        file.appendBytes(" world!".encodeToByteArray())
+        assertEquals(file.readText(), "Hello world!")
+    }
+
+    @Test
+    fun testAttributeGetters() {
+        val file = Files.createTempFile(null, null)
+        assertTrue(file.exists())
+        assertTrue(file.isRegularFile())
+        assertFalse(file.isDirectory())
```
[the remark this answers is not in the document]
ajalt: Good idea. I'll also add a test for a path that doesn't exist, so we can make sure `notExists` works as expected.

libraries/stdlib/jdk7/src/kotlin/io/PathReadWrite.kt
```diff
@@ -0,0 +1,255 @@
...
+ * Default charset is UTF-8.
+ *
+ * You may use this function on huge files.
+ *
+ * @param charset character set to use.
+ * @param action function to process file lines.
+ */
+@SinceKotlin("1.4")
+@ExperimentalStdlibApi
+public fun Path.forEachLine(charset: Charset = Charsets.UTF_8, action: (line: String) -> Unit): Unit {
```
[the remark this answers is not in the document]
ajalt: If I remove that overload, I get a `NoSuchMethodError` in the test that calls that function. IntelliJ doesn't show any errors, though, so maybe it's a problem on my setup. I'll remove it in the PR, and if CI shows the same error, I can add it back.

ilya-g:
>  In some edge cases (namely, when this is an empty path), Path.relativeTo will return a slightly different result than File.relativeTo would for the same paths.

I believe it's the following bug in the JDK implementation: https://bugs.openjdk.java.net/browse/JDK-8072495. It was fixed in JDK 9, but I think we can workaround it in our `relativeTo`. I'm going to try it.

ajalt:
So that bug could be fixed with something like this:

```kotlin
public fun Path.relativeTo(base: Path): Path {
    return if (base.toString().isEmpty()) this else base.relativize(this)
}
```

which matches the [fix in the JDK](https://github.com/openjdk/jdk/commit/980fb58833d51d504527498eff872c081fd5495c), although `toString` isn't always free on unix paths like is it on windows.

Even if we did fix that bug, there are other Path bugs fixed later (e.g. [this](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8140449) and [this](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8066943)). Do we need to fix all of them? If so I don't think we could use the platform `relativize` any more. The existence of all these bugs is why I left out `relativeTo` initially.

ilya-g:
Yes, I've also spotted them when testing `relativeTo`. The workaround is rather ugly, but still possible without reimplementing `relativize` completely: 6cd89c4e5dce91bbc37694737341db00d0b997d2.

ilya-g:
Would you mind if I squash commits in this PR into one or two?

ajalt:
Not at all, feel free to make whatever changes you'd like.

JakeWharton:
Any chance this makes 1.4.20? 😀

ilya-g:
Good news! I've squashed these changes into two commits and merged them together with our subsequent enhancements into master. The resulting changeset is https://github.com/JetBrains/kotlin/compare/0fa2cc15de9ffc0248ea9c3275e1dcb637157e95~11...0fa2cc15de9ffc0248ea9c3275e1dcb637157e95

> Any chance this makes 1.4.20? 😀

Chances are pretty high, indeed. We plan to show it in 1.4.20-RC.

ajalt:
Great, then I'll close the PR! Thanks!

PR #3345, opened 2020-04-28
FIR2IR: generate synthetic members for inline class.

jsjeon:
We can reuse the consolidated `DataClassMembersGenerator` to generate synthetic members for `inline` class: `equals`, `hashCode`, and `toString`.

Inline class lowering in the backend requires `overridenSymbols` for those synthetic members, which should have been part of https://github.com/JetBrains/kotlin/commit/57fe01c375d0f35c4d5746811669e96fcbc5a377, and this PR also addresses that.

One caveat is, in `light-tree`, it was assumed that `inline` modifier is only used as function modifier (and has always been added to the function modifier set). Adding an enum entry for the same modifier to `CLASS_MODIFIER` intercepts modifier modification, resulting in missed function modifier instead. To mitigate that, this PR introduces `INLINE_MODIFIER`, a singleton set, and uses it to branch out for `inline` modifier, and there, the use of that modifier for `inline` class is handled specially.


compiler/testData/codegen/box/reflection/callBy/inlineClassDefaultArguments.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND_FIR: JVM_IR
```
jsjeon, line 1: FYI, this is not a regression. This test was added today w/o this opt-out.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
```diff
@@ -379,7 +379,7 @@ class DeclarationsConverter(
                     isInner = modifiers.isInner()
                     isCompanion = modifiers.isCompanion() && classKind == ClassKind.OBJECT
                     isData = modifiers.isDataClass() && classKind != ClassKind.OBJECT
-                    isInline = modifiers.hasInline()
+                    isInline = modifiers.isInlineClass() && classKind != ClassKind.OBJECT
```
jsjeon, line 382: I wonder, though, that both `isData` and `isInline` should check `classKind == ClassKind.CLASS` instead?

compiler/testData/codegen/box/reflection/callBy/inlineClassDefaultArguments.kt
sfs: Sorry about that, I forgot to run the FIR tests before submitting the PR.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
sfs: This may be a silly question, but why is the `classKind` check necessary at all? The old frontend just [looks at the modifier list](https://github.com/JetBrains/kotlin/blob/1a26d2e157a31459d9bf3abca0123f03a6ae011d/compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/descriptors/LazyClassDescriptor.java#L156) and psi2ir sets the isInline flag [directly based on the descriptor](https://github.com/JetBrains/kotlin/blob/1a26d2e157a31459d9bf3abca0123f03a6ae011d/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/impl/IrClassImpl.kt#L74).

jsjeon: Not a silly question, thanks for the pointers! @mglukhikh any reasons we check that? I saw some diagnostic tests check invalid uses of that modifier, though.

Hm, I just checked the grammar: https://kotlinlang.org/docs/reference/grammar.html#classModifier and `inline` is not a class modifier?!

mglukhikh: I will think about this, thanks.

mglukhikh: Probably these checks are not needed, I will try to remove them.

mglukhikh:
I merged it together with related fix: https://github.com/JetBrains/kotlin/commit/ef8bcceb8eaa5aef5ee0c4306f672362cd6f9e24. Thank you!

compiler/testData/codegen/box/reflection/callBy/inlineClassDefaultArguments.kt
jsjeon: No problem. Hope `:compiler:test` trigger FIR tests as well...

punzki: @jsjeon AFAIK... `:compiler:test` will only run a subset of the JVM tests. You need to run `firCompilerTest` to run all the FIR-related tests, or `:compiler:fir:fir2ir:test` to run the box tests for FIR (along with other FIR2IR tests). `compilerTest` will run _all_ compiler-related tests, including FIR.

jsjeon: Ah, good to know that, thanks.

PR #3346, opened 2020-04-29
Add bunch for Android Studio 4.1 (based on IDEA 2020.1)

rsauciuc:
This is mostly preserving the diffs for as40.

The last commit in the series reflects platform differences which should go away as we catch up with 2020.1 stable, and is meant only as a temporary patch.


buildSrc/src/main/kotlin/IdeCompatibilityDsl.kt.as41
```diff
@@ -0,0 +1,99 @@
...
+    IJ191(Platform.P191),
+    IJ192(Platform.P192),
+    IJ193(Platform.P193),
+    IJ201(Platform.P201),
+
+    AS34(Platform.P183),
+    AS35(Platform.P183),
+    AS36(Platform.P192),
+    AS40(Platform.P193),
+    AS41(Platform.P201);
```
goodwinnk, line 54: We usually prefer to avoid creating bunches for build files to avoid full rebuild after switch operation. By adding additional entry to this file, it's possible to create build files with all supported platforms with conditions in code.

goodwinnk:
Thanks a lot for these pull requests! Unfortunately, I can't merge them till we have a working build, because I would like to modify commits before pushing them to master.

rsauciuc:
Thanks for the comment, it's a great tip. Clearly I haven't used bunches enough to start thinking about the impact on CI :)

I shared a build with you. Please feel free to take over the commits here, they were meant as a starting point only.

PR #3349, opened 2020-04-29
JVM IR: Create static default argument stubs

sfs:
Right now we are creating non-static default argument stubs in DefaultArgumentStubGenerator only to make them static in a later lowering.

Apart from being a bit messy, the fake overrides from non-static default argument stubs can break the invariant that every concrete fake override has a single most specific concrete implementation method. While we do have a workaround for this, I would like to avoid more problems in the future by generating default argument stubs as static in the first place.


udalov:
Sorry for the delay. It seems that this change leads to the following exception when bootstrapping Kotlin with JVM IR, during `:compiler:cli:compileKotlin`:
```
java.lang.RuntimeException: Exception while generating code for:
FUN FUNCTION_FOR_DEFAULT_PARAMETER name:analyzeFilesWithJavaIntegration$default visibility:public modality:FINAL <> (project:com.intellij.openapi.project.Project, files:kotlin.collections.Collection<org.jetbrains.kotlin.psi.KtFile>, trace:org.jetbrains.kotlin.resolve.BindingTrace, configuration:org.jetbrains.kotlin.config.CompilerConfiguration, packagePartProvider:kotlin.Function1<com.intellij.psi.search.GlobalSearchScope, org.jetbrains.kotlin.load.kotlin.PackagePartProvider>, declarationProviderFactory:kotlin.Function2<org.jetbrains.kotlin.storage.StorageManager, kotlin.collections.Collection<org.jetbrains.kotlin.psi.KtFile>, org.jetbrains.kotlin.resolve.lazy.declarations.DeclarationProviderFactory>?, sourceModuleSearchScope:com.intellij.psi.search.GlobalSearchScope?, klibList:kotlin.collections.List<org.jetbrains.kotlin.library.KotlinLibrary>?, $mask0:kotlin.Int, $handler:kotlin.Any?) returnType:org.jetbrains.kotlin.analyzer.AnalysisResult
```
with the following cause:
```
java.lang.IllegalStateException: No mapping for symbol: VALUE_PARAMETER name:<this> type:org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM
      at org.jetbrains.kotlin.backend.jvm.codegen.IrFrameMap.typeOf(irCodegenUtils.kt:55)
      at org.jetbrains.kotlin.backend.jvm.codegen.ExpressionCodegen.visitGetValue(ExpressionCodegen.kt:518)
```

Could you investigate please? If you need any help with bootstrapping, please let me know. Basically you can follow the instruction in the pinned post on the internal channel.

sfs:
Thanks for the report, I'll look into it!

sfs:
The problem was in `JvmStaticAnnotationLowering`, I've fixed it and added a test case for it.

I'm getting out of memory errors when trying to bootstrap the Kotlin compiler with the IR backend, though, but that doesn't seem to be related to this PR.

sfs:

udalov:
Looks like moving `jvmStaticAnnotationPhase` way above broke `addContinuationPhase` in case of `@JvmStatic suspend` functions in companion objects. `addContinuationPhase` expects all suspend functions to have attribute owner, with the attached internal name for the corresponding continuation class computed in `InventNamesForLocalClasses`, but now there are `JVM_STATIC_WRAPPER` without such information. Could you please take a look?

BTW this didn't come up in any real scenario, I noticed it accidentaly by looking at the list of phases between the old and the new position of `jvmStaticAnnotationPhase`. I wonder if there are some other phases there which might not expect lifted JvmStatic functions at that point...

udalov:
I'll close this because of the problem described above, and a bunch of conflicts (including with #3865). Feel free to reopen or send a new PR if you wish to work on this further. Thanks!

PR #3350, opened 2020-04-29
FIR2IR: populate overriddenSymbols and superQualifierSymbol for overriding properties.

jsjeon:
This PR extends https://github.com/JetBrains/kotlin/pull/3332 to overriding properties.


jsjeon:
I'm working on one regression in this PR: `kt2566_2.kt`. It looks like a labeled `super` is not handled correctly, e.g.,

```
open class A {
  open val foo
}

class B : A {
  inner class E {
    val foo = super<A>@B.foo
  }
}
```

`psi2ir`:
```
PROPERTY name:foo visibility:public modality:FINAL [val]
  FIELD PROPERTY_BACKING_FIELD name:foo type:kotlin.String visibility:private [final]
    EXPRESSION_BODY
      CALL 'public open fun <get-foo> (): kotlin.String declared in <root>.A' superQualifier='CLASS CLASS name:A modality:OPEN visibility:public superTypes:[kotlin.Any]' type=kotlin.String origin=GET_PROPERTY
        $this: GET_VAR '<this>: <root>.B declared in <root>.B' type=<root>.B origin=null
```

v.s.

`fir2ir`:
```
PROPERTY name:foo visibility:public modality:FINAL [val]
  FIELD PROPERTY_BACKING_FIELD name:foo type:kotlin.String visibility:private [final]
    EXPRESSION_BODY
      CALL 'public open fun <get-foo> (): kotlin.String declared in <root>.A' superQualifier='CLASS CLASS name:A modality:OPEN visibility:public superTypes:[kotlin.Any]' type=kotlin.String origin=GET_PROPERTY
        $this: GET_VAR '<this>: <root>.B.E declared in <root>.B.E' type=<root>.B.E origin=null
```

mglukhikh:
>  It looks like a labeled super is not handled correctly

I know that @semoro already have a fix for this problem.

mglukhikh:
Merged manually, thank you!

mglukhikh:
It's the fix of @semoro : https://github.com/JetBrains/kotlin/commit/a11866ab5a56ea2dd52a3fdf9a823304047a2b96, but your regression wasn't fixed. I will look more attentively.

mglukhikh:
Really this ^ problem is related to `FirSuperReference`, not to `FirThisReferenceExpression`. They are a bit similar. We should replace `lastDispatchReceiverExpression` with `dispatchReceiverExpression(irSuperClass)`. However I tried to do it and catched some strange side effects...

jsjeon:
Note that, both at the resolution phase and at the conversion phase, the stack for (implicit) receivers would look like: [`B`, `B.E` (peak)], and the use of `lastDispatchReceiver` is currently picking `B.E`. I wonder, though, we can't use `irSuperClass` directly, since the peak, `B.E`, is not a subtype of `A`. Instead, as the label pointed out, we should pick the receiver of type `B` and moves to its super type (`A`).

To handle `super<A>@B`, we really need `label` `B`. However, I found that, unlike `FirThisReference`, `FirSuperReference` doesn't have `labelName`. During resolution, I saw that, for `FirThisReference`, such label is used to retrieve the proper dispatch receiver from the stack, and I believe we should have similar logics for the labeled super.

But, that's not the only factor. At the conversion stage, we're seeing neither `FirThisReferenceExpression`, `FirThisReference`, nor `FirSuperReference` for this case. It's resolved qualified access whose callee symbol directly points to super type's property properly. I think `label` for `FirSuperReference` is modeled and propagated to such qualified access, and then the retrieval of dispatch receiver for `superQualifierSymbol` should use that if available.  I'll dig into it further.

PR #3352, opened 2020-04-29
[FIR] Support several annotation class diagnostics

McCreeMainer:
Added checker for next errors:
- ANNOTATION_CLASS_MEMBER
- LOCAL_ANNOTATION_CLASS_ERROR
- MISSING_VAL_ON_ANNOTATION_PARAMETER
- VAR_ANNOTATION_PARAMETER


compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationClassDeclarationChecker.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.jetbrains.kotlin.KtNodeTypes.*
+import org.jetbrains.kotlin.lexer.KtTokens.*
+import org.jetbrains.kotlin.psi.KtParameter
+
+object FirAnnotationClassDeclarationChecker : FirDeclarationChecker<FirDeclaration>() {
+    override fun check(declaration: FirDeclaration, context: CheckerContext, reporter: DiagnosticReporter) {
+        if (declaration !is FirRegularClass) return
+        if (declaration.classKind != ANNOTATION_CLASS) return
+        for (it in declaration.declarations)
+            when (it.source?.elementType) {
```
mglukhikh, line 31: Here it's much easier to check `it` type:
```
when {
    it is FirConstructor && it.isPrimary -> { ... }
    it is FirRegularClass -> { ... }
    else -> { ... }
}
```

mglukhikh: (when possible, we should not go to `source` to check something, but use own FIR properties and types)

McCreeMainer: > (when possible, we should not go to `source` to check something, but use own FIR properties and types)
Fixed.

mglukhikh:
Merged manually. Congratulations!

PR #3353, opened 2020-04-30
KT-38649 False positive quickfix "Assignment should be lifted out of when" in presence of smartcasts

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38649


yanex:
Review: 👍 .

yanex:
Merged into `master`. Thanks!

PR #3355, opened 2020-04-30
FIR: Provide dispatch receiver for property according to property type

jsjeon:
This extends https://github.com/JetBrains/kotlin/commit/078cf02c8a00088a41ed985f4a5ab994de89df24 (or part of https://github.com/JetBrains/kotlin/pull/3234) to `FirPropertySymbol` (previously, only for `FirBackingFieldSymbol`). This way, `Properties` case of `TowerLevels#processElementByName` covers both properties with and without (resolved) backing field.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/tower/TowerLevels.kt
```diff
@@ -209,7 +209,7 @@ class ScopeTowerLevel(
             }
         }
         return when {
-            candidate !is FirBackingFieldSymbol -> null
+            candidate !is FirBackingFieldSymbol && candidate !is FirPropertySymbol -> null
```
jsjeon, line 212: Actually, I tried to remove this condition at all (so that `Functions` can be treated in the same way), but it didn't have any differences (w.r.t. BB tests) but introduced many other diagnostic failures. So I guess there are other steps during the resolution that set/use dispatch receiver for function calls.

mglukhikh: Yet I can tell you it should not work this way. See comment above `ScopeTowerLevel` about dispatch receiver. I will investigate it to understand why it helps in some particular cases.

compiler/fir/analysis-tests/testData/resolve/expresssions/invoke/extensionOnObject.txt
```diff
@@ -13,7 +13,7 @@ FILE: extensionOnObject.kt
         public final fun f(op: R|X.() -> kotlin/Unit|): R|kotlin/Unit| {
             R|<local>/op|.R|FakeOverride<kotlin/Function1.invoke: R|kotlin/Unit|>|(Q|X|)
             lval x: R|X| = Q|X|
-            R|<local>/op|.R|FakeOverride<kotlin/Function1.invoke: R|kotlin/Unit|>|(R|<local>/x|)
+            R|<local>/op|.R|FakeOverride<kotlin/Function1.invoke: R|kotlin/Unit|>|(this@R|/Y|.R|<local>/x|)
```
mglukhikh, line 16: For example, `x` is a local variable here, it should not have dispatch receiver.

compiler/fir/analysis-tests/testData/resolve/delegates/delegateWithLambda.txt
```diff
@@ -33,7 +33,7 @@ FILE: delegateWithLambda.kt
                 }
             }
 
-            ^ R|<local>/y|
+            ^ this@R|/Test|.R|<local>/y|
```
mglukhikh, line 36: Here we have the same case: `y` is a local variable

compiler/fir/analysis-tests/testData/resolve/expresssions/invoke/implicitTypeOrder.txt
```diff
@@ -5,7 +5,7 @@ FILE: implicitTypeOrder.kt
         }
 
         public final fun bar(): R|A| {
-            ^bar R|/foo|.R|/A.invoke|()
+            ^bar this@R|/A|.R|/foo|.R|/A.invoke|()
```
mglukhikh, line 8: And here `foo` is a top-level property, again without dispatch receiver

mglukhikh:
To be short, the problem here is around smart-casts and intersection types. Simple example could be the following:
```
interface A
interface B {
    fun next(): B?
}
class Some {
    fun foo(s: A) {
        if (s !is B) return
        var ss: B? = s
        while (ss != null) {
            ss = ss.next()
        }
    }
}
```

Here `ss` in `ss.next()` is an `FirExpressionWithSmartCast`, and its smart cast type is intersection between `A` and `B`. At this moment FIR2IR converts intersection types by just taking first of them and inserts implicit cast operator to this type. Here if you take `A` as first, you encounter `NoSuchMethodError` because `A` has no `next()` method, but if you take `B` instead, everything will be Ok.

In your fix, you incorrectly added `this` receiver to `ss` and this change `deceives` smart cast system. Now it (probably) thinks that `ss` is not a local variable but mutable property, and does not insert any smart casts. Type of `ss` is left `B` as explicitly stated, and everything accidentally works.

To fix this situation correctly, we may do the following. When we encounter `FirExpressionWithSmartCast` in FIR2IR in dispatch receiver position, we should build scope for each of intersected type (using `type.scope()`) and check which one of these scopes contains the function or property accessed. After this we select this type as target type for implicit cast, and everything should work.

mglukhikh:
@jsjeon The preliminary fix working as stated above is here: https://github.com/JetBrains/kotlin/commit/e63feb4044dfd0ed88b085be937dbaa7a8e84ec8. All tests fixed in your PR work here, with two additional tests; however, it breaks two tests with smart cast on `this`, tomorrow I will look why (or you may do it today if you wish). See `FirBlackBoxCodegenTestGenerated/TypeInfo`.

jsjeon:
Thank you for the investigation. I think I need to learn how to read those rendered FIR; then I would spot those local variables or top-level property should not have dispatch receiver.

It was really helpful to see detailed explanations, along with corresponding implementation! Also, good to learn that we can build scopes again during conversion. Though, I'm not sure why intersection types for smart-cast are not determined during resolution yet carried over to conversion phase and finalized there.

Alas, I don't have enough time today to dig further into two regressions, but here are what I've found so far. Those two tests are easily(?) mitigated by:
```
$ git diff
diff --git a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
index ce4cc7b4029..4e7bc2d6a29 100644
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
@@ -361,7 +361,7 @@ class Fir2IrVisitor(
         if (castTypeRef is FirResolvedTypeRef && originalTypeRef is FirResolvedTypeRef) {
             val castType = castTypeRef.type
             val originalType = originalTypeRef.type
-            if (originalType == castType) {
+            if (value.type == originalType && originalType == castType) {
                 return value
             }
             if (castType is ConeIntersectionType) {
```
I noticed, previously, type from the converted expression is compared against `castType` so I just double-checked that and found that `value.type` could not be same as `originalType`.

But I have to admit this is a weird stopgap, and then I found it could originate from here: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt#L395 At that point, indeed `intersectedType` is equal to original type. Even `FirThisReference` has `typeRef` of `B`, and thus `intersectedType`, including the type of qualified access expression, is same. On the other hand, during conversion, `this` reference has type `A` (from the code context), and that's how `value.type` could be different from `originalType`.

An issue I can think of is, during resolution, `this` reference has too specific type: `B`, not `A`, even though the enclosing function is `A#foo`. According to some terms around there, like `data flow` analysis, I wonder it knows the only call site is `box()` where `B().foo()` is invoked, which guarantees that `this` is always of type `B`? Perhaps that is too smart for smart-cast. In some sense, that explains why smart-cast needs to be (re)visited during conversion.

Hope these investigation are useful, and feel free to close this PR. Thanks again for detailed explanations.

jsjeon:
Oh, btw, I'm talking about `TypeInfo.testSmartCastThis` test. :)

mglukhikh:
Thank you, it was helpful. I will probably just remove the check in question
```
if (originalType == castType) ...
```

PR #3356, opened 2020-05-01
FIR: handle labeled super reference properly.

jsjeon:
Similar to `this` reference with label (https://kotlinlang.org/docs/reference/grammar.html#THIS_AT), `super` reference can be labeled (https://kotlinlang.org/docs/reference/grammar.html#superExpression). Therefore, `FirSuperReference` should have the counter part of `FirThisReference.labelName`.

During resolution, such `labelName` in `FirThisReference` is used to retrieve the dispatch receiver from the implicit receiver stack, according to "label." Of course, `FirSuperReference.labelName` can be used in the same way.

During conversion, when converting `super`-based qualified access, if the label is specified, we should use it to retrieve the correct dispatch receiver. This addresses the regression from https://github.com/JetBrains/kotlin/pull/3350.


mglukhikh:
Yes, it's forgotten part of grammar, thank you.

mglukhikh:
@jsjeon I added some minor fixes, you may look at them if you wish (the second one is more important):

- https://github.com/JetBrains/kotlin/commit/e04d5d4e61b2c1610e2e404a2d430f5acf5ceeb2
- https://github.com/JetBrains/kotlin/commit/35490c2ff93d4a274c8db8b01314d001212c32e7

Yet they aren't in master, so your opinion is greatly appreciated. Probably I will push PR and my fixes tomorrow.

jsjeon:
Thanks for the pointers. IIUC, the last commit is moving the part of setting dispatch receiver for super reference from conversion to resolution, and a nice by-product is that now we don't need to have somewhat duplicate utils to retrieve dispatch receivers during conversion, because everything is set during resolution. It looks great, and wish things will be kept that way!

mglukhikh:
Merged manually.

PR #3358, opened 2020-05-02
KT-38492 False positive "Add import" intention for already imported class

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38492


yanex:
Review: OK.

yanex:
Merged into `master`. Thanks!

PR #3361, opened 2020-05-03
Include a scripts dependencies in project resolve scope

garyttierney:
This resolves an issue with completion/code-assistance not working inside scripts that are contained in a regular modules sources. Previously, only SDK dependencies were showing via auto-imports or code-completion (ctrl-space). This change includes script dependencies in the global project resolve scope to allow this to work.

When a script is included in a project as a regular file to be compiled
it produces a `ModuleSourceInfo` representing its metdata, as opposed to
`ScriptModuleInfo` for the special-cased scratch files and Gradle build
script.

This change includes script dependencies as part of the regular search
scope when attempting to resolve items.

The module info for a script contained within a regular modules sources is produced [here](https://github.com/JetBrains/kotlin/blob/7d5396ac95879062e8d9cfde8eab155c86e483b0/idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/project/getModuleInfo.kt#L261) (called by https://github.com/JetBrains/kotlin/blob/7d5396ac95879062e8d9cfde8eab155c86e483b0/idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/project/getModuleInfo.kt#L182).

This could maybe be changed to differentiate between `ModuleInfo` produced for a script, and `ModuleInfo` produced for a regular `KtFile`, however this change doesn't seem too intrusive.

I've put together an example repository demonstrating the problem (try to get code assistance in a script file without this patch): https://github.com/garyttierney/kotlin-scripts-imports-reproducer

### DIFF
--- a/idea/idea-analysis/src/org/jetbrains/kotlin/idea/stubindex/KotlinSourceFilterScope.kt
+++ b/idea/idea-analysis/src/org/jetbrains/kotlin/idea/stubindex/KotlinSourceFilterScope.kt
@@ -113,7 +113,7 @@ class KotlinSourceFilterScope private constructor(
             includeProjectSourceFiles = true,
             includeLibrarySourceFiles = false,
             includeClassFiles = true,
-            includeScriptDependencies = false,
+            includeScriptDependencies = true,
             includeScriptsOutsideSourceRoots = true,
             project = project
         )


NataliaUkhorskaya:
The major reason `getModuleInfo` for script in module returns `ModuleSourceInfo` is that they should see the other files in this module and otherwise - all other kotlin files should see declarations from the script.

As for your fix - I don't understand the original issue: script inside module should see only the dependencies of this module, all additional script dependencies produced by its definition should be included into module dependencies. I tried your example: am I right that your script should have a MyScript as definition? If so, this definition doesn't add any additional dependencies for such scripts. Which class do you expect to see in competion, that isn't seen before your fix?

garyttierney:
Hi @NataliaUkhorskaya, thanks for the detailed reply!

> they should see the other files in this module and otherwise - all other kotlin files should see declarations from the script.

Yep, that is exactly right and it appears to be working insofar building the code is concerned:

```kt
// script-examples/src/main/kotlin/test.myscript.kts
// `MyObject` declared in script-examples/src/main/kotlin/kotlintest/MyObject.kt
val obj = kotlintest.MyObject()
```

The IDE reports this code as correct and will build it successfully. However, if I write
```
val obj = MyObj<caret>
```
and invoke code completion (via ctrl-space) the only suggestions are ones from the SDK, not the module (or any of the modules dependencies).

> As for your fix - I don't understand the original issue: script inside module should see only the dependencies of this module

That's correct, but the behaviour right now is that code assistance doesn't see the dependencies of the module (it doesn't see classes declared in the same module actually, demonstrated with the `kotlintest.MyObject` case -- `kotlintest.MyCommonObject` declared in a dependant module shows it failing with a dependency).

> I tried your example: am I right that your script should have a MyScript as definition?

Yep, `MyScript` is the correct definition and scripts are compiled by `kotlinc` during the Gradle build process.

> If so, this definition doesn't add any additional dependencies for such scripts.

The expectation I've had so far (and this was how scripting behaved in 1.3.40, albeit before the configuration refinement API was introduced) was that a script `a.myscript.kts` defined in `scripting-impl` and using a definition `MyScript` in `scripting-api` should be able to access all types available to `scripting-impl` that a regular Kotlin class file would be able to access. The compiler and IDE do reflect this, but it does not extend to code assistance.

> Which class do you expect to see in competion, that isn't seen before your fix?

- `kotlintest.MyObject` - declared in the same module as the script (.kts, not the script definition)
- `kotlintest.MyCommonObject` declared in a module that the script (.kts) module depends on

Both of these types above do not appear in code assistance before the fix, but do afterwards.

garyttierney:
Here is a bug report that describes the problem as well: https://youtrack.jetbrains.com/issue/KT-35825

garyttierney:
Apologies for the notification spam, just one further update. I originally noticed the issue here: https://github.com/JetBrains/kotlin/blob/7f8774f68e6ada142aeb044d29c7cfa1abe050a0/idea/idea-core/src/org/jetbrains/kotlin/idea/core/KotlinIndicesHelper.kt#L390.

The index doesn't return the results because `scope` is different. Here is `scope` for a kts file:

```
KotlinSourceFilterScope(delegate=Union: (File :file:///home/gtierney/workspace/github/apollo2/game/server/src/main/scripts/org/test.script.kts,com.intellij.psi.search.NonClasspathDirectoriesScope@173d1c) src=true libSrc=false cls=true scriptDeps=false scripts=true)
```

vs. a Kt file

```
KotlinSourceFilterScope(delegate=Module-with-dependencies:kotlin-scripting-imports-reproducer.script-examples.main compile-only:true include-libraries:true include-other-modules:true include-tests:false src=true libSrc=false cls=true scriptDeps=false scripts=true)
```

NataliaUkhorskaya:
@garyttierney Yes, seems the root cause is in the wrong scope provided for completion. You can see `KotlinScriptResolveScopeProvider` if you are interested: it should provide enlarged scope only for scripts outside of source root. 

I think the issue became not so easy to fix, so I'll fix it by myself soon! 

Thanks a lot for the investigation and sorry for the delayed response due to the holidays.

garyttierney:
@NataliaUkhorskaya ah, I see. This was this missing piece for me. I couldn't find out how those resolve scopes were being produced!

Thanks for your insight into how it works. If you don't mind, I'm going to continue poking at this to see if I can get the correct fix if it's not as complicated as overhauling large parts of the IDEA integration, as it's quite the blocker for us. I'll add any further updates to the relevant YouTrack issue instead of updating this PR or opening any further ones.

NataliaUkhorskaya:
@garyttierney Ok, ask me any questions if you need some help. 
Fisrt advise you need to check if moduleInfo for relevant file is ScriptModuleInfo or ModuleInfo. In case of ModuleInfo you don't need to enlarge it's scope, because all script dependencies are already contained in the default one.

PR #3363, opened 2020-05-04
[FIR2IR] Allow implicit casts from Short and Byte to Int.

madsager:
This allows the generation of code that avoids boxing when
comparing primitives to boxed primitives.

Split out test expectations for primitive comparisons so the
different IR generation can be dealt with in tests.


madsager:
@mglukhikh can you have a look at the Fir2Ir change?

@dnpetrov can you have a look at the IR differences here. It looks like Psi2Ir puts in implicit conversions here, but it seems to work without them for Fir and looking at other ir text expectations there are more places where fir2ir does not add these implicit conversions. Will that cause trouble for some of the other backends, I wonder? Also, the test expected fir2ir and psi2ir to generate the same, that is why we lost the psi2ir changes when you landed this and the Ir tests are currently failing.

madsager:
The psi2ir change broke KotlinNative. I'll rework this some other way. Closing!

PR #3365, opened 2020-05-05
JVM IR: Take superQualifierSymbols into account when lowering inline classes

sfs:
Specifically, super calls with inline class parameters were broken.

The PR also adds a test for this case. It's honestly surprising that we didn't have such a test already.


udalov:
Thanks!

PR #3366, opened 2020-05-05
IR: cleanup unused implementation in WrappedDescriptors.

jsjeon:
This is partial revert of https://github.com/JetBrains/kotlin/commit/08ca6f5f99cbf7e90f8b584cf21b0173c0ee85c7 (as part of https://github.com/JetBrains/kotlin/pull/3261), which set metadata field of IR symbols during FIR2IR conversion, triggering some paths in IR codegen to use wrapped descriptors copy builder and others to serialize. Now FIR element-based metadata serialization is in, those additions were not needed anymore.


mglukhikh:
Merged manually 👍

PR #3367, opened 2020-05-06
Extend scripting REPL API and implement it

ileasile:
This PR adds the following features to scripring REPL API:
- new method `inspect` which allows to get info about symbols under cursor. Currently only types/signatures are supported;
- completion options which may be passed via compilation configuration. Currently, options include shadowed symbols filtering and names filter;
- `analyze` method now returns text representation of snippet result type, or `null`, if snippet returns nothing. Note that this API change is breaking.

Also, testing infrastructure was slightly refactored, and common code parts of `analyze`, `complete` and `inspect` methods were picked out to a separate private `analyzeWithCursor` method.


ligee:
Partially merged - the result type part. Please submit another PR with remaining functionality, as we agreed on offline.

PR #3369, opened 2020-05-06
FIR: make (K)SuspendFunctionX derived from (K)Function

jsjeon:
Here are counterparts in `IrBuiltsIns`:

* `SuspendFunctionN` uses `Function` as a base class: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrFunctionFactory.kt#L108 which is used to create a super type reference, just the same as `FunctionN` uses `Function`: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrFunctionFactory.kt#L95

* `KSuspendFunctionN` uses `KFunction` as a base class: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrFunctionFactory.kt#L138, just the same as `KFunctionN` uses `KFUnction`: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/descriptors/IrFunctionFactory.kt#L123


mglukhikh:
Merged manually 👍

PR #3370, opened 2020-05-07
KT-20795 "replace explicit parameter with it" creates invalid code in case of overload ambiguities

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-20795


yanex:
Broken case (1):

```
fun foo() {
    val a: (Int) -> Unit = { a -> println(a) }
}
```

t-kameyama:
PTAL

yanex:
Review: OK.

yanex:
Merged into `master`. Thanks!

PR #3373, opened 2020-05-07
[JVM_IR] Avoid safe-call conversions from Byte? and Short? to Int? for

madsager:
comparisons.

Having those conversions leads to unnecesary boxing and null checks.

This change does it only for JVM in the optimization lowering. It
is unclear to me if the other backends can get away with something
similar.

### DIFF
--- a/compiler/testData/codegen/bytecodeText/intrinsicsCompare/byteSmartCast_after.kt
+++ b/compiler/testData/codegen/bytecodeText/intrinsicsCompare/byteSmartCast_after.kt
@@ -1,6 +1,4 @@
 // !LANGUAGE: +ProperIeee754Comparisons
-// IGNORE_BACKEND: JVM_IR
-
 fun equals3(a: Byte?, b: Byte?) = a != null && b != null && a == b
 
 fun equals4(a: Byte?, b: Byte?) = if (a is Byte && b is Byte) a == b else null!!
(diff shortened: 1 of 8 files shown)


compiler/testData/codegen/bytecodeText/intrinsicsCompare/byteSmartCast_after.kt
```diff
@@ -13,9 +11,19 @@ fun less4(a: Byte?, b: Byte?) = if (a is Byte && b is Byte) a < b else true
...
 
+// JVM_TEMPLATES
 // 3 Intrinsics\.areEqual
 // 0 Intrinsics\.compare
 // 4 INVOKEVIRTUAL java/lang/Byte\.byteValue \(\)B
 // 2 INVOKEVIRTUAL java/lang/Number\.intValue \(\)I
 // 0 IFGE
 // 3 IF_ICMPGE
+
+// JVM_IR_TEMPLATES
```
madsager, line 22: The change here is that instead of `Number.intValue`, `Number.byteValue` is used when known to be a byte. Additionally, the IR backend actually generates 3 `areEqual` calls, but the asm optimization method visitor then ends up unboxing and using `Number.byteValue` on the arguments.

dnpetrov:
Ok, I'll run tests and merge manually.

dnpetrov:
Pushed manually.
1f4a3b0d1cb326815cc18eec227d3bfd28f481b2

PR #3375, opened 2020-05-08
[FIR] enable populateOverriddenSymbols for deserialized methods.

juan-chen:
Deserialized methods don't have isOverride flag set, even for
overriding methods, e.g., UIntIterator.next. After enabling
populateOverriddenSymbols for such methods, there is regression,
testCharBuffer, where CharBuffer.get is considered overriding
kotlin.CharSequence.get, because CharBuffer implements
java.lang.CharSequence and java.lang.CharSequence is mapped to
kotlin.CharSequence.

To avoid such overriding, I followed the psi frontend to use
JavaIncompatibilityRulesOverridabilityCondition to check method
overriding incompatibility. The only issue is
JavaIncompatibilityRulesOverridabilityCondityCondition checks if a
class is a Java Class by checking if the descriptor is a
JavaClassDescriptor, which Fir didn't create for Java Classes. This
commit adds WrappedJavaClassDescriptor for such cases.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -161,7 +162,8 @@ class Fir2IrClassifierStorage(
         parent: IrDeclarationParent? = null,
         origin: IrDeclarationOrigin = IrDeclarationOrigin.DEFINED
     ): IrClass {
-        val descriptor = WrappedClassDescriptor()
+        val descriptor = if (regularClass is FirJavaClass) WrappedJavaClassDescriptor()
```
jsjeon, line 165: If this is the only place `FirJavaClass` is used and that's why `compiler:fir:java` dependency is added, you can use `regularClass.origin == FirDeclarationOrigin.Java` instead.

juan-chen: Thanks for the suggestion! Files updated.

compiler/frontend/build.gradle.kts
```diff
@@ -6,6 +6,7 @@ plugins {
 dependencies {
     compile(project(":kotlin-annotations-jvm"))
     compile(project(":core:descriptors"))
+    compile(project(":core:descriptors.jvm"))
```
mglukhikh, line 9: I think this dependency should not be here

core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JavaIncompatibilityRulesOverridabilityCondition.kt
```diff
@@ -51,72 +51,71 @@ class JavaIncompatibilityRulesOverridabilityCondition : ExternalOverridabilityCo
...
+    override fun getContract() = ExternalOverridabilityCondition.Contract.CONFLICTS_ONLY
 
-        val overriddenBuiltin = superDescriptor.getOverriddenSpecialBuiltin()
+    companion object {
+        // This overridability condition checks two things:
+        // 1. Method accidentally having the same signature as special builtin has does not supposed to be override for it in Java class
+        // 2. In such Java class (with special signature clash) special builtin is loaded as hidden function with special signature, and
+        // it should not override non-special method in further inheritance
+        // See java.nio.Buffer
+        fun isIncompatibleInAccordanceWithBuiltInOverridabilityRules(
```
mglukhikh, line 62: I don't like FIR2IR wants to use this function. May be the better solution is to write FIR version of it. Let's discuss on today meeting.

compiler/frontend/build.gradle.kts
udalov: +1! Please be careful with adding such dependencies. Common compiler code shouldn't depend on platform-specific parts. It sometimes really turns out quite difficult to get rid of such dependencies afterwards.

jsjeon: As per my suggestion above, this dependency is removed at the last commit in this PR.

udalov: Looks like your suggestion was about the `fir2ir` -> `fir:java` dependency, which is indeed removed and it's great :). But this one is the `frontend` -> `descriptors.jvm` dependency.

jsjeon: Oh, there is another dependency addition, wasn't aware of that, sorry!

mglukhikh:
Refactored and added partially. Yet I leave this PR on hold to remember a possible way to fix regressions.

mglukhikh:
I'm closing the remainder because all tests fixed here are already fixed in master.

PR #3376, opened 2020-05-09
FIR2IR: fix offsets for calls.

jsjeon:
The motivation is `InRangeWithNonmatchingArgument`, which is the only test failure that shows [D8/R8](https://r8.googlesource.com/r8#) call stack. Currently, in `fir`-generated IR, every single call has default-ish offset, -1, for start/end offset, since `typeRef`, which doesn't have `source` at all :), is used, instead of the correct source.

As a result, IR backend produces bytecode with non-sense line numbers, like 10...then all the sudden 65,100 :o, along with non-sense local variable table. The master version of `d8` is lenient with such erroneous debug info, but the latest stable version still has an issue: https://issuetracker.google.com/issues/156142802

With the correct source, now no more `d8` error. Note that the original test still suffers from a resolution error, but, as a by-product, some debugging-related tests are unmuted.


mglukhikh:
Thank you 👍

PR #3378, opened 2020-05-11
FIR2IR: apply SAM conversion to arguments if needed.

jsjeon:
The motivation was `coroutines/async.kt`, which is failing due to the lack of SAM conversion (from `Function2` to `BiConsumer`), resulting in casting exception at runtime.

The first commit is about adding `SAM_CONVERSION` type operator call if needed. As a starting point, it covers basic cases: a lambda as an argument whose type doesn't match with the expected parameter type, which is not built-in functional types, but indeed SAM interface. We can check if the expected parameter type has one single abstract method by examining `FirClass.declarations` or `IrClass.functions`, but I chose the latter because `FunctionReferenceLowering` eventually retrieves the SAM definition in the IR class: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FunctionReferenceLowering.kt#L122

Such safety net---assuring the parameter type as SAM---discovered that the first commit is only applicable to Kotlin interfaces, since SAM interfaces from Java, such as `BiConsumer`, don't have any members at all at the conversion phase. The second commit attempts to force loading Java SAM on demand.

Alas, it requires a couple hacks. First, those SAM stubs are not part of files that `Fir2IrConverter` is converting. The best place we can detect them is when the corresponding IR class is generated, and to force loading SAM at that point, we need a connection between the converter and classifier storage. An alternative would be introducing queue-based logic through which any components queue the request to load certain class/member, which is in turn handled by the main converter.

Second, while converting such SAM as `FirJavaMethod` to `IrFunction`, we need to convert `FirJavaTypeRef` to `IrType`. It's simply error type for now. Instead of reinventing the wheel, I simply piggy-back on utils in `fir.java.JavaUtils`. To not change visibilities and module dependencies, when converting `JavaType` to `FirJavaTypeRef`, a corresponding `ConeType` is computed/attached, and then converted to `IrType`. This has so many unexpected conversion errors, and I left TODOs with brief explanations.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -43,6 +43,8 @@ class Fir2IrClassifierStorage(
 
     private val localStorage = Fir2IrLocalStorage()
 
+    var converter: Fir2IrConverter? = null
```
mglukhikh, line 46: `lateinit` should be better here

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -145,6 +147,15 @@ class Fir2IrClassifierStorage(
         val origin = regularClass.irOrigin(firProvider)
         val irClass = registerIrClass(regularClass, parent, origin)
         processClassHeader(regularClass, irClass)
+        // TODO: queue-based logic?
+        // Force loading Java methods in Java SAM stub, since FunctionReferenceLowering relies on it.
+        if (converter != null && regularClass.origin == FirDeclarationOrigin.Java) {
+            regularClass.getSamIfAny()?.let { sam ->
```
mglukhikh, line 153: I think this logic should be in `addDeclarationsToExternalClass` and not here. `createIrClass` should just create a class.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
```diff
@@ -54,6 +55,7 @@ class Fir2IrTypeConverter(
 
     fun FirTypeRef.toIrType(typeContext: ConversionTypeContext = ConversionTypeContext.DEFAULT): IrType {
         return when (this) {
+            is FirJavaTypeRef -> coneType?.toIrType(typeContext) ?: createErrorType()
```
mglukhikh, line 58: Yet it's Ok but I should investigate this place. As I understand we should not have `FirJavaTypeRef` at this point, and if we do, it points to some FIR resolve error.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/JavaUtils.kt
```diff
@@ -80,9 +82,23 @@ internal fun FirTypeRef.toConeKotlinTypeProbablyFlexible(
 
 internal fun JavaType.toFirJavaTypeRef(session: FirSession, javaTypeParameterStack: JavaTypeParameterStack): FirJavaTypeRef {
     val annotations = (this as? JavaClassifierType)?.annotations.orEmpty()
+    val type = this
     return buildJavaTypeRef {
         annotations.mapTo(this.annotations) { it.toFirAnnotationCall(session, javaTypeParameterStack) }
-        type = this@toFirJavaTypeRef
+        this.type = type
```
mglukhikh, line 88: I think this code is not needed at all. I'll investigate why `FirJavaTypeRef` remains in resolved FIR.

mglukhikh:
Looks like yet we can accept the first commit (apply SAM conversion ...) but not the second.

mglukhikh:
I pushed the first commit into master: https://github.com/JetBrains/kotlin/commit/6034fcdc4659651829fd455d562f41a3b55dfaaa.

mglukhikh:
I added also the second commit in the following form: https://github.com/JetBrains/kotlin/commit/cdac6157a92f216637947b9c8b93f4133116b59c. Closing this one.

PR #3380, opened 2020-05-11
JVM IR: Parcelize bug fixes

sfs:
- The previous code failed to remap function symbols in overridden symbols and function references, potentially leaving some psi2ir stubs in the generated IR.
- We have to resolve parcelers for properties lazily, as custom parcelizers might have properties for which resolution would fail.


sfs:
And another fix: The code should fall back to using `Parcel.readValue/writeValue` if no other serializer is available (even without the `@RawValue` annotation). This is going to result in code which fails at runtime, unless the user links against a different library than what we are compiling against, but at least the behavior matches the JVM backend.

sfs:
And two more fixes. The first is a big one though: The `ParcelableDeclarationChecker` was only running if the view extensions were enabled as well. This lead to wrong code in many of the tests and some confusion on my side.

The second commit fixes a problem with the tests that @rbares pointed out: We are not calling `setDataPosition(0)` after unmarshalling a parcel. This works because the Robolectric test framework is more permissive than android. On recent android versions this code would fail and there is no reason to expect that this will continue to work as we update Robolectric.

sfs:
@yanex I just noticed that you already merged this, thank you! Closing the PR.

PR #3381, opened 2020-05-12
FIR2IR: split property annotations according to use-site targets.

jsjeon:
Currently, annotations for properties are converted and stored to `IrProperty.annotations` (except for `CONSTRUCTOR_PARAMETER`). Rather, depending on use-site targets, those should be associated with correct targets. For example,
```
@get:Ann
@set:Ann
@setparam:Ann
var prop : T
```
those three should be set to getter, setter, and the value parameter of the setter, since their targets are `PROPERTY_GETTER`, `PROPERTY_SETTER`, and `SETTER_PARAMETER`, respectively. For now, they are stored to `prop`'s annotations, resulting in duplicate annotation error while dexing (both `dx` and `d8`), somewhat similar to https://issuetracker.google.com/issues/129241209


compiler/testData/ir/irText/declarations/annotations/primaryConstructorParameterWithAnnotations.fir.txt
```diff
@@ -25,8 +25,6 @@ FILE fqName:<root> fileName:/primaryConstructorParameterWithAnnotations.kt
         DELEGATING_CONSTRUCTOR_CALL 'public constructor <init> () [primary] declared in kotlin.Any'
         INSTANCE_INITIALIZER_CALL classDescriptor='CLASS CLASS name:Test modality:FINAL visibility:public superTypes:[kotlin.Any]'
     PROPERTY name:x visibility:public modality:FINAL [val]
-      annotations:
-        Ann
```
jsjeon, line 29: Note that this removal is legitimate, since the constructor above has the annotation already (for `class Test(@param:Ann val x`)`. All other IR dumps show that annotations are now associated with correct targets.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.kotlin.fir.FirAnnotationContainer
+import org.jetbrains.kotlin.fir.backend.Fir2IrVisitor
+import org.jetbrains.kotlin.fir.declarations.FirProperty
+import org.jetbrains.kotlin.fir.declarations.FirValueParameter
+import org.jetbrains.kotlin.ir.declarations.*
+import org.jetbrains.kotlin.ir.expressions.IrConstructorCall
+import org.jetbrains.kotlin.ir.util.isPropertyField
+import org.jetbrains.kotlin.ir.util.isSetter
+
+internal class AnnotationGenerator(private val visitor: Fir2IrVisitor) {
```
punzki, line 18: Could you please provide some high-level comments/documentation? Doesn't have to be long. 🙂

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    fun generate(irContainer: IrMutableAnnotationContainer, firContainer: FirAnnotationContainer) {
+        irContainer.annotations = firContainer.annotations.mapNotNull {
```
punzki, line 21: Just curious if we would ever expect the `it.accept(...)` call to return `null` or something other than an `IrConstructorCall`. What case would that be? If not, then maybe we should just assert that here.

jsjeon: Hm, I guess such other cases would indicate compilation/resolution errors at earlier stages, but not 100% sure. @mglukhikh will have better answers. :)

jsjeon:
@semoro Based on the offline discussion today, I thought of what other options we have, but I'm rather convinced that `fir2ir` conversion is the most reasonable place to distribute property annotations than any other places, such as resolution or FIR building.

During the resolution, even `FirProperty` is not created yet. Either light-tree-based builder or `RawFirBuilder` will build `FirProperty` when visiting either `LighterASTNode` or `KtProperty`, but still not all necessary pieces are ready, e.g., backing field. That is, e.g., annotations with `AnnotationUseSiteTarget.FIELD` should still wait and be bound to `FirProperty`, and then, when we prepare all corresponding IR elements, we can split property annotations accordingly.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
mglukhikh: It can return `IrErrorCallExpression` in case of resolution errors.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
mglukhikh: I cannot say that I personally need documentation here, but if Mark considers it to be necessary, please add it :)

compiler/testData/ir/irText/declarations/annotations/primaryConstructorParameterWithAnnotations.fir.txt
mglukhikh: Yes, I know that some annotations are at wrong targets now

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
jsjeon: Sounds like the current handling is needed, then.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
jsjeon: None of generators in the conversion module has such high-level comments, but that doesn't justify not adding one for this generator. :) This time, I'm doing my parts, including others that I added previously: https://github.com/JetBrains/kotlin/pull/3381/commits/8e45211aa5fc277ad01eeadd56b6b8816206386c

mglukhikh:
Merged, thank you!

PR #3382, opened 2020-05-12
FIR: set proper visibility of property accessors.

jsjeon:
If property accessors don't have specific visibility (a.k.a. `Visibilities.UNKNOWN`), their visibility should conform to property visibility, e.g.,
```
private val prop
    get
```
the getter should be `private`. Otherwise, necessary visibility bridges won't be generated.


mglukhikh:
Merged 👍

PR #3385, opened 2020-05-13
Add scripting REPL completion options

ileasile:
nameFilter - way of filtering descriptors by their names;
filterOutShadowedDescriptors - if true, filters out descriptors
shadowed by descriptors in latter-compiled snippets. May slow down
completion performance.

### DIFF
--- a/plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/CompletionOptions.kt
+++ b/plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/CompletionOptions.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.scripting.ide_services.compiler
+
+import kotlin.script.experimental.api.ScriptCompilationConfigurationKeys
+import kotlin.script.experimental.util.PropertiesCollection
+
+interface ReplCompletionOptionsKeys
+
+open class ReplCompletionOptionsBuilder : PropertiesCollection.Builder(), ReplCompletionOptionsKeys {
+    companion object : ReplCompletionOptionsKeys
+}
+
+fun ReplCompletionOptionsBuilder.filterOutShadowedDescriptors(value: Boolean) {
+    this[filterOutShadowedDescriptors] = value
+}
+
+fun ReplCompletionOptionsBuilder.nameFilter(value: (String, String) -> Boolean) {
+    this[nameFilter] = value
+}
+
+val ReplCompletionOptionsKeys.filterOutShadowedDescriptors by PropertiesCollection.key(true)
+val ReplCompletionOptionsKeys.nameFilter
+        by PropertiesCollection.key<(String, String) -> Boolean>({ name, namePart -> name.startsWith(namePart) })
+
+@Suppress("unused")
+val ScriptCompilationConfigurationKeys.completion
+    get() = ReplCompletionOptionsBuilder()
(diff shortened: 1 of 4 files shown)


plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/KJvmReplCompilerWithIdeServices.kt
```diff
@@ -48,13 +48,18 @@ class KJvmReplCompilerWithIdeServices(hostConfiguration: ScriptingHostConfigurat
...
+
             with(analyzeResult.valueOr { return it }) {
                 return getKJvmCompletion(
                     ktScript,
                     bindingContext,
                     resolutionFacade,
                     moduleDescriptor,
-                    cursorAbs
+                    cursorAbs,
+                    filterOutShadowed,
```
ligee, line 61: Why don't you pass the configuration right into the function and extract values from it right in place, inside `KJvmReplCompleter`? It will look much cleaner to me and will require fewer changes when you'll add more options to it.

plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/CompletionOptions.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlin.script.experimental.api.ScriptCompilationConfigurationKeys
+import kotlin.script.experimental.util.PropertiesCollection
+
+interface ReplCompletionOptionsKeys
+
+open class ReplCompletionOptionsBuilder : PropertiesCollection.Builder(), ReplCompletionOptionsKeys {
+    companion object : ReplCompletionOptionsKeys
+}
+
+@Suppress("unused")
```
ligee, line 17: Please explain why you need unused construct here

plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/CompletionOptions.kt
```diff
@@ -0,0 +1,32 @@
...
+
+fun ReplCompletionOptionsBuilder.nameFilter(value: (String, String) -> Boolean) {
+    this[nameFilter] = value
+}
+
+val ReplCompletionOptionsKeys.filterOutShadowedDescriptors by PropertiesCollection.key(true)
+val ReplCompletionOptionsKeys.nameFilter
+        by PropertiesCollection.key<(String, String) -> Boolean>({ name, namePart -> name.startsWith(namePart) })
+
+@Suppress("unused")
```
ligee, line 30: Is it really unused?
It is ok to have unused things if you explain why in comments, but it seems that in this case it will be safer to write a test that uses it.

ileasile: No, it's because of unused receiver. Unfortunately, I can't make suppress name more specific.
The test exists, obviously

plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/KJvmReplCompilerWithIdeServices.kt
ileasile: Tried to choose between "flat" structure and less files to change. Your arguments seem reasonable, I'll fix it.

plugins/scripting/scripting-ide-services/src/org/jetbrains/kotlin/scripting/ide_services/compiler/CompletionOptions.kt
ileasile: I'll write the test

PR #3386, opened 2020-05-13
[FIR] add support for varargs in delegated constructor calls

juan-chen:
Also add support for spread operators as named arguments.


demiurg906:
Thanks.

Merged manually.

PR #3387, opened 2020-05-13
[IR/FIR] Use `isExpect` from IR element instead of descriptor in ExpectDeclarationRemover

punzki:
This dependency on descriptors was easy to remove, but there is still a heavy dependency on descriptors in `ExpectDeclarationRemover` which is not trivial to untangle.


udalov:
Thanks!

PR #3388, opened 2020-05-13
FIR: update argument mapping while transforming integer operator.

jsjeon:
`FirIntegerOperator` is converted via integer approximator, which didn't take into account argument mapping in the call. As a result, transformed call has old argument mapping whose values refer to value parameters in the original call symbol. During argument reordering, conversion of such unknown value parameter led to out-of-bound error for now.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/IntegerLiteralTypeApproximationTransformer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/IntegerLiteralTypeApproximationTransformer.kt
@@ -77,13 +77,28 @@ class IntegerLiteralTypeApproximationTransformer(
         // e.g. Byte doesn't have `and` in member scope. It's an extension
         if (resultSymbol == null) return functionCall.compose()
         functionCall.resultType = data?.let { functionCall.resultType.resolvedTypeFromPrototype(it) } ?: resultSymbol.fir.returnTypeRef
+        // If the original call has argument mapping, values in that mapping refer to value parameters in that original symbol. We should
+        // map those original value parameters back to indices, and then renew the argument mapping with new value parameters in the result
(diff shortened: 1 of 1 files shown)


jsjeon:
However, the main motivation, `collectionLiterals/collectionLiteralsInArgumentPosition.kt`, which is the only test failure with out-of-index error, still fails, complaining that the element in the annotation value array has unexpected kind: function call.

```
@Foo(a = [1f, 2f, 1 / 0f])
fun test1() {}
```
Here, `psi2ir` replaced `1 / 0f` with float `infinity` via `ConstantExpressionEvaluator`, while `fir2ir` left it as `Int.div` call. I'm not sure when/where is a good timing/place to compute compile-time constants.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/IntegerLiteralTypeApproximationTransformer.kt
```diff
@@ -77,13 +77,28 @@ class IntegerLiteralTypeApproximationTransformer(
         // e.g. Byte doesn't have `and` in member scope. It's an extension
         if (resultSymbol == null) return functionCall.compose()
```
jsjeon, line 78: I found that the original `calleeReference.resolvedSymbol` (after resolution) is of `FirIntegerOperator` from `FirJAvaModuleBasedSession`, while this, kind of reloaded, result symbol is `FirSimpleFunctionImpl` from `FirLibrarySession`. I'm not sure 1) why the latter was not found as a resolution candidate; and 2) why the former was not reloaded again from the member scope lookup.

jsjeon:
> I'm not sure when/where is a good timing/place to compute compile-time constants.

Noticed that there are many similar cases, which span not only (integer) operators but also arrayOf and property accessors. I'll think of how to handle them altogether.

demiurg906:
Thanks for request. I merged it manually

PR #3389, opened 2020-05-13
[FIR] Create builtin primitive array types for properties

juan-chen:
If the primary constructor has a vararg parameter, the corresponding
property has an array type. This commit creates the builtin array
types for such properties if the vararg element type is primitive,
e.g., CharArray instead of Array<Char>.


juan-chen:
This and pr/3386 together unmutes testInnerExtendsOuter

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/ConversionUtils.kt
```diff
@@ -456,7 +457,12 @@ fun FirPropertyBuilder.generateAccessorsByDelegate(
 fun FirTypeRef.convertToArrayType(): FirUserTypeRef = buildUserTypeRef {
     source = this@convertToArrayType.source
     isMarkedNullable = false
-    qualifier += FirQualifierPartImpl(StandardClassIds.Array.shortClassName).apply {
+    val elementType = this@convertToArrayType as? FirUserTypeRef
+    val arrayQualifier =
+        if (elementType != null && elementType.qualifier.size == 1)
```
dzharkov, line 462: I would say that this change might be not enough.
For example, it's not gonna work in case it has a form of `vararg x: kotlin.Int` or `vararg x: MyTypeAliasToInt`.
Maybe it's worth trying to move this logic to `org.jetbrains.kotlin.fir.resolve.transformers.FirTypeResolveTransformer#transformValueParameter` when the type is already resolved properly and we can just use it without guessing

BTW, something similar is already happening there at `result.transformVarargTypeToArrayType()`.
Looks like we have similar logic in two different places and I would say it's worth investigating which one should work for all cases.

juan-chen:
FirTypeRef.convertToArrayType is now gone.

compiler/fir/analysis-tests/testData/resolve/varargInPrimaryConstructor.txt
```diff
@@ -5,7 +5,7 @@ FILE: varargInPrimaryConstructor.kt
         }
 
         public final val strings: R|kotlin/Array<out kotlin/String>| = R|<local>/strings|
-            public get(): R|kotlin/Array<out kotlin/String>|
+            public get(): R|kotlin/String|
```
dzharkov, line 8: I think this change is not very much correct, getter of the property should also have an array type

compiler/fir/tree/gen/org/jetbrains/kotlin/fir/declarations/FirProperty.kt
```diff
@@ -45,6 +45,7 @@ abstract class FirProperty : FirVariable<FirProperty>(), FirTypeParametersOwner,
     abstract override val symbol: FirPropertySymbol
     abstract val backingFieldSymbol: FirBackingFieldSymbol
     abstract val isLocal: Boolean
+    abstract val isFromVararg: Boolean
```
dzharkov, line 48: From my point-of-view being originated from a vararg parameter is not a substantial feature of a property.
It's really expected to be used and initialized just once while all the implementations should override it and all users will see it in completion.

For such kind of smaller features, you may use `org.jetbrains.kotlin.fir.declarations.FirDeclarationDataRegistry` and `org.jetbrains.kotlin.fir.declarations.FirDeclarationAttributes`

compiler/fir/analysis-tests/testData/resolve/varargInPrimaryConstructor.txt
juan-chen: Fixed getter types

compiler/fir/tree/gen/org/jetbrains/kotlin/fir/declarations/FirProperty.kt
juan-chen: Thanks for the suggestion! I didn't know about FirDeclarationAttributes. It's very handy for the boolean value I need.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
```diff
@@ -107,4 +107,7 @@ fun FirRegularClass.addDeclaration(declaration: FirDeclaration) {
         is FirSealedClassImpl -> declarations += declaration
         else -> throw IllegalStateException()
     }
-}
\ No newline at end of file
+}
+
+object isFromVarargKey: FirDeclarationDataKey()
```
dzharkov, line 112: Please, make this object private if possible and use capitalized names for objects (https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules)

juan-chen: Done. Thanks!

dzharkov:
I've merged your changes manually

PR #3390, opened 2020-05-14
Add samples for String/CharSequence filter & filterNot

segunfamisa:
[KT-35867](https://youtrack.jetbrains.com/issue/KT-35867) - Adds samples to Text `filter()` and `filterNot` functions.

### DIFF
--- a/libraries/stdlib/common/src/generated/_Strings.kt
+++ b/libraries/stdlib/common/src/generated/_Strings.kt
@@ -367,7 +367,7 @@ public inline fun String.dropWhile(predicate: (Char) -> Boolean): String {
 /**
  * Returns a char sequence containing only those characters from the original char sequence that match the given [predicate].
  * 
- * @sample samples.collections.Collections.Filtering.filter
+ * @sample samples.text.Strings.filter
  */
 public inline fun CharSequence.filter(predicate: (Char) -> Boolean): CharSequence {
     return filterTo(StringBuilder(), predicate)
@@ -376,7 +376,7 @@ public inline fun CharSequence.filter(predicate: (Char) -> Boolean): CharSequenc
 /**
  * Returns a string containing only those characters from the original string that match the given [predicate].
  * 
- * @sample samples.collections.Collections.Filtering.filter
+ * @sample samples.text.Strings.filter
  */
 public inline fun String.filter(predicate: (Char) -> Boolean): String {
     return filterTo(StringBuilder(), predicate).toString()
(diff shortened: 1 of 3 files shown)


libraries/stdlib/common/src/generated/_Strings.kt
```diff
@@ -367,7 +367,7 @@ public inline fun String.dropWhile(predicate: (Char) -> Boolean): String {
 /**
  * Returns a char sequence containing only those characters from the original char sequence that match the given [predicate].
  * 
- * @sample samples.collections.Collections.Filtering.filter
+ * @sample samples.text.Strings.filter
```
ilya-g, line 370: Please also change the stdlib generator code, otherwise this change will be overwritten the next time this code is regenerated.
It's located here: https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-stdlib-gen/src/templates/Filtering.kt#L551

segunfamisa: Thanks for the review @ilya-g 

I ran into a problem when applying your comment.

1. When I change the sample in the [line](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-stdlib-gen/src/templates/Filtering.kt#L551) you pointed to, it applied to all the types included in this template (Sequences, Arrays, etc.)

2. When I add the sample lines to the block, like this:

```
specialFor(Strings, CharSequences) {
    sample("samples.text.Strings.filter") // the new added line
    returns("SELF")
    doc { "Returns a ${f.collection} containing only those characters from the original ${f.collection} that match the given [predicate]." }
    body { """return filterTo(StringBuilder(), predicate)${toResult(f)}""" }
}
```
I noticed that two samples were generated. It turns out that in `MemberBuilder.sample()` we add the sample to the existing sample set here:

https://github.com/JetBrains/kotlin/blob/188bcf0e7bc04462ae8fd0c0c31f5f37f62f5ad8/libraries/tools/kotlin-stdlib-gen/src/templates/dsl/MemberBuilder.kt#L137-L139

What do you think about changing `MemberBuilder.sample()` to something like: 
```
fun sample(sampleRef: String, boolean replaceExisting = false ) {
    if (replaceExisting) {
        samples.clear()
    }
    samples += sampleRef
}
```
Or adding a different method (e.g `singleSample()`) that does the same?

Wanted to ask if there's a better way before implementing.

Thank you.

ilya-g: I think it's worth making this the default behavior of `sample` function. I've made such change in PR #3396.

segunfamisa: Okay. I'll revert my changes here then 👍🏽

segunfamisa: Done

ilya-g: That PR is now in master, so you can try to rebase on top of that.

segunfamisa: Done 👍

ilya-g:
Merged. Thank you for the sample.

PR #3392, opened 2020-05-15
FIR: fix default constructor visibility for enum entry.

jsjeon:
Based on `DescriptorUtils#getDefaultConstructorVisibility in core.descriptors`, the visibility for the default constructor is slightly updated to include enum entry as well.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
@@ -520,16 +520,12 @@ class Fir2IrDeclarationStorage(
     ): IrConstructor {
         val descriptor = WrappedClassConstructorDescriptor()
         val isPrimary = constructor.isPrimary
-        val visibility = when (irParent.modality) {
-            Modality.SEALED -> Visibilities.PRIVATE
-            else -> constructor.visibility
-        }
         classifierStorage.preCacheTypeParameters(constructor)
         val created = constructor.convertWithOffsets { startOffset, endOffset ->
             symbolTable.declareConstructor(startOffset, endOffset, origin, descriptor) { symbol ->
                 IrConstructorImpl(
                     startOffset, endOffset, origin, symbol,
-                    Name.special("<init>"), visibility,
+                    Name.special("<init>"), constructor.visibility,
                     constructor.returnTypeRef.toIrType(),
                     isInline = false, isExternal = false, isPrimary = isPrimary, isExpect = false
                 ).apply {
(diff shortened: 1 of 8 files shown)


compiler/fir/tree/src/org/jetbrains/kotlin/fir/FirEffectiveVisibility.kt
```diff
@@ -45,7 +45,8 @@ interface FirEffectiveVisibility {
         }
 
         override fun lowerBound(other: FirEffectiveVisibility): FirEffectiveVisibility {
-            throw AssertionError("Should not be called")
+            // This default object acts as TOP in the lattice, hence by-passing `other` as a result of `meet`.
+            return other
```
jsjeon, line 49: Maybe I'm wrong? Is it rather `BOTTOM`? In either case, for the least lower bound computation, I guess it's safe to return `other` as-is here.

compiler/testData/diagnostics/tests/typealias/typeAliasAsBareType.fir.kt
```diff
@@ -29,7 +29,7 @@ fun testWrong4(x: List<Any>) = x is ReadableList
 
 fun <T> testLocal(x: Any) {
     class C
-    typealias CA = C
+    typealias <!EXPOSED_TYPEALIAS_EXPANDED_TYPE!>CA<!> = C
```
jsjeon, line 32: I have to admit this seems like a regression...

mglukhikh:
Unfortunately I didn't have time yet to dig into this PR. At this moment I can say that changes in raw FIR building here look questionable for me. It seems to me that using `Visibilities.UNKNOWN` is more correct that using of default visibility, because there are some cases when unknown visibility should be replaced with non-default. Will try to spend more time on Wed or Thu.

mglukhikh:
I'd suggest that the main cause of the problem is the fact that anonymous initializer is not visited by some of our transformers, probably by `FirStatusResolveTransformer`. And that's why visibility inside remains unknown.

jsjeon:
> I'd suggest that the main cause of the problem is the fact that anonymous initializer is not visited by some of our transformers, probably by `FirStatusResolveTransformer`. And that's why visibility inside remains unknown.

Thank you for the critical hint! :) I need to polish it a bit more, but you're right; that's where `UNKNOWN` visibility left.

I can still use this PR to clean up default constructor visibility (for enum entry).

compiler/fir/tree/src/org/jetbrains/kotlin/fir/FirEffectiveVisibility.kt
jsjeon: Obsolete comment.

compiler/testData/diagnostics/tests/typealias/typeAliasAsBareType.fir.kt
jsjeon: Ditto.

jsjeon:
Thank you again. Fixing missed visibility places is handled at https://github.com/JetBrains/kotlin/pull/3406, and as mentioned above, this PR is now only dealing with default constructor visibility for enum entry.

mglukhikh:
Thank you! 👍

PR #3395, opened 2020-05-16
FIR: set function factory (before it's used in SAM lowering).

jsjeon:
Otherwise, `SingleAbstractMethodLowering` failed with `UninitializedPropertyAccessException` for that function factory.


jsjeon:
@mglukhikh I've tried to initialize it in `JvmBackendFacade.doGenerateFilesInternal` so that it can be set in a frontend-independent way, but not successful. Perhaps, stub generator and/or IR linker in `psi2ir` depend on it.

mglukhikh:
Ok, let's do it at this point now. May be later we will move this initialization.

PR #3398, opened 2020-05-18
stdlib: DeepRecursiveFunction

elizarov:
Introduces coroutine-based framework to execute deeply recursive functions that utilitie the heap and thus avoid StackOverflowError.

Fixes KT-31741

### DIFF
--- a/libraries/stdlib/src/kotlin/coroutines/CoroutinesIntrinsicsH.kt
+++ b/libraries/stdlib/src/kotlin/coroutines/CoroutinesIntrinsicsH.kt
@@ -44,6 +44,13 @@ public expect inline fun <R, T> (suspend R.() -> T).startCoroutineUninterceptedO
     completion: Continuation<T>
 ): Any?
 
+// Internal version that support arity-2 suspending functions, might be made public in the future if needed
+internal expect inline fun <R, P, T> (suspend R.(P) -> T).startCoroutineUninterceptedOrReturn(
+    receiver: R,
+    param: P,
+    completion: Continuation<T>
+): Any?
+
 @SinceKotlin("1.3")
 public expect fun <T> (suspend () -> T).createCoroutineUnintercepted(
     completion: Continuation<T>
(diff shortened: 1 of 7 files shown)


LifeIsStrange:
I wonder what are the pros and cons (mainly about performance) of storing deepling recursive functions on the heap versus on a growable stack.
The latter is the solution that GCC has implemented ->
https://gcc.gnu.org/wiki/SplitStacks
I don't know what would be the equivalent in JVM semantics

elizarov:
> I wonder what are the pros and cons (mainly about performance) of storing deeply recursive functions on the heap versus on a growable stack.

This is mostly irrelevant w.r.t. this issue since we are running on a JVM that does not support any kind of segmented stacks or any similar solution. If it were, then we would not have a problem of running deeply recursive functions on JVM, but it is not and is not planned to in any conceivable future.

libraries/stdlib/src/kotlin/coroutines/CoroutinesIntrinsicsH.kt
```diff
@@ -44,6 +44,13 @@ public expect inline fun <R, T> (suspend R.() -> T).startCoroutineUninterceptedO
     completion: Continuation<T>
 ): Any?
 
+// Internal version that support arity-2 suspending functions, might be made public in the future if needed
+internal expect inline fun <R, P, T> (suspend R.(P) -> T).startCoroutineUninterceptedOrReturn(
```
udalov, line 48: I'd prefer if we annotated the new internal API with `@SinceKotlin("1.4")` as well, to lessen the probability of accidentally generating access to it in the backend without checking if the API version is at least 1.4.

elizarov: Does the exepect or actual vesion or both need to be annotated this way? Also, it is _inline_ and its actuals are `@InlineOnly`. Does it really matter?

udalov: It's more of a custom than a requirement. Basically if I'm going to implement some new backend feature in 1.4.x and suddenly realize that using this intrinsic will help -- I'll happily use it. But with `@SinceKotlin("1.4")` on it, I'd think some more and either add a requirement on API version, or avoid using this function.

So since it's for documentation purposes only, I'd annotate both expect and actual. The fact that it's inline-only is not important, since the backend needs to generate a call to some intrinsic in the first place anyway (which can fail later during inlining, e.g. if the function body is not found).

PR #3399, opened 2020-05-18
JVM IR: Fix mangling of @JvmStatic internal function in companion objects

sfs:
Right now the same function gets mangled twice.


udalov:
Thank you!

PR #3400, opened 2020-05-18
KT-38961 "Useless call on collection type" for filterNotNull on non-null array where list return type is expected

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38961

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnCollectionInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnCollectionInspection.kt
@@ -7,7 +7,9 @@ package org.jetbrains.kotlin.idea.inspections.collections
 
 import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.openapi.util.TextRange
+import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.idea.quickfix.ReplaceSelectorOfQualifiedExpressionFix
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtQualifiedExpression
 import org.jetbrains.kotlin.psi.psiUtil.endOffset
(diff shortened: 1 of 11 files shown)


idea/src/org/jetbrains/kotlin/idea/inspections/collections/ReplaceUselessCallWithToListFix.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.intellij.codeInspection.LocalQuickFix
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.openapi.project.Project
+import org.jetbrains.kotlin.idea.KotlinBundle
+import org.jetbrains.kotlin.psi.KtPsiFactory
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.createExpressionByPattern
+
+class ReplaceUselessCallWithToListFix : LocalQuickFix {
```
nikitabobko, line 16: I'd generalize this quick fix and call it `ReplaceSelectorOfQualifiedExpressionFix` and put it to `org.jetbrains.kotlin.idea.quickfix` package. Then first line of class would look like:
```
class ReplaceSelectorOfQualifiedExpressionFix(val newSelector: String) : LocalQuickFix
```
I think you get the idea

idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnCollectionInspection.kt
```diff
@@ -77,7 +78,7 @@ class UselessCallOnCollectionInspection : AbstractUselessCallInspection() {
                 KotlinBundle.message("useless.call.on.collection.type"),
                 ProblemHighlightType.LIKE_UNUSED_SYMBOL,
                 isOnTheFly,
-                RemoveUselessCallFix()
+                if (KotlinBuiltIns.isArray(receiverType)) ReplaceUselessCallWithToListFix() else RemoveUselessCallFix()
```
nikitabobko, line 81: That's wrong logic. You should use `ReplaceUselessCallWithToListFix` in case when type of whole expression is `List` and when receiver type isn't `List` at the same time. E.g. consider such example:
```
val arr: List<Int> = setOf(1, 2, 3).filterNotNull()
```
Your quick fix will remove `filterNotNull` but it's expected to be replaced with `toList` just like with `arrayOf`

idea/src/org/jetbrains/kotlin/idea/inspections/collections/UselessCallOnCollectionInspection.kt
```diff
@@ -68,6 +71,11 @@ class UselessCallOnCollectionInspection : AbstractUselessCallInspection() {
             )
             holder.registerProblem(descriptor)
         } else {
+            val fix = if (receiverType.constructor.declarationDescriptor?.fqNameSafe == KotlinBuiltIns.FQ_NAMES.list) {
```
nikitabobko, line 74: You need to check for both conditions. You should replace selector with `toList()` when `whole exp` == `List` && `receiver` != `List`. Now you only check for `receiver` != `List`. So your code doesn't work for `sequenceOf` (because whole expression has type Sequence, not List):
```
val arr: Sequence<Int> = sequenceOf(1, 2, 3).filterNotNull()
```
Yes, we don't yet have "Useless call" check for sequences but your code will broke when we add it.

So I suggest to invert your if for improving readability and add full check.

PR #3402, opened 2020-05-18
[FIR] Use flexible type lower bound when approximating type for IntegerLiteral

punzki:


compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/ConeIntegerLiteralTypeImpl.kt
```diff
@@ -61,9 +61,12 @@ class ConeIntegerLiteralTypeImpl : ConeIntegerLiteralType {
     }
 
     override fun getApproximatedType(expectedType: ConeKotlinType?): ConeClassLikeType {
-        val approximatedType = when (expectedType) {
+        val expectedTypeForApproximation = if (expectedType is ConeFlexibleType) {
+            expectedType.lowerBound
+        } else expectedType
```
juan-chen, line 66: You may use expectedType.lowerBoundIfFlexible()

mglukhikh:
Thank you! 👍

PR #3403, opened 2020-05-19
[FIR] Use vararg element type when generating argument mapping

punzki:


juan-chen:
The unmuted tests are automatically generated. Maybe add updates to GenerateRangesCodegenTestData as well?

mglukhikh:
Thank you! 👍

PR #3404, opened 2020-05-19
FIR: transform resolved arrayOf call to FirArrayOfCall.

jsjeon:
Array creation from collection literals, such as `[1, 2, 3]`, is encoded as `FirArrayOfCall`, and then converted to `IrVararg`. As its name implies, `arrayOf` call can be encoded as `FirArrayOfCall` as well. After the call is resolved, i.e., the type of array is also resolved, we can chime in and replace that call with `FirArrayOfCall`, along with resolved array type. Note that the counter part, collection literals, doesn't have a correct array type yet.

Then, during the conversion, depending on the availability of the array type, we can infer element type and vice versa. Some of utils to get element type from array type and vice versa; utils to build `IrVararg` and `IrVarargElement` are introduced. With these, array creation via `arrayOf` in annotation values now becomes compile-time constants.


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
```diff
@@ -108,3 +108,19 @@ fun IrType.getArrayElementType(irBuiltIns: IrBuiltIns): IrType =
...
             ?: throw AssertionError("Primitive array expected: $classifier")
     }
+
+fun IrType.toArrayType(irBuiltIns: IrBuiltIns): IrType =
+    if (isPrimitiveType()) {
+        irBuiltIns.primitiveArrayForType[this]?.defaultType
+            ?: throw AssertionError("$this not in primitiveArrayForType")
+    } else {
+        IrSimpleTypeBuilder().apply {
+            classifier = irBuiltIns.arrayClass
```
sfs, line 118: This looses nullability information and I think that the return type of `arrayOf` is an invariant array. In this case, you could get away with the slightly shorter: `irBuiltIns.arrayClass.typeWith(this)`.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
```diff
@@ -108,3 +108,19 @@ fun IrType.getArrayElementType(irBuiltIns: IrBuiltIns): IrType =
...
+            arguments = listOf(
+                IrSimpleTypeBuilder().apply {
+                    classifier = this@toArrayType.classifierOrNull!!
+                    variance = Variance.OUT_VARIANCE
+                }.buildTypeProjection()
```
sfs, line 123: Again, I'm not sure if this has to have `OUT_VARIANCE`.

If it does, the call to `classifierOrNull!!` will fail here unless the extension receiver is an `IrSimpleType`, so why not change the signature to `fun IrSimpleType.toArrayType ...` and use `buildTypeProjection { variance = Variance.OUT_VARIANCE }` here?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -760,10 +758,20 @@ class Fir2IrVisitor(
 
     override fun visitArrayOfCall(arrayOfCall: FirArrayOfCall, data: Any?): IrElement {
         return arrayOfCall.convertWithOffsets { startOffset, endOffset ->
+            lateinit var elementType: IrType
+            lateinit var arrayType: IrType
+            if (arrayOfCall.typeRef is FirResolvedTypeRef) {
+                arrayType = arrayOfCall.typeRef.toIrType()
+                elementType = arrayType.getArrayElementType(irBuiltIns)
+            } else {
+                // TODO: this should be the least upper bound of all arguments' types, e.g., ["4", 2u, 0.42f]
```
sfs, line 778: Since this can only happen for annotation arguments, would it be possible to infer the correct type from the context?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -251,19 +251,29 @@ class Fir2IrVisitor(
...
-                            })
+        return varargArgumentsExpression.convertWithOffsets { startOffset, endOffset ->
+            IrVarargImpl(
+                startOffset,
+                endOffset,
+                varargArgumentsExpression.typeRef.toIrType(),
+                varargArgumentsExpression.varargElementType.toIrType(),
+                varargArgumentsExpression.arguments.map { arg ->
+                    convertToIrExpression(arg).run {
+                        convertToIrSpreadElementIfNeeded(arg)
```
sfs, line 262: I think this could be simplified to `convertToIrExpression(arg).convertToIrSpreadElementIfNeeded(arg)`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -772,8 +782,10 @@ class Fir2IrVisitor(
                 startOffset, endOffset,
                 type = arrayType,
                 varargElementType = elementType,
-                elements = arrayOfCall.arguments.map {
-                    convertToIrExpression(it)
+                elements = arrayOfCall.arguments.map { arg ->
+                    convertToIrExpression(arg).run {
+                        convertToIrSpreadElementIfNeeded(arg)
```
sfs, line 788: Same as above

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -251,19 +251,29 @@ class Fir2IrVisitor(
...
+                varargArgumentsExpression.varargElementType.toIrType(),
+                varargArgumentsExpression.arguments.map { arg ->
+                    convertToIrExpression(arg).run {
+                        convertToIrSpreadElementIfNeeded(arg)
+                    }
+                })
+        }
     }
 
+    private fun IrExpression.convertToIrSpreadElementIfNeeded(arg: FirExpression): IrVarargElement =
```
sfs, line 268: Actually, it might be nicer to have a function `convertToIrVarargElement(arg: FirExpression)` which calls `convertToIrExpression` internally. That is assuming that there is no point to passing anything other than the translation of `arg` as the extension receiver.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -187,9 +188,46 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
             }
 
         dataFlowAnalyzer.exitFunctionCall(completeInference, callCompleted)
+        if (callCompleted && completeInference.isArrayOfCall) {
+            return completeInference.toArrayOfCall().compose()
```
sfs, line 192: What happens to calls of the form `arrayOf<Int>(1,2,3)`?

This is supposed to produce an array of boxed integers, but if I read the code correctly, `FirArrayOfCall` will produce primitive arrays depending on the element type and would result in an `IntArray` in this example, instead of an `Array<Int>`.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
jsjeon: Didn't know that simple way, awesome!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -187,9 +188,46 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
...
+    private fun FirFunctionCall.toArrayOfCall(): FirArrayOfCall {
+        assert(isArrayOfCall) {
+            "Unexpected transformation from $this to FirArrayOfCall"
+        }
+        val functionCall = this
+        val typeRef = this.typeRef
+        return buildArrayOfCall {
+            source = functionCall.source
+            annotations += functionCall.annotations
+            // Note that the signature is: arrayOf(varag element). Hence, unwrapping the original argument list here.
```
punzki, line 218: Nit: Typo: `varag` -> `vararg`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
jsjeon: I saw some cases that have `OUT_VARIANCE`, but didn't dig enough to convince myself it has to be. I'll go with the simplest way you suggested above. Thanks!

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
```diff
@@ -90,3 +96,23 @@ fun IrType.isNullable(): Boolean = DFS.ifAny(
...
+fun IrType.getArrayElementType(irBuiltIns: IrBuiltIns): IrType =
+    if (isBoxedArray)
+        ((this as IrSimpleType).arguments.single() as IrTypeProjection).type
+    else {
+        val classifier = this.classOrNull!!
+        irBuiltIns.primitiveArrayElementTypes[classifier]
+            ?: throw AssertionError("Primitive array expected: $classifier")
+    }
+
+fun IrType.toArrayType(irBuiltIns: IrBuiltIns): IrType =
```
punzki, line 112: Suggest renaming to `toArrayOrPrimitiveArrayType`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
jsjeon: Perhaps I should have updated the (high-level) comment about this PR (will do so). As TODO above mentioned, this `else` case is for collection literals (according to other commit messages, or array literals), where we have not set type of the resulting array. So, we should infer the array type from the elements, and I believe it should be the least upper bound of all elements, not the first one. Note that this part was originally there, i.e., out of scope of this PR (I can revisit this later), and I guess, some cases like I mentioned, e.g., ["4", 2] which should have `Array<Any>`, would (silently) fail at this moment. :)

On the other hand, above `if` case corresponds to resolved `arrayOf` call where we know the return type of the call, which is now the array type, and then we rather infer the element type from it. This will be part of an answer to your other question (regarding `Array<Int>` v.s. `IntArray`), but I'll repeat anyway: once the call `arrayOf` is resolved and its return type is resolved as `Array<Int>`, `getArrayElementType` will return `Int` from that boxed array type, so no worries about unexpected conversion.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
jsjeon: Right; I don't know how I got kotlin readability. :p Anyway, removed as per the other suggestion about `convertToIrVarargElement`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
jsjeon: Ditto.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
jsjeon: Indeed! Thanks for the suggestion!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Thank you for asking this question.

Note that `arrayOf` call is resolved, so is its return type. Once it's resolved as boxed array, like `Array<Int>`, somewhere below, `FirArrayOfCall`'s `typeRef` will be also same as boxed array. Then, during the conversion (from `fir` to `ir`), we use array type as-is, and retrieve element type from the array type, which would be boxed `Int`, not primitive `int`, since the given type is boxed array.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Oops, thanks for spotting this.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeUtils.kt
jsjeon: Renamed.

jsjeon:
> I can only speculate about the FIR specific parts. The IR parts look good to me, apart from a small issue with type nullability.

Updated the main comment about this PR. Also, thanks for the comments about type nullability, much simpler now!

> I +1'd some of Steven's comments.

Yes, noticed, and answered.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
jsjeon: I found `IrTypeUtils` has `commonSuperclass`, which is a good fit here.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -187,9 +188,46 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
...
+        }
         return completeInference.compose()
     }
 
+    private val FirFunctionCall.isArrayOfCall: Boolean
+        get() {
+            val function: FirCallableDeclaration<*> = getOriginalFunction() ?: return false
+            return function is FirSimpleFunction &&
+                    // TODO: extend it to other variants, like emptyArray, intArrayOf, floatArrayOf, etc.
+                    !function.name.isSpecial && function.name.identifier == "arrayOf" &&
```
juan-chen, line 202: use name.asString() instead of name.identifier, so you don't have to check isSpecial first?

jsjeon: Right, changed, thanks.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
```diff
@@ -99,7 +99,7 @@ class Fir2IrTypeConverter(
                 intersectedTypes.first().toIrType(typeContext)
             }
             is ConeStubType -> createErrorType()
-            is ConeIntegerLiteralType -> createErrorType()
+            is ConeIntegerLiteralType -> getApproximatedType().toIrType(typeContext)
```
mglukhikh, line 102: Normally we should not have ILT at this stage; if we do, then it's (probably) a front-end resolve issue // cc @demiurg906

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/ir/IrArrayBuilder.kt
```diff
@@ -13,6 +13,8 @@ import org.jetbrains.kotlin.ir.expressions.IrFunctionAccessExpression
 import org.jetbrains.kotlin.ir.expressions.IrGetValue
 import org.jetbrains.kotlin.ir.types.IrType
 import org.jetbrains.kotlin.ir.types.classOrNull
+import org.jetbrains.kotlin.ir.types.getArrayElementType
+import org.jetbrains.kotlin.ir.types.isBoxedArray
```
mglukhikh, line 17: Probably these imports aren't required

jsjeon: The point of this commit (https://github.com/JetBrains/kotlin/pull/3404/commits/465e69683dad06ee5f6ebb7727f640c9df4cfc31) is to migrate `IrType` utils that convert array type to element type (and later the other commit added another util to retrieve element type from array type). These were in the same module, but then migrated to `ir.tree` so that `fir` conversion can use it (when converting `FirArrayOfCall` resolved from `arrayOf` call, which has resolved array type, yet doesn't have element type).

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
jsjeon: Thank you for pointing this out. (Actually I expected you will do so, since I found this one is intentionally generating error type previously.)

This is indeed related to inverted array type and element type for `IrVararg` and my change about computing the common super type of all elements as the array's element type. If I revert this, Ir text generation test with `arrayInAnnotationArguments.kt` fails, complaining that part of `commonSupertype` is error type. In that test, we can see the simple example: `@TestAnnWithIntArray([4, 5, 6])`.

So, even without this change, for `[4, 5, 6]`, we're seeing `FirArrayOfCall` with integer literal arguments. Creating an error type (from retrieving the first element and converting its type) hid the resolution issue, if ILT should not be at this stage.

Nonetheless, I still believe seeing the first element only should be an issue, and the element type should be the least upper bound of all arguments in the array.

jsjeon: Actually, returning a non-error type for ILT would be rather hiding an issue. Since I described the issue above, i.e., an example with the concrete failing test, I can revert this part, and leave some TODOs in element type computation. Updated this commit accordingly.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -187,9 +188,46 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
             }
 
         dataFlowAnalyzer.exitFunctionCall(completeInference, callCompleted)
+        if (callCompleted && completeInference.isArrayOfCall) {
```
mglukhikh, line 191: It seems to me that the following transformation should be done only if we are resolving annotation arguments now. Not quite sure how good or bad is the transformation of a regular `arrayOf` call (not inside an annotation) to `FirArrayOfCall`.

compiler/testData/ir/irText/expressions/vararg.txt
```diff
@@ -12,12 +12,10 @@ FILE fqName:<root> fileName:/vararg.kt
   PROPERTY name:test2 visibility:public modality:FINAL [val]
     FIELD PROPERTY_BACKING_FIELD name:test2 type:kotlin.Array<kotlin.String> visibility:private [final,static]
       EXPRESSION_BODY
-        CALL 'public final fun arrayOf <T> (vararg elements: T of kotlin.arrayOf): kotlin.Array<T of kotlin.arrayOf> [inline] declared in kotlin' type=kotlin.Array<kotlin.String> origin=null
-          <T>: kotlin.String
-          elements: VARARG type=kotlin.Array<out kotlin.String> varargElementType=kotlin.String
-            CONST String type=kotlin.String value="1"
-            CONST String type=kotlin.String value="2"
-            CONST String type=kotlin.String value="3"
+        VARARG type=kotlin.Array<kotlin.String> varargElementType=kotlin.String
```
mglukhikh, line 15: Please pay attention to the fact that this file was "FIR_IDENTICAL", and it means that FE 1.0 (aka old front-end) does not transform regular (not inside annotation arguments) `arrayOf` calls this way you suggest.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Make sense. Added one more commit that migrates transformation-related logic into a separate transformer, then tweak expression transformer a bit to use it only if we're resolving annotation call or default value of annotation constructor (as per two unmuted tests).

compiler/testData/ir/irText/expressions/vararg.txt
jsjeon: Sorry, didn't noticed this one doesn't have `fir` suffix in the file name. Reverted with additional guard regarding when to transform `arrayOf` calls.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Btw, feel free to squash last two commits if you think the intro of `FirArrayOfCallTransformer` and its application to expression transformer should stick together. And, that will remove unnecessary nosies of commit changes in those IR text files. But, hope the separate one is helpful for you to review.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -188,44 +194,14 @@ class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) :
...
-            source = functionCall.source
-            annotations += functionCall.annotations
-            // Note that the signature is: arrayOf(vararg element). Hence, unwrapping the original argument list here.
-            argumentList = buildArgumentList {
-                if (functionCall.arguments.isNotEmpty()) {
-                    (functionCall.argument as FirVarargArgumentsExpression).arguments.forEach {
-                        arguments += it
-                    }
+        if (callCompleted) {
+            if (enableArrayOfCallTransformation) {
```
jsjeon, line 198: Note that I intentionally split these conditions, since I'm working on another transformer (constant evaluator that will transform resolved numeric calls with const expressions into const expression) that will be added around here.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -41,6 +42,13 @@ import org.jetbrains.kotlin.name.Name
 open class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransformer) : FirPartialBodyResolveTransformer(transformer) {
     private var containingClass: FirRegularClass? = null
 
+    private fun <T> withFirArrayOfCallTransformer(block: () -> T): T {
```
demiurg906, line 45: Please make this function `inline` and put `transformer.expressionsTransformer.enableArrayOfCallTransformation = false` into `finally` block so it will be called even if there was a `return` inside `block`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -38,15 +36,23 @@ import org.jetbrains.kotlin.fir.visitors.*
...
 class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransformer) : FirPartialBodyResolveTransformer(transformer) {
     private inline val builtinTypes: BuiltinTypes get() = session.builtinTypes
+    private val arrayOfCallTransformer = FirArrayOfCallTransformer()
+    internal var enableArrayOfCallTransformation = false
 
     init {
         components.callResolver.initTransformer(this)
     }
 
+    private fun <T> withFirArrayOfCallTransformer(block: () -> T): T {
```
demiurg906, line 49: Same

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
jsjeon: Thank you for the suggestions. Done.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Ditto.

mglukhikh:
Finally I merged it, thank you (I squashed two last commits). 👍

PR #3405, opened 2020-05-19
JVM_IR: extract transient state (e.g. the visitor) from ClassCodegen

pyos:
Since JvmBackendContext now holds onto ClassCodegens in order to generate and cache inline method nodes in the same module, any state stored in ClassCodegen directly affects JVM_IR peak memory consumption. In particular, the visitor and the metadata serializer appear to reference objects totaling about half the ClassCodegens' memory consumption in my not very scientific tests.

This also has the benefit of making anything that would be inconsistent under incremental compilation due to inline methods more visible, as generateMethodNode is now statically guaranteed to not touch anything in ClassCodegen.State.

Most of the diff is indentation; the only change that is somewhat of substance is that fields are now generated only after all methods are done to avoid adding even more hacks to generateAssertFieldIfNeeded.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/AnnotationCodegen.kt
```diff
@@ -49,7 +49,7 @@ import org.jetbrains.org.objectweb.asm.*
 import java.lang.annotation.RetentionPolicy
 
 abstract class AnnotationCodegen(
-    private val innerClassConsumer: InnerClassConsumer,
+    private val innerClassConsumer: ClassCodegen,
```
udalov, line 52: Maybe rename it to `classCodegen` then?

pyos:
@udalov ping?

pyos:
@udalov while I've not tried bootstrapping, I think I know what happened: `generateMethodNode` does not cache inline lambdas under the assumption that there is only one call site, but with finally blocks technically there may be multiple uses of the same lambda:

```
fun f() = try { 1 } finally {
    1.let { object {} } // <- f$lambda-1 generated twice, thus ClassCodegen for `XKt$f$1` is needed again it's already marked as no longer needed
}
```

I've reverted `forgetClassCodegen` and the assertions for now because the `finally` handling in codegen needs some more thought.

udalov:
@pyos Sorry, I had to push (a bit overdue) change https://github.com/JetBrains/kotlin/commit/77e479fda836595c71bf956f4c3529f2937298c0 which conflicts a bit with your PR. Could you rebase once again please?

Regarding the change: I find the protected inner class with inner subclasses a bit of a complicated abstraction here. The name State doesn't help either, since some of the state is still stored in ClassCodegen.

Can we instead reverse this and leave all of the current `State`'s logic directly in `ClassCodegen`, but extract the rest to a new top level class `ClassCodegenFacade` (name can be improved), and change all external usages of `ClassCodegen` to `ClassCodegenFacade`? I think it would be a bit clearer but not 100% sure. What do you think?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -41,306 +43,297 @@ import org.jetbrains.org.objectweb.asm.commons.Method
...
+            val clinit = irClass.functions.singleOrNull { it.name.asString() == "<clinit>" }
+            if (clinit != null) {
+                generatingClInit = true
+                generateMethod(clinit)
+            }
+            // 3. Now we have all the fields:
+            for (declaration in irClass.fields) {
+                generateField(declaration)
+            }
+            // 4. Generate inner classes at the end, to ensure that when the companion's metadata is serialized
```
udalov, line 169: The previously used term "nested" is more correct here since it includes classes both with and without `this$0` (see https://kotlinlang.org/docs/reference/nested-classes.html)

pyos:
I initially made this an inner class to avoid making too many things in both ClassCodegen and State (especially `begin`) externally accessible; this is mostly not a problem now that `generateAssertFieldIfNeeded` doesn't actually do much, and `begin` can be removed altogether by making ClassCodegen non-abstract and instead having a separate interface for constructing metadata protobufs (the rest of the logic in DescriptorBasedClassCodegen and FirJvmClassCodegen is in fact the same).

>The name State doesn't help either, since some of the state is still stored in ClassCodegen. [...] a new top level class ClassCodegenFacade (name can be improved)

Not sure about the name. As the PR description says, *ideally* ClassCodegen should not in fact store any *state*. The fact that it keeps track e.g. of regenerated object indices is a bug as it causes instability under incremental compilation.

PR #3406, opened 2020-05-19
FIR: resolve local class in anonymous initializer.

jsjeon:
The current logic to set `local` bit of `ClassId` uses sort of scope notion: stacked function target in the context: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/Context.kt#L21

However, anonymous initializer shouldn't be a return target, and thus are not entered into/exited from the function target. As a result, when visiting a local class in an anonymous initializer, that stack of function target is empty, and concluded that class is non-local, even though we can determine it as `local` from other source of information, e.g., retrieving enclosing element in the source: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt#L677

This change extends the logic in the context to consider other source of information about `local`. If available, that bit is passed when the new scope for a class is stacked, and then popped when we leave the scope.


compiler/testData/codegen/box/localClasses/anonymousObjectInInitializer.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND_FIR: JVM_IR
```
jsjeon, line 1: This one has an unrelated, unresolved call, which is now visible to IR backend. :\

punzki: Which one is it, and if it was not present before, why was the test not failing?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirStatusResolveTransformer.kt
```diff
@@ -168,11 +174,10 @@ private class FirStatusResolveTransformer(
         return transformDeclaration(typeParameter, data)
     }
 
-    override fun transformBlock(block: FirBlock, data: FirDeclarationStatus?): CompositeTransformResult<FirStatement> {
```
punzki, line 171: Why was this taken out?

jsjeon: Because it bails out from the transformation. That is, statements inside a block (in particular, inside the anonymous initializer block) are not transformed by this transformer, if this is still there.

compiler/testData/codegen/box/localClasses/anonymousObjectInInitializer.kt
jsjeon: It was `toString` call with the explicit receiver of anonymous object. I thought that unresolved call was already there, but it turned out that it's due to this change. After investigating when/how the resolution result has been changed, found that indeed the initial change bothered that `toString` to be transformed by `TYPE`-phase transformer; then its return type is not resolved; then from the call site, call completer gave up, since the candidate still has non-resolved type (user-defined type in this example).

On the bright side, by comparing resolution differences before/after the initial change against the motivation test _and_ this regression, I learned about those resolution phases, phase-based transformers, etc., and found that there is a unified place to transform declarations inside local classes. That means, the declarations in question (with `unknown` visibility) were somehow not transformed, and the main reason was the containing class was determined as non-local if it's inside anonymous initializer.

That led me to find where such (should-be-)`local` class was incorrectly determined, and concluded that the current scope-based logic wasn't a good fit. Will update the top-level comment to explain that.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirStatusResolveTransformer.kt
jsjeon: It turns out this is intentionally here to stop this transformer from transforming declarations inside block. Reverted.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/Context.kt
```diff
@@ -18,7 +18,8 @@ import org.jetbrains.kotlin.name.Name
 class Context<T> {
     lateinit var packageFqName: FqName
     var className: FqName = FqName.ROOT
-    val currentClassId get() = ClassId(packageFqName, className, firFunctionTargets.isNotEmpty())
+    fun currentClassId(isLocal: Boolean = false) =
```
punzki, line 21: It's a little odd that the `currentClassId` depends on the caller knowing whether it's currently in a "local context" or not. Is it possible to add an `isLocal` bit to `Context`? This could be set within `BaseFirBuilder.withChildClassName` every time `className` is updated; e.g., change the call to `withChildClassName(className, isLocal)`. You probably need to use a stack of `isLocal` bits, so you can correctly restore when the child is "popped".

Let me know if this makes sense or if it's not possible.

jsjeon: Makes sense. Updated as per the suggestion.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
```diff
@@ -837,7 +837,7 @@ class DeclarationsConverter(
         }
 
         val typeAliasName = identifier.nameAsSafeName()
-        return withChildClassName(typeAliasName) {
+        return withChildClassName(typeAliasName, context.firFunctionTargets.isNotEmpty()) {
```
punzki, line 840: Would it work to make the default value of `isLocal` to be `firFunctionTargets.isNotEmpty()`? Would that work for the other places where you explicitly pass `false`? It's odd to have this logic outside of the `Context` class.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/Context.kt
```diff
@@ -18,7 +18,8 @@ import org.jetbrains.kotlin.name.Name
 class Context<T> {
     lateinit var packageFqName: FqName
     var className: FqName = FqName.ROOT
-    val currentClassId get() = ClassId(packageFqName, className, firFunctionTargets.isNotEmpty())
+    val localBits: MutableList<Boolean> = mutableListOf()
```
punzki, line 21: You could make this `internal` since it is not used outside the package. (In fact, maybe you could even copy `withChildClassName` in here to make `localBits` private, and then leave `withChildClassName` in `BaseFirBuilder` just to delegate it to `Context.withChildClassName`.)

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
jsjeon: > Would it work to make the default value of `isLocal` to be `firFunctionTargets.isNotEmpty()`? Would that work for the other places where you explicitly pass `false`?

Yes, it works. I still believe passing bits explicitly for enum and anonymous object make sense, but at least using the old way of using function targets as a default value made this change as minimal as possible.

> It's odd to have this logic outside of the `Context` class.

I think the whole purpose of `Context` is to make users of it "context-aware" (or "context-sensitive"), which means, it's up to call-site to tweak its logic, depending on under what context it is. So, the current data model makes perfect sense to me, and it's rather odd to have such _guideline_ of how to use context inside `Context`. Please note that `firFunctionTargets` are manually updated outside of `Context` when targetable functions are visited. (And there, again, anonymous initializer is (and should be) missed, which is the motivation of this PR.)

Some additional thoughts. In theory, `Context` should be also immutable so that any independent elements can be handled with different immutable, independent contexts. Well, maybe this mutable data model is also part of what `fir` gains performance against the old frontend, but I wonder, if we start trying to running things in parallel, those mutable data models will bother. Even in the `with` utils about context changes, I saw that, with incorrect manipulation, quite unexpected stack push/pop happened. (That's why I used `removeLast` instead of `pop`, which silently hides stack consistency.) Anyway, it's too early to think about this.

compiler/fir/raw-fir/fir-common/src/org/jetbrains/kotlin/fir/builder/Context.kt
jsjeon: It's reasonable to make it `internal`, but `withChildClassName`, which is `inline`, requires to access only public items. As mentioned at the other comment, I prefer the current data model of `Context`, without any default-ish, guideline logic to manipulate parts of context. Actually, such class-related utils are about creating (raw) `fir` elements (in particular, class), so I think those should belong to the builder(s), not context.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
punzki: > > Would it work to make the default value of `isLocal` to be `firFunctionTargets.isNotEmpty()`? Would that work for the other places where you explicitly pass `false`?
> 
> Yes, it works. I still believe passing bits explicitly for enum and anonymous object make sense, but at least using the old way of using function targets as a default value made this change as minimal as possible.

I agree, I only meant in the other places outside of anonymous object and enum. Current code looks good!

mglukhikh:
Great, thank you!

PR #3407, opened 2020-05-20
KT-38404 Add API for getting annotations and their locations

nerdsupremacist:
***Note** This PR is still a work in progress and needs to be tested properly*

This adds an additional key to ScriptCollectedData to get an array of annotations with their location in the script source code...

This addition allows the Script Developers to emit errors and warnings on the usage of their Annotations directly in the user's code. 

For the purpose of not making this a breaking change, it was added as a separate key in the collected data dictionary. However, if we consider the existence of two APIs to read the annotations to be too complicated, we can decide to make this a breaking change.

Additionally I added the optional location param to `ExternalDependenciesResolver.resolve`, to allow the dependency resolver to emit diagnostics on the precise location as well. This constitutes a breaking change on everyone who implemented their own `ExternalDependenciesResolver`, however this API does not seem to be widely used.

### DIFF
--- a/libraries/examples/scripting/jvm-maven-deps/script/src/org/jetbrains/kotlin/script/examples/jvm/resolve/maven/scriptDef.kt
+++ b/libraries/examples/scripting/jvm-maven-deps/script/src/org/jetbrains/kotlin/script/examples/jvm/resolve/maven/scriptDef.kt
@@ -38,10 +38,10 @@ object ScriptWithMavenDepsConfiguration : ScriptCompilationConfiguration(
 private val resolver = CompoundDependenciesResolver(FileSystemDependenciesResolver(), MavenDependenciesResolver())
 
 fun configureMavenDepsOnAnnotations(context: ScriptConfigurationRefinementContext): ResultWithDiagnostics<ScriptCompilationConfiguration> {
-    val annotations = context.collectedData?.get(ScriptCollectedData.foundAnnotations)?.takeIf { it.isNotEmpty() }
+    val annotations = context.collectedData?.get(ScriptCollectedData.collectedAnnotations)?.takeIf { it.isNotEmpty() }
         ?: return context.compilationConfiguration.asSuccess()
     return runBlocking {
-        resolver.resolveFromAnnotations(annotations)
+        resolver.resolveFromScriptSourceAnnotations(annotations)
     }.onSuccess {
         context.compilationConfiguration.with {
             dependencies.append(JvmDependency(it))
--- a/libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
+++ b/libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
@@ -51,7 +51,29 @@ interface SourceCode {
      * @param end optional range location end position (after the last char)
      */
     data class Location(val start: Position, val end: Position? = null) : Serializable
+
+    /**
+     * The source code location including the path to the file
+     * @param codeLocationId the file path or other script location identifier (see [SourceCode.locationId])
+     * @param locationInText concrete location of the source code in file
+     */
+    data class LocationWithId(val codeLocationId: String, val locationInText: Location) : Serializable
 }
+
+/**
+ * Annotation found during script source parsing along with its location
+ */
+data class ScriptSourceAnnotation<out A : Annotation>(
+    /**
+     * Annotation found during script source parsing
+     */
+    val annotation: A,
+
+    /**
+     * Location of annotation is script
+     */
+    val location: SourceCode.LocationWithId?
+)
 
 /**
  * The interface for the source code located externally
@@ -90,6 +112,13 @@ class ScriptCollectedData(properties: Map<PropertiesCollection.Key<*>, Any>) : P
  * The script file-level annotations found during script source parsing
  */
 val ScriptCollectedDataKeys.foundAnnotations by PropertiesCollection.key<List<Annotation>>()
+
+/**
+ * The script file-level annotations and their locations found during script source parsing
+ */
+val ScriptCollectedDataKeys.collectedAnnotations by PropertiesCollection.key<List<ScriptSourceAnnotation<*>>>(getDefaultValue = {
+    get(ScriptCollectedData.foundAnnotations)?.map { ScriptSourceAnnotation(it, null) }
+})
 
 /**
  * The facade to the script data for compilation configuration refinement callbacks
--- a/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
+++ b/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
@@ -7,6 +7,7 @@ package kotlin.script.experimental.dependencies
 
 import java.io.File
 import kotlin.script.experimental.api.ResultWithDiagnostics
+import kotlin.script.experimental.api.SourceCode
 import kotlin.script.experimental.api.valueOrNull
 
 open class RepositoryCoordinates(val string: String)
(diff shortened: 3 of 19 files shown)


libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -16,7 +17,7 @@ interface ExternalDependenciesResolver {
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
     fun acceptsArtifact(artifactCoordinates: String): Boolean
 
-    suspend fun resolve(artifactCoordinates: String): ResultWithDiagnostics<List<File>>
+    suspend fun resolve(artifactCoordinates: String, location: SourceCode.Location? = null): ResultWithDiagnostics<List<File>>
```
ligee, line 20: I'd name the parameter `sourceCodeLocation` - here it is not obvious from the context.

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
```diff
@@ -53,6 +55,21 @@ interface SourceCode {
     data class Location(val start: Position, val end: Position? = null) : Serializable
 }
 
+/**
+ * Annotation found during script source parsing along with its location
+ */
+data class CollectedScriptAnnotation<out A : Annotation>(
```
ligee, line 61: I do not like the name somehow. I feel to me that something like `ScriptSourceAnnotations` or `SourceCodeAnnotations` will sound better.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
```diff
@@ -33,22 +31,42 @@ annotation class Repository(vararg val repositoriesCoordinates: String)
...
  */
-suspend fun ExternalDependenciesResolver.resolveFromAnnotations(annotations: Iterable<Annotation>): ResultWithDiagnostics<List<File>> {
-    annotations.forEach { annotation ->
-        when (annotation) {
-            is Repository -> {
-                for (coordinates in annotation.repositoriesCoordinates) {
-                    if (!tryAddRepository(coordinates))
-                        return makeFailureResult("Unrecognized repository coordinates: $coordinates")
-                }
+suspend fun ExternalDependenciesResolver.resolveFromCollectedScriptAnnotations(
```
ligee, line 34: See the comment about the class name above, this probably should be renamed too.

plugins/scripting/scripting-compiler/tests/org/jetbrains/kotlin/scripting/compiler/test/CompileTimeFibonacciTest.kt
```diff
@@ -0,0 +1,180 @@
...
+import kotlin.script.experimental.annotations.KotlinScript
+import kotlin.script.experimental.api.*
+import kotlin.script.experimental.host.ScriptingHostConfiguration
+import kotlin.script.experimental.host.toScriptSource
+import kotlin.script.experimental.jvm.*
+
+
+private const val testDataPath = "plugins/scripting/scripting-compiler/testData/compiler/compileTimeFibonacci"
+
+class CompileTimeFibonacciTest : TestCase() {
```
ligee, line 34: I'd like to have a comment here on what this test actually tests. It is not obvious that the main purpose is to test annotation locations.

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
nerdsupremacist: I like `ScriptSourceAnnotation`

libraries/examples/scripting/jvm-maven-deps/script/src/org/jetbrains/kotlin/script/examples/jvm/resolve/maven/scriptDef.kt
```diff
@@ -38,10 +38,10 @@ object ScriptWithMavenDepsConfiguration : ScriptCompilationConfiguration(
 private val resolver = CompoundDependenciesResolver(FileSystemDependenciesResolver(), MavenDependenciesResolver())
 
 fun configureMavenDepsOnAnnotations(context: ScriptConfigurationRefinementContext): ResultWithDiagnostics<ScriptCompilationConfiguration> {
-    val annotations = context.collectedData?.get(ScriptCollectedData.foundAnnotations)?.takeIf { it.isNotEmpty() }
+    val annotations = context.collectedData?.get(ScriptCollectedData.collectedAnnotations)?.takeIf { it.isNotEmpty() }
```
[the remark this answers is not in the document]
nerdsupremacist: I think the easiest way to do this is to include the default value in the `collectedAnnotations` key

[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
```diff
@@ -53,6 +55,21 @@ interface SourceCode {
...
+    /**
+     * Annotation found during script source parsing
+     */
+    val annotation: A,
+
+    /**
+     * Location of annotation is script
+     */
+    val location: SourceCode.Location
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
```diff
@@ -151,3 +173,10 @@ data class ScriptEvaluationConfigurationRefinementContext(
     val evaluationConfiguration: ScriptEvaluationConfiguration,
     val contextData: ScriptEvaluationContextData? = null
 )
+
+inline fun <reified A : Annotation> Iterable<CollectedScriptAnnotation<*>>.filterByAnnotationType(
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
nerdsupremacist: ✅

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
nerdsupremacist: ✅

plugins/scripting/scripting-compiler/tests/org/jetbrains/kotlin/scripting/compiler/test/CompileTimeFibonacciTest.kt
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
```diff
@@ -33,22 +31,42 @@ annotation class Repository(vararg val repositoriesCoordinates: String)
...
+/**
+ * An extension function that configures repositories and resolves artifacts denoted by the [Repository] and [DependsOn] annotations
+ */
+suspend fun ExternalDependenciesResolver.resolveFromAnnotations(
+    annotations: Iterable<Annotation>
+): ResultWithDiagnostics<List<File>> {
+    return annotations.flatMapSuccess { resolveFromAnnotation(it, null) }
+}
+
+private suspend fun ExternalDependenciesResolver.resolveFromAnnotation(
```
ligee, line 51: You changed the logic of the annotation processing here. Before we were processing all `@Repository` annotations first and then all `@DependsOn` and now they are processed in one cycle. Which is also ok and maybe even result in more expected behavior, but if you do this, you need to add tests with `@DependsOn`, `@Repository`, `@DependsOn` sequence, to make sure that resolvers can accept repos after a first resolution. We need this test both for the maven and ivy (main-kts) resolvers.
Or you can return to the old logic, but then please add a TODO that proposes this change with a note about tests, as I described it.

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/resolve/refineCompilationConfiguration.kt
```diff
@@ -354,27 +359,58 @@ fun getScriptCollectedData(
                 jvmGetScriptingClass(ann, contextClassLoader, hostConfiguration) as? KClass<Annotation> // TODO errors
             }
         }.orEmpty()
-    val annotations = scriptFile.annotationEntries.construct(contextClassLoader, acceptedAnnotations, project)
+    val annotations = scriptFile.annotationEntries.construct(
+        contextClassLoader,
+        acceptedAnnotations,
+        project,
+        LineSet.createLineSet(scriptFile.text),
```
ligee, line 366: This is still too expensive, the goal is to reuse the data that already exists in the compiler. I'll send you a working patch separately.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -16,7 +17,11 @@ interface ExternalDependenciesResolver {
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
```
ligee, line 17: And why there is no location here? Maybe it makes sense to add one here too.

nerdsupremacist: Including location in `addRepository` as discussed

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
nerdsupremacist: I actually prefer the previous behavior more. It seems like it just allows for more flexibility, so I'll add a test for ivy and maven that tests that all repositories are added before

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
nerdsupremacist: ✅

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/resolve/refineCompilationConfiguration.kt
nerdsupremacist: Used your patch ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
nerdsupremacist: ✅

nerdsupremacist:
Merged manually

PR #3408, opened 2020-05-20
[JVM] Port line number tests to stepping framework.

madsager:
Allow specifying JVM and JVM_IR as well as shared expectations.

Add the method name to the step.

Discard steps in synthetic methods.


madsager:
This gets rid of all of the non-custom line number tests in favor of stepping tests. I'll do the custom ones next so we have one way of doing line number tests. Then I'll look into the TODOs left in this PR where there are differences between the JVM and JVM_IR backends.

compiler/testData/lineNumber/defaultParameter.kt
```diff
@@ -1,4 +0,0 @@
-class A {
-    fun foo(param: Int = test.lineNumber()) {
```
max-kammerer, line 2: Converted test doesn't test line number on default parameter initialization. It possible to make breakpoint on this line and this behaviour should be somehow tested (maybe in debugger tests if they doesn't have such test yet)

madsager: Yikes, thanks for catching that Mike!

The issue is that foo$default is synthetic and I decided to skip synthetic methods so that the stepping tests operate as going through the code as step-into would. However, if you just do step-into, you do not get the step in the synthetic foo$default method and therefore, we do not check that you can set a breakpoint there.

I have updated the pull request to add a flag for requesting force-step-into behavior where we see the synthetic methods. I have used that flag in the two tests where it makes sense: the two tests with $default methods.

What do you think?

max-kammerer: Thanks! Look OK!

max-kammerer:
NB: As you are working with these test maybe best to replace `TestCase.assertEquals` with `KotlinTestUtils.assertEqualsToFile`. So you could apply changes within compare dialog

max-kammerer:
Thank you!

PR #3412, opened 2020-05-21
FIR: unmute tests after bootstrap compiler update.

jsjeon:


jsjeon:
For the future record, once the bootstrap compiler is updated (in particular, containing incompatible ABI changes), we need to build `stdlib` again (via `gradlew dist`), otherwise, older `stdlib` is used and would see incorrect test results like this.

PR #3413, opened 2020-05-21
Store info if Android plugin is applied in a property

gavra0:
This avoids accessing project.plugins during task action
which reports no applied plugins in the run that has
cached configuration.


nav-nav:
Looks good, I need to run build  to check tests because of typo in branch name

gavra0:
Sorry about that, I can now see there is one `r` too many in the branch name :)

gavra0:
This is fixed by https://github.com/JetBrains/kotlin/commit/a766369e7283d7b89a5ca2960460dc20365f919a, closing this.

PR #3417, opened 2020-05-24
KT-20357 Add samples to kotlin.coroutines package

thumannw:
Some of the samples are heavily inspired by examples found in the [coroutines-examples](https://github.com/Kotlin/coroutines-examples) repository, others are new (e.g. `codeFlow`). In the latter case, my intention was to create samples which clearly show what happens when the code is run in order to get a feeling which part of the code is executed when and where. For me personally, this was an important step to understand the coroutine primitives in the Kotlin standard library.

### DIFF
--- a/libraries/stdlib/samples/test/samples/coroutines/coroutines.kt
+++ b/libraries/stdlib/samples/test/samples/coroutines/coroutines.kt
@@ -9,6 +9,7 @@ import samples.Sample
 import samples.assertPrints
 import java.util.concurrent.CompletableFuture
 import kotlin.coroutines.*
+import kotlin.test.assertTrue
 
 class Coroutines {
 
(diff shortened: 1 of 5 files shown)


libraries/stdlib/samples/test/samples/coroutines/coroutines.kt
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        val receiver = SuspensionScope()
+        block.startCoroutine(receiver, Continuation(EmptyCoroutineContext) {})
+        receiver.continuation!!.resume(Unit)
+        assertPrints(receiver.result, "foo")
+    }
+
+    /**
+     * Unfortunately, this class cannot be a local class in a function because of the companion object.
```
ilya-g, line 94: Unfortunately, this is a blocking obstacle for making this sample runnable in the documentation pages on kotlin website.

thumannw: I have a replacement for this sample which is now
* self-contained and
* also shows addition of context elements

The prototypical definition of a context element using `companion object` and `AbstractCoroutineContextElement` is now embedded in the kdoc of the latter.

PR #3418, opened 2020-05-24
KT-37333 Intention "Convert lambda to reference": NONE_APPLICABLE with overloaded functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37333


yanex:
Relying on a particular parent might be not a good idea:
```
class Random {
    fun nextInt(): Int = 42
    fun nextInt(bound: Int): Int = 42
}

fun main() {
    val random: (Int) -> Int = { n: Int -> Random().nextInt(n) } + { n: Int -> Random().nextInt(n) }
}

operator fun ((Int) -> Int).plus(other: (Int) -> Int): (Int) -> Int {
    return { n -> this(n) + other(n) }
}
```

t-kameyama:
Fixed in 2ecd54a, PTAL.

t-kameyama:
https://github.com/JetBrains/intellij-community/pull/1611

PR #3419, opened 2020-05-25
KT-37744 "Convert lambda to reference" inspection quick fix create incompilable code when type is inferred from lambda parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-37744


yanex:
Failing case 1:

```
fun overloadFun(p: Int) {}
fun overloadFun(p: String) {}

fun <T> ambiguityFun(vararg fn: (T) -> Unit) {}

fun overloadContext() {
    ambiguityFun({ <caret> x: String -> overloadFun(x) }, ::overloadFun)
}
```

yanex:
I would recommend moving out the type parameter check with early returns to avoid clunky constructs such as `?.let {}`, `?.any { it.type.isTypeParameter() } == true)` and others.

t-kameyama:
PTAL

yanex:
Review: OK.

yanex:
Merged into `master`. Thanks!

PR #3421, opened 2020-05-26
WIP: FIR2IR: apply SAM conversion to SamConstructor calls.

jsjeon:
If a resolved call's callee has `SamConstructor` origin, which is synthesized during SAM resolution, that's yet another candidate for type operation call with SAM conversion.


mglukhikh:
@jsjeon Have you seen this commit https://github.com/JetBrains/kotlin/commit/ac1769a4743338972d396e630f5bfa9b277aed67 from @semoro ?

jsjeon:
> @jsjeon Have you seen this commit [ac1769a](https://github.com/JetBrains/kotlin/commit/ac1769a4743338972d396e630f5bfa9b277aed67) from @semoro ?

My repo was synced till last Wed, May 20, and it looks like his commit was created around that time, and submitted yesterday. :'( Thank you for the pointer.

jsjeon:
Good to see that it's quite similar; one difference to note is, maybe it's safer to double-check the target type is indeed SAM type.

PR #3422, opened 2020-05-26
FIR: first step towards compile-time constant evaluation

jsjeon:
Numeric operations, such as `1 + 1`, are resolved as calls, e.g., `1.plus(1)`. As long as all operands are compile-time constants, a.k.a. `FirConstExpression`, such calls can be evaluated at compile-time. The visitor/transformer will naturally visit/transform tree-like formulae, e.g., `4 + 2 * 1` as `4.plus(2.times(1))` and visiting `times` and then `plus`.

The first commit explores naive ideas: 1) reflection-based operation lookup, which requires linear search every time we visit candidate numeric operations; 2) exhaustive type matching, which piggy-backs on compile-time call resolution with smart casts, which doesn't scale, though. At least, this commit serves as a check point, along with handy utils about `FirConstKind` conversions.

Later, after learning old frontend's hashing hack _and_ automatic generation of operations' hash map, the third commit mimics it by generating `fir` version of operations map and rewriting constant evaluator to use such operation map.

In the middle, the second commit applies constant evaluator to annotation serialization, as per the offline discussion. Similarly, the fourth commit applies it to constant initializer for property, only for serialization.

------

Due to the limitations, such as missing overflow check, distinguishing div-by-zero error v.s. infinity, not many tests, in particular, targeting tests (with remaining `div` call inside annotation argument), are muted yet. Also, `Boolean`, `Char`, `String`, and their operations are ignored at the moment. Note that the first commit achieved minimal type casting suppression, but it won't hold when we add types other than `Number`.


compiler/fir/fir-serialization/src/org/jetbrains/kotlin/fir/serialization/FirElementSerializer.kt
```diff
@@ -199,8 +198,8 @@ class FirElementSerializer private constructor(
         var hasGetter = false
         var hasSetter = false
 
-        //val compileTimeConstant = property.compileTimeInitializer
```
jsjeon, line 202: @mglukhikh It looks like FE 1.0 actually has a field in property descriptor and initializes it during field value resolution if the property has a constant. I think I found a counter part in `fir`, but I'm not sure we want to add `compileTimeInitializer` (of `ConstantValue` type perhaps) to `FirProperty` or just determining whether it has constant or not during serialization like this.

compiler/fir/evaluate/src/org/jetbrains/kotlin/fir/evaluate/CompileTimeConstantEvaluator.kt
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    private fun <T : Number> FirConstKind<T>.toConstExpression(source: FirSourceElement?, value: Number): FirConstExpression<T> =
+        @Suppress("UNCHECKED_CAST")
+        buildConstExpression(source, this, value as T)
+
+    // TODO: extension functions, e.g., Byte.and
+    // TODO: toChar?, toString?
+    companion object {
+        private val unaryOperatorNames: Set<String> = setOf(
```
punzki, line 190: Just an idea: Instead of hard-coding (prone to being out-of-sync), you could put both of these as properties in `FirOperationsMapGenerated`, something like:
```kt
internal val unaryOperationsMap = unaryOperations.map { it.opName }.toSet()
```

generators/evaluate/GenerateFirOperationsMap.kt
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+package org.jetbrains.kotlin.generators.evaluate
+
+import org.jetbrains.kotlin.generators.util.GeneratorsFileUtil
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.utils.Printer
+import java.io.File
+
+val FIR_DEST_FILE: File = File("compiler/fir/evaluate/src/org/jetbrains/kotlin/fir/evaluate/FirOperationsMapGenerated.kt")
```
punzki, line 13: Is the only difference between the 2 generators the destination file and excluded functions and types? Is it possible to refactor to reduce duplicated code?

jsjeon: Not really. Types that represent constant kinds are different. Therefore, components of operation keys are also different. In `fir`, as a first step, only numeric types are handled here, and to reduce unchecked cast suppression, `Number` type is used explicitly as many as possible. Since `FirConstKind` is representing only literal kinds, not sure we would end up with really similar generators. But, as I mentioned in the PR description, such kind includes `char` and `string` as well, and extending the evaluator to those types may require many changes as well. Maybe we could think of refactoring at that point.

compiler/fir/evaluate/src/org/jetbrains/kotlin/fir/evaluate/CompileTimeConstantEvaluator.kt
jsjeon: That's good idea, thanks! Changes added to the fifth commit..;;

punzki: Oops I actually meant `unaryOperations.keys.map { ... }` but your code works the same. 🙂

jsjeon:
I've added one more commit: 2345aacb897bd3a0299acddc532aaf95edfaa7de. So, `when` exhaustiveness transformer is missing some boolean exhaustive case due to the lack of constant evaluation. Unlike replacing compile-time constants inside annotation values (which may change source info), this one only affects the generation of the true constant branch in `when`. I wonder the unmuted test in that commit can be unmuted only in that way: using constant evaluation.

compiler/testData/codegen/box/when/exhaustiveBoolean.kt
```diff
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND_FIR: JVM_IR
```
mglukhikh, line 1: This is, well, very side-case, and I'm not sure we should do all of this just to support it. Exhaustive when on boolean is rare by itself, and using expressions instead `true` and `false` there is even more rare. https://youtrack.jetbrains.com/issue/KT-39883 created, to be discussed.

mglukhikh: Another case when IR evaluator can't help us is FIR serializer, also to be discussed.

PR #3423, opened 2020-05-26
JVM_IR: fix a bug when `isInlineParameter` is applied to default stubs

pyos:
If an inline parameter has a default value, its type is nullable. There's already code to handle this in `IrInlineCodegen`, but it really should be in `isInlineParameter` instead, otherwise e.g. SyntheticAccessorLowering fails.


max-kammerer:
Thank you!

PR #3425, opened 2020-05-26
coutLines cleanup

nikitabobko:


idea/testData/refactoring/extractFunction/controlFlow/evaluateExpression/evalWhenExpr.kt.after
```diff
@@ -8,10 +8,8 @@ fun foo(a: Int): Int {
     return i(a, b)
 }
 
-private fun i(a: Int, b: Int): Int {
```
nikitabobko, line 11: You may say that I should leave previous behaviour. But according to https://youtrack.jetbrains.com/issue/KT-6405 for me it seems that we had a bug and new behavior is correct

idea/idea-core/src/org/jetbrains/kotlin/idea/core/util/PsiLinesUtils.kt
```diff
@@ -47,18 +49,26 @@ fun PsiElement.getLineCount(): Int {
     if (doc != null) {
         val spaceRange = textRange ?: TextRange.EMPTY_RANGE
 
-        if (spaceRange.endOffset <= doc.textLength) {
+        if (spaceRange.endOffset <= doc.textLength && spaceRange.startOffset < spaceRange.endOffset) {
             val startLine = doc.getLineNumber(spaceRange.startOffset)
             val endLine = doc.getLineNumber(spaceRange.endOffset)
```
nikitabobko, line 54: BTW, `getLineNumber` JavaDoc is incorrect. It accepts ints from range [0, textLength]. I fixed JavaDoc in community: https://github.com/JetBrains/intellij-community/commit/9e4198f84cdbd3f67f6a8fb94bb3f46a4e9dbb45

idea/idea-core/src/org/jetbrains/kotlin/idea/core/util/PsiLinesUtils.kt
```diff
@@ -47,18 +49,26 @@ fun PsiElement.getLineCount(): Int {
...
-            return endLine - startLine
+            return endLine - startLine + 1
         }
     }
 
     return (text ?: "").count { it == '\n' } + 1
 }
 
-fun PsiElement.isMultiLine(): Boolean = getLineCount() > 1
+fun PsiElement.isMultiLine(): ThreeState {
```
dimonchik0036, line 63: Why introduce `TreeState`? The `UNSURE` value will never occur

idea/idea-core/src/org/jetbrains/kotlin/idea/core/util/PsiLinesUtils.kt
```diff
@@ -47,18 +49,26 @@ fun PsiElement.getLineCount(): Int {
...
+    val i = getLineCount().takeIf { it >= 0 } ?: return ThreeState.UNSURE
+    return ThreeState.fromBoolean(i > 1)
+}
+
+fun PsiElement.isOneLiner(): ThreeState {
```
dimonchik0036, line 68: Same as above

idea/idea-core/src/org/jetbrains/kotlin/idea/core/util/PsiLinesUtils.kt
```diff
@@ -47,18 +49,26 @@ fun PsiElement.getLineCount(): Int {
...
+        if (spaceRange.endOffset <= doc.textLength && spaceRange.startOffset < spaceRange.endOffset) {
             val startLine = doc.getLineNumber(spaceRange.startOffset)
             val endLine = doc.getLineNumber(spaceRange.endOffset)
 
-            return endLine - startLine
+            return endLine - startLine + 1
         }
     }
 
     return (text ?: "").count { it == '\n' } + 1
```
dimonchik0036, line 60: Redundant `:?`

dimonchik0036: It would be correct to use `com.intellij.openapi.util.text.StringUtil#getLineBreakCount` insted of `count`

nikitabobko: > Redundant :?

Not sure. It's `String!`. So for safety there's `?:`

idea/idea-core/src/org/jetbrains/kotlin/idea/core/util/PsiLinesUtils.kt
nikitabobko: Actually it supposed to be `?: return -1` in count lines but I missed it.

After thinking a little bit I removed `ThreeState` because it seems too verbose. Now I throw exception in case `text` is `null`

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantLetInspection.kt
```diff
@@ -249,7 +249,7 @@ private fun KtFunctionLiteral.valueParameterReferences(callExpression: KtCallExp
 private fun isSingleLine(element: KtCallExpression): Boolean {
     val qualifiedExpression = element.getQualifiedExpressionForSelector() ?: return true
     var receiver = qualifiedExpression.receiverExpression
-    if (receiver.lineCount() > 1) return false
+    if (receiver.getLineCount() > 1) return false
```
dimonchik0036, line 252: `isMultiLine()`?

idea/src/org/jetbrains/kotlin/idea/parameterInfo/TypeHints.kt
```diff
@@ -104,7 +104,7 @@ fun provideTypeHint(element: KtCallableDeclaration, offset: Int): List<InlayInfo
         return emptyList()
     }
 
-    if (element is KtProperty && element.isLocal && type.isUnit() && element.lineCount() > 1) {
+    if (element is KtProperty && element.isLocal && type.isUnit() && element.getLineCount() > 1) {
```
dimonchik0036, line 107: `isMultiLine()`?

nikitabobko: Fixed

idea/src/org/jetbrains/kotlin/idea/inspections/RedundantLetInspection.kt
nikitabobko: Fixed

PR #3426, opened 2020-05-26
[JVM IR] Use JVM8 support for unsigned int operations

kandersen:
Use the Java Standard Library functions for particular integer and long operations when available (language version 1.8+). 

Emit a JVM Switch instruction when Kotlin "switching" on UInt using an appropriate `when` statement.

See [YouTrack KT-36838](https://youtrack.jetbrains.com/issue/KT-36838).

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
@@ -43,48 +43,125 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
 
         val calls = callToLabels.map { it.call }
 
-        // Checks if all conditions are CALL EQEQ(tmp_variable, some_constant)
-        if (!areConstComparisons(calls))
-            return null
+        // To generate a switch from a when it must be a comparison of a single
+        // variable, the "subject", against a series of constants. We assume the
+        // subject is the left hand side of the first condition, provided the
+        // first condition is a comparison. If the first condition is of the form:
+        //
+        //     CALL EQEQ(<unsafe-coerce><UInt, Int>(var),_)
+        //
+        // we must be trying to generate an _unsigned_ int switch, and need to
+        // account for unsafe-coerce in all comparisons that arise from the
+        // wrapping and unwrapping of the UInt inline class wrapper. Otherwise,
+        // this is a primitive Int or String switch, with a condition of the form
+        //
+        //     CALL EQEQ(var,_)
 
-        // Subject should be the same for all conditions. Let's pick the first.
-        val subject = callToLabels[0].call.getValueArgument(0)!! as IrGetValue
+        val firstCondition = callToLabels[0].call
+        if (firstCondition.symbol != codegen.classCodegen.context.irBuiltIns.eqeqSymbol) return null
+        val subject = firstCondition.getValueArgument(0)
+        return when {
+            subject is IrCall && subject.isCoerceFromUIntToInt() ->
+                generateUIntSwitch(subject.getValueArgument(0)!! as? IrGetValue, calls, callToLabels, expressionToLabels, elseExpression)
+            subject is IrGetValue ->
+                generatePrimitiveSwitch(subject, calls, callToLabels, expressionToLabels, elseExpression)
+            else ->
+                null
+        }?.genOptimizedIfEnoughCases()
+    }
 
-        // Don't generate repeated cases, which are unreachable but allowed in Kotlin.
-        // Only keep the first encountered case:
-        val cases =
-            callToLabels.map { ValueToLabel((it.call.getValueArgument(1) as IrConst<*>).value, it.label) }.distinctBy { it.value }
+    fun IrCall.isCoerceFromUIntToInt(): Boolean =
+        symbol == codegen.classCodegen.context.ir.symbols.unsafeCoerceIntrinsic
+                && getTypeArgument(0)?.isUInt() == true
+                && getTypeArgument(1)?.isInt() == true
 
-        // Remove labels and "then expressions" that are not reachable.
-        val reachableLabels = HashSet(cases.map { it.label })
-        expressionToLabels.removeIf { it.label !in reachableLabels }
+    private fun generateUIntSwitch(
+        subject: IrGetValue?,
+        conditions: List<IrCall>,
+        callToLabels: ArrayList<CallToLabel>,
+        expressionToLabels: ArrayList<ExpressionToLabel>,
+        elseExpression: IrExpression?
+    ): Switch? {
+        if (subject == null) return null
+        // We check that all conditions are of the form
+        //    CALL EQEQ (<unsafe-coerce><UInt,Int>(subject),
+        //               <unsafe-coerce><UInt,Int>( Constant ))
+        if (!areConstUIntComparisons(conditions)) return null
+
+        // Filter repeated cases. Allowed in Kotlin but unreachable.
+        val cases = callToLabels.map {
+            val constCoercion = it.call.getValueArgument(1)!! as IrCall
+            val constValue = (constCoercion.getValueArgument(0) as IrConst<*>).value
+            ValueToLabel(
+                constValue,
+                it.label
+            )
+        }.distinctBy { it.value }
+
+        expressionToLabels.removeUnreachableLabels(cases)
+
+        return IntSwitch(
+            subject,
+            elseExpression,
+            expressionToLabels,
+            cases
+        )
+    }
+
+    private fun generatePrimitiveSwitch(
+        subject: IrGetValue,
+        conditions: List<IrCall>,
+        callToLabels: ArrayList<CallToLabel>,
+        expressionToLabels: ArrayList<ExpressionToLabel>,
+        elseExpression: IrExpression?
+    ): Switch? {
+        // Checks if all conditions are CALL EQEQ(var,constant)
+        if (!areConstantComparisons(conditions)) return null
 
         return when {
-            areConstIntComparisons(calls) ->
+            areConstIntComparisons(conditions) -> {
+                val cases = extractSwitchCasesAndFilterUnreachableLabels(callToLabels, expressionToLabels)
                 IntSwitch(
                     subject,
                     elseExpression,
                     expressionToLabels,
                     cases
                 )
-            areConstStringComparisons(calls) ->
+            }
+            areConstStringComparisons(conditions) -> {
+                val cases = extractSwitchCasesAndFilterUnreachableLabels(callToLabels, expressionToLabels)
                 StringSwitch(
                     subject,
                     elseExpression,
                     expressionToLabels,
                     cases
                 )
-            else -> null // TODO: Enum, etc.
-        }?.genOptimizedIfEnoughCases()
+            }
+            else ->
+                null
+        }
     }
 
-    private fun areConstComparisons(conditions: List<IrCall>): Boolean {
-        // All branches must be CALL 'EQEQ(Any?, Any?)': Boolean
-        if (conditions.any { it.symbol != codegen.classCodegen.context.irBuiltIns.eqeqSymbol })
-            return false
+    // Check that all conditions are of the form
+    //
+    //  CALL EQEQ (<unsafe-coerce><UInt,Int>(subject), <unsafe-coerce><UInt,Int>( Constant ))
+    //
+    // where subject is taken to be the first variable compared on the left hand side, if any.
+    private fun areConstUIntComparisons(conditions: List<IrCall>): Boolean {
+        val lhs = conditions.map { it.takeIf { it.symbol == codegen.context.irBuiltIns.eqeqSymbol }?.getValueArgument(0) as? IrCall }
+        if (lhs.any { it == null || !it.isCoerceFromUIntToInt() }) return false
+        val lhsVariableAccesses = lhs.map { it!!.getValueArgument(0) as? IrGetValue }
+        if (lhsVariableAccesses.any { it == null || it.symbol != lhsVariableAccesses[0]!!.symbol }) return false
 
-        // All LHS refer to the same tmp variable.
-        val lhs = conditions.map { it.getValueArgument(0) as? IrGetValue }
+        val rhs = conditions.map { it.getValueArgument(1) as? IrCall }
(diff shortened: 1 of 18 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -47,42 +66,96 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+                    && getTypeArgument(1)?.isInt() == true
+
+        val lhs = conditions.map { it.getValueArgument(0) }
+        if (lhs.any { !(it is IrCall && it.isCoerceFromIntToUInt()) }) return false
+        val lhsVariableAccesses = lhs.map { (it as IrCall).getValueArgument(0) as? IrGetValue }
+        if (lhsVariableAccesses.any { it == null || it.symbol != lhsVariableAccesses[0]!!.symbol })
+            return false
+
+        val rhs = conditions.map { it.getValueArgument(1) as IrCall }
+        if (rhs.any { !it.isCoerceFromIntToUInt() }) return false
```
punzki, line 142: You could combine this and the next line (`||` both conditions)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
```diff
@@ -63,6 +65,41 @@ class JvmOptimizationLowering(val context: JvmBackendContext) : FileLoweringPass
...
+            endOffset,
+            target,
+            context.ir.symbols.unsafeCoerceIntrinsic
+        ).also { call ->
+            call.putTypeArgument(0, type)
+            call.putTypeArgument(1, target)
+            call.putValueArgument(0, this)
+        }
+
+    private fun IrCall.replaceWithCallTo(replacement: IrFunctionSymbol) =
```
punzki, line 80: To be clear, the calls being replaced are all static function calls, right? So no need to copy receivers?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -47,42 +66,96 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
         if (!areConstComparisons(calls))
             return null
 
-        // Subject should be the same for all conditions. Let's pick the first.
-        val subject = callToLabels[0].call.getValueArgument(0)!! as IrGetValue
+        return when {
+            areConstUIntComparisons(calls) -> {
+                val subject = callToLabels[0].call.getValueArgumentThroughCoercion(0)!!.asThroughCoercion<IrGetValue>()
```
punzki, line 71: It would help to describe what the expected IR looks like for the call.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
kandersen: That's right! Should I make that more clear?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -122,6 +146,59 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+            )
+        }.distinctBy { it.value }
+        val reachableLabels = HashSet(cases.map { it.label })
+        expressionToLabels.removeIf { it.label !in reachableLabels }
+        return Triple(subject, expressionToLabels, cases)
+    }
+
+    private fun extractSwitchComponents(callToLabels: List<CallToLabel>, expressionToLabels: ArrayList<ExpressionToLabel>): Triple<IrGetValue, ArrayList<ExpressionToLabel>, List<ValueToLabel>> {
+        // Get the
+        val subject = callToLabels[0].call.getValueArgument(0)!! as IrGetValue
```
punzki, line 187: It seems to me the only difference between `extractUIntSwitchComponents` and `extractSwitchComponents` are there is a call to `getValueArgumentThroughCoercion()` for the arguments. But since it falls back to `getValueArgument()` if the call symbol is not the unsafe-coerce, then you could just always call `getValueArgumentThroughCoercion()` in both cases. Basically, I'm suggesting we just keep one version and always call `getValueArgumentThroughCoercion()`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,46 +43,70 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+                val (subject, expressionToLabels, cases) = extractUIntSwitchComponents(callToLabels, expressionToLabels)
+                IntSwitch(
+                    subject,
+                    elseExpression,
+                    expressionToLabels,
+                    cases
+                )
+            }
+            // All remaining switchable types (Int and String) are primitives and can be handled uniformly.
+            !areTmpVariableVSConstantComparisons(calls) ->
```
punzki, line 65: Don't think this is really necessary. It seems at this point you just need to check to see if all RHS are `IrConst`. But to accomplish this, you could just modify `checkTypeSpecifics` similar to what you did for the `UInt` version:

```
val rhs = conditions.map { it.getValueArgument(1) as? IrConst<*> }
if (rhs.any { it == null || !irConstPredicate(it) })
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,46 +43,70 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+            }
             else -> null // TODO: Enum, etc.
         }?.genOptimizedIfEnoughCases()
     }
 
-    private fun areConstComparisons(conditions: List<IrCall>): Boolean {
-        // All branches must be CALL 'EQEQ(Any?, Any?)': Boolean
-        if (conditions.any { it.symbol != codegen.classCodegen.context.irBuiltIns.eqeqSymbol })
-            return false
+    private fun areConstUIntComparisons(conditions: List<IrCall>): Boolean {
```
punzki, line 89: Do we need to check each condition to see if they are unsafe-coerce calls? Could we just check if the LHS is an unsafe-coerce call, and check that the type of the subject (`IrGetValue`) is `UInt`? Are there cases where you would have unsafe-coerce from `UInt` to `Int` on LHS, but _not_ an unsafe-coerce on the RHS?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -122,6 +146,59 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+            return this as T
+        }
+    }
+
+    // Extract the switch subject and calculate labels and expressions of conditions.
+    // Every condition is of the form:
+    //   CALL EQEQ(CALL <unsafe-coerce><UInt,Int>(tmp_variable),
+    //             CALL <unsafe-coerce><UInt,Int>(constant))
+    private fun extractUIntSwitchComponents(callToLabels: List<CallToLabel>, expressionToLabels: ArrayList<ExpressionToLabel>): Triple<IrGetValue, ArrayList<ExpressionToLabel>, List<ValueToLabel>> {
+        val subject = callToLabels[0].call.getValueArgumentThroughCoercion(0)!!.asThroughCoercion<IrGetValue>()
```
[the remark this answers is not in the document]
kandersen: I got rid of the the "throughCoercions". It was an attempt at unifying the logic, but I think I ended a better place.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -43,46 +43,70 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
             else -> null // TODO: Enum, etc.
         }?.genOptimizedIfEnoughCases()
     }
 
-    private fun areConstComparisons(conditions: List<IrCall>): Boolean {
-        // All branches must be CALL 'EQEQ(Any?, Any?)': Boolean
-        if (conditions.any { it.symbol != codegen.classCodegen.context.irBuiltIns.eqeqSymbol })
-            return false
+    private fun areConstUIntComparisons(conditions: List<IrCall>): Boolean {
+        fun IrCall.isCoerceFromIntToUInt(): Boolean =
```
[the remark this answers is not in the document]
kandersen: Good call!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
kandersen: You are right! I rolled it into the previous check, however.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
kandersen: I think the `isCoerce` check on the RHS as more of a precondition for accessing the constant expression that I expect to be there. On the way, it also checks the types.

I can't think of an example off the top of my head that does not conform to your second point, though. I think the front-end would need to put those coercions for it to type-check. Perhaps @sfs can think of a counter?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
kandersen: I went a different route and wrote out the UInt case explicitly as the exception. I feel it was more direct and, as you suggested elsewhere, made the expected IR structure clearer.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
sfs: These `unsafe-coerce` calls are put in by `JvmInlineClassLowering` in [specializeEqualsCall](https://github.com/JetBrains/kotlin/blob/3c8ef5749f3075c29f3a1a30b8a3b9c59cd7c73d/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmInlineClassLowering.kt#L330), so it should be fine. I'd still keep the check here, just in case.

kandersen: @sfs Thanks!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
```diff
@@ -63,6 +65,42 @@ class JvmOptimizationLowering(val context: JvmBackendContext) : FileLoweringPass
...
+    private val Jvm8builtInReplacements = mapOf(
+        ("kotlin.UInt" to "compareTo-WZ4Q5Ns") to context.ir.symbols.compareUnsignedInt,
+        ("kotlin.UInt" to "div-WZ4Q5Ns") to context.ir.symbols.divideUnsignedInt,
+        ("kotlin.UInt" to "rem-WZ4Q5Ns") to context.ir.symbols.remainderUnsignedInt,
+        ("kotlin.UInt" to "toString-impl") to context.ir.symbols.toUnsignedStringInt,
+        ("kotlin.ULong" to "compareTo-VKZWuLQ") to context.ir.symbols.compareUnsignedLong,
+        ("kotlin.ULong" to "div-VKZWuLQ") to context.ir.symbols.divideUnsignedLong,
+        ("kotlin.ULong" to "rem-VKZWuLQ") to context.ir.symbols.remainderUnsignedLong,
+        ("kotlin.ULong" to "toString-impl") to context.ir.symbols.toUnsignedStringLong
+    )
```
udalov, line 102: Can we invoke the name mangling methods here, passing the required signature information as strings, to get strings such as `-WZ4Q5Ns` back without manually hardcoding them here? So that it would be less painful if we change the name mangling scheme once again in the future.

dnpetrov: NB it might require some refactoring to isolate common functionality from inlineClassManglingUtils.kt, where it is implemented in terms of Descriptors and KotlinTypes. It all boils down to building a String from type FQNs and taking base64 MD5.

dnpetrov:
Looks good to me, by the modulo of hardcoded mangled names. Should be not so difficult to fix, and would be useful for some other work related to optimizations and unsigned types.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmOptimizationLowering.kt
kandersen: @udalov We could try something like that! But it would require some reshuffling of utility functions as @dnpetrov points out.

kandersen:
> Looks good to me, by the modulo of hardcoded mangled names. Should be not so difficult to fix, and would be useful for some other work related to optimizations and unsigned types.

Cool, I'll get to it!

kandersen:
@dnpetrov There, shuffled the lowerings around, do take a second look!

dnpetrov:
Ok, LGTM, I'll run tests and push manually.

dnpetrov:
Pushed manually ( f0ff8f202cd3d08f4c9d51ca422b3e0169d417d8 )

kandersen:
Awesome! Thanks for the reviews, all.

PR #3428, opened 2020-05-27
[JVM] Port remaining line number tests to stepping infrastructure.

madsager:
These line number tests only tested that a set of line numbers where
present in the java bytecode. Not that they would be hit in the
right order by the debugger. Moving them to stepping tests fixes that.

This exposes a couple of issues (in particular around try-catch-finally)
that should be fixed.

A number of tests are marked as failing now. Will investigate and
work on fixes next.


madsager:
@max-kammerer I also changed the infrastructure to use KotlinTestUtils.assertEqualsToFile. Thanks for that suggestion, that does make this easier to work with!

max-kammerer:
Thank you!

PR #3429, opened 2020-05-27
JVM IR: Implement the new inline class ABI

sfs:
Apart from top-level functions, methods returning a value of inline class type are now mangled. There are two pain points with the implementation in the IR backend.

- First, property references need a special case to avoid inline class mangling for properties without getters in the bytecode, since the runtime type mapper is incomplete and not aware of inline classes in getters.
- Second, compiler generated methods must sometimes be mangled as well, for example the invoke method in a generated callable reference class. In the long run we probably want to clean this up - either with a nicer API to create functions with correctly mangled names, or by handling all boxing/unboxing manually in compiler generated functions.


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
```diff
@@ -110,7 +110,7 @@ internal fun IrFunction.isInvokeSuspendOfContinuation(): Boolean =
     name.asString() == INVOKE_SUSPEND_METHOD_NAME && parentAsClass.origin == JvmLoweredDeclarationOrigin.CONTINUATION_CLASS
 
 private fun IrFunction.isInvokeOfSuspendCallableReference(): Boolean =
-    isSuspend && name.asString() == "invoke" && parentAsClass.origin == JvmLoweredDeclarationOrigin.FUNCTION_REFERENCE_IMPL
+    isSuspend && name.asString().startsWith("invoke") && parentAsClass.origin == JvmLoweredDeclarationOrigin.FUNCTION_REFERENCE_IMPL
```
dnpetrov, line 113: Shouldn't this condition be more specific? Probably something like `startsWith("invoke-")`, if you meant possibly mangled `invoke` (but not something like `invokeSuspend`, etc).

compiler/testData/codegen/box/ranges/stepped/unsigned/expression/downTo/emptyProgression.kt
```diff
@@ -1,6 +1,4 @@
 // Auto-generated by GenerateSteppedRangesCodegenTestData. Do not edit!
```
dnpetrov, line 1: This test file is auto-generated. Please, update GenerateSteppedRangesCodegenTestData and  GenerateRangesCodegenTestData (I don't see correspodning changes in this PR; sorry if I overlooked something).

mglukhikh: You can (probably) use partial revert of this commit: https://github.com/JetBrains/kotlin/commit/e625bb375f6a3a9d10e9d1aa23355d903c950849#diff-7642cac8c3abb165ed1add8bc6787560R73

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
sfs: Good point, done!

compiler/testData/codegen/box/ranges/stepped/unsigned/expression/downTo/emptyProgression.kt
sfs: Thanks for the pointers! Truth is, I didn't look at the tests, I just unmuted them with `AUTOMATICALLY_UNMUTE_PASSED_TESTS` in `KotlinTestUtils`. It's fixed now.

dnpetrov:
LGTM, I'll run tests and push manually.

dnpetrov:
Pushed manually ( dc0ef996b75d117885ad1f96eaef9e9214c23296 )

PR #3431, opened 2020-05-28
KT-31682 'Convert lambda to reference' intention inside class with function which return object produces uncompilable code

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-31682


yanex:
Review: 👍 .

t-kameyama:
8f3392d635a6b9e91a04e339d1754d0eb341e25c

PR #3432, opened 2020-05-28
FIR: extend arrayOf call transformation to other variants

jsjeon:
At https://github.com/JetBrains/kotlin/pull/3404, only `arrayOf` call is supported. This is a follow-up to extend it to other variants, such as `emptyArray`, `intArrayOf`, `floatArrayOf`, etc.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirArrayOfCallTransformer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirArrayOfCallTransformer.kt
@@ -23,11 +23,8 @@ internal class FirArrayOfCallTransformer {
         get() {
             val function: FirCallableDeclaration<*> = getOriginalFunction() ?: return false
             return function is FirSimpleFunction &&
-                    // TODO: extend it to other variants, like emptyArray, intArrayOf, floatArrayOf, etc.
-                    function.name.asString() == "arrayOf" &&
                     function.returnTypeRef.isArrayType &&
-                    function.valueParameters.size == 1 && function.valueParameters[0].isVararg &&
-                    arguments.size == 1 &&
+                    isArrayOf(function, arguments) &&
                     function.receiverTypeRef == null
         }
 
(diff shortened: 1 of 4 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirArrayOfCallTransformer.kt
```diff
@@ -51,4 +54,19 @@ internal class FirArrayOfCallTransformer {
...
     }
+
+    companion object {
+        private val arrayOfNames = listOf("emptyArray", "arrayOf") +
+                listOf(
+                    "boolean", "byte", "char", "double", "float", "int", "long", "short",
+                    "ubyte", "uint", "ulong", "ushort"
+                ).map { it + "ArrayOf" }
+
+        private fun arrayOfCallParameterCheckers(name: Name): Function2<List<FirValueParameter>, List<FirExpression>, Boolean> =
```
punzki, line 65: Seems unnecessary to create a new lambda instance every time. This could just be:

```kt
private fun isArrayOf(function: FirFunction, arguments: List<FirExpression>): Boolean =
  when (function.name.asString()) {
    "emptyArray" -> function.valueParameters.isEmpty() && arguments.isEmpty()
    in arrayOfNames -> function.valueParameters.size == 1 && function.valueParameters[0].isVararg && arguments.size == 1
    else -> false
  }
```

...or you could just fold this all into `isArrayOfCall` and combine the other conditions.

jsjeon: Thank you for the suggestion! Changed.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirArrayOfCallTransformer.kt
```diff
@@ -23,14 +24,15 @@ internal class FirArrayOfCallTransformer {
         get() {
             val function: FirCallableDeclaration<*> = getOriginalFunction() ?: return false
             return function is FirSimpleFunction &&
-                    // TODO: extend it to other variants, like emptyArray, intArrayOf, floatArrayOf, etc.
-                    function.name.asString() == "arrayOf" &&
+                    function.name.isArrayOf &&
```
punzki, line 27: This check is not necessary since you're already checking the name in `checkParametersAndArguments`. (You also don't need `"emptyArray"` in `arrayOfNames`.)

jsjeon: Aha, you're right. Thanks.

mglukhikh:
Merged 👍

PR #3434, opened 2020-05-29
FIR: introduce serializer extension base for protocol-based platform extensions

jsjeon:
One day, if `fir` is used for other platforms (as well as `JVM`), we may want the serializer extension to deal with annotation serializations in a centralized way, similar to how `SerializerExtensionBase` (under `:compiler:serialization`) does.

### DIFF
--- a/compiler/fir/fir-serialization/src/org/jetbrains/kotlin/fir/serialization/FirSerializerExtensionBase.kt
+++ b/compiler/fir/fir-serialization/src/org/jetbrains/kotlin/fir/serialization/FirSerializerExtensionBase.kt
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.fir.serialization
+
+import org.jetbrains.kotlin.fir.declarations.*
+import org.jetbrains.kotlin.fir.types.ConeKotlinType
+import org.jetbrains.kotlin.metadata.ProtoBuf
+import org.jetbrains.kotlin.metadata.serialization.MutableVersionRequirementTable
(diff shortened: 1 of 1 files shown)


compiler/fir/fir2ir/jvm-backend/src/org/jetbrains/kotlin/fir/backend/jvm/FirJvmSerializerExtension.kt
```diff
@@ -46,7 +47,7 @@ class FirJvmSerializerExtension @JvmOverloads constructor(
     state: GenerationState,
     private val irClass: IrClass,
     private val typeMapper: KotlinTypeMapperBase = state.typeMapper
-) : FirSerializerExtension() {
+) : FirSerializerExtension(BuiltInSerializerProtocol) {
```
udalov, line 50: Looks like this will lead to unnecessary protobuf extensions (defined in `builtins.proto`) being serialized for JVM metadata.

I'd rather prefer if we replicated the hierarchy of `SerializerExtension` with two immediate subtypes `JvmSerializerExtension` and `KotlinSerializerExtensionBase`, with all "serializer protocol"-affected platform extensions as subtypes of the latter.

jsjeon: Thank you for the review! I think I got the point; updated as suggested.

mglukhikh:
Guys, sorry, now I don't understand why this `FirKotlinSerializerExtensionBase` is introduced here and how do you want to use it. As far as I understand, at this moment the class will be unused.

udalov:
Indeed, it doesn't make much sense to introduce this class if we don't intend to add support for JS or Native under FIR in the nearest future. But if there are such plans, then it's required :).

mglukhikh:
Well, I don't know, but I don't expect JS or Native support will be added until August or September. @semoro , what do you think about these perspectives?

PR #3436, opened 2020-05-29
[JVM_IR] Generate line numbers and nops for init blocks.

madsager:
This seems to be what JVM does and it allows you to set a
breakpoint on the init line.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -6,6 +6,7 @@
 package org.jetbrains.kotlin.backend.jvm.codegen
 
 import org.jetbrains.kotlin.backend.common.lower.BOUND_RECEIVER_PARAMETER
+import org.jetbrains.kotlin.backend.common.lower.SYNTHESIZED_INIT_BLOCK
 import org.jetbrains.kotlin.backend.jvm.JvmLoweredDeclarationOrigin
 import org.jetbrains.kotlin.backend.jvm.intrinsics.IrIntrinsicMethods
 import org.jetbrains.kotlin.backend.jvm.intrinsics.JavaClassProperty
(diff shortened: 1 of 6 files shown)


madsager:
The first commit is: https://github.com/JetBrains/kotlin/pull/3428

Second commit is the fix for stepping on init block starts and ends.

max-kammerer:
@madsager Please update `org.jetbrains.kotlinx.serialization.SerializationIrBytecodeListingTestGenerated#testBasic` test

It's not clear how new logic should work with `clinit`s

max-kammerer:
cc @yanex

madsager:
> @madsager Please update `org.jetbrains.kotlinx.serialization.SerializationIrBytecodeListingTestGenerated#testBasic` test

I updated the expectations. I find it strange though that the start line number for the JVM_IR backend is 9 which is where the @Serializable annotation is and not the start of the class.

> It's not clear how new logic should work with `clinit`s
> 
> Could you also add test with class companion (its initialization logic is moved to class clinit)

I manually verified that JVM and JVM_IR generate the same debug information for the example that I have now added with a companion object with all the initialization code in `<clinit>`. Unfortunately, the stepping tests operate as step-into which for this example does not hit `<clinit>` at all. If I set a breakpoint on the first line in `<clinit>` the stepping works. However, the debug infrastructure doesn't have support for that yet. Maybe I should add support for that, but I should probably understand better where the `<clinit>` lines are filtered out for step into. I don't seem to get any events for those lines over the debug protocol at all when asking for step into line stepping.

max-kammerer:
>>    @madsager Please update org.jetbrains.kotlinx.serialization.SerializationIrBytecodeListingTestGenerated#testBasic test

>I updated the expectations. I find it strange though that the start line number for the JVM_IR backend is 9 which is where the @serializable annotation is and not the start of the class.

cc @sandwwraith

max-kammerer:
> Maybe I should add support for that, but I should probably understand better where the <clinit> lines are filtered out for step into. I don't seem to get any events for those lines over the debug protocol at all when asking for step into line stepping.

I don't have any expertise here. Maybe @yanex @goodwinnk @NataliaUkhorskaya  could help

max-kammerer:
> Maybe I should add support for that, but I should probably understand better where the <clinit> lines are filtered out for step into.

Looks like it's minor and could be fixed separately

max-kammerer:
Thank you!

madsager:
Thanks Mike. I'll look into checking the line stepping through `<clinit>` separately. 👍

PR #3437, opened 2020-05-29
Parcelize, JVM IR: Fix types in nested containers

sfs:
This PR makes the JVM IR backend of the Parcelize plugin more strict when it comes to IR types. This was necessary for boxed arrays nested in other container types, which ended up with missing casts. I've also added tests for these cases.

Just in case, I also made the serializer for booleans, chars, and shorts more strict, so that it is less likely to break with future changes to the JVM IR backend.


sfs:
@yanex Please let me know when you'll have time to look at this. I'd really like to get this PR into the next release, since we have some internal code that fails without it.

sfs:
@yanex Ping? (also for #3586)

sfs:
Rebased on master.

@yanex Did you have time to look at this (and #3586) yet?

yanex:
Merged to master manually (https://github.com/JetBrains/kotlin/commit/8ab6411b937df8f6364485f7ad2aea7304d43599). Will be merged to the 1.4.20 branch after M1.

PR #3438, opened 2020-05-29
JVM IR: Use language feature for inline class mangling rules

sfs:


dnpetrov:
LGTM. I'll run tests and merge manually.

PR #3439, opened 2020-05-29
change "// BUNCH: xxx" format

nikitabobko:


compiler/tests/org/jetbrains/kotlin/code/CodeConformanceTest.kt
```diff
@@ -138,7 +138,7 @@ class CodeConformanceTest : TestCase() {
         for (sourceFile in FileUtil.findFilesByMask(SOURCES_BUNCH_FILE_PATTERN, root)) {
             if (EXCLUDED_FILES_AND_DIRS.any { FileUtil.isAncestor(it, sourceFile, false) }) continue
 
-            val matches = Regex("BUNCH: (\\w+)").findAll(sourceFile.readText())
+            val matches = Regex("""BUNCH: (\d+)""").findAll(sourceFile.readText())
```
goodwinnk: There might be bunches like as35

compiler/tests/org/jetbrains/kotlin/code/CodeConformanceTest.kt
```diff
@@ -138,7 +138,7 @@ class CodeConformanceTest : TestCase() {
         for (sourceFile in FileUtil.findFilesByMask(SOURCES_BUNCH_FILE_PATTERN, root)) {
             if (EXCLUDED_FILES_AND_DIRS.any { FileUtil.isAncestor(it, sourceFile, false) }) continue
 
-            val matches = Regex("BUNCH (\\w+)").findAll(sourceFile.readText())
+            val matches = Regex("""BUNCH (\d+)""").findAll(sourceFile.readText())
```
nikitabobko, line 141: Actually it can be even better: """BUNCH \d\d\d""". But for some reason I left older behavior

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/coreApplicationEnvironmentCompat.kt
```diff
@@ -9,7 +9,7 @@ import com.intellij.core.CoreApplicationEnvironment
 import com.intellij.openapi.extensions.ExtensionsArea
 import java.io.File
 
-// BUNCH: 193
+// FIX ME WHEN BUNCH 193 REMOVED
```
goodwinnk, line 12: There are also `// BUNCH 192` already in use. Les's change them as well?

Also I see couple `// BUNCH: as36, 191`. Do you think they can also be modified?

compiler/tests/org/jetbrains/kotlin/code/CodeConformanceTest.kt
nikitabobko: Ah, yes, I missed them. Thanks

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/coreApplicationEnvironmentCompat.kt
nikitabobko: > There are also // BUNCH 192 already in use. Les's change them as well?

You have a sharp eye. Thanks, fixed them as well

> Also I see couple // BUNCH: as36, 191. Do you think they can also be modified?

Fixed

nikitabobko:
Sorry, I've found a little bit more wrongly used BUNCH messages :)

PR #3441, opened 2020-05-29
Use TypeSubstitutor to get the substituted underlying type for inline classes, instead of MemberScope

punzki:
The primary motivation was to fix issues around type-mapping for inline classes in FIR, which uses wrapped descriptors that have empty MemberScopes.


zarechenskiy:
Thank you! From my POV these changes look fine, @mglukhikh take a look and if everything is OK then please merge this PR

mglukhikh:
Merged 👍

PR #3443, opened 2020-06-01
Add info about the end of range in scripting REPL compiler messages

ileasile:


compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+private data class CompilerMessageExtendedLocationImpl(
+    override val path: String,
+    override val line: Int,
+    override val column: Int,
+    override val lineEnd: Int,
+    override val columnEnd: Int,
+    override val lineContent: String?,
+) : Serializable, CompilerMessageExtendedLocation {
```
ligee, line 54: The combination of private and serializable seems suspicious to me. I'd maybe make an interface extending Serializable, although at the moment it seems not needed, and adding it later seems shouldn't break things.

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
```diff
@@ -0,0 +1,70 @@
...
+        val end =
+            if (lineEnd == -1 && columnEnd == -1) ""
+            else if (lineEnd == line) " - $columnEnd"
+            else " - $lineEnd:$columnEnd"
+        val loc = if (start.isEmpty() && end.isEmpty()) "" else " ($start$end)"
+        return path + loc
+    }
+
+    companion object {
+        private const val serialVersionUID: Long = 8228357580L
```
ligee, line 68: With the new serializables, we start now with `serialVersionUID = 0L` and increment. The pseudo-random big numbers like these were used a long time ago and now considered unnecessary.

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollector.kt
```diff
@@ -21,9 +21,21 @@ interface MessageCollector {
 
     fun report(severity: CompilerMessageSeverity, message: String, location: CompilerMessageLocation? = null)
 
+    fun reportExt(severity: CompilerMessageSeverity, message: String, location: CompilerMessageExtendedLocation? = null) {
```
ligee, line 24: I do not like `...Ext` naming - remind me of bad MS APIs habits. I'd rather use `@JvmName` to resolve java ambiguity`.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/impl/errorReporting.kt
```diff
@@ -50,6 +51,37 @@ class ScriptDiagnosticsMessageCollector(private val parentMessageCollector: Mess
     }
 }
 
+class ScriptDiagnosticsMessageExtendedCollector(parentMessageCollector: MessageCollector?) :
```
ligee, line 54: `Extended` part of the name sounds weird to me. It seems that there is no good reason to have a separate "extended" variant, and you can simply add this functionality to the `ScriptDiagnosticsMessageCollector`.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/impl/jvmCompilationUtil.kt
```diff
@@ -71,6 +71,19 @@ inline fun <T> withMessageCollector(
     }
 }
 
+inline fun <T> withMessageExtendedCollector(
```
ligee, line 74: And here too, it seems that we don't need non-extended variant in scripting, so you can just update the `withMessageCollector` function instead.

ileasile:
@ligee Fixed all the comments, but had to convert one file from .java to .kt to overcome naming complications

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/GroupingMessageCollector.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private fun hasExplicitErrors(): Boolean {
+        return groupedMessages.entries().stream()
+            .anyMatch { entry: Map.Entry<CompilerMessageLocation, Message> -> entry.value.severity.isError }
+    }
+
+    private fun hasWarnings(): Boolean {
+        return groupedMessages.entries().stream()
+            .anyMatch { entry: Map.Entry<CompilerMessageLocation, Message> -> entry.value.severity.isWarning }
```
udalov, line 54: If you're converting to Kotlin, please also make the code a bit more idiomatic everywhere. Otherwise there's not much value in converting. For example here something like
```
    private fun hasWarnings(): Boolean = groupedMessages.values().any { it.severity.isWarning }
```

udalov: And also please extract the commit that does conversion

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollector.kt
```diff
@@ -21,9 +21,23 @@ interface MessageCollector {
 
     fun report(severity: CompilerMessageSeverity, message: String, location: CompilerMessageLocation? = null)
 
+    @Suppress("INAPPLICABLE_JVM_NAME")
```
udalov, line 24: Let's please avoid suppressing errors in the compiler codebase. What are you trying to achieve here?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.Serializable
+
+interface CompilerMessageExtendedLocation {
+    val path: String
+    val line: Int
+    val column: Int
+    val lineEnd: Int
+    val columnEnd: Int
+    val lineContent: String?
```
udalov, line 16: Is it possible to store `CompilerMessageLocation` + `lineEnd` + `columnEnd` here?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/GroupingMessageCollector.kt
```diff
@@ -0,0 +1,110 @@
...
+package org.jetbrains.kotlin.cli.common.messages
+
+import com.google.common.collect.LinkedHashMultimap
+import com.google.common.collect.Multimap
+import kotlin.collections.sortedWith
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector.Companion.reportExtended
+import java.util.Comparator
+
+open class GroupingMessageCollector(private val delegate: MessageCollector, private val treatWarningsAsErrors: Boolean) : MessageCollector {
+    // Note that the key in this map can be null
```
udalov, line 25: This comment hints that you might have converted this declaration to Kotlin incorrectly

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/GroupingMessageCollector.kt
ileasile: Do you mean that there should be 2 commits: one for renaming and one for converting? And then commit with changes?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollector.kt
ileasile: I want to have `JvmName` for interface method which is not allowed without suppression. I'd use the name different from `report` here,but @ligee insists on using this name: https://github.com/JetBrains/kotlin/pull/3443#discussion_r433151327

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/GroupingMessageCollector.kt
udalov: Yes

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
ileasile: The current approach seems to be more simple for usage and also more perfomant (because there is no one additional reference). Do you really want to change it?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollector.kt
udalov: Why should this be an interface method though?

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
udalov: I'm just concerned that this is a copy-paste of the internals of `CompilerMessageLocation`, and semantically this class seems to be exactly `CompilerMessageLocation` + line/column end.

As an example, if somebody changes something in `CompilerMessageLocation` in the future, there will be no hint that the corresponding change is also needed here in `CompilerMessageExtendedLocation`. Which is a thing that happens with copy-pasted code all the time.

To keep usages simple, we might still expose these properties like this:
```
val line: Int get() = location.line
val column: Int get() = location.column
```

ileasile: Ok, agree.

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollector.kt
ileasile: Mostly because it should be used in `MessageCollectorBasedReporter#report` which uses instance of MessageCollector, not an instance of a subclass

udalov: OK but why does it need some other JVM name? Why can't it be just `report`? If you're worried about ambiguities in Java code, you can fix them by casting `null` to an appropriate type:
```
messageCollector.report(..., ..., (CompilerMessageLocation) null);
```

udalov: Also, if you annotate it with `@JvmDefault`, you will avoid the need to override it in Java subclasses

ileasile: Added JvmDefault and got rid of error suppression

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/GroupingMessageCollector.kt
ileasile: Got rid of converting to Kotlin

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
ileasile: Fixed

udalov: Thanks! Only now I noticed that `path`, `line`, `column`, `lineContent` do not actually need to be interface members. Can we extract them to extensions? I.e.
```
val CompilerMessageExtendedLocation.path: String
    get() = location.path
```

ileasile: Really so. Extracted :) Thank you

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/CompilerMessageExtendedLocation.kt
ileasile: Removed Serializable

compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/messages/MessageCollectorExtensions.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.cli.common.messages
+
+fun MessageCollector.reportExtendedNullable(
```
ligee, line 8: This doesn't look nice to me.
@udalov, are you sure that it is better to have such boilerplate, instead of using @JvmName + Suppress?

ileasile: It's not because of it. Let me fix it

ileasile: Removed this extension method.

ligee:
After some discussions and coding, I merged another variant of the same change, closer to the original commit. Closing this as obsolete.

PR #3444, opened 2020-06-02
FIR2IR: distinguish substitution case when adding fake overrides for external class

jsjeon:
The motivation is `Accessors.topLevelPropertyAccessors` where the call to `KMutableProperty0.Setter#invoke` is referring to a defined function, which is supposed to be a fake override to `Function1#invoke`. When creating fake overrides for external class, that function is indeed recognized as a fake override, but the scope already picked another fake override, a.k.a. substitution case. Without a proper handling for such case, `fir` currently generates a fake override whose `overriddenSymbol` points to itself again.

Another change in this PR is to visit member class of external class, e.g., `KMutableProperty0.Setter`, which should be external as well, but didn't have a chance to populate fake overrides.


mglukhikh:
Looks good, I will merge it manually a bit later. BTW I've managed to re-use `FakeOverrideGenerator` at this point.

mglukhikh:
Merged manually 👍

PR #3445, opened 2020-06-02
FIR JVM serializer: serialize property signature

jsjeon:
The motivation is `Properties.privateClassVal` where reflection library can't find the accessor for the given property, which uses signature-based look-up. That means, in the associated metadata for the property, signatures are missing, and indeed that's the case. This PR adds `FirMetadataSource.Property`; binds an instance of it with the backing field; and uses it when serializing property.


mglukhikh:
Merged 👍

PR #3446, opened 2020-06-02
Script caching improvements

tunedal:
Makes the hashing a bit more robust and moves the cache directory to the appropriate platform-specific location for caches.


ligee:
Merged manually (with minor changes).

PR #3447, opened 2020-06-02
Polyfills on prototype should be non-enumerable

ilgonmic:
Fix [KT-39244](https://youtrack.jetbrains.com/issue/KT-39244)

On `Object.defineProperty` default value on `enumerable` is `false`


bashor:
LGTM

PR #3448, opened 2020-06-02
JVM IR: Mark lateinit fields as NotNull

sfs:
This is needed for compatibility with the JVM backend.


udalov:
Looks good, thanks!

PR #3450, opened 2020-06-03
[Scripting] Options API for External Dependencies Resolvers

nerdsupremacist:
This is a draft API...

Based on https://github.com/JetBrains/kotlin/pull/3407 to avoid conflicts, but I can rebase it out of the way...
The actual content of this PR is the last two commits

### DIFF
--- a/libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
+++ b/libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
@@ -51,7 +51,29 @@ interface SourceCode {
      * @param end optional range location end position (after the last char)
      */
     data class Location(val start: Position, val end: Position? = null) : Serializable
+
+    /**
+     * The source code location including the path to the file
+     * @param codeLocationId the file path or other script location identifier (see [SourceCode.locationId])
+     * @param locationInText concrete location of the source code in file
+     */
+    data class LocationWithId(val codeLocationId: String, val locationInText: Location) : Serializable
 }
+
+/**
+ * Annotation found during script source parsing along with its location
+ */
+data class ScriptSourceAnnotation<out A : Annotation>(
+    /**
+     * Annotation found during script source parsing
+     */
+    val annotation: A,
+
+    /**
+     * Location of annotation is script
+     */
+    val location: SourceCode.LocationWithId?
+)
 
 /**
  * The interface for the source code located externally
@@ -90,6 +112,13 @@ class ScriptCollectedData(properties: Map<PropertiesCollection.Key<*>, Any>) : P
  * The script file-level annotations found during script source parsing
  */
 val ScriptCollectedDataKeys.foundAnnotations by PropertiesCollection.key<List<Annotation>>()
+
+/**
+ * The script file-level annotations and their locations found during script source parsing
+ */
+val ScriptCollectedDataKeys.collectedAnnotations by PropertiesCollection.key<List<ScriptSourceAnnotation<*>>>(getDefaultValue = {
+    get(ScriptCollectedData.foundAnnotations)?.map { ScriptSourceAnnotation(it, null) }
+})
 
 /**
  * The facade to the script data for compilation configuration refinement callbacks
--- a/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
+++ b/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
@@ -7,6 +7,7 @@ package kotlin.script.experimental.dependencies
 
 import java.io.File
 import kotlin.script.experimental.api.ResultWithDiagnostics
+import kotlin.script.experimental.api.SourceCode
 import kotlin.script.experimental.api.valueOrNull
 
 open class RepositoryCoordinates(val string: String)
@@ -16,23 +17,21 @@ interface ExternalDependenciesResolver {
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
     fun acceptsArtifact(artifactCoordinates: String): Boolean
 
-    suspend fun resolve(artifactCoordinates: String): ResultWithDiagnostics<List<File>>
-    fun addRepository(repositoryCoordinates: RepositoryCoordinates)
+    suspend fun resolve(
+        artifactCoordinates: String,
+        sourceCodeLocation: SourceCode.LocationWithId? = null
+    ): ResultWithDiagnostics<List<File>>
+
+    fun addRepository(
+        repositoryCoordinates: RepositoryCoordinates,
+        sourceCodeLocation: SourceCode.LocationWithId? = null
+    ): ResultWithDiagnostics<Boolean>
 }
 
 fun ExternalDependenciesResolver.acceptsRepository(repositoryCoordinates: String): Boolean =
     acceptsRepository(RepositoryCoordinates(repositoryCoordinates))
 
-fun ExternalDependenciesResolver.addRepository(repositoryCoordinates: String) = addRepository(RepositoryCoordinates(repositoryCoordinates))
-
-suspend fun ExternalDependenciesResolver.tryResolve(artifactCoordinates: String): List<File>? =
-    if (acceptsArtifact(artifactCoordinates)) resolve(artifactCoordinates).valueOrNull() else null
-
-fun ExternalDependenciesResolver.tryAddRepository(repositoryCoordinates: String) =
-    tryAddRepository(RepositoryCoordinates(repositoryCoordinates))
-
-fun ExternalDependenciesResolver.tryAddRepository(repositoryCoordinates: RepositoryCoordinates) =
-    if (acceptsRepository(repositoryCoordinates)) {
-        addRepository(repositoryCoordinates)
-        true
-    } else false
+fun ExternalDependenciesResolver.addRepository(
+    repositoryCoordinates: String,
+    sourceCodeLocation: SourceCode.LocationWithId? = null
+) = addRepository(RepositoryCoordinates(repositoryCoordinates), sourceCodeLocation)
--- a/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
+++ b/libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
@@ -6,24 +6,34 @@
 package kotlin.script.experimental.dependencies
 
 import java.io.File
-import kotlin.script.experimental.api.ResultWithDiagnostics
-import kotlin.script.experimental.api.SourceCode
-import kotlin.script.experimental.api.valueOrNull
+import kotlin.script.experimental.api.*
+import kotlin.script.experimental.dependencies.ExternalDependenciesResolver.Options
 
 open class RepositoryCoordinates(val string: String)
 
 interface ExternalDependenciesResolver {
+    interface Options {
+        object Empty : Options {
+            override fun value(name: String): String? = null
+            override fun flag(name: String): Boolean? = null
+        }
+
+        fun value(name: String): String?
+        fun flag(name: String): Boolean?
+    }
 
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
     fun acceptsArtifact(artifactCoordinates: String): Boolean
 
     suspend fun resolve(
         artifactCoordinates: String,
+        options: Options = Options.Empty,
         sourceCodeLocation: SourceCode.LocationWithId? = null
     ): ResultWithDiagnostics<List<File>>
 
     fun addRepository(
         repositoryCoordinates: RepositoryCoordinates,
+        options: Options = Options.Empty,
         sourceCodeLocation: SourceCode.LocationWithId? = null
     ): ResultWithDiagnostics<Boolean>
 }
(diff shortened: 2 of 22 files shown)


libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/EqualityExternalDependenciesResolverOptionsParser.kt
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Simple lightweight default options parser for external dependency resolvers.
+ *
+ * This parser expects the input to be a series of equality statements:
+ * `foo=Foo bar=Bar`
+ *
+ * And additionally supports flags without any equality statement:
+ * `foo bar`
+ */
+object EqualityExternalDependenciesResolverOptionsParser : ExternalDependenciesResolver.Options.Parser {
```
ligee, line 21: "Equality" is too loaded here, I'd name it "Simple" instead.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/EqualityExternalDependenciesResolverOptionsParser.kt
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package kotlin.script.experimental.dependencies
```
ligee, line 6: Please move to the `impl` package

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -7,16 +7,30 @@ package kotlin.script.experimental.dependencies
...
 import java.io.File
 import kotlin.script.experimental.api.ResultWithDiagnostics
+import kotlin.script.experimental.api.SourceCode
 import kotlin.script.experimental.api.valueOrNull
 
 open class RepositoryCoordinates(val string: String)
 
 interface ExternalDependenciesResolver {
+    interface Options {
+        interface Parser {
```
ligee, line 17: I think that we do not need to expose the flexibility of parsing here. This simply gives an additional coupling between the resolver and annotations, which is not good to my mind.
It seems to me that if user wants to have its own way to parse options, he should rather create a new annotation and use his own helpers that resolve for the annotation, and put the parser he wants there. 
And in our own helpers, we'd better stick to a single consistent (albeit simple) syntax for the options.
This means that the parser interface and parameters should be dropped, the default parser put into the `impl` as an implementation detail and used internally directly.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
```diff
@@ -18,7 +16,7 @@ import kotlin.script.experimental.api.makeFailureResult
 @Target(AnnotationTarget.FILE)
 @Repeatable
 @Retention(AnnotationRetention.SOURCE)
-annotation class DependsOn(vararg val artifactsCoordinates: String)
+annotation class DependsOn(vararg val artifactsCoordinates: String, val options: String = "")
```
ligee, line 19: I sill think that having `val options: Array<String> = []` is better here. It will leave less room for misinterpretation (which separators to use, how to escape strings with spaces, etc.) and should make the parser simpler.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -35,4 +49,37 @@ fun ExternalDependenciesResolver.tryAddRepository(repositoryCoordinates: Reposit
...
+            "1", "true", "t", "yes", "y" -> true
+            "0", "false", "f", "no", "n" -> false
+            else -> null
+        }
+    }
+}
+
+// Empty default ExternalDependenciesResolver.Options
+
+object EmptyExternalDependenciesResolverOptions : ExternalDependenciesResolver.Options {
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -35,4 +49,37 @@ fun ExternalDependenciesResolver.tryAddRepository(repositoryCoordinates: Reposit
...
+fun makeExternalDependenciesResolverOptions(map: Map<String, String>): ExternalDependenciesResolver.Options =
+    MapExternalDependenciesResolverOptions(map)
+
+suspend fun ExternalDependenciesResolver.resolve(
+    artifactCoordinates: String,
+    sourceCodeLocation: SourceCode.LocationInFileSystem? = null,
+    options: Map<String, String>
+): ResultWithDiagnostics<List<File>> = resolve(artifactCoordinates, sourceCodeLocation, makeExternalDependenciesResolverOptions(options))
+
+private class MapExternalDependenciesResolverOptions(
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/EqualityExternalDependenciesResolverOptionsParser.kt
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/EqualityExternalDependenciesResolverOptionsParser.kt
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
nerdsupremacist: Sounds good ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/annotations.kt
nerdsupremacist: My mistake ✅

nerdsupremacist:
I also just changed the behavior of the parser to throw an error when the same options is set twice with different values

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
```diff
@@ -51,8 +52,30 @@ interface SourceCode {
      * @param end optional range location end position (after the last char)
      */
     data class Location(val start: Position, val end: Position? = null) : Serializable
+
+    /**
+     * The source code location including the path to the file
+     * @param filePath path to file with the source code
+     * @param locationInFile concrete location of the source code in file
+     */
+    data class LocationInFileSystem(val filePath: String, val locationInFile: Location) : Serializable
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
```diff
@@ -51,8 +51,30 @@ interface SourceCode {
      * @param end optional range location end position (after the last char)
      */
     data class Location(val start: Position, val end: Position? = null) : Serializable
+
+    /**
+     * The source code location including the path to the file
```
ligee, line 56: "the path to the file" -> "the file path or other script location identifier (see [SourceCode.locationId])"

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/resolve/refineCompilationConfiguration.kt
```diff
@@ -354,27 +361,56 @@ fun getScriptCollectedData(
                 jvmGetScriptingClass(ann, contextClassLoader, hostConfiguration) as? KClass<Annotation> // TODO errors
             }
         }.orEmpty()
-    val annotations = scriptFile.annotationEntries.construct(contextClassLoader, acceptedAnnotations, project)
+    val annotations = scriptFile.annotationEntries.construct(
+        contextClassLoader,
+        acceptedAnnotations,
+        project,
+        DocumentImpl(scriptFile.text),
```
ligee, line 368: I just realized that it could be quite an expensive way to extract coordinates. Sorry for not noticing it before. Could you, please, investigate it deeper, maybe it is actually a preferable and inexpensive way, or maybe there are cheaper ways to get it from the PSI.

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -7,16 +7,31 @@ package kotlin.script.experimental.dependencies
...
+        fun value(name: String): String?
+        fun flag(name: String): Boolean?
+    }
 
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
     fun acceptsArtifact(artifactCoordinates: String): Boolean
 
-    suspend fun resolve(artifactCoordinates: String): ResultWithDiagnostics<List<File>>
+    suspend fun resolve(
+        artifactCoordinates: String,
```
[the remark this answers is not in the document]
nerdsupremacist: I don't recall this being a varargs but it seem like a good idea, and while we're changing the API ¯\_(ツ)_/¯

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/impl/makeExternalDependenciesResolverOptions.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.script.experimental.api.ResultWithDiagnostics
+import kotlin.script.experimental.api.SourceCode
+import kotlin.script.experimental.dependencies.ExternalDependenciesResolver
+
+fun makeExternalDependenciesResolverOptions(map: Map<String, String>): ExternalDependenciesResolver.Options =
+    MapExternalDependenciesResolverOptions(map)
+
+suspend fun ExternalDependenciesResolver.resolve(
+    artifactCoordinates: String,
+    sourceCodeLocation: SourceCode.LocationInFileSystem? = null,
```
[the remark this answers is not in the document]
nerdsupremacist: Good catch!

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/resolve/refineCompilationConfiguration.kt
nerdsupremacist: The PSI only contains the reference to the file and the offsets without position information which makes sense. So we do need to convert the offset into line and column numbers.

I looked it up. I notice that this API is basically just using the LineSet API from IntelliJ, so I decided to move to using it directly, to avoid any of the additional overhead from the DocumentImpl class.

I also looked at the implementation of LineSet and I don't think we're can get any better performance than that tbh. 
I remember writing this exact same logic for my thesis: https://github.com/nerdsupremacist/Graphaello/blob/master/Sources/Graphaello/Processing/Pipeline/Component/1.%20Extraction/SourceCode/LineColumnIndex.swift

libraries/scripting/common/src/kotlin/script/experimental/api/scriptData.kt
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/impl/makeExternalDependenciesResolverOptions.kt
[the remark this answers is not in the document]
nerdsupremacist: ✅

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/impl/SimpleExternalDependenciesResolverOptionsParser.kt
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Simple lightweight options parser for external dependency resolvers.
+ *
+ * This parser expects the input to be a series of equality statements:
+ * `foo=Foo bar=Bar`
+ *
+ * And additionally supports flags without any equality statement:
+ * `foo bar`
+ */
+object SimpleExternalDependenciesResolverOptionsParser {
```
ligee, line 24: Shouldn't it be private/internal too?

nerdsupremacist: My bad 👍 
It's internal now

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/ExternalDependenciesResolver.kt
```diff
@@ -9,35 +9,50 @@ import java.io.File
...
+    }
 
     fun acceptsRepository(repositoryCoordinates: RepositoryCoordinates): Boolean
     fun acceptsArtifact(artifactCoordinates: String): Boolean
 
     suspend fun resolve(
         artifactCoordinates: String,
-        sourceCodeLocation: SourceCode.LocationWithId? = null
+        sourceCodeLocation: SourceCode.LocationWithId? = null,
+        options: Options = Options.Empty
```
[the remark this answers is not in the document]
nerdsupremacist: ✅

nerdsupremacist:
Merged manually

PR #3451, opened 2020-06-03
FIR: construct type with actual type arguments during GetClassCall transformation

jsjeon:
Without using the actual type arguments, `fir` has generated vague type, e.g., `Array<*>`, which bothers class reference comparison.


mglukhikh:
Looks good, thank you! 👍

mglukhikh:
While analyzing test history I noticed this PR broke two `testBareArray` BB tests. I will mute them. Please take a look.

jsjeon:
Oops, I saw them in my earlier trial, and then came up with the current one to handle them. Will take a look at them again.

PR #3454, opened 2020-06-04
FIR: consider the absence of type arguments when constructing GetClassCall type

jsjeon:
Before https://github.com/JetBrains/kotlin/pull/3451, when constructing type of `FirGetClassCall` expression, type arguments were always star projection with the number of type parameters. That is too conservative, but still necessary when type arguments are not given. This PR addresses the regressions that that PR added.


mglukhikh:
Merged 👍

PR #3455, opened 2020-06-05
KT-39393 "Convert anonymous function to lambda expression" intention does not add necessary lambda type parameter for outer function

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39393


igoriakovlev:
Merged. Thank you for your work.

PR #3456, opened 2020-06-05
Make repeat example more expressive

juliankotrba:
This commit adds the zero-based index of current iteration from the passed HOF "action" of the repeat function to its associated code sample.

related to KT-20357

### DIFF
--- a/libraries/stdlib/samples/test/samples/misc/controlFlow.kt
+++ b/libraries/stdlib/samples/test/samples/misc/controlFlow.kt
@@ -16,9 +16,9 @@ class ControlFlow {
             println("Hello")
         }
         
-        // greets with a count
-        repeat(3) { count ->
-            println("Hello with count $count")
+        // greets with an index
+        repeat(3) { index ->
+            println("Hello with index $index")
         }
 
         repeat(0) {
--- a/libraries/stdlib/samples/test/samples/misc/controlFlow.kt
+++ b/libraries/stdlib/samples/test/samples/misc/controlFlow.kt
@@ -13,7 +13,12 @@ class ControlFlow {
     fun repeat() {
         // greets three times
         repeat(3) {
-            println("Hello with count $it")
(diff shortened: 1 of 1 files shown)


juliankotrba:
Hi, since I prefer expressive examples, I have explicitly added the zero-based index of current iteration from the passed HOF `action` (see [repeat](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/repeat.html)) to the example. Any rewording recommendations are appreciated. In case you want to keep it as it is, please let me know, then I will close the PR.

ilya-g:
I believe it would be better to leave the former sample variant as well, because `repeat` is more commonly used without the index parameter. But the proposed addition is valuable too.

juliankotrba:
@ilya-g ok, I guess that's a matter of opinion. Thank you for the quick review though!

juliankotrba:
@ilya-g Just to be sure, is it desired to add the suggested change as an additional example to the sample or leave the sample as it is?

ilya-g:
Sure, you can add your example between the existing two, that's what I've meant.

juliankotrba:
friendly reminder @ilya-g just in case you have not seen the latest commit yet.

libraries/stdlib/samples/test/samples/misc/controlFlow.kt
```diff
@@ -15,9 +15,14 @@ class ControlFlow {
         repeat(3) {
             println("Hello")
         }
+        
+        // greets with a count
+        repeat(3) { count ->
```
ilya-g, line 20: Better to name the lambda parameter `iteration`/`index` or something like that, otherwise it makes an impression that `count` would be equal to 3.

juliankotrba: Done

ilya-g:
It's all good. Merged.

PR #3460, opened 2020-06-09
KT-35357 "Move lambda argument out of parentheses" does not preserve block comments

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-35357


klunnii:
Merged. Thank you!

PR #3462, opened 2020-06-09
[FIR] add support for implementation by delegation

juan-chen:
This commit handles "subclass: super-interface by delegate-expression".

During Psi2Fir, for each delegate, we add to the subclass a synthetic
field (which has type super-interface), and an assignment of the
delegate-expression to the synthetic field in the primary constructor,
so that the delegate-expression can be resolved and transformed along
the way.

During Fir2Ir, we look up delegatable members from the super-interface
and generate corresponding functions/properties for the subclass.

TODO: support for generic delegatable members and generic
super-interface.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
```diff
@@ -0,0 +1,191 @@
...
+ * It assumes a synthetic field with the super-interface type has been created for the delegate expression. It looks for delegatable
+ * methods and properties in the super-interface, and creates corresponding members in the subclass.
+ * TODO: generic super interface types and generic delegated members.
+ */
+internal class DelegatedMemberGenerator(
+    private val components: Fir2IrComponents
+) : Fir2IrComponents by components {
+
+    // Generate delegated members for [subClass]. The synthetic field [irField] has the super interface type.
+    fun generateDelegatedMembers(irField: IrField, subClass: IrClass) {
```
jsjeon, line 42: Nit: naming suggestion: `generate` would be good enough, since the call-site would look like `delegateMemberGenerator.generate(...)`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
```diff
@@ -406,6 +406,9 @@ fun isOverriding(
             // Not checking the field type (they should match each other if everything other match, otherwise it's a compilation error)
             target.name == superCandidate.name
         }
+        target is IrProperty && superCandidate is IrProperty -> {
```
jsjeon, line 409: Perhaps redundant, but just to make sure:
```
// Not checking the property return type (they should match each other if everything other match, otherwise it's a compilation error)
```
Is that right? Then, it'd be good to explicitly mention that as a comment. :)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSpecificTypeResolverTransformer.kt
```diff
@@ -6,6 +6,8 @@
 package org.jetbrains.kotlin.fir.resolve.transformers
 
 import org.jetbrains.kotlin.fir.FirSession
+import org.jetbrains.kotlin.fir.declarations.FirDeclaration
```
mglukhikh, line 9: Redundant?

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
```diff
@@ -7,10 +7,12 @@ package org.jetbrains.kotlin.fir.declarations
 
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Visibilities
+import org.jetbrains.kotlin.fir.declarations.builder.FirClassBuilder
```
mglukhikh, line 10: Redundant?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
```diff
@@ -94,14 +94,31 @@ class Fir2IrConverter(
...
+            if (it.origin != FirDeclarationOrigin.Synthetic) {
+                result.add(it)
+            }
+        }
+        declarations.forEach {
+            if (it.origin == FirDeclarationOrigin.Synthetic) {
+                result.add(it)
+            }
+        }
+        return result
```
jsjeon, line 112: These are readable, so I think it would be okay. I have a couple thoughts, though.

1. groupBy: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/group-by.html

This seems to me like grouping synthetic v.s. non-synthetic members, something like:
```
declarations.groupBy { it.origin == FirDeclarationOrigin.Synthetic }.values.flatten()
```
assuming `false` key is placed ahead (otherwise, well, flipping the condition :) ).

2. materialization

I'm not sure there will be a huge difference or a class with too many declarations, but I wonder we could avoid creating another materialized list here. There are only two uses (above and below), which are simply iteration. So, we could return either `Iterable` or `Sequence` instead?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
```diff
@@ -138,6 +155,13 @@ class Fir2IrConverter(
             is FirProperty -> {
                 declarationStorage.createIrProperty(declaration, parent)
             }
+            is FirField -> {
+                if (declaration.origin == FirDeclarationOrigin.Synthetic) {
```
jsjeon, line 159: This is repeated, and seems worth declaring as an extension property somewhere around here: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt#L42
```
inline val FirDeclaration.isSynthetic: Boolean
    get() = origin == FirDeclarationOrigin.Synthetic
```

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -77,6 +77,14 @@ class Fir2IrVisitor(
         TODO("Should not be here: ${element.render()}")
     }
 
+    override fun visitField(field: FirField, data: Any?): IrField {
+        if (field.origin == FirDeclarationOrigin.Synthetic) {
+            return declarationStorage.getCachedIrField(field)!!
+        } else {
+            TODO("Unexpected field: ${field.render()}")
```
jsjeon, line 84: This (and above one too) looks like an assertion error, not `TODO`. :)

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
```diff
@@ -64,6 +64,14 @@ internal class ClassMemberGenerator(
                     else -> null
                 }
             }
+            irClass.declarations.filter {
```
jsjeon, line 67: Maybe add some comments? I assume:
```
// Add delegated members *before* fake override generations.
// Otherwise, fake overrides for delegated members, which are redundant, will be added.
```
or something like that.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSpecificTypeResolverTransformer.kt
```diff
@@ -6,6 +6,8 @@
...
+import org.jetbrains.kotlin.fir.declarations.FirField
```
jsjeon, line 10: Are these unused import?

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
```diff
@@ -7,10 +7,12 @@ package org.jetbrains.kotlin.fir.declarations
...
 import org.jetbrains.kotlin.fir.declarations.builder.FirRegularClassBuilder
 import org.jetbrains.kotlin.fir.declarations.builder.FirTypeParameterBuilder
 import org.jetbrains.kotlin.fir.declarations.impl.FirRegularClassImpl
 import org.jetbrains.kotlin.fir.declarations.impl.FirFileImpl
+import org.jetbrains.kotlin.fir.expressions.FirExpression
```
jsjeon, line 15: I wonder these are also unused import?

Also, the line below looks not like a related change.

compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
```diff
@@ -511,8 +542,14 @@ class RawFirBuilder(
                 delegatedSuperTypeRef,
                 delegatedSelfTypeRef ?: delegatedSuperTypeRef,
                 owner = this,
-                containerTypeParameters
+                containerTypeParameters,
+                body = if (initializeDelegateStatements.size > 0) buildBlock {
```
jsjeon, line 546: Nit: `initializeDelegateStatements.isNotEmpty()`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
```diff
@@ -0,0 +1,191 @@
...
+                superFunction.visibility,
+                if (superFunction.modality == Modality.ABSTRACT) Modality.OPEN else superFunction.modality,
+                superFunction.returnType,
+                superFunction.isInline,
+                superFunction.isExternal,
+                superFunction.isTailrec,
+                superFunction.isSuspend,
+                superFunction.isOperator,
+                superFunction.isExpect
+            ).apply {
```
jsjeon, line 94: Since you're not using the default factory defined in `SymbolTable.declare*`, I'd like to double-check whether we won't need to set `@Metadata`. Please refer to https://github.com/JetBrains/kotlin/pull/3261

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSpecificTypeResolverTransformer.kt
jsjeon: Ah, already mentioned. :)

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
juan-chen: done

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
juan-chen: done

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSpecificTypeResolverTransformer.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/ClassMemberGenerator.kt
juan-chen: done

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
juan-chen: done

compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
juan-chen: Not sure about this: I did use SymbolTable.declareSimpleFunction?

compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
```diff
@@ -444,6 +447,8 @@ class RawFirBuilder(
         ): FirTypeRef {
             var superTypeCallEntry: KtSuperTypeCallEntry? = null
             var delegatedSuperTypeRef: FirTypeRef? = null
+            var delegateNumber = 0
```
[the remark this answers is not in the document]
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
```diff
@@ -94,14 +94,31 @@ class Fir2IrConverter(
         return irClass
     }
 
+    // Sort declarations so that all non-synthetic declarations are before synthetic ones.
+    // This is needed because converting synthtic fields for implementation delegation needs to know
+    // existing declarations in the class to avoid adding redundant delegated members.
+    private fun putSyntheticLast(declarations: List<FirDeclaration>) : List<FirDeclaration> {
```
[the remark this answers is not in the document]
juan-chen: Thanks for the suggestion, much simpler than mine!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
```diff
@@ -94,14 +94,31 @@ class Fir2IrConverter(
         return irClass
     }
 
+    // Sort declarations so that all non-synthetic declarations are before synthetic ones.
+    // This is needed because converting synthtic fields for implementation delegation needs to know
```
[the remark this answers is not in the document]
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
jsjeon: Yes, you're using `SymbolTable.declareSimpleFunction` with your own `functionFactory`: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt#L745

Okay, now there is no more default function factory at all. So you should set metadata info by yourself, like: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt#L482

When I mentioned my previous PR, around that time, `SymbolTable.declare*`s have default IR declaration factory where metadata is set. But, in `fir2ir`, we almost always pass our own factory without metadata setting. So, that's why I wondered the use of `declareSimpleFunction` with own function factory still misses metadata setting. Since default function factory is gone, it seems it's up to caller's responsibility to property set metadata source.

mglukhikh:
@juan-chen , so is it still a draft or not?

jsjeon:
> @juan-chen , so is it still a draft or not?

I just asked her to double-check one last thing: metadata setting for those delegated members.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
juan-chen: done.

mglukhikh:
@juan-chen , as I can see light tree tests (see `fir:raw-fir:light-tree2fir` module) are broken here. Also, `RawFirBuilderTotalKotlinTestCase` is broken.

mglukhikh:
Also, it influences (breaks test data) in `FirDiagnosticsSpecTestGenerated`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -136,6 +136,11 @@ class Fir2IrClassifierStorage(
         }
     }
 
+    internal fun findFirClass(irClass: IrClass): FirClass<*>? {
```
mglukhikh, line 139: This is very nasty function, I recommend to get rid of it

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
```diff
@@ -0,0 +1,260 @@
...
+                        this.isVar = firSuperProperty.isVar
+                    }
+                )
+                subClass.addMember(this)
+            }
+        }
+    }
+
+    private fun getCallableId(irClass: IrClass, name: Name): CallableId {
+        val classId = classifierStorage.findFirClass(irClass)?.classId
```
mglukhikh, line 257: You can take `classId` directly from `irClass` without using nasty `findFirClass`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/DelegatedMemberGenerator.kt
```diff
@@ -0,0 +1,260 @@
...
+                val function = generateDelegatedFunction(subClass, irField, it, declarationStorage.findOverriddenFirFunction(it))
+                subClass.addMember(function)
+            } else if (it is IrProperty) {
+                generateDelegatedProperty(subClass, irField, it, declarationStorage.findOverriddenFirProperty(it))
+            }
+        }
+    }
+
+    private fun IrDeclaration.isDelegatable(): Boolean {
+        // TODO: replace checks on descriptors
```
mglukhikh, line 70: Please do not add descriptor usages. Today evening I will push the opt-in for descriptors in IR, and after it you will have a compiler error here. I think you can do all of it using IR elements themselves

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConverter.kt
```diff
@@ -95,14 +95,21 @@ class Fir2IrConverter(
         return irClass
     }
 
+    // Sort declarations so that all non-synthetic declarations are before synthetic ones.
+    // This is needed because converting synthetic fields for implementation delegation needs to know
+    // existing declarations in the class to avoid adding redundant delegated members.
+    private fun putSyntheticLast(declarations: List<FirDeclaration>) : Iterable<FirDeclaration> {
```
mglukhikh, line 101: I don't like name, `put` is associated with adding something into container. Consider something like `deprioritizeSynthetics` or `sortedBySynthetic`

mglukhikh:
NB: continued in #3480

mglukhikh:
Finally I merged it, thank you!

PR #3463, opened 2020-06-09
KT-39263 False positive "Variable should be inlined" for override value in  initialized object

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39263


darthorimar:
@t-kameyama thank you!

PR #3464, opened 2020-06-09
FIR: load and convert annotations of property accessors in Library

jsjeon:
Many reflection tests are failing due to the wrong signature use for the property accessor in Library, in particular, getter with `@JvmName` annotation. (IR backend determines the signature by looking up that specific annotation.)

The first commit fixes the deserializer to build property accessors if non-default ones exist. Previously, default accessors have been always created.

The second commit extends the deserializer to deserialize/load annotations for property accessors, according to their signatures. For example,
```
val java ...
    @JvmName("getJavaClass")
    get() ...
```
that annotation is deserialized to `getJavaClass` method, not `getJava`. Thus, after loading all members' annotations, we need to look up the correct member via its signature. To avoid multiple loading of member annotations, this change adds a cache as well.

Finally, the third commit eagerly converts those annotations in Library members when creating containing IR elements. Note that, for now, only annotations in user-contributed classes and members have a chance to be converted.


jsjeon:
@mglukhikh @semoro These are changes I mentioned today (or yesterday, depending on the time zones) about deserialization of `@JvmName` and eager population.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -262,6 +263,9 @@ class Fir2IrDeclarationStorage(
                 irDeclaration.parent = irClass
             }
         }
+        if (regularClass.origin == FirDeclarationOrigin.Library) {
```
juan-chen, line 266: Seems the pattern is repeated several times (except for the propertyAccessor case). Maybe adding a util function?

jsjeon: Yes, I noticed that I did more repetition while reviewing your PR. :D Updated, thanks!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -53,6 +51,9 @@ class Fir2IrDeclarationStorage(
     conversionScope: Fir2IrConversionScope,
     fakeOverrideMode: FakeOverrideMode
 ) : Fir2IrComponents by components {
+
+    internal var annotationGenerator: AnnotationGenerator? = null
```
juan-chen, line 55: use "lateinit" instead of initializing with null and then null-check later?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -262,6 +263,9 @@ class Fir2IrDeclarationStorage(
                 irDeclaration.parent = irClass
             }
         }
+        if (regularClass.isFromLibrary) {
+            annotationGenerator?.generate(irClass, regularClass)
+        }
```
juan-chen, line 268: Should have been more clear in my earlier comment: I suggested a util function for the three-line pattern

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
jsjeon: Yeah, I tried that as well; Alas, for built-in classes, this storage is used even before the visitor is created (which is used to initialize `AnnotationGenerator`).

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/deserialization/FirMemberDeserializer.kt
```diff
@@ -172,13 +173,79 @@ class FirMemberDeserializer(private val c: FirDeserializationContext) {
...
+                        c.annotationDeserializer.loadPropertySetterAnnotations(c.containerSource, proto, local.nameResolver, setterFlags)
+                    this.symbol = FirPropertyAccessorSymbol()
+                    valueParameters += proto.setterValueParameter.let {
+                        val parameterFlags = if (it.hasFlags()) it.flags else 0
+                        buildValueParameter {
+                            session = c.session
+                            origin = FirDeclarationOrigin.Library
+                            this.returnTypeRef = returnTypeRef
+                            name = if (it.hasName()) c.nameResolver.getName(it.name) else Name.special("<default-setter-parameter>")
+                            this.symbol = FirVariableSymbol(CallableId(FqName.ROOT, name))
```
juan-chen, line 230: Any reason to give a package name "FqName.ROOT", instead of just FirVariableSymbol(name)?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
jsjeon: Aha, make sense. Thanks for the suggestion.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/deserialization/FirMemberDeserializer.kt
jsjeon: I referred to this: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/impl/FirDefaultPropertyAccessor.kt#L112 but to be honest, no idea why. Just tested, and using either doesn't show any differences. Any idea @mglukhikh ?

Btw, note that `FirVariableSymbol(name)` will use `CallableId(name)` creation, which is deprecated.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/deserialization/JvmBinaryAnnotationDeserializer.kt
```diff
@@ -5,18 +5,160 @@
...
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.protobuf.MessageLite
+import org.jetbrains.kotlin.serialization.deserialization.descriptors.DeserializedContainerSource
 
 class JvmBinaryAnnotationDeserializer(
-    session: FirSession
+    val session: FirSession
 ) : AbstractAnnotationDeserializer(session) {
+    private val storage: MutableMap<KotlinJvmBinaryClass, Storage> = mutableMapOf()
```
juan-chen, line 30: Nit: the names "storage" and "Storage" seem a bit confusing to me, as they mean different things. Maybe "MemberAnnotations" instead of "Storage"?

jsjeon: Right; since it represents only member annotations for now, that name makes sense. Thanks for the suggestion.

mglukhikh:
Merged, thank you! 👍

PR #3466, opened 2020-06-10
KT-38819 False positive "Unused symbol" for class, with inner class, type parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-38819


vladimirdolzhenko:
Thank you so much, merged manually into https://github.com/JetBrains/intellij-kotlin

PR #3467, opened 2020-06-10
[Scripting] KT-35411 Explicitly handle array annotation args in scripting pre-processing

nerdsupremacist:
This is the same PR as #3337.
With some additional tests to make sure we don't break this functionality. And a fix.

### DIFF
--- a/compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
+++ b/compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
@@ -63,8 +63,8 @@ private interface ArgsConverter<T> {
     }
 
     fun tryConvertSingle(parameter: KParameter, arg: NamedArgument<T>): Result
-    fun tryConvertVararg(parameter: KParameter, firstArg: NamedArgument<T>, restArgsIt: Iterator<NamedArgument<T>>): Result
-    fun tryConvertTail(parameter: KParameter, firstArg: NamedArgument<T>, restArgsIt: Iterator<NamedArgument<T>>): Result
+    fun tryConvertVararg(parameter: KParameter, firstArg: NamedArgument<T>, restArgs: Sequence<NamedArgument<T>>): Result
+    fun tryConvertTail(parameter: KParameter, firstArg: NamedArgument<T>, restArgs: Sequence<NamedArgument<T>>): Result
 }
 
 private enum class ArgsTraversalState { UNNAMED, NAMED, TAIL }
@@ -77,7 +77,7 @@ private fun <T> tryCreateCallableMapping(
     val res = mutableMapOf<KParameter, Any?>()
     var state = ArgsTraversalState.UNNAMED
     val unboundParams = callable.parameters.toMutableList()
-    val argIt = args.iterator()
+    val argIt = LookAheadIterator(args.iterator())
     while (argIt.hasNext()) {
         if (unboundParams.isEmpty()) return null // failed to match: no param left for the arg
         val arg = argIt.next()
@@ -102,7 +102,11 @@ private fun <T> tryCreateCallableMapping(
                     res[par] = cvtRes.v
                 } else if (par.type.jvmErasure.java.isArray) {
                     // try vararg
-                    val cvtVRes = converter.tryConvertVararg(par, arg, argIt)
+
+                    // Collect all the arguments that do not have a name
+                    val unnamed = argIt.sequenceUntil { it.name != null }
+
+                    val cvtVRes = converter.tryConvertVararg(par, arg, unnamed)
                     if (cvtVRes is ArgsConverter.Result.Success) {
                         res[par] = cvtVRes.v
                     } else return null // failed to match: no suitable param for unnamed arg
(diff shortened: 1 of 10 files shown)


compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
```diff
@@ -218,10 +219,31 @@ private class AnyArgsConverter : ArgsConverter<Any> {
                 else -> null
             }
 
-        if (value::class.isSubclassOf(parameter.type.jvmErasure)) return ArgsConverter.Result.Success(value)
+        fun evaluateValue(arg: Any): Any? {
+            if (arg::class.isSubclassOf(parameter.type.jvmErasure)) return arg
```
ligee, line 223: Are you sure that this line cannot return a valid `null` value? I'm not 100% sure about it, and if can, the logic below is incorrect.

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
```diff
@@ -218,10 +219,31 @@ private class AnyArgsConverter : ArgsConverter<Any> {
...
+            return convertPrimitivesArray(parameter.type, arg)
+        }
+
+        evaluateValue(value)?.let { return ArgsConverter.Result.Success(it) }
+
+        // Handle special scenario where [arg::class] is Array<Any> or Array<*> but
+        // but it's values all contain the proper values
```
ligee, line 230: Please check grammar, but in general, it's a very nice and valuable comment.

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
```diff
@@ -218,10 +219,31 @@ private class AnyArgsConverter : ArgsConverter<Any> {
...
+        val arrayComponentType = parameter
+            .type
+            .classifier
+            .safeAs<KClass<*>>()
+            ?.takeIf { it.java.isArray }
+            ?.java
+            ?.componentType
+            ?.kotlin
```
ligee, line 238: Please don't use such a chaining style, it hinder the readability. Group calls by some property, e.g. `parameter.type.classifier` can be placed on one line.

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
```diff
@@ -218,10 +219,31 @@ private class AnyArgsConverter : ArgsConverter<Any> {
...
 
-        return convertPrimitivesArray(parameter.type, value)?.let { ArgsConverter.Result.Success(it) }
-            ?: ArgsConverter.Result.Failure
+        if (value is Array<*> && arrayComponentType != null) {
+            // TODO: Idea! Maybe we should check if the values in the array are compatible with [arrayComponentType]
+            //  if they aren't perhaps we should fail silently
```
ligee, line 242: Yep, in general, more reporting from this code could be nice. But on the other hand, it partially reimplements the logic already present in the compiler, so ideally we'd like to reuse it, instead of extending this one.

plugins/scripting/scripting-compiler/tests/org/jetbrains/kotlin/scripting/compiler/test/ConstructAnnotationTest.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    fun testAnnotationWithEmptyArrayFunction() {
+        val annotations = annotations("TestAnnotationWithEmptyArrayFunction.kts", TestAnnotation::class)
+            .valueOrThrow()
+            .filterIsInstance(TestAnnotation::class.java)
+
+        assertEquals(annotations.count(), 1)
+        assert(annotations.first().options.isEmpty())
+    }
+
```
ligee, line 83: The missing case is our current `@DependsOn` annotation with a parameter after a `vararg`. It doesn't work with the current argument mapping logic, unfortunately. Let's discuss what to do about it offline.

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
nerdsupremacist: This logic was already there before. The case with null is handled in line 206 which I haven't changed

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
nerdsupremacist: ✅

compiler/cli/cli-common/src/org/jetbrains/kotlin/utils/parametersMap.kt
nerdsupremacist: ✅

plugins/scripting/scripting-compiler/tests/org/jetbrains/kotlin/scripting/compiler/test/ConstructAnnotationTest.kt
nerdsupremacist: I added a test that successfully constructs this annotation:

```kotlin
@file:AnnotationWithVarArgAndArray("option", moreOptions = ["otherOption"])
```

It will fail if the `moreOptions` label is missing, however, the regular compiler also won't accept it without the label:

From [docs](https://kotlinlang.org/docs/reference/functions.html):
```
Only one parameter may be marked as vararg. If a vararg parameter is not the last one in the list, values for the following parameters can be passed using the named argument syntax, or, if the parameter has a function type, by passing a lambda outside parentheses.
```

I'm not sure if this properly covers the scenario with a lambda outside the parenthesis, but we can't have this in our annotations anyway

ligee:
Merged manually

PR #3468, opened 2020-06-10
FIR: inherit property accessor modifiers from property and vice versa

jsjeon:
For certain modifiers (`inline` and `external`), putting it to property means its accessor is also, e.g., `inline`. Hence, downward propagation:
```
inline val foo: ...
    get() = ... // now, it is inline getter
```

It applies to the other way around, i.e., if both (or only getter) accessors are, e.g., `external`, the corresponding property is also, e.g., `external`. Hence, upward propagation:
```
val bar: ... // this is external (or native)
    external get
```


compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
```diff
@@ -1078,19 +1086,24 @@ class RawFirBuilder(
...
+                        } else null
+
                         status = FirDeclarationStatusImpl(visibility, modality).apply {
                             isExpect = hasExpectModifier()
                             isActual = hasActualModifier()
                             isOverride = hasModifier(OVERRIDE_KEYWORD)
                             isConst = hasModifier(CONST_KEYWORD)
                             isLateInit = hasModifier(LATEINIT_KEYWORD)
+                            isInline = hasModifier(INLINE_KEYWORD) ||
+                                    (!isVar && getter!!.isInline) || (isVar && getter!!.isInline && setter!!.isInline)
```
mglukhikh, line 1102: IMO this is over-complicated. Something like `getter!!.isInline && setter?.isInline == true` looks much simpler. Same thing for code below.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
```diff
@@ -958,6 +950,18 @@ class DeclarationsConverter(
...
                         } else null
 
+                    status = FirDeclarationStatusImpl(propertyVisibility, modifiers.getModality()).apply {
+                        isExpect = modifiers.hasExpect()
+                        isActual = modifiers.hasActual()
+                        isOverride = modifiers.hasOverride()
+                        isConst = modifiers.isConst()
+                        isLateInit = modifiers.hasLateinit()
+                        isInline = modifiers.hasInline() ||
+                                (!isVar && getter!!.isInline) || (isVar && getter!!.isInline && setter!!.isInline)
```
mglukhikh, line 960: IMO this is over-complicated. Something like `getter!!.isInline && setter?.isInline == true` looks much simpler. Same thing for code below.

compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
jsjeon: Agree, but I've tried to be as explicit as possible. :)

I worried `setter?.isInline == true` can't catch my intention, though. It implies setter is not null _and_ the specific modifier exists. However, for `val`, setter doesn't exist and it should trigger the upward propagation as well. Hence, `setter?.isInline != false` seems right to me. I added some comments to explain that as well.

compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
jsjeon: Ditto.

mglukhikh:
Merged manually 👍

PR #3470, opened 2020-06-11
KT-39490 'Wrap with ?.let' quickfix put the receiver in `let` lambda

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39490


igoriakovlev:
Thank you for you work.

PR #3471, opened 2020-06-11
Scripting REPL frontend performance improvments

ileasile:
This PR improves compilation (especially analyze) performance of REPL snippets:
- Fix imporing scopes so that every snippet has only one default importing scope (instead of N where N is a snippet number). It fastens symbols resolution significantly. It also fixes KT-35651
- Allow `TowerResolver` to skip time-consuming extensions resolution of implicits for some specific receivers. Implement such filter in REPL: it skips extensions resolution for snippet classes (`Line*`)
- Create scope for each implicit receiver from compilation configuration only once (not once per each compilation).


plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
```diff
@@ -0,0 +1,80 @@
...
+            val receiver = (scope as? LexicalScope)?.implicitReceiver?.value
+            val keep = receiver?.let {
+                val group = getGroup(it)
+                when (group.policy) {
+                    SkippingPolicy.DONT_SKIP -> true
+                    SkippingPolicy.SKIP_ALL -> false
+                    SkippingPolicy.SKIP_AFTER_FIRST -> {
+                        val processed = group in processedGroups
+                        processedGroups.add(group)
+                        !processed
```
dzharkov, line 57: `processedGroups.add(group)` returns the same boolean

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
```diff
@@ -0,0 +1,80 @@
...
+        if (receiver !is ImplicitClassReceiver) return DEFAULT_GROUP
+
+        val descriptor = receiver.declarationDescriptor
+        val descriptorFqName = descriptor.fqNameSafe.asString()
+        if (descriptorFqName in classesToSkipNames) return SKIP_GROUP
+        if (descriptorFqName in classesToSkipFirstTimeNames) return ReceiverGroup(descriptorFqName, SkippingPolicy.SKIP_AFTER_FIRST)
+
+        if (descriptor !is LazyScriptDescriptor) return DEFAULT_GROUP
+        val superClass = descriptor.getSuperClassOrAny()
+        if (superClass.fqNameSafe.asString() == baseScriptClassName) return LINE_CLASS_GROUP
```
dzharkov, line 38: I thought all of the lines extend the same superclass, don't they?

ileasile: Yes, they actually do, and it's a way to distinguish them from usual classes.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.kotlin.scripting.resolve.LazyScriptDescriptor
+import kotlin.script.experimental.api.KotlinType
+
+class BaseReplImplicitsExtensionsResolutionFilter(
+    baseScriptClass: KotlinType = KotlinType(Any::class),
+    classesToSkip: Collection<KotlinType> = emptyList(),
+    classesToSkipAfterFirstTime: Collection<KotlinType> = emptyList()
+) : ImplicitsExtensionsResolutionFilter {
+    private val baseScriptClassName = baseScriptClass.typeName
+    private val classesToSkipNames = classesToSkip.map { it.typeName }.toHashSet()
```
ligee, line 25: `mapTo` seems more appropriate here. Below as well.

compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/FileScopeFactory.kt
```diff
@@ -214,20 +217,24 @@ class FileScopeFactory(
...
                 scope, allUnderImportResolver, null, LazyImportScope.FilteringKind.VISIBLE_CLASSES,
                 "All under imports in $debugName (visible classes)"
             )
 
-            scope = LazyImportScope(
-                scope, defaultExplicitImportResolver, null, LazyImportScope.FilteringKind.ALL,
-                "Default explicit imports in $debugName"
-            )
+            if (createDefaultImportingScopes) {
+                scope = LazyImportScope(
```
ligee, line 233: does this scope include explicit imports from the same line? If yes, why it is not needed in repl when we have imports in the line?

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/extensions/ScriptingResolveExtension.kt
```diff
@@ -22,8 +26,17 @@ class ScriptingResolveExtension : SyntheticResolveExtension {
         declarationProvider: PackageMemberDeclarationProvider,
         result: MutableSet<ClassDescriptor>
     ) {
+        val previousImplicitReceivers = hashSetOf<KotlinType>()
+        for (file in declarationProvider.getPackageFiles()) {
```
ligee, line 30: This looks too expensive for me and probably incorrect too. 
What you probably should do instead is to collect descriptors compiling each line (see history handling or places where `earlierScriptsForReplInterpreter` is set) and get receivers from the script descriptors and pass them to the newly created descriptors.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
```diff
@@ -0,0 +1,76 @@
...
+        val superClass = descriptor.getSuperClassOrAny()
+        if (superClass.fqNameSafe.asString() == baseScriptClassName) return LINE_CLASS_GROUP
+
+        return DEFAULT_GROUP
+    }
+
+    override fun getScopesWithInfo(
+        scopes: Sequence<HierarchicalScope>
+    ): Sequence<ScopeWithImplicitsExtensionsResolutionInfo> {
+        val processedGroups = mutableSetOf<ReceiverGroup>()
```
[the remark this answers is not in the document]
ileasile: Actually, calling `add()` is reading, because it returns true/false depending on existance of element in the set, and that's what we need

plugins/scripting/scripting-ide-services-test/test/org/jetbrains/kotlin/scripting/ide_services/test_util/resolverUtil.kt
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.scripting.ide_services.test_util
+
+import com.sun.org.slf4j.internal.LoggerFactory
```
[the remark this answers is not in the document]
ileasile: Hmm... Seems to be ok for me, but I may drop off the logging

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
```diff
@@ -0,0 +1,76 @@
...
+            } ?: true
+
+            ScopeWithImplicitsExtensionsResolutionInfo(scope, keep)
+        }
+    }
+
+    /**
+     * Note that for groups with [SkippingPolicy.SKIP_AFTER_FIRST] policy, each group is treated separately.
+     */
+    private data class ReceiverGroup(val name: String, val policy: SkippingPolicy)
```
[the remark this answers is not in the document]
ileasile: Yes, they are used in `equals()` and `hashCode()` when the instances of this class are used in the set.

compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/FileScopeFactory.kt
ileasile: Yes, but in this case `createDefaultImportingScopes` is `true`, see implementation of `org.jetbrains.kotlin.scripting.compiler.plugin.repl.ReplCodeAnalyzerBase.ResettableAnalyzerState#computeFileScopes`

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/extensions/ScriptingResolveExtension.kt
ileasile: Ok, I will try to refactor this logic

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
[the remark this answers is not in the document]
ileasile: It's a complete implementation

plugins/scripting/scripting-ide-services-test/test/org/jetbrains/kotlin/scripting/ide_services/test_util/resolverUtil.kt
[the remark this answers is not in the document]
ligee: You don't need to drop logging from tests, if you need it. Just check that you haven't forgot to commit changes in the build files, if you added any dependencies.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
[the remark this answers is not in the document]
ligee: Ok, I see now, thanks. See the next comment though.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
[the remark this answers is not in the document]
ligee: Ok. I see it now.
But I think that the solution, in general, is too complicated and fragile.
Let's discuss whether you really need the `skipExtensionsResolutionForImplicits` and `skipExtensionsResolutionForImplicitsExceptInnermost` settings in the configuration, and if not, maybe this part could be significantly simplified.
The argument for that could be that if you will filter out repeating receivers from the previous lines, there will be no much need for filtering them again here.
I also doubt that the way you're detecting line descriptors is the correct one. E.g. I can easily imagine that in some DSL you may have base class for lines used as a base class for a receiver independent from the line classes.
So maybe it should boil down to the same mechanism of collecting previous lines in the script descriptor and use them here as well to filter out scopes.

compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/FileScopeFactory.kt
ligee: Sorry, I may be missing something again, but I don't see any relevant call to the `createScopesForFile` with `createDefaultImportingScopes = true`.

ileasile: Sorry for misleading, I've just haven't understood you correctly. Default imports **do not** include explicit imports. But for the first snippet in session it is called with `true`: in `org.jetbrains.kotlin.resolve.repl.ReplState#submitLine`, in override of `org.jetbrains.kotlin.resolve.lazy.FileScopesCustomizer#createFileScopes`.

compiler/resolution/src/org/jetbrains/kotlin/resolve/calls/tower/TowerResolver.kt
```diff
@@ -185,7 +185,8 @@ class TowerResolver {
                 TowerData.TowerLevel(localLevel).process()?.let { return it }
             }
 
-            for (scope in implicitScopeTower.lexicalScope.parentsWithSelf) {
+            for (scopeInfo in implicitScopeTower.allScopesWithImplicitsResolutionInfo()) {
```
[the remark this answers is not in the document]
ileasile: Actually no. It runs once per each name that is needed to be resolved, but `run()` may return before this line, and it often does.

plugins/scripting/scripting-ide-services-test/test/org/jetbrains/kotlin/scripting/ide_services/test_util/resolverUtil.kt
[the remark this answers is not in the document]
ileasile: Everything wascommited, but it was a strange thing with my dependencies. Hope that it's fixed now.

plugins/scripting/scripting-compiler/src/org/jetbrains/kotlin/scripting/compiler/plugin/repl/ReplImplicitsExtensionsResolutionFilter.kt
ileasile: Fixed

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/extensions/ScriptingResolveExtension.kt
ileasile: BTW, how should I get the script descriptor(s) here? By `ctx` or using more explicit way?

ileasile:
Merged manually into master and 1.4.0

PR #3474, opened 2020-06-13
FIR deserializer: load annotations for functions and constructors in Library

jsjeon:
The 1st commit applies a similar idea of https://github.com/JetBrains/kotlin/pull/3464 (which was for property accessors) to functions because annotations such as `@JvmName` can be used with functions as well. Indeed, bb tests that use `assertFails` whose JVM name differs are unmuted.

During that, a couple regressions were found, which help figure out a wrong raw fir building for class reference inside annotations with array value. The 2nd commit addresses that.

As a consequence, fir2ir conversion for GetClassCall needs update to handle remaining class reference from deserialized annotation. The 3rd commit adds that.

The 4th commit is a small addition and/or a tweak of the 1st commit: applying the same idea/logic to constructors.


compiler/fir/analysis-tests/testData/loadCompiledKotlin/annotations/ClassLiteralArguments.txt
```diff
@@ -15,7 +15,7 @@ public final annotation class Anno : R|kotlin/Annotation| {
 
 }
 
-@R|test/Anno|(klass = <getClass>(<getClass>(R|kotlin/String|)), klasses = <implicitArrayOf>(<getClass>(R|kotlin/Int|), <getClass>(R|kotlin/String|), <getClass>(R|kotlin/Float|)), sarKlass = <getClass>(<getClass>(R|kotlin/String|)), d2arKlass = <getClass>(<getClass>(R|kotlin/DoubleArray|))) public final class Klass : R|kotlin/Any| {
+@R|test/Anno|(klass = <getClass>(<getClass>(R|kotlin/String|)), klasses = <implicitArrayOf>(<getClass>(<getClass>(R|kotlin/Int|)), <getClass>(<getClass>(R|kotlin/String|)), <getClass>(<getClass>(R|kotlin/Float|))), sarKlass = <getClass>(<getClass>(R|kotlin/String|)), d2arKlass = <getClass>(<getClass>(R|kotlin/DoubleArray|))) public final class Klass : R|kotlin/Any| {
```
jsjeon, line 18: @mglukhikh So... both `FirGetClassCall` and `FirClassReferenceExpression` are rendered as `<getClass>`. :) Similar to other "reference", how about changing `FirClassReferenceExpression` to be rendered as, e.g., `C|...|`?

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/deserialization/JvmBinaryAnnotationDeserializer.kt
```diff
@@ -42,36 +45,69 @@ class JvmBinaryAnnotationDeserializer(
...
+                    JvmProtoBufUtil.getJvmConstructorSignature(proto, nameResolver, typeTable) ?: return null
+                )
+            }
+            is ProtoBuf.Function -> {
+                val signature = JvmProtoBufUtil.getJvmMethodSignature(proto, nameResolver, typeTable) ?: return null
+                // TODO: Investigate why annotations for accessors affect resolution, resulting in dangling type parameter.
+                //   regressions: Fir2IrTextTest.Declarations.test*LevelProperties
+                if (signature.name.startsWith("get") || signature.name.startsWith("set")) {
+                    return null
+                }
```
jsjeon, line 108: To be honest, all works have done a day ago, but it took me a while to deal with those regressions. :( Without these restriction, some with ambiguous resolutions are resolved (and unmuted, more clear IR texts, etc.), but some are simply failing with remaining type parameter at the conversion phase. I'll keep investigating it, but any hints, advices, insights, etc. would be appreciated, @mglukhikh

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/deserialization/JvmBinaryAnnotationDeserializer.kt
```diff
@@ -33,6 +35,7 @@ class JvmBinaryAnnotationDeserializer(
     private data class MemberAnnotations(val memberAnnotations: Map<MemberSignature, MutableList<FirAnnotationCall>>)
 
     private enum class CallableKind {
+        FUNCTION,
```
juan-chen, line 38: Seems only GETTER/SETTER are relevant kinds. Maybe use something like OTHER or NON_PROPERTY_ACCESSOR, instead of FUNCTION?

jsjeon: Yeah, this is not quite relevant. Thanks for the suggestion, I'll call it `OTHERS`.

mglukhikh:
NB: this PR slows modularized test down by 0.5-1.0 sec (1-2%).

mglukhikh:
We decided yesterday that @dzharkov will investigate the slow down and then either approve it or suggest possible optimization.

dzharkov:
I've merged your changes manually with some additional optimizations on top

jsjeon:
I've noticed https://github.com/JetBrains/kotlin/commit/429b2a9705cb92462ae0e325fdffbe9c458470b9 and https://github.com/JetBrains/kotlin/commit/efee0dae9474e63633bb22894a18af1fd16de15d. Thank you!

PR #3477, opened 2020-06-15
[FIR] fix mismatching type parameter symbols in generic class

juan-chen:
Type parameters in generic classes are given wrapped descriptors
whereas references to those type parameters inside the classes have
deserialized descriptors (from stub generation). This is a problem for
creating type substitutions for instantiated classes. This commit
fixes the mismatch by using deserialized descriptors if possible.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
@@ -22,7 +22,9 @@ import org.jetbrains.kotlin.ir.declarations.impl.IrClassImpl
 import org.jetbrains.kotlin.ir.declarations.impl.IrEnumEntryImpl
 import org.jetbrains.kotlin.ir.declarations.impl.IrTypeAliasImpl
 import org.jetbrains.kotlin.ir.declarations.impl.IrTypeParameterImpl
-import org.jetbrains.kotlin.ir.descriptors.*
+import org.jetbrains.kotlin.ir.descriptors.WrappedClassDescriptor
+import org.jetbrains.kotlin.ir.descriptors.WrappedEnumEntryDescriptor
+import org.jetbrains.kotlin.ir.descriptors.WrappedTypeParameterDescriptor
 import org.jetbrains.kotlin.ir.expressions.impl.IrEnumConstructorCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrExpressionBodyImpl
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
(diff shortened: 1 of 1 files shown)


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -282,11 +316,12 @@ class Fir2IrClassifierStorage(
     private fun createIrTypeParameterWithoutBounds(
         typeParameter: FirTypeParameter,
         index: Int,
-        typeContext: ConversionTypeContext = ConversionTypeContext.DEFAULT
+        typeContext: ConversionTypeContext = ConversionTypeContext.DEFAULT,
+        existingDescriptor: TypeParameterDescriptor? = null
     ): IrTypeParameter {
         require(index >= 0)
-        val descriptor = WrappedTypeParameterDescriptor()
         val origin = IrDeclarationOrigin.DEFINED
```
jsjeon, line 323: For those new handled case(s), I wonder the origin of type parameter, which would be the same as that of the enclosing class, i.e., external stub, should be propagated and recorded in that way. How about moving this `origin` to the param of `createIrTypeParameterWithoutBounds` with default origin `DEFINED` and pass that external stub if necessary?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -113,11 +115,43 @@ class Fir2IrClassifierStorage(
 
     private fun IrClass.declareTypeParameters(klass: FirClass<*>) {
         if (klass is FirRegularClass) {
-            preCacheTypeParameters(klass)
+            if (origin == IrDeclarationOrigin.IR_EXTERNAL_DECLARATION_STUB
```
jsjeon, line 118: There is external Java stub as well. Should we handle it in a similar way or not necessary?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
juan-chen: done

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
juan-chen: done

mglukhikh:
It will be probably easier if I think about getting symbol by signature myself. But think please about some test confirming this behavior.

mglukhikh:
NB: continued in #3480

PR #3478, opened 2020-06-15
Fix lookup tests

ileasile:
This PR fixes test failures introduced by https://github.com/JetBrains/kotlin/commit/b74692e96b241fdf3b7f2386eca9f25e299ceff1. 
It also changes the test engine to ignore the order of lookups, because semantically it doesn't matter.


ileasile:
Merged as a part of #3486

PR #3480, opened 2020-06-16
[FIR] add support for generic cases for delegation by implementation

juan-chen:
Add type parameters for generic delegated members and type
substitution when implementing instantiated super interfaces.
    
Test delegationToArrayList is muted again due to handling of type
parameters in Fir overriding computation: "remove(Int): E" in List is
regarded overriding "remove(E): Boolean" in MutableCollection, which
results in a different signature of the remove function.


juan-chen:
This is based on https://github.com/JetBrains/kotlin/pull/3462 and https://github.com/JetBrains/kotlin/pull/3477. The new bits are in 1ce906a.

mglukhikh:
Looks good, thank you.

mglukhikh:
Merged manually 👍

PR #3483, opened 2020-06-16
Fix build error messages

ileasile:


udalov:
Do we have a test to prevent this kind of problems in the future?

ligee:
@udalov I cannot think of an easy test for that, and the problem looks quite arbitrary to generalize.

PR #3484, opened 2020-06-17
FIR2IR: coerce to Unit when "when" expr is not effectively exhaustive

jsjeon:
Even though `when` expression is exhaustive, transformed `IrWhen` might not, since the constant true branch is not going to be added if it doesn't have body statements. That's effectively not exhaustive, and in that case, the return type of the expression is supposed to be `Unit`.


mglukhikh:
Merged, thank you!

PR #3486, opened 2020-06-17
Add missing definitelyDoesNotContainName and recordLookup methods

ileasile:
https://github.com/JetBrains/kotlin/commit/9c3ca7634c5d63b08c09c93fd4a5936452c46dcf fixes test logic so that lookup tests don't respect lookups orders (because it realy doesn't matter)
https://github.com/JetBrains/kotlin/commit/cb10fe4e7d40a7cacd16e82512326abf5e82a56f just adds some overrides to skip descriptors computation when only lookup record is needed.
https://github.com/JetBrains/kotlin/commit/496edce1139ff1522503282cd49bed49ba7a4b83 adds `definitelyDoesNotContainName` methods with a fix for kotlinx.serialization extension


compiler/frontend/src/org/jetbrains/kotlin/resolve/extensions/SyntheticResolveExtension.kt
```diff
@@ -136,6 +144,13 @@ interface SyntheticResolveExtension {
 
     fun getSyntheticNestedClassNames(thisDescriptor: ClassDescriptor): List<Name> = emptyList()
 
+    /**
+     * This method should return either superset of what [getSyntheticNestedClassNames] returns,
+     * or null in case it needs to run resolution and inference and/or it is very costly.
+     * Override this method if resolution started to fail with recursion.
+     */
+    fun maybeGetSyntheticNestedClassNames(thisDescriptor: ClassDescriptor): List<Name>? = getSyntheticNestedClassNames(thisDescriptor)
```
sandwwraith, line 152: imo, `getPossibleSyntheticNestedClassNames` is a slightly better name for this method

plugins/kotlin-serialization/kotlin-serialization-ide/src/org/jetbrains/kotlinx/serialization/compiler/extensions/SerializationIDEResolveExtension.kt
```diff
@@ -21,6 +21,9 @@ class SerializationIDEResolveExtension : SerializationResolveExtension() {
     override fun getSyntheticNestedClassNames(thisDescriptor: ClassDescriptor): List<Name> =
         getIfEnabledOn(thisDescriptor) { super.getSyntheticNestedClassNames(thisDescriptor) } ?: emptyList()
 
+    override fun maybeGetSyntheticNestedClassNames(thisDescriptor: ClassDescriptor): List<Name>? =
+        getIfEnabledOn(thisDescriptor) { super.maybeGetSyntheticNestedClassNames(thisDescriptor) } ?: emptyList()
```
sandwwraith, line 25: If `maybeGetSyntheticNestedClassNames` returns null, `emptyList()` still is the result because of `?:`. Is this a desirable behavior?

ileasile: Yes, it was a mistake. Thank you for pointing this out.

PR #3489, opened 2020-06-18
KT-39200 False positive "Redundant qualifier name" with same-named member object and companion property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39200


idea/src/org/jetbrains/kotlin/idea/inspections/RemoveRedundantQualifierNameInspection.kt
```diff
@@ -61,6 +61,7 @@ class RemoveRedundantQualifierNameInspection : AbstractKotlinInspection(), Clean
                 val expressionParent = expression.parent
                 if (expressionParent is KtDotQualifiedExpression || expressionParent is KtPackageDirective || expressionParent is KtImportDirective) return
                 val expressionForAnalyze = expression.firstExpressionWithoutReceiver() ?: return
+                if (expressionForAnalyze.selectorExpression?.text == expressionParent.safeAs<KtProperty>()?.name) return
```
dimonchik0036, line 64: This doesn't work for code:
```kotlin
sealed class Foo {
    object BAR : Foo()

    companion object {
        val BAR: Foo by lazy { Foo.BAR }
    }
}
```

PR #3491, opened 2020-06-18
The new release shouldn't have an existing version

LifeIsStrange:
Kotlin 1.3.72 has already been released, is this an unintentional mistake?


LifeIsStrange:
Well I find it weird but I understand the reasoning of not jumping the version, e.g everybody would benefit from the fixes for free without needed to e.g change a line in build.gradle.kts

PR #3492, opened 2020-06-18
FIR serializer: regard property accessors with modifiers as non-default

jsjeon:
Even though the name of `FirDefaultPropertyAccessor` implies that it's default, it just means it doesn't have body. It can have some other information, e.g., annotations, user-contributed visibility (like `val x private get`), and modifiers (like `val y external get`). Those are technically not default, and their serialized flags should indicate that. Otherwise, reflective inspection via property reference won't catch such attributes.


mglukhikh:
Merged, thank you!

PR #3494, opened 2020-06-21
KT-35732 KT-32403 Make string literals clicalble

shiraji:
https://youtrack.jetbrains.com/issue/KT-35732
https://youtrack.jetbrains.com/issue/KT-32403


vladimirdolzhenko:

it is merged as https://github.com/JetBrains/kotlin/commit/9844607d0f18ac88dc18b289791c5aacd33addb1

PR #3495, opened 2020-06-22
KT-39532 Support intention to convert reference to lambda and vice versa for adapted references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39532

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
@@ -8,6 +8,7 @@ package org.jetbrains.kotlin.idea.intentions
 import com.intellij.openapi.editor.Editor
 import org.jetbrains.kotlin.builtins.*
 import org.jetbrains.kotlin.config.LanguageFeature
+import org.jetbrains.kotlin.config.LanguageVersion
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.idea.KotlinBundle
 import org.jetbrains.kotlin.idea.caches.resolve.analyze
@@ -32,6 +33,7 @@ import org.jetbrains.kotlin.resolve.calls.callUtil.getParameterForArgument
 import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
 import org.jetbrains.kotlin.resolve.calls.components.hasDefaultValue
 import org.jetbrains.kotlin.resolve.calls.components.isVararg
+import org.jetbrains.kotlin.resolve.calls.model.DefaultValueArgument
 import org.jetbrains.kotlin.resolve.calls.model.ResolvedCall
 import org.jetbrains.kotlin.resolve.calls.model.VarargValueArgument
 import org.jetbrains.kotlin.resolve.descriptorUtil.isCompanionObject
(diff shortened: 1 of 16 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
```diff
@@ -80,7 +84,11 @@ open class ConvertLambdaToReferenceIntention(textGetter: () -> String) : SelfTar
 
         val lambdaParameterIsSuspend = lambdaParameterType?.isSuspendFunctionType == true
         val calleeFunctionIsSuspend = (calleeDescriptor as? FunctionDescriptor)?.isSuspend == true
-        if (lambdaParameterIsSuspend && !calleeFunctionIsSuspend || !lambdaParameterIsSuspend && calleeFunctionIsSuspend) return false
+        if (languageVersion >= LanguageVersion.KOTLIN_1_4) {
+            if (!lambdaParameterIsSuspend && calleeFunctionIsSuspend) return false
+        } else {
+            if (lambdaParameterIsSuspend && !calleeFunctionIsSuspend || !lambdaParameterIsSuspend && calleeFunctionIsSuspend) return false
+        }
```
fedochet, line 91: This is not the correct check, since this feature (passing reference to non-suspend method as suspend function parameter) can be enabled and disabled independently from the language version. You probably should rewrite the code like this:

```suggestion
        if (!lambdaParameterIsSuspend && calleeFunctionIsSuspend) return false
        if (!languageVersionSettings.supportsFeature(LanguageFeature.SuspendConversion)) {
            if (lambdaParameterIsSuspend && !calleeFunctionIsSuspend) return false
        }
```

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
```diff
@@ -94,17 +102,23 @@ open class ConvertLambdaToReferenceIntention(textGetter: () -> String) : SelfTar
...
 
         val callableArgumentsCount = (callableExpression as? KtCallExpression)?.valueArguments?.size ?: 0
-        if (calleeDescriptor.valueParameters.size != callableArgumentsCount) return false
-        val lambdaMustReturnUnit =
-            if (lambdaParameterType?.isFunctionType == true) lambdaParameterType.getReturnTypeFromFunctionType().isUnit() else false
-        if (lambdaMustReturnUnit) {
-            calleeDescriptor.returnType.let {
-                // If Unit required, no references to non-Unit callables
-                if (it == null || !it.isUnit()) return false
+        if (languageVersion >= LanguageVersion.KOTLIN_1_4) {
```
fedochet, line 109: This is also not a correct check -- this feature resides under the `LanguageFeature.FunctionReferenceWithDefaultValueAsOtherType` flag, which also can be disabled independently from the language version

idea/src/org/jetbrains/kotlin/idea/intentions/ConvertLambdaToReferenceIntention.kt
```diff
@@ -137,6 +151,7 @@ open class ConvertLambdaToReferenceIntention(textGetter: () -> String) : SelfTar
             if (lambdaValueParameterDescriptors.size < explicitReceiverShift + callableExpression.valueArguments.size) return false
             val resolvedCall = callableExpression.getResolvedCall(context) ?: return false
             resolvedCall.valueArguments.entries.forEach { (valueParameter, resolvedArgument) ->
+                if (resolvedArgument is DefaultValueArgument && languageVersion >= LanguageVersion.KOTLIN_1_4) return@forEach
```
fedochet, line 154: This too should be replaced with the `LanguageFeature.FunctionReferenceWithDefaultValueAsOtherType` flag check. I would advise to extract the flag check to some variable and reuse it twice

idea/testData/intentions/convertLambdaToReference/suspendFunctionParameter3.kt
```diff
@@ -0,0 +1,7 @@
+fun foo(a: suspend () -> Unit) {}
```
fedochet, line 1: By default this conversion will lead to red code, because currently it is disabled

Please enable it with this

```suggestion
// COMPILER_ARGUMENTS: -XXLanguage:+SuspendConversion
fun foo(a: suspend () -> Unit) {}
```

PR #3496, opened 2020-06-22
FIR2IR: handle type alias inside GetClassCall

jsjeon:
When a type alias is used as a class reference, e.g., [`(kotlin.)Error::class`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-error/), the symbol retrieval should refer to the aliased type, i.e., expanded type, e.g., [`java.lang.Error`](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html).


mglukhikh:
Merged, thank you!

PR #3498, opened 2020-06-24
[JVM] Extend stepping tests with <clinit> stepping.

madsager:
In addition, made the stepping information for constructor calls
consistent across JVM_IR and JVM. For JVM_IR that stepping behavior
is consistent for enum constructor calls in <clinit> for JVM it
is not.


madsager:
@max-kammerer follow up on the init block stepping for companion objects.

max-kammerer:
Thank you!

PR #3500, opened 2020-06-25
KT-30738 Inspection which replaces `.reduce { acc, i -> acc + i }` with `sum()`

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30738


nikitabobko:
Hi, thanks for PR! But code complexity which you bring with your PR is much more complex than the task it tries to solve. We are not ready to support this complex code just for sake of having this rare issue fixed. 

At least, I believe that this issue can be fixed much simpler with inspection based on "structural search and replace"

But ideally we should write as much as possible different expression simplification cases which are possible in Kotlin and understand whether structural search satisfies all our needs (https://youtrack.jetbrains.com/issue/KT-39797). After doing this research we can better understand our needs and determine the further direction of improvement. I mostly sure that "structural search and replace" should satisfy most of our needs and the ones it doesn't satisfy should be issued for improving "Structural search"

Again, thanks for PR and sorry that you wasted your time on it

nikitabobko:
You can help in this area by submitting more simplification cases into comment of KT-39797 and writing your thoughts about different approaches if you have any besides "Structural search and replace" that can be used as implementation.

PR #3501, opened 2020-06-25
Coroutines: Fix RedundantLocalsEliminationMethodTransformer

sfs:
`RedundantLocalsEliminationMethodTransformer` does not take control flow joins into account when collecting use sites for Unit values. For example, in code such as
```kotlin
x?.let { Unit } ?: Unit
```
the transformer will erroneously remove the `GETSTATIC kotlin/Unit.INSTANCE` instruction from the lambda argument to `let`, leading to incompatible stack heights at the join point after the safe call.

The transformer also does not preserve stores to local variables. This is probably a non-issue in practice, since we're talking about local variables of type Unit, but is still something we should take into account.

This PR changes `RedundantLocalsEliminationMethodTransformer` to use ASMs `SourceInterpreter` to compute usage information for all instructions, instead of a custom interpreter which only collects liveness information for use sites of unit values.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/RedundantLocalsEliminationMethodTransformer.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/RedundantLocalsEliminationMethodTransformer.kt
@@ -5,177 +5,127 @@
 
 package org.jetbrains.kotlin.codegen.coroutines
 
-import org.jetbrains.kotlin.codegen.inline.nodeText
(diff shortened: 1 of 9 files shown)


ilmirus:
Thanks for the PR, but I cannot merge it, sorry. The reason why I rewrote RedundantLocalsEliminationMethodTransformer is to remove all SourceInterpreters from coroutines codegen and the inliner. Otherwise, since SourceInterpreter's merge operation has N^2 complexity, it slows down compilation significantly if there are a lot of inlined code in a suspend function. So, I'll fix the bug and let you know.

PR #3502, opened 2020-06-26
FIR: resolve override conflicts when synthetic members are involved in

jsjeon:
As mentioned at https://youtrack.jetbrains.com/issue/KT-39070, super calls to `Any`'s synthetic members could be ambiguous when the class has super interface and super class, if, e.g., super class defines its own `equals` whereas the super type hierarchy through super interface led to `Any#equals`.

`OverloadingConflictResolver` in FE 1.0 has a step to filter out overrides (via `SyntheticMemberDescriptor`). This PR adds a similar (yet simpler) logic to the counterpart: `ConeOverloadConflictResolver`.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConeOverloadConflictResolver.kt
```diff
@@ -42,11 +44,78 @@ class ConeOverloadConflictResolver(
             else
                 candidates
 
+        // If one is an override of another, that overridden one is more specific.
+        val noOverrides = filterOverrides(fixedCandidates)
```
mglukhikh, line 48: I think base candidates should be filtered by use-site member scope instead of doing it once more in conflict resolver // cc @dzharkov

jsjeon: Let me know if I misunderstood something. The motivating example is:
```
interface A

open class B {
    override fun equals(other: Any?): Boolean {
        return super.equals(other)
    }
    ...
}

class C : A, B() {
    override fun equals(other: Any?): Boolean {
        return super.equals(other)
    }
    ...
}
```

The conflict happens while completing the `super.equal(...)` call, where, due to the unqualified `super`, the receiver has composite super type refs, which point to both interface `A` and class `B`. In the path to `B`, it's obvious that the resolution ends at `B#equal`. In the path to `A`, I wonder that call resolver is not able to see the resolution result of the other path (in this case `B#equal`)? Then, it's natural to return `Any#equal`.

Then, after collecting both candidates, this overload conflict resolver seems to me a good place to resolve such conflict.

dzharkov: Unfortunately, we have a bit more complicated algorithm here in FE 1.0, see https://youtrack.jetbrains.com/issue/KT-39599
It doesn't perform overload resolution but chooses a single supertype that contains functions we look for

See `org.jetbrains.kotlin.types.expressions.unqualifiedSuper.UnqualifiedSuperKt#resolveUnqualifiedSuperFromExpressionContext` and its usage at `org/jetbrains/kotlin/types/expressions/BasicExpressionTypingVisitor.java:527`

We've discussed the issue shortly in the team and decided that it requires more detailed discussion, maybe even with some changes in the language.

Sorry, but we think that for now, the best solution would be postponing your changes because the final resolution is likely to be very different.
I'm closing the issue, but please keep the branch for some time, just in case.

dzharkov:
I'm closing the PR because the issue addressed here needs some design resolutions (see the comment in the PR)

jsjeon:
Found https://github.com/JetBrains/kotlin/commit/376d3385d0f8ace0b4dacb27ee8b9af67ba4147c and https://github.com/JetBrains/kotlin/commit/cd896ae6c8221db47d7ce91ce7dfdde2b9e429b1 Deleting the branch too.

PR #3503, opened 2020-06-26
KAPT: Fix serialization of class structure data

gavra0:
When serializing ClasspathEntryData, ids assigned to types were
incorrect, resulting in corrupt data when deserializing files. This
happened if there were dependencies between classes.

Fixes: https://youtrack.jetbrains.com/issue/KT-39876
Test: ClasspathAnalyzerTest


abelkov:
Fixed in 1.4.0 branch: https://github.com/JetBrains/kotlin/commit/c6373987a90a410f139b153b7461e51ee5ea8dae

PR #3504, opened 2020-06-26
KAPT: Fix error reporting

gavra0:
When incremental analysis is unable to run, handle
failure gracefully. More info will be provided by
compiler diagnostics.

Fixes KT-36302

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
@@ -22,26 +22,36 @@ class MentionedTypesTaskListener(
 ) : TaskListener {
 
     var time = 0L
+    var failure: SourceAnalysisFailure? = null
+
     override fun started(e: TaskEvent) {
         // do nothing, we just process on finish
     }
 
     override fun finished(e: TaskEvent) {
+        if (failure != null) return // stop processing if we failed to analyze some sources
+
         if (e.kind != TaskEvent.Kind.ENTER || cache.isAlreadyProcessed(e.sourceFile.toUri())) return
 
-        val l = System.currentTimeMillis()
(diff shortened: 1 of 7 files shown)


plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/javacVisitors.kt
```diff
@@ -22,24 +22,32 @@ class MentionedTypesTaskListener(
...
-        compilationUnit.typeDecls.forEach {
-            it.accept(treeVisitor, Visibility.ABI)
+            val treeVisitor = TypeTreeVisitor(elementUtils, trees, compilationUnit, structure)
+            compilationUnit.typeDecls.forEach {
+                it.accept(treeVisitor, Visibility.ABI)
+            }
+            cache.addSourceStructure(structure)
+            time += System.currentTimeMillis() - l
+        } catch (t: Throwable) {
+            failureReason = "Running non-incrementally because analyzing ${e.sourceFile.toUri()} failed."
```
yanex, line 49: Should we print the exception to a debug log so we can understand what went wrong on a user machine?

gavra0: We can add that to debug log. I did not want to add it to info/warn as it is mostly obscure compiler-infra exception, but adding it to debug makes sense.

gavra0: This is now fixed in https://github.com/JetBrains/kotlin/compare/master...google:prr/gavra/kapt-better-error-message?expand=1 but PR is not updated for some reason.

abelkov:
Fixed in 1.4.0 branch: https://github.com/JetBrains/kotlin/commit/5d1727b68138babb42f85febaaef5a4f5699d9c3

PR #3505, opened 2020-06-27
FIR2IR: apply IrConstTransformer to other annotation containers with array values

jsjeon:
The motivation is bb test `CollectionLiterals.collectionLiteralsInArgumentPosition` that _was_ failing due to:
```
@Foo(a = [1f, 2f, 1 / 0f])
fun test1() {}
```

The first commit applies `IrConstTransformer` to all annotation containers, including functions.

------

That reveals another issue: annotation creation with named arguments, like
```
@Deprecated(level=... message="...")
fun foo() ...
```
where `Deprecated` constructor inputs `message`, `level`, and some other parameters in order. When `IrConstTransformer` visits such call in IR element form, no useful information is left, e.g., argument mapping.

Instead, during FIR resolution, arguments inside annotation should be converted to resolved argument list with mapping. The second commit does so. Also, during FIR2IR conversion, correct parameter index is computed for annotation call.

------

However, the target test _was_ still failing, and found that IR interpreter has a primitive array at end, but didn't support to convert it into `IrVararg`. The third commit adds that.

------

Alas, the target test _is_ still failing due to:
```
@Foo(b = ["Hello", ", ", "Kot" + "lin"])
fun test2() {}
```
that requires string concatenation at compile-time. It will also require me to learn more about IR interpreter, so I'd like to make a checking point for now.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/ConversionUtils.kt
@@ -446,9 +446,6 @@ internal fun FirReference.statementOrigin(): IrStatementOrigin? {
     }
 }
 
-fun FirClass<*>.getPrimaryConstructorIfAny(): FirConstructor? =
-    declarations.filterIsInstance<FirConstructor>().firstOrNull()?.takeIf { it.isPrimary }
-
 internal fun IrDeclarationParent.declareThisReceiverParameter(
     symbolTable: SymbolTable,
     thisType: IrType,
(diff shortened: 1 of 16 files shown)


jsjeon:
@vana06 PTAL 1st and 3rd commits. @mglukhikh as usual, PTAL. :)

mglukhikh:
Merged manually (in fact, I merged the 2nd commit only because other things were already done in @vana06 branch).

jsjeon:
@vana06 in https://github.com/JetBrains/kotlin/commit/de2b20482a4ff3d693278aa5d65d29c8bb713b52 I saw you left a TODO:
```
 // TODO this check can be removed after fix with annotation call arguments mapping
if ((0 until annotation.valueArgumentsCount).any { annotation.getValueArgument(it) == null }) return@forEach
```
which is now checked in: https://github.com/JetBrains/kotlin/commit/c3fc524c0d577610720299c0cc12dadc4818c480 (the 2nd commit @mglukhikh merged)

jsjeon:
Never mind. I found this one: https://github.com/JetBrains/kotlin/commit/d798071e06935c58c5c42d5b5f659020d19ce9db :)

PR #3508, opened 2020-06-29
[JVM_IR] Fix stepping behavior for assignments to local variables.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -502,15 +504,15 @@ class ExpressionCodegen(
...
         val initializer = declaration.initializer
         if (initializer != null) {
             var value = initializer.accept(this, data)
             initializer.markLineNumber(startOffset = true)
             value.materializeAt(varType, declaration.type)
+            declaration.markLineNumber(startOffset = true)
             mv.store(index, varType)
         } else if (declaration.isVisibleInLVT) {
+            declaration.markLineNumber(startOffset = true)
             pushDefaultValueOnStack(varType, mv)
```
max-kammerer, line 516: Shouldn't it be symmetrical to first `if` branch (emit linenumber just before `store` call)?

madsager: Yes, I agree, that would be better. That shifts labels a little bit in a few of the bytecode text tests, so there are a few updates there as well.

max-kammerer: On second look I'm a bit confused by changed tests on LVT. 
Looks very strange that compiler emit line numbers for simple variable declaration without initialization. 
Java doesn't emit anything for `z` declaration in next case:
```
    public static void main(String[] args) {
        int x = 1;
        int z;  //nothing about this linenumber
        z= 0;
    }
```
and look like Kotlin compiler shouldn't emit anything in this case too

madsager: That's a good point. The current backend does generate these line numbers and I agree that they are not useful. I removed the recording of the line number here so that the IR backend does not do so, and documented the behavior of both backends with a new stepping test.

max-kammerer:
Thank you!

PR #3509, opened 2020-06-29
FIR deserializer: proper annotation loading for value parameter of property setter

jsjeon:
At https://github.com/JetBrains/kotlin/pull/3464/files#diff-aca6498e9ef1b1d053cffda7fb32d750, when there is a non-default property accessor, a corresponding accessor has been built by FIR deserializer. That, however, should have read an annotation associated with the value parameter of setter, if any. (Instead, it always used kind of hard-coded, predefined value parameter.) This is a quick fix for that, by applying https://github.com/JetBrains/kotlin/commit/7834284bece00ffa8fe2b9d2211b963ff2448cb8 to the value parameter of setter.


jsjeon:
@demiurg906 PTAL

PR #3511, opened 2020-06-30
KT-39929 Add explicit this: false negative on variable call

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39929


nikitabobko:
Merged to new repo https://github.com/JetBrains/intellij-kotlin/commit/cf7e70f3928e5070d4b8e812ca305128b6552918

PR #3512, opened 2020-06-30
FIR deserializer: fix parameter shift for constructor of inner classes and enums

jsjeon:
When computing shifted index for value parameter in constructor, we need to know the kind of the enclosing class, e.g., whether it's enum or inner class. To be general, whenever value parameters are visited, the containing class's proto will be passed.


jsjeon:
@demiurg906 PTAL

jsjeon:
The reason that the affected test file is only about enum constructor is described here: https://youtrack.jetbrains.com/issue/KT-39837#focus=streamItem-27-4231425.0-0

demiurg906:
@jsjeon Thank you

PR #3513, opened 2020-06-30
FIR2IR: handle 'this' as reference to outer object

jsjeon:
`InnerClassesLowering` in backend IR expects _implicit_ `this` receiver to be retrieved via `IrGetValue`. During `fir2ir` conversion, _implicit_ `this` and explicit one are mangled, and for anonymous object as an implicit receiver, we passed `IrGetObjectValue` instead, resulting in missed lowering that causes dangling static call.

To distinguish implicit v.s. explicit `this` (without revealing corresponding internal declarations: `FirImplicitThisReference` and `FirExplicitThisReference`), this PR passes the context callee so that we can match the class bound to `this` reference against the context where such `this` receiver is going to be used. Then, for anonymous object case, we can avoid using `IrGetObjectValue` if it is used as an implicit receiver (so that `InnerClassesLowering` can pick it up and convert it to the proper `$outer` field access).

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
@@ -44,6 +44,7 @@ import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.ir.util.constructors
 import org.jetbrains.kotlin.ir.util.defaultType
+import org.jetbrains.kotlin.ir.util.parentClassOrNull
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtBinaryExpression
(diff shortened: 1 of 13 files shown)


jsjeon:
> Good try, but the code looks very nasty. I'd suggest one of the following modifications:

Thanks for the suggestions. I have to admit that I thought it's a bit ad-hoc too.

> 1. In `FirQualifiedAccess.findIrReceiver` we already know if the receiver is dispatch or not. We could introduce `FirThisExpression.toIrThisExpression(isInDispatchPosition: Boolean)` in visitor, and call it directly from `CallAndReferenceGenerator`. Probably this is the preferable way.

Well, that's the problem actually. It's still dispatch receiver; I saw determining `this` is used as dispatch receiver or extension receiver depends on `dispatchReceiver` or `extensionReceiver`, and for those unmuted cases, `dispatchReceiver` is set. However, called functions are not dispatching case. Those were... well, all inner-class's constructor!

So, I've tried totally different approach with very simpler logic. It worked, with two regressions. To mitigate those regressions, I ended up with searching for how `psi2ir` does and learning that we need to look upward through scopes.

Anyhow, the new approach will look very differently. PTAL.

mglukhikh:
IMO it looks much better, thanks.

PR #3514, opened 2020-07-01
FIR2IR: discard fake overrides for property accessors according to base visibility

jsjeon:
> Otherwise, IR validation failed: "Overrides private declaration ..."

This is supposed to be part of https://github.com/JetBrains/kotlin/pull/3252, which missed substitution case.


mglukhikh:
Merged, thank you!

PR #3516, opened 2020-07-01
[Scripting] Initial Type Provider API

nerdsupremacist:
This PR implements our basic initial Type Providers API. 

It includes:

- A Code Generation DSL
- An API to convert the results of the Code Generation DSL into a Artifacts that can be imported to a script
- An API for Annotation Based Type Providers using the DSL

## Code Generation DSL

The entire Code Generation DSL is based on the `GeneratedCode` interface:
```kotlin
interface GeneratedCode {
    fun Builder.body()
}
```

This allows users to build reusable constructions for generating code, without being exposed to any internals.

The rest of our DSL is a set of internal extension functions on `GeneratedCode.Builder` to cover the scenarios we need to cover. These include:
- Data Classes
- Objects
- Constants
- Companion Objects
- Functions
- Methods
- Computed Properties (with getters and optionally setters), also available as extensions
- Overriding members
- Enums
- Interfaces
- Inline code as string (as a fallback)

Example:
```kotlin
val myCode = GeneratedCode {
    dataClass("Foo") {
        property<Int>("bar")
        implement(MyInterface::class)

        method("baz") { argument: Int -> 
            val bar by property<Int>()
            // Do something
        }

        override {
            method("other") { first: Int, second: String -> 
                // Call baz using invoke on string
                val bazResult = "baz"(first)

                // Do something else
            }   
        }
    }
}
```

## Using Generated Code

Generated Code can be turned into artifacts, which will return a list of Scripts that should be imported along with a set if imports that should be added to the scripts

```kotlin
val code = GeneratedCode {
    ...
}

val artifacts = code.artifacts()
```

These artifacts can also we included in a `ScriptCompilationConfiguration`:

```kotlin
val newCompilationConfiguration = config.provide(artifacts)

// or

val newCompilationConfiguration = config.provide {
    // Full access to Code Generation DSL
    constant("meaning", 42)
}
```

## Annotation Based Type Providers

To create a Type Provider you define an annotation and implement `AnnotationBasedTypeProvider`

Example:

```kotlin

@Target(AnnotationTarget.FILE)
@Repeatable
@Retention(AnnotationRetention.SOURCE)
annotation class MyAnnotation(val number: Int)

object MyTypeProvider : AnnotationBasedTypeProvider<MyAnnotation> {
    override fun invoke(
        collectedAnnotations: List<ScriptSourceAnnotation<JSON>>,
        context: TypeProvider.Context
    ): ResultWithDiagnostics<GeneratedCode> {
        val number = collectedAnnotations.first().annotation.number 

        return GeneratedCode {
             dataClass("MyClass") {
                 property<Int>("number", default = number)
             }
        }.asSuccess()
    }
}
```

To use a type provider you add it to your script compilation configuration using `typeProviders`.
Example:

```kotlin
ScriptCompilationConfiguration {
    ...
    typeProviders {
        +MyTypeProvider
    }
}
```

## Future Work

- [ ] Caching of the generated code.
- [ ] The annotation based type providers don't check for basic name clashes before compiling to report on each annotation. These errors would then get caught by the compiler, but the error message might not be very helpful to the user

The code generation still doesn't support these scenarios
- [ ] Include a way to add default arguments to functions and methods
- [ ] Subclassing
- [ ] Sealed classes
- [ ] Generics on functions. We already support the receiver type to have a generic parameter, however this parameter can't be a generic parameter of the function, i.e.: `fun <T> Foo<T>.doSomething(): T`


libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * Generated code that is a type that can be used as an Argument
+ */
+interface IdentifiableCode : GeneratedCode, IdentifiableMember
+
+/**
+ * Builder for a provided type whose JVM Name can be set
+ */
+interface JVMNameableBuilder {
```
ligee, line 154: Why is it JVM-specific?
also, the function seems unused, so the intent is not clear.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+ */
+fun GeneratedTypeBuilder.implement(kClass: KClass<*>) {
+    require(kClass.java.isInterface) { "Provided type is not an interface" }
+    implement(IdentifiableMember(kClass).asInterface())
+}
+
+/**
+ * Add a companion object to the generated type
+ */
+fun GeneratedTypeBuilder.companionObject(name: String? = null, init: ObjectBuilder.() -> Unit) {
```
ligee, line 203: It would be nice to have tests that use all functions from the DSL, otherwise, they may rot over time and we may discover it too late.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+interface GeneratedTypeBuilder : IdentifiableMember {
+    /**
+     * Implement a generated interface
+     */
+    fun implement(generated: GeneratedInterface)
+
+    /**
+     * Add code to the namespace of the Type
+     */
+    fun inner(code: GeneratedCode)
```
ligee, line 170: I think this is a confusing name, the word `inner` has a different meaning in the Kotlin code.
In the compiler, we're using `child`/`children`  in such places.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+         */
+        fun composite(iterable: Collection<GeneratedCode>): GeneratedCode = if (iterable.count() == 1)
+            iterable.first()
+        else
+            CompoundGeneratedCode(iterable.toList())
+
+        /**
+         * Create a composition of multiple instances of Generated Code
+         */
+        fun composite(vararg code: GeneratedCode) = composite(code.toList())
```
ligee, line 101: Probably `compose` would be better.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/**
+ * Builder for a provided type. It can implement an interface and have some code inside.
+ */
+interface GeneratedTypeBuilder : IdentifiableMember {
+    /**
+     * Implement a generated interface
+     */
+    fun implement(generated: GeneratedInterface)
```
ligee, line 165: `implements` would look better for me

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+ * Builder for a provided type whose JVM Name can be set
+ */
+interface JVMNameableBuilder {
+    fun jvmName(name: String)
+}
+
+/**
+ * Builder for a provided type. It can implement an interface and have some code inside.
+ */
+interface GeneratedTypeBuilder : IdentifiableMember {
```
ligee, line 161: This seems to represent an analog of `ClassLike` things in the compiler, so maybe it makes sense to name it accordingly. `Type` is a bit more general.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Add code to the namespace of the Type
+     */
+    fun inner(code: GeneratedCode)
+}
+
+/**
+ * Builder for a provided data class. It can implement interfaces, have code inside and have a number of properties
+ */
+interface DataClassBuilder : JVMNameableBuilder, GeneratedTypeBuilder {
```
ligee, line 176: I'm not sure that there are benefits of having extracted interfaces from the leaf nodes. Having the public open builder inside the `DataClass` (in this case) will make it leaner, and also allow to reuse some functionality from the similar builders and containers.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeMethod.kt
```diff
@@ -0,0 +1,318 @@
...
+import kotlin.reflect.*
+import kotlin.script.experimental.typeProviders.generatedCode.GeneratedCode
+import kotlin.script.experimental.typeProviders.generatedCode.GeneratedTypeBuilder
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.useImport
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.useImports
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.GeneratedCodeVisitor
+import kotlin.script.experimental.typeProviders.generatedCode.internal.InternalGeneratedCode
+import kotlin.script.experimental.typeProviders.generatedCode.override
+
+private val storedValues = mutableMapOf<String, Any>()
```
ligee, line 17: this is an unacceptable global state. You need to put it into some context carried with the script.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeInlineClass.kt
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
ligee, line 4: I guess it should belong to a different commit.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeLazyProperty.kt
```diff
@@ -0,0 +1,129 @@
...
+import kotlin.reflect.KClass
+import kotlin.script.experimental.typeProviders.generatedCode.GeneratedCode
+import kotlin.script.experimental.typeProviders.generatedCode.GeneratedTypeBuilder
+import kotlin.script.experimental.typeProviders.generatedCode.internal.InternalGeneratedCode
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.GeneratedCodeVisitor
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.useImport
+import kotlin.script.experimental.typeProviders.generatedCode.override
+
+
+private val storedValues = mutableMapOf<String, Any>()
```
ligee, line 17: Global state.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeLazyProperty.kt
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
ligee, line 4: wrong commit

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeProperty.kt
```diff
@@ -0,0 +1,196 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
ligee, line 4: wrong commit

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeTypeAlias.kt
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
ligee, line 4: wrong commit

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeMethod.kt
```diff
@@ -0,0 +1,318 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
```
ligee, line 4: The approach here looks strange to me, especially in the lack of the usages to chack the intentions. Let's discuss it online.

ligee (changes requested):
There are quite many comments already, so I think that a big refactoring is imminent. So, I think it doesn't make sense to dive too deep into details now. Let's discuss the approaches online.

libraries/scripting/type-providers/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+tasks.withType<org.jetbrains.kotlin.gradle.dsl.KotlinCompile<*>> {
+    kotlinOptions.freeCompilerArgs += "-Xallow-kotlin-package"
+}
+
+publish()
+
+runtimeJar()
+sourcesJar()
+javadocJar()
+testsJar()
```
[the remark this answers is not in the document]
nerdsupremacist: My bad. I copied and pasted stuff until it worked.

I'll clean up the build.gradle ASAP.

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeInlineClass.kt
nerdsupremacist: Oh yeah. I'll redo the entire git history after we our call

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: This was me preparing for something like Data sciences or the shell type provider's extension function heavy generated code. It will often result in clashing names. So renaming the JVM name for things seems appropriate

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeDataClass.kt
```diff
@@ -0,0 +1,110 @@
...
+import kotlin.script.experimental.typeProviders.generatedCode.internal.visitor.useImports
+import kotlin.script.experimental.typeProviders.generatedCode.internal.write
+import kotlin.text.Regex.Companion.escape
+
+/**
+ * Create a data class
+ *
+ * @param name Name of your data class
+ */
+fun makeDataClass(name: String, init: DataClassBuilder.() -> Unit): IdentifiableCode {
```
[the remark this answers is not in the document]
nerdsupremacist: `makeInterface` is definitely planned. I wasn't sure if `makeRegularClass` is necessary though

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/generatedCodeFromFile.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Include file as code generated.
+ */
+fun File.asGeneratedCode(): GeneratedCode =
+    GeneratedCodeInFile(this)
+
+/**
+ * Include jar file as generated code.
+ */
+fun File.asGeneratedJar(packageName: String): GeneratedCode =
```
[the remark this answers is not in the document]
nerdsupremacist: I'll write some tests. But this is in case the type provider has an already compiled jar they want to include.

I'm not sure if it will come up. I'm fine with removing it for now and seeing if it comes up

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+ * - etc.
+ */
+interface GeneratedCode {
+    /**
+     * Generate the code that corresponds to your type
+     *
+     * @return An instance of [GeneratedCode].
+     * Important: don't return an infinite recursion in the form of return `this`
+     */
+    fun body(): GeneratedCode
```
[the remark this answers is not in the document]
nerdsupremacist: This is a new design I'm trying out, where the user can define their own `GeneratedCode`, but it has to be a composition of the code we provide.

It's inspired by a new API in Swift for UI programming: https://developer.apple.com/documentation/swiftui/view

So the idea is that versions that are not internal have to at some point be reduced to a collection of internal constructs that we can work with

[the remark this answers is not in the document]
nerdsupremacist: But you're right. Perhaps it should throw an exception instead of returning `this` in that moment

libraries/scripting/type-providers/build.gradle.kts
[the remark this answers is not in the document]
nerdsupremacist: I think it's a lot cleaner now, but I'd love to go over it with you if you don't mind

libraries/scripting/type-providers/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+    testCompile(projectTests(":compiler:tests-common"))
+
+
+    testCompile(project(":compiler:frontend"))
+    testCompile(project(":compiler:plugin-api"))
+    testCompile(project(":compiler:frontend.java"))
+    testCompile(project(":compiler:backend.js"))
+
+    testImplementation(intellijCoreDep()) { includeJars("intellij-core") }
+    testRuntimeOnly(intellijDep()) { includeJars("jps-model") }
```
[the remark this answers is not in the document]
nerdsupremacist: You're right

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/TypeProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Examples include:
+ * - Generate data classes that would parse a JSON file.
+ * - Generate a type-safe client for a GraphQL API based on the Server URL.
+ * - Generate kotlin stubs to call C libraries.
+ * - Generate a type-safe DSL for shell commands.
+ *
+ * To include a Type Provider in a Scripts definition see [ScriptCompilationConfiguration.Builder.typeProviders].
+ */
+interface TypeProvider<A : Annotation> {
```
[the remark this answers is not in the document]
nerdsupremacist: As discussed we renamed it to `AnnotaionBasedTypeProvider` and allow to create the necessary artifacts from the code generation DSL

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/TypeProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+    data class Context internal constructor(
```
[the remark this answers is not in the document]
nerdsupremacist: Done

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/TypeProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * If it is not necessary to to perform any of these tasks, the default implementation will leave the annotations as they arrive.
+     *
+     * @param collectedAnnotations Collection of Annotations and their locations in the Script Source.
+     * @param context Context information regarding the calling script and base directory. See [TypeProvider.Context] for more information.
+     *
+     * @return In case of success the list of annotations that should be turned into types.
+     * In case of a failure an appropriate message for the developer.
+     */
+    fun preProcess(
```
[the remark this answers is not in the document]
nerdsupremacist: Done

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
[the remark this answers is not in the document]
nerdsupremacist: Changed it to `GeneratedCode.Builder.body()` where user's can call our DSL to include code themselves

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
```diff
@@ -0,0 +1,302 @@
...
+    companion object {
+        /**
+         * Create a composition of multiple instances of Generated Code
+         */
+        fun composite(init: Builder.() -> Unit) = Builder().apply(init).build()
+
+        /**
+         * Create a composition of multiple instances of Generated Code
+         */
+        fun composite(iterable: Collection<GeneratedCode>): GeneratedCode = if (iterable.count() == 1)
```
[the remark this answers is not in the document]
nerdsupremacist: Went with invoke

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/generatedCodeFromFile.kt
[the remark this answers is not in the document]
nerdsupremacist: Removed since it was a bit of an additional case. If at some point we need it, I'll add it back

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeDataClass.kt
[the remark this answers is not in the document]
nerdsupremacist: Renamed to `dataClass`. Put it in the DSL and included ``interface`` as well

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Removed. Will add it again as a modifier just like override in case it becomes necessary again

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Added a bunch of tests. It should be just shy of 70% coverage now

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Got rid of this abstraction by making `ClassLikeBuilder` implement `GeneratedCode.Builder`

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Renamed to `invoke`

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Renamed to ClassLike

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/GeneratedCode.kt
nerdsupremacist: Reusing `ClassLike.Builder`
`DataClassBuilder` is still there. But the implementation reuses as much as possible from the `ClassLike.Builder`

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeMethod.kt
nerdsupremacist: Replaced global state with serialized values file

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeInlineClass.kt
nerdsupremacist: Redid history

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeLazyProperty.kt
nerdsupremacist: Removed

libraries/scripting/type-providers/src/kotlin/script/experimental/typeProviders/generatedCode/impl/makeProperty.kt
nerdsupremacist: Reordered

ligee:
I still think we might need this MR in the future, so converting it to a draft to remove it from the list of pending ones.

PR #3519, opened 2020-07-02
[JVM_IR]: Improve stepping for `when`.

madsager:
Additionally, use the line number of the class for default interface
dispatch methods.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -126,6 +126,7 @@ class ExpressionCodegen(
         get() = typeMapper.typeSystem
 
     override var lastLineNumber: Int = -1
+    var noLineNumberScope: Boolean = false
 
     private val closureReifiedMarkers = hashMapOf<IrClass, ReifiedTypeParametersUsages>()
 
@@ -150,6 +151,7 @@ class ExpressionCodegen(
     private fun getLineNumberForOffset(offset: Int): Int = fileEntry?.getLineNumber(offset)?.plus(1) ?: -1
 
     private fun IrElement.markLineNumber(startOffset: Boolean) {
+        if (noLineNumberScope) return
         val offset = if (startOffset) this.startOffset else endOffset
         if (offset < 0) {
             return
@@ -165,6 +167,13 @@ class ExpressionCodegen(
     }
 
     fun markLineNumber(element: IrElement) = element.markLineNumber(true)
+
+    fun noLineNumberScope(block: () -> Unit) {
+        val previousState = noLineNumberScope
+        noLineNumberScope = true
+        block()
+        noLineNumberScope = previousState
+    }
 
     // TODO remove
     fun gen(expression: IrExpression, type: Type, irType: IrType, data: BlockInfo): StackValue {
@@ -364,7 +373,7 @@ class ExpressionCodegen(
     private fun visitStatementContainer(container: IrStatementContainer, data: BlockInfo) =
         container.statements.fold(unitValue) { prev, exp ->
             prev.discard()
-            exp.accept(this, data).also { (exp as? IrExpression)?.markEndOfStatementIfNeeded() }
+            exp.accept(this, data)
         }
 
     override fun visitBlockBody(body: IrBlockBody, data: BlockInfo): PromisedValue {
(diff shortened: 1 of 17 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -845,19 +864,6 @@ class ExpressionCodegen(
         }
     }
 
-    private fun IrExpression.markEndOfStatementIfNeeded() {
```
madsager, line 848: These generate line numbers for the closing brace of `when`, `try`, `if`. However, this does not ensure that there is an instruction for that line number. Therefore, these line numbers will be there in the line number table, but will often not have instructions and can therefore not be observed. With the old backend, these are also emitted and typically cannot be observed. I opted for removing them entirely as I do not find it useful to be able to step and set breakpoints on the closing brace. The alternative is to go the other way and always emit a `nop` so that you will always get a step here (and the posibility of setting a breakpoint).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -348,7 +348,27 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
         override fun genSwitch() {
-            subject.accept(codegen, data).materialize()
+            // Do not generate line numbers for the table switching. In particular,
+            // the subject is extracted from the condition of the first branch which
+            // will give the wrong stepping behavior for code such as:
+            //
+            // when {
+            //   x == 42 -> 1
+            //   x == 32 -> 2
+            //   x == 24 -> 3
```
max-kammerer, line 358: Could you also add some tests for multiline `case`, e.g.

            when {
               x == 42 -> 1
               x == 32 -> 
                 2
               x == 24 -> 3

madsager: Sorry for the late reply on this. Vacation got in the way. :) 

Good idea. Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/SwitchGenerator.kt
```diff
@@ -348,7 +348,27 @@ class SwitchGenerator(private val expression: IrWhen, private val data: BlockInf
...
+            //   ...
+            // }
+            //
+            // If the subject line number is generated, we will not stop on the line
```
max-kammerer, line 362: Would it be line numbers for initialzier of when variable (`when (val x = smthTrickyCall()){...} or nested multiline  when(x = when {... } ) {...}`)? 
For example:
```   
   val y = 1;
    when (val x =
            when {
                y == 0 -> 1;
                y == 1 -> 2;
                else -> 0;
            }) {
        0 -> 123;
        1 -> {
        }
    }
```

madsager: `subject` in both `IntSwitch` and `StringSwitch` has to be an `IrGetValue`, so this code does not come into play for such more complicated subjects. I added your test case for good measure.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -166,6 +168,12 @@ class ExpressionCodegen(
 
     fun markLineNumber(element: IrElement) = element.markLineNumber(true)
 
+    fun noLineNumberScope(block: () -> Unit) {
+        noLineNumberScope = true
+        block()
+        noLineNumberScope = false
```
max-kammerer, line 174: I would suggest to keep old state and restore it here to avoid bugs in future

madsager: Done.

max-kammerer: Thank you! That was last from my side, but there is one more from TC:  failed SerializationIrBytecodeListingTestGenerated.testBasic test

madsager: There is a dump of the full bytecode for the IR backend. The only change is the removal of the end brace line numbers. The actual code is the same. Rebased so the test passes. As far as I can tell, the point of the test is to check that serialization code is correctly generated.

max-kammerer:
Thank you!

PR #3521, opened 2020-07-03
instant execution

nav-nav:


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/CompilerArgumentsContributor.kt
```diff
@@ -44,7 +44,7 @@ internal open class AbstractKotlinCompileArgumentsContributor<T : CommonCompiler
 
     protected val logger by taskProvider.map { it.logger }
 
-    private val isMultiplatform by taskProvider.map { it.isMultiplatform }
+    private val isMultiplatform = taskProvider.map { it.isMultiplatform }.get()
```
eskatos, line 47: 💄 Don't `.get()` the provider for uniformity with the other fields in this class

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptGenerateStubsTask.kt
```diff
@@ -69,6 +69,9 @@ open class KaptGenerateStubsTask : KotlinCompile() {
             error("KaptGenerateStubsTask.useModuleDetection setter should not be called!")
         }
 
+    @get:Input
+    val verbose = (project.hasProperty("kapt.verbose") && project.property("kapt.verbose").toString().toBoolean() == true)
```
eskatos, line 73: :o: This could be modelled as a `Provider` to account for later changes to that property:

```kotlin
project.providers.gradleProperty("kapt.verbose").map { it.toBoolean() }.orElse(false)
```

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/TransformKotlinGranularMetadata.kt
```diff
@@ -15,22 +15,28 @@ import org.jetbrains.kotlin.gradle.plugin.sources.KotlinDependencyScope.*
...
+        project.buildDir.resolve("kotlinSourceSetMetadata/${kotlinSourceSet.name}")
+    }
 
     @Suppress("unused") // Gradle input
     @get:InputFiles
     @get:PathSensitive(PathSensitivity.RELATIVE)
-    internal val allSourceSetsMetadataConfiguration = project.configurations.getByName(ALL_COMPILE_METADATA_CONFIGURATION_NAME)
+    internal val allSourceSetsMetadataConfiguration: FileCollection by lazy {
+        project.files(project.configurations.getByName(ALL_COMPILE_METADATA_CONFIGURATION_NAME))
+    }
```
eskatos, line 39: :x: `FileCollection` is already a "lazy" type, no need to wrap it inside a `by lazy {}`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinWithJavaTarget.kt
```diff
@@ -44,10 +45,9 @@ open class KotlinWithJavaTarget<KotlinOptionsType : KotlinCommonOptions>(
             KotlinWithJavaCompilationFactory(project, this, kotlinOptionsFactory)
         )
 
-    internal val defaultArtifactClassesListFile: File
-        get() {
+    internal val defaultArtifactClassesListFile: Provider<File> = project.provider {
             val jarTask = project.tasks.getByName(artifactsTaskName) as Jar
-            return File(File(project.buildDir, KOTLIN_BUILD_DIR_NAME), "${sanitizeFileName(jarTask.archiveFileName.get())}-classes.txt")
+            File(File(project.buildDir, KOTLIN_BUILD_DIR_NAME), "${sanitizeFileName(jarTask.archiveFileName.get())}-classes.txt")
         }
```
eskatos, line 51: ❓ I'm not sure where this `val` is wired, but the above code is losing task dependency and reference to `jar.archiveFileName` and could cause ordering problems.

Another way to write this would be:

```kotlin
project.tasks.named("artifactsTaskName", Jar::class).flatMap { jar ->
    project.layout.buildDirectory.file("$KOTLIN_BUILD_DIR_NAME/${sanitizeFileName(jar.archiveFileName.get())}-classes.txt")
}
```
This would return a `Provider<RegularFile>`. If you need it to be a `Provider<File>` just chain with `.map { it.asFile }`

Again, I'm not sure where this gets used and if my comment is relevant.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/ParsedGradleVersion.kt
```diff
@@ -56,4 +57,12 @@ internal data class ParsedGradleVersion(val major: Int, val minor: Int) : Compar
 
 fun isGradleVersionAtLeast(major: Int, minor: Int) =
     ParsedGradleVersion.parse(GradleVersion.current().version)
-        ?.let { it >= ParsedGradleVersion(major, minor) } ?: false
\ No newline at end of file
+        ?.let { it >= ParsedGradleVersion(major, minor) } ?: false
+
+fun isConfigurationCacheAvailable(gradle: Gradle) =
```
eskatos, line 62: :x: you should not need that check at all, see my other comments at usage sites

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/BuildEventsListenerRegistryHolder.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import org.gradle.api.Project
+import org.gradle.build.event.BuildEventsListenerRegistry
+import org.jetbrains.kotlin.gradle.utils.isConfigurationCacheAvailable
+import javax.inject.Inject
+
+open class BuildEventsListenerRegistryHolder @Inject constructor(val listenerRegistry: BuildEventsListenerRegistry?) {
+    companion object {
+        fun getInstance(project: Project) = run {
+            if (isConfigurationCacheAvailable(project.gradle)) {
```
eskatos, line 16: :x: the check should be `gradleVersion >= 6.1` instead where `BuildEventsListenerRegistry` is available

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleBuildServices.kt
```diff
@@ -34,36 +32,69 @@ internal class KotlinGradleBuildServices private constructor(
...
                 log.kotlinDebug(ALREADY_INITIALIZED_MESSAGE)
                 return instance!!
             }
 
+            val gradle = project.gradle
             val services = KotlinGradleBuildServices(gradle)
-            gradle.addBuildListener(services)
-            instance = services
-            log.kotlinDebug(INIT_MESSAGE)
+            if (isConfigurationCacheAvailable(gradle)) {
```
eskatos, line 76: :x: the check should be `gradleVersion >= 6.1` instead where `BuildEventsListenerRegistry` is available

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/statistics/KotlinBuildStatsService.kt
```diff
@@ -106,21 +105,26 @@ internal abstract class KotlinBuildStatsService internal constructor() : BuildAd
                             )
                             instance = JMXKotlinBuildStatsService(mbs, beanName)
                         } else {
+                            val kotlinBuildStatProvider = project.provider{ KotlinBuildStatListener(beanName) }
                             val newInstance = DefaultKotlinBuildStatsService(gradle, beanName)
+
+                            if (isConfigurationCacheAvailable(gradle)) {
```
eskatos, line 111: :x: the check should be `gradleVersion >= 6.1` instead where `BuildEventsListenerRegistry` is available

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleFinishBuildHandler.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.kotlin.utils.addToStdlib.sumByLong
+import java.lang.management.ManagementFactory
+import kotlin.math.max
+
+// move to KotlinGradleBuildListener when min supported version is 6.1
+class KotlinGradleFinishBuildHandler {
+
+    private val log = Logging.getLogger(this.javaClass)
+    private var startMemory: Long? = null
+    private val shouldReportMemoryUsage = System.getProperty(KotlinGradleBuildServices.SHOULD_REPORT_MEMORY_USAGE_PROPERTY) != null
```
gavra0, line 25: Use `ObjectFactory.systemProperty` to get the value to allow reconfiguration if it changes.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleBuildListener.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.gradle.plugin
+
+import org.gradle.tooling.events.FinishEvent
+import org.gradle.tooling.events.OperationCompletionListener
+
+//Available since Gradle 6.1
+class KotlinGradleBuildListener(
```
gavra0, line 12: This can be a `BuildService` and if you make it `AutoCloseable` you can implement clean-up logic in `close()` method.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleBuildServices.kt
```diff
@@ -72,57 +103,11 @@ internal class KotlinGradleBuildServices private constructor(
     }
 
     override fun buildFinished(result: BuildResult) {
```
gavra0, line 105: It looks like configuration cached runs still rely on this method being invoked (please correct me if I'm wrong). If you'd like to be notified when build finished, build services provide you with such lifecycle (see my comment on `KotlinGradleBuildListener`).

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/GranularMetadataTransformation.kt
```diff
@@ -23,9 +23,13 @@ import java.util.zip.ZipOutputStream
 import javax.xml.parsers.DocumentBuilderFactory
 
 internal sealed class MetadataDependencyResolution(
+    @field:Transient // can't be used with Gradle Instant Execution, but fortunately not needed when deserialized
     val dependency: ResolvedComponentResult,
     val projectDependency: Project?
```
gavra0, line 28: Project is not serializable either, so probably needs `@Transient` as well.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/TransformKotlinGranularMetadata.kt
```diff
@@ -15,22 +15,28 @@ import org.jetbrains.kotlin.gradle.plugin.sources.KotlinDependencyScope.*
...
 @Inject constructor(
     @get:Internal
+    @field:Transient
     val kotlinSourceSet: KotlinSourceSet
 ) : DefaultTask() {
 
     @get:OutputDirectory
-    val outputsDir: File = project.buildDir.resolve("kotlinSourceSetMetadata/${kotlinSourceSet.name}")
+    val outputsDir: File by project.provider {
+        project.buildDir.resolve("kotlinSourceSetMetadata/${kotlinSourceSet.name}")
```
gavra0, line 31: Use `ProjectLayout` methods to compute this. E.g. this will not detect changes if buildDir changes.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleFinishBuildHandler.kt
eskatos: Should be `Providers.systemProperty(name)`

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt
```diff
@@ -50,9 +51,21 @@ const val COULD_NOT_CONNECT_TO_DAEMON_MESSAGE = "Could not connect to Kotlin com
 internal fun kotlinCompilerExecutionStrategy(): String =
     System.getProperty(KOTLIN_COMPILER_EXECUTION_STRATEGY_PROPERTY) ?: DAEMON_EXECUTION_STRATEGY
 
-internal open class GradleCompilerRunner(protected val task: Task) {
-    protected val project: Project
-        get() = task.project
+/*
+Using real taskProvider cause "field 'taskProvider' from type 'org.jetbrains.kotlin.compilerRunner.GradleCompilerRunner':
+value 'fixed(class org.jetbrains.kotlin.gradle.tasks.KotlinCompile_Decorated, task ':compileKotlin')'
+is not assignable to 'org.gradle.api.tasks.TaskProvider'" exception
```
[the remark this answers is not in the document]
eskatos: It looks more like somewhere an arbitrary `Provider<SomeTask>` is created instead of using the real `TaskProvider<SomeTask>`. Basically what the error message says.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/KotlinNativeTargetPreset.kt
```diff
@@ -88,12 +89,14 @@ abstract class AbstractKotlinNativeTargetPreset<T : KotlinNativeTarget>(
...
 
-        SingleActionPerProject.run(project, "setUpKotlinNativePlatformDependencies") {
-            project.gradle.addListener(object : BuildAdapter() {
-                override fun projectsEvaluated(gradle: Gradle) {
-                    project.setUpKotlinNativePlatformDependencies()
-                }
-            })
+        if (!isConfigurationCacheAvailable(project.gradle)) {
+            SingleActionPerProject.run(project, "setUpKotlinNativePlatformDependencies") {
+                project.gradle.addListener(object : BuildAdapter() {
```
eskatos, line 94: You can't use `gradle.addListener(BuildListener)` but in that case you only need to react to `projectsEvaluated` and that can be done using `gradle.projectsEvaluated(action)`.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/targets/native/performance/KotlinPerformancePlugin.kt
```diff
@@ -66,17 +66,26 @@ open class KotlinPerformancePlugin : Plugin<Project> {
...
-        performanceExtension.trackedBinaries.forEach { binary ->
-            val perfReport = project.registerTask<NativePerformanceReport>(binary.target.disambiguateName(lowerCamelCaseName("perfReport", binary.name))) {
-                it.binary = binary
-                it.settings = performanceExtension
-                it.timeListener = timeListener
-                it.group = TASK_GROUP
-                it.description = "Report performance measurement results for binary '${binary.name}' of target '${binary.target.name}'."
+        if (!isConfigurationCacheAvailable(project.gradle)) {
+            val timeListener = TaskTimerListener(project)
+            project.gradle.addListener(timeListener)
```
eskatos, line 71: You can't use a `TaskExecutionListener` with the configuration cache, but you can replace it with a build service as a `OperationCompletionListener`, and it'll work both without and with the configuration cache.
See https://docs.gradle.org/release-nightly/userguide/build_services.html#operation_listener

LifeIsStrange:
I am a total noob, but as an end user of Kotlin 1.4 this PR took my curiosity:
Kotlin plugin will make use of Instant execution (an alias for Gradle cache system I believe) 
What benefits should this bring for end users?
Will it bring:
Quicker build time?
Rebuild time?
Gradle.kts editing performance?
Make the Kotlin frontend faster? (e.g faster autocomplete)
Quicker app program startup time?

eskatos:
Hi @LifeIsStrange,
This PR should really be titled "Gradle Configuration Cache support in the Kotlin Gradle Plugin".
You can read about Gradle's Configuration Cache at https://docs.gradle.org/release-nightly/userguide/configuration_cache.html

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/statistics/KotlinBuildStatsService.kt
```diff
@@ -106,21 +105,26 @@ internal abstract class KotlinBuildStatsService internal constructor() : BuildAd
...
+                                listenerRegistryHolder.listenerRegistry!!.onTaskCompletion(kotlinBuildStatProvider)
```
eskatos, line 112: :x: this won't work with the configuration cache, the listener passed to `onTaskCompletion()` must be a registered build service as explained at https://docs.gradle.org/current/userguide/build_services.html#receiving_information_about_task_execution, not an arbitrary provider

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/ParsedGradleVersion.kt
```diff
@@ -56,4 +57,12 @@ internal data class ParsedGradleVersion(val major: Int, val minor: Int) : Compar
...
+    try {
+        val startParameters = gradle.startParameter
+        startParameters.javaClass.getMethod("isConfigurationCache").invoke(startParameters) as? Boolean
```
eskatos, line 65: :x: that check is using reflection on internal API and exceptions for control flow and doesn't cache the result. This could have a bad effect on performance.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/ParsedGradleVersion.kt
nav-nav: I need an ability to check configuration cache because statistic listeners don't work for it right now and it was moved to another issue.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/BuildEventsListenerRegistryHolder.kt
nav-nav: New listener is not ready for configuration cache

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinGradleBuildServices.kt
nav-nav: New listener is not ready for configuration cache

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/statistics/KotlinBuildStatsService.kt
nav-nav: New listener is not ready for configuration cache

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/statistics/KotlinBuildStatsService.kt
nav-nav: Yes, stats isn't supported yet. fix for is is move to a separate issue

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/ParsedGradleVersion.kt
eskatos: I see, deferring implementation of non essential listeners like statistic collection sounds like a good tradeoff. Thanks for the information.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/statistics/KotlinBuildStatsService.kt
eskatos: Got it. But, then why register this at all? It won't work

eskatos:
This is now superseded by https://github.com/JetBrains/kotlin/pull/3571 and should be closed

nav-nav:
Updated PR is https://github.com/JetBrains/kotlin/pull/3571

PR #3522, opened 2020-07-04
Explicitly specify id of `KotlinNonJvmSourceRootConverterProvider`

develar:
IDEA 2020.3 uses `id` from extension definition to avoid creating instance of converter without need.

I do not change `KotlinNonJvmSourceRootConverterProvider` because using of deprecated constructor `ConverterProvider(id)` is still required as you don't have a special branch for 2020.3 IDE.


vladimirdolzhenko:
LGTM

PR #3523, opened 2020-07-05
KT-40016 Replace 'a.toLowerCase() == b.toLowerCase()' with 'a.equals(b, ignoreCase = true)' inspection

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40016

### DIFF
--- a/idea/resources-en/inspectionDescriptions/ReplaceWithIgnoreCaseEquals.html
+++ b/idea/resources-en/inspectionDescriptions/ReplaceWithIgnoreCaseEquals.html
@@ -1,0 +1,16 @@
+<html>
+<body>
+This inspection reports case insensitive comparison that can be replaced with <b>equals(..., ignoreCase = true)</b>.
+By using <b>equals()</b> you won't have to allocate any extra strings with <b>toLowerCase()/toUpperCase()</b> to do the comparison.
+<br><br>
+For example:
+<br>
+<pre>
+fun main() {
+    val a = "KoTliN"
+    val b = "KOTLIN"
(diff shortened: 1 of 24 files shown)


vladimirdolzhenko:
LGTM. Thank you for your contribution.

idea/resources-en/inspectionDescriptions/ReplaceWithIgnoreCaseEquals.html
```diff
@@ -0,0 +1,16 @@
...
+This inspection reports case insensitive comparison that can be replaced with <b>equals(..., ignoreCase = true)</b>.
+By using <b>equals()</b> you won't have to allocate any extra strings with <b>toLowerCase()/toUpperCase()</b> to do the comparison.
+<br><br>
+For example:
+<br>
+<pre>
+fun main() {
+    val a = "KoTliN"
+    val b = "KOTLIN"
+    println(a.toLowerCase() == b.toLowerCase()) // can be replaced with "a.equals(b, ignoreCase = true)"
```
ilya-g, line 12: Note that it's not an equivalent replacement in general, because `String.toLowerCase()` considers the current locale and potentially may transform one char into several during lowercasing. 

In particular this example results `false` in Turkish locale, whereas `a.equals(b, ignoreCase = true)` returns `true`.

yanex: @vladimirdolzhenko Should we then add the "may change semantics" text?

vladimirdolzhenko: Yes, we need to warn about it.

PR #3316, opened 2020-04-20
KT-25807: Fix Parcelize annotation implementing Parcelable.Creator<T>

rbares:
The JVM backend for `@Parcelize` currently ignores the generic type on `Parcelable.Creator<T>`, Consequently code requiring the correct `Parcelable` return type must cast from `Object`. This is particularly unsuitable for generated code such as AIDL, see https://youtrack.jetbrains.com/issue/KT-25807

Until the IR implementation in https://github.com/JetBrains/kotlin/pull/3271 is merged and becomes default, this PR provides a fix by implementing the generic interface as follows:
- Returning the relevant `Parcelable` type (`T`)
- Specifying the overridden methods, enabling the backend to generate suitable synthetic bridges
- Providing explicit class and field signatures to specify the type

Tests are updated accordingly, supported by a partial cherry-pick of https://github.com/JetBrains/kotlin/pull/3271/commits/e3c3e67e687f84886882048e7e339c28166b5056 from https://github.com/JetBrains/kotlin/pull/3271, with Steven's blessing. It should be noted that the unit tests for `android-extensions-compiler` were broken on master and do not appear to be included in CI.

These can be run as follows:
```
gradlew :plugins:android-extensions-compiler:test --tests "org.jetbrains.kotlin.android.parcel.*"
```

PR #3524, opened 2020-07-06
KT-39981: Fix JVM Parcelize use of Parceler

rbares:
The JVM backend for `@Parcelize` updated in https://github.com/JetBrains/kotlin/pull/3316 does not use `Parceler` correctly. Currently some calls are made to synthetic methods returning `Object`, rather than those returning the generic type `T`. Consequently attempts to use `Parceler` can fail as seen in https://youtrack.jetbrains.com/issue/KT-39981

This PR provides the following:
- Adds an explicit test for KT-39981
- Calls the correct variant of `create` and `newArray`
- Adds box tests for `newArray` handling, with and without custom `Parceler`
- Updates Parcelize codegen expected values for JVM
- Also updates Parcelize codegen expected value for the IR variant of customParcelablesSameModule


Mygod:
Can confirm it still is not working on v1.4.0.

rbares:
> Can confirm it still is not working on v1.4.0.

Understood, unfortunately https://github.com/JetBrains/kotlin/pull/3316 was reverted. This PR is intended to fix the issue which resulted in the revert but hasn't made it in. Have commented on youtrack regarding which branch this should be targeting. Suspect this is now likely to conflict with https://youtrack.jetbrains.com/issue/KT-40030.

yanex:
Merged to master manually (9d000529e70527b3f320ccb4baf8490e6efbb2c1). Will be merged to 1.4.20 after M1.

PR #3525, opened 2020-07-06
[FIR] Fix for the effective visibility checker

lunakoly:
There's a problem with the performance of the effective visibility checker and this line of code may be the problem.


mglukhikh:
Merged manually

PR #3526, opened 2020-07-06
FIR deserializer: load annotations on inner classes

jsjeon:
This is a follow-up work on https://github.com/JetBrains/kotlin/pull/3512, which fixed parameter shifts for constructors of inner classes and enums, but only enum tests were affected.

It turns out that, when visiting inner classes, deserializer in the parent context, i.e., outer class, has been always used. When creating "child" context, we need a fresh deserializer that loads that child class of interest. To do so, in the 1st commit, container source and annotation deserializer are passed down to child context creation accordingly.

That revealed one subtle case: annotations on fields inside a companion object whose actual field declared in the outer class. In that case, we need loaded annotation info from the outer, enclosing class, in addition to the companion class. The 2nd commit adds a logic to inherit annotation info from the parent deserializer.


jsjeon:
@demiurg906 PTAL

demiurg906:
@jsjeon Thank you

PR #3527, opened 2020-07-07
KT-40033 "Redundant empty class body" and "Cleanup code" remove final semicolon on enum values list

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40033


nikita-nazarov:
Unfortunately, this pr cannot be merged to the new repo directly, because `compiler/psi/src/org/jetbrains/kotlin/psi/ktElementUtils.kt` lies in the compiler submodule, which is separated from the new repo. I suggest you file a separate issue for the bug in `KtElement.deleteSemicolon()` or rewrite the current fix differently.

t-kameyama:
@nikita-nazarov 
https://github.com/JetBrains/intellij-kotlin/pull/32

PR #3528, opened 2020-07-07

t-kameyama:


vladimirdolzhenko:
Thank you so much for the contribution and addressing so old ticket

Merged manually as https://github.com/JetBrains/intellij-kotlin/commit/f53489e9d11ef8ba831ea379a98dd41b06299386

PR #3529, opened 2020-07-07
FIR: avoid cast exception in Candidate.prepareExpectedType

jsjeon:
The motivation is `Classes.kt1980` where `Int` extension, `times`, that inputs a lambda, is there to test resolutions against `Int.times(Int)`. For that top-level `times`, `fir` simply failed with a cast exception while retrieving the expected type for the argument. That is easy to avoid, since the underlying assumption was, when parameter is integer literal type, the corresponding argument type has been resolved, which is never guaranteed. (In the test, a lambda body is given as an argument, and thus we should discard that candidate, instead of cast exception.)


mglukhikh:
Merged, thank you!

PR #3530, opened 2020-07-07
FIR: handle getClassCall with integer literals

jsjeon:
Currently, class reference with integer literals, e.g., `1u::class`, is not converted to IR due to the remaining ILT type on the constant expression. However, ILT type bound to a constant expression can be replaced with the corresponding literal types if such constant expression is visited with expected type. Inspired by that, for getClassCall with integer literals, we can safely assume types of those literals and use them to remove ILT types.


mglukhikh:
Merged, thank you!

PR #3533, opened 2020-07-07
FIR deserializer: load annotations on extension receiver parameters

jsjeon:
Annotations can be targeted to extension receiver, like
```
fun @receiver:Ann String.myLength(@Ann q:String) ...
```
and then serialized at receiver type. This PR extends FIR deserializer to deserialize it.


dzharkov:
Merged manually!

PR #3534, opened 2020-07-08
KT-18125 "Wrap with let {...}" intention not available in all cases

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-18125


vladimirdolzhenko:
LGTM

Thank you

PR #3535, opened 2020-07-08
JVM IR: Mangle delegated properties with inline class type (fix KT-40125)

sfs:
We have to avoid mangling local delegated properties in `MemoizedInlineClassReplacements`, since they do not have corresponding properties. It turns out that the check for this was overly broad and also excluded delegated properties in classes.


udalov:
Thanks!

udalov:
Will push the following commits to 1.4.0: #3542

PR #3536, opened 2020-07-09
FIR: adjust type of integer operator call as property initializer

jsjeon:
The motivations are bb tests `evaluate.unary{Plus|Minus}` where unary operations ended up with the original integer literal type, while its usage (property initializer) implied different types:
```
const val prop4: Long = +1
```
Such unary operations are transformed to unary operator calls, and we should update such call's return type as property type.

------

My first trial was, at `FirCallCompleter#completeCall`, passing the expected type---property type, passed from declaration resolver---to the call completion transformation. That approach rather showed why we should _not_ pass the expected type, e.g., integer overflow, like:
```
val a: Long = 2147483647 + 1
if (a != -2147483648L) return "fail: in this case we should add to ints and than cast the result to long - overflow expected"
```

------

Next, I tried to reorder transformation of property initializer with various transformers, including `IntegerLiteralTypeApproximationTransformer` and `IntegerOperatorsTypeUpdater`, but no luck. In doing so, at least I learned the target expression `+1` was a call with an integer operator, and then transformed to a resolved, `FirIntegerOperatorCall`. Once resolved, neither transformer bailed out early during `transformFunctionCall`.

------

Hence, the current approach: during declaration resolution, passing the property type as an expected type, again, to the ILT approximator. That step is right after overall children transformations, including initializer. That order is necessary, otherwise integer operator in the initializer is still not resolved. Now, in the ILT approximator, which waits for only integer operator, we can handle integer operator call specially, since it _was_ an integer operator, technically. The original logic in the ILT approximator is to resolve the integer operator, but the integer operator call is already resolved. So, we can just adjust return type there.


demiurg906:
Hi @jsjeon, thank you for fix and sorry for very delayed answer

PR #3537, opened 2020-07-09
JVM IR: Fix generic signatures for special bridge methods

sfs:
Second attempt at #3163.

We produce special bridge methods for overrides of Java methods which have different names or signatures in the Kotlin standard library. Since these methods are visible to Java (they are non-synthetic bridge methods) we need to produce proper generic signatures to ensure that they are callable from Java.

The JVM backend currently does *not* do this in 6 out of the 7 cases where we have special bridge methods which are generic in Java. This can lead to Java code seeing less precise types for Kotlin collections. It does not break compilation, because of a quirk of javac, which skips the usual override checks for bridge methods (which, as far as I know, are always synthetic in Java).

Rather than replicating the behavior of the JVM backend, this PR tries to produce correct generic types for all generic special bridge methods. This requires additional bridges, but is technically required in order for Java interoperability.

It's worth noting that the behavior of the JVM backend cannot be replicated in a lowering, since the generic signatures created by the JVM backend for special bridges don't necessarily match the method signatures in some cases. See the example in #3163.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
  */
 
@@ -19,7 +19,14 @@ import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 
 data class SpecialMethodWithDefaultInfo(
-    val defaultValueGenerator: (IrSimpleFunction) -> IrExpression, val argumentsToCheck: Int, val needsArgumentBoxing: Boolean = false
+    val defaultValueGenerator: (IrSimpleFunction) -> IrExpression,
+    val argumentsToCheck: Int,
+    val needsArgumentBoxing: Boolean = false,
+    val needsGenericSignature: Boolean = false,
+)
+
+class BuiltInWithDifferentJvmName(
+    val needsGenericSignature: Boolean = false,
 )
 
 class SpecialBridgeMethods(val context: CommonBackendContext) {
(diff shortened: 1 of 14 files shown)


compiler/testData/codegen/box/jvm8/treeMapBridge.kt
```diff
@@ -1,4 +1,5 @@
 // TARGET_BACKEND: JVM
+// IGNORE_BACKEND_FIR: JVM_IR
```
sfs, line 2: I'm not sure why this test was passing. The generated IR is not correct, e.g., the fake override for `IntervalTreeMap.get` takes an argument of type `Any?` instead of `String`.

sfs:
I've recorded the reasoning behind the differences in behavior compared to the JVM backend in [KT-40277](https://youtrack.jetbrains.com/issue/KT-40277). Hopefully we can converge on the same behavior for the JVM backend as well.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
```diff
@@ -19,7 +19,14 @@ import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
 
 data class SpecialMethodWithDefaultInfo(
-    val defaultValueGenerator: (IrSimpleFunction) -> IrExpression, val argumentsToCheck: Int, val needsArgumentBoxing: Boolean = false
+    val defaultValueGenerator: (IrSimpleFunction) -> IrExpression,
+    val argumentsToCheck: Int,
+    val needsArgumentBoxing: Boolean = false,
+    val isGeneric: Boolean = false,
```
udalov, line 25: Perhaps `needsGenericSignature`? To distinguish from methods which are generic but don't need one

compiler/testData/codegen/bytecodeListing/specialBridges/signatures/nonGenericClass_ir.txt
```diff
@@ -0,0 +1,57 @@
...
+    public <null> method getSize(): int
+    public <null> method isEmpty(): boolean
+    public @org.jetbrains.annotations.Nullable <null> method put(@org.jetbrains.annotations.NotNull p0: java.lang.String, @org.jetbrains.annotations.NotNull p1: java.lang.String): java.lang.String
+    public synthetic bridge <null> method put(p0: java.lang.Object, p1: java.lang.Object): java.lang.Object
+    public @org.jetbrains.annotations.Nullable <null> method remove(@org.jetbrains.annotations.NotNull p0: java.lang.String): java.lang.String
+    public synthetic bridge <null> method remove(p0: java.lang.Object): java.lang.Object
+    public bridge final <null> method remove(p0: java.lang.Object): java.lang.String
+    public bridge final <null> method size(): int
+    private synthetic final field <Ljava/util/HashMap<Ljava/lang/String;Ljava/lang/String;>;> $$delegate_0: java.util.HashMap
+}
```
udalov, line 57: Please mention KT-40277 in the source file for this test and `partiallySpecializedClass.kt`, for future refernece

compiler/testData/codegen/bytecodeText/builtinFunctions/removeAt.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND: JVM_IR
```
udalov, line 1: Maybe we can encapsulate the difference in bytecode text in JVM/JVM_IR into `JVM_TEMPLATES` / `JVM_IR_TEMPLATES` directives in this test?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
```diff
@@ -127,19 +129,34 @@ private class BridgeLowering(val context: JvmBackendContext) : FileLoweringPass,
...
+    // one which the JVM backend currently handles correctly (MutableList.removeAt). For the rest, it's impossible
+    // to reproduce the behavior of the JVM backend in this lowering.
+    //
+    // Finally, we sometimes need to use INVOKESPECIAL to invoke an existing special bridge implementation in a
+    // superclass, which is what `superQualifierSymbol` is for.
     private data class SpecialBridge(
         val overridden: IrSimpleFunction,
         val signature: Method,
-        val specializedReturnType: IrType? = null,
+        // Does the underlying Java method contains generic type parameters.
```
udalov, line 148: This comment could be interpreted as if the Java method declares its own generic type parameters, which is not the case here (e.g. `Map.getOrDefault` does not declare new type parameters, it uses `Map`'s `V`). Maybe also use more abstract concepts such as "needs generic signature"?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
sfs: Good point, done.

compiler/testData/codegen/bytecodeListing/specialBridges/signatures/nonGenericClass_ir.txt
sfs: Done!

compiler/testData/codegen/bytecodeText/builtinFunctions/removeAt.kt
sfs: Done.

I forgot about this test - the reason for the difference is that the new bridge methods for generic special  bridges end up calling the instantiated version of `remove(I)` instead of `removeAt(I)`. This is necessary for type-safe special bridges - since there is additional logic in the wrapper - but unnecessary for methods which were merely renamed. We might want to match the behavior of the old backend in this case (apart from some missing bridges in this test).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/BridgeLowering.kt
sfs: I've expanded the comment and changed the field name to avoid potential confusion.

udalov:
Thank you!

PR #3538, opened 2020-07-09
gradle.kts: improve service initializtion

snrostov:
- prevent analyzing .gradle.kts files until all services are loaded
- remove services caches where it is not required
- replace cached services with cache only during vfs events batch processing
- prevent services loading in actions updating

### DIFF
--- a/idea/idea-core/src/org/jetbrains/kotlin/idea/core/script/ScriptConfigurationManager.kt
+++ b/idea/idea-core/src/org/jetbrains/kotlin/idea/core/script/ScriptConfigurationManager.kt
@@ -17,6 +17,7 @@
 package org.jetbrains.kotlin.idea.core.script
 
 import com.intellij.openapi.components.ServiceManager
+import com.intellij.openapi.components.serviceIfCreated
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.projectRoots.Sdk
 import com.intellij.openapi.util.Key
@@ -108,8 +109,8 @@ interface ScriptConfigurationManager {
     fun getAllScriptDependenciesSources(): List<VirtualFile>
 
     companion object {
-        fun getAlreadyInitializedInstance(project: Project): ScriptConfigurationManager? =
-            ServiceManager.getServiceIfCreated(project, ScriptConfigurationManager::class.java)
+        fun getServiceIfCreated(project: Project): ScriptConfigurationManager? =
+            project.serviceIfCreated()
 
         @JvmStatic
         fun getInstance(project: Project): ScriptConfigurationManager =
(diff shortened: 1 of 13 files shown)


idea/idea-core/src/org/jetbrains/kotlin/idea/core/script/ScriptConfigurationManager.kt
```diff
@@ -108,6 +108,9 @@ interface ScriptConfigurationManager {
     fun getAllScriptDependenciesSources(): List<VirtualFile>
 
     companion object {
+        fun getAlreadyInitializedInstance(project: Project): ScriptConfigurationManager? =
+            ServiceManager.getServiceIfCreated(project, ScriptConfigurationManager::class.java)
```
develar, line 112: In Kotlin you can use `serviceIfCreated<ScriptConfigurationManager>()`.

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/definitions/KotlinScriptDefinitionProvider.kt
```diff
@@ -26,6 +26,9 @@ interface ScriptDefinitionProvider {
     companion object {
         fun getInstance(project: Project): ScriptDefinitionProvider? =
             ServiceManager.getService(project, ScriptDefinitionProvider::class.java)
```
develar, line 28: In Kotin you can use `project.service<ScriptDefinitionProvider>()`.

idea/idea-core/src/org/jetbrains/kotlin/idea/core/script/ScriptConfigurationManager.kt
snrostov: Thanks! Fixed

plugins/scripting/scripting-compiler-impl/src/org/jetbrains/kotlin/scripting/definitions/KotlinScriptDefinitionProvider.kt
snrostov: I cannot use it here because it is compiler module, there is no platform-impl on the classpath (and in deepdencies)

idea/idea-core/src/org/jetbrains/kotlin/idea/core/script/ScriptConfigurationManager.kt
NataliaUkhorskaya: As I remember this method isn't present in 192 (or even 193), so we'll need a bunch file then

snrostov: Yep, it is:
https://buildserver.labs.intellij.net/viewLog.html?buildId=79223509&tab=buildResultsDiv&buildTypeId=Kotlin_KotlinDev_IdePlugin_IdeaUltimate_192

Will revert then

snrostov:
Merged manually

PR #3539, opened 2020-07-10
FIR2IR: avoid duplicate fake overrides for objects

jsjeon:
Members for `object` have been processed twice for some bb tests, resulting in duplicate fake overrides. Apparently, [KT-35730](https://youtrack.jetbrains.com/issue/KT-35730) has been fixed these days, and a temporary hack for that, which (re)visits objects as receiver, is the culprit.


compiler/tests-common/tests/org/jetbrains/kotlin/ir/AbstractIrTextTestCase.kt
```diff
@@ -164,7 +164,9 @@ abstract class AbstractIrTextTestCase : AbstractIrGeneratorTestCase() {
             private val elements = HashSet<IrElement>()
 
             override fun visitElement(element: IrElement) {
-                require(elements.add(element)) { "Non-unique element: ${element.render()}" }
+                require(elements.add(element) || element is IrConst<*>) {
```
jsjeon, line 167: I've tracked down where this was added, and it's for checking if there is a duplicate array access (represented as `IrGetVariable`). I strongly believe having same const should not matter. Actually, "constant canonicalization" (so that the local function can avoid loading the same constants again) is well-known optimization.

mglukhikh:
Sorry but I've just fixed this issue, and removed these lines.

mglukhikh:
See https://github.com/JetBrains/kotlin/commit/6b964cb61de316cf154527a41380dc6a78ca8df4 and https://github.com/JetBrains/kotlin/commit/fbbf4e06ba667ef7055ce08362d3f3d4c8252f3c

PR #3540, opened 2020-07-10
KT-39815 Intention to swap true and false statements (flip `if`)

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39815


vladimirdolzhenko:
I have some objection in ticket itself.

t-kameyama:
PTAL

nikitabobko:
This PR is declined because we decided that we don't want to have such intention. If you have strong arguments for this intention then please write them into the issue. Thanks!

https://youtrack.jetbrains.com/issue/KT-39815#focus=Comments-27-4403521.0-0

PR #3543, opened 2020-07-12
FIR2IR: don't create synthetic class for enum entry w/o members

jsjeon:
The motivation is `Constructors.enumEntry`:
```
...
enum class TestEnum(val id: String? = null) {
  ENUM1(id = "enum_id1")
  ENUM2(id = "enum_id2") { ... // members }
  ...
}

fun box() {
  assertEquals(listOf("fun <init>(kotlin.String?): test.TestEnum"), TestEnum.ENUM1::class.constructors.map { it.toString() })
  assertEquals(listOf(), TestEnum.ENUM2::class.constructors.map { it.toString() })
}
```
which intentionally checks if `ENUM2`, an enum entry with (overridden) members, doesn't have its own primary constructor. At the same time, it checks, for `ENUM1`, an enum entry without any members, doesn't introduce a synthetic class and has a delegating constructor (to its owner).

------

First of all, `fir` serializer has a bug that doesn't serialize enum entries at all, which is also fixed in this PR.

------

`fir` always created synthetic class for enum entries by building an anonymous object and setting it as entry initializer. Inside that anonymous object, the primary constructor has a delegating constructor call. So, one approach I tried was to tweak that step such that an anonymous object is created only if there are members; if not, a similar delegating call is set as an initializer. However, `FirDelegatedConstructorCall` is a statement, not an expression. That is, a wrapping block as an expression is needed, which bothers expression resolution (since that `block` introduces an extra scope).

Fortunately, `IrEnumEntryImpl.correspondingClass` is a knob to introduce a synthetic class. Without affecting `fir` building and resolution, i.e., while still building and using an anonymous object as a holder for an enum entry, during `fir2ir`, we can drop it and avoid creating a synthetic class for an enum entry if it doesn't have members. The only caveat is not to forget to create `IrEnumConstructorCallImpl` for that case. As mentioned, a delegating call is stored inside the anonymous object's primary constructor, so we can extract it and convert it to the corresponding IR element.


mglukhikh:
Looks good, thank you!

mglukhikh:
Merged manually 👍

PR #3544, opened 2020-07-12
Add dependency scopes option for scripting (Ivy and Maven) resolvers

ileasile:


libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/resolverNamedOptions.kt
```diff
@@ -0,0 +1,31 @@
...
+    flag(name.key)
+
+operator fun MutableMap<String, String>.set(key: DependenciesResolverOptionsName, value: String) {
+    put(key.key, value)
+}
+
+
+enum class DependenciesResolverOptionsName {
+    TRANSITIVE,
+    DEPENDENCY_SCOPES;
```
ligee, line 21: Using underscores seems unusual for the maven configuration, and in maven this parameter called just `scope`, so I'd rename it, so the people can guess the param name easier.
If we want to have the naming flexibility - I'd add the string name parameter to the enum (and use `toLowerCase` as a default value).

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/resolverNamedOptions.kt
```diff
@@ -0,0 +1,31 @@
...
+
+fun ExternalDependenciesResolver.Options.flag(name: DependenciesResolverOptionsName) =
+    flag(name.key)
+
+operator fun MutableMap<String, String>.set(key: DependenciesResolverOptionsName, value: String) {
+    put(key.key, value)
+}
+
+
+enum class DependenciesResolverOptionsName {
```
nerdsupremacist, line 19: These options seem very specific to the individual dependencies resolvers. Perhaps it would be more appropriate to prefix this with a `Maven*` or `Ivy*` respectively.

This API is not implemented by many other resolvers that I know of. But what I want to avoid is someone implementing a dependencies resolver and thinking they:
1. have to support these exact options
1. are only allowed to support these exact options

@ligee what do you think?

nerdsupremacist:
other then that little philosophical discussion: lgtm

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/resolverNamedOptions.kt
ileasile: Fixed this name and added a parameter

libraries/scripting/dependencies/src/kotlin/script/experimental/dependencies/resolverNamedOptions.kt
ligee: Thanks, @nerdsupremacist, I think the concern is valid.
But we can probably get away right now simply by moving it into the `impl` package, signalling that it is a mere implementation detail now.
@ileasile, what do you think about this?

ileasile: I see these options names only as a convenient and safe way to access "text" properties. Maybe it's worth adding somewhere in comments that these options are not mandatory? Also, I'm OK with moving it into the `impl` package.

ligee: I understand your intention here, but we don't want to expose it as a generic API. So, please move it to the `impl`.

ileasile: Sure. Done ✔️

PR #3545, opened 2020-07-12
KT-40215 "Create abstract function" quick fix suggested even though surrounding class is non-abstract

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40215


fedochet:
@t-kameyama thanks for your contribution!

PR #3548, opened 2020-07-13
FIR: handle vararg in annotations

jsjeon:
The motivation was `Annotations.varargInAnnotationParameter`, which failed with unsupported const element of `IntArray` type.

The 1st commit relaxes the condition to create argument mapping for annotation calls. Previously, only named arguments are transformed to an argument mapping.

The 2nd commit then adds support for vararg in annotation calls. It mainly reuses the logic in `FirCallCompletionResultsWriterTransformer`, which converts an argument into `FirVarargArgument` if the corresponding parameter is vararg. Such vararg argument will be properly translated during `fir2ir`.

During that, there was one regression, which revealed a corner case that `IrConstTransformer` didn't handle well: vararg with spread elements, e.g., `@Ann(varargP = *["/"])`. The 3rd commit addresses that.

Finally, the remaining error in the original test revealed that `FirArrayOfCallTransformer` didn't transform an arrayOf call with empty argument, e.g., `@Ann(varargP = *intArray())`, into `FirArrayOfCall`, making `IrConstTransformer` give up that call. The 4th commit relaxes the condition to determine arrayOf calls.


jsjeon:
@vana06 PTAL 3rd commit. @mglukhikh as usual, PTAL. :)

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/evaluate/IrConstTransformer.kt
```diff
@@ -60,18 +61,33 @@ class IrConstTransformer(irBuiltIns: IrBuiltIns) : IrElementTransformerVoid() {
...
-            this.putElement(i, irVarargElement.transformSingleArg(this.varargElementType))
+    private fun IrVararg.transformVarArg(): IrVararg {
+        if (elements.isEmpty()) return this
+        val newIrVararg = IrVarargImpl(this.startOffset, this.endOffset, this.type, this.varargElementType)
+        for (element in this.elements) {
+            when (element) {
+                is IrExpression -> newIrVararg.addElement(element.transformSingleArg(this.varargElementType))
+                is IrSpreadElement -> {
+                    when (val expression = element.expression) {
+                        is IrVararg -> {
```
vana06, line 79: I would prefer here single line representation
`is IrVararg -> expression.transformVarArg().elements.forEach { newIrVararg.addElement(it) }`

jsjeon: Thank you for the suggestion! Done.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -691,19 +693,40 @@ open class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransform
...
-                            mapArguments(it.arguments, calleeReference.resolvedSymbol.fir as FirFunction<*>)
-                                .toArgumentToParameterMapping()
+                annotationCall.getCorrespondingConstructorReferenceOrNull(session)?.let { calleeReference ->
+                    val callee = calleeReference.resolvedSymbol.fir as FirFunction<*>
+                    val argumentMapping = mapArguments(it.arguments, callee).toArgumentToParameterMapping()
+                    val varargParameter = callee.valueParameters.firstOrNull { param -> param.isVararg }
+                    if (varargParameter == null) {
                         it.replaceArgumentList(buildResolvedArgumentList(argumentMapping))
+                    } else {
+                        // TODO: refactor/reuse [Candidate#handleVarargs] in [FirCallCompletionResultsWriterTransformer]
```
mglukhikh, line 703: I will do it to avoid code duplication

mglukhikh:
Merged manually 👍

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Found https://github.com/JetBrains/kotlin/commit/07b0ffef8473d7c8518a3a24f753605bdfd9cc8a. Thanks!

PR #3552, opened 2020-07-14
KT-39604 "Package directive doesn't match file location" quick fix does not insert a space between keyword `package` and the package name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39604


darthorimar:
Thanks!

PR #3556, opened 2020-07-15
Fix KotlinCoreEnvironment projectCount disposable

gharrma:
The disposable which decrements ourProjectCount was a non-capturing
lambda, meaning it had a single identity and could only be registered
in the Disposer tree once. Therefore the application environment
would never be disposed if ourProjectCount ever went above 1.


gharrma:
Note: I think this is a regression introduced in `63c51331678dff4780c2b34c85cac230a5e394ab`

ligee:
Merged manually

ligee:
@gharrma thank you for the contribution and sorry for the delay.

udalov:
Note that this caused an issue in parallel builds in 1.4.30: https://youtrack.jetbrains.com/issue/KT-45007#focus=Comments-27-4735847.0-0

ligee:
https://youtrack.jetbrains.com/issue/KT-45289

chao2zhang:
is it possible to update Kotlin documentation warning the behavioral change for Kotlin SAM conversion?
https://kotlinlang.org/docs/fun-interfaces.html#sam-conversions does not mention this at all and it may appear seemingly safe to always convert object to SAM when possible.

udalov:
^ @p7nov Please consider. :)

p7nov:
@chao2zhang 
Thanks for pointing, I've created an [issue for the doc update](https://youtrack.jetbrains.com/issue/KT-45708) in YouTrack.

PR #3557, opened 2020-07-15
KT-39421 KJS: False positive "Implicit (unsafe) cast from dynamic" inspection for Any? parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39421


nikita-nazarov:
Merged to the new repo https://github.com/JetBrains/intellij-kotlin/commit/477053a955d16c31076cb00a11de6ee5118bbc26. Thank you for the contribution!

PR #3558, opened 2020-07-15
FIR: consider more function types during SAM resolution

jsjeon:
The motivation is `annotatedLambda.samFunReference`:

```
...
testClass(Test.applyABC(::annotatedABC), "1")
...
val abcReference = ::annotatedABC
testClass(Test.applyABC(abcReference), "3")
```

The first call is resolved well with the necessary SAM conversion, whereas the second one, which just uses the stored local function reference, is not resolved. The main reason is, once stored to the local property, its type is `kotlin.reflect.KFunction2`. It is a subtype of `kotlin.jvm.functions.Function2`, which is a subtype of `kotlin.Function`. Inference util simply considered only `Function` and `SuspendFunction`, which should have included `KFunction` and `KSuspendFunction` as built-in functional types.

------

This reveals a bug in `fir` serializer, which considers only `SuspendFunction` when transforming it to runtime function type.


compiler/fir/fir-serialization/src/org/jetbrains/kotlin/fir/serialization/FirElementSerializer.kt
```diff
@@ -607,9 +607,11 @@ class FirElementSerializer private constructor(
 
     private fun transformSuspendFunctionToRuntimeFunctionType(type: ConeClassLikeType): ConeClassLikeType {
         val suspendClassId = type.classId!!
-        val runtimeClassId = FunctionClassDescriptor.Kind.Function.let {
-            ClassId(it.packageFqName, Name.identifier(suspendClassId.relativeClassName.asString().drop("Suspend".length)))
-        }
+        val relativeClassName = suspendClassId.relativeClassName.asString()
+        val kind =
+            if (relativeClassName.startsWith("Suspend")) FunctionClassDescriptor.Kind.Function
```
punzki, line 612: Very minor optimizations: You could just check if it starts with `'K'` to determine if it's a `KSuspendFunction`. And use `replaceFirst` instead of `replace` below.

jsjeon: Thanks for the suggestions, and good to learn `replaceFirst`. Both are applied.

mglukhikh:
Merged, thank you!

PR #3560, opened 2020-07-15
Make linker visible to compiler plugins

jimgoog:


erokhins:
@romanart Please take a look. For me -- it is ok.

romanart:
Could you please provide a small description why do you need it?

jimgoog:
We ran into a whole bunch of situations where we found ourselves needing to bind symbols from within our IR transforms.  You can see our usages by searching for [.linker](https://cs.android.com/search?q=.linker&sq=&ss=androidx) and [bindIfNecessary](https://cs.android.com/search?q=.bindIfNecessary&sq=&ss=androidx).  We still need to do some additional cleanup of that code, but if you see anything we're doing that's has a better solution, let us know!

romanart:
Do you have any ETA when this hack could be dropped? BTW, I wonder why `context.reference{Class/Functions}` don't work for you?

jimgoog:
I think maybe this hack could be dropped in a few days but I'm not 100% sure.  A large part of it is that we were migrating from code where we were manually binding everything with the irProviders, so switching to irLinker was exactly the API that matched our old code.  I think `context.reference*` might work for us, but when we were doing the migration we had lots of cases where we got symbols and it wasn't always obvious how to go from a symbol to a fqName, but I'm happy to take another look at it and let you know how it goes.

romanart:
@jimgoog Thanks a lot for explanation. Am I correct that some descriptors are still being created and such code is already considered as a debt and scheduled to be refactored soon? Until it done you need a hacky access to `irLinker` to keep your plugin work during the transition period.

jimgoog:
Yes, the 1.4 rebase introduced a lot of technical debt.  Our transforms do a bunch of things that are artifacts of the old APIs (like using descriptors everywhere and for everything).  Accessing descriptors on IR nodes is now deprecated, so we need to move away from that.  The first step was just to land the rebase though, suppress the deprecation errors, and then start cleaning up the code so we aren't using deprecated APIs.

PR #3561, opened 2020-07-15
Keep com.google dependencies intact

jimgoog:
It would be much more preferable if the `com.google` dependencies could be repackaged to live in a different namespace so they don't conflict with the Google-provided packages.  But at the very least, if we're going to include them in the Google package, it's helpful to keep all of them, to avoid subtle compatibility issues when only some classes are overridden.

### DIFF
--- a/prepare/compiler/compiler.pro
+++ b/prepare/compiler/compiler.pro
@@ -100,6 +100,8 @@
     public protected *;
 }
 
+-keep class com.google.common.** { *; }
+
 -keep class org.jetbrains.kotlin.container.** { *; }
 
 -keep class org.jetbrains.org.objectweb.asm.Opcodes { *; }


erokhins:
It seems that only guava library are affected by this change. @jimgoog Did you saw any problems with another libraries?

jimgoog:
No, just guava for us.

erokhins:
I'm going to check the size of our comiper after that change.

jimgoog:
Another option would be to rejar/jarjar the guava library so the package name is different and thus doesn't conflict.  Then it wouldn't impact compiler size.

erokhins:
Yes, but for 1.4.0 it not seems feasible.

erokhins:
Well, it is +5.5mb (114.3 -> 120 mb). Maybe there is another way how to tweak it?

erokhins:
Hm. that strange -- guava lib by itselv -- 2 mb. Keep digging

erokhins:
I've got reasonable result (+2mb) with 
> -keepclassmembers class com.google.common.** { *; }
@udalov @jimgoog WDYT?

jimgoog:
That still seems fragile, but I suspect that might resolve our issue.  I will make that change locally and re-run our tests to see if it breaks us.

udalov:
+2Mb still seems like a lot, honestly, and I'd prefer if we shaded the dependency, or even got rid of it (the latter might not be easy though because of its usages in intellij-core). But for 1.4.0, it's probably the safest option.

jimgoog:
@erokhins Yes, I just tested out `-keepclassmembers class com.google.common.** { *; }` and that does indeed seem to keep everything passing for us.  Thanks!

erokhins:
Merged manually: https://github.com/jetbrains/kotlin/commit/814bd48148ae99199a78ceae1e5cd4da1fd7bdfc

PR #3564, opened 2020-07-16
[FIR] refactoring for running extended checkers

vldF:


mglukhikh:
Branch isn't in consistent state.

PR #3568, opened 2020-07-16
FIR2IR: store and convert annotations on enum entry

jsjeon:
Annotations on enum entry should be associated with that enum entry _and_ a synthetic class (if any) for that enum entry. `fir` missed two things.

First, during raw FIR building, annotations were bound to the initializing anonymous object only. Those should be stored at the enum entry itself too. Once it's there, the translation of annotations on classes is already in effect.

Second, annotations on enum entries are not converted during `fir2ir`. With the unified annotation generator, it's just one-liner.


compiler/fir/raw-fir/psi2fir/src/org/jetbrains/kotlin/fir/builder/RawFirBuilder.kt
```diff
@@ -653,12 +653,12 @@ class RawFirBuilder(
                     isExpect = containingClassOrObject?.hasExpectModifier() == true
                 }
                 symbol = FirVariableSymbol(callableIdForName(nameAsSafeName))
-                // NB: not sure should annotations be on enum entry itself, or on its corresponding object
```
jsjeon, line 656: Well, actually annotations should be on "both" places. :)

mglukhikh:
Merged 👍

PR #3569, opened 2020-07-17
KT-39467 False negative "Move variable declaration into when" if a variable declaration is placed on a new line

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-39467


idea/testData/inspectionsLocal/moveVariableDeclarationIntoWhen/multiLinePropertywithSingleLineInitializer2.kt
```diff
@@ -0,0 +1,10 @@
+fun test() = 42
+
+fun foo() {
+    val <caret>a =
```
dimonchik0036, line 4: I think this is incorrect behavior. Inspection should be available in cases:
* this is one-line property
* this is property with one-line initializer and action == INLINE

dimonchik0036 (changes requested):

dimonchik0036 (approved):

dimonchik0036:
Thanks!

PR #3570, opened 2020-07-17
FIR2IR: set proper visibility of backing fields with @JvmField

jsjeon:
Property with `@JvmField` will not have accessors, and thus the visibility of its backing field should conform to the property visibility, not the default-ish `private`.

This is a belated follow-up of https://github.com/JetBrains/kotlin/pull/3243


mglukhikh:
Ok, I'll wait for the end of RR and merge it.

mglukhikh:
Merged 👍

PR #3574, opened 2020-07-17
JVM IR: Do not create accessors for enum entry constructors

sfs:
Enum entry constructors are package private and only called from the enclosing enum class, so we don't need to generate accessors.


udalov:
Thanks!

PR #3575, opened 2020-07-17
KT-40403: return annotations for ultra-light extension method receiver parameter

kevin1e100:
Proposed fix for https://youtrack.jetbrains.com/issue/KT-40403


kevin1e100:
added previously forgotten test for extension property, sorry about that

xiexed:
Similar fix is applied by  2c16ae968db5b00ced46b516da9cc314442fc617 . Thank you!

PR #3576, opened 2020-07-18
FIR2IR: convert annotations on type aliases

jsjeon:
`typealias` can have annotations, e.g.,
```
@Ann(2)
typealias TA = Any
```
`fir` should store the converted annotation to the corresponding IR element for the type alias.

One caveat is, `typealias` is registered at `Fir2IrClassifierStorage`, which is not a good place to have an annotation generator. Instead, similar to enum entries, we can cache the registered IR type alias, and (re)visit that inside visitor.


mglukhikh:
Merged, thank you!

PR #3579, opened 2020-07-18
Set wantsDiagnostics() in AbstractFilteringTrace

gharrma:
report() already delegates to the parentTrace, so
wantsDiagnostics() should delegate to it as well.

cc @dsavvinov who wrote `AbstractFilteringTrace`.
(Does this change look right to you?)

This change came about while investigating Android Lint performance,
which uses `TopDownAnalyzerFacadeForJVM`.

### DIFF
--- a/compiler/frontend/src/org/jetbrains/kotlin/resolve/AbstractFilteringTrace.kt
+++ b/compiler/frontend/src/org/jetbrains/kotlin/resolve/AbstractFilteringTrace.kt
@@ -39,4 +39,6 @@ abstract class AbstractFilteringTrace(
     override fun report(diagnostic: Diagnostic) {
         parentTrace.report(diagnostic)
     }
+
+    override fun wantsDiagnostics(): Boolean = parentTrace.wantsDiagnostics()
 }


dsavvinov:
Hi, @gharrma !

Could you please elaborate, what exactly you're trying to achieve? I guess, you want diagnostics not to be reported if `parentTrace.wantsDiagnostics() == false`? If that's the case, then it looks a little bit suspicious for me that all clients will be obliged to check `wantsDiagnostics()` before reporting it. Other traces do not behave like that - in particular, `DelegatingBindingTrace` implements `report` as no-op, if `wantsDiagnostics() == false`. 

Again, if that's what you're actually trying to achieve, then maybe it's better to call super-constructor of `DelegatingBindingTrace` with proper arguments - in particular, with `withParentDiagnostics = false` and with `filter = BindingTraceFilter.NO_DIAGNOSTICS`.

Alternatively, you could add something like `if (!parentTrace.wantsDiagnostics()) return` in `report`-overload.

gharrma:
Hi @dsavvinov  !

First, I can explain what I want to achieve. Here's the background:

- Android Lint uses `TopDownAnalyzerFacadeForJVM` to be able to inspect Kotlin sources.
- While profiling Lint I noticed that there was some time spent in some Kotlin compiler checks that we don't need, including subclasses of `CallChecker` such as `CallReturnsArrayOfNothingChecker`.
- I noticed that `CallCompleter.completeCall()` checks the value of `context.trace.wantsDiagnostics()` before running call checkers. So, in Lint we can save some time by using a custom `BindingTrace` which returns false from `wantsDiagnostics()`.
- This worked well, however the call checkers were still running in some cases. The cause turned out to be `EffectsFilteringTrace`, which wraps the original `BindingTrace` but does not forward its `wantsDiagnostics()` value.

So, the PR is really a fix for `EffectsFilteringTrace`, if my understanding is correct.

Also, the PR is based on `TemporaryBindingTrace`, which already does something similar: it overrides `wantsDiagnostics()` and uses the value from its delegate, even though it also sets `withParentDiagnostics=true`.

Let me know what you think!

dsavvinov:
Ok, an example of `TemporaryBindingTrace` sounds convincing enough for me :) Merged, thanks!

PR #3580, opened 2020-07-18
Kotlin-test: floating point assertions.

ReneeVandervelde:
Add Float and Double implementations of `assertEquals` and `assertNotEquals` that take a delta for comparing floating point values within a tolerance.

Fixes [KT-8364](https://youtrack.jetbrains.com/issue/KT-8364)

### DIFF
--- a/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
+++ b/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
@@ -75,6 +75,18 @@ fun <@OnlyInputTypes T> assertNotEquals(illegal: T, actual: T, message: String? 
     asserter.assertNotEquals(message, illegal, actual)
 }
 
+/** Asserts that [actual] is not within a [delta] value of the [illegal] value with an optional [message]. */
+@SinceKotlin("1.4")
+fun assertNotEquals(illegal: Double, actual: Double, delta: Double, message: String? = null) {
+    asserter.assertNotEquals(message, illegal, actual, delta)
+}
+
+/** Asserts that [actual] is not within a [delta] value of the [illegal] value with an optional [message]. */
+@SinceKotlin("1.4")
+fun assertNotEquals(illegal: Float, actual: Float, delta: Float, message: String? = null) {
+    asserter.assertNotEquals(message, illegal, actual, delta)
+}
+
 /** Asserts that [expected] is the same instance as [actual], with an optional [message]. */
 fun <@OnlyInputTypes T> assertSame(expected: T, actual: T, message: String? = null) {
     asserter.assertSame(message, expected, actual)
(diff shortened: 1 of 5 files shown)


JakeWharton:
Have you considered adding a default for `delta` so that it only need specified if the error-bound is known? There's more reading on it here: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#floating-point-comparison

ReneeVandervelde:
@JakeWharton Yeah I had that thought too. I saw there's [another issue] that discussed adding a general extension for that. @elizarov seemed to think it didn't provide much value there, and I somewhat agree. I left it out here to keep it simple and keep it in-line with most testing frameworks' asserters.

If we did go that route, it would also introduce a little ambiguity to the signature if that parameter were optional, so I might suggest just having separate methods with different names anyway.

[another issue]: https://youtrack.jetbrains.com/issue/KT-14911

JakeWharton:
Cool just wanted to make sure it had been seen. I learned about it from https://github.com/google/truth/issues/690. I suppose an `assertNear` could be added in the future if it's desired.

qurbonzoda:
The issue is updated:

> We've discussed this issue and decided to introduce overloads for assertEquals and assertNotEquals taking three numbers, e.g.
```
assertEquals(expected: Double, actual: Double, absoluteTolerance: Double, message: String? = null)
assertNotEquals(illegal: Double, actual: Double, absoluteTolerance: Double, message: String? = null)

// same for Float
```

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -57,11 +58,35 @@ fun <@OnlyInputTypes T> assertEquals(expected: T, actual: T, message: String? =
     asserter.assertEquals(message, expected, actual)
 }
 
+/** Asserts that [actual] is within a [delta] value of the [expected] value with an optional [message]. */
+@SinceKotlin("1.4")
```
qurbonzoda, line 62: `@SinceKotlin("1.5")`

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -57,11 +58,35 @@ fun <@OnlyInputTypes T> assertEquals(expected: T, actual: T, message: String? =
...
+fun assertEquals(expected: Double, actual: Double, delta: Double, message: String? = null) {
```
qurbonzoda, line 63: Please rename all `delta` arguments name to `absoluteTolerance`

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -87,6 +87,46 @@ class BasicAssertionsTest {
...
+
+    @Test
+    fun testAssertNotEqualsFloat() {
+        assertNotEquals(0.1f, 0.3f, 0.1f)
+    }
+
+    @Test
+    fun testAssertNotEqualsFloatFails() {
+        checkFailedAssertion { assertNotEquals(0.1f, 0.11f, .1f) }
+    }
```
qurbonzoda, line 128: Please add tests for `NaN` and `Infinite` values

qurbonzoda (changes requested):
Changes requested

qurbonzoda:
Moved to https://github.com/JetBrains/kotlin/pull/4275

PR #3581, opened 2020-07-19
FIR: account for vararg when creating KFunction type for callable reference

jsjeon:
The motivation is `callableReference.kt37604`:

```
fun useUnit(fn: () -> Unit) {
    fn.invoke()
}
...
var cWithVarargInit = false

class CWithVararg(vararg x: Int) {
    init {
        cWithVarargInit = true
    }
}

fun box(): String {
    ...
    useUnit(::CWithVararg)
    if (!cWithVarargInit) throw AssertionError("cWithVarargInit")

    return "OK"
```

Other class references with no parameter or default value parameter were resolved well, but that one with `vararg` wasn't. It turns out that, when computing the resulting type for callable reference of the candidate, if it is a class reference, its primary constructor is the call target, and while collecting value parameter of interest, `vararg` is simply missed. Since its use is callable reference, i.e., no arguments, we can ignore `vararg` when computing the `KFunction` argument number, similar to a value parameter with a default value.

With a concrete example, before this change, the resulting type for the candidate was `KFunction1<IntArray, Unit>` even though `vararg x` is not given (since it's a callable reference). After this change, its resulting type is correctly `KFunction0<Unit>`, which is what `useUnit` is expecting.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ResolutionStages.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ResolutionStages.kt
@@ -295,7 +295,10 @@ private fun FirSession.createKFunctionType(
         else -> expectedParameterNumberWithReceiver
     }
     for ((index, valueParameter) in function.valueParameters.withIndex()) {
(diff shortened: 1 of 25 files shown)


mglukhikh:
Merged, thank you!

PR #3584, opened 2020-07-20
KT-20357: Add a sample for requireNotNull

yurano:
In the sample code, I think it's easier to understand if we emphasize the effect that requireNotNull can convert an argument to a non-null value.

### DIFF
--- a/libraries/stdlib/samples/test/samples/misc/preconditions.kt
+++ b/libraries/stdlib/samples/test/samples/misc/preconditions.kt
@@ -30,17 +30,26 @@ class Preconditions {
     @Sample
     fun failRequireNotNullWithLazyMessage() {
 
-        fun execWithParams(params: Map<String, String?>) {
-            val required = requireNotNull(params["required"]) { "Required key must be non-null" } // returns a non-null value
+        fun printRequiredParam(params: Map<String, String?>) {
+            val required: String = requireNotNull(params["required"]) { "Required value must be non-null" } // returns a non-null value
             println(required)
             // ...
         }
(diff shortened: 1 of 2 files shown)


yurano:
@ilya-g 
Do you think this sample is not necessary?
Should I close this PR?

ilya-g (approved):
No, everything seems fine with these PR, we can merge it soon.

libraries/stdlib/samples/test/samples/misc/preconditions.kt
```diff
@@ -27,6 +27,22 @@ class Preconditions {
         assertPrints(getIndices(3), "[1, 2, 3]")
     }
 
+    @Sample
+    fun failRequireNotNullWithLazyMessage() {
+
+        fun execWithParams(params: Map<String, String?>) {
+            val required = requireNotNull(params["required"]) { "Required key must be non-null" } // returns a non-null value
```
ilya-g, line 34: Though, if you have a bit more time, could you add another variant of `requireNotNull` usage that utilizes its contract allowing to smartcast its parameter. Something like:
```
val requiredParam: String? = params["required"]
requireNotNull(requiredParam) { "Required value must be non-null" }
// now requiredParam is smartcast to String
```

I think it could be done with another local function.

yurano: OK, I work on it soon.

ilya-g (approved):

ilya-g:
Thanks for the sample!

PR #3586, opened 2020-07-20
Parcelize: Handle (suspend) function types

sfs:
Before #3380, there was a bug in the kotlin-android-extensions where we would not produce error messages for Parcelize under certain circumstances. In most case this just resulted in broken code which failed at runtime, but for `@Parcelize` classes containing function types we ended up producing perfectly valid code (calling `read/writeValue` as far as I can tell). Since functions in Kotlin are always serializable, this works and introducing error messages broke some working code.

This PR adds support for function types in Parcelize. I'm not convinced that this is a good idea, since it'll use java serialization. Maybe we should also add a deprecation warning at the same time?

---

The other two commits in this PR are just updates for the test runner and expectations.

### DIFF
--- a/plugins/android-extensions/android-extensions-compiler/src/org/jetbrains/kotlin/android/parcel/ir/IrParcelSerializerFactory.kt
+++ b/plugins/android-extensions/android-extensions-compiler/src/org/jetbrains/kotlin/android/parcel/ir/IrParcelSerializerFactory.kt
@@ -228,7 +228,9 @@ class IrParcelSerializerFactory(symbols: AndroidSymbols) {
             classifier.isEnumClass ->
                 return wrapNullableSerializerIfNeeded(irType, IrEnumParcelSerializer(classifier))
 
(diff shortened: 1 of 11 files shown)


sfs:
Rebased on master, updated test expectation and added another test (for a Parcelable sealed class).

@yanex Did you have time to look at this (and #3437) yet?

yanex:
Merged to master manually:

https://github.com/JetBrains/kotlin/commit/77fb0ddd329cf9d6174a17c16f0eb3868842a8b6
https://github.com/JetBrains/kotlin/commit/9bf3f105d96b9e5fec2d1534ebdc41c4a877a903
https://github.com/JetBrains/kotlin/commit/a2418484bbde648c5ad357193457a22f6e2e6611
https://github.com/JetBrains/kotlin/commit/aca9478998f3bcef34e97648f3e93cd051696478

Will be merged to 1.4.20 as well after M1.

PR #3588, opened 2020-07-21
FIR: special visibility handling for monitor{Enter|Exit}

jsjeon:
`monitor{Enter|Exit}` are defined in `kotlin.jvm.internal.unsafe.MonitorKt` with `private` visibility, but they're "intrinsics." During visibility check in resolution, we need a special handling to allow them to be accessed from anywhere.

IR functions for those, however, had an incorrect `parent`: external package fragment, in lieu of facade class. `SymbolTable` walks through registered symbols and returns those with `DescriptorWithContainerSource` (which usually refers to such facade class), and then stub generator will update `parent` of those symbols with the facade class. Therefore, `fir` should have used `DescriptorWithContainerSource` if container source is available.

Such correct `parent` class setting matters because backend's synthetic accessor lowering depends on it. Proper accessor lowering also matters because the final codegen, which looks up pre-defined intrinsics and replaces them with pre-defined instructions, searches for lowered accessor, like `access$monitor{Enter|Exit}`.


mglukhikh:
Merged, thank you!

PR #3589, opened 2020-07-21
FIR: handle object invoke via type alias

jsjeon:
The motivation is bb test `typealias/typeAliasObjectCallable.kt`:
```
object O {
    ...
    operator fun invoke() = ...
}
typealias A = O
fun box(): String = A()
```
Currently, `A()` is resolved as object `O`'s private constructor, resulting in `HIDDEN` applicability of the candidate. Rather, `O.invoke()` should be chosen as the call target. This is because, during `invoke` candidate resolution, only callable symbol and class symbol were assumed and handled. Since type alias has the expand type on it, it could be treated similar to regular class. This PR adds necessary handling of type alias symbol to `invoke` resolution.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/tower/FirTowerResolverSession.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/tower/FirTowerResolverSession.kt
@@ -7,17 +7,15 @@ package org.jetbrains.kotlin.fir.resolve.calls.tower
 
 import org.jetbrains.kotlin.fir.FirSession
 import org.jetbrains.kotlin.fir.asReversedFrozen
+import org.jetbrains.kotlin.fir.declarations.FirRegularClass
 import org.jetbrains.kotlin.fir.declarations.isInner
 import org.jetbrains.kotlin.fir.expressions.FirExpression
 import org.jetbrains.kotlin.fir.expressions.FirQualifiedAccessExpression
 import org.jetbrains.kotlin.fir.expressions.FirResolvedQualifier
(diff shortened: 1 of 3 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/tower/FirTowerResolverSession.kt
```diff
@@ -564,10 +563,15 @@ class FirTowerResolverSession internal constructor(
 
         for (invokeReceiverCandidate in candidateFactoriesAndCollectors.invokeReceiverCollector!!.bestCandidates()) {
             val symbol = invokeReceiverCandidate.symbol
-            if (symbol !is FirCallableSymbol<*> && symbol !is FirRegularClassSymbol) continue
+            if (symbol !is FirCallableSymbol<*> && symbol !is FirRegularClassSymbol && symbol !is FirTypeAliasSymbol) continue
 
-            val isExtensionFunctionType =
-                (symbol as? FirCallableSymbol<*>)?.fir?.returnTypeRef?.isExtensionFunctionType(components.session) == true
+            val returnTypeRef =
```
punzki, line 568: Suggest rewriting as:
```kt
val returnTypeRef = when (symbol) {
    is FirTypeAliasSymbol -> symbol.fir.expandedTypeRef
    is FirCallableSymbol<*> -> symbol.fir.returnTypeRef
    is FirRegularClassSymbol -> null
    else -> continue
}
```

jsjeon: Even `symbol` can be moved into `when`. Much simpler now, thanks!

mglukhikh:
Merged 👍

PR #3590, opened 2020-07-22
[IR][copy] prevent loose of location information for non inlinable lambdas

vvlevchenko:


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeepCopyIrTreeWithSymbols.kt
```diff
@@ -65,6 +65,14 @@ open class DeepCopyIrTreeWithSymbols(
     }
 
     protected fun mapDeclarationOrigin(origin: IrDeclarationOrigin) = origin
+    protected fun mapDeclarationOrigin(declaration: IrDeclaration):IrDeclarationOrigin {
```
homuroll, line 68: Won't it be enough to override this method in the inliner copier?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeepCopyIrTreeWithSymbols.kt
```diff
@@ -65,6 +65,14 @@ open class DeepCopyIrTreeWithSymbols(
...
+        return when {
+            declaration.origin == IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA && (declaration is IrFunction) && !declaration.isInline -> {
```
homuroll, line 70: Do really such functions exist - with origin = LOCAL_FUNCTION_FOR_LAMBDA and inline?

vvlevchenko: Hm... good question. will investigate it.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeepCopyIrTreeWithSymbols.kt
vvlevchenko: will try

vvlevchenko:
replaced with #3604

PR #3591, opened 2020-07-22
Add support for RxJava3's nullability annotations

Bencodes:
The annotations in RxJava3 moved from `io.reactivex.annotations` to `io.reactivex.rxjava3.annotations`. This pull pull request is just adding support to `JvmAnnotationNames` for the repacked annotations.

### DIFF
--- a/compiler/testData/foreignAnnotations/tests/rxjava3.kt
+++ b/compiler/testData/foreignAnnotations/tests/rxjava3.kt
@@ -1,0 +1,42 @@
+// !DIAGNOSTICS: -UNUSED_VARIABLE -UNUSED_PARAMETER
+// FILE: A.java
+
+import io.reactivex.rxjava3.annotations.*;
+
+public class A<T> {
+    @Nullable public String field = null;
+
+    @Nullable
+    public String foo(@NonNull String x, @Nullable CharSequence y) {
+        return "";
+    }
+
+    @NonNull
+    public String bar() {
+        return "";
+    }
+
+    @Nullable
+    public T baz(@NonNull T x) { return x; }
+}
+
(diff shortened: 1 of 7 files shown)


PetukhovVictor:
@Bencodes thank you for the pull request!

Unfortunately, such change is breaking change. Simple example:
```
// FILE: Test.java
import org.jetbrains.annotations.Nullable;

class Test {
    @Nullable
    static String foo() { return null; }
}

// FILE: main.kt

fun takeFoo(x: String?) {} // 1
fun takeFoo(x: String, y: Int = 10) {} // 2

fun main() {
    takeFoo(Test.foo()) // resolve to (2) before your change, to (1) after that
}
```

I suggest you create a language feature for it with target version 1.5 and write a checker which will be warn about future changing of resolution.

dzharkov:
Related https://youtrack.jetbrains.com/issue/KT-44261

petukhovv:
We've added support for rxjava3 annotations with common mechanism to specify specific annotations in CLI arguments with their report level. Unfortunately, we could do it only in 1.5.30 due to the fact that we've decided to support new nullability annotations only by that new managing mechanism.

For instance, to enable rxjava3 annotations in the strict mode you can use in 1.5.30: `-Xnullability-annotations=@io.reactivex.rxjava3.annotations:strict`. You can use this syntax for any nullability annotations. Nullability annotation settings are located [here](https://github.com/JetBrains/kotlin/blob/6cfb1f40d6881a90a6e8ac632347775a23d49b0a/core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JavaNullabilityAnnotationSettings.kt#L17).

PR #3592, opened 2020-07-22
KT-20421 fixing code generation for the case when "object" extends "class"

achmyr:
The fix will add braces so the code look like: 
```
internal object A : Base() {
    const val F1 = 0
}
```
https://youtrack.jetbrains.com/issue/KT-20421

### DIFF
--- a/nj2k/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
+++ b/nj2k/src/org/jetbrains/kotlin/nj2k/printing/JKCodeBuilder.kt
@@ -222,7 +222,7 @@ internal class JKCodeBuilder(context: NewJ2kConverterContext) {
                             ?.let { it as? JKDelegationConstructorCall }
                     if (delegationCall != null) {
                         printer.par { delegationCall.arguments.accept(this) }
-                    } else if (!superType.isInterface() && primaryConstructor != null) {
+                    } else if (!superType.isInterface() && (primaryConstructor != null || parentClass.isObjectOrCompanionObject)) {
                         printer.print("()")
                     }
                 }
--- a/nj2k/testData/newJ2k/issues/kt-20421.java
+++ b/nj2k/testData/newJ2k/issues/kt-20421.java
@@ -1,0 +1,6 @@
+class Base {
+    static final int F = 0;
+}
+final class A extends Base {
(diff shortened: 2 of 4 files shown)


darthorimar:
@achmyr  hi and thank you for your PR :)

On the following example proposed fix will add `()` for interface implementaion:

```
interface Base { // note it is interface now
    static final int F = 0;
}
final class A extends Base {
    static final int F1 = 0;
}

```
will be converted to 

```
internal interface Base {
    companion object {
        const val F = 0
    }
}

internal object A : Base() /*ERROR: This class does not have a constructor*/ {
    const val F1 = 0
}
```

darthorimar:
Looks, like the problem is missing parens around `primaryConstructor != null || parentClass.isObjectOrCompanionObject`
Will fix it and merge manually, if you don't mind :)

achmyr:
Hi @darthorimar please consider your change carefully. 

As my original understanding was that it is irrelevant if `primaryConstructor == null` or `!=null` for this new case. And it is more likely that condition `primaryConstructor!=null` would be true all the time, because of the filter applied earlier for all the classes to have default primaryConstructor [InsertDefaultPrimaryConstructorConversion.kt](https://github.com/JetBrains/kotlin/blob/b2002d56bdf27bd709d642c38bb8cce4805af3da/nj2k/src/org/jetbrains/kotlin/nj2k/conversions/InsertDefaultPrimaryConstructorConversion.kt)

This is the reason why the changes to the code made this way, not to break original logic.

darthorimar:
We still need to check if we aren't adding super call `()` for interfaces to not to have situation like [that](https://github.com/JetBrains/kotlin/pull/3592#issuecomment-674824242): 
`!superType.isInterface() && (primaryConstructor != null || parentClass.isObjectOrCompanionObject)`

Btw, primary constructor is not mandatory in Kotlin :) it is possible to have a class without it. In that case you cannot call a super constructor:
```
open class A
class B : A() /* Error: Supertype initialization is impossible without primary constructor*/ {
    constructor() {} // this is a secondary constructor
}
```
So it is better to check if we have one before delegating it :)

achmyr:
Thanks @darthorimar, I'm with you now. I have update the PR with parentheses.

darthorimar:
@achmyr thank you!

PR #3595, opened 2020-07-23
ForLoopsLowering: Assume `step` == 1 for *Range (e.g., IntRange)

punzki:
...and handle accordingly (e.g., do not read `step` property).


udalov:
Thanks!

PR #3596, opened 2020-07-23
KT-40515 Flip equals intention doesn't work for String `equals` without ignoreCase argument

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40515


dimonchik0036 (approved):
Thanks! Merged as https://github.com/JetBrains/intellij-kotlin/commit/79b35bf7e5c879c8eac6f67039a507c83dc9ec8f

PR #3599, opened 2020-07-24
FIR: initial support of suspend conversion for function reference

jsjeon:
The motivation is `adaptedReference/suspendConversion/simple.kt`:
```
fun runSuspend(c: suspend () -> Unit) {
    ...
}

var test = "failed"

fun foo() { test = "OK" }

fun box(): String {
    runSuspend(::foo)
    ...
}
```
Note that the function reference `::foo`, which is not a suspend function, is passed as an argument to the function where the (expected) type of the corresponding value parameter is a suspend function. Unlike usual function references, which normally have `KFunction` type, in this case, a reference that is passed to a suspend function parameter should be `KSuspendFunction` instead. Since 1.4, a feature named "suspend conversion" literally converts a function reference to a suspend function reference for such case.

This PR adds an initial support of suspend conversion for function reference. To do so, during resolution, we allow the resulting type of `KFunction` as a counter part for the expecting type of suspend function. Later, during conversion, we catch that and replace that reference with _adapted_ reference. See IrText test results which show how prior `FUNCTION_REFERENCE` is replaced with a call to the adapter that wraps the original function reference.

Note that `vararg`, spread, and bound receivers are not supported yet. Also, referring to feature flags is in question.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
```diff
@@ -87,12 +98,18 @@ class CallAndReferenceGenerator(
                 }
                 is IrFunctionSymbol -> {
                     val function = symbol.owner
-                    IrFunctionReferenceImpl(
-                        startOffset, endOffset, type, symbol,
-                        typeArgumentsCount = function.typeParameters.size,
-                        valueArgumentsCount = function.valueParameters.size,
-                        reflectionTarget = symbol
-                    )
+                    // TODO: should refer to LanguageVersionSettings.SuspendConversion
```
jsjeon, line 101: @mglukhikh I think both resolution and conversion should refer to feature flags, but didn't so far. Any recommended place to add it? E.g., `FirSession` or some other context?

compiler/testData/diagnostics/tests/suspendConversion/suspendConversionCompatibilityInDisabledMode.fir.kt
```diff
@@ -11,7 +11,7 @@ object Test1 {
 
         fun test() {
             val result = foo(::bar)
-            <!DEBUG_INFO_EXPRESSION_TYPE("kotlin.Int")!>result<!>
+            <!DEBUG_INFO_EXPRESSION_TYPE("kotlin.String")!>result<!>
```
jsjeon, line 14: Note that test changes under `...Disabled*` are regression, since they should not be converted. Once we add feature flag look-up, these can be reverted.

mglukhikh:
Merged manually 👍 Note: @udalov introduced `IrFactory` recently (see e.g. https://github.com/JetBrains/kotlin/commit/8c41ba8ee4e0ff76dae653dc9394869d050936ef as an example); in future please use it for creating IR declarations and block / expression bodies (IrBlockBody / IrExpressionBody), because soon IrDeclaration*Impl will be inaccessible from `fir2ir` module. In this PR, I edited the code myself.

jsjeon:
> Merged manually 👍 Note: @udalov introduced `IrFactory` recently (see e.g. [8c41ba8](https://github.com/JetBrains/kotlin/commit/8c41ba8ee4e0ff76dae653dc9394869d050936ef) as an example); in future please use it for creating IR declarations and block / expression bodies (IrBlockBody / IrExpressionBody), because soon IrDeclaration*Impl will be inaccessible from `fir2ir` module. In this PR, I edited the code myself.

Thank you for the pointer as well as editions. I noticed where you changed, and will use it.

PR #3600, opened 2020-07-25
KT-39502 Scripting: reverse order of Severity enum so that ERROR > INFO

rnett:
Pull request for [KT-39502](https://youtrack.jetbrains.com/issue/KT-39502).

Tested with tests in `kotlin.kotlin-scripting-compiler.test`, `kotlin.kotlin-scripting-ide-services-test.test`, and tests from `idea` in `org.jetbrains.kotlin.idea.script`.

### DIFF
--- a/core/script.runtime/src/kotlin/script/dependencies/resolvers_deprecated.kt
+++ b/core/script.runtime/src/kotlin/script/dependencies/resolvers_deprecated.kt
@@ -26,7 +26,7 @@ typealias Environment = Map<String, Any?>
 
 interface ScriptDependenciesResolver {
 
-    enum class ReportSeverity { FATAL, ERROR, WARNING, INFO, DEBUG }
+    enum class ReportSeverity { DEBUG, INFO, WARNING, ERROR, FATAL }
 
     fun resolve(script: ScriptContents,
                 environment: Environment?,
--- a/core/script.runtime/src/kotlin/script/experimental/dependencies/resolvers.kt
+++ b/core/script.runtime/src/kotlin/script/experimental/dependencies/resolvers.kt
@@ -49,7 +49,7 @@ interface DependenciesResolver : ScriptDependenciesResolver {
 
 data class ScriptReport(val message: String, val severity: Severity = Severity.ERROR, val position: Position? = null) {
     data class Position(val startLine: Int, val startColumn: Int, val endLine: Int? = null, val endColumn: Int? = null)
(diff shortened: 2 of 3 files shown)


core/script.runtime/src/kotlin/script/dependencies/resolvers_deprecated.kt
```diff
@@ -26,7 +26,7 @@ typealias Environment = Map<String, Any?>
 
 interface ScriptDependenciesResolver {
 
-    enum class ReportSeverity { FATAL, ERROR, WARNING, INFO, DEBUG }
+    enum class ReportSeverity { DEBUG, INFO, WARNING, ERROR, FATAL }
```
ligee, line 29: This change is outside the initial scope of the #KT-39502, and making this change will require much wider testing with 3-party projects since it is used in obscured places. On the other hand, this API is in fact obsolete and will be replaced with the  `kotlin.script.experimental.api.ScriptDiagnostic.Severity` soon.
@rnett unless you need this obsolete API to be changed too, I'd drop this one. Otherwise please provide some more details.

core/script.runtime/src/kotlin/script/experimental/dependencies/resolvers.kt
```diff
@@ -49,7 +49,7 @@ interface DependenciesResolver : ScriptDependenciesResolver {
 
 data class ScriptReport(val message: String, val severity: Severity = Severity.ERROR, val position: Position? = null) {
     data class Position(val startLine: Int, val startColumn: Int, val endLine: Int? = null, val endColumn: Int? = null)
-    enum class Severity { FATAL, ERROR, WARNING, INFO, DEBUG }
+    enum class Severity { DEBUG, INFO, WARNING, ERROR, FATAL }
```
ligee, line 52: The same comment as above, I'd drop this change unless absolutely necessary.

core/script.runtime/src/kotlin/script/dependencies/resolvers_deprecated.kt
rnett: That seems fine, especially as its obsolete.

core/script.runtime/src/kotlin/script/experimental/dependencies/resolvers.kt
rnett: Same as above, I see no reason not to drop it.

ligee:
Merged manually.
@rnett thank you for the contribution!

PR #3601, opened 2020-07-25
KT-22420 Intention to replace a = b with b.also { a = it }

devbridie:
https://youtrack.jetbrains.com/issue/KT-22420

This commit adds `ConvertVariableAssignmentToExpressionIntention`, which converts `a = b` to `b.also { a = it }`.


vladimirdolzhenko:
LGTM

Merged manually

PR #3602, opened 2020-07-26
KT-24797 "Remove explicit type specification": suggest on explicit type argument from property initializer

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-24797


vladimirdolzhenko:
Thanks, merged as https://github.com/JetBrains/intellij-kotlin/commit/7b62558f7a8d8d7393a00b1d5acd9e2bb71f7975

PR #3604, opened 2020-07-27
[IR][copy] prevent loose of location information for non inlinable lambdas

vvlevchenko:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/inline/DeepCopyIrTreeWithDescriptors.kt
```diff
@@ -9,8 +9,7 @@ import org.jetbrains.kotlin.backend.common.CommonBackendContext
 import org.jetbrains.kotlin.backend.common.DescriptorsToIrRemapper
 import org.jetbrains.kotlin.backend.common.WrappedDescriptorPatcher
 import org.jetbrains.kotlin.ir.IrElement
-import org.jetbrains.kotlin.ir.declarations.IrDeclarationParent
-import org.jetbrains.kotlin.ir.declarations.IrTypeParametersContainer
+import org.jetbrains.kotlin.ir.declarations.*
```
alexluz321, line 12: Not a review, just trying to understand. Why import everything instead of only what is needed? Or is everything needed?

vvlevchenko: Just an Idea's style of imports management.

PR #3605, opened 2020-07-28
FIR2IR: don't declare dispatch receiver for local functions

jsjeon:
because "local functions must not have dispatch receiver": https://github.com/JetBrains/kotlin/blob/master/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LocalDeclarationsLowering.kt#L589


mglukhikh:
Merged manually 👍

PR #3607, opened 2020-07-29
FIR2IR: convert annotations on FirTypeRef

jsjeon:
The motivation is `JvmName.classMembers`, where a receiver to the call with `with` is incorrectly type-casted due to the missing type annotation, `@ExtensionFunctionType`, on the following argument block. Since such annotations are stored in `FirTypeRef`, we can easily extend the type converter to convert that via, again, unified annotation generator.


mglukhikh:
Merged, thank you!

PR #3608, opened 2020-07-29
FIR2IR: make local storage track scopes, including anonymous init

jsjeon:
because, e.g., local function could be defined in an anonymous initializer, which is not a callable. In that case, `cacheStack` in the local storage is empty, and thus no callable cache to put parameter, variable, and local function.

------

An alternative is, whenever we want to put those into a callable cache, we can make sure there is indeed a callable in the conversion scope. This sounds over engineered, time-consuming task without visible effects. :\


mglukhikh:
I'd suggest another alternative: we should be able to create a callable cache for anonymous initializer too (of course, it's better to rename callable cache to something else). With your solution, we will have problems with retrieving local variables from anonymous initializer scope.

jsjeon:
> I'd suggest another alternative: we should be able to create a callable cache for anonymous initializer too (of course, it's better to rename callable cache to something else).

I thought that too, but wasn't sure about extending/renaming callable cache. Thank you for the suggestion.

mglukhikh:
Merged, thank you!

PR #3609, opened 2020-07-30
FIR: distinguish anonymous object as enum entry when scoping

jsjeon:
so that `this` reference with the enum entry name as label can be resolved.

The motivation is `enum.innerClassMethodInEnumEntryClass2`:
```
enum class A {
    X {
        val x = ...
        inner class Inner {
            fun foo() = this@X.x
        }
        ...
    };
    ...
}
```
where `this@X` was not resolved because, in the implicit receiver stack, there was nothing associated with the label `X`. Declaration resolver definitely visited the enum entry, but it is an anonymous object, and for all anonymous objects, labels were ignored. However, an anonymous object as enum entry has the name for label: enum entry name itself! That can be retrieved from the primary constructor.


mglukhikh:
Merged, thank you!

PR #3610, opened 2020-07-30
Fix NPE in KaptJavaLog

ZacSweers:
https://youtrack.jetbrains.com/issue/KT-44254

See https://issuetracker.google.com/issues/162446295 for more context, but in short: this will fail if the project:
* targets java 8
* consumes external libraries targeting a higher version (java 9+). This includes Android SDK 30's android.jar, which targets java 9
* has `mapDiagnosticLocations` enabled for kapt

targetElement is a nullable type, so this seems like a pretty cut-and-dry NPE fix

CC @gavra0 @yanex


ZacSweers:
@yanex should I close this?

ZacSweers:
It's quite frustrating trying to contribute at times and some sort of ack would go a long way

udalov:
Merged in https://github.com/JetBrains/kotlin/commit/12a6352bef01d5e65bc452a20f91d4efb77a199a

PR #3611, opened 2020-07-31
JVM IR: Optimize delegated properties (KT-36814)

sfs:
We want to remove unused, cached KProperties. This can happen with inline `get/setValue` functions on property delegates. Since inlining happens at the bytecode level, this optimization is implemented as a bytecode transformation.

I *tried* other ways of implementing this, without first generating code only to throw it away and there doesn't appear to be a good option in the IR backend. For example, hooking into the inliner to inspect the bytecode as is done in the JVM backend is possible, but this still has to happen during code generation and ends up being more code. Before code generation we won't necessarily know the correct method signatures to look up (since the IR method signature mapper offloads much of this work to lowerings), which precludes an implementation as a lowering. Finally, lowering class initializers right before code generation does not work, since there are usually uses of cached KProperties in the class initializer and we don't want to generate the same method twice...

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
@@ -116,12 +116,18 @@ abstract class ClassCodegen protected constructor(
         if (generated) return reifiedTypeParametersUsages
         generated = true
 
+        // We remove unused cached KProperties.
+        val classDelegatedPropertiesArray = irClass.fields.singleOrNull {
+            it.origin == JvmLoweredDeclarationOrigin.GENERATED_PROPERTY_REFERENCE
+        }
+        val delegatedPropertyTracker = if (classDelegatedPropertiesArray != null) DelegatedPropertyOptimizer() else null
+
         val smap = context.getSourceMapper(irClass)
         for (declaration in irClass.declarations) {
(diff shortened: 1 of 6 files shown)


dnpetrov:
LGTM, I'll run tests and push manually.

dnpetrov:
Pushed manually ( b15e32936e22f9510c9a21ebe6f194c40cdd9620 )

udalov:
Looks like this causes the following issue: https://youtrack.jetbrains.com/issue/KT-40815.

I've found this by running full bootstrap with JVM IR (by publishing and using the compiler which has `useIR = true` by default). Delegated properties are used in this way in `buildSrc` already, e.g. in `KotlinBuildPublishingPlugin.kt:56`.

@sfs Could you please take a look?

sfs:
Thanks for the reduced test case. I'll have a look at the bytecode to see how it differs and fix the optimization.

udalov:
Another issue is that (it seems) the local variable table is not adapted, leading to stray local variables in debugger and errors from D8. The latter can be reproduced for example by running the test `box/delegatedProperty/optimizedDelegatedProperties/definedInSources.kt`:
```
Stripped invalid locals information from 1 method.
In DefinedInSourcesKt.class:
  Methods with invalid locals information:
  java.lang.String DefinedInSourcesKt.box()
  Attempt to define local of type int as kProperty$iv:java.lang.Object
Some warnings are typically a sign of using an outdated Java toolchain. To fix, recompile the source with an updated toolchain.
```

PR #3613, opened 2020-08-01
KT-40558 False positive "Move to class body" intention on data class constructor property

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40558


igoriakovlev:
Thank you for your work!

PR #3614, opened 2020-08-01
FIR: support adapted callable reference with vararg

jsjeon:
When a function has `vararg` parameter, its reference can be adapted, depending on the expected type from the call site. E.g., from bb test `adaptedReferences/innerConstructorWithVararg.kt`:
```
class Outer(val o: String) {
    inner class Inner1(val i: Int, vararg v: String) {...}
    ...
}

fun <T> use0(f: (Int) -> T) = f(11)
fun <T> use1(f: (Int, String) -> T) = f(12, "B")

fun box(): String {
    val outer = Outer("O")
    val r1 = use0(outer::Inner1).result
    ...
    val r2 = use1(outer::Inner1).result
    ...
```
For `r1` that calls `use0` (which means using zero argument), `vararg` parameter `v` is not used. By assuming nothing is passed, `::Inner1` is still applicable. (That is supported after https://github.com/JetBrains/kotlin/pull/3581.) For `r2` that calls `use1`, `vararg` parameter `v` is supposed to pass a single element, and the current logic can't infer the correct, adapted reference type for that.

Interestingly, there could be, say, `use2` that would look like: `fun <T> use2(f: (Int, String, String) -> T) = ...`. `::Inner` is even applicable in this case, since we can retrieve two elements from `vararg` parameter `v`. In addition, there could be, say, `useAll` that would look like: `fun <T> useAll(f: (Int, Array<String>) -> T) = ...`. In that case, `::Inner1` is still applicable, since `vararg` parameter `v` can be passed as-is.

In sum, `vararg` parameter can be used in a function reference in three (or four) ways: 1) nothing passed, i.e., ignorable; 2) an element or a series of elements (with the same type, of course) is passed; or 3) an array as a whole is used.

To handle all cases, when inferring the adapted function reference type, we need expected type parameters exactly, not just a number of expected parameters. Also, we need to check the corresponding expected type, and then adapt `vararg` parameter to such usage.

Note that, in this PR, only the adaptation on reference type is done. For some other cases, we will need a different conversion for such adapted callable reference, which will reuse https://github.com/JetBrains/kotlin/pull/3599.


mglukhikh:
Looks good IMO. I will fix failed test and merge it a bit later.

mglukhikh:
Merged manually 👍

PR #3615, opened 2020-08-03
KT-40660 "Convert to anonymous object" intention does not work with anonymous lambda parameters

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40660


vladimirdolzhenko:
Thanks for the contribution

Merged as https://github.com/JetBrains/intellij-kotlin/commit/9ce0cbaffce57217aa5f80ac8217c1e99587b88a

PR #3616, opened 2020-08-03
FIR: prepare constructor for Java annotation classes

jsjeon:
The motivation is bb test `typeAnnotations/typeUseAnnotation`, which checks if specific annotations on parameter type and return type are indeed encoded in the generated JVM bytecode. Due to https://youtrack.jetbrains.com/issue/KT-30066, we thought such tests are failing, since we know `ConeKotlinType` doesn't have annotations on it and we know we don't (or can't) convert annotations on types.

However, from https://github.com/JetBrains/kotlin/pull/3607, we found annotations are stored in `FirTypeRef`, a reference wrapper that encompasses `ConeKotlineType`. So, by changing _when_ to serialize types, we can still retrieve annotations on types.

------

It turned out that those tests that check annotations on types actually examine both `javaMethod` and Kotlin's own reflective inspection. For annotations inside `javaMethod`, we still need conversions at IR element level, which has been done at https://github.com/JetBrains/kotlin/pull/3607. That means, some of annotation calls were not converted well, and the main reason was that, for Java annotation classes, no constructors were prepared.


mglukhikh:
Annotation constructors were implemented by @demiurg906 (and in general case, not only for Java) , but this branch isn't yet in master (probably I will merge it this week). I think I will add the remainder of this PR to Dmitry' branch. Thank you.

mglukhikh:
Partially merged (only the commit about the serializer). The second commit was discarded because this feature is implemented in much more general case in branch of Dmitry Novozhilov (also merged today).

jsjeon:
> Partially merged (only the commit about the serializer). The second commit was discarded because this feature is implemented in much more general case in branch of Dmitry Novozhilov (also merged today).

I found https://github.com/JetBrains/kotlin/commit/bc1fa8ed7f09c45e3fc98ffa85c8a50db62a3658 is what you meant about general case for Java annotation constructor. Also found how tests that were unmuted in this PR are indeed unmuted: https://github.com/JetBrains/kotlin/commit/8379b3794a902b44ede7eda1ea3cee438701e033

mglukhikh:
There are some more details regarding annotation resolve and some more unmuted tests in commits from 4 Aug: https://github.com/JetBrains/kotlin/commits?author=demiurg906

PR #3618, opened 2020-08-03
[FIR] Fixes for RedundantReturnUnitType checker

vldF:
Fixed misbehaver in cases with property in data class constructors and in cases with annotated unit return type


mglukhikh:
Merged 👍

PR #3623, opened 2020-08-04
FIR2IR: use type arguments from fully expanded type

jsjeon:
because, if the type is a nested type alias (e.g., one of type argument is another type alias), converting the type argument at the current level may lead to mismatched number of type parameters.


mglukhikh:
Merged, thank you!

PR #3624, opened 2020-08-04
KT-40546 False positive "Convert lambda to reference" with multiple `it` references

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40546


nikitabobko:
Merged to new repo
https://github.com/JetBrains/intellij-kotlin/commit/6b1bfd52b34d74de41b2edde57e5c2229395d4cc

PR #3625, opened 2020-08-05
FIR2IR: apply adapted reference conversion to coercion-to-unit

jsjeon:
The main purpose of the 1st commit was to extend https://github.com/JetBrains/kotlin/pull/3599 (initial support of suspend conversion for function reference) to coercion-to-unit case: when computing functional type for the function reference whose target function doesn't return `Unit` while the use site requires `Unit` return type, so-called coercion-to-unit adaptation happened: https://github.com/JetBrains/kotlin/commit/e1e149a0621645ba40da86a17aa7920f0e00138d. In addition to reference type adaptation, we need to wrap it with adapter function and replace the original reference with an adapted reference, similar to suspend conversion: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/ReflectionReferencesGenerator.kt#L84

While doing so, a couple regressions were found, if coercion-to-unit reference has bound receivers. Thus, the 1st commit also extends the initial support of suspend conversion to handle bound receivers. Still, a transient receiver, like a class symbol (e.g., `A::foo`), is not handled yet, and marked as TODO (and one regression). A couple of minor fixes, such correction of `suspend` and checking `Unit` return type, are added too.

The 2nd commit is a simple refactoring to share `IrExpression.isSafeToUseWithoutCopying`.

As mentioned at https://github.com/JetBrains/kotlin/pull/3614, the remaining work on adapted references and conversion is to convert adapted reference with `vararg`.


compiler/testData/codegen/box/callableReference/serializability/adaptedReferences.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND_FIR: JVM_IR
```
mglukhikh, line 1: Do you know why this test became broken?

jsjeon: When adding an adaptee call inside an adapter function, transient receiver (`A` in `A::foo`) is not handled well, and I left a TODO.

For example, the last one `allOfTheAbove(A::foo)`, the expected IR dump looks like:
```
CALL 'public final fun check (x: kotlin.Any): kotlin.Unit declared in <root>.AdaptedReferencesKt' type=kotlin.Unit origin=null
          x: CALL 'public final fun allOfTheAbove (f: kotlin.Function1<<root>.A, kotlin.Unit>): kotlin.Any declared in <root>.AdaptedReferencesKt' type=kotlin.Any origin=null
            f: BLOCK type=kotlin.Function1<<root>.A, kotlin.Unit> origin=ADAPTED_FUNCTION_REFERENCE
              FUN ADAPTER_FOR_CALLABLE_REFERENCE name:foo visibility:local modality:FINAL <> (p0:<root>.A) returnType:kotlin.Unit
                VALUE_PARAMETER ADAPTER_PARAMETER_FOR_CALLABLE_REFERENCE name:p0 index:0 type:<root>.A
                BLOCK_BODY
                  TYPE_OP type=kotlin.Unit origin=IMPLICIT_COERCION_TO_UNIT typeOperand=kotlin.Unit
                    CALL 'public final fun foo (s: kotlin.String, vararg xs: kotlin.Long): kotlin.String declared in <root>.A' type=kotlin.String origin=null
                      $this: GET_VAR 'p0: <root>.A declared in <root>.AdaptedReferencesKt.box.foo' type=<root>.A origin=null
              FUNCTION_REFERENCE 'local final fun foo (p0: <root>.A): kotlin.Unit declared in <root>.AdaptedReferencesKt.box' type=kotlin.Function1<<root>.A, kotlin.Unit> origin=ADAPTED_FUNCTION_REFERENCE reflectionTarget=null
```
v.s. `fir` result with this PR looks like:
```
CALL 'public final fun check (x: kotlin.Any): kotlin.Unit declared in <root>.AdaptedReferencesKt' type=kotlin.Unit origin=null
          x: CALL 'public final fun allOfTheAbove (f: kotlin.Function1<<root>.A, kotlin.Unit>): kotlin.Any declared in <root>.AdaptedReferencesKt' type=kotlin.Any origin=null
            f: BLOCK type=kotlin.Function1<<root>.A, kotlin.Unit> origin=ADAPTED_FUNCTION_REFERENCE
              FUN ADAPTER_FOR_CALLABLE_REFERENCE name:foo visibility:local modality:FINAL <> (p0:<root>.A) returnType:kotlin.Unit
                VALUE_PARAMETER ADAPTER_PARAMETER_FOR_CALLABLE_REFERENCE name:p0 index:0 type:<root>.A
                BLOCK_BODY
                  CALL 'public final fun foo (s: kotlin.String, vararg xs: kotlin.Long): kotlin.String declared in <root>.A' type=kotlin.String origin=null
              FUNCTION_REFERENCE 'local final fun foo (p0: <root>.A): kotlin.Unit declared in <root>.AdaptedReferencesKt.box' type=kotlin.Function1<<root>.A, kotlin.Unit> origin=ADAPTED_FUNCTION_REFERENCE reflectionTarget=null
```
The missing part is the bound receiver. I found that a class symbol `A` is set as an explicit receiver, but not converted to IR receiver. Then, we give up setting the receiver for the adaptee call.

I know, a real question is, why it worked when we _don't_ apply this adapted reference conversion to coercion-to-unit case. That's still mysterious, and I don't fully understand how reference lowering in the backend deals with adapted reference... On the other hand, another real question is, when we really apply reference conversion to coercion-to-unit? I don't know either... :( `ReflectionReferenceGenerator` in `psi2ir` doesn't have good answers either.

mglukhikh:
Merged 👍

jsjeon:
Thanks, I'll keep working on that regression as well as `vararg` supports.

compiler/testData/codegen/box/callableReference/serializability/adaptedReferences.kt
jsjeon: Will be resolved at https://github.com/JetBrains/kotlin/pull/3630

PR #3626, opened 2020-08-05
RangeContainsLowering: Optimize calls to contains() on ClosedRanges

punzki:
Notable TODOs in this initial version:
1. Unsigned ranges
2. Comparable ranges
3. `contains()` on non-literal floating point ranges
4. Refactoring/renaming code common to ranges and loops

CC @sfs


udalov:
Looks like this causes the following failure when compiling `:compiler:compileTestKotlin` after bootstrap with JVM IR:
```
e: java.lang.IllegalStateException: Validation failed
    at org.jetbrains.kotlin.backend.common.IrValidator.error(IrValidator.kt:79)
    at org.jetbrains.kotlin.backend.common.IrValidator.access$error(IrValidator.kt:60)
    at org.jetbrains.kotlin.backend.common.IrValidator$elementChecker$1.invoke(IrValidator.kt:83)
    at org.jetbrains.kotlin.backend.common.IrValidator$elementChecker$1.invoke(IrValidator.kt:60)
    at org.jetbrains.kotlin.backend.common.CheckIrElementVisitor.visitElement(CheckIrElementVisitor.kt:48)
    at org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid$DefaultImpls.visitExpression(IrElementVisitorVoid.kt:104)
    at org.jetbrains.kotlin.backend.common.CheckIrElementVisitor.visitExpression(CheckIrElementVisitor.kt:358)
    at org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid$DefaultImpls.visitMemberAccess(IrElementVisitorVoid.kt:158)
    at org.jetbrains.kotlin.backend.common.CheckIrElementVisitor.visitMemberAccess(CheckIrElementVisitor.kt:36)
...
```

Most likely the lowering produces the IR where the same expression is attached to several nodes, which is one of the things the validator checks against. But investigation is needed.

See the pinned post on the internal Slack channel for the instructions how to perform the bootstrap with JVM IR. Feel free to reach out for any help!

punzki:
@udalov Good catch! Rebased, and fixed in latest commit, with added tests.

compiler/testData/codegen/bytecodeText/ranges/inMixedUnsignedRange.kt
```diff
@@ -50,3 +48,14 @@ fun n_ul_ul(x: ULong, a: ULong, b: ULong) = x !in a..b
 // 22 SIPUSH 255
 // 24 LDC 65535
 // 2 LDC 4294967295
+
+// JVM_IR_TEMPLATES
```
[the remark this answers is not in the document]
punzki: Do you mean just to split this into 2 files? How would that be different from using `JVM(_IR)_TEMPLATES`?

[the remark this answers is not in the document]
udalov: Sorry for being unclear. Basically this test checks that `in` in ranges is optimized. However, there are six commented lines where it's not optimized in the old backend. I suggest to extract those six lines into a separate test, with the correponding `JVM_TEMPLATES/JVM_IR_TEMPLATES` which would be stricter. And then only leaving something like `// 0 contains` in this test, which is also going to be a bit stricter, and more to the point of what the test is actually checking.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/RangeContainsLowering.kt
```diff
@@ -0,0 +1,444 @@
...
+        //     $this: CALL 'public open fun contains (value: kotlin.Int): kotlin.Boolean [operator] declared in kotlin.ranges.IntRange' type=kotlin.Boolean origin=NOT_IN
+        //
+        // We only want to lower the call to contains(); even in the `!in` case, the call to not() should be preserved).
+        val origin = expression.origin
+        if (origin == null || (origin != IrStatementOrigin.IN && origin != IrStatementOrigin.NOT_IN)) {
+            return super.visitCall(expression)  // The call is not an `in` expression.
+        }
+        if (origin == IrStatementOrigin.NOT_IN && expression.symbol == context.irBuiltIns.booleanNotSymbol) {
+            return super.visitCall(expression)
+        }
```
udalov, line 84: I don't quite understand why can't these be replaced with just
```
        if (origin != IrStatementOrigin.IN) return super.visitCall(expression)
```
?

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/RangeContainsLowering.kt
```diff
@@ -0,0 +1,444 @@
...
+
+        if (expression.extensionReceiver != null && !stdlibExtensionContainsCallMatcher(expression)) {
```
udalov, line 86: Perhaps this would convey the meaning more clearly?
```suggestion
        if (expression.extensionReceiver == null || !stdlibExtensionContainsCallMatcher(expression)) {
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/RangeContainsLowering.kt
```diff
@@ -0,0 +1,444 @@
...
+        val argument = expression.getValueArgument(0)!!
+
+        val builder = context.createIrBuilder(getScopeOwnerSymbol(), expression.startOffset, expression.endOffset)
+
+        return builder.run {
+            var replacement = buildContainsComparison(headerInfo, argument, origin)
+                ?: return super.visitCall(expression)  // The call cannot be lowered.
+            if (argument.type.isNullable()) {
+                // There are stdlib extension functions that return false for null arguments, e.g., IntRange.contains(Int?).
+                replacement = irIfNull(context.irBuiltIns.booleanType, argument.deepCopyWithSymbols(), irFalse(), replacement)
```
udalov, line 108: Is it possible to avoid deep copy? For some reason it fails on a complex expression here:
```
fun foo(x: Int?, a: Int, b: Int): Boolean =
    { x }() in a..b

fun main() {
    foo(null, 0, 5)
}
```
```
java.lang.NullPointerException: declarationParentsStack.peekFirst() must not be null
	at org.jetbrains.kotlin.ir.util.PatchDeclarationParentsVisitor.patchParent(PatchDeclarationParents.kt:57)
	...
	at org.jetbrains.kotlin.backend.common.lower.Transformer.visitCall(RangeContainsLowering.kt:452)
```

BTW I was actually going to construct a test case related to https://youtrack.jetbrains.com/issue/KT-40467, which is the reason we currently mostly avoid `deepCopyWithSymbols` in JVM IR...

udalov: Oh, looks like it also breaks semantics here:
```
var state = ""
fun sideEffect(): Int? {
    state += "OK"
    return 1
}

fun box(): String {
    sideEffect() in 1..2
    return state
}
```
returns "OKOK"

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/RangeContainsLowering.kt
punzki: Currently in PSI2IR, for `!in`, _both_ the `not()` and the `contains()` calls have `origin=NOT_IN`. In this case, we only want to transform the `contains()` call, and not the `not()` call.

Sorry, it's not very obvious in the IR dump in the preceding comment. I'll modify the comment and explicitly call out the `NOT_IN` origins in the `!in` case. I also noticed just now that the `origin == null` clause is superfluous, so I'll remove that too.

compiler/testData/codegen/bytecodeText/ranges/inMixedUnsignedRange.kt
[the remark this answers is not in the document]
punzki: Ok, done. Let me know if I got what you meant. 🙂

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/RangeContainsLowering.kt
punzki: I believe I fixed this now in latest version of PR (last commit) and added tests. Please take a look!

udalov: Could you clarify why are we optimizing the case of nullable subject? Is it a pattern worth optimizing and were there any requests to optimize it?

I'm worried that the code of the lowering became much more involved, and the motivation is not clear to me. Maybe we could avoid optimizing these kinds of expressions, and force the user to refactor the code in a way that would make the subject non-nullable (`x != null && x in range`).

punzki: I'm very glad you asked this question! There weren't any requests to do optimize this, and it is also not optimized on the old JVM backend. During development, one of the initial failures I saw was `nullableInPrimitiveRange.kt`, because I wasn't handling nullable subjects. I considered at first to just let it remain unoptimized and let the stdlib extension functions take care of it (e.g., `IntRange.contains(Int?)`), but I thought it was simple enough to handle, hence the initial version which was indeed very simple.

But after the problems you've helped identify, and the corresponding solution, it has become much more complicated to handle. It's great you asked the question since I was also getting a little wary of the complexity. I'll just punt on handling it for now.

udalov:
Thank you for the very detailed explanation!

I think we do need the generated bytecode to be optimized as much as possible in the most popular cases, and something like `for (i in 0 until n)` is a very popular case. When I say "optimized" I mean not only the CPU clocks needed to execute it, but also the bytecode size since it directly influences HotSpot inlining decisions. My main concern is that subtracting 1 and checking for MIN_VALUE takes significantly more bytecode than the exclusive bounds check. We have a lot of inline functions with such `for` loops in the stdlib and I suspect it might be a bottleneck in some cases. Same for range `in` check, although I suspect it's much less critical.

I've reported this request to YouTrack here: https://youtrack.jetbrains.com/issue/KT-41352.

That said, it's a further optimization and right now we explicitly avoid working on optimizations and are focusing on finishing up the parts of JVM IR needed for enabling it as default in the next release. So it's not critical at all right now.

As of `step` on `until` ranges, I'd prefer the option with less work involved (the first). After all, `until`+`step` is far less popular than just `until` (`git grep '\<for\>.*\<until\>.*\<step\>'` returns just one non-test usage in the project).

PR #3629, opened 2020-08-05
JVM IR: Fix delegated property optimization (KT-40815)

sfs:
This PR fixes the two issues pointed out by @udalov in #3611 
- We need to track accesses to cached delegated properties from nested classes. We can exclude inline functions here, since these will not use cached KProperties.
- We have to keep stores to local variables which are introduced for inline function parameters. The JVM backend uses `null` in place of an unused KProperty here, so that's what we do in the JVM IR backend as well. I'm not sure how to test this behavior though. Maybe a stepping test?


udalov:
Thanks for the investigation, but looks like it didn't fully help. The full bootstrap with JVM IR is failing with the same error while configuring `:kotlin-allopen`. I might've oversimplified the repro, here's the one that still fails with `java.lang.NoSuchFieldError: $$delegatedProperties`:
```
import kotlin.reflect.KProperty

inline operator fun String.getValue(thiz: Any?, property: KProperty<*>): String = this

operator fun Int.provideDelegate(thiz: Any?, property: KProperty<*>): String = property.name

fun box(): String = with(42) {
    {
        val OK by this
        OK
    }()
}
```

sfs:
Thanks for the test case. I'll try to run the bootstrap with JVM IR myself next time.

The problem is that this assumption:

> We can exclude inline functions here, since these will not use cached KProperties.

is incorrect, since we need to handle lambda arguments passed to inline functions which *can* contain references to cached KProperties.

sfs:
I fixed this problem, removed the error prone default arguments, and checked that the compiler can be used to bootstrap kotlinc.

udalov:
Thanks for the quick resolution!

udalov:
Regarding testing local variables, maybe we could add a test case to `compiler/testData/debug/localVariables`? Or otherwise just a bytecode text test

PR #3630, opened 2020-08-05
FIR2IR: handle unbound reference with adapted arguments

jsjeon:
This is a follow-up of https://github.com/JetBrains/kotlin/pull/3625, where I said:
> Still, a transient receiver, like a class symbol (e.g., A::foo), is not handled yet, and marked as TODO (and one regression).

If the explicit receiver of the reference access is a class symbol, the counter part of receiver won't be a form of IR element. Rather, we can treat it as a transient receiver, and depending on dispatch/extension receiver requirement of the adaptee function, we can set either dispatch or extension receiver for the adaptee call properly.


mglukhikh:
Merged manually 👍

PR #3632, opened 2020-08-06
JVM IR: Don't produce CHECKCASTs on null constants (KT-36650)

sfs:


compiler/testData/codegen/bytecodeText/coroutines/cleanup/nullCleanup.kt
```diff
@@ -12,6 +12,12 @@ suspend fun test() {
...
+// 2 PUTFIELD .*L\$0 : Ljava/lang/Object;
+
+// JVM_TEMPLATES:
+// just before suspension point
 // 1 ACONST_NULL
-// 2 PUTFIELD .*L\$0 : Ljava/lang/Object;
\ No newline at end of file
+
+// JVM_IR_TEMPLATES:
+// two stores to initialize the `a` variable and one null constant to store in the spill slot.
```
sfs, line 22: @ilmirus Could you please take a look at this test and tell me if this looks correct to you? The bytecode produced by the JVM and JVM IR backend only differs by two (probably unnecessary) `ACONST_NULL; ASTORE 1` instruction sequences on the JVM IR backend.

compiler/testData/codegen/box/classes/kt1018.kt
```diff
@@ -1,3 +1,6 @@
+// IGNORE_BACKEND_FIR: JVM_IR
```
max-kammerer, line 1: @demiurg906 Hope you aware about this problem

compiler/testData/codegen/bytecodeText/coroutines/cleanup/nullCleanup.kt
ilmirus: @sfs Thanks for pointing this out! There is indeed a bug, but in Old JVM BE: https://youtrack.jetbrains.com/issue/KT-40886. Bytecode, generated by JVM_IR BE is completely fine.

udalov:
Thanks!

PR #3636, opened 2020-08-07
JVM IR: Fixes to IMPLICIT_NOTNULL casts

sfs:
There are two parts to this PR.

The first is a fix for KT-39520, which shows that we have to be very careful when compiling casts, as the type information might be outright wrong and superfluous `CHECKCAST` instructions can lead to runtime errors. The second is a fix for another Java interop issue around `IMPLICIT_NOTNULL` casts on special bridge methods. This issue affects the Kotlin compiler itself and leads to test failures when running the compiler test suite with a version of kotlinc that was compiled with the JVM IR backend.

I've added tests for all cases that are affected.

As a side note, the behavior around `IMPLICIT_NOTNULL` casts is a bit inconsistent. We elide the checks for all special bridge methods except for those which are `@PlatformDependent` (i.e. `MutableMap.remove` with two arguments and `Map.getOrDefault`). We probably want to change this at some point, but for now the behavior of the JVM and JVM IR backends should match.

### DIFF
--- a/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
+++ b/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
@@ -8,6 +8,7 @@ package org.jetbrains.kotlin.backend.common.lower
 import org.jetbrains.kotlin.backend.common.CommonBackendContext
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
+import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
 import org.jetbrains.kotlin.ir.expressions.IrConstKind
 import org.jetbrains.kotlin.ir.expressions.IrExpression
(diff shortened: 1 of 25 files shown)


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SpecialBridgeMethods.kt
```diff
@@ -103,11 +104,16 @@ class SpecialBridgeMethods(val context: CommonBackendContext) {
...
-        irFunction.allOverridden().forEach { overridden ->
-            val description = overridden.toDescription()
-            SPECIAL_METHODS_WITH_DEFAULTS_MAP[description]?.let {
-                return Pair(overridden, it)
+    fun findSpecialWithOverride(
+        irFunction: IrSimpleFunction,
+        includeSelf: Boolean = false
+    ): Pair<IrSimpleFunction, SpecialMethodWithDefaultInfo>? {
+        irFunction.allOverridden(includeSelf).forEach { overridden ->
+            if (overridden.parent is IrClass) {
```
udalov, line 112: How is it possible to have an override with a parent not inside an `IrClass`?

sfs: It isn't, but if `includeSelf` is `true` we're also looking at the `irFunction` which might be an external top-level declaration or intrinsic.

I've hoisted the check outside of the `forEach` to make a code a bit cleaner.

udalov:
Merged manually in https://github.com/JetBrains/kotlin/commit/0328fcaf5d7f3e3aef39ddd549f09d68914ed455.

Thanks!

PR #3639, opened 2020-08-07
FIR: smartcast after null check

jsjeon:
_Disclaimer_: this is very subtle, so would be my description. Thus, I'm open to long discussion, reviews, other suggestions, etc.

------

The motivation is bb test `nullSmartCast`:
```kt
fun String?.foo() = this ?: "OK"

fun foo(i: Int?): String {
    if (i == null) return i.foo() // Unresolvable
    return "$i"
}

fun box() = foo(null)
```
As expected, candidate `String?.foo()` expects `String?` as an extension receiver type, whereas the actual receiver, `i`, is of `Int?` type. Due to the `null` check, `i`'s type would be technically `Nothing?` after null smartcast, hence the candidate should be applicable.

First, I found that, in the data-flow analysis, while processing `== null`, adding `Nothing?` cast (in an inverted way, which was wrong, btw) has been commented out. With proper type implication after null equality check, the target bb test became green.

------

However, there were a couple regressions which looked like this:
```kt
if (result == NON_NULL_CONST) fail(...)
return result as String
```
At the last return statement, the type of `result` became `Nothing?` that I just added, which caused the runtime casting error. It turned out that, in the data-flow analysis, while processing a general equality, like `left == right`, the logic attempted to add smarter implications: if `left == right` and `right != null`, then `left != null`. Yes, that's right, but it relied on the aforementioned logic of processing `== null` to add the resulting implication.

The problem of reusing `== null` is, since `fir` isn't SSA form, all information inside data-flow analysis is sort of _guided_ by the entering condition. Therefore, the way `== null` adds implications is enumerating all cases: when the condition holds, the involving variable is null, of `Nothing?` type, etc., v.s. otherwise, it is not null, should not be of `Nothing?` type, etc. And there, unwanted condition and effect is added.

So, while processing `left == right`, if we know `right != null`, we can infer `left != null`. But, if we simply reuse `== null` to insert `left != null` against the condition `left == right`, it actually added: `left == right` -> `left != null` _and_ `left != right` -> `left == null` and the latter suddenly made `result` be of `Nothing?` type! Please note that, if `right != null` still holds and if `left != right`, it simply means `left` is something different from `right`, including `null`. It doesn't mean `left` should be `null`. Going back to the original regression example, at the return statement, `result` is simply not equal to non-null constant (including `null`), but we shouldn't conclude `result` is `null`.

To mitigate this, I extended `== null` processing to input condition tester so that users like processing general equality can pick effects under which conditions it wants to add. At this point, all bb tests are good, no more regressions. But then, I found literally tons of regressions in diagnostics tests. :\

------

I didn't investigate every single regression in diagnostic tests (because, really, it was too many), but after seeing couple examples, I noticed what dilemma `fir`'s null smartcast has:
```kt
if (out == null) {
  out?.println()  // type of `out` should remain as-is. With Nothing? cast, unresolvable.
} else {
  out.println()
}
``` 
So, I studied how FE 1.0 works with those cases, and found that
1) for the cast like bb test, `i` after null check is indeed of `Nothing?` type, and that is used to resolve `i.foo()`. BUT, the big difference is, it's not by insertion of type casting. That info is just used during resolution. At the end, there is a separate step that revisits receivers/arguments and adds necessary implicit casts if their type mismatch.
2) for the case like `out?` safe call, since the call candidate's expected type is same as the actual type of the variable, no visible type change is added. Of course no implicit cast at the end, either.

On the other hand, in `fir`, smartcast info accumulated inside data-flow analysis is always _materialized_ into actual type casting (of course, via `FirExpressionWithSmartcast` and then IR type operation with `IMPLICIT_CAST`). Such materialized smartcast affects the following resolution on that specific variable, even though sometimes we need the original type.

Perhaps, the first issue---always materializing smartcast info from data-flow analysis---is out of scope of this PR, and needs bigger discussion. For the second issue---when to smartcast or not, or when to use original type (inside smartcast), one might think of retrieving the original type inside smartcast (if we decide to stick to the current, always-materializing smartcast) under certain conditions, but I can't find good conditions that satisfy all corner cases yet don't look hacky.

Instead, one (maybe equally ad-hoc) solution to resort the issues is to differentiate the use of `Nothing?` from `null` check: for smartcast, don't materialize it. For resolution, if the expected type and argument type are known to be different, and they're both nullable, and the smartcast info from the data-flow analysis indicates the argument is null (via `Nothing?` type), take it! With those, all regressions are gone, and resolutions that must use the fact that the receiver or argument is literally `null` are resolved.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
@@ -247,9 +247,35 @@ private fun BodyResolveComponents.typeFromSymbol(symbol: AbstractFirBasedSymbol<
     }
 }
 
-fun BodyResolveComponents.transformQualifiedAccessUsingSmartcastInfo(qualifiedAccessExpression: FirQualifiedAccessExpression): FirQualifiedAccessExpression {
+fun BodyResolveComponents.transformQualifiedAccessUsingSmartcastInfo(
+    qualifiedAccessExpression: FirQualifiedAccessExpression
+): FirQualifiedAccessExpression {
     val typesFromSmartCast = dataFlowAnalyzer.getTypeUsingSmartcastInfo(qualifiedAccessExpression) ?: return qualifiedAccessExpression
     val originalType = qualifiedAccessExpression.resultType.coneType
+    // For example, if (x == null) { ... },
+    //   we don't want to smartcast to Nothing?, but we want to record the nullability to its own kind of node.
+    // TODO: should we differentiate x == null v.s. x is Nothing?
+    if (typesFromSmartCast.any { it.isNullableNothing }) {
+        val typesFromSmartcastWithoutNullableNothing =
+            typesFromSmartCast.filterTo(mutableListOf()) { !it.isNullableNothing }.also {
+                it += originalType
+            }
+        val intersectedTypeWithoutNullableNothing =
+            ConeTypeIntersector.intersectTypes(session.inferenceComponents.ctx, typesFromSmartcastWithoutNullableNothing)
+        val intersectedTypeRefWithoutNullableNothing = buildResolvedTypeRef {
+            source = qualifiedAccessExpression.resultType.source?.fakeElement(FirFakeSourceElementKind.SmartCastedTypeRef)
+            type = intersectedTypeWithoutNullableNothing
+            annotations += qualifiedAccessExpression.resultType.annotations
+            delegatedTypeRef = qualifiedAccessExpression.resultType
+        }
+        return buildExpressionWithSmartcastToNull {
+            originalExpression = qualifiedAccessExpression
+            // TODO: Use Nothing? during resolution?
+            typeRef = intersectedTypeRefWithoutNullableNothing
+            // NB: Nothing? in types from smartcast in DFA is recorded here (and the expression kind itself).
+            this.typesFromSmartCast = typesFromSmartCast
+        }
+    }
     val allTypes = typesFromSmartCast.also {
         it += originalType
     }
@@ -259,6 +285,7 @@ fun BodyResolveComponents.transformQualifiedAccessUsingSmartcastInfo(qualifiedAc
         source = qualifiedAccessExpression.resultType.source?.fakeElement(FirFakeSourceElementKind.SmartCastedTypeRef)
         type = intersectedType
         annotations += qualifiedAccessExpression.resultType.annotations
+        delegatedTypeRef = qualifiedAccessExpression.resultType
     }
     return buildExpressionWithSmartcast {
         originalExpression = qualifiedAccessExpression
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
@@ -149,7 +149,7 @@ abstract class FirDataFlowAnalyzer<FLOW : Flow>(
 
     private var contractDescriptionVisitingMode = false
 
-    protected val any = components.session.builtinTypes.anyType.type
+    private val any = components.session.builtinTypes.anyType.type
     private val nullableNothing = components.session.builtinTypes.nullableNothingType.type
 
     @PrivateForInline
@@ -472,11 +472,18 @@ abstract class FirDataFlowAnalyzer<FLOW : Flow>(
     ) {
         val leftIsNullable = leftOperand.coneType.isMarkedNullable
         val rightIsNullable = rightOperand.coneType.isMarkedNullable
+
         // left == right && right not null -> left != null
+        // [processEqNull] adds both implications: operator call could be true or false. We definitely need the matched case only.
+        fun shouldAddImplicationForStatement(operationStatement: OperationStatement): Boolean {
+            // Only if operation statement is == True, i.e., left == right
+            return (operation.isEq() && operationStatement.operation == Operation.EqTrue) ||
+                    (!operation.isEq() && operationStatement.operation == Operation.EqFalse)
+        }
         when {
             leftIsNullable && rightIsNullable -> return
-            leftIsNullable -> processEqNull(node, leftOperand, operation.invert())
-            rightIsNullable -> processEqNull(node, rightOperand, operation.invert())
+            leftIsNullable -> processEqNull(node, leftOperand, operation.invert(), ::shouldAddImplicationForStatement)
+            rightIsNullable -> processEqNull(node, rightOperand, operation.invert(), ::shouldAddImplicationForStatement)
         }
 
         if (operation == FirOperation.IDENTITY || operation == FirOperation.NOT_IDENTITY) {
@@ -484,7 +491,25 @@ abstract class FirDataFlowAnalyzer<FLOW : Flow>(
         }
     }
 
-    private fun processEqNull(node: EqualityOperatorCallNode, operand: FirExpression, operation: FirOperation) {
+    /*
+     * Process x == null in general: add implications for both cases.
+     * E.g., say d1 is an eq operator call node, d2 represents `x`, the variable inside the operator call. So, d1: x == null
+     * This util adds: "d1 == True -> d2 == null" and "d1 == False -> d2 != null"
+     * so that both branches after the operator call can infer the type of x.
+     *
+     * However, users can specify what conditions are of interest.
+     * E.g., say left == right _and_ right != null, then we can conclude left != null.
+     * In this example, say d1 is an eq operator call (left == right), and d2 is left.
+     * Unlike general cases, we want to add: "d1 == True -> d2 != null", and nothing more because the counter part,
+     * "d1 == False -> d2 == null" doesn't hold. That is, left != right and right != null don't mean left == null. It just means, left is
+     * something different from right, including null. By filtering "d1 == True" condition only, all the remaining logic can be shared.
+     */
+    private fun processEqNull(
+        node: EqualityOperatorCallNode,
+        operand: FirExpression,
+        operation: FirOperation,
+        shouldAddImplicationForStatement: (OperationStatement) -> Boolean = { true }
+    ) {
         val flow = node.flow
         val expressionVariable = variableStorage.createSyntheticVariable(node.fir)
         val operandVariable = variableStorage.getOrCreateVariable(node.previousFlow, operand)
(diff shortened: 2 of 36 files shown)


jsjeon:
Just in case, I filed https://youtrack.jetbrains.com/issue/KT-41015 and https://youtrack.jetbrains.com/issue/KT-41017, and split the commits, respectively.

jsjeon:
@demiurg906 I saw you started `FirDataflowAnalyzer` and `processEq`. Could you take a look at the 1st commit?

@dzharkov I saw you added nullability smartcast after `!is`. Could you take a look at this overall?

Thanks!

mglukhikh:
Just to mention: I didn't forget about this PR, but I'm not sure these things should work this way in FIR. May be it's a good idea for @demiurg906 (or @dzharkov ) to look at this, and then to organize some design discussion about these "null smartcasts" (?!).

cc @semoro

demiurg906:
Hi @jsjeon. Thank you for big job, but I suggest to delay merging this PR, because we have lack of language design in this case. I, personally, think that cases from first black box test should **not** work, because smartcast to `Nothing?` is something strange and not intuitive for users (it's hard for them to understand that `Nothing` is a subtype for each possible type). So I think we can return to this PR after we make a design decision about such smartcasts.

@dzharkov what do you think?

dzharkov:
I agree with @demiurg906 that this issue is mostly a matter of language design

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
```diff
@@ -247,9 +247,35 @@ private fun BodyResolveComponents.typeFromSymbol(symbol: AbstractFirBasedSymbol<
...
+    qualifiedAccessExpression: FirQualifiedAccessExpression
+): FirQualifiedAccessExpression {
     val typesFromSmartCast = dataFlowAnalyzer.getTypeUsingSmartcastInfo(qualifiedAccessExpression) ?: return qualifiedAccessExpression
     val originalType = qualifiedAccessExpression.resultType.coneType
+    // For example, if (x == null) { ... },
+    //   we don't want to smartcast to Nothing?, but we want to record the nullability to its own kind of node.
+    // TODO: should we differentiate x == null v.s. x is Nothing?
+    if (typesFromSmartCast.any { it.isNullableNothing }) {
+        val typesFromSmartcastWithoutNullableNothing =
+            typesFromSmartCast.filterTo(mutableListOf()) { !it.isNullableNothing }.also {
```
jsjeon, line 260: Here, I only discarded `Nothing?`, but an alternative I tried was:
```kt
val nullableTypesFromSmartcastWithoutNullableNothing =
    typesFromSmartcast.mapNotNullTo(mutableListOf()) {
        if (it.isNullableNothing) null else it.withNullability(...NULLABLE)
    }...
```
i.e., making everything else _nullable_ (since it's cast to literally `null`).

But, encountered non-sense cases, e.g.,
```kt
if (x != null) {
  ...
  if (x == null) argShouldNotNull(<!>x<!>)
}
```
So, using the old info (from `x != null`) without `Nothing?` doesn't introduce any change, so I ended up with that.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
```diff
@@ -247,9 +247,35 @@ private fun BodyResolveComponents.typeFromSymbol(symbol: AbstractFirBasedSymbol<
...
+                it += originalType
```
jsjeon, line 261: An alternative I tried was:
```kt
it += originalType.withNullability(...NULLABLE)
```
because, again, it's cast to literally `null`.

But, encountered other kind of non-sense cases, like,
```kt
fun foo(x : NonNull) {
  if (x == null) { ... } // ?!
}
```
It's better to see `<!DEBUG_INFO_EXPRESSION_TYPE(NonNull & NonNull)!><!>` (where the 2nd one is the original type) instead of `... NonNull & NonNull?`. So, again, ended up with the current form.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/ResolveUtils.kt
```diff
@@ -247,9 +247,35 @@ private fun BodyResolveComponents.typeFromSymbol(symbol: AbstractFirBasedSymbol<
...
+        val intersectedTypeRefWithoutNullableNothing = buildResolvedTypeRef {
+            source = qualifiedAccessExpression.resultType.source?.fakeElement(FirFakeSourceElementKind.SmartCastedTypeRef)
+            type = intersectedTypeWithoutNullableNothing
+            annotations += qualifiedAccessExpression.resultType.annotations
+            delegatedTypeRef = qualifiedAccessExpression.resultType
+        }
+        return buildExpressionWithSmartcastToNull {
+            originalExpression = qualifiedAccessExpression
+            // TODO: Use Nothing? during resolution?
+            typeRef = intersectedTypeRefWithoutNullableNothing
```
jsjeon, line 274: Just for the record, If we want to use `Nothing?` for resolution, this is the place. We can use the intersection type that uses the all types from smartcast.

But, for some cases like
```kt
if (out == null) {
  out?.println(...) // not resolvable with Nothing? cast
} else {
  out.println(...)
}
```
we may need some ad-hoc application.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -500,20 +530,35 @@ abstract class FirDataFlowAnalyzer<FLOW : Flow>(
...
+            }
+            operationStatementTester.invoke(expressionVariable notEq isEq) {
+                flow.addImplication((expressionVariable notEq isEq) implies (operandVariable typeEq any))
+            }
 
-//            TODO: design do we need casts to Nothing?
-//            flow.addImplication((expressionVariable eq !isEq) implies (operandVariable typeEq nullableNothing))
-//            flow.addImplication((expressionVariable notEq !isEq) implies (operandVariable typeNotEq nullableNothing))
+            operationStatementTester.invoke(expressionVariable eq !isEq) {
+                flow.addImplication((expressionVariable eq !isEq) implies (operandVariable typeNotEq nullableNothing))
```
jsjeon, line 557: If we want to differentiate `== null` and `is Nothing?`, I think one option is, here, we can use a different cone type with a special type attribute, say, `NullableNothingFromConstantNull` to indicate this `Nothing?` came from `null`, not `is Nothing?`.

compiler/testData/diagnostics/tests/smartCasts/smartcastToNothingAfterCheckingForNull.fir.kt
```diff
@@ -24,6 +24,6 @@ fun g(x: B<Int>) {
 
     if (y is Nothing?) {
         f(y)
-        <!NONE_APPLICABLE!>g<!>(y)
+        g(y)
```
jsjeon, line 27: Well, at least, we are consistent now (see line 22 above).

compiler/testData/diagnostics/testsWithStdLib/contracts/smartcasts/throwsEffect.fir.kt
```diff
@@ -26,7 +26,7 @@ fun testSpilling(x: Any?) {
         myAssert(x is String)
         x.length
     }
-    x.<!UNRESOLVED_REFERENCE!>length<!>
+    x<!UNSAFE_CALL!>.<!>length
```
jsjeon, line 29: This one seems interesting. From the path that passed through `if`, `x is String`. From the synthetic `else`, `x == null` (or `x is Nothing?`). After type `or`ing, now `x is String?`. :\

I updated `isAcceptableForSmartcast` accordingly, but seems not good enough. I expect type `or` also needs to drop `Nothing?` at every step?

jsjeon (commented):
@demiurg906 This PR is reincarnated. As per your offline suggestion, a new kind of expression is introduced to wrap `null` check information, but not materialized any further. `Nothing?` is not included in the expression type, hence no resolution change either.

cc @punzki see if this is good enough for your quickfix work.

compiler/fir/tree/tree-generator/src/org/jetbrains/kotlin/fir/tree/generator/FirTreeBuilder.kt
```diff
@@ -108,6 +108,7 @@ object FirTreeBuilder : AbstractFirTreeBuilder() {
     val callableReferenceAccess = element("CallableReferenceAccess", Expression, qualifiedAccessExpression)
     val thisReceiverExpression = element("ThisReceiverExpression", Expression, qualifiedAccessExpression)
     val expressionWithSmartcast = element("ExpressionWithSmartcast", Expression, qualifiedAccessExpression)
+    val expressionWithSmartcastToNull = element("ExpressionWithSmartcastToNull", Expression, qualifiedAccessExpression)
```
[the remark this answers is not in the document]
jsjeon: I thought of that too, and decided to make it one-of-a-kind in order to avoid inception that this is kind of smartcast. Well, it's kind of smartcast, but not in the same way: no materialization during conversion to backend IR, not all smartcast types are intersected. So, during its introduction, I'd like to visit every use of existing smartcast expression kind, and make sure this new kind of expression is treated properly per situation.

Well, now every use/override point is visited, maybe we can restructure that on top of the current state. @demiurg906 WDYT?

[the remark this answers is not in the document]
demiurg906: Originally I thought that `FirExpressionWithSmartcastToNull` will fully delegate to original expression and won't have any smartcast type. But after review I see that there may be situations where there are situations when expression has meaningful smartcasts and `Nothing?` in same time. So current approach that `FirExpressionWithSmartcastToNull` is independent from `FirExpressionWithSmartcast` but fully copy it is not good. I see two options here:
1. Take Marks suggestion and inherit `FirExpressionWithSmartcastToNull` from `FirExpressionWithSmartcast` (since it adds new semantic to existed one)
2. Make `FirExpressionWithSmartcastToNull` dummy delegate to original expression without any `smartcasted` types and in case when expression has e.g. smartcast to `A?` and `Nothing?` wrap it twice: firstly in `FirExpressionWithSmartcast` and then in `FirExpressionWithSmartcastToNull`

I prefer option 1.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
```diff
@@ -472,19 +474,47 @@ abstract class FirDataFlowAnalyzer<FLOW : Flow>(
...
+     * In this example, say d1 is an eq operator call (left == right), and d2 is left.
+     * Unlike general cases, we want to add: "d1 == True -> d2 != null", and nothing more because the counter part,
+     * "d1 == False -> d2 == null" doesn't hold. That is, left != right and right != null don't mean left == null. It just means, left is
+     * something different from right, including null. By filtering "d1 == True" condition only, all the remaining logic can be shared.
+     */
+    private fun processEqNull(
+        node: EqualityOperatorCallNode,
+        operand: FirExpression,
+        operation: FirOperation,
+        operationStatementTester: (OperationStatement, () -> Unit) -> Unit = { _, block -> block() }
```
demiurg906, line 516: This lambda looks strange. It will be cleaner if this lambda will have `(OperationStatement) -> Boolean` type.

demiurg906: Check [rr/demiurg906/jsjeon/fir-null-smartcast](https://github.com/JetBrains/kotlin/tree/rr/demiurg906/jsjeon/fir-null-smartcast) branch, where I added this change

demiurg906 (changes requested):

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt
jsjeon: Makes sense. Thanks for the branch with an update.

demiurg906:
Merged manually

PR #3640, opened 2020-08-10
FIR: record use of backing field symbol to indeed add a backing field

jsjeon:
The motivation was `secondaryConstructors.withoutPrimarySimple`:
```
class A {
    val value: String
        get() = field + "K"

    constructor(o: String) {
        value = o
    }
}

fun box() = A("O").value
```
Class `A` doesn't have a primary constructor; property `value` is initialized in the secondary constructor; and in the non-default `get`ter of `value`, its initial value stored at the backing field is accessed as `field`. That is, we need a backing field for a property, if it is referred via `field` in its accessors.

To that end, when we transform `field` reference to a qualified access expression, we can mark that that property is referred via `field`. Then, during conversion, we can add the corresponding backing field.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
@@ -668,8 +668,10 @@ class Fir2IrDeclarationStorage(
                     val setter = property.setter
                     // TODO: this checks are very preliminary, FIR resolve should determine backing field presence itself
                     if (property.isConst || (property.modality != Modality.ABSTRACT && (irParent !is IrClass || !irParent.isInterface))) {
-                        if (initializer != null || getter is FirDefaultPropertyGetter ||
-                            property.isVar && setter is FirDefaultPropertySetter
+                        if (initializer != null ||
+                            getter is FirDefaultPropertyGetter ||
+                            property.isVar && setter is FirDefaultPropertySetter ||
+                            property.backingFieldSymbol.isReferenced
                         ) {
                             backingField = createBackingField(
                                 property, IrDeclarationOrigin.PROPERTY_BACKING_FIELD, descriptor,
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/FirCallResolver.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/FirCallResolver.kt
@@ -515,6 +515,7 @@ class FirCallResolver(
                 val candidate = candidates.single()
                 val coneSymbol = candidate.symbol
                 if (coneSymbol is FirBackingFieldSymbol) {
+                    coneSymbol.isReferenced = true
                     return buildBackingFieldReference {
                         this.source = source
                         resolvedSymbol = coneSymbol
--- a/compiler/fir/tree/src/org/jetbrains/kotlin/fir/symbols/impl/FirVariableSymbol.kt
+++ b/compiler/fir/tree/src/org/jetbrains/kotlin/fir/symbols/impl/FirVariableSymbol.kt
@@ -32,7 +32,10 @@ open class FirPropertySymbol(
     constructor(name: Name) : this(CallableId(name))
 }
 
-class FirBackingFieldSymbol(callableId: CallableId) : FirVariableSymbol<FirProperty>(callableId)
+class FirBackingFieldSymbol(
+    callableId: CallableId,
+    var isReferenced: Boolean = false
+) : FirVariableSymbol<FirProperty>(callableId)
 
 class FirDelegateFieldSymbol<D : FirVariable<D>>(callableId: CallableId) : FirVariableSymbol<D>(callableId) {
     val delegate: FirExpression
--- a/compiler/testData/codegen/box/operatorConventions/augmentedAssignmentInInitializer.kt
+++ b/compiler/testData/codegen/box/operatorConventions/augmentedAssignmentInInitializer.kt
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND_FIR: JVM_IR
 abstract class A {
     val b = B("O")
 
--- a/compiler/testData/codegen/box/secondaryConstructors/withoutPrimarySimple.kt
+++ b/compiler/testData/codegen/box/secondaryConstructors/withoutPrimarySimple.kt
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND_FIR: JVM_IR
 class A {
     val value: String
         get() = field + "K"


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -69,6 +68,7 @@ open class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransform
         qualifiedAccessExpression.transformAnnotations(this, data)
         qualifiedAccessExpression.transformTypeArguments(transformer, ResolutionMode.ContextIndependent)
 
+        val wasField = (qualifiedAccessExpression.calleeReference as? FirNamedReference)?.name == Name.identifier("field")
```
punzki, line 71: There is the constant `NAME_FOR_BACKING_FIELD` in `compiler/fir/tree/.../DefaultNames.kt`.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -662,8 +662,11 @@ class Fir2IrDeclarationStorage(
                     val setter = property.setter
                     // TODO: this checks are very preliminary, FIR resolve should determine backing field presence itself
                     if (property.isConst || (property.modality != Modality.ABSTRACT && (irParent !is IrClass || !irParent.isInterface))) {
-                        if (initializer != null || getter is FirDefaultPropertyGetter ||
-                            property.isVar && setter is FirDefaultPropertySetter
+                        if (initializer != null ||
+                            getter is FirDefaultPropertyGetter ||
+                            property.isVar && setter is FirDefaultPropertySetter ||
+                            property.isReferredViaField == true ||
+                            property.hasJvmFieldAnnotation
```
punzki, line 669: Should there be some test that would be enabled with this additional condition `hasJvmFieldAnnotation`?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: Good to know that, thanks.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
jsjeon: It should, but as mentioned above in the TODO, it's rather better to extend these conditions/tests separately during resolution. I'd file an issue later, like "when/how to determine backing field presence during resolution". I'll remove this for now, since it's out of scope of PR anyway.

jsjeon: Filed https://youtrack.jetbrains.com/issue/KT-41014

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
```diff
@@ -69,6 +68,7 @@ open class FirExpressionsResolveTransformer(transformer: FirBodyResolveTransform
         qualifiedAccessExpression.transformAnnotations(this, data)
         qualifiedAccessExpression.transformTypeArguments(transformer, ResolutionMode.ContextIndependent)
 
+        val wasField = (qualifiedAccessExpression.calleeReference as? FirNamedReference)?.name == NAME_FOR_BACKING_FIELD
```
mglukhikh, line 71: This can work in 99% cases but looks very hacky. E.g. it can be broken by a local `val field`.

mglukhikh (commented):
I will try to think about better solution here. Yet I don't like how it looks, sorry.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
```diff
@@ -116,6 +116,8 @@ fun FirRegularClass.addDeclaration(declaration: FirDeclaration) {
 
 private object IsFromVarargKey : FirDeclarationDataKey()
 var FirProperty.isFromVararg: Boolean? by FirDeclarationDataRegistry.data(IsFromVarargKey)
+private object IsReferredViaField: FirDeclarationDataKey()
+var FirProperty.isReferredViaField: Boolean? by FirDeclarationDataRegistry.data(IsReferredViaField)
```
mglukhikh, line 120: Well, it may be possible but it's better to avoid such code....

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: I didn't know that `field` isn't a hard keyword: https://kotlinlang.org/docs/reference/keyword-reference.html#soft-keywords 😮  That means, the condition below should also check if the resolved property indeed matches the enclosing property.

We don't have a special construct to wrap backing field `field`; it's just named reference up to this point, and that's why I thought it's inevitable to look at name `field` itself. But, speaking of checking the enclosing property, I just came up with an alternative: once resolved, if that is a property, and if that backing field access is within its own accessors, now we know that property needs the presence of the backing field. Will try this.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
jsjeon: Not sure you don't like this kind of attribute addition or marking the presence of backing field for a property. As per the TODO in the conversion: `FIR resolve should determine backing field presence itself`, it is okay to mark the presence of backing field somehow...? E.g., adding a bit to `FirProperty` or encoding as a resolved state change of `FirProperty`?

mglukhikh: Tomorrow we will discuss it and decide something.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirExpressionsResolveTransformer.kt
jsjeon: I have no idea why I have checked the name `field` and backing field reference below together. Checking the backing field reference is simply enough, since it literally means a backing field of a property is referenced! Also, considering where to mark, I found it's better to do something when the backing field reference is generated, not here. Updated the other place instead.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/declarations/FirDeclarationUtil.kt
jsjeon: After seeing how resolve state works, I found it doesn't make sense to add the presence of a backing field as a resolved state. Also, instead of `FirProperty`, I found that a backing field symbol is always bound, so when that symbol is used to create a backing field reference, actually that is a good timing to indicate that this property's backing field is referenced. Hence, added a flag to the symbol.

mglukhikh:
Looks much better, thank you. We'll discuss backing fields last time today, and then I'll merge this PR (may be with some simple changes).

mglukhikh:
Merged manually 👍 During today discussion, colleagues convinced me to return attribute back: https://github.com/JetBrains/kotlin/commit/c6417696cfbd1ccc47a3e828ccd4f56906aedaf4. Also I introduced 'hasBackingField': https://github.com/JetBrains/kotlin/commit/962059a878e9881a0d4df7c07168e30f3c590d94 and fixed attribute exception: https://github.com/JetBrains/kotlin/commit/5702fb3b2230d68d56ea0ef97ad5a062b892573a.

jsjeon:
Thank you for the update!

PR #3641, opened 2020-08-10
FIR2IR: make the scope of body of do-while loop transparent

jsjeon:
by using `IrComposite`, not just `IrBlock`.

Otherwise, tests like `kt3280`, which has a loop condition that refers to a local variable inside the loop body, would fail due to the missing variables in the scope stack (in IR backend).


mglukhikh:
Merged manually 👍

PR #3642, opened 2020-08-10
[FIR] Checkers performance improvement 

vldF:


compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/CanBeReplacedWithOperatorAssignmentChecker.kt
```diff
@@ -22,16 +22,19 @@ import org.jetbrains.kotlin.fir.types.coneType
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 
 object CanBeReplacedWithOperatorAssignmentChecker : FirExpressionChecker<FirVariableAssignment>() {
-    override fun check(functionCall: FirVariableAssignment, context: CheckerContext, reporter: DiagnosticReporter) {
-        val lValue = functionCall.lValue
+    override fun check(assignment: FirVariableAssignment, context: CheckerContext, reporter: DiagnosticReporter) {
```
demiurg906, line 28: You should name value parameters of overriden function same as they named in original function (or you can get errors with function calls with named arguments)

It's better to rename `functionCall` in `FirExpressionChecker` to `expression`

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/CanBeReplacedWithOperatorAssignmentChecker.kt
```diff
@@ -22,16 +22,19 @@ import org.jetbrains.kotlin.fir.types.coneType
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
 
 object CanBeReplacedWithOperatorAssignmentChecker : FirExpressionChecker<FirVariableAssignment>() {
```
demiurg906, line 27: Also it's better to inherit from `FirVariableAssignmentChecker` alias

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/CanBeReplacedWithOperatorAssignmentChecker.kt
```diff
@@ -22,16 +22,19 @@ import org.jetbrains.kotlin.fir.types.coneType
...
-    override fun check(functionCall: FirVariableAssignment, context: CheckerContext, reporter: DiagnosticReporter) {
-        val lValue = functionCall.lValue
+    override fun check(assignment: FirVariableAssignment, context: CheckerContext, reporter: DiagnosticReporter) {
+        val lValue = assignment.lValue
         if (lValue !is FirResolvedNamedReference) return
-        val operator = functionCall.psi?.children?.getOrNull(1) ?: return
-        if (operator.text != "=") return
+
+        val operator = assignment.psi?.children?.getOrNull(1) ?: return
```
demiurg906, line 32: This is bad, because that checker won't work in light tree mode. I see, that this code was written before, so it's ok to push it in this PR, but you should fix it in future

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/CanBeReplacedWithOperatorAssignmentChecker.kt
```diff
@@ -56,9 +59,12 @@ object CanBeReplacedWithOperatorAssignmentChecker : FirExpressionChecker<FirVari
 
             val isLeftMatch = isHierarchicallyTrue(operationToken, leftExpression?.operationToken)
                     && leftExpression?.matcher(variable) == true
+            if (isLeftMatch) return true
             val isRightMatch = isHierarchicallyTrue(operationToken, rightExpression?.operationToken)
                     && rightExpression?.matcher(variable) == true
-            isLeftMatch or isRightMatch
+            if (isRightMatch) return true
```
demiurg906, line 65: Just return `isRightMatch`

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/EmptyRangeChecker.kt
```diff
@@ -41,4 +41,28 @@ object EmptyRangeChecker : FirBasicExpresionChecker() {
         }
     }
 
+    private val FirFunctionCall.rangeLeft: Long?
```
demiurg906, line 44: Add comment for proper integer operator calls checking

demiurg906 (commented):
Please, name commits properly. `Refactoring` says nothing. `Refactor some extended checker` makes sence

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
```diff
@@ -0,0 +1,99 @@
...
+    private val symbols: MutableSet<FirPropertySymbol> = mutableSetOf()
+
+    override fun visitNode(node: CFGNode<*>) {}
+
+    override fun visitVariableDeclarationNode(node: VariableDeclarationNode) {
+        symbols += node.fir.symbol
+    }
+}
+
+class DataCollector(private val localProperties: Set<FirPropertySymbol>) :
```
demiurg906, line 55: `DataCollector` is too common name. Please rename it to something like `PropertyInitializationInfoCollector`

demiurg906 (changes requested):

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/CanBeReplacedWithOperatorAssignmentChecker.kt
vldF: Fix for this issue will be presented in special PR (I will add support for light tree in all extended checkers in it)

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/EmptyRangeChecker.kt
vldF: This comment was added few lines before

PR #3647, opened 2020-08-12
KT-40985 "Remove explicit type arguments" is suggested when type has an annotation

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40985


dimonchik0036 (approved):
Thanks! https://github.com/jetbrains/kotlin/commit/5e9333773cab661a21f7907b598b1b6a220180c8

PR #3649, opened 2020-08-13
Support .srcjar files as source roots

gharrma:
@udalov @ausatiy let me know what you think. Thanks!

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/K2JVMCompiler.kt
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/K2JVMCompiler.kt
@@ -190,7 +190,7 @@ class K2JVMCompiler : CLICompiler<K2JVMCompilerArguments>() {
                     addCommonSourceFiles(arg)
                 }
 
-                if (File(arg).isDirectory) {
+                if (arg.endsWith(".srcjar") || File(arg).isDirectory) {
                     addJavaSourceRoot(JavaRootPath(arg, args.javaPackagePrefix))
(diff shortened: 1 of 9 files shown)


udalov:
Could you please explain what would be the alternative (and its costs) to having this support built into the Kotlin compiler? So that we would get a clear picture why we are adding and will be supporting this.

gharrma:
Yes, good point.

I think there are two alternatives:

1. We can patch the Kotlin compiler ourselves to include this change.

2. We can unzip .srcjar files to disk before passing them off to the Kotlin compiler, and then delete the unzipped files when finished. This should work fine, it's just somewhat awkward and requires extra disk I/O.

Right now we are using option #1 for Android Lint, since Lint already bundles a custom build of the Kotlin compiler. So the cost is not too high.

As for other teams at Google using Kotlin with Bazel, I'm not sure how they currently deal with .srcjar files. I will reach out to them and see if they want to comment.

udalov:
Thanks! Yes, it'll be very helpful to hear from other teams using Bazel.

Also how does it work for other compilers? E.g. javac probably doesn't support .srcjar files, and so I'd imagine there's some kind of a workaround already which might be reused as is for Kotlin.

udalov:
Closing since there's been no movement on this. Feel free to reopen if it's still relevant. At this point though, I'm leaning more to avoid having this supported in the compiler, and implement this in build tools only, since similar support is likely required for other JVM languages.

gharrma:
Thanks for discussing; I agree this can stay closed.

PR #3650, opened 2020-08-14
FIR: introduce & use various comparators to sort members while de/serialization

jsjeon:
As per https://youtrack.jetbrains.com/issue/KT-20980, it's not easy to remove member sorting in deserialization, which also implies that we need to do member sorting in serialization as a counterpart. To sort members, FE 1.0 uses [MemberComparator](https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/resolve/MemberComparator.java) that is based on descriptor rendering. `fir` also has its own element rendering (`FirElement.render(...)`), but it is obvious that such rendering will visit unnecessary parts too. Thus, in this PR, an old-fashioned, tedious, yet less exhaustive comparators are introduced to sort member declarations in `fir`.

As a result, test data that list deserialized members are sorted. In particular,
1) they are "clustered", based on element kind. The notion of _priority_ and actual numbers conform to what FE 1.0 use; and
2) inside the cluster, they are "sorted" in an organized way, starting from name, return type, value parameters, etc.

In particular, you can see 1) less important (?) members are placed later, like constructor, compared to properties and functions that need more frequent searching; and 2) even in properties and functions, now extension ones with the same name are clustered.

### DIFF
--- a/compiler/fir/analysis-tests/testData/builtIns/kotlin-collections.txt
+++ b/compiler/fir/analysis-tests/testData/builtIns/kotlin-collections.txt
@@ -136,6 +136,15 @@ public abstract class LongIterator : R|kotlin/collections/Iterator<kotlin/Long>|
 }
 
 public abstract interface Map<K, out V> : R|kotlin/Any| {
+    public abstract interface Entry<out K, out V> : R|kotlin/Any| {
+        public abstract val key: R|K|
+            public get(): R|K|
+
+        public abstract val value: R|V|
+            public get(): R|V|
+
+    }
+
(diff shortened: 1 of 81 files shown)


jsjeon:
Friendly ping. Rebased atop `fir-deserialization` module migration.

Actually, I'm not sure things are well sorted and/or how to test them. Any suggestions, tests, improvements, etc., are welcome!

zarechenskiy:
Sorry for the long delay, changes seem fine to me! 
Let me check your changes on our tests and I'll be ready to merge them

PR #3652, opened 2020-08-14
JVM IR: Fix special bridge generation with external Kotlin dependencies

sfs:
In the descriptor hierarchy we replace special bridge methods with their Kotlin equivalents, which leads to some differences between external and local Kotlin declarations. This carries over to the IR declarations, where we use `initialSignatureFunction` to map a declaration back to its original (non-substituted) declaration. For fake overrides, this has the unintended effect of mapping to the original non-substituted version in a superclass, which causes the method signature mapper to return wrong results for fake overrides (which are only used in bridge generation).

Ideally, we should avoid `initialSignatureFunction` altogether and this is what I tried first. This currently fails because there are too many places in the IR backend which still use descriptors. When the IR structures deviate from what is expected for descriptors we then run into assertion failures when applying descriptor utilities on wrapped descriptors.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
@@ -201,7 +201,7 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
         mapSignature(function, false)
 
     private fun mapSignature(function: IrFunction, skipGenericSignature: Boolean, skipSpecial: Boolean = false): JvmMethodGenericSignature {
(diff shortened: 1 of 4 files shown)


udalov:
Thanks! The change looks good and I've merged it.

While testing, I've found another problem, unrelated to this change, on JVM-JVM_IR interop, and reported it here: https://youtrack.jetbrains.com/issue/KT-41123

sfs:
Thanks for the report! This is a problem with [KT-40277](https://youtrack.jetbrains.com/issue/KT-40277) that I didn't consider before.

The JVM IR backend generates a final method `remove(Ljava/lang/Object;)Ljava/lang/String;` in `A`, while the JVM backend generates a final method `remove(Ljava/lang/Object;)Ljava/lang/Object;`. The latter is overridden in subclasses on JVM IR, so we just have to be more careful with creating bridge methods for special bridges (as the `Object` versions could have been created as final).

PR #3654, opened 2020-08-15
Kotlin-test: Add assertIs assertion

benasher44:
Kotlin 1.4 supports using the type parameter in its contract. This will allow code like this in tests:
```kotlin
val s: Any = "test"
assertIs<String>(s)
// can now print s.length because of contract in assertIs
println("${s.length}")
```

[KT-45296](https://youtrack.jetbrains.com/issue/KT-45296)

### DIFF
--- a/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
+++ b/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
@@ -17,6 +17,7 @@ import kotlin.internal.*
 import kotlin.jvm.JvmName
 import kotlin.native.concurrent.ThreadLocal
 import kotlin.reflect.KClass
+import kotlin.reflect.typeOf
 
 /**
  * Current adapter providing assertion implementations
@@ -74,6 +75,31 @@ fun <@OnlyInputTypes T> assertSame(expected: T, actual: T, message: String? = nu
 /** Asserts that [actual] is not the same instance as [illegal], with an optional [message]. */
 fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? = null) {
     asserter.assertNotSame(message, illegal, actual)
+}
+
+/**
+ * Asserts that [value] is of type [T], with an optional [message].
+ *
+ * Note that due to type erasure the type check may be partial (e.g. assertIs<List<String>>(value)
+ * only checks for the class being [List] and not the type of its elements because it's erased).
+ */
+@SinceKotlin("1.5")
+@InlineOnly
+inline fun <reified T> assertIs(value: Any?, message: String? = null) {
+    contract { returns() implies (value is T) }
+    asserter.assertTrue({ messagePrefix(message) + "Expected value to be of type <${typeOf<T>()}>, actual <${value?.let { it::class }}>." }, value is T)
+}
+
+/**
+ * Asserts that [value] is not of type [T], with an optional [message].
+ *
+ * Note that due to type erasure the type check may be partial (e.g. assertNotIs<List<String>>(value)
+ * only checks for the class being [List] and not the type of its elements because it's erased).
+ */
+@SinceKotlin("1.5")
+@InlineOnly
+inline fun <reified T> assertNotIs(value: Any?, message: String? = null) {
+    asserter.assertFalse({ messagePrefix(message) + "Expected value to not be of type <${typeOf<T>()}>" }, value is T)
 }
 
 /** Asserts that the [actual] value is not `null`, with an optional [message]. */
--- a/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Utils.kt
+++ b/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Utils.kt
@@ -5,6 +5,7 @@
 
 package kotlin.test
 
+@PublishedApi // called from inline assert functions
 internal fun messagePrefix(message: String?) = if (message == null) "" else "$message. "
 internal expect fun lookupAsserter(): Asserter
 
--- a/libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
+++ b/libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
@@ -205,6 +205,33 @@ class BasicAssertionsTest {
     fun testExpectFails() {
         checkFailedAssertion { expect(1) { 2 } }
     }
+
+    @Test
+    fun testAssertIs() {
+        val s: Any = "test"
+        assertIs<String>(s)
+        assertEquals(4, s.length)
+    }
+
+    @Test
+    fun testAssertIsFails() {
+        val error = checkFailedAssertion { assertIs<Int>("test") }
+        assertTrue(error.message.startsWith("Expected value to be of type"))
+        assertTrue(error.message.contains("Int"))
+        assertTrue(error.message.contains("String"))
+    }
+
+    @Test
+    fun testAssertNotIs() {
+        assertNotIs<Int>("test")
+    }
+
+    @Test
+    fun testAssertNotIsFails() {
+        val error = checkFailedAssertion { assertNotIs<Int>(1) }
+        assertTrue(error.message.startsWith("Expected value to not be of type"))
+        assertTrue(error.message.contains("Int"))
+    }
 }
 
 


benasher44:
The main value of this over `assertTrue(value is T)` is the built-in nice error message that tells you what type the value ended up being upon failure

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
     asserter.assertNotSame(message, illegal, actual)
 }
 
+/** Asserts that [value] is [T], with an optional [message]. */
+@SinceKotlin("1.4")
```
ilya-g, line 76: Now it's time for `SinceKotlin("1.5")`  : )

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
...
+@InlineOnly
+inline fun <reified T: Any> assertIs(value: Any, message: String? = null) {
```
ilya-g, line 78: It would be valuable to support it for nullable type and value.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
...
+    contract { returns() implies (value is T) }
+    asserter.assertTrue({ messagePrefix(message) + "Expected value to be of type <${T::class}>, actual <${value::class}>." }, value is T)
```
ilya-g, line 80: `typeOf<T>()` provides more precise information about the type, especially its nullability.

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -205,6 +205,28 @@ class BasicAssertionsTest {
...
+    @Test
+    fun testAssertIs() {
+        val s: Any = "test"
+        assertIs<String>(s)
+        assertEquals(4, s.length)
+    }
+    
+    @Test
+    fun testAssertIsFails() {
+        checkFailedAssertion { assertIs<Int>("test") }
```
ilya-g, line 218: Could you inspect the assertion returned by `checkFailedAssertion` and check that its message contains the information we expect?

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
     asserter.assertNotSame(message, illegal, actual)
 }
 
+/** Asserts that [value] is [T], with an optional [message]. */
```
ilya-g, line 75:
```suggestion
/** Asserts that [value] is of type [T], with an optional [message]. */
```

Also makes sense to add a note that due to type erasure the type check may be partial, e.g. `assertIs<List<String>>(value)` only checks for the class being `List` and doesn't check the type of elements because it's erased.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
...
+@SinceKotlin("1.4")
+@InlineOnly
+inline fun <reified T: Any> assertIs(value: Any, message: String? = null) {
+    contract { returns() implies (value is T) }
+    asserter.assertTrue({ messagePrefix(message) + "Expected value to be of type <${T::class}>, actual <${value::class}>." }, value is T)
+}
+
+/** Asserts that [value] !is [T], with an optional [message]. */
```
ilya-g, line 83:
```suggestion
/** Asserts that [value] is not of type [T], with an optional [message]. */
```

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +72,21 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
...
+@SinceKotlin("1.4")
+@InlineOnly
+inline fun <reified T: Any> assertNotIs(value: Any, message: String? = null) {
+    asserter.assertFalse({ messagePrefix(message) + "Expected value to not be of type <${T::class}>" }, value is T)
```
ilya-g, line 87: `messagePrefix` is internal and needs a `@PublisedApi` annotation to be used in an inline function.

ilya-g (changes requested):

benasher44:
Thanks for the feedback! I should be able to get to it in the next week.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
benasher44: Would you recommend that? I could also add an internal non-inline helper function that calls it. What do you think?

benasher44:
Fixed some of the initial issues. Left one question, which may have impact on the other feedback, and then I'll fix those final ones too.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
benasher44: Oh never mind. That's actually the same problem. Do you think `@PublishedApi` is the way to go?

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
```diff
@@ -72,6 +73,31 @@ fun <@OnlyInputTypes T> assertNotSame(illegal: T, actual: T, message: String? =
...
+ * Asserts that [value] is of type [T], with an optional [message].
+ * 
+ * Note that due to type erasure the type check may be partial (e.g. assertIs<List<String>>(value)
+ * only checks for the class being [List] and not the type of its elements because it's erased).
+ */
+@SinceKotlin("1.5")
+@InlineOnly
+inline fun <reified T> assertIs(value: Any?, message: String? = null) {
+    contract { returns() implies (value is T) }
+    asserter.assertTrue({ messagePrefix(message) + "Expected value to be of type <${typeOf<T>()}>, actual <${value::class}>." }, value is T)
```
ilya-g, line 86: Please ensure that the library compiles and its tests pass. `coreLibsTest` gradle task can be used for that.
In particular, `value::class` won't compile for nullable `value` because `::class` requires a non nullable LHS argument. We can use `value?.let { it::class }` instead.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
ilya-g: Yes, I think it's ok to expose this function (`messagePrefix`) to public binary API.

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt
benasher44: Yes of course. Sorry I wasn't done yet— just wanted to get the initial improvements in

libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Utils.kt
```diff
@@ -5,8 +5,9 @@
...
 package kotlin.test
 
+@PublishedApi // called from inline assert functions
 internal fun messagePrefix(message: String?) = if (message == null) "" else "$message. "
 internal expect fun lookupAsserter(): Asserter
 
 @PublishedApi // required to get stable name as it's called from box tests
-internal fun overrideAsserter(value: Asserter?): Asserter? = _asserter.also { _asserter = value }
\ No newline at end of file
+internal fun overrideAsserter(value: Asserter?): Asserter? = _asserter.also { _asserter = value }
```
benasher44, line 13: 😫 will fix

benasher44:
Okay fixed final issues, cleaned up some white space things, and squashed it all into one neat commit. I tested locally in a scratch file, and I think everything should be compiling now 🤞

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -205,11 +205,35 @@ class BasicAssertionsTest {
...
+        val message = "Expected value to not be of type <kotlin.Int>."
+        checkFailedAssertion(message) { assertNotIs<Int>(1) }
+    }
 }
 
 
-internal fun checkFailedAssertion(assertion: () -> Unit): AssertionError {
-    return assertFailsWith<AssertionError> { withDefaultAsserter(assertion) }
+internal fun checkFailedAssertion(message: String? = null, assertion: () -> Unit): AssertionError {
+    return assertFailsWith(message)<AssertionError> { withDefaultAsserter(assertion) }
```
ilya-g, line 236: This is a message for failed `assertFailsWith`, not the message to check in the successfully caught exception.
Use the `AssertionError` returned from `checkFailedAssertion` to verify its properties in particular tests.

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
```diff
@@ -205,11 +205,35 @@ class BasicAssertionsTest {
...
+    @Test
+    fun testAssertIs() {
+        val s: Any = "test"
+        assertIs<String>(s)
+        assertEquals(4, s.length)
+    }
+
+    @Test
+    fun testAssertIsFails() {
+        val message = "Expected value to be of type <kotlin.Int>, actual <class kotlin.String>."
```
ilya-g, line 218: Note that different platforms can have different `toString` representations of `KType` and `KClass`. Better to check that the message contains essential information rather than to expect it in exact form.

libraries/kotlin.test/common/src/test/kotlin/kotlin/test/tests/BasicAssertionsTest.kt
benasher44: 🤦‍♂️ thanks!

ilya-g:
Thanks for the PR, I've merged it manually with some changes: https://github.com/JetBrains/kotlin/compare/e66eeefe~2...e66eeefe

In particular we've decided to rename `assertNotIs` to `assertIsNot` because the latter reads better.

benasher44:
Awesome thanks!!

PR #3655, opened 2020-08-16
KT-15262: Allow generate toString() to include properties with non-default getters

n-p-s:
[KT-15262](https://youtrack.jetbrains.com/issue/KT-15262). This addresses the comment raised in the previous [PR](https://github.com/JetBrains/kotlin/pull/3134#discussion_r428025016). Custom getters are suggested in the toString/ hashCode/equals dialog but they are unselected by default.


vladimirdolzhenko:

PR has been merged manually.

PR #3656, opened 2020-08-16
KT-20768: Add Method.kotlinGetter and Method.kotlinSetter

HeartPattern:
Add jvm to kotlin mapping for Method.kotlinGetter and kotlinSetter.
Refactor common code to Method.toKotlin.

 #KT-20768

### DIFF
--- a/core/reflection.jvm/src/kotlin/reflect/jvm/ReflectJvmMapping.kt
+++ b/core/reflection.jvm/src/kotlin/reflect/jvm/ReflectJvmMapping.kt
@@ -18,18 +18,17 @@
 
 package kotlin.reflect.jvm
 
+import org.jetbrains.kotlin.descriptors.runtime.components.ReflectKotlinClass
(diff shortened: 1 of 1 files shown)


udalov:

I propose to make the following changes:
- leave the implementation where it was, in `Method.kotlinFunction`
- instead of calling `.functions` in it or filtering by KFunction, call `.members` and consider all options for a member: only one function if the member is a function, but getter and setter if the member is a property
- implement `kotlinGetter` in the following way (and `kotlinSetter` similarly): `return kotlinFunction as? KProperty.Getter`

This way, we'll also make sure that `kotlinFunction` works correctly for property accessors.

Also, we need to add some tests to `compiler/testData/codegen/box/reflection/mapping`, run `./gradlew generateTests` and then ensure that they pass.

Would you be willing to contribute some or all of these or should I do it myself (probably significantly later)?

PR #3657, opened 2020-08-17
KT-19321 Warn if method with nullable return type always returns non-null

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-19321


vladimirdolzhenko:
LGTM. 

Thank you

vladimirdolzhenko:
Resolved small test failure, adjusted commit message and merged manually

PR #3658, opened 2020-08-17
FIR serializer: approximate remaining exotic types

jsjeon:
In FIR serialization, some special kinds of types are regarded as errors (after https://github.com/JetBrains/kotlin/commit/a545ec41a0303971ac06be0e0a9093edce99687e). They could be approximated via `AbstractTypeApproximator`.


mglukhikh:
We discussed it with @semoro internally and decided that exotic types should not appear in serializer. Situations mentioned in this PR should be fixed in resolve, not in serializer.

jsjeon:
I see. I'll revisit those tests and see if I can find where those exotic types are left.

jsjeon:
Just for the records. Tests were unmuted in different ways:
* https://github.com/JetBrains/kotlin/commit/e2cebb24e8fba9c753515e2908aaf942ca74177b
* https://github.com/JetBrains/kotlin/commit/0c13d3197ce213c9e10e1958003b860df282373a

and type approximation is also added:
https://github.com/JetBrains/kotlin/commit/0436a555d543d72e90bb26d2635dd1b8818e18ff

PR #3659, opened 2020-08-18
FIR2IR: coerce-to-unit expressions in IrBlock(Body)?

jsjeon:
Expressions inside `IrBlock` or `IrBlockBody` are expected to be implicitly casted to `Unit` in backend IR: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt#L177 and https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt#L193 (and to be honest, not sure what lowerings depend on that assumptions...) Anyway, a similar coercion during `fir2ir` conversion results in many `FIR_IDENTICAL` IR text listings and a couple bb tests being unmuted.


mglukhikh:
Merged manually. Looks like this PR does not have significant performance impact.

PR #3660, opened 2020-08-18
JVM IR: Generate LVT entries for parameters in inline $default methods (KT-36769)

sfs:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -296,7 +296,7 @@ class ExpressionCodegen(
     }
 
     private fun writeParameterInLocalVariableTable(startLabel: Label, endLabel: Label) {
-        if (irFunction.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER) return
+        if (!irFunction.isInline && irFunction.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER) return
```
max-kammerer, line 299: Could you explain a bit (because in my head all variable declarations should be inlined from original function to default one)?

sfs: It's mostly about matching the behavior of the JVM BE, which normally does not produce LVT entries for the synthetic `$default` methods. For inline methods these entries are required for debugging, though. We just hide the `$mask` and `$handler` arguments in this case.

I don't think there would be anything wrong with always producing LVT entries, except that we have several bytecode text tests which check that we don't.

max-kammerer: In current pr added generation for LVT in case of inline $default function, but same variables should be also inlined from original function. Looks like here would be a duplication

sfs: There isn't, since these variables are not shifted and so we don't generate LVT entries for them. This may be specific to `$default` methods. I don't know enough about the inliner to say for sure. :/

max-kammerer: @sfs Could you look a bit deeper in $default inline function compilation?
During compilation of $default inline function call to original one is inlined into it.
And all variables and parameters from original one should be inlined into $default one. It's strange that `p`  declaration was not inlined.
There are at least one more inconsistency: all locals variables are renamed with $iv suffix, e.g. :+1: 

```
inline fun test(p: String = "OK"): String {
    var x = 1
    return p
}

fun box() : String {
    return test()
}
//  public static synthetic test$default(Ljava/lang/String;ILjava/lang/Object;)Ljava/lang/String;
//    LOCALVARIABLE $i$f$test I L3 L5 3
//    LOCALVARIABLE x$iv I L4 L5 4
// old backend: x
//    LOCALVARIABLE p Ljava/lang/String; L0 L6 0


//  public final static box()Ljava/lang/String;
//    LOCALVARIABLE $i$f$test I L4 L6 3
//    LOCALVARIABLE x$iv$iv I L5 L6 4
// old backend: x$iv
//    LOCALVARIABLE p$iv Ljava/lang/String; L2 L7 0

```
But in old backend they are kept with original names (see InlineCodegenForDefaultBody)
cc @pyos

sfs: I'll look into it some more, especially the suffix for local variables looks like it might become a problem (or does the debugger ignore it?).

What I already know is that the variables are inlined into the $default method, but we only create LVT entries of the form `p$ip` for variables with kind `SHIFT` (see org/jetbrains/kotlin/codegen/inline/LocalVarRemapper.kt:93). For this particular call the arguments end up with kind `REMAPPED` and that's why the LVT entries are missing.

This seems different from what the JVM BE is doing. I'll step through it and see what's going on.

max-kammerer: > or does the debugger ignore it?

Maybe @yanex can help

sfs:
I changed the inliner in the JVM IR backend to match the behavior of the JVM backend. This affected the LVT entries for local variables in $default stubs as well as other inliner behavior such as anonymous object regeneration (see the added tests).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -194,6 +195,41 @@ class IrInlineCodegen(
...
+        val argsSize =
+            (Type.getArgumentsAndReturnSizes(jvmSignature.asmMethod.descriptor) ushr 2) - if (function.isStatic) 1 else 0
+        node.accept(object : InlineAdapter(transformedMethod, 0, childSourceMapper) {
+            override fun visitLocalVariable(name: String, desc: String, signature: String?, start: Label, end: Label, index: Int) {
+                // We only copy LVT entries for local variables, since we already generated entries for the method parameters,
+                if (index >= argsSize) super.visitLocalVariable(name, desc, signature, start, end, index)
+            }
+        })
+
+        transformedMethod.accept(MethodBodyVisitor(codegen.visitor))
```
pyos, line 229: Looks equivalent to just passing `MethodBodyVisitor` to `InlineAdapter`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -194,6 +195,41 @@ class IrInlineCodegen(
             ::IrDefaultLambda
         )
     }
+
+    override fun inlineCall(
```
pyos, line 199: Instead of hacking `IrInlineCodegen`, maybe create a separate implementation of `IrCallGenerator` that uses `ClassCodegen.generateMethodNode`?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
```diff
@@ -234,7 +238,7 @@ abstract class InlineCodegen<out T : BaseExpressionCodegen>(
         val inliner = MethodInliner(
             node, parameters, info, FieldRemapper(null, null, parameters), isSameModule,
             "Method inlining " + sourceCompiler.callElementText,
-            SourceMapCopier(sourceMapper, nodeAndSmap.classSMAP, callSite.takeIf { !isCallOfFunctionInCorrespondingDefaultDispatch }),
+            SourceMapCopier(sourceMapper, nodeAndSmap.classSMAP, callSite),
```
pyos, line 241: That parameter was not used for anything else; in the override, you can check that using IR structures

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
sfs: Good call, I didn't even look into the method visitors. This simplified the code a bit, thanks. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
sfs: That's a really good idea, thanks. I've changed the code to use a different call generator.

One thing I'm a bit unclear on is whether we still need to replicate the inline cycle check or not. I'm thinking it shouldn't matter, since in the worst case we'll just put the throw instruction in a slightly different place. What do you think?

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/InlineCodegen.kt
sfs: Right. And with the custom call generator the parameter is not needed at all and I've gotten rid of it. :)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
pyos: Well, the only way we could get an error *here* is if an inline function with defaults calls itself, but direct recursion is already checked by the frontend way before that. You can see that because the error message is different:
```
x.kt:1:28: error: inline function 'public inline fun f(x: Int = ...): Unit defined in root package in file x.kt' cannot be recursive
inline fun f(x: Int = 1) { f() }
                           ^
```
vs
```
x.kt:1:28: error: the 'g' invocation is a part of inline cycle
inline fun f(x: Int = 1) { g() }
                           ^
x.kt:3:18: error: the 'f' invocation is a part of inline cycle
inline fun g() = f()
                 ^
```
With indirect recursion, the check would trigger when generating `f` or `g`, not `f$default`.

pyos: On second though, if we start generating from `f`, then the cycle will indeed be detected at the call of `f` in `f$default`, and calling `generateMethodNode` may be fatal.

sfs: Alright, it looks like this can't happen right now, but I've added the check anyway to be on the safe side. Thanks for the feedback @pyos!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineDefaultCodegen.kt
```diff
@@ -0,0 +1,55 @@
...
+        assert(argumentExpression is IrGetValue || argumentExpression is IrTypeOperatorCall && argumentExpression.argument is IrGetValue)
+    }
+
+    override fun genInlineCall(callableMethod: IrCallableMethod, codegen: ExpressionCodegen, expression: IrFunctionAccessExpression) {
+        val function = expression.symbol.owner
+        val nodeAndSmap = codegen.classCodegen.generateMethodNode(function, codegen.delegatedPropertyOptimizer)
+        val childSourceMapper = SourceMapCopier(codegen.smap, nodeAndSmap.classSMAP)
+
+        val argsSize =
+            (Type.getArgumentsAndReturnSizes(callableMethod.asmMethod.descriptor) ushr 2) - if (function.isStatic) 1 else 0
```
max-kammerer, line 43: Looks like there is possible to copy all variable and remove first commit (it's not necessery to update test data in first commit in any case)

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -1161,11 +1163,16 @@ class ExpressionCodegen(
...
     ): IrCallGenerator {
         if (!element.symbol.owner.isInlineFunctionCall(context) ||
-            classCodegen.irClass.fileParent.fileEntry is MultifileFacadeFileEntry
+            classCodegen.irClass.fileParent.fileEntry is MultifileFacadeFileEntry ||
+            irFunction.isInvokeSuspendOfContinuation()
         ) {
             return IrCallGenerator.DefaultCallGenerator
         }
 
+        if (irFunction == context.mapping.defaultArgumentsDispatchFunction[element.symbol.owner]) {
```
max-kammerer, line 1172: Minor, but I would prefer to have more precise check here because current frontend diagnostic is too robust, but next code could become valid in some feature:
```
@Suppress("RECURSION_IN_INLINE")
inline fun test(p: String = test("OK")): String {
    return p
}

fun box() : String {
    return test()
}
```
Now it fails with
```
Caused by: java.lang.AssertionError: Assertion failed
	at org.jetbrains.kotlin.backend.jvm.codegen.IrInlineDefaultCodegen.genValueAndPut(IrInlineDefaultCodegen.kt:34)
```

Please also provide assertion message

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineDefaultCodegen.kt
sfs: The problem with that is that we would need to change the start label (see `InlineCodegenForDefaultBody`). Both are a bit of a hack, and I thought that this solution was a bit cleaner. What do you think?

In any case, I could move the test unmute to a later commit, but given that the first commit fixes the test in question I would rather leave it there.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
sfs: I didn't like this check either and you're right, since we're changing the code we might as well make it more robust. I've added a more precise check and a test with your example in c067aba.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineDefaultCodegen.kt
max-kammerer: >The problem with that is that we would need to change the start label (see InlineCodegenForDefaultBody). Both are a bit of a hack, and I thought that this solution was a bit cleaner. What do you think?

Oh yes, I've missed it

> In any case, I could move the test unmute to a later commit, but given that the first commit fixes the test in question I would rather leave it there.

As I see in current state (after all commits) this change is unneccessery:
-// 1 LOCALVARIABLE p\$iv Ljava/lang/String; L2 L4 0
+// 1 LOCALVARIABLE p\$iv Ljava/lang/String; L2 L\d+ 0

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: Thank you!

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineDefaultCodegen.kt
sfs: > As I see in current state (after all commits) this change is unneccessery:

Thanks, I missed that! I've moved the unmute to a later commit.

max-kammerer:
@sfs There is a bunch of failed tests on TC:
```
Failed
IrBytecodeTextTestGenerated$DefaultArguments.testLocalVariablesInInlinedDefaultStubsorg.jetbrains.kotlin.codegen.ir
71ms
1

Failed
IrBytecodeTextTestGenerated$DefaultArguments.testNoAnonymousObjectRegenerationInDefaultStuborg.jetbrains.kotlin.codegen.ir
85ms
2


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testInlinInDefaultorg.jetbrains.kotlin.codegen.ir
167ms
3


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testKt21827org.jetbrains.kotlin.codegen.ir
189ms
4


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testInlineAnonymousInDefaultorg.jetbrains.kotlin.codegen.ir
204ms
5


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testInlineAnonymousInDefault2org.jetbrains.kotlin.codegen.ir
214ms
6


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testInlinInDefault2org.jetbrains.kotlin.codegen.ir
349ms
7


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testNestedorg.jetbrains.kotlin.codegen.ir
317ms
8


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testDefaultLambdaInAnonymousorg.jetbrains.kotlin.codegen.ir
153ms
9


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testSimple2org.jetbrains.kotlin.codegen.ir
162ms
10


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap$DefaultLambda.testSimpleorg.jetbrains.kotlin.codegen.ir
173ms
11


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap.testDefaultFunctionWithInlineCallorg.jetbrains.kotlin.codegen.ir
95ms
12


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap.testDefaultFunctionorg.jetbrains.kotlin.codegen.ir
77ms
13


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testInlinInDefaultorg.jetbrains.kotlin.codegen.ir
108ms
14


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testKt21827org.jetbrains.kotlin.codegen.ir
161ms
15


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testInlineAnonymousInDefaultorg.jetbrains.kotlin.codegen.ir
92ms
16


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testInlineAnonymousInDefault2org.jetbrains.kotlin.codegen.ir
201ms
17


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testInlinInDefault2org.jetbrains.kotlin.codegen.ir
213ms
18


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testNestedorg.jetbrains.kotlin.codegen.ir
112ms
19


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testDefaultLambdaInAnonymousorg.jetbrains.kotlin.codegen.ir
94ms
20


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testSimple2org.jetbrains.kotlin.codegen.ir
116ms
21


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap.testAssertionorg.jetbrains.kotlin.codegen.ir
122ms
22


Failed
JvmOldAgainstIrBoxInlineTestGenerated$Smap$DefaultLambda.testSimpleorg.jetbrains.kotlin.codegen.ir
107ms
23


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap.testDefaultFunctionWithInlineCallorg.jetbrains.kotlin.codegen.ir
142ms
24


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap.testDefaultFunctionorg.jetbrains.kotlin.codegen.ir
109ms
25


Failed
IrCompileKotlinAgainstInlineKotlinTestGenerated$Smap.testAssertionorg.jetbrains.kotlin.codegen.ir
138ms
26


Failed
IrSteppingTestGenerated.testFunctionCallWithDefaultorg.jetbrains.kotlin.codegen.debugInformation
140ms
27


Failed
IrSteppingTestGenerated.testSimpleDefaultArgWithInlineorg.jetbrains.kotlin.codegen.debugInformation
272ms
28


Failed
IrSteppingTestGenerated.testAssertionorg.jetbrains.kotlin.codegen.debugInformation
147ms
29


Failed
IrSteppingTestGenerated.testLambdaStepInlineWithDefaultsorg.jetbrains.kotlin.codegen.debugInformation
```

sfs:
@max-kammerer Thanks for reporting, I've rebased on master and fixed these issues. Some were due to a bad revision, and there was a problem with the check for calls to the underlying method in a $default method.

max-kammerer:
@sfs Thank you for pr!
@pyos Thank you for help in review!

PR #3662, opened 2020-08-19
[JVM IR] Avoid Redundant special collection bridges

kandersen:
This PR avoids creating bridges when the resulting bridge signature is precisely that of the bridge-ee, and hence redundant.

The test expectations are a little inscrutable, but the separate expectations for the IR backend is needed as the two backends still disagree on a number of issues orthogonal to this one, like nullability assertions. The key point is the absence of a "getSize" bridge to the size property.

### DIFF
--- a/compiler/testData/codegen/bytecodeListing/specialBridges/redundantStubForSize.kt
+++ b/compiler/testData/codegen/bytecodeListing/specialBridges/redundantStubForSize.kt
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
 open class A1 {
     open val size: Int = 56
 }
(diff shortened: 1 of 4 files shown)


compiler/testData/codegen/bytecodeListing/specialBridges/redundantStubForSize.kt
```diff
@@ -1,4 +1,3 @@
-// IGNORE_BACKEND: JVM_IR
```
madsager, line 1: Would it be better to leave this one as not passing if there are still outstanding issues? If any of the remaining differences are not as expected, I would vote that we leave this one as not passing? It looks like the previous test covers the case that is fixed with this PR, so we have a regression test?

kandersen: Sure, good call.

kandersen:
@udalov Hi Alex, thanks for taking a look! Going to implement Mads' suggestion, I notice I must have been looking at a stale version of test results -- something is not right! Let me ping you when I think this is ready for review.

udalov:
Closing in favor of https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4483

PR #3663, opened 2020-08-19
JVM IR: Fix compilation of inline functions in anonymous objects inside of field initializers

sfs:
The corresponding classes end up nested in the class initializer of the surrounding class and we need to take this into account when creating instances of ClassCodegen.

This fixes KT-40332 on the JVM IR backend.


udalov:
Thanks!

PR #3664, opened 2020-08-20
Fix @Language injection when using named parameters

liutikas:
Start using names of arguments if they are available when fetching
the @Language injection information.

Fixes KT-35859


goodwinnk:
Thank you for the contribution! Merged to master as [777b16e0a](https://github.com/JetBrains/kotlin/commit/777b16e0a).

PR #3667, opened 2020-08-21
FIR: consider functions with defaults to be more specific 

pyos:
so that calls with missing arguments are correctly directed to the default implementation.

A function with more defaults and a function with a more specific return type are incomparable and should both be considered during overload resolution.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/scopes/impl/FirTypeIntersectionScope.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/scopes/impl/FirTypeIntersectionScope.kt
@@ -65,16 +65,13 @@ class FirTypeIntersectionScope private constructor(
         }
 
         val membersByScope = scopes.mapNotNull { scope ->
-            val resultForScope = mutableListOf<D>()
+            val resultForScope = mutableListOf<MemberWithBaseScope<D>>()
             scope.processCallables(name) {
                 if (it !is FirConstructorSymbol) {
-                    resultForScope.add(it)
+                    resultForScope.add(MemberWithBaseScope(it, scope))
                 }
             }
-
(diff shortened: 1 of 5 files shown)


dzharkov:
Hello!
Unfortunately. while this solution looks rather reasonable it might work not very correctly in subtle cases when a more specific candidate doesn't have default parameters while the least specific does
```
interface A {
    fun foo(x: Int = 1): Any
}

interface B {
    fun foo(x: Int): String
}

interface C : A, B

fun C.bar() {
    println(foo().length)
}

fun main() {
    object : C {
        override fun foo(x: Int) = x.toString()
    }.bar()
}
```

One of the workarounds, we are thinking of right now is stopping propagating inherited default values through the FIR in the scopes, but rather computing them online when resolving a call (of course with some sort of cache).
Also, currently, we're reviewing some design language issues like [KT-36188](https://youtrack.jetbrains.com/issue/KT-36188) that may severely affect the resulting implementation.
Thus, at the moment I'd propose to postpone the change until we have some resolution (I'll notify you once we have it if you don't mind)
Please, do not remove the branch for some time.

pyos:
Yes, I realized this might be a problem too. To clarify, though, there would not be a more specific function here: `fun foo(x: Int = 1): Any` and `fun foo(x: Int): String` would be incomparable, and then the former would be chosen based on the arguments (and if you call `foo(1)`, the latter would be chosen instead). The problem here is the backend, actually: to make it notice the default value *and* have a String return type, the callee cannot be either of those functions but rather must be a fake override of both, which currently doesn't exist at all until `FakeOverrideGenerator`.

dzharkov:
I see the backend issue, but it seems reasonable to have a consistent design and then implementing it both for frontend and backend
We'll try hard to have the design ASAP

PR #3669, opened 2020-08-25
FIR2IR: introduce & use ClassId-based lookup for local class as IrParent

jsjeon:
Members that belong to local classes can't set up `parent`, since symbol provider doesn't allow `ClassId` lookup for local classes. At least, we can look up cached local classes in the classifier storage.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
@@ -35,6 +35,7 @@ import org.jetbrains.kotlin.ir.types.IrType
 import org.jetbrains.kotlin.ir.types.impl.IrSimpleTypeImpl
 import org.jetbrains.kotlin.ir.util.IdSignature
 import org.jetbrains.kotlin.ir.util.constructors
+import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.Name
 
 class Fir2IrClassifierStorage(
@@ -139,6 +140,13 @@ class Fir2IrClassifierStorage(
         } else {
             classCache[klass]
         }
+    }
+
+    internal fun getCachedLocalClass(classId: ClassId): IrClass? {
+        require(classId.isLocal) {
+            "As the function name implies, it ought to be used to look up _local_ classes."
+        }
+        return localStorage.getLocalClass(classId)
     }
 
     private fun FirRegularClass.enumClassModality(): Modality {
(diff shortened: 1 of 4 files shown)


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
```diff
@@ -141,6 +142,14 @@ class Fir2IrClassifierStorage(
         }
     }
 
+    internal fun getCachedIrClass(classId: ClassId): IrClass? {
```
mglukhikh, line 145: As I can see, this function is called *only* with local class id. So may be it's worth renaming it to `getCachedLocalClass` and removing the second if branch.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrLocalStorage.kt
```diff
@@ -49,6 +47,10 @@ class Fir2IrLocalStorage {
         return localClassCache[localClass]
     }
 
+    fun getLocalClass(classId: ClassId): IrClass? {
+        return localClassCache.keys.find { firClass -> firClass.classId == classId }?.let { localClassCache[it] }
```
mglukhikh, line 51: I think it's a bit faster to search for entry from the beginning:

```
        return localClassCache.entries.find { (firClass, _) -> firClass.classId == classId }?.value
```

mglukhikh (changes requested):
Good shot, thank you! Exception from `getClassLikeSymbolByFqName` is the common failure case also in full pipeline test. NB: could you please look at other use-sites of this function inside declaration storage? May be it's worth changing them in the same manner.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrClassifierStorage.kt
jsjeon: Right; renamed and also added `require` to make sure.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrLocalStorage.kt
jsjeon: Of course! 🤦  Thanks!

jsjeon:
> could you please look at other use-sites of this function inside declaration storage? May be it's worth changing them in the same manner.

There were three use-sites inside the declaration storage, and I think two of them can be changed in the same way. The remaining one explicitly said it wants to retrieve class by FqName.

mglukhikh (approved):

mglukhikh:
Merged 👍

PR #3670, opened 2020-08-25
FIR2IR: allow array expression as named argument for vararg

jsjeon:
The motivation is bb test `kt37779`:
```
fun test(vararg s: String) = s[1]

fun box(): String =
    test(s = arrayOf("", "OK"))
```

Since 1.4, an expression with an array type can be passed to _named_ _vararg_ parameter.


mglukhikh:
Merged, thank you!

PR #3671, opened 2020-08-25
[FIR] Add light tree support to extended checkers

vldF:


mglukhikh:
Merged manually with some fixes.

PR #3672, opened 2020-08-25
KT-41313: Only load .jar files when locating annotation processors

gavra0:
Previously, we'd only check if file exists and try to load it. However,
some compile classpaths may contain .tar.gz files.

Test: ProcessorLoaderTest

### DIFF
--- a/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
+++ b/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
@@ -100,7 +100,7 @@ open class ProcessorLoader(private val options: KaptOptions, private val logger:
                         processSingleInput(it.inputStream())
                     }
                 }
-                file.isFile -> {
+                file.isFile && file.extension.equals("jar", ignoreCase = true) -> {
                     ZipFile(file).use { zipFile ->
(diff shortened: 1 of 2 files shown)


gavra0:
//cc @yanex

plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/ProcessorLoader.kt
```diff
@@ -100,7 +100,7 @@ open class ProcessorLoader(private val options: KaptOptions, private val logger:
                         processSingleInput(it.inputStream())
                     }
                 }
-                file.isFile -> {
+                file.isFile && file.extension == "jar" -> {
```
yanex, line 103: Maybe we should check it case-insensitively?

gavra0: Good catch, updated in 627380798c8026d6625e4edc3a669a6dbbae2fc5.

yanex:
Thanks for your contribution!

The commit was merged to [master](https://github.com/JetBrains/kotlin/commit/91b99da7a098c611120471c8818bd3f239c40eee) and [1.4.0](https://github.com/JetBrains/kotlin/commit/f1e2def771b44a33909892879b5f6db061a0a9fe) branches. In other words, it will be included into Kotlin 1.4.10 release.

PR #3673, opened 2020-08-26
FIR2IR: convert adapted callable reference with vararg

jsjeon:
This is the follow-up of https://github.com/JetBrains/kotlin/pull/3614, where I said:

> Note that, in this PR, only the adaptation on reference type is done. For some other cases, we will need a different conversion for such adapted callable reference, which will reuse #3599.

That PR adapted callable reference type with `vararg` such that the reference can be used in a different form at the use site. This PR now converts such adapted callable reference. All necessary pieces of callable reference conversion were done at #3599. This PR just extends how value argument from the adapter function can be mapped to `vararg` parameter in the adaptee.


mglukhikh (approved):

jsjeon:
Oops, noticed the typo in the commit message, thanks to your title change. 👍  Updated the commit message as well.

PR #3674, opened 2020-08-26
KT-41338 False positive "Redundant 'asSequence' call" when Map.Entry properties are used.

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41338


vladimirdolzhenko:
LGTM

PR #3675, opened 2020-08-26
Add Window.isSecureContext declaration (KT-41366)

SerVB:
Hi! I guess this closes <https://youtrack.jetbrains.com/issue/KT-41366>. Could you take a look?


trilis:
Hi! Thanks for PR, but unfortunately this part of stdlib is auto-generated from WebIDL specifications, so I can't accept this PR. However, WebIDL specification in question (https://w3c.github.io/webappsec-secure-contexts/#idl-index) is quite small and simple, so I'll try to add it to our sources later.

PR #3676, opened 2020-08-26
Split changelog to reduce file size (KT-41399)

SerVB:
A possible solution is in this PR.


4u7:
Looks like this is already done in master

PR #3679, opened 2020-08-28
JVM IR: Fixes for nested inner classes

sfs:
- Fix the lookup for the `this` reference in `InnerClassesLowering`. The fix was found by @pyos. This previously caused problems with the compilation of nested inner classes, see the added test.
- Create fields for captured values as package private, same as in the JVM backend. Otherwise, nested inner classes would require accessors. There were also some cases where the fields needed to be package private instead of private anyway, see the unmuted tests.


udalov:
Thanks!
Regarding the second issue, we initially [thought](https://youtrack.jetbrains.com/issue/KT-36987#focus=Comments-27-3984332.0-0) that generating them as private in the old backend would be more "correct", but we missed that that would lead to more synthetic accessors which is completely unnecessary.

PR #3680, opened 2020-08-28
JVM_IR: extract initializer statements into a separate method

pyos:
if they need to be executed from multiple constructors.

This reduces code size, removes the need to move classes out of initializers, and fixes a case where copying causes a single lambda to be inlined into two locations (KT-41465).


udalov:
Nice idea, but if such initializer contains a final field assignment, it leads to `putfield` outside of the constructor. [JVMS11 §6.5](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-6.html#jvms-6.5.putfield) says:

> If the field is final, it must be declared in the current class, and the instruction must occur in an instance initialization method of the current class (§2.9.1).

Although the latest HotSpot verifier seems to accept such bytecode, it might break in the future. For example, the symmetric `putstatic`/`<clinit>` case was tolerated for a long time, but verifier started to enforce it in JDK 9.

udalov:
Oh, in fact verifier enforces the `putfield`/`<init>` rule as well, but for class files with version >= 53. So this example:
```
class A {
    val s: String

    constructor(s: String) {}
    constructor(x: Int) {}

    init {
        s = "OK"
    }
}

fun main() {
    println(A("").s)
}
```
fails if compiled with `-jvm-target 9` or later, and run on the corresponding JDK:
```
java.lang.IllegalAccessError: Update to non-static final field A.s attempted from a different method ($init) than the initializer method <init>
	at A.$init(2.kt:8)
	at A.<init>(2.kt:4)
```

pyos:
Ah, so that's why it wasn't done before...obvious in hindsight. Oh well, I suppose KT-41465 can also be fixed by having JvmLocalClassPopupLowering extract classes from inline lambdas too.

PR #3682, opened 2020-08-29
FIR2IR: handle more vararg spreads for adapted callable references

jsjeon:
Since https://github.com/JetBrains/kotlin/pull/3673, we can convert adapted callable reference with `vararg`, but only for a one-to-one mapping, like passing a parameter in the adapter of either array type or element type to the adaptee. However, as described in https://github.com/JetBrains/kotlin/pull/3614, there are more cases: no parameter passing (which is like passing an empty array to the `vararg` parameter) or multiple elements (a.k.a. spread). This PR extends/fixes the way https://github.com/JetBrains/kotlin/pull/3673 spreads parameter(s).


mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3684, opened 2020-08-31
JVM IR: Fix mangling for suspend functions with inline class params (KT-41374)

sfs:


udalov:
Thanks. Looks good as a temporary solution. I'll check the remote run and merge.

Meanwhile, I'm adding the JVM->JVM_IR and JVM_IR->JVM mixed box tests (on `compileKotlinAgainstKotlin` + `boxInline` test data), will add this case there too.

PR #3686, opened 2020-09-01
FIR: hide local types if needed

jsjeon:
The motivation was bb test `coroutines.kt21605` where the only difference between IR text listings from `fir2ir` v.s. `psi2ir` is the type of a certain function that just returns an anonymous object which acts like a single, one-time interface implementer:

```
inline fun crossInlineBuilder(crossinline block: (String) -> Unit) = object : Consumer {
    override fun consume(s: String) {
        block(s)
    }
}
```
This function's return type is `Consumer` from `psi2ir` v.s. `<no name provided>` (that anonymous object) from `fir2ir`. Since a similar pattern could appear in property, e.g.,
```
val prop
    get() = object : Consumer { ... }
```
conversion is not the right place (actually, too late). Hence, unwrapping the anonymous object's super type during declaration resolution.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -509,6 +511,44 @@ open class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransfor
...
+        val anonymousObject =
+            ((((this as? FirResolvedTypeRef)
+                ?.type as? ConeClassLikeType)
+                ?.lookupTag as? ConeClassLookupTagWithFixedSymbol)
+                ?.symbol as? FirAnonymousObjectSymbol)?.fir
+                ?: return this
+        // TODO: replace this ad-hoc condition with a better handling
+        //   Conflict: diagnostic test properties.inferenceFromGetters.objectExpression
+        //     has a property whose getter returns an anonymous object with own members, which are later invoked.
+        //     If we unwrap the type here, those member accesses become unresolved calls.
```
jsjeon, line 534: Discussed with other folks (@madsager @pyos), probably the current result---resolving a member in an anonymous object---is not a desired behavior, setting the right return type here, which results in unresolved calls there, would be correct?

jsjeon: I just learned when to _not_ expose anonymous object types. Now that remaining bb test is working; also, some `invisible` members are correctly hidden/unresolved.

jsjeon: Btw, special thanks to @pyos !

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -509,6 +515,42 @@ open class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransfor
...
+        if (visibility == Visibilities.Public ||
+            visibility == Visibilities.Protected ||
+            visibility == Visibilities.Internal ||
+            (visibility == Visibilities.Private && function?.isInline == true)
+        ) {
+            val anonymousObjectSupertype =
+                ((((this as? FirResolvedTypeRef)
+                    ?.type as? ConeClassLikeType)
+                    ?.lookupTag as? ConeClassLookupTagWithFixedSymbol)
+                    ?.symbol as? FirAnonymousObjectSymbol)?.fir
```
mglukhikh, line 544: Generally I would say that situation with local class should be similar, so (probably) we should hide type with local visibility if outer callable is public / protected / internal / private inline.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -496,8 +498,12 @@ open class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransfor
 
         val body = result.body
         if (result.returnTypeRef is FirImplicitTypeRef) {
+            val simpleFunction = function as? FirSimpleFunction
```
mglukhikh, line 501: Should we do the similar for property with initializer?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
jsjeon: Extended to general local types, thanks!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
jsjeon: Oops, the place I added covered functions _and_ accessors, so I thought that would be enough. Didn't think about that, added. Thanks!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
pyos: Returning a local non-anonymous type from a public function is actually just an error (EXPOSED_FUNCTION_RETURN_TYPE). Anonymous types with multiple supertypes should result in the same, by the way, and it's weird that FirExposedVisibilityDeclarationChecker isn't currently throwing an error on any public function with anonymous object return type.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
```diff
@@ -510,6 +517,47 @@ open class FirDeclarationsResolveTransformer(transformer: FirBodyResolveTransfor
...
+        isInlineFunction: Boolean = false
+    ): FirTypeRef {
+        if (containingCallableVisibility == null) {
+            return this
+        }
+        // Approximate types for non-private (all but package private or private) members.
+        // Also private inline functions, as per KT-33917.
+        if (containingCallableVisibility == Visibilities.Public ||
+            containingCallableVisibility == Visibilities.Protected ||
+            containingCallableVisibility == Visibilities.Internal ||
```
punzki, line 542: Somehow an `internal` _property_'s local type is not hidden. See unambiguousObjectExpressionType.fir.kt#L30 and #L140. (This is unresolved in old frontend.)

jsjeon: Good catch! It turned out that I missed some other places where property's return type is updated. Correctly unresolved now. Thanks!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirDeclarationsResolveTransformer.kt
jsjeon: Good to know there is a designated checker for that, and it looks like there are more detailed cases there. I'm not sure either why it isn't currently throwing any errors. Are those checkers invoked/involved in `fir` pipeline?

Regarding multiple supertypes, changed to return an error type instead.

mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3687, opened 2020-09-02
Add Map<K, V?>.filterValuesNotNull()

kageru:
Since `myMap.filterValues { it != null }` does not change the type to `Map<K, V>`, users either have to do an unchecked cast or implement this functionality themselves.

### DIFF
--- a/libraries/stdlib/src/kotlin/collections/Maps.kt
+++ b/libraries/stdlib/src/kotlin/collections/Maps.kt
@@ -483,6 +483,15 @@ public inline fun <K, V> Map<out K, V>.filterValues(predicate: (V) -> Boolean): 
     return result
 }
 
+/**
+ * Returns a map containing all key-value pairs with nonnull values.
+ *
+ * The returned map preserves the entry iteration order of the original map.
+ * @sample samples.collections.Maps.Filtering.filterValuesNotNull
+ */
+@Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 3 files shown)


libraries/stdlib/src/kotlin/collections/Maps.kt
```diff
@@ -483,6 +483,21 @@ public inline fun <K, V> Map<out K, V>.filterValues(predicate: (V) -> Boolean):
     return result
 }
 
+/**
+ * Returns a map containing all key-value pairs with nonnull values.
+ *
+ * The returned map preserves the entry iteration order of the original map.
+ * @sample samples.collections.Maps.Filtering.filterValuesNotNull
+ */
+public fun <K, V> Map<out K, V?>.filterValuesNotNull(): Map<K, V> {
```
pa-valentino, line 492: If I am not mistaken, you could use the existing function `filterTo` to shorten the code a bit.
It could look like this then
```kotlin
fun <K, V> Map<out K, V>.filterValuesNotNull(): Map<K, V> = this.filterTo(LinkedHashMap<K, V>()) { (_, value) -> value != null }
```

kageru: That’s true. I also thought about just returning
```kotlin
this.filterValues { it != null } as Map<K, V>
```
because I’m pretty sure that’s a safe cast.
If either of those would be preferred, I’ll update the PR accordingly.

kageru:
Updated the implementation following @pa-valentino’s suggestion.

PR #3688, opened 2020-09-02
doc: update collection example

bodin:
Update example for default buildList to not use capacity

KT-41136


qurbonzoda:
Thank you!

PR #3690, opened 2020-09-03
[JS_IR] Enable RangeContainsLowering optimizations

punzki:
Also unmuted some tests in `box/ranges/contains` by changing `assert()` calls (not available in JS) to `assertTrue()`. It seems to be the only reason they were muted in the first place.


romanart:
Thank you a lot :) 
LGTM
Will run CI and merge

romanart (approved):

PR #3691, opened 2020-09-03
[IR] Add IR support for setting parameters.

madsager:
Use it in the JVM_IR backend for default values code. The parameter
local has to be overwritten for the inliner to work.


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/LowerUtils.kt
```diff
@@ -57,6 +57,13 @@ abstract class AbstractVariableRemapper : IrElementTransformerVoid() {
         remapVariable(expression.symbol.owner)?.let {
             IrGetValueImpl(expression.startOffset, expression.endOffset, it.type, it.symbol, expression.origin)
         } ?: expression
+
+    override fun visitSetValue(expression: IrSetValue): IrExpression {
```
madsager, line 61: Actual change: when remapping parameter symbols, we now need to look at SetValue as well as GetValue.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SharedVariablesLowering.kt
```diff
@@ -118,11 +118,11 @@ class SharedVariablesLowering(val context: BackendContext) : BodyLoweringPass {
...
                 }
 
-                override fun visitSetVariable(expression: IrSetVariable, data: IrDeclarationParent?) {
-                    super.visitSetVariable(expression, data)
+                override fun visitSetValue(expression: IrSetValue, data: IrDeclarationParent?) {
+                    super.visitSetValue(expression, data)
 
                     val variable = expression.symbol.owner
-                    if (variable.initializer == null && getRealParent(variable) != data && variable in relevantVals) {
+                    if (variable is IrVariable && variable.initializer == null && getRealParent(variable) != data && variable in relevantVals) {
```
madsager, line 125: Actual change: adapt code to have same behavior as before.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -711,22 +711,22 @@ class ExpressionCodegen(
         return false
     }
 
-    override fun visitSetVariable(expression: IrSetVariable, data: BlockInfo): PromisedValue {
+    override fun visitSetValue(expression: IrSetValue, data: BlockInfo): PromisedValue {
         if (!handleIntVariableSpecialCases(expression)) {
             expression.value.markLineNumber(startOffset = true)
             expression.value.accept(this, data).materializeAt(expression.symbol.owner.type)
-            expression.markLineNumber(startOffset = true)
+            // We set the value of parameters only for default values. The inliner accepts only
```
madsager, line 718: Real change: be careful with the bytecode patterns for default arguments. This is a little too indirect for my taste. If we start setting parameters for some other reason the store will not have line numbers.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
```diff
@@ -157,7 +157,8 @@ class IrInlineCodegen(
         try {
             performInline(
                 expression.symbol.owner.typeParameters.map { it.symbol },
-                function.origin == IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER,
+                // Always look for default lambdas to allow custom default argument handling in compiler plugins.
```
madsager, line 160: Real change: In order to allow compiler plugins to have default argument handling in methods that are not FUNCTION_FOR_DEFAULT_PARAMETER, we always look for default lambdas (and bail out if there is no mask below, so there should be no compile-time performance penalty).

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/EnumClassLowering.kt
```diff
@@ -162,6 +163,13 @@ private class EnumClassLowering(val context: JvmBackendContext) : ClassLoweringP
                     IrGetValueImpl(expression.startOffset, expression.endOffset, it.type, it.symbol, expression.origin)
                 } ?: expression
 
+            override fun visitSetValue(expression: IrSetValue): IrExpression {
```
madsager, line 166: Real change: when remapping parameter symbols we now need to look at SetValue as well as GetValue.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmDefaultArgumentStubGenerator.kt
```diff
@@ -39,11 +38,7 @@ class JvmDefaultArgumentStubGenerator(override val context: JvmBackendContext) :
         //
         // This control flow limits us to an if-then (without an else), and this together with the
         // restriction on loading the parameter in the default case means we cannot create any temporaries.
-        +irIfThen(defaultFlag, irCall(this@JvmDefaultArgumentStubGenerator.context.ir.symbols.reassignParameterIntrinsic).apply {
-            putTypeArgument(0, parameter.type)
-            putValueArgument(0, irGet(parameter))
-            putValueArgument(1, default)
-        })
+        +irIfThen(defaultFlag, irSet(parameter.symbol, default))
```
madsager, line 41: Real change: remove the JVM-specific intrinsic and use the IR structure instead.

madsager (commented):
The purpose of this change is to allow compiler plugins (specifically the Compose compiler plugin) to implement its own default parameter handling and have the JVM inliner understand the code that is generated.

There are two things currently prohibiting that:

1. there is no IR support for reassigning value parameters which makes it impossible to generate bytecode patterns the inliner understands.

2. the inliner only looks for default lambdas in $default stubs marked with the FUNCTION_FOR_DEFAULT_PARAMETER origin.

For 1) this changes IrSetVariable to IrSetValue and allows assignment to parameters.

For 2) this change looks for default lambdas in the IR version of the bytecode inliner always (but with a fast bailout if there is no mask).

Because of renaming, there are many changes, I have marked the actual changes with comments to make it easier to look at.

I have not renamed anything in the serialization code. I'm unsure about the implications for compatibility, any pointers on that would be helpful. Also, if we agree to take this route, I need to make sure that this works for Kotlin/Native as well.

madsager:
@udalov @dnpetrov Could you have a look at this draft PR and let me know what you think. Would you be OK with the proposed IR changes here?

udalov:
Personally I think it looks good and it's better than having it hacked via an intrinsic. But I'll also ask colleagues from the team to see if there are objections and will write back to you.

Regarding
> there is no IR support for reassigning value parameters which makes it impossible to generate bytecode patterns the inliner understands

Just so that we understand why we need this change, could you please explain what exactly doesn't work with using the `ReassignParameter` intrinsic for the Compose team? E.g. similar to `JvmDefaultArgumentStubGenerator`, the plugin could produce calls to `JvmSymbols.reassignParameterIntrinsic` with proper arguments, and this should be handled automatically later in the codegen.

madsager:
> Personally I think it looks good and it's better than having it hacked via an intrinsic. But I'll also ask colleagues from the team to see if there are objections and will write back to you.
> 
> Regarding
> 
> > there is no IR support for reassigning value parameters which makes it impossible to generate bytecode patterns the inliner understands
> 
> Just so that we understand why we need this change, could you please explain what exactly doesn't work with using the `ReassignParameter` intrinsic for the Compose team? E.g. similar to `JvmDefaultArgumentStubGenerator`, the plugin could produce calls to `JvmSymbols.reassignParameterIntrinsic` with proper arguments, and this should be handled automatically later in the codegen.

I should have been more clear, sorry. You are right that if we were to make the compose plugin aware of the target it is compiling for, that would work. Currently, the compose plugin is run before we have a JVM codegen context, so there is no way to get to the JVM-specific intrinsic. Changing that would be another way to get there: make sure that compose has access to a target specific codegen context and then have separate code for the JVM backend in Compose as well as the compiler.

udalov:
Thanks, I missed that we have no access to `JvmSymbols` in plugins.

But since `IrIntrinsicMethods` works by transforming all methods to their FQ name-based representation (see `toKey`), the Compose plugin could probably just basically copy the `JvmSymbols.reassignParameterIntrinsic` declaration, and generate calls to it. The fact that there will be two `IrSimpleFunction`s now that represent the same function won't matter because the only way they're meaningfully used in JVM IR is when being mapped to an FQ name-based intrinsic key. Could you confirm that this logic works and check if it's applicable for Compose? :)

madsager:
> Thanks, I missed that we have no access to `JvmSymbols` in plugins.
> 
> But since `IrIntrinsicMethods` works by transforming all methods to their FQ name-based representation (see `toKey`), the Compose plugin could probably just basically copy the `JvmSymbols.reassignParameterIntrinsic` declaration, and generate calls to it. The fact that there will be two `IrSimpleFunction`s now that represent the same function won't matter because the only way they're meaningfully used in JVM IR is when being mapped to an FQ name-based intrinsic key. Could you confirm that this logic works and check if it's applicable for Compose? :)

That makes sense and that would work (with the minor change to the inliner to check for default lambdas always).

The only issue that I see with that is that the compose code generation would become JVM specific as this intrinsic only exists in the IR backend. I wonder if the plugin infrastructure allows easy checks of target language (jvm, js, native)?

madsager:
@udalov I believe this is ready now. 

Please note that it requires a corresponding change in kotlin-native which I have uploaded here: https://github.com/JetBrains/kotlin-native/pull/4369. Once we are happy with this change, we need to get that reviewed as well so we can land them together.

madsager:
Added validation that only the default value stub parameters are assignable.

madsager:
Rebased since there were conflicting changes.

compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/lower/EnumClassLowering.kt
```diff
@@ -171,6 +171,14 @@ private fun JsCommonBackendContext.fixReferencesToConstructorParameters(irClass:
 
             return super.visitGetValue(expression)
         }
+
+        override fun visitSetValue(expression: IrSetValue): IrExpression {
+            expression.transformChildrenVoid()
+            mapping.enumConstructorOldToNewValueParameters[expression.symbol.owner]?.let {
+                return builder.irSet(it.symbol, expression.value)
+            }
+            return super.visitSetValue(expression)
```
udalov, line 180: Looks like this would result in a second transformChildren call, potentially leading to exponential complexity

madsager: Thank you! Fixed.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrInlineCodegen.kt
udalov: (This was extracted and merged separately in #3707.)

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt
```diff
@@ -174,8 +174,24 @@ open class DefaultArgumentStubGenerator(
...
+        //     IAND
+        //     IFEQ Lx
+        //     -- any code inserted here is removed if the call site specifies the parameter
+        //     STORE <n>
+        //     -- no jumps here
+        //   Lx
+        //
+        // This control flow limits us to an if-then (without an else), and this together with the
+        // restriction on loading the parameter in the default case means we cannot create any temporaries.
+        +irIfThen(irNotEquals(defaultFlag, irInt(0)), irSet(parameter.symbol, default))
```
udalov, line 193: It's interesting that adding `irNotEquals` here doesn't (seemingly) change the generated JVM bytecode (which is good)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrValueParameter.kt
```diff
@@ -38,6 +38,11 @@ abstract class IrValueParameter : IrValueDeclaration(), IrSymbolDeclaration<IrVa
     // Once we are able to load any top-level declaration from klib this hack should be deprecated and removed.
     abstract val isHidden: Boolean
 
+    // Parameters are not assignable by default. However, in the IR, assignments to parameters are used
+    // to implement default argument stubs. If the default value is to be used, the value is assigned
+    // to the parameter.
+    abstract val isAssignable: Boolean
```
romanart, line 44: I would suggest hoist this property into `IrValueDeclaration`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/impl/IrSetValueImpl.kt
```diff
@@ -17,23 +17,23 @@
...
+class IrSetValueImpl(
     override val startOffset: Int,
     override val endOffset: Int,
     override val type: IrType,
-    override val symbol: IrVariableSymbol,
+    override val symbol: IrValueSymbol,
     override var value: IrExpression,
     override val origin: IrStatementOrigin?
-) : IrSetVariable() {
+) : IrSetValue() {
```
romanart, line 34: Could you please add an assert that checks whether value is assignable

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeepCopyIrTreeWithSymbols.kt
```diff
@@ -348,7 +348,8 @@ open class DeepCopyIrTreeWithSymbols(
             declaration.type.remapType(),
             declaration.varargElementType?.remapType(),
             declaration.isCrossinline,
-            declaration.isNoinline
+            declaration.isNoinline,
+            false
```
romanart, line 352: Shouldn't it be `declaration.isAssignable`?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/RenderIrElement.kt
```diff
@@ -510,7 +510,7 @@ class RenderIrElementVisitor(private val normalizeNames: Boolean = false) : IrEl
         renderFlagsList(
             "vararg".takeIf { varargElementType != null },
             "crossinline".takeIf { isCrossinline },
-            "noinline".takeIf { isNoinline }
+            "noinline".takeIf { isNoinline },
```
romanart, line 513: Seems like it would be good adding `isAssignagle` into render too

compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/IrFileDeserializer.kt
```diff
@@ -1009,7 +1009,8 @@ abstract class IrFileDeserializer(
                 if (proto.hasVarargElementType()) deserializeIrType(proto.varargElementType) else null,
                 flags.isCrossInline,
                 flags.isNoInline,
-                flags.isHidden
+                flags.isHidden,
+                false
```
romanart, line 1013: Shouldn't we put that flag into library too? I mean what if some plugin decides to use this?

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
```diff
@@ -227,7 +226,7 @@ internal fun IrFactory.buildValueParameter(builder: IrValueParameterBuilder, par
         return createValueParameter(
             startOffset, endOffset, origin,
             IrValueParameterSymbolImpl(wrappedDescriptor),
-            name, index, type, varargElementType, isCrossInline, isNoinline, isHidden
+            name, index, type, varargElementType, isCrossInline, isNoinline, isHidden, false
```
romanart, line 229: I would also suggest to support new flag in builder too.

romanart:
BTW, is there any plan to include this changes into 1.4.20?

romanart:
@madsager As you mentioned before it also worth to fix Kotlin/Native too. If you need any help with that feel free to reach me in slack.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrValueParameter.kt
madsager: We decided to not do that. The reason is that we cannot currently check that we only perform assignments to variables marked as `isVar`. Therefore, we can use `isVar` but then there is no benefit as nothing will be checked. For `IrVariable`, `isVar` is used for other things during codegen; but there is no checking that only `isVar` variables are assigned.

So, we ended up here, so that we could introduce the check that parameters are only assigned if they are explicitly marked as assignable.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/impl/IrSetValueImpl.kt
madsager: Done.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeepCopyIrTreeWithSymbols.kt
madsager: Yes, good catch. Done.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/RenderIrElement.kt
madsager: Done.

compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/IrFileDeserializer.kt
madsager: Yeah, we probably should. At this point it should never end up there, but I see no reason to not add support for it. Done.

compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt
madsager: Yes, good idea. Done.

madsager:
> BTW, is there any plan to include this changes into 1.4.20?

That would be nice since we would like to use this in Compose for custom handling of default arguments in Compose.

madsager:
> @madsager As you mentioned before it also worth to fix Kotlin/Native too. If you need any help with that feel free to reach me in slack.

I have that over here: https://github.com/JetBrains/kotlin-native/pull/4369

I'll update that one now and mark it ready for review when done. Thanks for the review Roman!

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrValueParameter.kt
madsager: Come to think of it we already do check via 'is' checks on the IrValueDeclaration, so we could use `isVar` and hoist it to `IrValueDeclaration` and then only check that a value is `isVar` on assignment if it is a value parameter. The only remaining issue that I have is that value parameters cannot be `var` and therefore I like the `isAssignable` bit on value parameters better. But we *could* use `isVar` if that is preferable. 

Thoughts: @udalov @romanart ?

romanart: I have a couple of thinks here:
 - I would like that we have only one flag which says whether value is assignable or mutable.
 - It should be checked likely without explicit cast in `IrSetValue*` 
 - Probably `isVar` isn't the best name for that flag.

madsager: Unfortunately, we cannot have all three of those. We can have one flag called `isAssignable`, but then we need the type checks and casts in `IrSetValue`. In the JVM_IR backend, we do assign to variables that have `isVar=false` and we have to in order to generate code that can be inlined by the bytecode inliner I believe. So, if we unify on one flag, we need the type checking and casts. Currently, there are no checks that we only use IrSetVariable with variables that have `isVar=true` and the reason is that we actually do have to do that. :-\

madsager: As discussed offline, I have moved `isAssignable` to `IrValueDeclaration` and have it always be true for `IrVariable` so that we can unconditionally assert in `IrSetValueImpl` that the declaration is assignable.

udalov:
Merged to master (1.4.30). Will look into merging to 1.4.20 soon. Thank you!

PR #3693, opened 2020-09-04
FIR: rework callable reference adaptation

jsjeon:
The motivation was bb test `.../equality/varargAsArrayWithDefaults`:
```kt
fun target(s1: String, vararg xs: Int, s2: String = "") {}

fun capture2(fn: (String, Int, Int) -> Unit): Any = fn

capture2(::target)
```
Then, I realized that the work I did at https://github.com/JetBrains/kotlin/pull/3614 assumed `vararg` is the last parameter (of course not!). Also, the accumulated logic had a couple more wrong parts: up to the number of expected parameter, value parameters from the candidate were packed, without any further (type) checks. Actually, we don't know whether the given candidate is adaptable before trying all possible configurations of parameters with default values and vararg.

Thus, I completely rewrote the logic in a natural, recursive way. At a high-level, while iterating both of candidate's value parameters and expected parameter types,
1) if the value parameter is `vararg`,
explore all cases: `vararg` as an array, spread it any arbitrary number of times, or no element passing
2) for normal parameter,
try type-matched value parameter; if the parameter has a default value, try using it


compiler/testData/ir/irText/expressions/callableReferences/funWithDefaultParametersAsKCallableStar.fir.txt
```diff
@@ -62,7 +62,7 @@ FILE fqName:<root> fileName:/funWithDefaultParametersAsKCallableStar.kt
   FUN name:testRegularAndDefaultsStar visibility:public modality:FINAL <> () returnType:kotlin.Unit
     BLOCK_BODY
       CALL 'public final fun useKCallableStar (fn: kotlin.reflect.KCallable<*>): kotlin.Unit declared in <root>' type=kotlin.Unit origin=null
-        fn: FUNCTION_REFERENCE 'public final fun regularAndDefaults (x1: kotlin.String, x2: kotlin.String): kotlin.Int declared in <root>' type=kotlin.reflect.KFunction1<kotlin.String, kotlin.Int> origin=null reflectionTarget=<same>
+        fn: FUNCTION_REFERENCE 'public final fun regularAndDefaults (x1: kotlin.String, x2: kotlin.String): kotlin.Int declared in <root>' type=kotlin.reflect.KFunction2<kotlin.String, kotlin.String, kotlin.Int> origin=null reflectionTarget=<same>
```
mglukhikh, line 65: This change is good, but please pay attention to other function references in this file. They differ from FE 1.0 version in arity.

mglukhikh (commented):
Please look at `funWithDefaultParametersAsKCallableStar.kt`. May be it's reasonable to make the whole file working similar to FE 1.0? Or it looks too difficult?

compiler/testData/ir/irText/expressions/callableReferences/funWithDefaultParametersAsKCallableStar.fir.txt
jsjeon: `FIR_IDENTICAL` now!

jsjeon:
> Please look at `funWithDefaultParametersAsKCallableStar.kt`. May be it's reasonable to make the whole file working similar to FE 1.0? Or it looks too difficult?

Thank you for pushing this further. It turns out that adapted callable reference is prohibited against reflective type: https://github.com/JetBrains/kotlin/commit/0c79de1a98d9689442f3ad8eaacfa79fa6961591#diff-06f0225b88a210e80ee17ad51c2668ab. Implemented at the 2nd commit: https://github.com/JetBrains/kotlin/pull/3693/commits/63570f6b2a6efa409d6621c5ea52fc57d87be304

mglukhikh (approved):

jsjeon:
Looks like it had a (self) conflict with https://github.com/JetBrains/kotlin/pull/3710/commits/0fcbd3f8efff7ce7ab0b16737d36641aa38ee236#diff-79d11049d10c40ac5d1655cd9cafda08. Rebased.

semoro:
Unfortunately, this bug is already fixed by @demiurg906, so we had a clash
So I propose to close this PR as Dmitry's fix covers more cases

PR #3694, opened 2020-09-04
[FIR] Add trueIn contract

socolio:


RoundNutz4 (approved):

.idea/dictionaries/Oleg_Ivanov.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="Oleg.Ivanov">
+    <words>
+      <w>dont</w>
```
Enteerman, line 4: why is this here?

socolio: Why not? There is already some dictionaries!

Enteerman: because "dont" is not grammatically correct, you use don't! you don't use dont!

PR #3700, opened 2020-09-07
JVM IR: Optimize static property references (KT-36975)

sfs:
This just generalizes `staticLambdaPhase` to handle generated property reference classes. The bulk of the PR is renaming things to reflect that we now handle arbitrary callable references.


sfs:
Note that there is an ABI difference here between the JVM and JVM IR backends. In the former, the instance field for an unbound property reference class has type `kotlin.reflect.KProperty1` while on the JVM IR backend we use the type of the generated class. I don't think this matters, since the class is synthetic to begin with.

udalov:
Thanks! Looks good, I'll merge after a remote run.

Regarding the ABI difference, I agree, it's not important. By the way, maybe we could now also add an `_ir.txt` test data dump for the test `bytecodeListing/noReceiverInCallableReferenceClasses.kt`, now that the only real problem there is fixed? Looks like all remaining differences are either in synthetic signatures, or stuff where JVM_IR behaves better (inner classes & enclosing method attributes).

sfs:
> By the way, maybe we could now also add an _ir.txt test data dump for the test bytecodeListing/noReceiverInCallableReferenceClasses.kt, now that the only real problem there is fixed? Looks like all remaining differences are either in synthetic signatures, or stuff where JVM_IR behaves better (inner classes & enclosing method attributes).

Good point, I missed the `noReceiverInCallableReferenceClasses` test. I've added the test expectation for the IR backend.

udalov:
Thank you!

PR #3702, opened 2020-09-08
KT-40861 "Convert to secondary constructor" intention expected on class name

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40861

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPrimaryConstructorToSecondaryIntention.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPrimaryConstructorToSecondaryIntention.kt
@@ -33,11 +33,12 @@ class ConvertPrimaryConstructorToSecondaryIntention : SelfTargetingRangeIntentio
     KotlinBundle.lazyMessage("convert.to.secondary.constructor")
 ) {
     override fun applicabilityRange(element: KtClass): TextRange? {
-        val nameIdentifier = element.nameIdentifier ?: return null
         val primaryCtor = element.primaryConstructor ?: return null
+        val startOffset =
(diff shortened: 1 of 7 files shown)


idea/src/org/jetbrains/kotlin/idea/intentions/ConvertPrimaryConstructorToSecondaryIntention.kt
```diff
@@ -21,21 +22,22 @@ import org.jetbrains.kotlin.psi.*
...
 ) {
-    override fun isApplicableTo(element: KtPrimaryConstructor, caretOffset: Int): Boolean {
-        val containingClass = element.containingClassOrObject as? KtClass ?: return false
-        if (containingClass.isAnnotation() || containingClass.isData()
-            || containingClass.superTypeListEntries.any { it is KtDelegatedSuperTypeEntry }
-        ) return false
-        return element.valueParameters.all { !it.hasValOrVar() || (it.name != null && it.annotationEntries.isEmpty()) }
+    override fun applicabilityRange(element: KtClass): TextRange? {
+        val nameIdentifier = element.nameIdentifier ?: return null
+        val primaryCtor = element.primaryConstructor ?: return null
```
dimonchik0036, line 37:
```suggestion
        val primaryCtor = element.primaryConstructor ?: return null
        val nameIdentifier = primaryCtor.getConstructorKeyword() ?: element.nameIdentifier ?: return null
```

dimonchik0036 (changes requested):
The intention shouldn't be available on the class name if the constructor keyword exists
```kotlin
class Fo<caret>o constructor()
```

dimonchik0036 (approved):
Thanks! https://github.com/jetbrains/kotlin/commit/ce407471ec0607415f6abc2b19b94e8889e73162

PR #3703, opened 2020-09-08
KT-41499 "Convert receiver to parameter" produces code with incorrect order of generic type and function invocation  in case of generic function with lambda as a parameter

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41499


dimonchik0036 (approved):
Thanks!

PR #3705, opened 2020-09-08
Introduce plugin for incremental not null verification of Java in our…

AlexeyTsvetkov:
… build

Currently adding runtime not null verification
(for methods/args with NotNull annotation) is done non-incrementally
by calling ant task from IJ not null instrumenter in doLast callback of
every JavaCompile task (see gradle/javaInstrumentation.gradle.kts).

That way the instrumentation breaks Gradle Java IC:
it adds additional `checkNotNull<N>` methods into every class
that was not recompiled by the IC, which triggers Gradle Java IC
to consider a lot of files as affected by changes.
According to our benchmarks, this may result in
incremental builds being 2-3 times slower than clean builds.

In order to fix it, this commit introduces a new internal plugin,
which compares JavaCompile's output classes before and after
compilation. Then it performs the instrumentation only on changed
(recompiled) classes.

After bootstrapping the plugin should be used instead of
gradle/javaInstrumentation.gradle.kts script.

    #KTI-129

### DIFF
--- a/libraries/tools/kotlin-project-build-internal-helper-plugin/build.gradle.kts
+++ b/libraries/tools/kotlin-project-build-internal-helper-plugin/build.gradle.kts
@@ -1,0 +1,41 @@
+
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import org.jetbrains.kotlin.pill.PillExtension
+
+plugins {
+    java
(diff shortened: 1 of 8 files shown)


gavra0:
Is it possible to implement this in `buildSrc`?

PR #3707, opened 2020-09-08
[JVM_IR] Always look for default lambdas in the inliner.

madsager:
This allows compiler plugins such as Compose to do custom
default argument handling in methods that do not have origin
`FUNCTION_FOR_DEFAULT_PARAMETER`.

The code still bails out early is there is not defaults
mask, so there should be no compile-time penalty.


madsager:
@udalov @max-kammerer I extracted this from https://github.com/JetBrains/kotlin/pull/3691. This change is enough to make it possible for Compose to implement custom default argument handling also for inline functions. Therefore, landing this first gives us a path forward even if the larger change takes longer to land.

udalov:
Looks good, thanks!

PR #3709, opened 2020-09-08
Document coroutines codegen

ilmirus:


madsager (commented):
Thank you for creating this documentation Ilmir! :+1: While reading through it I left a number of nit comments.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3266 @@
...
+
+Lastly, most of the examples presented in the document actually suspend, so one is sure every piece is in place since coroutines is a broad
+and complex topic, and it is easy to forget one piece, which will lead to a runtime error or even worse, semantically wrong code execution.
+
+## Suspend Lambda
+Let us begin by introducing suspend lambdas.
+
+Suspend lambda is an example of a coroutine, and the compiler turns ordinary, sequential code inside the lambda into suspendable.
+The example shows a simple suspend lambda:
+```kotlin
```
KvanTTT, line 39: [MD031 - Fenced code blocks should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines) here and below.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3266 @@
...
+and when it requires the typical definition, it says explicitly "a coroutine in a broad sense."
+
+The document often uses the term "undefined behavior," which means that we consciously rejected defining it. Thus, the behavior may vary
+from version to version, from back-end to back-end, and one should use it with extreme caution.
+
+Lastly, most of the examples presented in the document actually suspend, so one is sure every piece is in place since coroutines is a broad
+and complex topic, and it is easy to forget one piece, which will lead to a runtime error or even worse, semantically wrong code execution.
+
+## Suspend Lambda
+Let us begin by introducing suspend lambdas.
```
KvanTTT, line 35: [MD022 - Headings should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines) here and below.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3266 @@
...
+```kotlin
+fun main() {
+    val local: suspend (Int) -> Unit = {}
+}
+```
+Doing so allows us to reuse the logic of captured variables and simplify the logic of code generation. However, because of the limitations
+of old BE, its `create` and `invoke` are unerased. In other words, the compiler duplicates them, generating unerased and erased copies.
+Unerased copy accepts typed parameters, and it contains the logic of lambda's `invoke` or `create`. The other accepts only `Any?`
+parameters since they override supertype's functions, and delegates call to an unerased copy. Thus, the layout of local suspend functions is:
+1. supertypes: `kotlin/coroutines/jvm/internal/SuspendLambda` and `kotlin/jvm/functions/Function{N}`
```
KvanTTT, line 1600: [MD032 - Lists should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md032---lists-should-be-surrounded-by-blank-lines) here and below.

KvanTTT (commented):
I recommend using [mardownlint](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md) rules of markdown formatting. You can check them [online](https://dlaa.me/markdownlint/) or in VSCode with installed [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) extension.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
ilmirus: Thanks, but I don't think the linter will be much of use. The reason is simple: block codes do not introduce new paragraphs. Mostly, they are in the middle of a paraghraph. Besides, the documentation is not intended to be read raw. IntelliJ shows formatted version when you open it.
The same applies to lists.
Heading should be closer to the next paragraph.

KvanTTT: > Besides, the documentation is not intended to be read raw.

In this way, it does not matter how any source code is written because it's not intended to be read raw 😃 

> Heading should be closer to the next paragraph.

Anyway, there is inconsistency in your source: sometimes headers are bound to the next paragraph, sometimes are not.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3266 @@
...
+   |  +-----------+
+   +--+completion |
+      +-----------+
+```
+That is right; there is only one continuation, generated by the compiler: `main$1`. Moreover, it is passed to `generic` and then to `tx`,
+since these functions are tail-call and do not create a continuation. In `tx`, it is saved to resume it in `main`. When we call `resume` on
+it, its `resumeWith` calls `invokeSuspend` and passes `Dummy` as `$result`. The value will be on the stack at the beginning of the last
+state inside the state-machine. It would appear that the suspend function returning `Unit` (in this case `generic`) returns `Dummy`.
+
+To fix the issue, we generate `POP; GETSTATIC kotlin/Unit.INSTANCE` on the call site, when we are sure that callee returns `Unit`. By the
```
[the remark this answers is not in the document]
madsager: Thanks for the clarification Ilmir. I see what you mean now.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
ilmirus: Yes, I will fix the inconsistency. Thank you for pointing this out!

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3263 @@
...
+```kotlin
+fun main() {
+    val i = 1
+    val lambda: suspend (Int) -> Int = { i + it }
+}
+```
+the generated `create` will look like
+```kotlin
+public fun create(value: Any?, completion: Continuation): Continuation {
+    val resutl = main$lambda$1(this.$i, completion)
```
jsjeon, line 1021: s/resutl/result :)

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3263 @@
...
+    result.I$0 = value as Int
+}
+```
+note that the constructor, in addition to captured parameters, accepts a completion object.
+
+In Old JVM BE, `create` is generated for every suspend lambda even when we do not need it. I.e., even for suspending lambdas with
+more than one parameter. There are only two versions of `createCoroutineUnintercepted`, and there are no other places where we call
+`create` (apart from compiler-generated `invoke`s). Thus, in JVM_IR BE, we fixed the slip-up, and it generates the `create` function only
+for functions with zero on one parameter.
```
jsjeon, line 1030: I guess you meant "zero or one parameter"?

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3263 @@
...
+
+##### Lambda Parameters
+We need to put the suspend lambda arguments into fields since there can be only one argument of `invokeSuspend` - `$result`. The compiler
+moves the lambda body into `invokeSuspend`. Thus, `invokeSuspend` does all the computation. We reuse fields for spilled variables for
+parameters as well. For example, if we have a lambda with type `suspend Int.(Long, Any) -> Unit`, then `I$0` hold value of extension
+receiver,' `J$0` - the first argument, `L$1` - the second one.
```
jsjeon, line 1036: > receiver,'
The last quote looks dangling, i.e., not matched with any other quote mark around here.

jsjeon: > receiver,'

The last quote looks dangling, i.e., not matched with any other quote mark around here.

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3263 @@
...
+
+All restricted generated suspend lambdas extend `RestrictedSuspendLambda`:
+```kotlin
+abstract class RestrictedSuspendLambda(
+    public override val arity: Int,
+    completion: Continuation<Any?>?
+) : RestrictedContinuationImpl(completion), FunctionBase<Any?>, SuspendFunction
+```
+the only difference from `SuspendLambda` is superclass. `SuspendLambda` inherits `ContinuationImpl`, while
+`RestrictedSuspendLambda` inherits `RestrictedSuspendLambda`.
```
jsjeon, line 1285: This supposes to be... `RestrictedSuspendLambda` inherits `RestrictedContinuationImpl`?

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3263 @@
...
+        }
+        else -> {
+            throw IllegalStateException("call to 'resume' before 'invoke' with coroutine")
+        }
+    }
+}
+```
+After we resume the coroutine (its `label` value is `1`), `$result` is `42`, and it will get printed. That is right, a function returning
+`Unit` appers returning non-`Unit`. To fix the issue, we replace returns unit markers with `POP; GETSTATIC kotlin/Unit.INSTANCE` sequence.
+That way, we ignore the value, passed to `resume` the same way as if the was no suspension. By the way, we do the same in `callSuspend` and
```
jsjeon, line 1898: as if _there_ was no suspension?

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3240 @@
...
+dummy()
+----------
+println(2)
+==========
+```
+where function boundaries are represented by `==========` and state boundaries are represented by `----------`. The compiler after splitting
+the function generates the following code (simplified for now):
+```kotlin
+val $result: Any? = null
+when(this.label) {
```
LouisCAD, line 215:
```suggestion
when (this.label) {
```

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3240 @@
...
+```
+if one runs the program, it prints `42`. However, `suspendMe` does not return `42`. It just suspends and returns nothing. By the way,
+`suspendMe`'s continuation has type `Continuation<Int>`, i.e., the compiler moves function's return to type argument of the `Continuation`
+interface I mentioned in the previous section (about continuation-passing style).
+
+The state-machine section touched upon the `$result` variable inside the `invokeSuspend` function. The listing shows the `invokeSuspend`
+function of `a`, but, unlike the previous example, with its signature:
+```kotlin
+fun invokeSuspend($result: Any?): Any? {
+    when(this.label) {
```
LouisCAD, line 421:
```suggestion
    when (this.label) {
```

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3240 @@
...
+(because of `println(result.exceptionOrNull())`). There are a couple of things happening here: inside the generated state machine and inside
+`BaseContinuationImpl`'s `resumeWith`.
+
+First, we change the generated state machine. As explained before, the type of `$result` variable is `Int | COROUTINE_SUSPENDED |
+Result$Failue(Throwable)`, but when we resume, by convention, its type cannot be `COROUTINE_SUSPENDED`. Still, the type is `Int |
+Result$Failure(Throwable)`, which we cannot just pass to `plus`, at least, without a check and `CHECKCAST`. Otherwise, we will get CCE at
+runtime. Thus, we check the `$result` variable and throw the exception if the variable holds it.
+```kotlin
+fun invokeSuspend($result: Any?): Any? {
+    when(this.label) {
```
LouisCAD, line 494:
```suggestion
    when (this.label) {
```

compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/coroutines-codegen.md
```diff
@@ -0,0 +1,3240 @@
...
+INVOKESTATIC InlineMarker.mark(I)V
+INVOKESTATIC InlineMarker.afterInlineCall
+INVOKEVIRTUAL println(Ljava/lang/Object;)V
+GETSTATIC kotlin/Unit.INSTANCE
+ARETURN
+```
+I replaced inlined `println` with a call for clarity. After turning into a state-machine, the code becomes: 
+```kotlin
+fun invokeSuspend($result: Any?): Any? {
+    when(this.label) {
```
LouisCAD, line 1856:
```suggestion
    when (this.label) {
```

PR #3710, opened 2020-09-08
FIR: initial support of suspend conversion on arguments

jsjeon:
The motivation is `coroutines/suspendConversion/onArgument.kt`:
```kt
suspend fun runS(fn: suspend (String) -> String) = fn("O")
val lambda: (String) -> String = { it + "K" }
fun box(): String {
    var test = "Failed"
    builder {
        test = runS(lambda)
    }
    return test
}
```
where `runS`, which expects a suspend function, can still input `lambda` of non-suspend functional type, with support of _suspend conversion_. Note that https://github.com/JetBrains/kotlin/pull/3599 dealt with a similar suspend conversion for callable reference (and part of accumulated logics were reused here).


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
```diff
@@ -183,6 +185,7 @@ class CallAndReferenceGenerator(
         return coneType.toIrType() as IrSimpleType
     }
 
+    // TODO: refactor to share some logic with suspend conversion on arguments (maybe introduce AdapterGenerator?)
```
jsjeon, line 188: @mglukhikh If splitting all adapter-related code into `AdapterGenerator` sounds reasonable, I can do that in a follow-up PR.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
```diff
@@ -219,12 +218,29 @@ fun Candidate.resolvePlainArgumentType(
...
-    val argumentTypeForApplicabilityCheck =
+    var argumentTypeForApplicabilityCheck =
         if (useNullableArgumentType)
             capturedType.withNullability(ConeNullability.NULLABLE, session.typeContext)
         else
             capturedType
 
+    // If the argument is of functional type and the expected type is a suspend function type, we need to do "suspend conversion."
+    // TODO: should refer to LanguageVersionSettings.SuspendConversion
+    // TODO: should prefer another candidate without suspend conversion when ambiguous
```
jsjeon, line 229: @mglukhikh I think we can add `useSuspendConversion` to `Candidate`, somewhat similar to `useSam`, and then filter out if some of candidates have it. If this sounds like a good direction, I can also do this in another follow-up PR, and probably at that time, will consider suspend conversion on arguments and callable reference together.

compiler/testData/diagnostics/tests/suspendConversion/overloadResolutionBySuspendModifier.fir.kt
```diff
@@ -14,6 +14,6 @@ fun test1() {
 
 // candidate without suspend conversions is more specific
 fun test2(f: () -> Int, g: suspend () -> Int) {
-    foo(f)
+    <!AMBIGUITY!>foo<!>(f)
```
mglukhikh, line 17: It's a regression, we should not have ambiguity here

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
mglukhikh: > If splitting all adapter-related code into AdapterGenerator sounds reasonable, I can do that in a follow-up PR.

Yes, it sounds good.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
mglukhikh: Probably yes, this direction looks good.

compiler/testData/diagnostics/tests/suspendConversion/overloadResolutionBySuspendModifier.fir.kt
jsjeon: This is what I meant about `useSuspendConversion` flag inside `Candidate`. As mentioned in the comment
```kt
// candidate without suspend conversions is more specific
```
we should distinguish candidates with and without suspend conversion.

mglukhikh: Are you planning to implement this a bit later?

jsjeon: I locally implemented it, and am able to resolve those ambiguities. Will work on refactoring as well.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
jsjeon: Implemented at the 2nd commit: https://github.com/JetBrains/kotlin/pull/3710/commits/b17f7f51a59a85ef7fd80aec0541c066857a026c

compiler/testData/diagnostics/tests/suspendConversion/overloadResolutionBySuspendModifier.fir.kt
jsjeon: Implemented at the 2nd commit: https://github.com/JetBrains/kotlin/pull/3710/commits/b17f7f51a59a85ef7fd80aec0541c066857a026c where all ambiguity regressions added by the 1st commit are resolved.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
jsjeon: Done at the 3rd commit, which is mostly mechanical migration, plus a couple visibility adjustments and small function/check renaming for consistency.

mglukhikh (approved):

mglukhikh:
Merged, thank you!

PR #3711, opened 2020-09-09
FIR: map arguments for overloading indexed access operator

jsjeon:
Users can be overloading [Indexed access operator](https://kotlinlang.org/docs/reference/operator-overloading.html#indexed), and in that case, for `setter`, the last parameter is supposed to be the updated value. By wrapping that one as a named argument, we can handle any other parameters in the middle, which could have default values.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
@@ -6,11 +6,14 @@
 package org.jetbrains.kotlin.fir.resolve.calls
 
 import org.jetbrains.kotlin.fir.declarations.FirFunction
+import org.jetbrains.kotlin.fir.declarations.FirSimpleFunction
 import org.jetbrains.kotlin.fir.declarations.FirValueParameter
+import org.jetbrains.kotlin.fir.declarations.isOperator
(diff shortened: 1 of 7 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ResolutionStages.kt
```diff
@@ -153,8 +155,28 @@ internal object MapArguments : ResolutionStage() {
     override suspend fun check(candidate: Candidate, callInfo: CallInfo, sink: CheckerSink, context: ResolutionContext) {
         val symbol = candidate.symbol as? FirFunctionSymbol<*> ?: return sink.reportDiagnostic(HiddenCandidate)
         val function = symbol.fir
+        val simpleFunction = function as? FirSimpleFunction
+
+        var arguments = callInfo.arguments
+        // If this is an overloading indexed access operator, it could have default values in the middle.
+        // For proper argument mapping, wrap the last one, which is supposed to be the updated value, as a named argument.
+        if (simpleFunction?.isOperator == true &&
```
mglukhikh, line 163: Good shot! Please, consider moving this logic into `mapArguments`. Not quite sure but think it could be more logical...

jsjeon: No problem! Moved to `mapArguments`.

mglukhikh (approved):

mglukhikh:
Merged, thank you!

PR #3713, opened 2020-09-10
KT-41298 "Remove redundant 'with' call" intention works incorrectly with non-local returns and single-expression functions

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41298


idea/src/org/jetbrains/kotlin/idea/inspections/RedundantWithInspection.kt
```diff
@@ -86,7 +93,13 @@ private class RemoveRedundantWithFix : LocalQuickFix {
...
         val callExpression = descriptor.psiElement.parent as? KtCallExpression ?: return
         val lambdaBody = callExpression.valueArguments.getOrNull(1)?.lambdaExpression()?.bodyExpression ?: return
-        val replaced = callExpression.replaced(lambdaBody)
+        val singleReturnedExpression = (lambdaBody.statements.singleOrNull() as? KtReturnExpression)?.returnedExpression
+        val newElement = if (singleReturnedExpression != null && callExpression.isDeclarationInitializer()) {
+            singleReturnedExpression
+        } else {
+            lambdaBody
+        }
+        val replaced = callExpression.replaced(newElement)
```
klunnii, line 102: @t-kameyama would it be also possible to include the case mentioned in `testData/inspectionsLocal/redundantWith/notApplicable_inFunctionBodyWithReturn.kt`?
```
val lambdaExpression = callExpression.valueArguments.getOrNull(1)?.lambdaExpression()
KtPsiUtil.skipSiblingsBackwardByPredicate(callExpression) { element -> element.text != "=" }?.delete()
val replaced = callExpression.replaced(lambdaExpression!!)
```

t-kameyama: Fixed in e9fb229

klunnii:
Thank you! Merged (commits were squashed)

PR #3718, opened 2020-09-11
FIR: extend suspend conversion to subtype of functional type

jsjeon:
As shown in bb test `coroutines/suspendConversion/subtypeOfFunctionalTypeToSuspendConversion`, an argument, which is expected to be of `suspend` functional type, could be a class constructor when that class has overridden `invoke`:
```kt
suspend fun useSuspendFun(fn : suspend () -> String) = fn()

class Test : () -> String, (Int) -> String {
    override fun invoke(): String = ...
}

fun box(): String {
   ...
   useSuspendFun(Test())
}
```

This PR adds necessary changes to resolution (for argument resolution) as well as adapter generation to consider that.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
```diff
@@ -398,32 +399,35 @@ internal class AdapterGenerator(
         }
     }
 
-    // TODO: refactor to share some logic with suspend conversion for callable reference
```
jsjeon, line 401: Because of this addition, now I feel like they diverge...

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
```diff
@@ -448,19 +452,41 @@ internal class AdapterGenerator(
...
+                        .filterIsInstance<FirFunctionSymbol<*>>()
+                        .find { firFunctionSymbol ->
+                            firFunctionSymbol.fir.valueParameters.size == type.arguments.size - 1
+                        }?.let { firFunctionSymbol ->
+                            declarationStorage.getIrFunctionSymbol(firFunctionSymbol) as? IrSimpleFunctionSymbol
+                        }
+                }
+        }
+
+        // Or is a class-like type whose corresponding class is a user-defined and has an overridden `invoke`.
```
dzharkov, line 472: I'm concerned about this part.
AFAIK, suspend conversion works only on a value of a function type `(P1,.., PN) -> R` that is being used as a value for `suspend (P1,.., PN) -> R`, i.e. it shouldn't for an arbitrary class `X` that has an invoke irrelevant to one from function type.
So, I'd check that current `argumentType` is a subtype of function type and then looked for a symbol into a scope instead of declarations because the symbol might be inherited from supertypes, for example

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
```diff
@@ -239,27 +242,63 @@ fun Candidate.resolvePlainArgumentType(
...
+        val typeParameters = argumentType.typeArguments.map { it as ConeKotlinType }
+        return createFunctionalType(
+            typeParameters.dropLast(1), null, typeParameters.last(),
+            isSuspend = true,
+            isKFunctionType = argumentType.isKFunctionType(session)
+        )
+    }
+
+    // Or is a class-like type whose corresponding class is a user-defined and has an overridden `invoke`.
+    if (argumentType is ConeClassLikeType) {
```
dzharkov, line 278: This part looks very similar to the second part in `findInvokeSymbol`.
Please, try extracting such things into utility methods for better reusing.

dzharkov (changes requested):

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
jsjeon: I should have paid attention to the test name, which itself implies that subtype of functional type is applied to suspend conversion. :o It looks like a user class can implement multiple functional types, so the expected number of type parameter should be accounted as well.

The 2nd commit, https://github.com/JetBrains/kotlin/pull/3718/commits/ebe92c7a6457d21156dca30916b96d1dd6e4e066, addressed the comment, including the use of `scope`.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
jsjeon: Utilized at the 2nd commit: https://github.com/JetBrains/kotlin/pull/3718/commits/ebe92c7a6457d21156dca30916b96d1dd6e4e066

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
```diff
@@ -448,19 +452,38 @@ internal class AdapterGenerator(
...
-     * At the use site, instead of the argument, we can put the suspend lambda as an adapter.
-     */
-    private fun needSuspendConversion(argument: FirExpression, parameter: FirValueParameter): Boolean =
-        // TODO: should refer to LanguageVersionSettings.SuspendConversion
-        parameter.returnTypeRef.coneType.isSuspendFunctionType(session) &&
-                argument.typeRef.coneType.isBuiltinFunctionalType(session) &&
-                !argument.typeRef.coneType.isSuspendFunctionType(session)
+    private fun findInvokeSymbol(type: IrSimpleType, argument: FirExpression): IrSimpleFunctionSymbol? {
+        val argumentType = argument.typeRef.coneType
+        // Either the argument type is a non-suspend functional type
```
dzharkov, line 457: Looks like these two cases might be unified: we just need to check that the value is a subtype of FunctionN, but not a SuspendFunctionN and then find invoke in its scope (it should work for both cases as far as I can see)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
```diff
@@ -239,25 +236,59 @@ fun Candidate.resolvePlainArgumentType(
...
-    if (expectedType?.isSuspendFunctionType(session) == true &&
-        argumentTypeForApplicabilityCheck.isBuiltinFunctionalType(session) &&
-        !argumentTypeForApplicabilityCheck.isSuspendFunctionType(session)
+    // Expect the expected type to be a suspend functional type.
+    if (!expectedType.isSuspendFunctionType(session)) {
+        return null
+    }
+
+    // Either the argument type is a non-suspend functional type
+    if (argumentType.isBuiltinFunctionalType(session) && !argumentType.isSuspendFunctionType(session)) {
```
dzharkov, line 267: The same logic about unification may be applied here as well

dzharkov (changes requested):

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
```diff
@@ -45,6 +47,30 @@ fun ConeKotlinType.isKFunctionType(session: FirSession): Boolean {
             kind == FunctionClassKind.KSuspendFunction
 }
 
+fun ConeClassLikeType.isSubtypeOfFunctionalType(session: FirSession, expectedTypeParameterCount: Int? = null): Boolean {
+    val fullyExpandedType = fullyExpandedType(session)
+    val klass = fullyExpandedType.lookupTag.toSymbol(session)?.fir as? FirClass<*> ?: return false
+    return klass.superTypeRefs.any {
```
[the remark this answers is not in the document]
jsjeon: That's right, thanks! Extended the bb test a little bit to cover that case.

Used `AbstractTypeChecker.isSubtypeOf(...)`, but note that it doesn't handle generic types well, e.g.,

`isSubtypeOf(..., Function1<Int, String>, Function1<T, String>) == false`, which seems to me a subtype relation (since type variable `T` could be `Int`). If I dig all the way down to where it returns `false`, type variance of `T` is `in`, and then it chose to check the subtype of 1st type argument the other way around, like `isSubtypeOf(T, Int)`. If this is an expected behavior, fair enough. It simply means we still need to check if the argument type is a functional type. That also bothers unification below; see my other comments.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
```diff
@@ -448,19 +452,38 @@ internal class AdapterGenerator(
...
+                        }
+                }
+        }
+
+        // Or is a subtype of functional type, along with user-defined overridden `invoke`.
+        if (argumentType is ConeClassLikeType &&
+            argumentType.isSubtypeOfFunctionalType(session, type.arguments.size)
+        ) {
+            val klass = argumentType.fullyExpandedType(session).lookupTag.toSymbol(session)?.fir as? FirClass<*> ?: return null
+            if (!klass.isFromSource) {
```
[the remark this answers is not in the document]
jsjeon: Sorry, even I didn't remember where it came from. Thanks for spotting this. Removed.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
```diff
@@ -45,6 +47,30 @@ fun ConeKotlinType.isKFunctionType(session: FirSession): Boolean {
...
+}
+
+fun ConeClassLikeType.findInvokeSymbol(
+    session: FirSession,
+    scopeSession: ScopeSession,
+    expectedTypeParameterCount: Int
+): FirFunctionSymbol<*>? {
+    var result : FirFunctionSymbol<*>? = null
+    scope(session, scopeSession)?.processFunctionsByName(Name.identifier("invoke")) { functionSymbol ->
+        if (result == null && functionSymbol.fir.valueParameters.size == expectedTypeParameterCount - 1) {
```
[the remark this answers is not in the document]
jsjeon: Thanks. But, I had hard time to figure out how to use `processOverriddenFunctions(symbol)`. If I understood correctly, that util is finding functions in the given scope, which override the given symbol, and then giving the control back to the processor. So, while looking for an invoke symbol, how can I find functions whose overridden symbol is that invoke symbol?

One way I ended up is to find the declared invoke symbol in the scope first, and starting from there, to process overridden invoke symbol. Will continue the discussion around the edited code.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
jsjeon: I spent several hours to unify these, but two cases---a direct functional type or a subtype of functional type---always conflict to each other. :'( Maybe I'm still incorrectly finding invoke symbol, or maybe there is something more on IR lowering, but according to my trials, different symbols are required for those cases:
1) for a functional type, not sure what 'overridden' invoke symbol even means; it only works when using the declared invoke symbol. Finding it is somewhat strange too: only works when using `firSymbolProvider.getClassDeclaredCallableSymbols`, not through `scope.processFuncitonByName`.
2) for a subtype of functional type, using the overridden symbol by the current implementation or declared symbol by `scope.processFunctionByName` work.

So... I ended up splitting them, with as much intuitive function/variable names as possible.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
jsjeon: I tried this place as well, but no luck. In particular, for the case of direct functional type, building the functional type through an invoke symbol bothers resolving valid suspend conversion cases with generic types. Maybe this is a hint of finding invoke symbol is somewhat wrong? Anyhow, I ended up splitting them again, and this time, for proper type constraint generation, even not using invoke symbol for functional type.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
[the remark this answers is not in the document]
dzharkov: Sorry, I didn't get the where the problem arises
Actually, now your branch has changed drastically since the comment was applied
Let's synchronize our views on the issue.

We've got an argument `argument`  and its `argumentType`
Also, there's a parameter of `parameterType`

We need to check if suspend conversion is needed (and new `expectedType` if it is needed).
The conversion is needed if and only if 
1. `parameterType` is suspend function type `SuspendFunctionN<Args...>` (not a subtype of it, and not kFunctionType). Then we construct `expectedType` as `FunctionN<Args...>`
2. `argumentType` might be a subtype of that `expectedType` and it might _not_ be a subtype of `SuspendFunctionN<Args...>`

The second condition is a bit complicated to check, but as an approximation we can use check subtyping from `argumentType` to `FunctionN<*, *, *>` (all star projections are used instead of real arguments because they  always fit in supertype position)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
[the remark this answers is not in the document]
dzharkov: The idea is the following:
1. Obtain the symbol for FunctionN::invoke `baseInvoke` (just look into its unsubsituted scope)
2. In the scope for `argumentType`, find the `invoke` symbol that is either equal to `baseInvoke`, or `baseInvoke` is contained among the symbols returned from `processOverriddenFunctions(symbol)`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
dzharkov: Answered above.

The idea is to find in the scope the `invoke` that either an override of `Function::invoke` or it is `Function::invoke` itself

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
dzharkov: Maybe my previous comments may help here as well
If not, please do not hesitate to ask again

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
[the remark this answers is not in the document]
jsjeon: So, `replaceArgumentsWithStarProjections` is the magic wand. Now checking `isSubtypeOfFunctionalType` looks really simple and intuitive. Thanks!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/InferenceUtils.kt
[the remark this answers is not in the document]
jsjeon: I think I implemented that way, but still not sure what `symbol` you meant in `processOverriddenFunctions(symbol)`. If I use `baseInvoke` as a symbol, nothing visited. When I searched for a contributed `invoke` first in the scope of `argumentType` and then visited overridden functions from it, now I can visit functions (and among them, I can find a function whose `overriddenSymbol` is the base invoke).

Renamed it a bit to match with here: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/ArgumentsGenerationUtils.kt#L427

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AdapterGenerator.kt
jsjeon: Two tests that conflict are the one I unmuted in this PR and the other one I unmuted in the previous PR
* https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/coroutines/suspendConversion/onArgument.kt
* https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/coroutines/suspendConversion/subtypeOfFunctionalTypeToSuspendConversion.kt

For `onArgument`, neither an override of `Function::invoke` or `Function::invoke` in the scope of argument type (`Function1<String, String>`) works. It only works when the type-unsubstituted, base `invoke` symbol (`Function1<P1, R>`) is used. Note that the base invoke symbol won't work for the other use-site: type constraint building for arguments, because the type argument (`String`) is needed as constraints.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Arguments.kt
jsjeon: Finding base invoke from unsubstituted scope and finding contributed invoke from the scope of argument type help me figure out the difference between two. Here, as constraints, we need type substituted one, and I finally managed to unify the logic.

dzharkov:
I've merged your changes manually.
Thanks for the contribution!

jsjeon:
Thank you for the thorough reviews!

PR #3719, opened 2020-09-11
[JVM IR] Do not add collection stubs on specialized implementations

kandersen:
Returning user provided types from `iterator`, `listIterator` and
`sublist` should prevent the compiler from adding corresponding
collection stubs, avoiding a JVM Platform declaration clash.


dnpetrov:
I see that your changes are all in JVM_IR, but many of the tests have `// IGNORE_BACKEND: JVM_IR` directive.
This is Ok from documenting perspective, but it'd be better to 1) file remaining issues as YouTrack tickets 2) provide JVM_IR-specific golden data files (`foo_ir.txt` for `foo.kt`), so that changes in JVM_IR behavior would be observable.

dnpetrov:
Superseded by https://github.com/JetBrains/kotlin/commit/c03573fc18c21c044a0845d5c4eeabbc71c3cc7e

PR #3720, opened 2020-09-11
FIR2IR: pick return target based on matched IrFunction

jsjeon:
The motivation is bb test `coroutines/nonLocalReturnFromInlineLambda`:
```kt
inline fun foo(x: (Int) -> Unit) {
    for (i in 1..2) {
        x(i)
    }
}

fun box(): String {
    var result = ""

    val controllerResult = builder {
        result += "-"
        foo {
            result += suspendHere(it).toString()
            if (it == 2) return@builder 56
        }
        // Should be unreachable
        result += "+"
        1
    }.cResult

   if (result != "-24") return "fail 1: $result"
   ...
}
```
`result` was `-24+`, which means we hit the line:
```kt
// Should be unreachable
result += "+"
```
which implies `return@builder` didn't return from `builder`. It turns out that the return target was computed based on "name", which will never work for anonymous lambdas. An exact matching-based lookup for return target resolves it (3rd commit).

------

I initially investigated how `foo` is converted, and found two small discrepancy about `IrOrigin`. First two commits are sort-of by-product.


mglukhikh:
Merged manually. I had to modify your first commit in the following way: https://github.com/JetBrains/kotlin/commit/578e707ab9251021a90b4c7fb76200b18768ec2e, because original commit often set `VARIABLE_AS_FUNCTION` origin erroneously (and even one test failed due to this reason).

jsjeon:
Oops, I thought it over-changed origin, but didn't catch any test failures. I should have updated the master more often. Thank you for the fix!

PR #3721, opened 2020-09-12
FIR: update suspend lambda type during declaration resolution

jsjeon:
The motivation is bb test `coroutine/featureIntersection/suspendFunctionIsAs`, where
```kt
val suspendLambda0: suspend () -> Unit = {}
```
had a casting exception, since the lambda's type remains as non-suspend functional type. During declaration resolution, when visiting anonymous functions and updating their type, we should have considered if its expected type is `suspend` or not.


mglukhikh:
Merged, thank you! 👍

workfilemmfloe (approved):
Merging

PR #3722, opened 2020-09-13
FIR: support several annotation argument diagnostics

McCreeMainer:
### Added next fir-checks:
- ANNOTATION_ARGUMENT_KCLASS_LITERAL_OF_TYPE_PARAMETER_ERROR
- ANNOTATION_ARGUMENT_MUST_BE_CONST
- ANNOTATION_ARGUMENT_MUST_BE_ENUM_CONST
- ANNOTATION_ARGUMENT_MUST_BE_KCLASS_LITERAL
- NON_CONST_VAL_USED_IN_CONSTANT_EXPRESSION (only for annotation argument)

### Problems:
- cannot check initializer of Java fields
- conflicts between fields and functions with the same names declared inside annotated class and declared outside after declaring of this class

Example:
```
annotation class Ann(val a: Int)

@Ann(b)                      // There is 2 possible vals
class SomeClass() {
    val b = 1                // Annotation refer on this 
}

val b = 2                    // But must on this
```


mglukhikh:
Thank you, I will check all problems a bit later.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationArgumentChecker.kt
```diff
@@ -0,0 +1,246 @@
...
+import org.jetbrains.kotlin.fir.expressions.*
+import org.jetbrains.kotlin.fir.types.*
+import org.jetbrains.kotlin.psi.KtExpression
+
+object FirAnnotationArgumentChecker : FirBasicDeclarationChecker() {
+    override fun check(declaration: FirDeclaration, context: CheckerContext, reporter: DiagnosticReporter) {
+        val annotationList = (declaration as FirMemberDeclaration).annotations
+            .plus(
+                ((declaration as? FirProperty)
+                    ?.getter
```
mglukhikh, line 30: This way to handle getter annotations is too hacky; moreover, there are also setters which are unaffected by this hack. It's much easier to add your checker to `basicDeclarationCheckers` list, not too `memberDeclarationsCheckers`, and your checker will see property and both its accessors separately.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationArgumentChecker.kt
```diff
@@ -0,0 +1,246 @@
...
+    private fun checkAnnotationArgumentWithSubElements(
+        expression: FirExpression,
+        session: FirSession,
+        reporter: DiagnosticReporter
+    ): FirDiagnosticFactory0<FirSourceElement, KtExpression>? {
+        when (expression) {
+            is FirArrayOfCall -> {
+                var usedNonConst = false
+
+                for (arg in (expression as FirCall).argumentList.arguments) {
```
mglukhikh, line 55: Is this cast to `FirCall` necessary here?

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationArgumentChecker.kt
```diff
@@ -0,0 +1,246 @@
...
+                        FirErrors.ANNOTATION_ARGUMENT_MUST_BE_CONST
+                }
+            }
+            else ->
+                return FirErrors.ANNOTATION_ARGUMENT_MUST_BE_CONST
+        }
+        return null
+    }
+
+    private fun FirCallableDeclaration<*>?.getFirSymbol(session: FirSession): FirSymbolOwner<*>? =
```
mglukhikh, line 209: Some name like `getReferencedAnnotationClass` would be much better

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationArgumentChecker.kt
```diff
@@ -0,0 +1,246 @@
...
+            }
+            expression is FirFunctionCall -> {
+                val calleeReference = expression.calleeReference
+                if (calleeReference is FirErrorNamedReference)
+                    return null
+
+                when (calleeReference.name.identifier) {
+                    "toString" ->
+                        return checkAnnotationArgument(expression.dispatchReceiver, session)
+                    in listOf("dec", "div", "inc", "minus", "plus", "rem", "times", "unaryMinus", "unaryPlus", "compareTo") -> {
```
mglukhikh, line 158: `object OperatorNameConventions`  includes some helpful constants and lists.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirAnnotationArgumentChecker.kt
```diff
@@ -0,0 +1,246 @@
...
+                    ?.typeRef
+                    ?.coneType
+
+                if (coneType is ConeClassErrorType)
+                    return FirErrors.ANNOTATION_ARGUMENT_MUST_BE_CONST
+
+                while (coneType?.classId == StandardClassIds.Array)
+                    try {
+                        coneType = coneType.arrayElementType()
+                    } catch (_: ClassCastException) {
```
mglukhikh, line 126: It's much better to fix this exception (will do myself).

mglukhikh (changes requested):

mglukhikh:
Ok, I will probably do some more tests and accept this PR.

mglukhikh (approved):

mglukhikh:
Merged manually, thank you!

PR #3724, opened 2020-09-14
KT-40251 Intention action to evaluate compile time expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-40251


klunnii:
Merged! Thank you!

PR #3725, opened 2020-09-14
[JVM_IR] Generate line numbers in synthetic bridges. 

madsager:


madsager:
@yanex @max-kammerer It doesn't seem useful to put in the line number of a surrounding class for synthetic bridges. This rebases a test that assumes that. Are there things in the debugger that rely on line numbers in synthetic bridges?

udalov:
Hey Mads, I just found out that IntelliJ has "Skip synthetic methods" option, true by default, in Preferences | Build, Execution, Deployment | Debugger | Stepping. It can be disabled and I suppose some users are doing that.

Looks like the test checks this option via the `// SKIP_SYNTHETIC_METHODS: false` directive. Another test at `stepping/stepInto/accessors.kt` also checks exactly this option.

I don't know the exact use cases and would be interested to learn about them as well. But at the moment it looks like a regression not to support this in JVM IR.

madsager:
Thanks Alex. I'm not sure how useful these line numbers in synthetic code are, but let's just generate them. The JVM backend is not consistent in the generation though and there are certain bridges that do not have line numbers (bridges for the invoke method for a lambda for instance). Also, the line number used seems a little strange. This change now always uses the offset of the class the bridge is added to which is easy to explain and reason about.

Let me know what you think.

madsager:
Updated to generate line numbers for synthetic accessors as well which makes stepping behavior the same as the old backend for a couple of other stepping tests.

idea/jvm-debugger/jvm-debugger-test/testData/stepping/custom/functionCallStoredToVariable.ir.out
```diff
@@ -0,0 +1,21 @@
+LineBreakpoint created at functionCallStoredToVariable.kt:7
+LineBreakpoint created at functionCallStoredToVariable.kt:14
+LineBreakpoint created at functionCallStoredToVariable.kt:21
+LineBreakpoint created at functionCallStoredToVariable.kt:28
+Run Java
+Connected to the target VM
+functionCallStoredToVariable.kt:7
+functionCallStoredToVariable.kt:4
+functionCallStoredToVariable.kt:14
```
madsager, line 9: This one is interesting. The code is:

```
11: val b = {
12:   // STEP_OVER: 2
13:   // Breakpoint!
14:   1
15: }
```

We are breaking on line 14 and then we take two steps before resuming.

On JVM_IR those two steps are line 11 which is leaving the accessibility bridge, followed by line 43 leaving the caller.

On JVM those two steps are line 43 leaving the caller (this particular bridge has no line numbers) followed by line 11 for the assignment.

So it looks like line 11 and line 43 are reversed, but they are not, there is just an extra line 11 step on JVM_IR because of the consistent line numbers in bridges.

compiler/testData/debug/stepping/namedCallableReference.kt
```diff
@@ -10,10 +10,19 @@ fun f(block: () -> Unit) {
...
+// For some reason, when the bridge does not have line numbers, there is no method entry event
+// for the invoke method bridged to. Therefore, the entry line number for invoke only shows
+// up for JVM_IR.
+
 // LINENUMBERS
 // test.kt:3 box
 // test.kt:4 box
 // test.kt:8 f
+// LINENUMBERS JVM_IR
+// test.kt:4 invoke
```
max-kammerer, line 24: @madsager Is this test affected by #3735 or not?

madsager: It is not, no.

The difference in this test comes from the `invoke` bridge that returns an `Object` having line numbers with JVM_IR and not with JVM.

JVM:

```
  public java.lang.Object invoke();
    descriptor: ()Ljava/lang/Object;
    flags: ACC_PUBLIC, ACC_BRIDGE, ACC_SYNTHETIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokevirtual #12                 // Method invoke:()V
         4: getstatic     #18                 // Field kotlin/Unit.INSTANCE:Lkotlin/Unit;
         7: areturn

  public final void invoke();
    descriptor: ()V
    flags: ACC_PUBLIC, ACC_FINAL
    Code:
      stack=0, locals=1, args_size=1
         0: invokestatic  #23                 // Method FlafKt.g:()V
         3: return
      LineNumberTable:
        line 4: 0
      LocalVariableTable:
            0       4     0  this   LFlafKt$box$1;
```

JVM_IR
```
  public final void invoke();
    descriptor: ()V
    flags: ACC_PUBLIC, ACC_FINAL
    Code:
      stack=0, locals=1, args_size=1
         0: invokestatic  #23                 // Method FlafKt.g:()V
         3: return
      LineNumberTable:
        line 4: 0
      LocalVariableTable:
            0       4     0  this   LFlafKt$box$1;

  public java.lang.Object invoke();
    descriptor: ()Ljava/lang/Object;
    flags: ACC_PUBLIC, ACC_BRIDGE, ACC_SYNTHETIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokevirtual #27                 // Method invoke:()V
         4: getstatic     #32                 // Field kotlin/Unit.INSTANCE:Lkotlin/Unit;
         7: areturn
      LineNumberTable:
        line 4: 0
      LocalVariableTable:
            0       8     0  this   LFlafKt$box$1;
```

I must admit that I'm confused as to why the debugging filters do not stop on line 4 in the usual non-bridge `invoke` method for JVM. There must be some kind of filtering going on.

max-kammerer: @yanex @goodwinnk @NataliaUkhorskaya Maybe you now what is going on with filtering?

max-kammerer:
@madsager Thank you for pr!

max-kammerer:
@madsager There are 4 failing test after merging pr, could you look into them?
```
ParcelizeIrBytecodeListingTestGenerated.testCustomParcelablesSameModuleorg.jetbrains.kotlin.parcelize.test
ParcelizeIrBytecodeListingTestGenerated.testCustomSimpleWithNewArrayorg.jetbrains.kotlin.parcelize.test
ParcelizeIrBytecodeListingTestGenerated.testCustomSimpleDeprecatedorg.jetbrains.kotlin.parcelize.test
ParcelizeIrBytecodeListingTestGenerated.testCustomSimpleorg.jetbrains.kotlin.parcelize.test
```

madsager:
I'll take a look, the generated code probably has bridges that now have no line numbers. Thanks for the heads up!

madsager:
Fix here: https://github.com/JetBrains/kotlin/pull/3754

max-kammerer:
@madsager Thank you!

PR #3726, opened 2020-09-14
Android Synthetics: Update to the new plugin API (KT-40018)

sfs:


udalov:
Thank you very much Steven! Merged to master and 1.4.20

PR #3728, opened 2020-09-14
FIR deserializer: load inline/external flags properly

jsjeon:
The motivation is bb test `coroutines/intrinsicSemantics/coroutineContext`, which fails with access error, complaining that the test file is trying to access coroutine context from inaccessible class. It turns out that the context should be inlined: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/coroutines/Continuation.kt#L157 FIR deserializer should load `inline` flags for properties properly. Found `external` flags need similar loading.


mglukhikh:
Merged, thank you!

PR #3729, opened 2020-09-15
KT-41878 False positive `Redundant nullable return type` inspection in case of elvis operator

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41878


vladimirdolzhenko:
LGTM, Merged manually

Thank you for providing fix quite fast

PR #3730, opened 2020-09-15
KT-41897 `Redundant nullable return type` inspection should not be applied to overridden methods

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41897

### DIFF
--- a/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantNullableReturnTypeInspection.kt
+++ b/idea/src/org/jetbrains/kotlin/idea/inspections/RedundantNullableReturnTypeInspection.kt
@@ -20,6 +20,7 @@ import org.jetbrains.kotlin.idea.core.isOverridable
 import org.jetbrains.kotlin.idea.project.languageVersionSettings
 import org.jetbrains.kotlin.idea.resolve.getDataFlowValueFactory
 import org.jetbrains.kotlin.idea.util.textRangeIn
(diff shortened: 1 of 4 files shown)


klunnii:
@t-kameyama issue seems to controversial. 
It's quite normal to narrow return type at override. Should one design the class for extension and mark it as `open` inspection won't be applied (there could be undesired consequences). Please, provide more details/examples for the case(s) you tried to solve.

t-kameyama:
> It's quite normal to narrow return type at override

I think so, but it's also normal not to narrow return type. 
I don't think we should be forced to narrow return type at override.

t-kameyama:
> It's quite normal to narrow return type at override.

I think so, but it's also normal not to narrow return type. I think this inspection should not force to narrow return type at override.

klunnii:
> ... but it's also normal not to narrow return type.

The case in question is about nullability. Why should clients of class C be misleaded and handle `null`? Could you please provide as example where inspection is undesired?

``` kotlin
interface I {
    fun f(): String?
    val v: String?
}

class C : I {
    override fun f(): String?/*caret*/ {
        return ""
    }
    override val v: String?/*caret*/
        get() = ""
}

```

t-kameyama:
I think the above case is undesired.
In general, is it better to remove nullability in overridden methods if nullability is unnecessary?
If yes, please close this PR and the ticket.

klunnii:
Closing

PR #3732, opened 2020-09-15
JVM_IR: move common metadata serialization code to ClassCodegen

pyos:
Part of the forgotten #3405.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -791,7 +791,6 @@ class Fir2IrDeclarationStorage(
                     isExternal = false,
                     isStatic = field.isStatic
                 ).apply {
-                    metadata = FirMetadataSource.Variable(field)
```
pyos, line 794: Not 100% sure about this one, though it looks like it only affects imported Java fields and delegated "synthetic" fields, whatever that means, so it should not have any effect on emitted metadata.

That said, the fact that FIR fields, unlike descriptor fields, are not properties is slightly concerning, because e.g. they currently cannot be referenced. The IR simply doesn't allow it - it *requires* a property symbol for an `IrPropertyReferenceImpl`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/DescriptorMetadataSerializer.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    override fun serialize(metadata: MetadataSource): MessageLite? {
+        val localDelegatedProperties = (irClass.attributeOwnerId as? IrClass)?.let(context.localDelegatedProperties::get)
+        if (localDelegatedProperties != null && localDelegatedProperties.isNotEmpty()) {
+            context.state.bindingTrace.record(
+                CodegenBinding.DELEGATED_PROPERTIES_WITH_METADATA,
+                // When serializing metadata for interfaces, `JvmSerializerExtension.serializeClass`
+                // looks at `$DefaultImpls` for some reason that's probably related to the old backend.
+                if (irClass.isInterface) context.typeMapper.mapClass(context.cachedDeclarations.getDefaultImplsClass(irClass)) else type,
+                localDelegatedProperties.mapNotNull { (it.owner.metadata as? DescriptorMetadataSource.LocalDelegatedProperty)?.descriptor }
```
pyos, line 50: Given that this is descriptor-specific metadata serialization, I actually don't see why someone changed it from simply `it.descriptor` to this. (There was similar code for FIR, but it didn't have any effect because FIR metadata serialization does not use the binding context, and should've been simply removed.)

mglukhikh (approved):

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/FirMetadataSource.kt
```diff
@@ -10,55 +10,24 @@ import org.jetbrains.kotlin.fir.declarations.*
 import org.jetbrains.kotlin.ir.declarations.MetadataSource
 import org.jetbrains.kotlin.name.Name
 
-interface FirMetadataSource : MetadataSource {
-
+sealed class FirMetadataSource(open val fir: FirDeclaration) : MetadataSource {
```
udalov, line 13: `open val` here + `override val` in subclass means there will be two fields with the same value. Maybe make it abstract here (or make it final, and remove `override val` in subclasses)? Same for `DescriptorMetadataSource`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -216,7 +228,55 @@ abstract class ClassCodegen protected constructor(
...
+        val entry = irClass.fileParent.fileEntry
+        val kind = when {
+            facadeClassName != null -> KotlinClassHeader.Kind.MULTIFILE_CLASS_PART
+            metadata is MetadataSource.Class -> KotlinClassHeader.Kind.CLASS
+            metadata is MetadataSource.File -> KotlinClassHeader.Kind.FILE_FACADE
+            metadata is MetadataSource.Function -> KotlinClassHeader.Kind.SYNTHETIC_CLASS
+            entry is MultifileFacadeFileEntry -> KotlinClassHeader.Kind.MULTIFILE_CLASS
+            else -> KotlinClassHeader.Kind.SYNTHETIC_CLASS
+        }
+        writeKotlinMetadata(visitor, state, kind, extraFlags) {
```
udalov, line 251: Note that there's a minor change in behavior in that we'll write the `METADATA_JVM_IR_FLAG`/`METADATA_JVM_IR_STABLE_ABI_FLAG` to synthetic classes too (previously we used `writeSyntheticClassMetadata` which passed extraFlags=0).

But I think the new behavior is more useful, so let's keep it. It doesn't affect anything anyway, other than manual bytecode introspection.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/DescriptorBasedClassCodegen.kt
```diff
@@ -1,144 +0,0 @@
...
-        val localDelegatedProperties = (irClass.attributeOwnerId as? IrClass)?.let(context.localDelegatedProperties::get)
-        if (localDelegatedProperties != null && localDelegatedProperties.isNotEmpty()) {
-            state.bindingTrace.record(
-                CodegenBinding.DELEGATED_PROPERTIES_WITH_METADATA,
-                type,
-                localDelegatedProperties.mapNotNull { (it.owner.metadata as? MetadataSource.LocalDelegatedProperty)?.descriptor }
-            )
-        }
-
-        // TODO: if `-Xmultifile-parts-inherit` is enabled, write the corresponding flag for parts and facades to [Metadata.extraInt].
```
udalov, line 54: This TODO doesn't seem fixed, but disappeared for some reason. Let's return it to `DescriptorMetadataSerializer`

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/FirMetadataSource.kt
pyos: `final` doesn't work because then the subclasses will not have more specific types. `abstract` is indeed the best option here, thanks.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
pyos: Yes, I noticed the 0 but thought that it looked like an oversight.

udalov:
Thank you!

PR #3733, opened 2020-09-15
JVM_IR: implement FIR local delegated property reflection

pyos:
Also, fix a bug in serialization of lambdas' metadata: the "free" in the name of the function means that all referenced type parameters are serialized as well, but this was not done. (The same fix is used for local delegated properties, which are serialized at class level but may refer to function-scope type parameters, and so need their own copies of type parameter metadata.)


mglukhikh:
Merged manually 👍

PR #3734, opened 2020-09-15
FIR2IR: approximate type argument if reified

jsjeon:
The motivation is bb test `coroutines/featureIntersection/callableReference/function/genericCallableReferenceArguments`:
```kt
suspend fun <T> fooReturnLong(x: T): Long = 1L
suspend inline fun <reified T, reified R> check(x: T, y: R, f: suspend (T) -> R, tType: String, rType: String) {
  ...
}
fun box(): String {
   ...
   check("", "", ::fooReturnLong, "String", "Any")
}
```
By the 2nd argument `""` of `String`, the NI's type constraint has: `String <: R` . Also, from the reference (`::fooReturnLong`) to parameter `f`'s return type, it also has: `Long <: R`. From there, fixed type for type variable `R` is `it(Comparable<*>, Serializable)`. At the end, `fir2ir` chose the 1st type among intersection types, hence runtime error: expected `Any` but `Comparable`. Rather, exotic types left in type arguments need to be approximated.


semoro:
It is not correct to just approximate type arguments during the completion
Seems that problem here caused by R is a refined type parameter. The design here is undefined yet, so let's just approximate type arguments in FIR2IR if the corresponding type parameter is reified

jsjeon:
> The design here is undefined yet, so let's just approximate type arguments in FIR2IR if the corresponding type parameter is reified

Thanks for the clarification and suggestion. Rewrote the 2nd commit accordingly.

mglukhikh (approved):

mglukhikh:
Merged manually, thank you!

PR #3735, opened 2020-09-16
[JVM_IR] Fix line numbers for function reference invoke methods.

madsager:


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/FunctionReferenceLowering.kt
```diff
@@ -398,6 +398,7 @@ internal class FunctionReferenceLowering(private val context: JvmBackendContext)
 
             body = context.createJvmIrBuilder(symbol).run {
                 var unboundIndex = 0
+                at(this@createFunctionReferenceInvokeMethod.startOffset, this@createFunctionReferenceInvokeMethod.startOffset)
```
udalov, line 401: Maybe pass them to the `createJvmIrBuilder` call two lines above?

Also, why `startOffset, startOffset` instead of `startOffset, endOffset`?

madsager: Ah, didn't realize you could pass it to `createJvmIrBuilder`, thank you! Fixed to used `startOffset, endOffset`.

madsager: Updated #3725 as well, there we really do want to use the start offset for both start and end (as this is the first line of the class that we want to use for the entire bridge body). I think that is why I ended up using `startOffset,startOffset` in this change as well. Cleaned up now! Thanks!

udalov:
Merged, thanks!

PR #3736, opened 2020-09-16
[JVM_IR] Rebase fwBackingField stepping test.

madsager:
The difference is that JVM_IR generates line numbers for the
constructor field initialization.

In this case:

```
29: class B {
30:   // comment
31:   val bPropVal: Int,
32:   // comment
33:   var bProbVar: Int
34: ) {
```

the line numbers for the field initializations are 31 and 33 where
in the current backend the whole constructor has line number 29.


udalov:
Thanks!

PR #3737, opened 2020-09-17
[JVM_IR] Mark delegated member functions as bridges.

madsager:
This allows them to have line numbers and at the same time have
smart step into work. Smart step into will step through bridges
to the final target which seems appropriate for these delegating
methods.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
@@ -168,7 +168,9 @@ class FunctionCodegen(
         }
 
         val isVararg = valueParameters.lastOrNull()?.varargElementType != null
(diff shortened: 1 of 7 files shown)


madsager:
@udalov proposed fix for: https://youtrack.jetbrains.com/issue/KT-41960

It seems reasonable to me that these synthetic methods are marked as bridges, but what do you think. The alternative is to attempt to avoid generating line numbers for these synthetic methods.

udalov:
To be honest, I think it's too risky. We had some problems in the past with changing the ACC_BRIDGE flag which affected external tools: https://youtrack.jetbrains.com/issue/KT-28684. Also, for reasons of easier ABI comparison, I'd prefer if we kept the bytecode shape equivalent to the old backend.

I suppose we should rather not generate the line number for the delegated members. In KT-41960, you're saying that not having line numbers here breaks expression null checks. Maybe we could patch those instead?

madsager:
That is fair, given that bug I do tend to agree with you. Closing this in favor of:

https://github.com/JetBrains/kotlin/pull/3739

PR #3738, opened 2020-09-17
[JVM_IR] Ignore fwBackingField stepping test.

madsager:
The test relies on fixes not yet landed, so it currently fails.


madsager:
@udalov Sorry about that. This relies on: https://github.com/JetBrains/kotlin/pull/3725. I'll update that change to remove this ignore flag so that the test actually passes.

PR #3742, opened 2020-09-21
JVM_IR: fix lifting of arguments to object super constructors

pyos:
If the arguments are reordered, they can not only be reads of temporary variables, but also implicit type casts of said reads.

Fixes #KT-42002

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlin.ir.declarations.IrVariable
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstructorCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrGetValueImpl
+import org.jetbrains.kotlin.ir.expressions.impl.IrTypeOperatorCallImpl
 import org.jetbrains.kotlin.ir.util.transformInPlace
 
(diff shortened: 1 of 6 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/AnonymousObjectSuperConstructorLowering.kt
```diff
@@ -78,16 +79,19 @@ private class AnonymousObjectSuperConstructorLowering(val context: JvmBackendCon
...
-            for (i in 0 until symbol.owner.valueParameters.size) {
-                val argument = getValueArgument(i) ?: continue
-                val mapped = when (argument) {
-                    is IrConst<*> -> null
-                    is IrGetValue -> remapping[argument.symbol.owner]
-                    else -> addArgument(argument)
-                } ?: continue
-                putValueArgument(i, IrGetValueImpl(argument.startOffset, argument.endOffset, mapped.symbol))
+            for (i in symbol.owner.valueParameters.indices) {
+                putValueArgument(i, getValueArgument(i)?.transform(remapping))
```
max-kammerer, line 94: What means `null` value for `getValueArgument(i)`? Shouldn't it be getValueArgument(i)!!?

pyos: defaultArgumentInjectorPhase is after anonymousObjectSuperConstructorPhase, so it could be null if the super constructor has defaults.

max-kammerer: Yes, they are supported.
...so just reassigning null value that is identical to original case.

max-kammerer:
@pyos Thank you!

PR #3743, opened 2020-09-21
[JVM_IR] Fix offsets in constant propagation optimization.

madsager:
Loads of temporary variables that contain constants are replaced
with a copy of the constant. This avoids locals loads and stores.
However, the copy of the constant needs to have the offset of
the load and not of the original constant.

Fixes KT-41963.


udalov:
Thanks Mads!

PR #3744, opened 2020-09-22
JVM_IR: fix bound suspend-converted references

pyos:
Also, treat adapted function references as lambdas for inlining.

#KT-42021 Fixed
#KT-38536 Fixed
#KT-38535 Fixed

### DIFF
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -2207,6 +2207,11 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
                     runTest("compiler/testData/codegen/box/callableReference/adaptedReferences/suspendConversion/bound.kt");
                 }
 
+                @TestMetadata("boundExtension.kt")
+                public void testBoundExtension() throws Exception {
+                    runTest("compiler/testData/codegen/box/callableReference/adaptedReferences/suspendConversion/boundExtension.kt");
+                }
+
                 @TestMetadata("crossInline.kt")
                 public void testCrossInline() throws Exception {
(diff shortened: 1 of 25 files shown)


max-kammerer:
@pyos Could you add tests for second commit?

pyos:
Done

compiler/testData/codegen/bytecodeText/inline/adaptedReference.kt
```diff
@@ -0,0 +1,11 @@
...
+// FILE: a.kt
+fun nonSuspend() {}
+
+inline suspend fun f(x: suspend () -> Unit) = x()
+
+// FILE: b.kt
+suspend fun call() = f(::nonSuspend)
+
+// @BKt.class:
+// 0 INVOKEINTERFACE kotlin/jvm/functions/Function1.invoke
```
max-kammerer, line 11: Minor. Maybe convert it into boxInline test? It will automatically check that there is no additional anonymous objects

pyos: Unfortunately this check seems to not work properly for tests marked `WITH_COROUTINES` - something inside the injected `helpers` package breaks it.

max-kammerer: Is it possible to get rid of suspend at all, look like they are not neccessery according to https://youtrack.jetbrains.com/issue/KT-38536 issue?

pyos: Oh right, I forgot there was a test already. Still, the check seems very unreliable - for some reason, on that test it only fails in the `JvmIrAgainstOld` configuration, even though `IrBlackBoxInline` contains a lambda class too.

pyos: Never mind, it also triggers in `IrCompileKotlinAgainstInlineKotlin`, I just used the wrong test filter.

max-kammerer:
@pyos Many thanks!

PR #3745, opened 2020-09-22
JVM IR: Allow the debugger to evaluate expression with IR dependencies

sfs:
This PR changes the debugger to ignore `IR_COMPILED_CLASS` errors, in the same way as it is already ignoring errors about invisible references or experimental APIs. This is required to use the evaluate expression functionality in projects compiled with the JVM IR backend.

The bulk of the PR is just about enabling the evaluate expression test suite for the JVM IR backend. Right now, the evaluate expression functionality of the debugger is still using the JVM backend to compile expressions, but outside of a few corner cases this works surprisingly well even for projects which are compiled with the JVM IR backend. Looking at the tests, there are essentially only three problems, relating to local functions, captured variables, and local delegated properties. Fixing these is left for future work and for now I've just muted the corresponding tests.


max-kammerer:
@sfs Please rebase on master, there is a conflict reported by github

sfs:
@max-kammerer Rebased on master and fixed conflicts. Some of the recent work on line numbers in the IR backend ended up fixing an additional test, as well. 👍

max-kammerer:
Thank you!

PR #3746, opened 2020-09-23
FIR: record type aliased nested class to nested classifier scope

jsjeon:
(Excerpted from the internal slack message)

The motivation is bb test `typealias/innerClassTypeAliasConstructor`: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/typealias/innerClassTypeAliasConstructor.kt where the inner class is exposed via type alias. By replacing the alias, `OI`, with the actual inner class name, `Inner`, I can track down what is the expected behavior v.s. what’s happening now.

The outer class instantiation, `Outer(...)`, is set as an explicit receiver, and when the correct constructor of `Inner` is processed at `ConstructorProcessing#processConstructors`, the `processor` is from `MemberScopeTowerLevel`, and that receiver is regarded as a dispatch receiver for inner class construction.

When the alias is used instead, that processing with member scope doesn’t work as expected. From `Outer`’s member scope, it goes down to `FirNestedClassifierScope`, where mapping from nested class name (`Inner`) to its corresponding class symbol doesn’t know about type alias (`OI`) of course. Rather, after the member scope handling, `ScopeTowerLevel` is used, and from the package scope, now `OI` is resolved as type alias symbol. But, when reaching `ConstructorProcessing#processConstructors`, the `processor` is from `ScopeTowerLevel`, which treats the explicit receiver as an extension receiver, hence the candidate is discarded with an inconsistent receiver.

The ideal point to remedy this is when trying to process constructor name `OI` with `Outer`’s use-site scope and `MemberScopeTowerLevel`. To do so, we need to let the nested classifier scope know bound type alias. It won’t be possible to replace such name with resolved, expanded type name any time, because we don’t know in what package that name could be a type alias. It’s also too late if we wait until the right package scope is used to retrieve the matched type alias symbol.

It seems `FirApplySupertypesTransformer#transformTypeAlias` is the first time type alias's expanded type is resolved. Around that time, a nested classifier scope in question is also created/cached during super type resolution. We can directly look up that cached scope and inform it of type-aliased nested class.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSupertypesResolution.kt
```diff
@@ -129,8 +132,22 @@ private class FirApplySupertypesTransformer(
             "Expected single supertypeRefs, but found ${supertypeRefs.size} in ${typeAlias.symbol.classId}"
         }
 
+        val supertypeRef = supertypeRefs.single()
+        // TODO: Need better handling for type alias with type parameters.
+        if (typeAlias.typeParameters.isEmpty()) {
+            (supertypeRef.firClassLike(session) as? FirRegularClass)?.let {
```
dzharkov, line 138: NB: It can be not a regular class, but for example a different type alias

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSupertypesResolution.kt
```diff
@@ -129,8 +132,22 @@ private class FirApplySupertypesTransformer(
...
+                if (it.classId.isNestedClass) {
+                    val outerClassFir = session.firProvider.getFirClassifierByFqName(it.classId.outerClassId!!) as? FirClass<*>
+                    if (outerClassFir != null) {
+                        (it.scopeProvider.getNestedClassifierScope(outerClassFir, session, scopeSession) as? FirNestedClassifierScope)
```
dzharkov, line 142: I didn't get why it is necessary but modifying the scope of some class that potentially belongs to an independent module and doesn't know about our type alias looks suspicious

dzharkov (changes requested):
Please, clarify why these changes are necessary, why we need modifying the scope of independent class

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSupertypesResolution.kt
```diff
@@ -129,8 +132,22 @@ private class FirApplySupertypesTransformer(
             "Expected single supertypeRefs, but found ${supertypeRefs.size} in ${typeAlias.symbol.classId}"
         }
 
+        val supertypeRef = supertypeRefs.single()
+        // TODO: Need better handling for type alias with type parameters.
+        if (typeAlias.typeParameters.isEmpty()) {
```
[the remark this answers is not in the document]
jsjeon: Sorry, it's also hard for me to remember why I added that two weeks ago. :\

Without this guard, I can see more `UNRESOLVED_REFERENCE`s are resolved, which looks good, but just reminded me of the mixed signal from `typealias/innerClassTypeAliasConstructorInSupertypes`. Still, more super types are resolved, but not quite sure about two regressions there.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSupertypesResolution.kt
jsjeon: Thanks. Also noticed that it could be type parameter. Changed to use `fullyExpandedType`, and deferred to symbolize it.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/FirSupertypesResolution.kt
jsjeon: Well, that's the point: the scope of a class that doesn't know about the type alias that we're currently resolving here could lead to unresolved constructor call later. I added comments to articulate the situation.

jsjeon:
> Please, clarify why these changes are necessary, why we need modifying the scope of independent class

I think the initial PR explanation articulates the situation, but maybe not enough. Added comments to the code as well, but I can do that again here.

```kt
class Outer {
  inner class Inner { ... }
}
typealias OI = Outer.Inner
fun foo() {
  Outer().OI() // `OI` is currently not resolved
}
```

With `Outer()` as an explicit receiver, the call resolver first collects candidates with the member scope of `Outer`, which in turn goes down to the nested classifier scope, where `Inner` is clearly cached, but no information about the type alias, of course. So, `OI` is simply not resolved at that moment, and the overall purpose of this PR is to make the type alias be resolved at that point, which I believe is the earliest, and correct: in the sense that `MemberScopeTowerLevel` will correctly handle the explicit receiver as a dispatch receiver.

After trying the member scope, the call resolver tries again with `ScopeTowerLevel`, where the package scope can resolve the type alias, where it is literally defined. But, now, the explicit receiver is treated as an extension receiver, hence rejecting the candidate with an inconsistent receiver. Well, we can add an exception there, that if the resolved type alias points to a nested class, that seemingly inconsistent receiver is actually a dispatch receiver and moving on, which sounds awful. Another alternative is to revoke the whole resolution and retry it with the resolved type alias. Then, the call resolver revisits the member scope with `Inner` instead, which should be resolved in the nested classifier scope. But... I'm not sure there is an easy way of revoke-and-retry in the call resolution. So I ended up with kind of populating the nest classes that can be exposed via type alias, directly into the nested classifier scope.

Btw, I'm not sure what "independent" class means. Inner/nested classes are exposed to outer scopes via type alias, and when we know which (type alias) is which (inner/nested class), aren't they now "bound" (to each other)?

dzharkov:
Sorry, I haven't noticed the description under the PR title :(
Now, I understood the motivation.
It seems that the chosen approach is a bit fragile: for example, it would work wrong if there was a nested class having the same name as our type alias.
The latter one would rewrite the former and it becomes unavailable through the nested classifier scope.

The more generalized idea would be rewriting `org.jetbrains.kotlin.fir.resolve.calls.ConstructorProcessingKt#processConstructors` so that it makes inner constructor based on type alias having not a dispatch but an extension receiver.

So, resolving the call like `outer.TypeAlias()` would work just like there would be declarations similar to
```
typealias TypeAlias = Outer.Inner 
fun Outer.TypeAlias(): TypeAlias = ...
```

Namely, I'd suggest creating a copy of inner constructors (when they came from typealias) somewhere beside `org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt:164`, so that these copies would have an extension receiver instead of a dispatch one.

Consider the following way of creating a copies for constructors: https://github.com/JetBrains/kotlin/commit/c7d6a79c252660610001a5cf0568f03e532fc442#diff-d97615bfdeec91327d39e8f27bb472c3R89

But, it should be a different `buildConstructorCopy` function and probably it's worth somehow preserving a reference to the original constructor among `attributes`

jsjeon:
Thank you for the thorough comments! I'm truly amazed how this could be (re-)implemented in a totally different way. Since github overwrites all the commits in the history, I rather upload the rework in a different branch/PR: https://github.com/JetBrains/kotlin/pull/3800

It turns out that some changes that were newly resolved in this PR are supposed to be _not_ resolved, and as you worried, adding typealias to a nested classifier symbol may overwrite existing names.

If my other rework looks better than this, feel free to close this one. Anything that needs to be improved can be discussed there.

PR #3747, opened 2020-09-23
JVM_IR: use the correct asm type when emitting constant values

pyos:
#KT-42092 Fixed


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -78,8 +80,9 @@ class MaterialValue(codegen: ExpressionCodegen, type: Type, irType: IrType) : Pr
 
 // A value that can be branched on. JVM has certain branching instructions which can be used
 // to optimize these.
-abstract class BooleanValue(codegen: ExpressionCodegen) :
-    PromisedValue(codegen, Type.BOOLEAN_TYPE, codegen.context.irBuiltIns.booleanType) {
+abstract class BooleanValue(codegen: ExpressionCodegen, type: Type, irType: IrType) : PromisedValue(codegen, type, irType) {
```
max-kammerer, line 83: Maybe rename to BranchedValue (similar to one in old backend)?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/PromisedValue.kt
```diff
@@ -98,17 +101,60 @@ abstract class BooleanValue(codegen: ExpressionCodegen) :
...
+        }
+    }
+
+    override fun discard() {
+        // Add something for the line number to be attached to.
+        mv.nop()
     }
 
-    override fun discard() {}
+    override fun jumpIfFalse(target: Label) = if (isDefaultValueForType(type, value)) mv.goTo(target) else mv.nop()
```
max-kammerer, line 126: ConstanValue class  and especially `jumpIfXXXX` method a bit confusing, cause only boolean constant should be branched  values. Maybe add assertion about types here? 
Or refactor values similar to old backend approached, where all constants except boolean are StackValue.Constant and boolean constants are BranchedValue.Companion.TRUE/FALSE

pyos: Technically any value can be converted to a branched value via `coerceToBoolean`, and treating 0s as false is consistent with that.

max-kammerer: Technically yes, but could such situation arised in bytecode generation?

pyos: With generics yes, but with constants no. Removed this inheritance for non-boolean constants.

max-kammerer:
@pyos Could you rebase on master? Github reports conflicts

pyos:
Done. Looks like someone already fixed the integer condition in default stubs.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
```diff
@@ -764,20 +764,37 @@ class ExpressionCodegen(
...
+                when (value) {
+                    is Char -> mv.iconst(value.toInt())
+                    is Long -> mv.lconst(value)
+                    is Float -> mv.fconst(value)
+                    is Double -> mv.dconst(value)
+                    is Number -> mv.iconst(value.toInt())
+                    else -> mv.aconst(value)
+                }
+                // `null` is of any reference type.
+                if (value != null || isPrimitive(target)) {
```
max-kammerer, line 791: Why non-primitive types are not cast?

pyos: Because `aconst null` pushes a value of type "bottom", which is not representable as an asm type but is a subtype of any reference type and does not need casting.

max-kammerer: Yes, sorry for stupid question. 
But `isPrimitive(target)` check looks redundant, no?

max-kammerer:
Please look into `FirBlackBoxCodegenTestGenerated$Intrinsics.testDefaultObjectMapping` test TC reports fail

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
pyos: Probably, unboxing null into a primitive does not sound like a valid operation.

pyos:
I'll mute this test for now, FIR is doing something weird: `java.lang.Byte.MAX_VALUE` produces a reference to a field that has an `IrConst` initializer with a kind of `Int` (and an `Integer` value) and an error type. The error type is later replaced by `Byte` somewhere, but the value and the kind are still incorrect.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: Yes, thank you!

max-kammerer:
cc @demiurg906 ^^^

demiurg906:
Thanks, will fire a ticket for it

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
max-kammerer: Looks like this change exposes additional problems with types in Fir,
there are at least 3 failing test in running build:
FirBlackBoxCodegenTestGenerated$Classes.testKt1018
FirBytecodeTextTestGenerated$NullCheckOptimization.testRedundantSafeCall_1_4
FirBytecodeTextTestGenerated$NullCheckOptimization.testRedundantSafeCall

max-kammerer: There is also fail with `IrKotlinSteppingTestGenerated$Custom.testConstantConditions`

pyos: `RedundantSafeCall`s are unmuted by the PR, some recent commit must've broken them in another way.

pyos: Oh, right, those tests have precisely the problem that null is unboxed into a primitive - on an unreachable code path.

max-kammerer:
@pyos  Could you rebase pr?

pyos:
Seems like this PR is not needed anymore since @udalov proposes to replicate the old backend's behavior in a related issue (https://youtrack.jetbrains.com/issue/KT-42321).

udalov:
Some other tests are fixed here for FIR, are these changes also not needed?

Moreover, I had the feeling that it'd be easier to merge this PR, and then add a kludge above to replicate the old backend's behavior, but maybe I'm wrong

pyos:
The FIR changes are accidental - these tests contain redundant safe-call operators that are (correctly) inferred to have non-null result type but are still translated to conditionals in the IR, which results in a cast of `null` to a primitive type on a dead code path.

pyos:
This PR might be needed after all - there are other cases with constant expressions of non-Int type where the IR backend currently fails to insert a cast, but the old backend handles them fine. I added one such case as a test.

udalov:
@pyos FYI I'm working on restoring the incorrect old backend's behavior in [`rr/udalov3`](https://github.com/JetBrains/kotlin/compare/rr/udalov3)

pyos:
Yep, fails with a ClassCastException now, so I've removed it. The other test is still valid until KT-38895 takes effect.

compiler/testData/codegen/box/casts/javaInterop/enhancedNullabilityConstant.kt
```diff
@@ -0,0 +1,24 @@
...
+fun box(): String {
+    val x = X()
+    val y = id(x.y ?: 0 * 1)
+    return if (y == 0L) "OK" else "fail: $y"
+}
+
+// FILE: Nullable.java
+package android.support.annotation;
+
+public @interface Nullable {}
```
udalov, line 15: If the issue is reproducible with our existing `org.jetbrains.annotations.Nullable`, I'd prefer to use it

pyos: Done

max-kammerer:
@pyos  IrKotlinSteppingTestGenerated$Custom.testConstantConditions test fails

max-kammerer:
@pyos, gently ping

pyos:
I still can't run IDEA debugger tests due to weird JDK errors :-/

madsager:
Maybe the issue will show in a IrStepping or IrLocalVariables test as well. We could try to take the code for that test and create a simpler non-idea debugger test for it?

PR #3748, opened 2020-09-24
FIR2IR: put @JvmField on field, not property

jsjeon:
See https://youtrack.jetbrains.com/issue/KT-42118 for more details. The previous work (https://github.com/JetBrains/kotlin/pull/3381) on splitting property annotations simply missed `@JvmField` annotation.

------

With that, 4 tests with `@JvmField` property in interface companion failed, which reveals the remaining TODO in the serializer. The 2nd commit adds necessary addition of version requirement in a `fir` way.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/AnnotationGenerator.kt
```diff
@@ -48,7 +49,8 @@ internal class AnnotationGenerator(private val components: Fir2IrComponents) : F
         irProperty.annotations +=
             property.annotations
                 .filter {
-                    it.useSiteTarget == null || it.useSiteTarget == AnnotationUseSiteTarget.PROPERTY
+                    !it.isJvmFieldAnnotation &&
```
mglukhikh, line 52: This `isJvmFieldAnnotation` check, as well as the next one, looks too specific for me. Normally we should check `@Target` of an annotation class, and if it's `FIELD`, then annotation with `null` use-site target isn't applicable to a property but applicable to its field.

mglukhikh: Just checked in the debugger: you can take `it.annotationTypeRef`, get its related `FirRegularClass`, and this class has `@Target` annotation. So let's do it in a generalized way :).

mglukhikh: Cite from Kotlin reference just to be precise:

```
If you don't specify a use-site target, the target is chosen according to the @Target annotation of the annotation being used. If there are multiple applicable targets, the first applicable target from the following list is used:

- param;
- property;
- field.
```

jsjeon: Thanks for the suggestion!

I saw you already made some utils to retrieve meta annotation, like `@Retention`, so I migrated all the annotation-related utils into one place; extended it to retrieve `@Target` (a bit different, since it has array value); and then generalize property-v.s.-field targeting with targets in `@Target`. Now I can unmute one more. :)

mglukhikh (approved):

mglukhikh:
Merged manually, thank you!

PR #3749, opened 2020-09-24
JVM_IR: do not use fields' superQualifierSymbol to cast the receiver

pyos:
It's only for computing the field owner.

 #KT-42131 Fixed


udalov:
Thanks!

PR #3750, opened 2020-09-24
KGP: Clean up configurations setup

gavra0:
KAPT classpath now uses extendsFrom in order to
resolve all configurations as a single graph.

Also, use Configuration.extendsFrom instead of
DependencyHandler.add that is being deprecated
in Gradle 6.7.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
@@ -447,7 +447,13 @@ class Kapt3GradleSubplugin @Inject internal constructor(private val registry: To
                 }
             }
 
-            kaptTask.kaptClasspathConfigurations = kaptClasspathConfigurations
+            val kaptClasspathConfiguration =
+                project.configurations.create("_kaptClasspath_" + kaptTask.name).setExtendsFrom(kaptClasspathConfigurations).also {
+                    it.isVisible = false
+                    it.isCanBeConsumed = false
+                }
+            kaptTask.kaptClasspath.from(kaptClasspathConfiguration)
+            kaptTask.kaptClasspathConfigurationNames.set(kaptClasspathConfigurations.map { it.name })
 
             KaptWithAndroid.androidVariantData(this)?.annotationProcessorOptionProviders?.let {
                 kaptTask.annotationProcessorOptionProviders.add(it)
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/gradleConfigurationUtils.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/gradleConfigurationUtils.kt
@@ -6,11 +6,14 @@
 package org.jetbrains.kotlin.gradle.utils
 
 import org.gradle.api.Project
+import java.lang.RuntimeException
 
 fun Project.addExtendsFromRelation(extendingConfigurationName: String, extendsFromConfigurationName: String, forced: Boolean = true) {
-    if (extendingConfigurationName != extendsFromConfigurationName) {
-        if (forced || configurations.findByName(extendingConfigurationName) != null) {
-            project.dependencies.add(extendingConfigurationName, project.configurations.getByName(extendsFromConfigurationName))
-        }
-    }
+    if (extendingConfigurationName == extendsFromConfigurationName) return
+
+    val extending = configurations.findByName(extendingConfigurationName)
+        ?: if (forced) throw RuntimeException("Configuration $extendingConfigurationName does not exist.")
+        else return
+
+    extending.extendsFrom(configurations.getByName(extendsFromConfigurationName))
 }
--- a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptTask.kt
+++ b/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptTask.kt
@@ -1,8 +1,9 @@
 package org.jetbrains.kotlin.gradle.internal
 
-import org.gradle.api.artifacts.Configuration
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.FileCollection
 import org.gradle.api.internal.ConventionTask
+import org.gradle.api.provider.ListProperty
 import org.gradle.api.tasks.*
 import org.gradle.api.tasks.incremental.IncrementalTaskInputs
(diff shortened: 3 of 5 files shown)


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/kotlinCompilations.kt
```diff
@@ -122,12 +122,12 @@ abstract class AbstractKotlinCompilation<T : KotlinCommonOptions>(
 
                 // Use `forced = false` since `api`, `implementation`, and `compileOnly` may be missing in some cases like
```
hungvietnguyen, line 123: Update comment

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
```diff
@@ -447,7 +447,13 @@ class Kapt3GradleSubplugin @Inject internal constructor(private val registry: To
                 }
             }
 
-            kaptTask.kaptClasspathConfigurations = kaptClasspathConfigurations
+            val kaptClasspathConfiguration =
+                project.configurations.create("_kaptClasspath" + kaptTask.name).setExtendsFrom(kaptClasspathConfigurations).also {
```
hungvietnguyen, line 451: nit: add "_" before kaptTask.name (looks like it starts with lowercase?)

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/gradleConfigurationUtils.kt
```diff
@@ -7,10 +7,10 @@ package org.jetbrains.kotlin.gradle.utils
 
 import org.gradle.api.Project
 
-fun Project.addExtendsFromRelation(extendingConfigurationName: String, extendsFromConfigurationName: String, forced: Boolean = true) {
```
hungvietnguyen, line 10: If `configurations.findByName(extendingConfigurationName) == null`, by default the existing code would still run `project.dependencies.add(extendingConfigurationName,...)` (which may succeed or fail depending on the contract of the `add` method, but in the new code it will just ignore it. Is the new behavior expected?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/kotlinCompilations.kt
gavra0: Done.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt
gavra0: done

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/gradleConfigurationUtils.kt
gavra0: `add` invokes `ConfigurationContainer.getByName(...)` which fails if that configuration does not exist i.e previous code:
- would not run if `force = false` if configuration does not exist; which is done in this PR
- would throw if `force = true` and conf does not exist => this was never the case.

Because of that, this change should not change the semantic of this code.

hungvietnguyen: > would throw if force = true and conf does not exist => this was never the case.

I agree this may not have happened, but the intention of `force` is to prevent this case from happening *in the future* right? So previously, calling `addExtendsFromRelation("nonExistentConfig", "existingConfig", force=true)` would fail, whereas the new code will just ignore it.

gavra0: I find it hard to justify usage of `force=true` TBH. Tests should ensure correctness of the implementation.

@h0tk3y Can you please chime in here, considering you are the original author of this utility?

hungvietnguyen: I think throwing is usually better than swallowing so normally we would write `configurations.findByName(extendingConfigurationName)!!.extendsFrom(...)` instead of `configurations.findByName(extendingConfigurationName)?.extendsFrom(...)`. But for some old projects as mentioned in the code comment we actually have to ignore the error, so maybe that's why `force` was introduced (and defaulted to `true`). Anyway, I'll defer to @h0tk3y on this.

gavra0: I've changed this to have a more helpful error message in case `forced=true`, failing early does make sense.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/utils/gradleConfigurationUtils.kt
```diff
@@ -6,11 +6,14 @@
...
-            project.dependencies.add(extendingConfigurationName, project.configurations.getByName(extendsFromConfigurationName))
-        }
-    }
+    if (extendingConfigurationName == extendsFromConfigurationName) return
+
+    val extending = configurations.findByName(extendingConfigurationName)
+        ?: if (forced) throw RuntimeException("Configuration $extendingConfigurationName does not exist.")
+        else return
+
+    extending.extendsFrom(configurations.getByName(extendsFromConfigurationName))
```
nav-nav, line 18: Do we need search it again? Is "extending" enough?

gavra0: We are searching for `extendsFromConfigurationName`, not `extendingConfigurationName`, so this `getByName` cannot be avoided.

gavra0:
This is merged in master as https://github.com/JetBrains/kotlin/commit/075ad9ea664c89d53159c5707323a05a8f6b1147.

ericoporto:
In which version of Kotlin does the Warning goes away?

TWiStErRob:
1.4.30 released few days ago. Commit on master: https://github.com/JetBrains/kotlin/commit/33a0ec9b4f40f3d6f1f96b2db504ade4c2fafe03 which has the tags.
(Found the commit by opening the file that was changed and blaming the line that was changed.)

PR #3751, opened 2020-09-24
JVM_IR: check for null when converting unboxed inline classes to strings

pyos:
`C?` can be unboxed into `T?` if if wraps a reference type `T!!`, but in this case `null` is not a valid value to pass to `toString-impl`.

 #KT-42005 Fixed


compiler/testData/codegen/bytecodeText/inlineClasses/toStringIsCalledByInlineClass.kt
```diff
@@ -1,13 +1,25 @@
 // !LANGUAGE: +InlineClasses
+// Completely incorrect bytecode in some of these - see `box/inlineClasses/toStringOfUnboxedNullable.kt`
+// IGNORE_BACKEND: JVM
```
udalov, line 3: Please extract the parts that do work in JVM, into a separate non-muted test

pyos: Moved the class that wraps `Any` to a separate test

udalov:
Thank you!

PR #3753, opened 2020-09-24
JVM IR: Fix types in generated function reference constructor

sfs:
This fixes a build error when compiling [arcs](https://github.com/PolymerLabs/arcs/blob/73da1581cfb4871b89925f67929a86ccd44f2a26/java/arcs/core/storage/DirectStoreMuxer.kt#L77) with the JVM IR backend.

Interestingly enough, the added `testSuspendFunctionMethodReference` currently fails for the JVM backend, but works fine when compiled from the command line.


udalov:
Merged with a minor correction in f8f2be3d9fb. The test was failing for the JVM backend because there's no stdlib in the classpath in box tests by default. (I also found [a bug in kotlin-reflect](https://youtrack.jetbrains.com/issue/KT-42199) while testing.)

Thank you!

PR #3754, opened 2020-09-24
[JVM_IR] Fix expectations for parcelize tests.

madsager:
The generated bridges now have line numbers as for the JVM
backend. There are some extra (irrelevant) labels in the JVM_IR
bytecode listing compared to JVM so expectations can still not
be shared even though they are closer.


madsager:
@max-kammerer thanks for pointing these out!

udalov:
Thanks! Since this fixes tests in master, I'll merge right away.

PR #3755, opened 2020-09-24
[FIR] Capture array and indices, handle varargs in overloaded indexed access operator, keep argument order

punzki:
This also includes a fix (2nd commit) that ensures that vararg arguments are kept in the right order in resolved calls. They were being added to the end (https://youtrack.jetbrains.com/issue/KT-17691), which is the behavior with the non-IR backend with both old and new inference. However with the JVM IR backend, the bug is fixed, hence it should be fixed with FIR as well.

### DIFF
--- a/compiler/fir/raw-fir/raw-fir.common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
+++ b/compiler/fir/raw-fir/raw-fir.common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
@@ -55,6 +55,8 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
     abstract fun T.getChildNodeByType(type: IElementType): T?
     abstract val T?.receiverExpression: T?
     abstract val T?.selectorExpression: T?
+    abstract val T?.arrayExpression: T?
+    abstract val T?.indexExpressions: List<T>?
 
     /**** Class name utils ****/
     inline fun <T> withChildClassName(
@@ -435,6 +437,17 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
             )
         }
 
+        if (unwrappedArgument.elementType == ARRAY_ACCESS_EXPRESSION) {
+            return generateIncrementOrDecrementBlockForArrayAccess(
+                baseExpression,
+                operationReference,
+                unwrappedArgument,
+                callName,
+                prefix,
+                convert
+            )
+        }
+
         return buildBlock {
             val baseSource = baseExpression?.toFirSourceElement()
             val desugaredSource = baseSource?.fakeElement(FirFakeSourceElementKind.DesugaredIncrementOrDecrement)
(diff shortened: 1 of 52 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
```diff
@@ -74,6 +78,7 @@ fun mapArguments(
                 name = function.valueParameters.last().name
             }
             argumentsInParenthesis = argumentsInParenthesis.dropLast(1) + listOf(namedV)
+            oldToNewArgumentMap[v] = namedV
```
punzki, line 81: This is used to replace the set argument in the call's argument list (see changes in `FirCallCompletionResultsWriterTransformer.kt`). Otherwise `remapArgumentsWithVararg` in `BodyResolveUtils.kt` simply loses the argument (it remaps based on the call argument list). I'm open to other suggestions on how to solve this.

punzki:
Added a 3rd commit to capture array and indices for postfix/prefix increment/decrement of array element (including overloaded indexed access operators, e.g., `a[b, c]++`).

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Candidate.kt
```diff
@@ -101,6 +101,7 @@ class Candidate(
     var usesSuspendConversion: Boolean = false
 
     var argumentMapping: Map<FirExpression, FirValueParameter>? = null
+    var oldToNewArgumentMapping: Map<FirExpression, FirExpression>? = null
```
demiurg906, line 104: I think it's better to rename that mapping to something more specific, lile `argumentMappingForArraySetCall`, and also make this property nullable

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
```diff
@@ -27,6 +27,7 @@ data class ArgumentMapping(
     //      foo(b = bar(), a = qux())
     // parameterToCallArgumentMap.values() should be [ 'bar()', 'foo()' ]
     val parameterToCallArgumentMap: Map<FirValueParameter, ResolvedCallArgument>,
+    val oldToNewArgumentMap: Map<FirExpression, FirExpression>,
```
demiurg906, line 32: Same for nullability and name

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
```diff
@@ -26,6 +26,8 @@ data class ArgumentMapping(
     //      fun foo(a: Int, b: Int) {}
     //      foo(b = bar(), a = qux())
     // parameterToCallArgumentMap.values() should be [ 'bar()', 'foo()' ]
+    // TODO: Consider changing this (and other similar declarations like Candidate.argumentMapping) to LinkedHashMap to signify that
```
demiurg906, line 29: This TODO is important and it's good to implement it now

compiler/fir/raw-fir/raw-fir.common/src/org/jetbrains/kotlin/fir/builder/BaseFirBuilder.kt
```diff
@@ -623,6 +636,147 @@ abstract class BaseFirBuilder<T>(val baseSession: FirSession, val context: Conte
...
+     *     val <array> = a
+     *     val <index0> = b
+     *     val <index1> = c
+     *     val <unary-result> = <array>.get(b, c).inc()
+     *     <array>.set(b, c, <unary-result>)
+     *     ^<unary-result>
+     * }
+     *
+     */
+    private fun generateIncrementOrDecrementBlockForArrayAccess(
```
demiurg906, line 667: Very important change, thanks for it

demiurg906 (changes requested):
@punzki thanks for this PR. They're basically OK modulo minor remarks

demiurg906:
Also note that `IrTextTestCaseGenerated$Expressions.testGenericConstructorCallWithTypeArguments` is failing in your branch. I suppose there was `FIR_IDENTICAL` flag, and after your changes it's not relevant any more

punzki:
> Also note that `IrTextTestCaseGenerated$Expressions.testGenericConstructorCallWithTypeArguments` is failing in your branch. I suppose there was `FIR_IDENTICAL` flag, and after your changes it's not relevant any more

I think this was broken by https://github.com/JetBrains/kotlin/commit/10e7c0f1b26e00a878c35a1a7fe6b667e85dbac3. It is not related to the changes in these PR. But I can fix them here, or it might be better to do it on your end since it's a small, fast change.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/Candidate.kt
punzki: I ended up finding a different solution. I removed this and instead we can use `Candidate.argumentMapping` instead of the `Fir*Call.argumentList` to remap vararg arguments to a `FirVarargArgumentExpression`.

The `argumentList` was used to preserve the order, but we can rely on the mapping to have its keys in order. See latest 2 commits.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/FirArgumentsToParametersMapper.kt
punzki: Done. See last commit.

demiurg906:
> I think this was broken by [10e7c0f](https://github.com/JetBrains/kotlin/commit/10e7c0f1b26e00a878c35a1a7fe6b667e85dbac3). It is not related to the changes in these PR. But I can fix them here, or it might be better to do it on your end since it's a small, fast change.

Try to rebase your branch to master. This test should be fixed there

demiurg906:
I ran `rr` for rebased branch, and there are two tests are failing:
- `FirBlackBoxInlineCodegenTestGenerated$ArgumentOrder.testDefaultParametersAndLastVararg`
- `FirBlackBoxInlineCodegenTestGenerated$ArgumentOrder.testVarargAndDefaultParameters`

Please fix them. Everything else is fine

demiurg906:
Also check two new commits about varargs added today
There may be useful information about vararg handling and new tests which also may change after your commits:
- https://github.com/JetBrains/kotlin/commit/d62c665e9968b372b97889591051036e3c1f14bc
- https://github.com/JetBrains/kotlin/commit/d02432cf93df9e61eba0c7aa40396a94334defdf

punzki:
> > I think this was broken by [10e7c0f](https://github.com/JetBrains/kotlin/commit/10e7c0f1b26e00a878c35a1a7fe6b667e85dbac3). It is not related to the changes in these PR. But I can fix them here, or it might be better to do it on your end since it's a small, fast change.
> 
> Try to rebase your branch to master. This test should be fixed there

I did just that, the test still fails. Even with a clean checkout of `origin/master`, it still fails.

punzki:
> I ran `rr` for rebased branch, and there are two tests are failing:
> 
> * `FirBlackBoxInlineCodegenTestGenerated$ArgumentOrder.testDefaultParametersAndLastVararg`
> * `FirBlackBoxInlineCodegenTestGenerated$ArgumentOrder.testVarargAndDefaultParameters`
> 
> Please fix them. Everything else is fine

Fixed these tests, added new ones with `UseCorrectExecutionOrderForVarargArguments` enabled.

demiurg906:
> I did just that, the test still fails. Even with a clean checkout of origin/master, it still fails.

On current master this test is passing

demiurg906:
Merged manually

PR #3756, opened 2020-09-25
FIR2IR: set dispatch receiver for companion member reference

jsjeon:
As described at https://youtrack.jetbrains.com/issue/KT-42132, a callable reference to companion member needs `GET_OBJECT` as dispatch receiver.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
```diff
@@ -361,14 +361,35 @@ class CallAndReferenceGenerator(
...
         val classSymbol = (qualifier.typeRef.coneType as? ConeClassLikeType)?.lookupTag?.toSymbol(session)
-        if (callableReferenceMode && classSymbol is FirRegularClassSymbol) {
-            if (classSymbol.classId != qualifier.classId) {
+        if (callableReferenceAccess != null && classSymbol is FirRegularClassSymbol) {
+            val classIdMatched = classSymbol.classId == qualifier.classId
+            val companionClassIdMatched = classSymbol.fir.isCompanion && classSymbol.classId.outerClassId == qualifier.classId
+            val lastClassIfCompanion = conversionScope.lastClass()?.takeIf { it.isCompanion }
+            val callable =
+                ((callableReferenceAccess.calleeReference as FirResolvedNamedReference).resolvedSymbol as FirCallableSymbol<*>)
+            val companionMemberReferenceThatNeedsDispatchReceiver = companionClassIdMatched &&
```
mglukhikh, line 379: I must confess all this code looks like a hack, sorry. Will try to invent something better...

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
```diff
@@ -361,14 +361,35 @@ class CallAndReferenceGenerator(
...
+                    classSymbol.fir.declarations.contains(callable.fir) &&
+                    // Built-in extensions are stored in the companion, e.g., String.Companion.
+                    // Make sure this is not a reference to an extension function.
+                    !callable.isExtension &&
+                    // Dispatch receiver is not needed if we're converting a callable reference inside that companion, e.g.,
+                    // class Foo {
+                    //   companion object {
+                    //      val cp = Foo::member
+                    // } }
+                    lastClassIfCompanion?.classId != classSymbol.classId
```
mglukhikh, line 391: After some consideration I dropped this check. It looks very inconsistent, and does not help any BB / text test to pass.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
```diff
@@ -361,14 +361,35 @@ class CallAndReferenceGenerator(
...
+        if (callableReferenceAccess != null && classSymbol is FirRegularClassSymbol) {
+            val classIdMatched = classSymbol.classId == qualifier.classId
+            val companionClassIdMatched = classSymbol.fir.isCompanion && classSymbol.classId.outerClassId == qualifier.classId
+            val lastClassIfCompanion = conversionScope.lastClass()?.takeIf { it.isCompanion }
+            val callable =
+                ((callableReferenceAccess.calleeReference as FirResolvedNamedReference).resolvedSymbol as FirCallableSymbol<*>)
+            val companionMemberReferenceThatNeedsDispatchReceiver = companionClassIdMatched &&
+                    // TODO: this can't cover delegated members. See bb test kt38664.kt
+                    // Make sure the reference indeed refers to a member of that companion
+                    classSymbol.fir.declarations.contains(callable.fir) &&
```
mglukhikh, line 382: This can be replaced with `callableId` parent check. It's easier to perform, and it covers this `kt38664.kt` you've reminded above.

mglukhikh:
Merged manually. I simplified your code with the following commit: https://github.com/JetBrains/kotlin/commit/cebbd21a1e8116259d944ca52e9240b7b90d819c. See also my comments here in PR.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
jsjeon: It helped prevent any regressions, but probably those regressions were due to missed parent check you mentioned below. Thanks.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
jsjeon: Yeah, that was my intention; good to learn that 👍. Thanks!

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/generators/CallAndReferenceGenerator.kt
jsjeon: Thank you for the all simplifications!

PR #3757, opened 2020-09-25
JVM_IR: slightly refactor ClassCodegen

pyos:
* reduce code duplication for constructing JvmDeclarationOrigin;
 * move all `visitor.visitInnerClass` calls to before `visitor.done`;
 * make the order of members more strict & remove 3 mutable fields, as order-dependent mutation can lead to differences with incremental compilation due to inline calls (thus keeping mutable state in ClassCodegen is in general unsafe).

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
@@ -18,7 +18,7 @@ import org.jetbrains.kotlin.config.LanguageVersionSettings
 import org.jetbrains.kotlin.descriptors.DescriptorVisibilities
 import org.jetbrains.kotlin.descriptors.DescriptorVisibility
 import org.jetbrains.kotlin.descriptors.Modality
-import org.jetbrains.kotlin.ir.builders.declarations.buildFun
+import org.jetbrains.kotlin.ir.builders.declarations.addFunction
 import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.descriptors.toIrBasedDescriptor
 import org.jetbrains.kotlin.ir.expressions.IrBlockBody
@@ -27,8 +27,6 @@ import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrFunctionReference
 import org.jetbrains.kotlin.ir.expressions.impl.IrBlockBodyImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrSetFieldImpl
-import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
-import org.jetbrains.kotlin.ir.types.IrSimpleType
 import org.jetbrains.kotlin.ir.util.*
 import org.jetbrains.kotlin.load.java.JvmAbi
 import org.jetbrains.kotlin.load.java.JvmAnnotationNames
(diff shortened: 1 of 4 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
```diff
@@ -124,15 +102,12 @@ class ClassCodegen private constructor(
         }
     }
 
-    private var hasAssertField = irClass.hasAssertionsDisabledField(context)
-    private var classInitializer = irClass.functions.singleOrNull { it.name.asString() == "<clinit>" }
-    private var generatingClInit = false
     private var generated = false
 
-    fun generate(parentDelegatedPropertyTracker: DelegatedPropertyOptimizer? = null): ReifiedTypeParametersUsages {
+    fun generate(parentDelegatedPropertyTracker: DelegatedPropertyOptimizer? = null) {
```
pyos, line 107: FYI passing `DelegatedPropertyOptimizer` like this is incorrect - the description of #3629 is wrong, and inline functions cannot in fact be excluded:
```
import kotlin.reflect.KProperty

inline operator fun String.getValue(t:Any?, p: KProperty<*>): String = p.name + this

object ForceOutOfOrder {
    // This call forces `C$inlineFun$1` to be generated before `C` itself.
    // If `IrSourceCompilerForInline` passes a `null` optimizer to `generateMethodNode`,
    // the use of `C.$$delegatedProperties` will not be recorded anywhere.
    fun callInline() = C.inlineFun()
}

object C {
    inline fun inlineFun() = {
        val O by "K"
        O
    }()
}

fun box(): String = ForceOutOfOrder.callInline() // NoSuchFieldError
```
I will fix this in a separate PR.

pyos: This is also problematic if `inlineFun` is called from a different package, as `$$delegatedProperties` has package-private visibility...

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -328,42 +318,30 @@ private val KOTLIN_MARKER_INTERFACES: Map<FqName, String> = run {
...
     val sw = BothSignatureWriter(BothSignatureWriter.Mode.CLASS)
-
-    typeMapper.writeFormalTypeParameters(irClass.typeParameters, sw)
+    writeFormalTypeParameters(irClass.typeParameters, sw)
 
     sw.writeSuperclass()
-    val irType = superClassInfo.irType
-    if (irType == null) {
-        sw.writeClassBegin(superClassInfo.type)
+    val superClassType = irClass.superTypes.find { it is IrSimpleType && (it.classifier as? IrClassSymbol)?.owner?.isJvmInterface == false }
```
udalov, line 326:
```suggestion
    val superClassType = irClass.superTypes.find { it.getClass()?.isJvmInterface == false }
```

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ClassCodegen.kt
udalov: Thanks! Reported the first problem as [KT-42253](https://youtrack.jetbrains.com/issue/KT-42253), the second one looks like [KT-39434](https://youtrack.jetbrains.com/issue/KT-39434).

udalov:
Thanks!

PR #3758, opened 2020-09-25
Produce deterministic jar files.

mpetrov:
This resets all timestamps present in jars produced by kotlinc.

This is important for build systems like bazel that rely on
deterministic outputs. One can revert to the legacy behavior
by using the `-Xno-reset-jar-timestamps` flag.

Before:

```
$ kotlinc ~/test.kt -d /tmp/a.jar
$ kotlinc ~/test.kt -d /tmp/b.jar
$ shasum /tmp/a.jar /tmp/b.jar
1ba022697317f796bd123fb4dc95418a18bcb51a  /tmp/a.jar
6d2a2683470c24928f3fbd6768a4c57f55b0d196  /tmp/b.jar
$ unzip -l /tmp/a.jar
Archive:  /tmp/a.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
       75  09-25-2020 16:48   META-INF/MANIFEST.MF
      683  09-25-2020 16:48   TestKt.class
       28  09-25-2020 16:48   META-INF/main.kotlin_module
---------                     -------
      786                     3 files
```

After:

```
$ kotlinc ~/test.kt -d /tmp/a.jar
$ kotlinc ~/test.kt -d /tmp/b.jar
$ shasum /tmp/a.jar /tmp/b.jar
86b9614fab6973d2d637b804fd460d2d184c9b08  /tmp/a.jar
86b9614fab6973d2d637b804fd460d2d184c9b08  /tmp/b.jar
$ unzip -l /tmp/a.jar
Archive:  /tmp/a.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
       75  01-01-1980 00:00   META-INF/MANIFEST.MF
      590  01-01-1980 00:00   TestKt.class
       36  01-01-1980 00:00   META-INF/main.kotlin_module
---------                     -------
      701                     3 files
```

See https://github.com/JetBrains/kotlin/pull/3226 for a similar change.

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/CompileEnvironmentUtil.java
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/CompileEnvironmentUtil.java
@@ -37,6 +37,9 @@ import java.util.jar.*;
 import static org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity.ERROR;
 
 public class CompileEnvironmentUtil {
+
+    public static long DOS_EPOCH = new GregorianCalendar(1980, Calendar.JANUARY, 1, 0, 0, 0).getTimeInMillis();
+
     @NotNull
     public static ModuleChunk loadModuleChunk(File buildFile, MessageCollector messageCollector) {
         if (!buildFile.exists()) {
@@ -66,13 +69,13 @@ public class CompileEnvironmentUtil {
             JarOutputStream stream = new JarOutputStream(fos);
             JarEntry manifestEntry = new JarEntry(JarFile.MANIFEST_NAME);
             long dosEpoch = new GregorianCalendar(1980, Calendar.JANUARY, 1, 0, 0, 0).getTimeInMillis();
-            manifestEntry.setTime(dosEpoch);
+            manifestEntry.setTime(DOS_EPOCH);
             stream.putNextEntry(manifestEntry);
             manifest.write(new BufferedOutputStream(stream));
 
             for (OutputFile outputFile : outputFiles.asList()) {
                 JarEntry entry = new JarEntry(outputFile.getRelativePath());
-                entry.setTime(dosEpoch);
+                entry.setTime(DOS_EPOCH);
                 stream.putNextEntry(entry);
                 stream.write(outputFile.asByteArray());
             }
@@ -119,6 +122,7 @@ public class CompileEnvironmentUtil {
                 if (e == null) {
                     break;
                 }
+                e.setTime(DOS_EPOCH);
                 if (FileUtilRt.extensionEquals(e.getName(), "class")) {
                     stream.putNextEntry(e);
                     FileUtil.copy(jis, stream);
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/CompileEnvironmentUtil.java
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/CompileEnvironmentUtil.java
@@ -126,7 +126,7 @@ public class CompileEnvironmentUtil {
                     break;
                 }
                 if (resetJarTimestamps) {
-                  e.setTime(DOS_EPOCH);
+                    e.setTime(DOS_EPOCH);
                 }
                 if (FileUtilRt.extensionEquals(e.getName(), "class")) {
                     stream.putNextEntry(e);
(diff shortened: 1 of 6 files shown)


mpetrov:
Could someone have a look at this? It's affecting our bazel remote caches.

udalov:
Sorry for delay!

I'm not entirely sure that using the same timestamp for all files _by default_ is a good idea. This timestamp is further set as the file modification date if you unzip the archive, and having all files with the default 0 timestamp is unintuitive and might even trip up some other use cases. Maybe enable this behavior only if some compiler argument is specified, e.g.`-Xreset-jar-timestamps`, and use it in Bazel for Kotlin builds? I'd even argue that the same could be done for the klib changes in #3226.

Also, I'm not sure what should happen if `-include-runtime` is specified -- would that also reset timestamps in the bundled stdlib?

We'll also need some tests that check this behavior, but probably after we agree on the implementation.

Do you have any examples of this behavior in other languages or tools? The only data point that I could find is that the Scala compiler also doesn't reset timestamps when outputting to a jar.

madsager:
I don't have a strong opinion on the subject, but as for data points, the D8 dexer and R8 shrinker writes zip/jar files with reset timestamps by default. Using the D8/R8 programmatic APIs you can build a version of D8 and R8 that puts in current timestamps if desired. I don't know of anyone who does.

mpetrov:
Thanks for having a look!

My general thought is that the timestamps being included is just an unfortunate artifact of using zip as a file format. The compilation time is not included in Elf/Dwarf files written by `clang` for example; two identical clang invocations will produce the same output. I'm not familiar with how common this practice of resetting timestamps is in the wider community, though a lot of Google authored tools I've seen reset timestamps (e.g. https://github.com/bazelbuild/bazel/tree/master/third_party/ijar).

FWIW I made a similar change in https://github.com/JetBrains/kotlin/pull/3226 for `klib`s specifically, but those don't need to interact with external tooling in the same way jars might. I don't work on developer tooling as my day job though, so it's possible I'm missing some use cases. I'm not opposed to being more conservative and introducing a `-Xreset-jar-timestamp` if y'all think that's best.

I updated the PR so that `-include-runtime` also explicitly resets timestamps. In practice this still generated deterministic output though since multiple invocations would merge the same runtime lib jar.

I also took a stab at adding a test. Feedback on the testing strategy would be appreciated. Unfortunately just comparing the output of two invocations isn't sufficient to reliably produce an error with the existing implementation. DOS timestamps don't have much resolution. I could introduce a 2 second `sleep` if needed to force a timestamp increment, but that doesn't seem great.

mpetrov:
@udalov could you take a look?

udalov:
Sorry for the delay. Yeah, I read up a little bit on the timestamps in ZIP and can say that I'm convinced now :). Those timestamps indeed seem fairly useless, and the deterministic output use case looks far more important.

Regarding the test, I think it's OK as long as we're still verifying that the timestamps are reset.

I'll merge the change after a remote run. Thanks!

udalov:
After an internal discussion, we figured that we still need to have at least an opt-out flag for this behavior, to fallback to the old mode if it's needed in some use cases of which we are not aware. Could you please add such flag? I'd propose a name `-Xenable-jar-timestamps`, but not sure that it's the best one.

mpetrov:
I updated the PR to add a flag.

Thoughts on `-Xno-reset-jar-timestamps` instead as naming? Not sure if `enable`-prefixing makes it sound too much like a feature 🤷‍♂️. I'm happy to change the name to whatever if y'all have preferences.

compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/CompileEnvironmentUtil.java
```diff
@@ -94,21 +110,24 @@ public static void writeToJar(File jarPath, boolean jarRuntime, FqName mainClass
...
-    private static void copyJarImpl(JarOutputStream stream, File jarPath) throws IOException {
+    private static void copyJarImpl(JarOutputStream stream, File jarPath, boolean resetJarTimestamps) throws IOException {
         try (JarInputStream jis = new JarInputStream(new FileInputStream(jarPath))) {
             while (true) {
                 JarEntry e = jis.getNextJarEntry();
                 if (e == null) {
                     break;
                 }
+                if (resetJarTimestamps) {
+                  e.setTime(DOS_EPOCH);
```
udalov, line 129:
```suggestion
                    e.setTime(DOS_EPOCH);
```

udalov:
Thanks! Merged manually in 68fdeaf865f.

Also, I've reported [KT-43406](https://youtrack.jetbrains.com/issue/KT-43406) for changelog purposes & migration-related issues.

PR #3759, opened 2020-09-26
[FIR] Match behavior of prefix inc/decrement with old frontend

punzki:
...retrieving and returning the new argument value.

Rmoved the temporary variable that was used to store and return the result of the increment. Note: The previous behavior was correct according to spec, but KT-42077 proposes changing the behavior so that getters are only called once.


mglukhikh:
I observe failing test here: https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/3127633

punzki:
> I observe failing test here: https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/3127633

The error in `IrTextTestCaseGenerated` is not related to this change (that is with old front-end). Fixed the other test data.

mglukhikh:
The code itself is Ok. @dzharkov suggested to delay accepting this PR until some decision will be taken on https://youtrack.jetbrains.com/issue/KT-42077

punzki:
Rebased on top of #3755

mglukhikh:
We discussed this situation and decided that the current K2 (FIR) behavior matched the Kotlin spec and thus we stay on it. So I close this PR.

PR #3760, opened 2020-09-27
Added samples for the kotlin.streams package

uzilan:
Added samples for the kotlin.streams package

### DIFF
--- a/libraries/stdlib/samples/test/samples/streams/streams.kt
+++ b/libraries/stdlib/samples/test/samples/streams/streams.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package samples.streams
+
+import samples.Sample
+import samples.assertPrints
+import java.util.stream.DoubleStream
+import java.util.stream.IntStream
+import java.util.stream.LongStream
+import java.util.stream.Stream
+import kotlin.streams.asSequence
+import kotlin.streams.asStream
+import kotlin.streams.toList
+
+class Streams {
+
+    @Sample
+    fun streamAsSequence() {
+        val stringStream = Stream.of("Never", "gonna", "give", "you", "up")
+        val stringSequence = stringStream.asSequence()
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/streams/streams.kt
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2010-2018 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package samples.streams
+
+import samples.Sample
+import samples.assertPrints
```
ilya-g, line 9: Please use a star-import from `samples` package, this is a requirement of the runnable sample converter.
For the other two packages, a star-import won't hurt either.

ilya-g (commented):
For the samples of conversion functions, e.g. `toList` or `asSequence`, it may be a good idea to show the types of the original value and the resulting value.

libraries/stdlib/samples/test/samples/streams/streams.kt
uzilan: Thanks for the comment. I will change the samples imports to a star but when it comes to the rest I find it much easier to code review when the imports are explicit. That way you always know which classes are being used. Don't you agree?

ilya-g: The way how the runnable sample converter works is that it just copies all imports from a file, no matter whether they are used in the particular sample or not. Also in a short sample where at most two-three types are used, it usually is clear where they come from.
So importing all types explicitly just adds noise to a sample.

libraries/stdlib/samples/test/samples/streams/streams.kt
```diff
@@ -0,0 +1,82 @@
...
+import kotlin.streams.asStream
+import kotlin.streams.toList
+
+class Streams {
+
+    @Sample
+    fun streamAsSequence() {
+        val stringStream = Stream.of("Never", "gonna", "give", "you", "up")
+        val stringSequence = stringStream.asSequence()
+        assertPrints(stringSequence.joinToString(" "), "Never gonna give you up")
```
johanrydstrom, line 18: 😁

ilya-g (approved):

ilya-g:
Thank you!

PR #3761, opened 2020-09-28
FIR: carry annotations on delegated property accessors

jsjeon:
The motivation is bb test `annotations/delegatedPropertySetter`:
```kt
class MyClass() {
    public var x: String by Delegate()
        @First set
}
```
The annotation `@First` was parsed, converted, and stored at a property setter, but when (re)generating property accessors by delegate, it's simply not carried.


mglukhikh:
Merged 👍

PR #3762, opened 2020-09-29
KT-42255 "Replace elvis expression with 'if' expression" intention shouldn't introduce unnecessary variable if 'error' expression is used

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-42255


dimonchik0036 (approved):
Thanks!

PR #3763, opened 2020-09-29
[FIR] Enable more codegen test suites

punzki:
We are at around 5% failing for the current blackbox tests. Enabling more tests to get more coverage and add more potential work. :slightly_smiling_face:

BlackBoxInlineCodegen tests: 16 out of 638 tests (2.5%) are currently failing.
BlackBoxAgainstJavaCodegen tests: 23 out of 156 tests (14.7%) are currently failing.
BytecodeText tests: 143 out of 767 tests (18.6%) are currently failing.


compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractFirBlackBoxAgainstJavaCodegenTest.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors.
```
jsjeon, line 2: Out of curiosity, is this auto generated? I wonder it should be 2020 or somewhere in the generator is using an old one?

mglukhikh (approved):

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractFirBlackBoxAgainstJavaCodegenTest.kt
mglukhikh: I think it was just copied from `AbstractFirBlackBoxCodegenTest.kt`. "New File" generates correct comment, at least on my PC.

mglukhikh:
Merged manually. I think these tests should be helpful, thank you.

mglukhikh:
BTW, I updated these years Jinseong has pointed to.

compiler/tests-common/tests/org/jetbrains/kotlin/codegen/ir/AbstractFirBlackBoxAgainstJavaCodegenTest.kt
punzki: Yes, you caught me copy-pasting. 😁

compiler/testData/codegen/bytecodeText/lineNumbers/ifConsts.kt
```diff
@@ -1,3 +1,4 @@
+// IGNORE_BACKEND_FIR: JVM_IR
```
pyos, line 1: this shifts the line numbers, so either the line needs to be at the bottom or the expected results should be updated

mglukhikh: Fixed, thank you.

mglukhikh: BTW I unmuted 15 FIR bytecode text tests in process (they were muted directly because of this line shift).

punzki: Oh sorry about that! I should've known better. 🙁  Thanks @pyos for the callout and @mglukhikh for the quick fix!

PR #3764, opened 2020-09-29
JVM_IR: fix copying of receivers when unboxing inline class parameters

pyos:
If an extension receiver is copied into a normal parameter, this may later cause default argument masks to be shifted.


udalov:
Please regenerate tests, and mute the newly added test for FIR if necessary

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/inlineclasses/MemoizedInlineClassReplacements.kt
```diff
@@ -137,18 +137,19 @@ class MemoizedInlineClassReplacements(private val mangleReturnTypes: Boolean, pr
             require(function.dispatchReceiverParameter != null && function is IrSimpleFunction)
             val newValueParameters = ArrayList<IrValueParameter>()
             for ((index, parameter) in function.explicitParameters.withIndex()) {
```
udalov, line 139: Maybe don't use `explicitParameters` at all now and handle dispatch/extension receivers outside of the loop (which can go only through `valueParameters`)?

pyos: Good point, and the loop below doesn't really need to construct this list either.

udalov:
Thanks!

PR #3765, opened 2020-09-29
Sort class members to ensure deterministic builds

hungvietnguyen:
In an incremental build, class members are already sorted when they are
deserialized from the serialized data (see
DeserializedMemberScope.addMembers).

However, in a clean build where the class members are read directly from
the source files, the class members are not yet sorted. Therefore, there
can be a difference in the contents of the stubs file between a clean
build and the next incremental build, making the subsequent task run
unnecessarily.

To fix that, this commit ensures that we always sort the class members.

Bug: KT-40882 (there are actually 2 issues in here; this commit fixes
     the first one)
Test: New DeterministicBuildIT

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/DeterministicBuildIT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/DeterministicBuildIT.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.gradle
+
+import org.jetbrains.kotlin.gradle.util.allJavaFiles
+import org.junit.Test
+import java.io.File
+import kotlin.test.assertEquals
+
+/** Tests that the outputs of a build are deterministic. */
+class DeterministicBuildIT : BaseGradleIT() {
+
+    @Test
+    fun `test KaptGenerateStubsTask - KT-40882`() = with(
+        Project("simple", directoryPrefix = "kapt2")
+    ) {
+        setupWorkingDir()
+        projectDir
+            .resolve("src/main/java/Foo.kt")
+            .writeText(
+                """
+                class Foo : Bar {
+                    // The fields and methods are ordered such that any sorting by KGP will be detected.
+                    val fooField1 = 1
+                    val fooField3 = 3
+                    val fooField2 = 2
+                    fun fooMethod1() {}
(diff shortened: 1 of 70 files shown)


compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/descriptors/AbstractLazyMemberScope.kt
```diff
@@ -138,12 +139,12 @@ protected constructor(
...
     }
 
-    private fun doGetProperties(name: Name): Collection<PropertyDescriptor> {
+    private fun doGetProperties(name: Name): List<PropertyDescriptor> {
         val result = LinkedHashSet(declaredPropertyDescriptors(name))
 
         getNonDeclaredProperties(name, result)
 
-        return result.toList()
+        return sort(result.toList())
```
gavra0, line 147: Note: I don't think this should not incur measurable overhead. Sorting only when incremental compilation is enabled seems like an overkill.

gavra0 (approved):

compiler/frontend/src/org/jetbrains/kotlin/resolve/lazy/descriptors/AbstractLazyMemberScope.kt
hungvietnguyen: Ack. (Btw, I've changed where we sort this in my latest commit.)

gavra0 (approved):

gavra0:
@yanex Can you please take a look as well?

hungvietnguyen:
@yanex Hi Yan, this PR ensures that Kotlin tasks are correctly UP-TO-DATE. As you know, there is a huge impact on build time when a task should be UP-TO-DATE but runs unnecessarily because of this issue. It would be great to get your feedback on this.

(If you are busy with other work, can you maybe recommend someone else to look at this? Thanks!)

hungvietnguyen:
@nav-nav Hi Nataliya, I've updated this PR to fix the previously failed tests that you found. Can you please check if it is passing this time?

Also, can you or someone else review this updated PR? It looks like a large one, but actually there are only 3 types of changes:
1. `ClassFileToSourceStubConverter.kt`: The main change to add sorting (a few lines of code)
2. `DeterministicBuildIT.kt`: Integration test
3. `plugins/kapt3/kapt3-compiler/testData/**/*.txt`: Various expectation files need to be updated as the contents of stub files are now sorted.

// cc @gavra0 in case you want to take another look.

Thank you!

hungvietnguyen:
@nav-nav Hi Nataliya, I just updated this PR to fix these failed tests: `Kapt3WorkersIT.testLocationMapping`, `Kapt3IT.testLocationMapping`. Please take another look.

nav-nav:
Cherry-picked into master

hungvietnguyen:
Commit in master: https://github.com/JetBrains/kotlin/commit/4bf63a953958fc1342312274cbdfe23d085347a3.

Thank you, Nataliya!

PR #3766, opened 2020-09-29
FIR: fix vararg remapping that merged named vararg

jsjeon:
The motivation is bb test `vararg/assigningArrayToVarargInAnnotation`, the last failure in `vararg`:
```kt
@interface JavaAnn {
    String[] value() default {};
    String[] path() default {};
}

@JavaAnn(value = ["value5"], path = ["value6"])

@JavaAnn("value7", path = ["value8"])
```
where arguments for those two annotation calls are merged, like `["value5", "value6"]`. When remapping arguments with `vararg`, we should not pull out of named `vararg`.


jsjeon:
Rebased atop https://github.com/JetBrains/kotlin/pull/3755

mglukhikh:
Merged manually 👍

PR #3768, opened 2020-09-29
FIR serializer: special handling of Continuation and typealias

jsjeon:
During serialization, suspend functions are transformed to runtime function types. Serializing types that came from such transformation fails currently, since we can't find the corresponding classifier symbol if coroutine lib is not given. We can avoid that specific case or in general.

Also, when classifier lookup succeeds, we can distinguish either class classifier or `typealias` classifier and set appropriate fields.


mglukhikh:
Looks good, merged, thank you. 👍

PR #3769, opened 2020-09-30
[IR] Avoid extra lists creation during getting explicit/all parameters

LepilkinaElena:


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +199,12 @@ val IrClass.primaryConstructor: IrConstructor?
 val IrDeclarationContainer.properties: Sequence<IrProperty>
     get() = declarations.asSequence().filterIsInstance<IrProperty>()
 
-val IrFunction.explicitParameters: List<IrValueParameter>
-    get() = (listOfNotNull(dispatchReceiverParameter, extensionReceiverParameter) + valueParameters)
+val IrFunction.explicitParameters: MutableList<IrValueParameter>
+    get() = mutableListOf<IrValueParameter>().apply {
```
romanart, line 203: This change seems like something made due to performance reasons. If so I think it good if exact list size is computed before creation to avoid extra memory allocations.

SvyatoslavScherbina: Approximation might be good enough too.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +199,12 @@ val IrClass.primaryConstructor: IrConstructor?
 val IrDeclarationContainer.properties: Sequence<IrProperty>
     get() = declarations.asSequence().filterIsInstance<IrProperty>()
 
-val IrFunction.explicitParameters: List<IrValueParameter>
-    get() = (listOfNotNull(dispatchReceiverParameter, extensionReceiverParameter) + valueParameters)
+val IrFunction.explicitParameters: MutableList<IrValueParameter>
```
SvyatoslavScherbina, line 202: Please avoid making return type `MutableList`.
Consider declaring `IrFunction.addExplicitParametersTo(MutableList)` and using it here and in `allParameters` instead.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +199,16 @@ val IrClass.primaryConstructor: IrConstructor?
 val IrDeclarationContainer.properties: Sequence<IrProperty>
     get() = declarations.asSequence().filterIsInstance<IrProperty>()
 
+fun IrFunction.addExplicitParametersTo(parametersList: MutableList<IrValueParameter>) {
+    dispatchReceiverParameter?.let { parametersList.add(it) }
```
SvyatoslavScherbina, line 203: `parametersList.addIfNotNull(dispatchReceiverParameter)`?
Same below.

SvyatoslavScherbina (approved):

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +200,16 @@ val IrClass.primaryConstructor: IrConstructor?
 val IrDeclarationContainer.properties: Sequence<IrProperty>
     get() = declarations.asSequence().filterIsInstance<IrProperty>()
 
+fun IrFunction.addExplicitParametersTo(parametersList: MutableList<IrValueParameter>) {
```
romanart, line 203: please make it `private`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -206,6 +206,12 @@ fun IrFunction.addExplicitParametersTo(parametersList: MutableList<IrValueParame
     parametersList.addAll(valueParameters)
 }
 
+internal fun Boolean.toInt(): Int = if (this) 1 else 0
```
SvyatoslavScherbina, line 209: Isn't `private` enough here?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +200,22 @@ val IrClass.primaryConstructor: IrConstructor?
...
 
+fun IrFunction.addExplicitParametersTo(parametersList: MutableList<IrValueParameter>) {
+    parametersList.addIfNotNull(dispatchReceiverParameter)
+    parametersList.addIfNotNull(extensionReceiverParameter)
+    parametersList.addAll(valueParameters)
+}
+
+internal fun Boolean.toInt(): Int = if (this) 1 else 0
+
+val IrFunction.explicitParametersNumber: Int
```
SvyatoslavScherbina, line 211: Maybe `explicitParametersCount`?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -199,8 +200,22 @@ val IrClass.primaryConstructor: IrConstructor?
...
+    get() = (dispatchReceiverParameter != null).toInt() + (extensionReceiverParameter != null).toInt() +
+            valueParameters.size
+
 val IrFunction.explicitParameters: List<IrValueParameter>
-    get() = (listOfNotNull(dispatchReceiverParameter, extensionReceiverParameter) + valueParameters)
+    get() = ArrayList<IrValueParameter>(valueParameters.size + 2).also {
```
SvyatoslavScherbina, line 216: (Not sure if it makes sense but) now it is easy to use exact size here. Same for `allParameters`.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/ir/IrUtils.kt
```diff
@@ -432,14 +431,20 @@ fun IrFunction.createDispatchReceiverParameter(origin: IrDeclarationOrigin? = nu
 
 val IrFunction.allParameters: List<IrValueParameter>
     get() = if (this is IrConstructor) {
-        listOf(
-            this.constructedClass.thisReceiver
-                ?: error(this.render())
-        ) + explicitParameters
+        ArrayList<IrValueParameter>(valueParameters.size + 3).also {
```
SvyatoslavScherbina, line 434:
```suggestion
        ArrayList<IrValueParameter>(allParametersCount).also {
```

SvyatoslavScherbina (approved):

PR #3773, opened 2020-10-01
FIR CFG: make finally block post-dominate try/catch blocks

jsjeon:
The motivation is https://youtrack.jetbrains.com/issue/KT-42350:
```kt
fun println(arg: Int) {}

fun foo() {
    val x: Int
    try {
        x = 0
    } finally {}

}
```
It turns out that, at the try-expression exit node, its incoming edges from `try` block and `finally` block are treated equally. That's not true because `finally` block post-dominates `try` block and `catch` clauses _unless_ they escape, e.g., `return` or rethrow exceptions.

1st commit reproduces the issue; 2nd commit adjusts edges between try/catch/finally so that `finally` block can post-dominates others; and 3rd commit handles escaping cases, inspired by the regression from the 2nd commit (`diagnostics/.../kt2845`).


demiurg906:
Hi @jsjeon.

Before your changes finally blocks was not properly supported in CFG, but with them they are still incorrect. Check this case from your tests:
```kotlin
fun foo(s: Any) {
    val a: Int
    try {
        s as String
    } catch (e: Exception) {
        s as String
    }
    finally {
        s.length // OK, should be an error
    }
    s.length // OK, should be OK
}
```

For this code you build next graph:

assignedInTryWithCatch

Here in perspective of build graph we can enter to `finally` after full execution of `try` or `catch` block, so there is smartcast, but it's incorrect, because e.g. in `try` block before cast was throw some `Error`, so we don't enter to `catch` and enter directly to `finally`. So, my idea is that you can't just remove edge from `EnterTryMainBlock` to `EnterFinallyBlock`.

Actually `finally` blocks are very hard to implement because of their semantics and implementation: in generated code we have a copy of finally block to each possible exit from function. But in source code we have only one `finally` block and we want to analyze it only once and build graph for it only once too, because building graph inextricably connected with code analysis. Note that in FE 1.0 we build graph after analysis, and there graph had separate copies for each try exit.

When I was thinking about this problem I found next solution: when we meet `try` with `finally` then we can create only one copy of CFG for `finally` block, but for different enters to it we can create colored edges, so when CFG traverser enters to finally block by edge of one color so it should leave it only by edge of same color. This is like simulation of multiple copies of `finally`. See example:

```kotlin
fun test() {
    ...
    try {
        ...
    } catch (e: IOException) { // catch 1
        ...
    } catch (e: ClassCastException) { // catch 2
        ...
        return
    } finally {
        ...
    }
    ...
}
```

TryWithFinallyCfg

But I still didn't implement this decision, so you can try to do it.

demiurg906 (changes requested):
Also note that you can dump CFG for every diagnostic test that you want using `// !DUMP_CFG` directive inside testdata. With it testrunner dumps graph to `.dot` file, which you can observe using any dot tool you want. There is an instruction in `AbstractFirDiagnosticsTest` how to add action to IDEA which opens graph dump in XDot viewer for opened `.kt` testdata file (works for linux and mac). Also there is a plugin for .dot files for IDEA, but it very useless to me, so I recommend using XDot or it's alternative

jsjeon:
Thanks @demiurg906 for the thorough reviews (and drawings)!

I accidentally(?) dived into another checker-related issue, so before doing the context switching, I may dig it further until reaching some checking points. Then, I'll revisit this and ponder a bit more. While working on that, I also learned `// !DUMP_CFG` and `XDot` external tool setting; thanks for the kind reminder.

I also noticed that side effects of every single expression in each block are not considered at all. And I agree that it's not preferred to link _all_ of those expressions with side effects to finally block enter. So, I admit we need a connection from try-main enter to finally bloc enter somehow sometimes, but not always.

Going back to the original motivation, well, `CfgTraverser` and all the related logic in checkers seem not path-sensitive, so it would be a huge change to make them be aware of paths. Even with the simulated path from try-main through finally block to the qualified access, we will have things like:
```
{ path1 (direct to finally, assuming exception inside try block): {x: ZERO},
  path2 (full execution of try block and then finally): {x: EXACTLY_ONCE} }
```
and the best approximation is that x is initialized `AT_MOST_ONCE`, resulting in `UNINITIALIZED_VARIABLE`. I still believe, in this case, an edge from try-main to finally is spurious, since the try block is side effect free. And, not having such spurious paths, which I tried to do at the 2nd commit, will let the checker see the correct information.

So, when I resume, I'll start with side effect checks first, which I think will be able to cover your example with cast exception (or any other side effects) in try/catch blocks. Labeling or coloring different paths can be done on top of it, then making CFG traverser path-sensitive would be done after that. Well, sounds like the last two topics are worth being a separate/follow-up PR? :)

demiurg906:
>  well, CfgTraverser and all the related logic in checkers seem not path-sensitive, so it would be a huge change to make them be aware of paths. Even with the simulated path from try-main through finally block to the qualified access, we will have things like

Yes, CFA is not ready for such changes in part of traversing graph, this logic should be implemented somehow

>  I still believe, in this case, an edge from try-main to finally is spurious, since the try block is side effect free.

> So, when I resume, I'll start with side effect checks first

Side-effect free is a very rare case, because almost every expression or statement in Kotlin may have side effects: set or get variable may have custom accessors/delegate, each function call may throw exception. So, I don't think it's worth to add some complex logic to very few usecases. So I don't think it's a good idea to implement some side-effect analysis

> Labeling or coloring different paths can be done on top of it, then making CFG traverser path-sensitive would be done after that. Well, sounds like the last two topics are worth being a separate/follow-up PR

I suggest you to implement coloring for specific case of `finally` blocks and update CFA according to those changes and send it as one PR. This request I will discard

jsjeon:
Hm, I expected using the same branch would end up with the same PR, and planned to revamp this. Apparently, it's not. PTAL my other rework here: https://github.com/JetBrains/kotlin/pull/3813

PR #3774, opened 2020-10-01
Avoid iterating over all idea modules.

nskvortsov:
Use hash table to quickly lookup idea module by gradle project path.

Colleagues from Android Studio team have a CPU snapshot, where line `KotlinGradleProjectResolverExtension.kt:161` takes almost 3 minutes (166 000 ms) of CPU time during import. This patch should mitigate the problem


vladimirdolzhenko:
LGTM

PR #3775, opened 2020-10-01
Implemented the merge function for maps

excitement-engineer:
The merge function on maps is something that I use a lot. It is similar to the zip extension function on List. 

Let me know if you think that this could be a candidate for the standard library.

### DIFF
--- a/libraries/stdlib/src/kotlin/collections/Maps.kt
+++ b/libraries/stdlib/src/kotlin/collections/Maps.kt
@@ -777,6 +777,26 @@ public inline operator fun <K, V> MutableMap<K, V>.minusAssign(keys: Sequence<K>
     this.keys.removeAll(keys)
 }
 
+/**
+ * Returns a map containing the keys of `this` map and the [other] map. Where there are values for a certain key the transform function is
+ * applied. If there is a value for only key in a map then the value of the corresponding entry is taken.
+ */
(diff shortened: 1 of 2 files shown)


AlexJuca:
@excitement-engineer I believe changes to the standard library need additional input from the core team. @ivakub Is this correct?

libraries/stdlib/src/kotlin/collections/Maps.kt
```diff
@@ -777,6 +777,26 @@ public inline operator fun <K, V> MutableMap<K, V>.minusAssign(keys: Sequence<K>
     this.keys.removeAll(keys)
 }
 
+/**
+ * Returns a map containing the keys of `this` map and the [other] map. Where there are values for a certain key the transform function is
+ * applied. If there is a value for only key in a map then the value of the corresponding entry is taken.
+ */
+public inline fun <K, V>Map<out K, V>.merge(other: Map<out K, V>, transform: (a: V, b: V) -> V): Map<K, V> {
```
ilya-g, line 784: The name is to be discussed. Currently, when you use JDK8 or higher, there's already [MutableMap.merge](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#merge-K-V-java.util.function.BiFunction-) function available coming from JDK. It has the similar signature, but applies the merge operation in-place.
It would be very confusing to have two `merge` operations on a mutable map, where one merges in-place and the other returns a new map.

PR #3776, opened 2020-10-01
FIR checker: report val reassignment

jsjeon:
While working on false positives from `FirPropertyInitializationAnalyzer`, I noticed we missed to report `val` reassignment. This is a simple extension that visits variable assignments and reports an error if the lvalue of the assignment is `val` and can be revisited.

In most cases, `// FIR_IDENTICAL` are added, which is good, but I can see a couple of false postives as well. Some are related to CFG construction for try/catch/finally: discussion about that is on-going at https://github.com/JetBrains/kotlin/pull/3773. Some are very deep, like considering loop condition as well. Will leave comments on test files directly.


compiler/fir/analysis-tests/testData/extendedCheckers/CanBeValChecker.kt
```diff
@@ -105,9 +105,9 @@ fun foo() {
     <!CAN_BE_VAL!>var<!> <!VARIABLE_NEVER_READ!>a<!>: Int
     val bool = true
     if (bool) <!ASSIGNED_VALUE_IS_NEVER_READ!>a<!> = 4 else <!ASSIGNED_VALUE_IS_NEVER_READ!>a<!> = 42
-    val <!UNUSED_VARIABLE!>b<!>: String
+    val <!VARIABLE_NEVER_READ!>b<!>: String
 
-    <!ASSIGNED_VALUE_IS_NEVER_READ!>bool<!> = false
```
jsjeon, line 110: `bool` has been read above, so I'm pretty much sure the intention was `b`, not `bool`.

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
```diff
@@ -3,21 +3,21 @@
 fun test() {
     do {
         val s: String
-        s = ""
+        <!VAL_REASSIGNMENT!>s<!> = ""
```
jsjeon, line 6: Due to the loop ending condition below (`s == ""`), this shouldn't be the case, but no idea yet where to look at, how to integrate value assignment to CFG building/pruning, etc.

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
```diff
@@ -3,21 +3,21 @@
...
     } while (s == "")
 }
 
 fun test2() {
     do {
         val s: String
-        s = "1"
-        s = s + "2"
+        <!VAL_REASSIGNMENT!>s<!> = "1"
+        <!VAL_REASSIGNMENT!>s<!> = s + "2"
```
jsjeon, line 14: These look quite strange to me. The 2nd assignment is definitely `VAL_REASSIGNMENT`, but because of that, the loop ending condition, `s == "1"`, holds and the 1st assignment is regarded as valid assignment at old FE.

compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.fir.kt
```diff
@@ -23,14 +23,14 @@ fun f2(flag: Boolean) {
         val n: Int
         try {
             if (flag) {
-                n = 1
+                <!VAL_REASSIGNMENT!>n<!> = 1
```
jsjeon, line 26: Due to the `break` below, this one wouldn't be revisited via the outer `while` loop. I'll keep this in mind while re-working https://github.com/JetBrains/kotlin/pull/3773

compiler/testData/diagnostics/tests/controlFlowAnalysis/varInitializationInIfInCycle.kt
```diff
@@ -3,10 +3,10 @@ fun foo(numbers: Collection<Int>) {
     for (i in numbers) {
         val b: Boolean
         if (1 < 2) {
-            b = false
+            <!VAL_REASSIGNMENT!>b<!> = false
         }
         else {
-            b = true
+            <!VAL_REASSIGNMENT!>b<!> = true
```
jsjeon, line 9: I guess this one (and below) is about the wrong CFG construction for `while` loop v.s. `for` loop where their bodies are (non) transient?

compiler/testData/diagnostics/tests/when/ExhaustiveValOverConditionalInit.fir.kt
```diff
@@ -4,8 +4,8 @@ fun foo(a: Boolean, b: Boolean): Int {
         x = 1
     }
     when (b) {
-        true -> x = 2
-        false -> x = 3
+        true -> <!VAL_REASSIGNMENT!>x<!> = 2
+        false -> <!VAL_REASSIGNMENT!>x<!> = 3
```
jsjeon, line 8: This 2nd error raises the question on how often we want to report errors. It seems like old FE reported an error only once for the above `true` case.

compiler/testData/diagnostics/testsWithStdLib/contracts/controlflow/flowInlining/breakContinuesInInlinedLambda.fir.kt
```diff
@@ -33,12 +33,12 @@ fun test() {
...
                 myRun { x.inc() }
 
-                myRun { x = 42 }
+                myRun { <!VAL_REASSIGNMENT!>x<!> = 42 }
                 break
             }
             // x is I?D here because loop could've been execited
             // VAL_REASSIGNMENT isn't reported because of repeating diagnostic
-            x = 42
+            <!VAL_REASSIGNMENT!>x<!> = 42
```
jsjeon, line 41: Aha, very clear evidence here. According to the comment, the repeating diagnostic error is not reported. So, this one is `VAL_REASSIGNMENT`, but old FE somehow collects and reports an error per error kind maybe.

compiler/testData/diagnostics/testsWithStdLib/contracts/controlflow/flowInlining/labeledReturns.fir.kt
```diff
@@ -42,7 +42,7 @@ fun threeLevelsReturnNoInitialization(x: Int?): Int? {
             }
         }
         // Possible to report unreachable here
-        y = 54
+        <!VAL_REASSIGNMENT!>y<!> = 54
```
jsjeon, line 45: See the comments above that explain the complex structure of inner-outer entanglement. No idea yet...

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
mglukhikh: This diagnostic is definitely wrong and it prevents this PR from being accepted.

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
mglukhikh: This (the first one) too.

compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.fir.kt
mglukhikh: We shouldn't have reassignment here

mglukhikh: But *should* have it reported in catch, both in `f1` and in `f2`

compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.fir.kt
```diff
@@ -41,14 +41,14 @@ fun f3(flag: Boolean) {
...
```
mglukhikh, line 44: This one is also wrong

compiler/testData/diagnostics/tests/controlFlowAnalysis/varInitializationInIfInCycle.kt
mglukhikh: Both of these two are wrong

compiler/testData/diagnostics/tests/when/ExhaustiveValOverConditionalInit.fir.kt
mglukhikh: I think yet we can report both of them, it's not a major problem.

compiler/testData/diagnostics/testsWithStdLib/contracts/controlflow/flowInlining/labeledReturns.fir.kt
mglukhikh: Probably FE 1.0 doesn't report reassignment here because the code is unreachable

compiler/testData/diagnostics/testsWithStdLib/contracts/controlflow/initialization/atLeastOnce/varIndefiniteIntialization.fir.kt
```diff
@@ -24,6 +24,6 @@ fun indefiniteFlow() {
 
 fun shadowing() {
     var x: Int
-    runTwice { val x: Int; x = 42; x.inc() }
+    runTwice { val x: Int; <!VAL_REASSIGNMENT!>x<!> = 42; x.inc() }
```
mglukhikh, line 27: This one is also wrong

mglukhikh (changes requested):
I see that all problems are provoked by CFG building and/or analyzing bugs and not your changes directly. However, we cannot merge this PR because it will add some false errors that can break compilation of green code. We have two choices:

* wait until @demiurg906 or somebody other of us will fix these bugs
* you can try to fix them yourself

demiurg906:
As I can see, there is no problems with building graph, but they are with collecting data for variable assign info: we don't consider scope where variable was declared when traversing the graph until fixed point.

Right now I have some other stuff to do, so it may take long time to fix it by me

jsjeon:
Thanks to the work done for https://youtrack.jetbrains.com/issue/KT-44804, in particular, `PreliminaryLoopVisitor`, I got an idea of avoiding all the false alarms this PR had before: valid `val` assignment inside a loop, which is visited/accumulated again and again. E.g.,

```kt
do {
  val x: String
  x = ""  // x -> [1, 1] -> [2, 2] -> [3, 3] -> MORE_THAN_ONCE
} while (x == "")
```

So, inspired by `PreliminaryLoopVisitor`, we can collect what variables are declared in capturing statements. Note that the one in FIR DFA collects potentially reassigned variable names. So the collected info/position is slightly different.

Then, in loop's merge points, where two flows---the one from outside of the loop and the other from the loop body itself---will merge, we can (partially) filter out variable init info for those that are _declared_ inside the loop. The caveat is, merge points for loops vary: all different kinds of loops---`for`, `while`, `do-while`---have different positions. Also, the timing of retrieving variables declared in the loop v.s. the timing of visiting/processing loop merge points are different.

To that end, we need a couple changes to CFG as well. First, when loop condition node is created, the containing loop expression is also recorded, in addition to the loop condition expression. So, when data-flow collector visits the loop condition node, which is a merge point for some loop kinds, it can retrieve what variables are declared in the loop.

Second, when back edges are added for loops, we need a special edge label so that the flow from the loop body can be differentiated. Since it's literally "back" edge, we can use that edge kind too, but CFG traversals abstract away edges; edge labels are bound instead (to handle exception handling paths in `try-catch-finally`), hence a new edge label for loop-back edge.

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
jsjeon: It's gone! (And all others!)

compiler/testData/diagnostics/tests/controlFlowAnalysis/doWhileAssignment.fir.kt
jsjeon: `// FIR_IDENTICAL` !

compiler/testData/diagnostics/tests/controlFlowAnalysis/varInitializationInIfInCycle.kt
jsjeon: Not reported anymore.

compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.fir.kt
jsjeon: No longer reported.

compiler/testData/diagnostics/testsWithStdLib/contracts/controlflow/initialization/atLeastOnce/varIndefiniteIntialization.fir.kt
jsjeon: Not reported anymore.

compiler/tests-spec/testData/diagnostics/notLinked/contracts/analysis/controlFlow/initialization/pos/3.fir.kt
```diff
@@ -155,7 +155,7 @@ fun case_10() {
     val x: Int
     funWithExactlyOnceCallsInPlace outer@ {
         funWithAtLeastOnceCallsInPlace {
-            x = 42
+            <!VAL_REASSIGNMENT!>x<!> = 42
```
jsjeon, line 158: This is the only one that differs from FE 1.0 behavior, but I can't tell whether it's a false alarm or not. Due to the labeled return below, this lambda exits to outer, and thus won't be invoked multiple times. However, as the function name indicates, I guess the contract for the caller is that the lambda can be invoked _at least_ once, hence the event occurrence count for this variable would be `AT_LEAST_ONCE`, which also means this can be revisited.

I don't think there is anything we can do regarding CFG/contract here. cc @demiurg906

compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.fir.kt
jsjeon: > We shouldn't have reassignment here

Yep, no longer report anything here.

> But _should_ have it reported in catch, both in f1 and in f2

Nothing reported at all, but I'm not sure why we _should_ report `val` reassignment in this test file:
https://github.com/JetBrains/kotlin/blob/master/compiler/testData/diagnostics/tests/controlFlowAnalysis/reassignmentInTryCatchWithJumps.kt

So, every `n = 3` in `catch` clause can be reached if, e.g., there is OOM thrown before `n = 1`.

jsjeon: Okay, after thinking about these a bit more, I have to admit we have no false positives (which is good), but yet not reporting those assignments in `catch` clauses are false negatives because every `n = 3` in `catch` clause can be reached if, e.g., there is OOM thrown _after_ `n = 1`.

The pattern of test functions in this file is that `try` main block has a `when` expression whose main part will escape the `try` main block (through `return`, `continue`, or `break`). Because of that, the `WhenBranchResultExitNode` is marked as dead, and the edge from it is also `DeadForward`, which gets that node discarded when collecting previous CFG nodes for the `when` exit node. Thus, assignment info inside `if` is always discarded. Therefore, assignment info inside `catch` clause is always `AT_MOST_ONCE`, which should be `AT_LEAST_ONE` instead.

So, this is related TODOs I left at https://github.com/JetBrains/kotlin/commit/440cf7888460aa4d691c15c3dfe358465a1e59fb: when adding jump nodes, if they're inside try-catch-finally, we should be more careful. As shown in this case, we need an edge towards the next node (for unexpected error at any place) in addition to the edge toward the jump target (we're adding this only for now). I'll keep this thread of comments open, not to forget the issue, and will try a bit more while waiting for the reviews. But, since there are no false errors, this PR should be good to go too.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
```diff
@@ -120,10 +127,156 @@ class PropertyInitializationInfoCollector(private val localProperties: Set<FirPr
...
+
+            override fun visitAnonymousFunction(anonymousFunction: FirAnonymousFunction, data: FirStatement?) {
+                visitCapturingStatement(anonymousFunction, data)
+            }
+
+            override fun visitSimpleFunction(simpleFunction: FirSimpleFunction, data: FirStatement?) {
+                visitCapturingStatement(simpleFunction, data)
+            }
+
+            override fun <F : FirFunction<F>> visitFunction(function: FirFunction<F>, data: FirStatement?) {
```
demiurg906, line 192: You don't need to override `visitFuinction` after you override `visitXXX` for all `FirFunction<*>` inheritors

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
```diff
@@ -120,10 +127,156 @@ class PropertyInitializationInfoCollector(private val localProperties: Set<FirPr
 
     private fun processVariableWithAssignment(
         dataForNode: PathAwarePropertyInitializationInfo,
-        symbol: FirPropertySymbol
+        symbol: FirPropertySymbol,
+        reset: Boolean = false,
```
demiurg906, line 131: `reset` is very unclear name. Maybe rename it to something like `overwriteRange`?

demiurg906 (approved):

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
jsjeon: Right, thanks for the suggestion. Changed.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
jsjeon: Since lambda and local declarations are targets to visit, I think overriding `visitAnonymousFunction` and `visitSimpleFunction` is good enough (and that's what/why `PreliminaryLoopVisitor` did so). This general one, though, is added while mimicking `PreliminaryLoopVisitor`, so I guess it's safe to remove both ones? :) Added as the 2nd commit.

demiurg906:
Full pipeline test is hanging with this changes. Can you investigate cause of it?

jsjeon:
> Full pipeline test is hanging with this changes. Can you investigate cause of it?

Thank you for catching that. On my side, with an old test data, around the end of full pipeline test, it failed with a GC limit error. Then, I noticed that I defined the declared variable collector as an object inside property init info collector. Note that `PreliminaryLoopVisitor` is a regular class, and instantiated on demand.

So, I assumed such design---inner object---made that capturing statement visitor (and its cached map) is not properly garbage-collected when its owner, property init info collector, finished its job. Defining the declared variable collector as a regular class and instantiating it when property init info collector is instantiated (which now align the lifespan of those two collectors) resolved the issue on my side. Let me know that also works for you.

demiurg906:
Merged manually

PR #3777, opened 2020-10-01
FIR CFG traverser: do not skip graph enter node

jsjeon:
The motivation is https://youtrack.jetbrains.com/issue/KT-42348:
```kt
fun println(arg: Int) {}

fun foo(arg: Boolean) {
    val x: Int
    if (arg) {
        x = 1
    } else {
        x = 2
    }

    class Local {
        fun bar() {
        }
    }
}
```
It turns out that `CfgTraverser` skipped the enter node of subgraph, while visiting all other nodes in that subgraph. This causes inaccurate data-flow information propagation because information on incoming edges to the entrance of the subgraph is ignored.

In fact, skipping the entrance of graph is not necessary, since `initialiInfo` is going to be used anyway.


demiurg906:
@jsjeon Thank you. I merged request manually

PR #3778, opened 2020-10-01
Added samples for the kotlin.system package

uzilan:
Added samples for the kotlin.system package

### DIFF
--- a/libraries/stdlib/samples/test/samples/system/timing/timing.kt
+++ b/libraries/stdlib/samples/test/samples/system/timing/timing.kt
@@ -2,17 +2,17 @@ package samples.system.timing
 
 import samples.*
 import kotlin.system.*
-import kotlin.test.*
 
 class Timings {
 
     @Sample
     fun measureBlockTimeMillis() {
         val timeInMillis = measureTimeMillis {
-            (0..10).reduce(Int::plus)
+            val numbers = (0..100).toMutableList()
+            numbers.shuffle()
+            numbers.sort()
         }
-        assertTrue(timeInMillis >= 0)
-        assertTrue(timeInMillis < 1e9)
+        println("The operation took $timeInMillis ms")
     }
 
     @Sample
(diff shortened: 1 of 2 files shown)


libraries/stdlib/samples/test/samples/system/timing/timing.kt
```diff
@@ -0,0 +1,26 @@
...
+import samples.*
+import kotlin.system.*
+import kotlin.test.*
+
+class Timings {
+
+    @Sample
+    fun measureBlockTimeMillis() {
+        val timeInMillis = measureTimeMillis {
+            (0..10).reduce(Int::plus)
```
ilya-g, line 12: I suppose given the resolution of millisecond timer this result is going to be 0 every time. Perhaps we can find some other task that lasts for 10-20 ms in Kotlin playground?

libraries/stdlib/samples/test/samples/system/timing/timing.kt
```diff
@@ -0,0 +1,26 @@
...
+        }
+        assertTrue(timeInMillis >= 0)
+        assertTrue(timeInMillis < 1e9)
```
ilya-g, line 15: I think such assertions are not that informative for the purpose of sample. Would be better just to print out the value, something like `println("The operation took $timeInMillis ms")`

libraries/stdlib/samples/test/samples/system/timing/timing.kt
uzilan: Cool. I'll try to find something else to use.

libraries/stdlib/samples/test/samples/system/timing/timing.kt
uzilan: Good suggestions. I'll try that.

ilya-g:
Thanks, I've merged it manually with some corrections

PR #3779, opened 2020-10-02
KT-41246 False positive "Receiver parameter is never used" with anonymous function expression

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41246


dimonchik0036 (approved):
Thanks!

PR #3781, opened 2020-10-02
FIR2IR: correct dispatch receiver inside inner class constructor

jsjeon:
The motivation is `secondaryConstructors/innerClasses`:
```kt
class Outer {
  inner class A1() {
    var prop: ...
    constructor(x: String): this() {
      prop = x + ...
    }
  }
}
```
The (secondary) constructor of `Outer.A1` needs `Outer` instance as a dispatch receiver. But, when (implicit) `this` receiver inside that constructor is converted, e.g., `<set-prop>`, the same `Outer`'s `<this>` is used, which supposed to be `Outer.A1`'s `<this>`.

In general, the dispatch receiver for the top function in the conversion scope is the one we're looking for as `this` receiver, but the inner class's constructor should be handled differently.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConversionScope.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConversionScope.kt
@@ -111,6 +111,11 @@ class Fir2IrConversionScope {
     fun dispatchReceiverParameter(irClass: IrClass): IrValueParameter? {
         for (function in functionStack.asReversed()) {
             if (function.parentClassOrNull == irClass) {
+                // An inner class's constructor needs an instance of the outer class as a dispatch receiver.
+                // However, if we are converting `this` receiver inside that constructor, now we should point to the inner class instance.
+                if (function is IrConstructor && irClass.isInner) {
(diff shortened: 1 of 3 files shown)


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrConversionScope.kt
```diff
@@ -111,6 +111,11 @@ class Fir2IrConversionScope {
     fun dispatchReceiverParameter(irClass: IrClass): IrValueParameter? {
         for (function in functionStack.asReversed()) {
             if (function.parentClassOrNull == irClass) {
+                // An inner class's constructor needs an instance of the outer class as a dispatch receiver.
+                // However, if we are converting `this` receiver inside that constructor, now we should point to the inner class instance.
+                if (function is IrConstructor && irClass.isInner) {
+                    irClass.thisReceiver?.let { return it }
```
jsjeon, line 117: I realized `irClass.thisReceiver` is returned as a default-ish dispatch receiver at the end, so we may just skip here if `function is IrConstructor` in general. But, it's better to return early I think.

mglukhikh:
Merged, thank you!

PR #3782, opened 2020-10-02
[JVM_IR] Rebase init blocks stepping test that is working as intended.

madsager:


madsager:
@udalov @max-kammerer 

The JVM_IR backend consistently ensures that there is a nop instruction for the exit from each init block. It looks like JVM does not do that for the last init block which looks inconsistent to me.

In this particular example:

```
3: class Foo {
4:     init {
5:         //Breakpoint!
6:         val a = 5
7:     }
8:
9:     init {
10:        val b = 7
11:    }
12: }
```

When constructing an instance, the stepping sequence for JVM is:

3, 4, 6, 7, 9, 10, 11

For JVM_IR we step back to the line number three for the return from the synthesized constructor:

3, 4, 6, 7, 9, 10, 11, 3

I think that makes sense and I would prefer to note have special handling for the last init block.

For full details, here is the JVM bytecode:

```
  public Foo();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=2, args_size=1
         0: aload_0
         1: invokespecial #8                  // Method java/lang/Object."<init>":()V
         4: nop
         5: iconst_5
         6: istore_1
         7: nop
         8: nop
         9: bipush        7
        11: istore_1
        12: return
      LineNumberTable:
        line 3: 0
        line 4: 4
        line 6: 5
        line 7: 7
        line 9: 8
        line 10: 9
        line 11: 12
      LocalVariableTable:
            0      13     0  this   LFoo;
```

And here is the JVM_IR bytecode:

```
  public Foo();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=2, args_size=1
         0: aload_0
         1: invokespecial #8                  // Method java/lang/Object."<init>":()V
         4: nop
         5: iconst_5
         6: istore_1
         7: nop
         8: nop
         9: bipush        7
        11: istore_1
        12: nop
        13: return
      LineNumberTable:
        line 3: 0
        line 4: 4
        line 6: 5
        line 7: 7
        line 9: 8
        line 10: 9
        line 11: 12
        line 3: 13
      LocalVariableTable:
            0      14     0  this   LFoo;
```

udalov:
Yeah, this looks like a minor difference, so I'm OK with declaring this to be working as intended. @max-kammerer What do you think?

max-kammerer:
@madsager Thank you!

PR #3783, opened 2020-10-03
Added Samples for Kotlin.text last() functions

zekehul:
Added samples for Kotlin.text `last()` and `last(predicate)` functions


ilya-g (changes requested):
Please reformat the new sample according to the Kotlin style guide (`Ctrl/Cmd+Alt+L` shortcut can do it).

qurbonzoda:
You can run gradle task `./gradlew :kotlin-stdlib:samples:check` to make sure your sample compiles and actually works as expected.

libraries/stdlib/samples/test/samples/text/char.kt
```diff
@@ -124,4 +124,20 @@ class Chars {
         assertTrue('a'.equals('A', true))
     }
 
-}
\ No newline at end of file
+    @sample
```
qurbonzoda, line 127: @**S**ample

libraries/stdlib/samples/test/samples/text/char.kt
```diff
@@ -124,4 +124,20 @@ class Chars {
...
+    fun last() {
+	    val chars = listOf('a', 'z', '1', '2')
```
qurbonzoda, line 129: We use 4 spaces indentation

qurbonzoda:
The `last` and `last(predicate)` functions are not related to `Char`.
It would be very useful to provide samples for `String.last()` and `String.last(predicate)`. For example:
```
    @Sample
    fun last() {
        val string = "Kotlin 1.4"
        assertEquals('4', string.last())
        assertEquals('n', string.last { it.isLetter() })
    }
```

libraries/stdlib/samples/test/samples/text/char.kt
zekehul: Fixed

libraries/stdlib/samples/test/samples/text/char.kt
zekehul: This should be fixed as well

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -408,4 +408,18 @@ class Strings {
         assertPrints(matchDetails(inputString, toFind, 2), "Searching for 'ever' in 'Never ever give up' starting at position 2: Found at 6")
         assertPrints(matchDetails(inputString, toFind, 10), "Searching for 'ever' in 'Never ever give up' starting at position 10: Not found")
     }
+
+    @Sample
+    fun last() {
+        val str0 = "Sample String 0"
+        val str1 = ""
+        assertEquals('0', str0.last())
+        assertEquals('g', str0.last( {it.isLetter()} ))
```
ilya-g, line 417: - Use `assertPrints(actualValue, expectedString)` instead of `assertEquals`
- The trailing lambda should be formatted as `str0.last { it.isLetter() }`

libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -408,4 +408,18 @@ class Strings {
...
+        assertFailsWith<NoSuchElementException>{
+            str1.last()
+        }
+        assertFailsWith<NoSuchElementException>{
+            str0.last( {it > 'z'} )
+        }
```
ilya-g, line 423: Please place the entire `assertFailsWith` call on the single line, otherwise it won't be correctly formatted after a conversion to a runnable sample.

ilya-g (changes requested):

libraries/stdlib/samples/test/samples/text/char.kt
```diff
@@ -123,5 +123,4 @@ class Chars {
         assertFalse('a'.equals('A', false))
         assertTrue('a'.equals('A', true))
     }
-
-}
\ No newline at end of file
+}
```
ilya-g, line 126: Better to revert unrelated changes

qurbonzoda:
The notes were fixed and changes were merged manually.
Thanks for the effort!

PR #3784, opened 2020-10-03
FIR2IR: distinguish constructor when picking return target

jsjeon:
The motivation is `secondaryConstructors/withNonLocalReturn`, which literally has a non-local return in secondary constructor, which reveals that, when retrieving IR target for the return target, constructor and normal functions should be treated differently (since they're stored in different caches). This should have been done at https://github.com/JetBrains/kotlin/pull/3720.


RoundNutz4 (approved):

mglukhikh:
Merged, thank you!

PR #3785, opened 2020-10-03
Added samples for doubles in the kotlin.math package

uzilan:
Added samples for doubles in the kotlin.math package


ilya-g (commented):
Good work!
I have two suggestions: 
- perhaps we could improve these samples even more by showing not only what these functions return provided with the given arguments, but rather by showing them in context of solving some math problem that requires to use them?
- the overloads of these functions for Float argument do not differ in usage much. I believe we can use the same sample for both overloads of a particular function provided for Double and Float arguments.

uzilan:
Cool. I'll try to think of some more examples and add the samples to the Float methods.

uzilan:
What do you think of examples like this one:
 `println("""In a triangle with opposite = 2.8, hypotenuse = 4.9 and angle θ = 35°, the sine of the angle θ is ${sin(toRadians(35.0))}""")`

ilya-g:
@uzilan Sorry for a late reply

> What do you think of examples like this one:

Yes something like this would be better, although I don't get how `opposite` and `hypotenuse` values are related to this sample? Or are they used in other expressions in the same sample function?

uzilan:
It's just to illustrate that this is a real triangle and to give the opportunity to calculate the angle in different means. I'll try to make it clearer.

fzhinkin:
There were no updates for quite a while, closing the PR as inactive. Feel free to reopen it.

PR #3786, opened 2020-10-03
KT-41230 False positive "Use Property Access Syntax" after adding function to excluded methods in inspection's settings

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41230


t-kameyama:
I've created new pull request in intellij-community, please review https://github.com/JetBrains/intellij-community/pull/1606.

PR #3789, opened 2020-10-05
FIR deserializer: load annotations on default accessors

jsjeon:
because even _default_ accessors can have annotations.

The motivation is bb test `jvmPackageName.metadataField`, which fails to map the access to property `Metadata.packageName` to the correct JVM getter signature because `@get:JvmName`, which is recorded on the default getter, is not loaded property.


mglukhikh:
Merged, thank you!

PR #3790, opened 2020-10-05
JVM_IR: make `primitive == object` slightly less lazy.

pyos:
Discarding the value used to leave an unused-but-never-destroyed temporary variable. It's best to not separate calls to `enterTemp` and `leaveTemp`.

Not sure what kind of test to add though, since this is minor -- if the result of the comparison is discarded, then the entire statement is more or less pointless.

 #KT-42251 Fixed


max-kammerer:
@pyos Could you add test from https://youtrack.jetbrains.com/issue/KT-42251 ?

pyos:
Added the simpler test from https://youtrack.jetbrains.com/issue/KT-42456

max-kammerer:
Thank you!

PR #3792, opened 2020-10-05
[IR] Save several last calculated line numbers not to recalculate them several times during code and debug info generation

LepilkinaElena:


udalov:
Maybe use something more robust such as `com.intellij.util.containers.SLRUCache`?

udalov (approved):

alexander-gorshenev (approved):

PR #3793, opened 2020-10-05
IR: add an emptiness check to all unsigned `until` loops 

pyos:
Unlike signed integers, a larger unsigned type does not mean a lower minimum value, so `x - 1` can overflow even if `x` is casted to a larger type.

 #KT-42186 Fixed


udalov:
cc @punzki

punzki:
@pyos Thank you for this, especially for adding the new tests! Note: I am working on https://youtrack.jetbrains.com/issue/KT-41352 currently, which should supersede the changes in `UntilHandler`. But I think we should accept this PR now, in case I am unable to finish my work on it in time, and the changes in the tests are also good.

libraries/stdlib/test/ranges/RangeIterationTest.kt
```diff
@@ -467,12 +467,26 @@ public class RangeIterationTest : RangeIterationTestBase() {
     }
 
     @Test fun overflowZeroToMinValue() {
+        doTest(0.toByte()..MinB step 3, 0, MinB, 3, listOf())
```
udalov, line 470: This test fails on our buildserver with
```
  java.lang.AssertionError: expected: java.lang.Byte<-128> but was: java.lang.Integer<-128>
```

It also fails on JS with
```
AssertionError: Expected <-128>, actual <-129>.
```

udalov:
The following tests are also failing:
* `IrBytecodeTextTestGenerated$ForLoop$Unsigned.testForInUntilWithMixedTypeBoundsNoBoundCheckNeededForUIntRangeIR`
* `FirBytecodeTextTestGenerated$ForLoop$Unsigned.testForInUntilWithMixedTypeBoundsNoBoundCheckNeededForUIntRangeIR`

libraries/stdlib/test/ranges/RangeIterationTest.kt
pyos: The numbers should be fixed now...to be honest, I don't know how to run this test locally.

pyos:
Yeah, that test explicitly confirms that the check is elided, which as explained is not a correct optimization. I'll just remove it, because IMO the number of branches is kind of a meaningless expectation anyway.

libraries/stdlib/test/ranges/RangeIterationTest.kt
pyos: Never mind, I just ran the entire `coreLibsTest` and it was pretty fast. Definitely fixed now.

udalov:
Thanks!

PR #3795, opened 2020-10-06
[FIR]: fix KT-42351

juan-chen:
FirVisibilityChecker::isVisible checked if a private declaration can be
accessed at a use site by matching class ids of the private
declaration's owner with the use site's containing class
declarations. When the private declaration is defined in an enum
entry and used in the same entry, its owner class id has the enum
entry name, but the use site is in an FirAnonymousObject, which has
"anonymous" as the class id. This causes visibility check to fail.

This PR adds handling of enum entries in isVisible: it checks if any
containing FirAnonymousObject of the use site contains the private
declaration.

### DIFF
--- a/compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
+++ b/compiler/fir/raw-fir/light-tree2fir/src/org/jetbrains/kotlin/fir/lightTree/converter/DeclarationsConverter.kt
@@ -643,7 +643,12 @@ class DeclarationsConverter(
                     )
                     superTypeRefs += enumClassWrapper.delegatedSuperTypeRef
                     convertPrimaryConstructor(null, null, enumClassWrapper, null)?.let { declarations += it.firConstructor }
-                    classBodyNode?.also { declarations += convertClassBody(it, enumClassWrapper) }
+                    classBodyNode?.also {
+                        // Use ANONYMOUS_OBJECT_NAME for the owner class id of enum entry declarations
+                        this@DeclarationsConverter.context.className =
+                            this@DeclarationsConverter.context.className.parent().child(ANONYMOUS_OBJECT_NAME)
+                        declarations += convertClassBody(it, enumClassWrapper)
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/FirVisibilityChecker.kt
```diff
@@ -67,6 +67,8 @@ abstract class FirVisibilityChecker : FirSessionComponent {
                     } else {
                         // Member: visible inside parent class, including all its member classes
                         canSeePrivateMemberOf(containingDeclarations, ownerId, session)
+                                // Handle enum entries, as they are represented as anonymous objects (KT-42351)
+                                || containingDeclarations.any { it is FirAnonymousObject && it.declarations.contains(declaration) }
```
demiurg906, line 71: This check looks like ad-hock hack

demiurg906:
Hello @juan-chen. This issue happens not because of `FirVisiblityChecker` don't support members of enum entries, but because of raw fir builder incorrectly builds callable ids for them.

Check that such case is working:
```kotlin
fun test() {
    val x = object : Any {
        private fun foo() {}
        
        fun bar() {
            foo()
        }
    }
}
```

I can't accept your fix, but you can dig this problem deeper

demiurg906:
We discussed this problem in a team and decided that this problems should be fixed in RawFirBuilder next way: self type which passed to declarations of enum entry should have special class id same as for regular anonymous objects

demiurg906:
@juan-chen this fix looks much cleaner, thanks.

Can you please also support it in `lightTree2fir` builder?
Also check failing FIR tests in your branch:
- `Fir2IrTextTestGenerated$Declarations.testInlineCollectionOfInlineClass`
- `TreesCompareTest.testCompareDiagnostics`

juan-chen:
Thanks Dmitriy!

> Can you please also support it in `lightTree2fir` builder?

Done.

> Also check failing FIR tests in your branch:
> 
> * `Fir2IrTextTestGenerated$Declarations.testInlineCollectionOfInlineClass`

This failure seems unrelated to the PR, fixed by beac3757 already.

> * `TreesCompareTest.testCompareDiagnostics`

Fixed

demiurg906:
Merged manually

PR #3798, opened 2020-10-06
FIR2IR: element-wise SAM conversion for vararg

jsjeon:
The motivation is bb test `sam/kt31908`:
```kt
fun box() {
  val r = java.lang.Runnable { ... }
  J().foo({ result = "O" }, r) // 1st arg of Function1 type
}
```

```java
public class J {
  public void foo(Runnable... rs) { ... }
}
```
The lambda as the 1st argument needs SAM conversion. This PR extends `fir2ir`'s SAM conversion to elements inside `vararg`.


jsjeon:
Friendly ping @mglukhikh

PR #3799, opened 2020-10-06
JVM_IR / PSI2IR / FIR2IR: align the IR for bound function references

pyos:
There are currently 4 ways they are represented:

1. normal function references have the dispatch/extension receiver set;
2. inline bound references are converted by JVM_IR into lambdas, with the receiver stored in a temporary and then captured;
3. adapted function references created by PSI2IR or FIR2IR add an extension receiver to the adapter function and then bind it in `IrFunctionReference` - it may or may not be stored in a temporary first, which is unnecessary since, just like function call arguments, function reference arguments should only be evaluated once;
4. suspend-converted (language feature `SuspendConversion`) `FunctionN` objects are wrapped by PSI2IR and FIR2IR equivalently to an adapted function reference for `callee::invoke`, but just like case 2, the object itself is stored in a temporary and captured by the adapter function.

This PR aligns all of these cases to 1: the receiver (or the FunctionN object to suspend-convert) is always set as `extensionReceiver` of the `IrFunctionReference`, and there is never a temporary for it in the same block.

#KT-42492 Fixed - note that no new tests are needed since the first 2 commits break 4.


compiler/testData/ir/irText/expressions/callableReferences/suspendConversion.fir.txt
```diff
@@ -108,8 +108,6 @@ FILE fqName:<root> fileName:/suspendConversion.kt
     BLOCK_BODY
       CALL 'public final fun useSuspend (fn: kotlin.coroutines.SuspendFunction0<kotlin.Unit>): kotlin.Unit declared in <root>' type=kotlin.Unit origin=null
         fn: BLOCK type=kotlin.coroutines.SuspendFunction0<kotlin.Unit> origin=ADAPTED_FUNCTION_REFERENCE
-          VAR IR_TEMPORARY_VARIABLE name:tmp_0 type:<root>.C [val]
-            CONSTRUCTOR_CALL 'public constructor <init> () [primary] declared in <root>.C' type=<root>.C origin=null
```
pyos, line 112: Funny thing about these FIR tests: in the FUNCTION_REFERENCE below, the constructor call used to be copied instead of reading this variable. I have no idea how this is possible since the old code clearly assigns an IrGetValueImpl to extensionReceiver.

udalov:
> note that no new tests are needed since the first 2 commits break 4.

Not sure if I understand this... Since all commits combined seem to fix KT-42492, can we add a test from that issue to prevent regressions here?

pyos:
KT-42492 occurs because it hits case 3, but the temporary variable is not created since the receiver is already a read of another variable, and so IrInlineCodegen breaks (the current version looks for an IrVariable in the same block to determine whether the reference is bound). There are already tests that do *almost* the same, but simply happen to hit a case with a temporary variable - box/callableReference/bound/adapted.kt, for example, uses `C()` as the receiver. The very first commit makes it so that a temporary is *never* generated, so whether it's `C()` or `c` is irrelevant.

mglukhikh (approved):

compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/generators/ReflectionReferencesGenerator.kt
```diff
@@ -153,35 +153,26 @@ class ReflectionReferencesGenerator(statementGenerator: StatementGenerator) : St
...
             }
             val receiver = irDispatchReceiver ?: irExtensionReceiver
-
             if (receiver == null) {
                 IrFunctionExpressionImpl(
                     startOffset, endOffset, irFunctionalType, irAdapterFun, IrStatementOrigin.ADAPTED_FUNCTION_REFERENCE
                 )
             } else {
-                val statements = SmartList<IrStatement>()
-                if (receiver.isSafeToUseWithoutCopying()) {
```
udalov, line 174: Let's remove `isSafeToUseWithoutCopying` now that it's unused

udalov (approved):

max-kammerer:
@pyos is it possible to add test for last commit?

pyos:
Done. See `onInlineArgument_ir.txt` - there should be no `OnInlineArgumentKt$box$1$1` function reference class. (BTW, the test fails on JS_IR due to an AbstractMethodError in InlineClassLowering for some reason.)

max-kammerer:
@romanart  ^^
@pyos What is 'OnInlineArgumentKt$box$lambda$1' class?

pyos:
The value of `val lambda`.

max-kammerer:
Don't know why but got next fail on TC:
```
IrJsCodegenBoxTestGenerated$Coroutines$SuspendConversion.testOnInlineArgument | 4 failures in one build
-- | --
The test was run 4 times in the build, 4 failuresFirst failure: rr/mb/pr3799#1.4.30-dev-1773  pyos (4) 21 Oct 20 13:50======= Failed test run #1 ========== java.lang.AssertionError: Looks like this test can be unmuted. Remove IGNORE_BACKEND directive. 	at org.jetbrains.kotlin.test.KotlinTestUtils.lambda$testWithCustomIgnoreDirective$6(KotlinTestUtils.java:855) 	at org.jetbrains.kotlin.test.MuteWithFileKt$testWithMuteInFile$1.invoke(muteWithFile.kt:41) 	at org.jetbrains.kotlin.test.KotlinTestUtils.runTestImpl(KotlinTestUtils.java:773) 	at org.jetbrains.kotlin.test.KotlinTestUtils.runTest0(KotlinTestUtils.java:754) 	at org.jetbrains.kotlin.js.test.ir.semantics.IrJsCodegenBoxTestGenerated$Coroutines$SuspendConversion.runTest(IrJsCodegenBoxTestGenerated.java:7359) 	at org.jetbrains.kotlin.js.test.ir.semantics.IrJsCodegenBoxTestGenerated$Coroutines$SuspendConversion.testOnInlineArgument(IrJsCodegenBoxTestGenerated.java:7378) 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) |   | First failure: | rr/mb/pr3799 | #1.4.30-dev-1773 | pyos (4) | 21 Oct 20 13:50
  | First failure: | rr/mb/pr3799 | #1.4.30-dev-1773 | pyos (4) | 21 Oct 20 13:50

```
Maybe IC problem?

pyos:
Rebasing on master again indeed did fix the test, but I don't see why...

max-kammerer:
Most likely incremental compilation problem

max-kammerer:
@pyos Many thanks!

PR #3800, opened 2020-10-06
FIR: copy constructor for typealias'ed inner/nested class

jsjeon:
This PR is a complete rework of https://github.com/JetBrains/kotlin/pull/3746 as per the [suggestion](https://github.com/JetBrains/kotlin/pull/3746#issuecomment-704364170).

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
@@ -9,6 +9,7 @@ import org.jetbrains.kotlin.fir.FirSession
 import org.jetbrains.kotlin.fir.declarations.*
 import org.jetbrains.kotlin.fir.declarations.builder.buildConstructedClassTypeParameterRef
 import org.jetbrains.kotlin.fir.declarations.builder.buildConstructor
+import org.jetbrains.kotlin.fir.declarations.builder.buildConstructorCopy
 import org.jetbrains.kotlin.fir.declarations.builder.buildValueParameter
 import org.jetbrains.kotlin.fir.resolve.*
 import org.jetbrains.kotlin.fir.resolve.substitution.ConeSubstitutor
@@ -42,7 +43,7 @@ internal fun FirScope.processConstructorsByName(
             substitutor,
             processor,
             session,
-            bodyResolveComponents.scopeSession,
+            bodyResolveComponents,
             includeInnerConstructors
         )
 
@@ -135,7 +136,7 @@ private fun processConstructors(
     substitutor: ConeSubstitutor,
     processor: (FirFunctionSymbol<*>) -> Unit,
     session: FirSession,
-    scopeSession: ScopeSession,
+    bodyResolveComponents: BodyResolveComponents,
     includeInnerConstructors: Boolean
 ) {
     try {
@@ -144,7 +145,7 @@ private fun processConstructors(
                 is FirTypeAliasSymbol -> {
                     matchedSymbol.ensureResolved(FirResolvePhase.TYPES, session)
                     val type = matchedSymbol.fir.expandedTypeRef.coneTypeUnsafe<ConeClassLikeType>().fullyExpandedType(session)
-                    val basicScope = type.scope(session, scopeSession)
+                    val basicScope = type.scope(session, bodyResolveComponents.scopeSession)
 
                     if (basicScope != null && type.typeArguments.isNotEmpty()) {
                         prepareSubstitutingScopeForTypeAliasConstructors(
@@ -154,14 +155,16 @@ private fun processConstructors(
                 }
                 is FirClassSymbol ->
                     (matchedSymbol.fir as FirClass<*>).scopeForClass(
-                        substitutor, session, scopeSession
+                        substitutor, session, bodyResolveComponents.scopeSession
                     )
             }
 
             //TODO: why don't we use declared member scope at this point?
             scope?.processDeclaredConstructors {
                 if (includeInnerConstructors || !it.fir.isInner) {
-                    processor(it)
+                    val constructorSymbolToProcess =
+                        prepareCopyConstructorForTypealiasNestedClass(matchedSymbol, it, session, bodyResolveComponents) ?: it
+                    processor(constructorSymbolToProcess)
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
```diff
@@ -161,6 +162,38 @@ private fun processConstructors(
...
+                    //
+                    // the chances are that `processor` belongs to [ScopeTowerLevel] (to resolve type aliases at top-level), which treats
+                    // the explicit receiver (`Outer()`) as an extension receiver, whereas the constructor of the nested class may regard
+                    // the same explicit receiver as a dispatch receiver (hence inconsistent receiver).
+                    // Here, we add a copy of the nested class constructor, along with the outer type as an extension receiver, so that it
+                    // can be seen as if resolving:
+                    //
+                    //   fun Outer.OI(): OI = ...
+                    //
+                    if (it.callableId.classId?.isNestedClass == true && matchedSymbol is FirTypeAliasSymbol) {
```
dzharkov, line 181: Wow! That looks even simpler than I thought

dzharkov: I'd extract it into a separate function like`prepareConstructor` that would return the original one if no replacement is needed

dzharkov: Otherwise, both (copy and original) are being processed now

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
```diff
@@ -161,6 +162,38 @@ private fun processConstructors(
...
+                        val outerClassId = it.callableId.classId?.outerClassId!!
+                        val outerClass = session.firProvider.getFirClassifierByFqName(outerClassId) as? FirClass<*>
+                        val outerType = outerClass?.getPrimaryConstructorIfAny()?.returnTypeRef
+                        if (outerType != null) {
+                            val extCopy = buildConstructorCopy(it.fir) {
+                                origin = FirDeclarationOrigin.FakeOverride
+                                receiverTypeRef =
+                                    outerType.withReplacedConeType(
+                                        outerType.coneTypeUnsafe<ConeClassLikeType>().replaceArgumentsWithStarProjections()
```
dzharkov, line 190: I think `replaceArgumentsWithStarProjections()` is not quite correct here.
I would just obtain `fullyExpendedType` and call `org.jetbrains.kotlin.fir.resolve.FirOuterClassManager#outerType` on it

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
```diff
@@ -161,6 +162,38 @@ private fun processConstructors(
...
+                    //
+                    //   fun Outer.OI(): OI = ...
+                    //
+                    if (it.callableId.classId?.isNestedClass == true && matchedSymbol is FirTypeAliasSymbol) {
+                        val outerClassId = it.callableId.classId?.outerClassId!!
+                        val outerClass = session.firProvider.getFirClassifierByFqName(outerClassId) as? FirClass<*>
+                        val outerType = outerClass?.getPrimaryConstructorIfAny()?.returnTypeRef
+                        if (outerType != null) {
+                            val extCopy = buildConstructorCopy(it.fir) {
+                                origin = FirDeclarationOrigin.FakeOverride
```
dzharkov, line 187: Probably, it's worth having a different kind of origin here

dzharkov (changes requested):

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
jsjeon: It's even simpler now!

I intentionally processed both, but after trying a copied one only (if available), it turns out that it actually made supposed-to-be-error case indeed a resolution error, nice! See line 29 at `typeAliasConstructorWrongClass`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
jsjeon: Good to learn the outer class manager! Quite handy.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
jsjeon: How about Synthetic? :)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
```diff
@@ -289,3 +292,43 @@ private fun <F : FirFunction<F>> prepareSubstitutorForTypeAliasConstructors(
...
+    //
+    // the chances are that `processor` belongs to [ScopeTowerLevel] (to resolve type aliases at top-level), which treats
+    // the explicit receiver (`Outer()`) as an extension receiver, whereas the constructor of the nested class may regard
+    // the same explicit receiver as a dispatch receiver (hence inconsistent receiver).
+    // Here, we add a copy of the nested class constructor, along with the outer type as an extension receiver, so that it
+    // can be seen as if resolving:
+    //
+    //   fun Outer.OI(): OI = ...
+    //
+    if (originalSymbol.callableId.classId?.isNestedClass == true && matchedSymbol is FirTypeAliasSymbol) {
```
dzharkov, line 318: For such utility functions, it may be easier to handle pre-requisites first in a form `if (!necessaryCondition) return null`.
E.g., this one is only expected to work with type-aliases to nested classes, so it can be rewritten like:
```
if (matchedSymbol !is FirTypeAliasSymbol) return null
val innerTypeRef = originalSymbol.fir.returnTypeRef
val innerType = innerTypeRef.coneType.fullyExpandedType(session) as? ConeClassLikeType
// `outerType()` returns null for not-inner class
val outerType = bodyResolveComponents.outerClassManager.outerType(innerType) ?: return null
/* useful code */
```
This would help to avoid code nestedness that (subjectively) doesn't make code more readable.
But it's cosmetics and quite a subjective statement, so I don't insist here

dzharkov:
I've merged changes manually.
Thanks for your contribution!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/ConstructorProcessing.kt
jsjeon: Ahhh, you're right; and I also prefer that style! No idea why I forgot that at the moment.

Not quite worth having a PR to clean up; if I have a chance to revisit around here with another issue, I may rewrite it.

jsjeon:
Thank you for the reviews and comments!

PR #3803, opened 2020-10-07
PSI2IR: do not generate when subjects multiple times

pyos:
`in x` is represented as `<subject expression> in x` in psi, so generating the entire call and then replacing the argument with a read of a temporary results in redundant regenerations of the subject.

 #KT-42054 Fixed
 #KT-42455 Fixed


dnpetrov:
LGTM. I'll run tests and push manually.

dnpetrov:
Pushed manually ( 98088f739d53d98b669f77cabb73c087d82bd4e5 ).

PR #3805, opened 2020-10-07
JVM_IR: move MoveOrCopyCompanionObjectFields down a bit

pyos:
More specifically, it should be done *just before* JvmPropertiesLowering, as walking the IR tree between them will result in visiting the moved backing fields twice (once via the companion's parent, once via the IrProperty).

Ideally, this group of 3 lowerings should be merged into 1 as they are completely inseparable, but this is slightly harder to do properly.

 #KT-42527 Fixed


udalov:
:+1:

PR #3806, opened 2020-10-07
added samples for Random functions

zekehul:
nextBits, nextBoolean, nextBytes, nextDouble, nextFloat, nextInt, and nextLong

### DIFF
--- a/libraries/stdlib/samples/test/samples/random/randoms.kt
+++ b/libraries/stdlib/samples/test/samples/random/randoms.kt
@@ -41,62 +41,62 @@ class Randoms {
 
     @Sample
     fun nextBits() {
-	// always generates a 0
-	println(Random.nextBits(0))
-	// get a random 1 bit value
-	println(Random.nextBits(1))
-	// get a random 8 bit value
-	println(Random.nextBits(8))
-	// get a random 16 bit value
-	println(Random.nextBits(16))
-	// get a random 32 bit value
-	println(Random.nextBits(32))
+		// always generates a 0
+		println(Random.nextBits(0))
+		// get a random 1 bit value
+		println(Random.nextBits(1))
+		// get a random 8 bit value
+		println(Random.nextBits(8))
+		// get a random 16 bit value
+		println(Random.nextBits(16))
+		// get a random 32 bit value
+		println(Random.nextBits(32))
     }
 
     @Sample
     fun nextBoolean() {
-	// get a random Boolean value
-	println(Random.nextBoolean())
+		// get a random Boolean value
+		println(Random.nextBoolean())
     }
 
     @Sample
     fun nextBytes() {
-	var bytes = ByteArray(4)
-	assertPrints(bytes[0], "0")
-	assertPrints(bytes[1], "0")
-	assertPrints(bytes[2], "0")
-	assertPrints(bytes[3], "0")
-	Random.nextBytes(bytes, 1, 3)
-	for(i in 1..2)println("bytes[${i}]: ${bytes[i]}")
-	Random,nextBytes(bytes)
-	for(i in 0..3)println("bytes[${i}]: ${bytes[i]}")
-	val new_bytes = Random.nextBytes(3)
-	for(i in 0..2)println("new_bytes[${i}]: ${new_bytes[i]}")
+		var bytes = ByteArray(4)
+		assertPrints(bytes[0], "0")
+		assertPrints(bytes[1], "0")
+		assertPrints(bytes[2], "0")
+		assertPrints(bytes[3], "0")
+		Random.nextBytes(bytes, 1, 3)
+		for(i in 1..2)println("bytes[${i}]: ${bytes[i]}")
+		Random,nextBytes(bytes)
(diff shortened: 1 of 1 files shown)


libraries/stdlib/samples/test/samples/random/randoms.kt
```diff
@@ -38,4 +38,65 @@ class Randoms {
         // random with another seed produce another sequence
         assertPrints(randomValues3, "[14, 48, 57, 67, 82, 7, 61, 27, 14, 59]")
     }
-}
\ No newline at end of file
+
+    @Sample
+    fun nextBits() {
+	// always generates a 0
+	println(Random.nextBits(0))
```
qurbonzoda, line 45: 4 spaces indentation

qurbonzoda (changes requested):
Please run `./gradlew :kotlin-stdlib:samples:check --i` from the project's root directory to test your samples

zekehul:
@qurbonzoda gradle samples check came back green

libraries/stdlib/samples/test/samples/random/randoms.kt
zekehul: They should be fixed now

libraries/stdlib/samples/test/samples/random/randoms.kt
```diff
@@ -38,4 +38,65 @@ class Randoms {
...
-}
\ No newline at end of file
+
+    @Sample
+    fun nextBits() {
+        // always generates a 0
+        println(Random.nextBits(0))
+        // get a random 1 bit value
+        println(Random.nextBits(1))
+        // get a random 8 bit value
```
qurbonzoda, line 48: These comments do not seem to be very informative, they don't clarify what the function does.
Consider writing something like "always generates 0 or 1" for `nextBits(1)`, and "generates a positive Int value less than 256" for `nextBits(8)`. 
`nextBits(16)` and `nextBits(32)` seem to be redundant as they do not add clarity, though `nextBits(32)` could be kept with comments alike "can generate any Int value"

libraries/stdlib/samples/test/samples/random/randoms.kt
```diff
@@ -38,4 +38,65 @@ class Randoms {
...
+        // get a random 16 bit value
+        println(Random.nextBits(16))
+        // get a random 32 bit value
+        println(Random.nextBits(32))
+    }
+
+    @Sample
+    fun nextBoolean() {
+        // get a random Boolean value
+        println(Random.nextBoolean())
```
qurbonzoda, line 59: Samples are meant to clarify function behavior and provide usage examples for documentation readers. This sample do not seem to comply the criteria.

Consider writing more or less realistic use cases. For example:
```
val isAliceFirst = Random.nextBoolean()
if (isAliceFirst) {
    println("Alice goes first.")
} else {
    println("Bob goes first.")
}
```

qurbonzoda:
Purpose of a sample is to be linked in the documentation of the corresponding declaration, so that reader could see usages of the API.
Excerpt from https://youtrack.jetbrains.com/issue/KT-20357
> See an example here:
>* how the sample looks like in code: https://github.com/JetBrains/kotlin/blob/1.2.0/libraries/stdlib/samples/test/samples/collections/collections.kt#L38-L49
>* how it is referenced in the function's kdoc: https://github.com/JetBrains/kotlin/blob/1.2.0/libraries/stdlib/src/generated/_Collections.kt#L1143
>* how it looks like in the resulting documentation on site: http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/group-by.html 

That is, sample for `Random.nextBits(bitCount: Int): Int` should be referenced from [the function's documentation]( https://github.com/JetBrains/kotlin/blob/95edcea9a97464c6ae2b7f5a911e4ad17d28802e/libraries/stdlib/src/kotlin/random/Random.kt#L29).

[Here](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/samples/ReadMe.md) is the guide to write sound samples.

libraries/stdlib/samples/test/samples/random/randoms.kt
zekehul: The comment for `Random.nextBits(0)` is just there because I thought it was an interesting bit of information.  I'll update the other comments to be more informative

libraries/stdlib/samples/test/samples/random/randoms.kt
```diff
@@ -38,4 +38,65 @@ class Randoms {
...
+    }
+
+    @Sample
+    fun nextBytes() {
+        var bytes = ByteArray(4)
+        assertPrints(bytes[0], "0")
+        assertPrints(bytes[1], "0")
+        assertPrints(bytes[2], "0")
+        assertPrints(bytes[3], "0")
```
ilya-g, line 68: Please use `bytes.contentToString()` instead of printing each byte separately.

qurbonzoda:
The notes were fixed and changes were merged manually.
Thanks for the effort!

PR #3801, opened 2020-10-07
small change: DELEGATE_SPECIAL_FUNCTION_NONE_APPLICABLE: is -> are

KotlinIsland:
fixed grammar in DELEGATE_SPECIAL_FUNCTION_NONE_APPLICABLE error message

PR #3807, opened 2020-10-08
small change: UNRESOLVED_REFERENCE_WRONG_RECEIVER: is -> are

KotlinIsland:
Grammar fix.

Can you please tag this pr with `hacktoberfest-accepted`


KotlinIsland:
@ivakub can this be merged?

nikita-nazarov:
The fix is redundant.  The phrase can be transformed in two ways:

- Not any of the following candidates are applicable...
- Not one of the following candidates is applicable...

So in this case both 'is' and 'are' are acceptable.

KotlinIsland:
@nikita-nazarov That's not right, "are" is correct here and "is" is incorrect.
Here's a similar change: #3801
And here is someone else mentioning the issue: [KT-38709](https://youtrack.jetbrains.com/issue/KT-38709)

PR #3808, opened 2020-10-08
ForLoopsLowering: Use last-exclusive for-loops for optimized `until` progressions instead of decrementing "last"

punzki:
#KT-41352 Fixed

The following combinations will only be _partially_ optimized (i.e., a progression will be constructed):
- `for (A until B step C)`
- `for ((A until B).reversed())`

The latter is fully optimized (i.e., no progression constructed) in the non-IR backend but is buggy (see KT-42533), hence we are also not fully optimizing for IR, for the same reasons described in the bug.


udalov:
Looks great, thanks! I'll take another look and will merge ~tomorrow.

The critical part of KT-41352 is fixed and it's great, but the bytecode is still a bit longer in the JVM IR version, for the most part because of the fact that we're generating a condition + do-while loop in JVM IR, instead of the while loop. Do you think this is something we could improve in the future, i.e. generate the while loop as in the old backend? If yes, I'll report another issue to track this.
We're also doing manual addition instead of using `iinc`, but that might be fixed with [KT-42010](https://youtrack.jetbrains.com/issue/KT-42010).

punzki:
@udalov The reason it was if+do-while was because the old K/N version of the lowering was doing the same, and changing it to do-while caused a regression in the benchmarks. I don't think it made a difference for JVM IR. It was a priority to make sure K/N didn't regress at the time, and it should still be a priority now. I haven't ran the K/N benchmarks in a loooong time, but I'll check the numbers again.

I think you can open the issue to track this if bytecode size is important for JVM. We could add a boolean (`alwaysAddIfCondition` or something) to switch between versions for K/N vs JVM.

Also, I _think_ I should be able to use `IINC` by just adding the origin to the `IrSetVariable`. I'll try that out.

punzki:
@udalov Last commit enables use of `IINC`. I haven't been able to run the benchmarks yet. Since I'm going out-of-office for awhile, could we merge this PR as-is (if it's good) and track the work to use while instead of if+do-while (including benchmarking) in KT-36837?

udalov:
The changes look good, thanks! Enjoy your time off work! :)

PR #3809, opened 2020-10-08
[JVM_IR] Fix mangling of default argument stubs for internal methods.

madsager:
The MethodSignatureMapper expected to be able to look at the body
of the default argument stub. That is of course not possible when
it is from an external dependency.

Instead, we go through the attribute owner to get to the method
the stub is a default argument adapter for.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
@@ -105,12 +105,12 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
     private fun mangleMemberNameIfRequired(name: String, function: IrSimpleFunction): String {
         val newName = JvmCodegenUtil.sanitizeNameIfNeeded(name, context.state.languageVersionSettings)
 
-        val suffix = when {
-            function.isTopLevel ->
-                if (function.isInvisibleInMultifilePart()) function.parentAsClass.name.asString() else null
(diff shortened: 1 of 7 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/MethodSignatureMapper.kt
```diff
@@ -127,15 +127,19 @@ class MethodSignatureMapper(private val context: JvmBackendContext) {
                 (DescriptorVisibilities.isPrivate(suspendFunctionOriginal().visibility) ||
                         originalForDefaultAdapter?.isInvisibleInMultifilePart() == true)
 
-    private fun IrSimpleFunction.shouldMangleAsInternal(): Boolean =
-        (origin != JvmLoweredDeclarationOrigin.STATIC_INLINE_CLASS_CONSTRUCTOR &&
-                visibility == DescriptorVisibilities.INTERNAL &&
-                !isPublishedApi())
-                || originalForDefaultAdapter?.shouldMangleAsInternal() == true
+    private fun IrSimpleFunction.shouldMangleAsInternal(): IrSimpleFunction? {
```
udalov, line 130: Maybe rename it to something like `getBaseNameForMangling` (`...IfNeeded`?), since it no longer returns boolean?

madsager: Good idea. I went for: `getInternalFunctionForManglingIfNeeded`.

udalov:
Merged to master (1.4.30) in afd710292a8. Thanks!

PR #3810, opened 2020-10-08
Fix whitespace in Standard.kt

Xapphire13:
When looking at `Standard.kt`, I noticed there was inconsistent whitespace (a doc comment not touching the documented function for example).

This PR cleans that up


Xapphire13:
@udalov, could you give context on why you closed this PR? I read the contributing guidelines and don't believe this breaks them.

udalov:
Sorry. We tend to avoid merging such changes because they basically don't add any value, yet each takes a bit of our resources (people & CI time). Imagine if a lot of people started sending PRs that fix whitespaces and typos in comments, that would get tiresome pretty quickly. I guess we should add some note about this to our contributing guidelines, for some reason I was sure we had that explained somewhere...

PR #3811, opened 2020-10-08
JVM_IR: do not deep-copy suspend lambdas in initializers

pyos:
This loses reflection metadata (and also sometimes fails). Which was missing anyway - this is also fixed now.

 #KT-42554 Fixed


udalov:
Thanks for the quick resolution!

PR #3812, opened 2020-10-09
small change: update with fix of KT-12448

KotlinIsland:
please tag with `hacktoberfest-approved`

### DIFF
--- a/libraries/stdlib/src/kotlin/reflect/typeOf.kt
+++ b/libraries/stdlib/src/kotlin/reflect/typeOf.kt
@@ -8,7 +8,6 @@ package kotlin.reflect
 /**
  * Returns a runtime representation of the given reified type [T] as an instance of [KType].
  */
-@Suppress("unused") // KT-12448
 @SinceKotlin("1.3")
 @ExperimentalStdlibApi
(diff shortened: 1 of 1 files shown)


libraries/stdlib/src/kotlin/reflect/typeOf.kt
```diff
@@ -8,8 +8,7 @@ package kotlin.reflect
 /**
  * Returns a runtime representation of the given reified type [T] as an instance of [KType].
  */
-@Suppress("unused") // KT-12448
 @SinceKotlin("1.3")
 @ExperimentalStdlibApi
-public inline fun <reified T> typeOf(): KType =
+public inline fun <@Suppress("unused") reified T> typeOf(): KType =
```
ilya-g, line 13: Thanks, but actually this suppress is not needed at all, because the `T` type parameter is references in the docs above.

ilya-g: So it would be enough just to remove the suppress from the function.

ilya-g (approved):

ilya-g:
Merged, thanks for the improvement!

PR #3813, opened 2020-10-09
FIR checker: make property init analyzer path-sensitive

jsjeon:
The motivation is  https://youtrack.jetbrains.com/issue/KT-42350:
```kt
fun println(arg: Int) {}

fun foo() {
    val x: Int
    try {
        x = 0
    } finally {}

}
```

This is a complete rework of https://github.com/JetBrains/kotlin/pull/3773, as per the [suggestion](https://github.com/JetBrains/kotlin/pull/3773#issuecomment-701978117).

Note that this doesn't fully follow the suggestion, though. The differences are:
* Only label the path from `try-main` through `finally` to the exit target, a.k.a., an exception throwing path

Other paths, like the normal path from `try-main-exit` through `finally` to `try-exit` or paths from (non-escaping) `catch` clauses to `try-exit`, will be merged at `try-exit` anyway, so not needed to distinguish them.

* No proxy enter/exit nodes

Since we need one single label (for an exception throwing path) for now, proxy nodes are not quite needed (yet). If we start allowing multiple labels between nodes with multiple labels, proxy nodes would be handy for the CFG traversal logic. I left some places that would be affected in the future.

------

The 1st commit adds the reproduction of the issue, along with some more variants.

Before these changes, CFG for try/catch/finally block is missing edges from try-main block and catch clauses to the finally block. The 2nd commit adds them.

Towards path-sensitive CFA, the 3rd commit extends edges in CFG to _labeled_ ones. It is intentional to use `null` as the default label, which indicates the lack of path info; also, it's handy to use as a special key in the map.

The 4th commit adds path-sensitive version of CFG traversal (not sure how to generalize both of path-insensitive and path-sensitive ones, in particular, `visitor`'s type is quite different) and makes property initialization analyzer _path-sensitive_.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/CFGNode.kt
```diff
@@ -19,25 +19,34 @@ import org.jetbrains.kotlin.utils.addToStdlib.runIf
...
+            edgeExists: Boolean,
+            label: String? = null
+        ) {
+            // It's hard to define label merging, hence overwritten with the latest one.
+            // One day, if we allow multiple edges between nodes with different labels, we won't even need kind merging.
+            if (kind != EdgeKind.Forward || label != null) {
+                val fromToKind = from._outgoingEdges[to]?.kind ?: runIf(edgeExists) { EdgeKind.Forward }
                 merge(kind, fromToKind)?.let {
-                    from._outgoingEdges[to] = it
+                    from._outgoingEdges[to] = Edge(it, label)
```
demiurg906, line 45: Since almost all edges in user code without labels, there will be a lot of same objects of `Edge` with four different kinds and `null` as label which is a waste of memory. One of possible solutions is creating four static instances of `Edge` and using them if `label == null`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
```diff
@@ -804,7 +807,15 @@ class ControlFlowGraphBuilder {
         levelCounter--
         val node = createTryMainBlockExitNode(tryExpression)
         popAndAddEdge(node)
-        addEdge(node, tryExitNodes.top())
+        // NB: If this block can escape, not needed to connect to the finally block.
+        val tryMainBlockEscaped = blockEscaped(node) { n -> n is TryMainBlockEnterNode && n.level == levelCounter }
```
demiurg906, line 811: I still didn't get why we need such condition and what problems it solves

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
```diff
@@ -793,7 +796,7 @@ class ControlFlowGraphBuilder {
 
         if (tryExpression.finallyBlock != null) {
             val finallyEnterNode = createFinallyBlockEnterNode(tryExpression)
-            addEdge(enterTryNodeBlock, finallyEnterNode)
+            addEdge(enterTryNodeBlock, finallyEnterNode, label = "${FINALLY_PREFIX}_${tryExitNode.level}")
```
demiurg906, line 799: I just realize that we actually have only two different labels: 
- after finally go to function exit
- after finaly go to next line

So maybe it's ok to replace type of label with some enum or even add two new EdgeKinds? This way may help us to save some memory and don't create new entities just for one case, which simplify understating of this part of analisys

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
```diff
@@ -52,17 +53,91 @@ class LocalPropertyCollector private constructor() : ControlFlowGraphVisitorVoid
...
+            //   | l1         // path entry
+            // { l1 -> ... }  // now, every info bound to the label
+            constructor(persistentMapOf(label to (map[null] ?: PropertyInitializationInfo.EMPTY)))
+        }
+    }
+
+    fun merge(other: PathAwarePropertyInitializationInfo): PathAwarePropertyInitializationInfo {
+        var resultMap = persistentMapOf<String?, PropertyInitializationInfo>()
+        for (label in keys.union(other.keys)) {
+            // disjoint merging to preserve paths. i.e., merge the property initialization info iff both have the key.
```
demiurg906, line 106: Typo: `iff`

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
```diff
@@ -75,3 +80,60 @@ private fun <I : ControlFlowInfo<I, K, V>, K : Any, V : Any> ControlFlowGraph.co
         }
     }
 }
+
+// ---------------------- Path-sensitive data collection -----------------------
+
+fun <I : ControlFlowInfo<I, K, V>, K : Any?, V : Any> ControlFlowGraph.collectPathAwareDataForNode(
```
demiurg906, line 86: Actually all control flow analysises should be path aware, so it's good to totally replace `collectDataForNode` with `collectPathAwareDataForNode` (and rename last one to `collectDataForNode`)

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/FirPropertyInitializationAnalyzer.kt
```diff
@@ -48,7 +48,20 @@ object FirPropertyInitializationAnalyzer : AbstractFirPropertyInitializationChec
             val symbol = reference.resolvedSymbol as? FirPropertySymbol ?: return
             if (symbol !in localProperties) return
             if (symbol.fir.isLateInit) return
-            val kind = data.getValue(node)[symbol] ?: EventOccurrencesRange.ZERO
+            val info = data.getValue(node)
+            if (info.containsKey(null)) {
+                // Check accumulated info at the normal path
+                investigate(info[null]!!, symbol, node)
```
demiurg906, line 54: This API (`info[null]` as most common call) looks wierd. Can you make API of `PathAwarePropertyInitializationInfo` more friendly?

compiler/testData/diagnostics/tests/controlFlowAnalysis/assignedInTryWithCatch.kt
```diff
@@ -0,0 +1,70 @@
...
+    val a: Int
+    try {
+        s as String // Potential cast exception
+        a = 42
+    } catch (e: Exception) {
+        s as String // Potential cast exception
+        <!VAL_REASSIGNMENT!>a<!> = 41
+    } finally {
+    }
+    a.hashCode()
```
demiurg906, line 43: It will be good add here `s.length` call, so we can check behavior smartcast from finally blocks also (or create separate test for it)

demiurg906 (changes requested):

demiurg906:
@jsjeon despite all notes, this is a great job and thank you for that

demiurg906:
Also I recommend you to regularly run `FirDiagnosticsTestSpecGenerated`, it contains a lot of useful testcases. In your branch `FirDiagnosticsTestSpecGenerated.NotLinked.Contracts.Analysis.ControlFlow.Initialization.Neg#test3` is failinig

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/CFGNode.kt
jsjeon: Good idea, thanks for the suggestion. I assume four kinds you mentioned are these: https://github.com/JetBrains/kotlin/blob/master/compiler/tests-common/tests/org/jetbrains/kotlin/fir/AbstractFirDiagnosticTest.kt#L324, but I thought it won't hurt to cover all edge kinds when there is no label. Added a creating util that returns static ones appropriately.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
jsjeon: So, in your diagram in my previous PR, consider the path from `catch 2` through `finally` to the function exit. Without this condition, such path will be flown to the `finally` and needs to flow out to the function exit. This needs either handling multiple edges with the same label or labeling edges on that path with another unique label. With this condition, we can avoid adding a spurious path in the beginning, which also makes the path-sensitive version of CFG traversal as well as the addition of path-aware property init info a lot simpler, because we can focus on only normal v.s. uncaught-exception paths.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
jsjeon: `enum` sounds good to me. That also removed `K: Any?` change I added to control-flow info. Plus, APIs to check/retrieve the property init info at the normal path look much simpler too. Thanks for the suggestion!

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
jsjeon: It was intentional: https://en.wikipedia.org/wiki/If_and_only_if

if both `info`s have the same key, their corresponding property init info will be merged. If a certain key has a merged form of property init info, that means that key appears at both paths.

Anyway, I changed to explicitly say that term.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
jsjeon: Okay... good to hear that we want all CFA checkers to be path-sensitive. However, augmenting path info to existing control-flow info implies changes on how each checker interprets the collected info. It's unclear how easy or non-trivial it would be to migrate all analyses. Can I do that as a follow-up PR? I'd like to tackle one by one, or certain number/group of checkers.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/FirPropertyInitializationAnalyzer.kt
jsjeon: Right; thanks to `enum`eration of edge label, things look cleaner I guess.

compiler/testData/diagnostics/tests/controlFlowAnalysis/assignedInTryWithCatch.kt
jsjeon: It would be better to put assignment-related tests together, and separate smartcasts-related tests. I found at https://github.com/JetBrains/kotlin/pull/3776 that many existing tests do that way.

To that end, I tried to add similar smartcast tests, and found that... FIR DFA doesn't behave same as old FE, even on master. At a glance, type info at FIR DFA looks much more aggregated form, and once type _approved_, one variable's changed type is used/retrieved everywhere. See https://github.com/JetBrains/kotlin/pull/3843. I think it's a different issue. Even with these changes, smartcast info is not changed. I'll keep learning DFA and investigating those tests in that PR.

jsjeon:
> Also I recommend you to regularly run `FirDiagnosticsTestSpecGenerated`, it contains a lot of useful testcases. In your branch `FirDiagnosticsTestSpecGenerated.NotLinked.Contracts.Analysis.ControlFlow.Initialization.Neg#test3` is failing

I always run `firCompilerTest`, but was not aware of `:compiler:tests-spec:test`. Thanks for the pointer. The good news is that failure just indicates test data update (in a good way).

Btw, thanks for the CFG rendering fix: https://github.com/JetBrains/kotlin/commit/f794ced8881df28ba1d1a9d462b43f26c93362fa I noticed that as well, but had no idea where to look at; just manually added two closing brackets every time. :D

compiler/testData/diagnostics/tests/controlFlowAnalysis/assignedInTryWithCatch.kt
jsjeon: As mentioned internally, I think the way FIR DFA handles try-catch-finally needs some improvement. In particular, around here: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/FirDataFlowAnalyzer.kt#L656, different paths need _fork_ed flows; otherwise, all nodes inside try-catch-finally use the same persistent flow, resulting in incorrect type info at the end. Again, I believe this is a separate issue, and can be tracked at that PR. Though, I need this PR to be merged first, since it will depend on CFG changes in this PR

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
```diff
@@ -804,7 +804,15 @@ class ControlFlowGraphBuilder {
         levelCounter--
         val node = createTryMainBlockExitNode(tryExpression)
         popAndAddEdge(node)
-        addEdge(node, tryExitNodes.top())
+        // NB: If this block can escape, not needed to connect to the finally block.
+        val tryMainBlockEscaped = blockEscaped(node) { n -> n is TryMainBlockEnterNode && n.level == levelCounter }
```
demiurg906, line 809: I still don't agree with that this check. Why it needed?  If all exits from try main blocks are returns then execution still be passed to finally before return so there **should** be an edge from that jump to finally

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
```diff
@@ -52,17 +53,91 @@ class LocalPropertyCollector private constructor() : ControlFlowGraphVisitorVoid
     }
 }
 
+class PathAwarePropertyInitializationInfo(
+    map: PersistentMap<String?, PropertyInitializationInfo> = persistentMapOf()
+) : ControlFlowInfo<PathAwarePropertyInitializationInfo, String?, PropertyInitializationInfo>(map) {
+    companion object {
+        val INIT = PathAwarePropertyInitializationInfo(persistentMapOf(null to PropertyInitializationInfo.EMPTY))
```
demiurg906: Please rename it to `EMPTY`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/CFGNode.kt
```diff
@@ -87,18 +108,18 @@ sealed class CFGNode<out E : FirElement>(val owner: ControlFlowGraph, val level:
     val previousNodes: List<CFGNode<*>> get() = _previousNodes
     val followingNodes: List<CFGNode<*>> get() = _followingNodes
 
-    private val _incomingEdges = mutableMapOf<CFGNode<*>, EdgeKind>().withDefault { EdgeKind.Forward }
-    private val _outgoingEdges = mutableMapOf<CFGNode<*>, EdgeKind>().withDefault { EdgeKind.Forward }
+    private val _incomingEdges = mutableMapOf<CFGNode<*>, Edge>().withDefault { Edge.create(EdgeLabel.Normal, EdgeKind.Forward) }
```
demiurg906, line 111: There is already `Edge.Normal_Forward`, so why don't just use it?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraph.kt
```diff
@@ -86,6 +86,49 @@ class ControlFlowGraph(val declaration: FirDeclaration?, val name: String, val k
...
+                    }
+                }
+                else -> {
+                    Edge(label, kind)
+                }
+            }
+    }
+}
+
+enum class EdgeLabel(val label: String?) {
```
demiurg906, line 120: I just realize that `EdgeLabel` can not be enum, because of labeled returns:
```kotlin
fun test(b: Boolean) {
    run {
        try {
            if (b) return
            if (b) return@run 
        } finally {
            println("Hello")
        }
    }
}
```
Here we have three kinds of edges:
- normal exit
- return from containing function
- return from lambda `run`

And there may be multiple returns from different lambdas. What do you think about making `EdgeLabel` a sealed class with two `object` implementations (for normal exit and simple return) and one class for labeled returns, which would hold symbol of targeted lambda as key?

demiurg906 (changes requested):
Also, could you please squash e51f6903 and 141d1458 to single commit?

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
jsjeon: continued at the other comment.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraphBuilder.kt
jsjeon: Thank you for keeping pointing out this. It seems like I still didn't fully comprehend try-catch-finally. :( Now I'm convinced, and removing this revealed that one of changes in `neg/3` was actually wrong. Also, interpretation in the analyzer that only checks the normal flow was wrong. Both places are fixed, thanks to this place. Reverted escaping checks.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/CFGNode.kt
jsjeon: Aha! Thanks :)

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraph.kt
jsjeon: I agree that `enum` is quite limited, and a sealed class sounds much better!

I found that even simple return---the return from containing function---has a symbol of targeted function. So, I think `ReturnLabel` in general can be declared as a holder of that target symbol.

Instead of simple return as `object` implementation, we need uncaught exception flow as an object, and that's the whole point of this chain of commits. Thus, I added `NormalPath` and `UncaughtExceptionPath` as `object`s. `ReturnLabel` is currently not used; I prefer a manageable size of cluster that has related items. Left TODOs, and will do so afterward.

jsjeon:
> Also, could you please squash [e51f690](https://github.com/JetBrains/kotlin/commit/e51f6903a9c4834d057ce6a3ce8ba4722cbc2fb2) and [141d145](https://github.com/JetBrains/kotlin/commit/141d145859e7d71e8be3f3ca527dd20c60ef2598) to single commit?

It wasn't trivial, since the 4th commit depends on the 3rd one a lot. But, it was also my mistake to use `String?` as a label type, so squashing makes sense. Still, I'd like to separate 3rd and 4th ones because they're worth being a standalone commit. Anyway, I'm glad I managed to squash the aforementioned two commits eventually!

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/cfg/ControlFlowGraph.kt
demiurg906: Note that return without label path is equal to `UncaughtExceptionPath`, so there is no need create special labels for such return, only for labeled ones

demiurg906:
Thanks, I gonna to push master your changes keeping in mind that you will return to try/finally problem and finalize this branch.

BTW I checked tests, and notices one thing:
```kotlin
fun assignedInTryAndFinally() {
    val a: Int
    try {
        a = 42
    } finally {
        a = 41 // should be VAL_REASSIGNMENT
    }
    a.hashCode()
}
```
I don't remember, do we have such checker in FIR, just want to draw your attention to it

demiurg906:
Merged manually

jsjeon:
Thank you for the reviews so far!

Re: VAL_REASSIGNMENT, yes, that's on my radar: https://github.com/JetBrains/kotlin/pull/3776

I'm going to work on smartcasts in try/catch/finally, and then revamp VAL_REASSIGNMENT. Meanwhile, I'll try to migrate existing checkers to path-sensitive versions.

PR #3814, opened 2020-10-09
JVM_IR: add local delegated property metadata to non-synthetic classes

pyos:
Otherwise kotlin-reflect won't find it (and it won't even be serialized anyway). Note that this is separate from the `$$delegatedProperties` field.

Also, refactor PropertyReferenceLowering so that it's not quadratic in the nesting level of classes (90% of that is changing indentation levels).

 #KT-42562 Fixed

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
@@ -53,10 +53,6 @@ internal val propertyReferencePhase = makeIrFilePhase(
 )
 
 private class PropertyReferenceLowering(val context: JvmBackendContext) : IrElementTransformerVoidWithContext(), FileLoweringPass {
-    // Reflection metadata for local properties is serialized under the signature "<v#$N>" attached to the containing class.
-    // This maps properties to values of N.
-    private val localPropertyIndices = mutableMapOf<IrSymbol, Int>()
-
     // TODO: join IrLocalDelegatedPropertyReference and IrPropertyReference via the class hierarchy?
     private val IrMemberAccessExpression<*>.getter: IrSimpleFunctionSymbol?
         get() = (this as? IrPropertyReference)?.getter ?: (this as? IrLocalDelegatedPropertyReference)?.getter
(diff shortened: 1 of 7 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -201,223 +184,253 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
+            // if not possible, though, putting it right here will at least allow non-reflective uses.
+            val metadataOwner = generateSequence(this) { it.parent }.find { !it.isSynthetic } ?: this
+            metadataOwner.localPropertyIndices[property.getter.symbol] = metadataOwner.localProperties.size
+            metadataOwner.localProperties.add(property.symbol)
+        }
+    }
 
-            override fun visitLocalDelegatedPropertyReference(expression: IrLocalDelegatedPropertyReference): IrExpression =
-                cachedKProperty(expression)
+    private var currentClassData: ClassData? = null
```
udalov, line 207: It'd be great to avoid mutable state here by passing this within the visitor's `data` variable. However that would also require using `IrElementTransformer` instead of `IrElementTransformerVoidWithContext` and passing `currentScope`/`currentDeclarationParent` as well, and I'm not sure that it'll be pretty, so I think it's not needed for this PR

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -201,223 +184,253 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
+    //        constructor(boundReceiver: C) : super(boundReceiver, C::class.java, "property", "getProperty()LType;", 0)
+    //        override fun get(): T = receiver.property
+    //        override fun set(value: T) { receiver.property = value }
+    //    }
+    //
+    // and then `C()::property` -> `C$property$0(C())`.
+    //
+    private fun createSpecializedKProperty(expression: IrCallableReference<*>): IrExpression {
+        // We do not reuse classes for non-reflective property references because they would not have
+        // a valid enclosing method if the same property is referenced at many points.
```
udalov, line 298: Strictly speaking, we don't have (and JLS/JVMS don't impose) any such requirements on EnclosingMethod, so we can start reusing classes for references at some point ([KT-15690](https://youtrack.jetbrains.com/issue/KT-15690)). But I'd prefer to do this after JVM IR is stabilized.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/PropertyReferenceLowering.kt
```diff
@@ -201,223 +184,253 @@ internal class PropertyReferenceLowering(val context: JvmBackendContext) : Class
...
+
+        fun localPropertyOwner(getter: IrSymbol): IrClass? =
+            if (getter in localPropertyIndices) irClass else parent?.localPropertyOwner(getter)
 
-            override fun visitPropertyReference(expression: IrPropertyReference): IrExpression =
-                cachedKProperty(expression)
+        fun rememberLocalProperty(property: IrLocalDelegatedProperty) {
+            // Prefer to attach metadata to non-synthetic classes, because it won't be serialized otherwise;
+            // if not possible, though, putting it right here will at least allow non-reflective uses.
+            val metadataOwner = generateSequence(this) { it.parent }.find { !it.isSynthetic } ?: this
```
udalov, line 201: 👍 Hopefully it now corresponds to the old backend's behavior in `CodegenAnnotatingVisitor.getMetadataOwner`

udalov:
@pyos Thank you!

udalov:
BTW great find about the quadratic complexity of `PropertyReferenceLowering`. I found two more lowerings that suffer from this issue and fixed them in [KOTLIN-CR-4553](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4553).

PR #3815, opened 2020-10-09
[IR] Eliminated expensive calculating of fqNames

LepilkinaElena:


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
```diff
@@ -85,11 +85,21 @@ fun IrSimpleFunction.overrides(other: IrSimpleFunction): Boolean {
 private val IrConstructorCall.annotationClass
     get() = this.symbol.owner.constructedClass
 
+val IrClass.packageFqName
+    get() = parent.getPackageFragment()?.fqName
+
+tailrec fun IrDeclarationWithName.hasEqualFqName(fqName: FqName): Boolean {
```
udalov, line 91: This function doesn't seem `tailrec` to me

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
```diff
@@ -85,11 +85,21 @@ fun IrSimpleFunction.overrides(other: IrSimpleFunction): Boolean {
 private val IrConstructorCall.annotationClass
     get() = this.symbol.owner.constructedClass
 
+val IrClass.packageFqName
```
udalov, line 88: Maybe use explicit return type? Otherwise one needs to check the implementation to see if it's FqName or String

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
udalov: Also, I propose to refactor a bit:
```
fun IrDeclarationWithName.hasEqualFqName(fqName: FqName): Boolean =
    name == fqName.shortName() && when (val parent = parent) {
        is IrPackageFragment -> parent.fqName == fqName.parent()
        is IrDeclarationWithName -> parent.hasEqualFqName(fqName.parent())
        else -> false
    }
```

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/AbstractSuspendFunctionsLowering.kt
```diff
@@ -80,7 +80,10 @@ abstract class AbstractSuspendFunctionsLowering<C : CommonBackendContext>(val co
     // Suppress since it is used in native
     @Suppress("MemberVisibilityCanBePrivate")
     protected fun IrCall.isReturnIfSuspendedCall() =
-        symbol.owner.run { fqNameWhenAvailable == context.internalPackageFqn.child(Name.identifier("returnIfSuspended")) }
+        symbol.owner.run {
+            parent.getPackageFragment()?.fqName == context.internalPackageFqn &&
+                    name == Name.identifier("returnIfSuspended")
```
udalov, line 85:
```suggestion
                    name.asString() == "returnIfSuspended"
```

udalov: BTW maybe this whole function could rather be refactored to something like `return symbol.signature == context.returnIfSuspendedSignature` where the latter is created in the context as something like `IdSignature.PublicSignature(internalPackageFqn, "returnIfSuspended", null, 0)`. (But I don't know if this works.)

In general I think we should move more backend code to dealing with `IdSignature`s where `owner` is not needed, as in this case, since it could be more performant.

udalov: Now that I think about, maybe we could refactor the whole `irTypePredicates.kt` and similar utilities to use `IdSignature` as well, to only deal with strings and avoid `FqName` (whose `parent` is also not very cheap sometimes)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
LepilkinaElena: > This function doesn't seem tailrec to me

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/AbstractSuspendFunctionsLowering.kt
LepilkinaElena: > Now that I think about, maybe we could refactor the whole irTypePredicates.kt and similar utilities to use IdSignature as well, to only deal with strings and avoid FqName

Thank you for the idea. FqName is really quite expensive. Getting parent isn't very cheap, but in the huge number of cases it'd be called quite rare (I can told about this looking at profile results of K/N compiler). I'll try to use IdSignature a bit later and  try to measure if it's cheaper.

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/AbstractSuspendFunctionsLowering.kt
```diff
@@ -80,7 +80,10 @@ abstract class AbstractSuspendFunctionsLowering<C : CommonBackendContext>(val co
     // Suppress since it is used in native
     @Suppress("MemberVisibilityCanBePrivate")
     protected fun IrCall.isReturnIfSuspendedCall() =
-        symbol.owner.run { fqNameWhenAvailable == context.internalPackageFqn.child(Name.identifier("returnIfSuspended")) }
+        symbol.owner.run {
+            parent.getPackageFragment()?.fqName == context.internalPackageFqn &&
+                    name.asString() == "returnIfSuspended"
```
udalov, line 85: Looks like the simple name check can come first now, since it's cheaper

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IdSignature.kt
```diff
@@ -88,6 +88,9 @@ sealed class IdSignature {
             other is PublicSignature && packageFqName == other.packageFqName && declarationFqName == other.declarationFqName &&
                     id == other.id && mask == other.mask
 
+        fun equalsWithNames(other: Any?): Boolean =
```
udalov, line 91: Why do we need this function? Why doesn't normal `equals` work?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
```diff
@@ -85,11 +85,21 @@ fun IrSimpleFunction.overrides(other: IrSimpleFunction): Boolean {
 private val IrConstructorCall.annotationClass
     get() = this.symbol.owner.constructedClass
 
+val IrClass.packageFqName: FqName?
```
udalov, line 88: Looks like this is now unused

compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/AbstractSuspendFunctionsLowering.kt
```diff
@@ -80,9 +80,10 @@ abstract class AbstractSuspendFunctionsLowering<C : CommonBackendContext>(val co
     // Suppress since it is used in native
     @Suppress("MemberVisibilityCanBePrivate")
     protected fun IrCall.isReturnIfSuspendedCall() =
-        symbol.owner.run { fqNameWhenAvailable == context.internalPackageFqn.child(Name.identifier("returnIfSuspended")) }
+        symbol.isPublicApi && context.returnIfSuspendedSignature.equalsWithNames(symbol.signature)
```
udalov, line 83: Sorry, I didn't notice the first time that there's already `returnIfSuspended` in `Symbols`. Maybe do something like this instead then: `context.ir.symbols.returnIfSuspended.signature == symbol.signature`?

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypePredicates.kt
```diff
@@ -5,20 +5,59 @@
...
 import org.jetbrains.kotlin.ir.symbols.IrClassifierSymbol
+import org.jetbrains.kotlin.ir.util.IdSignature
 import org.jetbrains.kotlin.ir.util.fqNameWhenAvailable
+import org.jetbrains.kotlin.ir.util.hasEqualFqName
+import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.FqNameUnsafe
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.utils.newHashMapWithExpectedSize
+
+object IdSignatureValues {
```
udalov, line 21: Please add `@Suppress("ObjectPropertyName")`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypePredicates.kt
```diff
@@ -29,55 +68,79 @@ private fun classFqNameEquals(symbol: IrClassSymbol, fqName: FqNameUnsafe): Bool
...
+fun IrType.isArray(): Boolean = isNotNullClassType(IdSignatureValues.array)
+fun IrType.isNullableArray(): Boolean = isNullableClassType(IdSignatureValues.array)
+fun IrType.isCollection(): Boolean = isNotNullClassType(IdSignatureValues.collection)
+fun IrType.isNothing(): Boolean = isNotNullClassType(IdSignatureValues.nothing)
+fun IrType.isKClass(): Boolean = isNotNullClassType(IdSignatureValues.kClass)
 
-fun IrType.isPrimitiveType(): Boolean = StandardNames.FqNames.fqNameToPrimitiveType.keys.any { isNotNullClassType(it) }
-fun IrType.isNullablePrimitiveType(): Boolean = StandardNames.FqNames.fqNameToPrimitiveType.keys.any { isNullableClassType(it) }
+fun IrType.isPrimitiveType(): Boolean = idSignatureToPrimitiveType.keys.any { isNotNullClassType(it) }
+fun IrType.isNullablePrimitiveType(): Boolean = idSignatureToPrimitiveType.keys.any { isNullableClassType(it) }
```
udalov, line 102: This is optional and can be fixed separately, but maybe do something like this to avoid traversing the map?
```
classOrNull?.takeIf(IrClassSymbol::isPublicApi)?.signature in idSignatureToPrimitiveType
```
(but also handle nullability)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IdSignature.kt
LepilkinaElena: I made this because faced a problem with KClass in cases of annotations. Mask in backend usage for it equals 1, but for FIR - 0. I didn't debug deeply in FIR and just made this hack. If  you have an idea how to fix this it'd be cool.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/AdditionalIrUtils.kt
LepilkinaElena: It's needed for part in K/N, I'm now in process with these changes. I'll a bit change this method trying to use firstly data from signature if it's possible to escape extra iterations.

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IdSignature.kt
udalov: Looks like it's the same problem which you worked around by keeping the old FqName-based `isClassType`?

Just replace `isKClass` with
```
fun IrType.isKClass(): Boolean = isClassType(StandardNames.FqNames.kClass, false)
```

and then we can use normal `IdSignature.equals` (at least all JVM IR and all FIR tests pass for me)

udalov (approved):

PR #3817, opened 2020-10-10
FIR checker: attempt to reproduce KT-42352

jsjeon:
i.e., not reproducible.

`FirAbstractSuperCallChecker` is quite simple in the sense that it visits `super` reference and checks if the corresponding resolved callable member declaration is `abstract` or not: https://github.com/JetBrains/kotlin/blob/master/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/expression/FirAbstractSuperCallChecker.kt#L35

For the case mentioned in https://youtrack.jetbrains.com/issue/KT-42352, `super<AbstractKotlinUVariable>.getInitializer()` is resolved as `AbstractKotlinUVariable.getInitializer`, an `open` fake override, hence no report.

### DIFF
--- a/compiler/fir/analysis-tests/testData/resolve/diagnostics/abstractSuperCall.kt
+++ b/compiler/fir/analysis-tests/testData/resolve/diagnostics/abstractSuperCall.kt
@@ -39,3 +39,24 @@ abstract class J : A() {
         super.<!ABSTRACT_SUPER_CALL!>y<!>
     }
 }
+
+// KT-42352
+
+interface PsiVariable {
+    fun getInitializer(): Any?
+}
+
+interface UVariable : PsiVariable {
+    override fun getInitializer(): Any? {
+        return null
+    }
+}
+
+abstract class AbstractKotlinUVariable : PsiVariable, UVariable {
+}
+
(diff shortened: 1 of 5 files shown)


jsjeon:
Running the full pipeline test with `uast-kotlin` module fails with a different reason:
```
Analyzing uast-kotlin.main: 0% (0/1), ETA: Unknown, Elapsed: 6ms
[IR VALIDATION] Overrides private declaration org.jetbrains.kotlin.ir.declarations.impl.IrFunctionImpl@24452e28
FUN FAKE_OVERRIDE name:nearestCommentSibling visibility:private modality:FINAL <> ($this:org.jetbrains.uast.kotlin.internal.KotlinUElementWithComments, $receiver:com.intellij.psi.PsiElement, forward:kotlin.Boolean) returnType:com.intellij.psi.PsiComment? [fake_override]
```

mglukhikh:
Really, can't reproduce more. Probably was fixed by some commit about supertypes.

mglukhikh:
I closed https://youtrack.jetbrains.com/issue/KT-42352 as Obsolete.

PR #3818, opened 2020-10-12
JVM: Allow the JvmSynthetic annotation on file classes (KT-41884)

sfs:


udalov:
Not sure if we can add a new target to an existing annotation in a 1.4.x release. @ilya-g What do you think?

ilya-g:
Yes, I believe such change can be targeted to as early as 1.5.

udalov:
@sfs We have to postpone adding a new annotation target until 1.5. We can merge the backend changes right now though, and suppress the "inapplicable target" diagnostic in tests with the `// !DIAGNOSTICS` directive.

sfs:
@udalov I've udpated the PR to remove the changes to the standard library. I've also added an `@file:Suppress` directive to the test, since the `!DIAGNOSTIC` didn't seem to work.

udalov:
With these changes, the old backend supports `JvmSynthetic` on file facades only, but JVM IR will generate any annotated class as synthetic. It's not a critical problem because we don't plan to support 100% compatibility between backends for new features starting from 1.5. But it makes me wonder what's the expected behavior going to be when we add the new target to `JvmSynthetic`? The safest option probably would be to add a new frontend error (not in these changes, but when we'll add the new target), restricting the usage of `JvmSynthetic` to file classes, but maybe you have another opinion.

sfs:
That's a good point. Maybe we should just allow `@JvmSynthetic` annotations on all classes? This could be useful for library developers in general and it wouldn't be that much work to implement.

udalov:
Merged, thanks!

Yeah, we might support it for all classes, but then we'll need more tests to check that it works in all cases. But it'll be relevant closer to the major release along with adding the new target to `JvmSynthetic`.

PR #3844, opened 2020-10-13
[JVM_IR] Use line number 1 for multifile bridge methods.

madsager:
This is necessary for the intellij debugger filters to hide
these bridges.

Fixes KT-41961.

### DIFF
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
@@ -84,15 +84,20 @@ class SourceMapper(val sourceInfo: SourceInfo?) {
     val resultMappings: List<FileMapping>
         get() = fileMappings.values.toList()
 
+    companion object {
+        const val FAKE_FILE_NAME = "fake.kt"
+        const val FAKE_PATH = "kotlin/jvm/internal/FakeKt"
+    }
+
     init {
         sourceInfo?.let { sourceInfo ->
-            // If 'sourceFileName' is null, this class doesn't have debug information
-            // (e.g., multi-file class facade with multiple parts).
-            sourceInfo.sourceFileName?.let { sourceFileName ->
-                // Explicitly map the file to itself -- we'll probably need a lot of lines from it, so this will produce fewer ranges.
-                getOrRegisterNewSource(sourceFileName, sourceInfo.pathOrCleanFQN)
-                    .mapNewInterval(1, 1, sourceInfo.linesInFile)
-            }
+            // If 'sourceFileName' is null we are dealing with a synthesized class
+            // (e.g., multi-file class facade with multiple parts). Such classes
+            // only have synthetic debug information and we use a fake file name.
+            val sourceFileName = sourceInfo.sourceFileName ?: FAKE_FILE_NAME
+            // Explicitly map the file to itself -- we'll probably need a lot of lines from it, so this will produce fewer ranges.
+            getOrRegisterNewSource(sourceFileName, sourceInfo.pathOrCleanFQN)
+                .mapNewInterval(1, 1, sourceInfo.linesInFile)
         }
     }
 
@@ -110,7 +115,7 @@ class SourceMapper(val sourceInfo: SourceInfo?) {
     }
 
     fun mapSyntheticLineNumber(id: Int): Int {
-        return mapLineNumber(SourcePosition(id, "fake.kt", "kotlin/jvm/internal/FakeKt"), null)
+        return mapLineNumber(SourcePosition(id, FAKE_FILE_NAME, FAKE_PATH), null)
     }
 }
 
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/ExpressionCodegen.kt
@@ -213,6 +213,10 @@ class ExpressionCodegen(
     fun generate() {
         mv.visitCode()
         val startLabel = markNewLabel()
+        if (irFunction.origin == JvmLoweredDeclarationOrigin.MULTIFILE_BRIDGE) {
+            // Multifile bridges need to have line number 1 to be filtered out by the intellij debugging filters.
+            mv.visitLineNumber(1, startLabel)
+        }
         val info = BlockInfo()
         val body = irFunction.body!!
         generateNonNullAssertions()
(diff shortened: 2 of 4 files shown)


madsager:
@udalov There are a couple of failures I have to look into here. I'm not sure that adding this to the source mapper in the way that I have done here is the best thing. I'll update here when I have looked into that. Sorry for the noise.

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
```diff
@@ -66,7 +66,7 @@ class SourceMapCopier(val parent: SourceMapper, private val smap: SMAP, val call
             return mappedLineNumber
         }
 
-        val range = lastVisitedRange?.takeIf { lineNumber in it } ?: smap.findRange(lineNumber) ?: return -1
+        val range = lastVisitedRange?.takeIf { lineNumber in it } ?: smap.findRange(lineNumber) ?: return lineNumber
```
madsager, line 69: @pyos I think you looked into source maps. I need to force the use of line number 1 for a multifile facade which has no source mappings (as it doesn't exist in the source). Do you see any bad consequences of just letting the line number get through here instead of returning -1? I could map the line number using a fake file. However, that would put in a source mapping for the line number which seems unnecessary since it will not make sense anyway.

pyos: When inlining, `lineNumber` is a line number in the original while the return value should be a line number in the target. Those are different meanings, so returning `lineNumber` doesn't really make sense, as in the target it might have been mapped to something already.

Your problem should be solvable by changing this value to 1: https://github.com/JetBrains/kotlin/blob/8c42372/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt#L94

madsager: I actually tried that and that didn't work. Let me have another look at it. :)

pyos: Never mind, this is an inclusive 0-based index, so 0 should already allow line number 1 to be mapped to itself. The reason it doesn't is that the line below produces a `null` file name, so it's filtered out: https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt#L89-L91

>However, that would put in a source mapping for the line number which seems unnecessary since it will not make sense anyway.

Inserting line number 1 into the facade does not make sense either. However, note that if the SMAP only contains one entry that maps all line numbers to itself, it will not be written to the class file.

madsager: Thanks @pyos. This is where my investigations led me as well. I was using the smap.mapSyntheticLineNumber utility which is not the right thing here. It creates a fake line number 2 which mapped to 1 in a fake file. Maybe we should just use the name "MultiFileFacade" for the actual multi-file facades. That will get a source map that maps line 1 to line 1 which is really what we are after. It will lead to the facade having a source file attribute, but since the debugger filters them out based on the fake 1 line number, that should not be much of an issue. What do you think?

pyos: I think it's probably best to leave the name as null and just add `?: "<fake>"` or something in this call to `getOrRegisterNewSource`. That way there'll be no source file attribute and the source map still won't be written because it's trivial.

madsager: Thanks, I used the `fake.kt` file name when there is no filename in the source info. This leads to generation that matches the current backend.

madsager:
@udalov this is ready for review now, @pyos and I have reached a fixed-point. :)

udalov (approved):

madsager:
@max-kammerer Friendly ping. :)

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
max-kammerer: I'm a bit confused: actually neither nothing is inlined into facade, neither inlined from facade. So smap tranformation in facade could be present only for simplicity of code but actual smap is not neccessery to write into bytecode. Could we check that it's not present otherwise it should be removed from class file to be aligned with old backedn?

max-kammerer:
@madsager Sorry for delay, I've missed this pr

compiler/backend/src/org/jetbrains/kotlin/codegen/inline/SMAP.kt
madsager: That is correct, there is no source map in the output. `ClassCodegen.kt:363` always goes through the SMAP code. For multifile facades the file name doesn't exist. With the code in this PR we then generate a source mapping that uses a fake file name, maps the fake file name to itself, and all lines to themselves. Such trivial source mappings are not written to the output. The `javap -v` dump of the multifile facade is here (I truncated the constant pool to make it easier to read); there is no source mapping:

```
$ javap -v customLib/oneFunSameClassName/SameNameOneFunSameFileName.class 
Classfile /tmp/flaf/customLib/oneFunSameClassName/SameNameOneFunSameFileName.class
  Last modified Oct 29, 2020; size 571 bytes
  MD5 checksum 9f0d6ec1d20124a939e88a93621e21ab
public final class customLib.oneFunSameClassName.SameNameOneFunSameFileName
  minor version: 0
  major version: 50
  flags: ACC_PUBLIC, ACC_FINAL, ACC_SUPER
Constant pool:
   #1 = Utf8               customLib/oneFunSameClassName/SameNameOneFunSameFileName
   ...
  #27 = Utf8               RuntimeVisibleAnnotations
{
  public static final int oneFunSameFileNameFun();
    descriptor: ()I
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=1, locals=0, args_size=0
         0: invokestatic  #10                 // Method customLib/oneFunSameClassName/SameNameOneFunSameFileName__A1Kt.oneFunSameFileNameFun:()I
         3: ireturn
      LineNumberTable:
        line 1: 0

  public static final int oneFunSameFileNameFun2();
    descriptor: ()I
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=1, locals=0, args_size=0
         0: invokestatic  #15                 // Method customLib/oneFunSameClassName/SameNameOneFunSameFileName__A2Kt.oneFunSameFileNameFun2:()I
         3: ireturn
      LineNumberTable:
        line 1: 0
}
RuntimeVisibleAnnotations:
  0: #16(#17=[I#18,I#19,I#18],#20=[I#18,I#21,I#22],#23=I#19,#24=[s#7,s#12])
```

max-kammerer: Thank you for check! (I've just remember that we have test for that https://github.com/JetBrains/kotlin/commit/1772f20d488f71bd5eeee9a59ba61ad509871dad), sorry for additional work (

max-kammerer:
Thank you for your work!

udalov:
@madsager Great job fixing all the IDE stepping tests! Thanks :)

PR #3849, opened 2020-10-14
FIR CLI: initialize module and dependencies

jsjeon:
Otherwise, `getProgressionLastElement` from built-ins is not found.

Note that https://github.com/JetBrains/kotlin/pull/3228 did for test set up, and we should have added the same set up for CLI.

Running full pipeline tests with the filter `.*metadata.*` shows:
```
SUCCESSFUL MODULES
------------------

metadata.main: main-java-production, 12 files (589 lines)
metadata.jvm.main: main-java-production, 11 files (893 lines)

...

SUCCESSFUL MODULES
------------------

metadata.main: main-java-production, 12 files (589 lines)
metadata.jvm.main: main-java-production, 11 files (893 lines)
```

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt
@@ -50,6 +50,8 @@ import org.jetbrains.kotlin.codegen.KotlinCodegenFacade
 import org.jetbrains.kotlin.codegen.state.GenerationState
 import org.jetbrains.kotlin.codegen.state.GenerationStateEventCallback
 import org.jetbrains.kotlin.config.*
+import org.jetbrains.kotlin.container.get
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.diagnostics.*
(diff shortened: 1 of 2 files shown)


jsjeon:
Just in case, I compared what other paths are doing.

`FIR` is returned here: https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt#L189

For non-fir, from `repeatAnalysisIfNeeded` at https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt#L192, it goes to `TopDownAnalyzerFacadeForJVM#analyzerFilesWithJavaIntegration` (https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinToJVMBytecodeCompiler.kt#L558), which calls `createContainer` (https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/TopDownAnalyzerFacadeForJVM.kt#L95).

mglukhikh:
Thank you, it's great I told you about this problem yesterday! I will recheck and merge.

mglukhikh:
Merged manually 👍

PR #3850, opened 2020-10-14
JVM_IR: refine the condition for caching method nodes

pyos:
Don't cache non-inline `suspend fun`s if there is no `$$forInline`; do cache `invokeSuspend` if there *is* `invokeSuspend$$forInline`.


max-kammerer:
Thank you!

PR #3851, opened 2020-10-14
[JVM_IR] Copy offsets from the original function to invokeSuspend.

madsager:
Otherwise, implicit returns in void returning suspend functions
will not have the line number of the end brace.

Fixes KT-41962 and KT-40464.

### DIFF
--- a/idea/jvm-debugger/jvm-debugger-test/testData/stepping/stepOver/coroutines/sequenceNested2.ir.out
+++ b/idea/jvm-debugger/jvm-debugger-test/testData/stepping/stepOver/coroutines/sequenceNested2.ir.out
@@ -7,6 +7,7 @@ sequenceNested2.kt:10
 sequenceNested2.kt:11
 sequenceNested2.kt:23
 sequenceNested2.kt:23
+sequenceNested2.kt:23
 sequenceNested2.kt:24
 Disconnected from the target VM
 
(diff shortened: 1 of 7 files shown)


idea/jvm-debugger/jvm-debugger-test/testData/stepping/stepOver/coroutines/sequenceNested2.ir.out
```diff
@@ -0,0 +1,14 @@
+LineBreakpoint created at sequenceNested2.kt:8
+Run Java
+Connected to the target VM
+sequenceNested2.kt:8
+sequenceNested2.kt:9
+sequenceNested2.kt:10
+sequenceNested2.kt:11
+sequenceNested2.kt:23
+sequenceNested2.kt:23
+sequenceNested2.kt:23
```
madsager, line 10: The difference here is an extra step on line number 23. That is because bridges on JVM_IR consistently have line numbers. The `invoke` bridge for the `{ it.toList() }` lambda does not have line numbers on the current backend. The line number on JVM_IR leads to the extra step.

I proposed rebasing here as we have done that for other similar cases. However, we can attempt to add another special case to avoid the line number in these bridges if we think it matters.

udalov: I'm OK with this behavior, as long as we log it into a YouTrack issue (just for the sake of having all differences in behavior reported in one place)

madsager: Filed an issue here: https://youtrack.jetbrains.com/issue/KT-42696

ilmirus (approved):
Thanks!

udalov:
👍

PR #3852, opened 2020-10-14
FIR2IR: add support for callable reference to Java field

jsjeon:
As shown at https://youtrack.jetbrains.com/issue/KT-42656, callable reference to Java field is simply not supported yet.


mglukhikh:
Merged, thank you!

PR #3854, opened 2020-10-15
KT-41138 Batch replacing of "json" calls to "buildJsonObject" does not work

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-41138


t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-17921#focus=Comments-27-4997246.0-0

PR #3855, opened 2020-10-16
KT-30937 Quick fix could suggest to change fully qualified name with existing import alias

t-kameyama:
https://youtrack.jetbrains.com/issue/KT-30937


t-kameyama:
https://youtrack.jetbrains.com/issue/KTIJ-13716

saranya8june (commented):
Ok

PR #3856, opened 2020-10-16
[JVM] Fix the nop removal optimization.

madsager:
Never remove a nop if that would cause a line number to move
across a local lifetime boundary.

This fixes KT-42725.

### DIFF
--- a/compiler/testData/codegen/bytecodeText/coercionToUnitOptimization/nopInlineFuns.kt
+++ b/compiler/testData/codegen/bytecodeText/coercionToUnitOptimization/nopInlineFuns.kt
@@ -23,4 +23,4 @@ fun simpleFunVoid(f: () -> Unit): Unit {
     return f()
 }
 
-// 3 NOP
+// 0 NOP
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantNopsCleanupMethodTransformer.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/optimization/RedundantNopsCleanupMethodTransformer.kt
@@ -16,7 +16,6 @@
 
 package org.jetbrains.kotlin.codegen.optimization
 
-import org.jetbrains.kotlin.codegen.optimization.common.InsnSequence
 import org.jetbrains.kotlin.codegen.optimization.common.findNextOrNull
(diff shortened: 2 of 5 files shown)


compiler/testData/codegen/bytecodeText/coercionToUnitOptimization/nopInlineFuns.kt
```diff
@@ -23,4 +23,4 @@ fun simpleFunVoid(f: () -> Unit): Unit {
     return f()
 }
 
-// 3 NOP
```
madsager, line 26: Stepping through this code works beautifully without the nops, so I don't think any nops are needed. Do you see any reason there would have to be nops in the generated code for this?

madsager:
I should probably add a bit more of an explanation for this. The issue that there is a line number on a return instruction in an inline function in the standard library `filter` function:

```
L1:
   // Code for the inline function.
L2:
   LINENUMBER 6
   RETURN
L3:
```

The locals table says that the inlining variable `$i$f$filter` goes from label 1 to label 3.

Now, when inlining this code, we remove the `return` and replace it with a `nop`:

```
L1:
   // Code for the inline function.
L2:
   LINENUMBER 6
   NOP
L3:
  CHECKCAST 
  // Code that uses the result.
```

The locals table still says that the inlining variable goes from label 1 to label 3.

Unfortunately, the nop removal algorithm does not work correctly for this case and it ends up removing the `nop`. That puts the line number on the `checkcast` instruction. However, that is now outside of the range of the inlining variable which leads the debugger to break there. Because it is outside of the range of the inlining variable, the debugger does not filter out the break as being from the standard library. So using step-over, we actually end up stepping into the `filter` library inline function.

compiler/testData/codegen/bytecodeText/coercionToUnitOptimization/nopInlineFuns.kt
dnpetrov: I think that's fine. @yanex could you please take a look?

dnpetrov:
Overall, looks good to me. I'll run tests and push manually.

dnpetrov:
Pushed manually.

318d11dcecdac5c2bc137e5a1b3abd5faba7f9ce
0505bd958a835a3be45a6224d734b1f2689efcdd

PR #3857, opened 2020-10-19
FIR2IR: add implicit NOT_NULL cast if needed

jsjeon:
The motivation is bb test `nullCheckOptimization/javaNullCheckThrowsNpe.kt`:
```kt
import org.jetbrains.annotations.NotNull;
public class A {
    @NotNull
    public static String foo() { return null; }
}
// FILE: test.kt
fun box(): String {
    try {
        val s: String = A.foo()
        return "Fail: NPE should have been thrown"
    } catch (e: Throwable) {
        if (e::class != NullPointerException::class) return "Fail: exception class should be NPE: ${e::class}"
        return "OK"
    }
}
```
The initializer of `s`, `A.foo()`, is supposed to be `@NotNull`, but returns `null`. The returned value should be checked via an assertion. Due to the lack of that assertion insertion, currently `"Fail: NPE ..."` is returned. During the conversion, we need to wrap the expression (in this case, `A.foo()`) with an implicit `NOT_NULL` cast.

To do so, first, we need to know that the return type, `String`, of `A.foo()` actually has `@NotNull` annotation. Special annotations, including that `@NotNull` annotation, are not loaded/converted/serialized intentionally (IIUC). Instead, we can record it as an attribute, and use it to determine whether or not the converted expression needs `NOT_NULL` cast.


compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
```diff
@@ -95,13 +96,23 @@ fun List<FirAnnotationCall>.computeTypeAttributes(): ConeAttributes {
...
+            CompilerConeAttributes.EnhancedNullability.ANNOTATION_CLASS_ID -> attributes += CompilerConeAttributes.EnhancedNullability
             CompilerConeAttributes.ExtensionFunctionType.ANNOTATION_CLASS_ID -> attributes += CompilerConeAttributes.ExtensionFunctionType
             CompilerConeAttributes.UnsafeVariance.ANNOTATION_CLASS_ID -> attributes += CompilerConeAttributes.UnsafeVariance
+            in NOT_NULL_ANNOTATION_IDS -> attributes += CompilerConeAttributes.EnhancedNullability
         }
     }
     return ConeAttributes.create(attributes)
 }
 
+// Not to load [org/jetbrains/kotlin/load/java/JvmAnnotationNames.JETBRAINS_NOT_NULL_ANNOTATION] at :compiler.common.jvm
```
jsjeon, line 108: I guess we prefer less dependencies? Not sure how to deal with other annotations in `NOT_NULL_ANNOTATIONS`, though.

compiler/testData/codegen/box/javaInterop/notNullAssertions/enhancedNullability/inLambdaReturnWithExpectedType.kt
```diff
@@ -1,4 +1,5 @@
 // !LANGUAGE: +StrictJavaNullabilityAssertions
+// IGNORE_BACKEND_FIR: JVM_IR
```
jsjeon, line 2: I'll take a look at this regression. This fails at line 36, which seems correct to me in the sense that a function that said `@NotNull` actually returns `null`. But I guess this is related to this change: https://github.com/JetBrains/kotlin/commit/ee020ef2903ca25154973ff7d9f93a1213068bf2

That is, if the expected type has `@EnhancedNullability` (in this case, I guess so), according to that change, we shouldn't add implicit cast, and that's why it worked so far (since we've never added implicit `NOT_NULL` cast). However, since the expected type is involved in, I wonder it would be way deeper in the resolution, and instead of `fir2ir`, we may need implicit cast during the resolution?

compiler/testData/codegen/bytecodeText/nullCheckOptimization/deterministicNotNullChecks.kt
```diff
@@ -1,4 +1,7 @@
 // FILE: test/CallableDescriptor.java
+// IGNORE_BACKEND_FIR: JVM_IR
+// Here FIR adds implicit NOT_NULL cast for `origin`, resulting in an assertion being added,
+// which is the correct (yet mismatching) behavior, according to https://youtrack.jetbrains.com/issue/KT-35656
```
jsjeon, line 4: According to the comment below about `JVM_IR` behavior, I guess this is also related to expected type checking. One thing I'm confused is, as mentioned in TODO below, "there should be an assertion," but it will be a breaking change?

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
mglukhikh: This looks quite alien here. I think we should introduce some JVM-specific `computeTypeAttributes` and call it from enhancement. Basic `fir:tree` should know nothing about these annotations.

mglukhikh (changes requested):
Full pipeline test produces a lot of IOOBE here. Will send the log in DM.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
jsjeon: Ah, didn't think of that, and `fir:java` has a dependency on `compiler.common.jvm` ([link](https://github.com/JetBrains/kotlin/blob/master/compiler/fir/java/build.gradle.kts#L8)) of course. 🤦 

Added a JVM-specific version with an additional processor notion.

jsjeon:
> Full pipeline test produces a lot of IOOBE here. Will send the log in DM.

Thanks for the sharing. As mentioned in the channel, I wonder test data generation for full pipeline is not up-to-date. Still, I saw some of modules in the log are reproducing the index out-of-bound error. Every case points to `null` source of `FirElement`s eventually, and the 2nd commit addressed that.

Once the script for test data generation is updated, I can retry (or you can let me know it goes well or there are still some other modules left). I guess, as long as I can reproduce locally, it won't be a big issue. Actually, setting such source, which will be converted to offsets, will resolve/unmute debugging tests in the future. :)

jsjeon:
> > Full pipeline test produces a lot of IOOBE here. Will send the log in DM.
> 
> Thanks for the sharing. As mentioned in the channel, I wonder test data generation for full pipeline is not up-to-date. Still, I saw some of modules in the log are reproducing the index out-of-bound error. Every case points to `null` source of `FirElement`s eventually, and the 2nd commit addressed that.
> 
> Once the script for test data generation is updated, I can retry

I can regenerate the test data with the up-to-date script, and I can confirm no more IOOBE is there. Alas, two modules are failing with assertion errors. I'm working on it now.

jsjeon:
> I can regenerate the test data with the up-to-date script, and I can confirm no more IOOBE is there. Alas, two modules are failing with assertion errors. I'm working on it now.

Those are also resolved. I left some comments regarding my findings on the 2nd commit.

mglukhikh:
Ok, I will recheck everything I can.

mglukhikh:
Looks like full pipeline test is Ok now after this PR.

mglukhikh (approved):

mglukhikh:
Merged manually, thank you. NB: looks like `EnhancedNullability` is still added not everywhere it should be. For example, see this newly added test https://github.com/JetBrains/kotlin/commit/46cc01602e9afc70bca0d94e9780e34f0f344120#diff-9ffee154eb6e498b57980a68abd191d9363089743b18ed6f55a0a2134ad9dd81R2 and its IR text analogue.

jsjeon:
Thank you for the adjustment and pointers! I'll take a look at those.

PR #3860, opened 2020-10-19
[IR] Made signature nullable

LepilkinaElena:


compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/irTypePredicates.kt
```diff
@@ -57,8 +57,7 @@ fun getPublicSignature(packageFqName: FqName, name: String) =
 private fun IrType.isClassType(signature: IdSignature.PublicSignature, hasQuestionMark: Boolean? = null): Boolean {
     if (this !is IrSimpleType) return false
     if (hasQuestionMark != null && this.hasQuestionMark != hasQuestionMark) return false
-    if (!classifier.isPublicApi || !classifier.signature.isPublic) return false
-    return signature == classifier.signature
+    return classifier.signature?.let { signature == it } ?: false
```
udalov, line 60:
```suggestion
    return classifier.signature == signature
```

compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/IrModuleDeserializer.kt
```diff
@@ -37,9 +37,9 @@ abstract class IrModuleDeserializer(val moduleDescriptor: ModuleDescriptor) {
     abstract fun deserializeIrSymbol(idSig: IdSignature, symbolKind: BinarySymbolData.SymbolKind): IrSymbol
 
     open fun declareIrSymbol(symbol: IrSymbol) {
-        assert(symbol.isPublicApi) { "Symbol is not public API: ${symbol.descriptor}" }
+        assert(symbol.signature != null && symbol.isPublicApi) { "Symbol is not public API: ${symbol.descriptor}" }
```
udalov, line 40: Here and below, please extract `symbol.signature` to a variable. Also, replace `assert`s  with `require` here, to make use of smart casts to avoid extra `!!`

udalov (approved):

udalov:
I'd go further and make `isPublicApi` an extension which is implemented as `signature != null`, but it's not required for these changes

udalov:
There are also usages of `isPublicApi` in `FqNameEqualityChecker` which are not needed anymore

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/types/IrTypeSystemContext.kt
```diff
@@ -347,7 +347,7 @@ interface IrTypeSystemContext : TypeSystemContext, TypeSystemCommonSuperTypesCon
     override fun TypeConstructorMarker.getPrimitiveType(): PrimitiveType? {
         if (this !is IrClassSymbol || !isPublicApi) return null
 
-        val signature = signature.asPublic()
+        val signature = signature?.asPublic()
```
udalov, line 350: The call to `isPublicApi` above is no longer needed here and in `getPrimitiveArrayType`

LepilkinaElena:
> There are also usages of isPublicApi in FqNameEqualityChecker which are not needed anymore

But what would be in case when both are null?

udalov:
You're right, we need to handle that case, but it's still going to be a bit simpler:
```
    override fun areEqual(left: IrClassifierSymbol, right: IrClassifierSymbol): Boolean =
        left === right ||
                left.signature != null && left.signature == right.signature

    override fun getHashCode(symbol: IrClassifierSymbol): Int =
        symbol.signature?.hashCode() ?: symbol.hashCode()
```

udalov (approved):

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
```diff
@@ -76,8 +76,8 @@ class DeclarationStubGenerator(
         if (symbol is IrFieldSymbol && (symbol.descriptor as? WrappedPropertyDescriptor)?.isBound() == true) {
             return generateStubBySymbol(symbol, symbol.descriptor)
         }
-        val descriptor = if (symbol.descriptor is WrappedDeclarationDescriptor<*>)
-            findDescriptorBySignature(symbol.signature)
+        val descriptor = if (symbol.descriptor is WrappedDeclarationDescriptor<*> && symbol.signature != null)
```
romanart, line 79: Why an additional condition is required here?

compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/IrModuleDeserializer.kt
```diff
@@ -166,8 +167,9 @@ class IrModuleDeserializerWithBuiltIns(
     }
 
     override fun declareIrSymbol(symbol: IrSymbol) {
-        if (symbol.isPublicApi && checkIsFunctionInterface(symbol.signature))
-            resolveFunctionalInterface(symbol.signature, symbol.kind())
+        val signature = symbol.signature
+        if (signature != null && checkIsFunctionInterface(symbol.signature!!))
```
romanart, line 171: Could you please fix the usage of `symbol.signature!!` with `signature`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
```diff
@@ -76,8 +76,8 @@ class DeclarationStubGenerator(
...
+            findDescriptorBySignature(symbol.signature!!)
```
romanart, line 80: It would good if you put `symbol.signature` into a separate variable and use it without null-assertion

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
LepilkinaElena: Does WrappedDeclarationDescriptor guarantee that symbol has signature?

LepilkinaElena: I understood that before it was supposed to be so. But I don't see guarantees in code. May be made this condition as require or assert?

romanart: No there is no such guarantee in general. It could be symbol for private declaration.
I think it worth to add something like 
` symbol.signature ?: error("expected signature... bla bla bla")`

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
```diff
@@ -76,8 +76,10 @@ class DeclarationStubGenerator(
         if (symbol is IrFieldSymbol && (symbol.descriptor as? WrappedPropertyDescriptor)?.isBound() == true) {
             return generateStubBySymbol(symbol, symbol.descriptor)
         }
-        val descriptor = if (symbol.descriptor is WrappedDeclarationDescriptor<*> && symbol.signature != null)
-            findDescriptorBySignature(symbol.signature!!)
+        val signature = symbol.signature
+        signature ?: error("Symbol is not public API. Expected signature for symbol: ${symbol.descriptor}")
```
udalov, line 80:
```suggestion
            ?: error("Symbol is not public API. Expected signature for symbol: ${symbol.descriptor}")
```

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DeclarationStubGenerator.kt
```diff
@@ -76,8 +76,10 @@ class DeclarationStubGenerator(
         if (symbol is IrFieldSymbol && (symbol.descriptor as? WrappedPropertyDescriptor)?.isBound() == true) {
             return generateStubBySymbol(symbol, symbol.descriptor)
         }
+        val signature = symbol.signature
+            ?: error("Symbol is not public API. Expected signature for symbol: ${symbol.descriptor}")
```
romanart, line 80: I mean 
`findDescriptorBySignature(symbol.signature
            ?: error("Symbol is not public API. Expected signature for symbol: ${symbol.descriptor}")`

romanart (approved):
LGTM

PR #3861, opened 2020-10-20
kotlin.native.cocoapods.targets can be a custom list of architectures

MikeKulasinski-visa:
As sdk developers we would like to have possibility to specify a concrete list of architectures to build as fat framework. When clients of ours would like to use the framework they need to be able to run it on both iPhone and Simulator.

Therefore in this PR I am proposing a change that allows to dynamically specify a list of architectures with comma separated manner. Rest of the functionality stays untouched, meaning user still can provide just one target if they wish.

### DIFF
--- a/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt
+++ b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt
@@ -1259,7 +1259,7 @@ class CocoaPodsIT : BaseGradleIT() {
                                 set -ev
                                 REPO_ROOT="${'$'}PODS_TARGET_SRCROOT"
                                 "${'$'}REPO_ROOT/../gradlew" -p "${'$'}REPO_ROOT" :kotlin-library:syncFramework \
-                                    -Pkotlin.native.cocoapods.targets=${'$'}KOTLIN_TARGET \
+                                    -Pkotlin.native.cocoapods.target=${'$'}KOTLIN_TARGET \
                                     -Pkotlin.native.cocoapods.configuration=${'$'}CONFIGURATION \
                                     -Pkotlin.native.cocoapods.cflags="${'$'}OTHER_CFLAGS" \
                                     -Pkotlin.native.cocoapods.paths.headers="${'$'}HEADER_SEARCH_PATHS" \
@@ -1301,7 +1301,7 @@ class CocoaPodsIT : BaseGradleIT() {
                                 set -ev
                                 REPO_ROOT="${'$'}PODS_TARGET_SRCROOT"
                                 "${'$'}REPO_ROOT/../gradlew" -p "${'$'}REPO_ROOT" :second-library:syncFramework \
-                                    -Pkotlin.native.cocoapods.targets=${'$'}KOTLIN_TARGET \
+                                    -Pkotlin.native.cocoapods.target=${'$'}KOTLIN_TARGET \
                                     -Pkotlin.native.cocoapods.configuration=${'$'}CONFIGURATION \
                                     -Pkotlin.native.cocoapods.cflags="${'$'}OTHER_CFLAGS" \
                                     -Pkotlin.native.cocoapods.paths.headers="${'$'}HEADER_SEARCH_PATHS" \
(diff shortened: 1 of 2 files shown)


crisredfi:
This is a key feature for us. For our company releases we require to pack everything into frameworks and we need to specify 3 architectures to include in the framework (32, 64 and X86_64). This PR would solve some headaches we have with custom scripts to achieve our needs.

MikeKulasinski-visa:
@abelkov @ilmat192 this relatively simple problem is blocking us from releasing our product, do you think it is possible to increase priority of it a bit? Thanks a lot

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt
```diff
@@ -1251,7 +1251,7 @@ class CocoaPodsIT : BaseGradleIT() {
                                 set -ev
                                 REPO_ROOT="${'$'}PODS_TARGET_SRCROOT"
                                 "${'$'}REPO_ROOT/../gradlew" -p "${'$'}REPO_ROOT" :kotlin-library:syncFramework \
-                                    -Pkotlin.native.cocoapods.target=${'$'}KOTLIN_TARGET \
+                                    -Pkotlin.native.cocoapods.targets=${'$'}KOTLIN_TARGET \
```
ilmat192, line 1254: Renaming this property makes the plugin incompatible with Xcode projects configured by older versions of the plugin. Could you support both property names or at least don't rename it?

ilmat192 (changes requested):

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt
MikeKulasinski-visa: I decided to revert it back to singular form

ilmat192:
LGTM but could you please also add a test to [`CocoaPodsIT`](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt) for this feature? I think building a `DEBUG` framework for iosX64+iosArm64 will be enough to test it. You can use [`testCinteropExtraOpts`](https://github.com/JetBrains/kotlin/blob/d4d93b709138b682efb7a5b1c5cd13965c4ec881/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt#L723) or [`testCustomPackageName`](https://github.com/JetBrains/kotlin/blob/d4d93b709138b682efb7a5b1c5cd13965c4ec881/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/native/CocoaPodsIT.kt#L703) as a reference.

Feel free to ask me about the test infrastructure if you have any questions.

MikeKulasinski-visa:
@ilmat192 I have spent quite some time already for analyzing the tests. I have managed to run them and I think I understand what is happening there. So it seems that you are creating a temporary project on which you execute some tasks and compare results. 

I have not found a single test that tests current behaviour of targets, like a test confirming that `iphoneos` option is producing `IOS_ARM32,` IOS_ARM64` architectures that are packed in one fat framework. Therefore I can't really tell how I should write this test. 

So I need a little bit more help since I am not quite sure how to do it.

MikeKulasinski-visa:
@ilmat192 perhaps we can merge without test? Since original TARGETS were not tested anyway? I would normally took time to write the test, however in this situation it requires quite a lot of time to get on board with your testing setup.

ilmat192:
@MikeKulasinski-visa Sorry for the delay: I was sick and had no chance to answer.

You're right, the test machinery in the CocoaPods tests is a little bit complicated. So I've prepared a simple test for this change, you can include it to your PR: https://github.com/JetBrains/kotlin/commit/193b4f2643d9f05866a04f6ff2682070e8b0c305.

As for lack of tests for `arm32`+`arm64` fat frameworks, we really don't test this case but this is a flaw that should be eventually fixed.

MikeKulasinski-visa:
@ilmat192 thanks for helping out. I had to rebase on top of master since there were conflicts with tests. Someone added a new test :) I hope it is ok if we have a different commit hash than the one you had?

ilmat192 (approved):
Ok, I'll wait until the CI run and merge this PR once it finishes.

ilmat192:
One more remark since you're going to distribute K/N modules as frameworks. Please note that including several K/N frameworks _with_ _dependent_ _APIs_ into a single app may not work as you expect. The point is that Kotlin/Native follows the "closed world" compilation model. It means that all dependencies of a K/N module are packed into a single final framework.

Consider two Kotlin modules `A` and `B` both depending on a third module `lib`. Let `lib` contains only one class `Foo`. In this model, frameworks produced from `A` and `B` will contain their own copies of `Foo` (`A.Foo` and `B.Foo` in terms of Swift symbols). Thus if `Foo` is used in APIs of these modules they will not be able to interact with each other because each of them expects its own `Foo`.

Please take this into account when working on your SDK. Some additional info can be found in 
https://github.com/JetBrains/kotlin-native/issues/2423.

Note also that this restriction affects only frameworks with common dependencies other than the Kotlin stdlib. So Kotlin frameworks without common API dependencies work fine in a single application (see [1.3.70 blog post](https://blog.jetbrains.com/kotlin/2020/03/kotlin-1-3-70-released/#kotlin-native)).

MikeKulasinski-visa:
Thanks for the update :)

vkormushkin:
Please note that starting v1.5.30 syncFramework task will not support kotlin.native.cocoapods.target property. It will use kotlin.native.cocoapods.platform and  
@MikeKulasinski-visa FYI

PR #3864, opened 2020-10-21
[FIR] fix kt-42517

juan-chen:
After a cast on "this" of a local anonymous function,
FirDataflowAnalyzer replaces the type of the corresponding
FirThisReceiverExpressionImpl with the target type. Before the fix,
Fir2Ir conversion generates directly an IrGetValueImpl with the target
type for references to "this". During LocalDeclarationsLowering, the
anonymous function is rewritten, parameters copied including "this",
and the IrGetValueImpl replaced with a new one referring to the new
parameter whose type is the original type, not the target type (see
FunctionBodiesRewriter::visitGetValue).

Note that "this" of non-local functions or other value parameters
don't have similar problems.

There are two possible fixes:
1. Change Fir2Ir conversion to keep the cast, as in this PR;
2. Change LocalDeclarationsLowering to handle the case when
IrGetValueImpl contains an implicit cast. This option is more
intrusive than the first one.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
@@ -7,6 +7,7 @@ package org.jetbrains.kotlin.fir.backend
 
 import org.jetbrains.kotlin.KtNodeTypes
 import org.jetbrains.kotlin.descriptors.ClassKind
+import org.jetbrains.kotlin.descriptors.DescriptorVisibilities
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.Visibilities
 import org.jetbrains.kotlin.fir.*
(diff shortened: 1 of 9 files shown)


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
```diff
@@ -485,7 +486,8 @@ class Fir2IrVisitor(
                 }
             }
         }
-        return implicitCastOrExpression(value, castTypeRef.toIrType())
+        return implicitCastOrExpression(value, castTypeRef.toIrType(),
+                                        expressionWithSmartcast.originalExpression is FirThisReceiverExpression)
```
mglukhikh, line 490: Why don't you check here that `this` expression is from local anonymous function?

mglukhikh:
Also you forgot to regenerate other black box tests in this PR.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrVisitor.kt
juan-chen: added checks for local functions

mglukhikh:
Thanks, merged manually with some my simplifications.

PR #3865, opened 2020-10-21
JVM_IR: fix a couple of @JvmStatic-related bugs

pyos:
#KT-42816 Fixed
#KT-42846 Fixed

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
@@ -343,6 +343,7 @@ private val jvmFilePhases = listOf(
     interfaceSuperCallsPhase,
     interfaceDefaultCallsPhase,
     interfaceObjectCallsPhase,
+    jvmStaticAnnotationPhase,
 
     tailCallOptimizationPhase,
     addContinuationPhase,
@@ -359,7 +360,6 @@ private val jvmFilePhases = listOf(
     initializersPhase,
     initializersCleanupPhase,
     functionNVarargBridgePhase,
-    jvmStaticAnnotationPhase,
     staticDefaultFunctionPhase,
     bridgePhase,
     syntheticAccessorPhase,
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
@@ -143,6 +143,7 @@ internal fun IrFunction.shouldContainSuspendMarkers(): Boolean = !isInvokeSuspen
         // These are tail-call bridges and do not require any bytecode modifications.
         origin != IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER &&
         origin != JvmLoweredDeclarationOrigin.JVM_OVERLOADS_WRAPPER &&
+        origin != JvmLoweredDeclarationOrigin.JVM_STATIC_WRAPPER &&
         origin != JvmLoweredDeclarationOrigin.MULTIFILE_BRIDGE &&
         origin != JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR &&
         origin != JvmLoweredDeclarationOrigin.SYNTHETIC_ACCESSOR_FOR_HIDDEN_CONSTRUCTOR &&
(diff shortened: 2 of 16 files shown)


udalov:
> I still don't get why running the lowerings per-module is bad for
incremental compilation.

The incremental compilation-related argument is that in whole-module lowerings it's easier to introduce a bug that would make the code compile or behave differently in clean and partial compilation scenarios. For example, you can accidentally create a cache of some sort of synthetic declarations needed among several files, which could lead to all kinds of problems when recompiling the module incrementally, from accidental changes in ABI to changes in behavior.

But it's neither a big, nor the main reason we aren't looking into making all lowerings per-module anymore. The main reason is probably that to achieve the situation where everything is lowered up to a certain point, you also need to lower dependent code (lazy IR). And doing that without introducing performance regressions is a major undertaking, which I'm not sure is warranted at this point, given how well the current JVM IR handles both lowered & unlowered code in general.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
```diff
@@ -343,6 +343,7 @@ private val jvmFilePhases = listOf(
     interfaceSuperCallsPhase,
     interfaceDefaultCallsPhase,
     interfaceObjectCallsPhase,
+    jvmStaticAnnotationPhase,
```
udalov, line 346: Can we add some prerequisites on this phase, if this move is necessary?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/CoroutineCodegen.kt
```diff
@@ -143,6 +143,7 @@ internal fun IrFunction.shouldContainSuspendMarkers(): Boolean = !isInvokeSuspen
         // These are tail-call bridges and do not require any bytecode modifications.
         origin != IrDeclarationOrigin.FUNCTION_FOR_DEFAULT_PARAMETER &&
         origin != JvmLoweredDeclarationOrigin.JVM_OVERLOADS_WRAPPER &&
+        origin != JvmLoweredDeclarationOrigin.JVM_STATIC_WRAPPER &&
```
udalov, line 146: Something tells me it might be a good time to introduce a `Set` of origins here

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmLower.kt
pyos: It's not strictly necessary, this is just to ignore `$$forInline` versions of suspend functions created by AddContinuationLowering below since creating static versions of those is obviously pointless.

pyos:
In this case, the cause of the bug is that the referenced declaration is *completely* lowered, i.e. the function itself was already processed by `JvmStaticAnnotationLowering` while a reference to it wasn't. So that comes back to my previous argument that currently JVM_IR needs to be able to handle *three* states of referenced declarations (unlowered, lowered up to current point, lowered completely) while with per-module lowerings it would only need to correctly handle the first two.

pyos:
By the way, this PR does *not* fix anything about https://youtrack.jetbrains.com/issue/KT-42858 on JVM_IR. The first call can be fixed by making both versions of `MetadataSerializer.bindMethodMetadata(MetadataSource.Function, Method)` use the global serialization binding storage, just like the property version, because the functions have been moved out to the parent class. This doesn't seem like an entirely good idea to me though.

udalov:
Looks like this fails on something like this:
```
// FILE: 1.kt

fun box(): String = O.x

// FILE: 2.kt

object O {
    @JvmStatic
    val x: String
        get() = "OK"
}
```
```
java.lang.RuntimeException: Exception while generating code for:
FUN name:<get-x> visibility:public modality:FINAL <> () returnType:kotlin.String
  correspondingProperty: PROPERTY name:x visibility:public modality:FINAL [val]

	at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.generate(FunctionCodegen.kt:53)
	at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generateMethodNode(ClassCodegen.kt:321)
	at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generateMethod(ClassCodegen.kt:336)
	at org.jetbrains.kotlin.backend.jvm.codegen.ClassCodegen.generate(ClassCodegen.kt:125)
	at org.jetbrains.kotlin.backend.jvm.JvmBackendFacade.doGenerateFilesInternal$backend_jvm(JvmBackendFacade.kt:162)
	... 24 more
Caused by: java.lang.NullPointerException
	at org.jetbrains.kotlin.backend.jvm.codegen.ExpressionCodegen.generate(ExpressionCodegen.kt:217)
	at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.doGenerate(FunctionCodegen.kt:107)
	at org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.generate(FunctionCodegen.kt:51)
	... 28 more
```

udalov:
The main part of this PR is no longer needed after https://github.com/JetBrains/kotlin/commit/71179326237d7ae5b7f00a568ba186e27a5dd31a, could you please extract useful parts and rebase on master?

PR #3867, opened 2020-10-23
[JVM_IR] Rebase inline function and defaults args stepping test.

madsager:


compiler/testData/debug/stepping/simpleInlineDefaultArg.kt
```diff
@@ -10,9 +10,29 @@ fun box(): String {
...
+// codegen, the line number is reset. Therefore any code generated with line
+// number information will emit a new line number so that stepping continues
+// in the caller.
+//
+// For JVM none of the code after `alsoInline` leads to line number generation.
+// This seems problematic as the line number for the inline function is then
+// in effect for instructions outside the scope of the `$i$f$alsoInline`
+// variable life time.
+//
+// For JVM_IR the parameter non-null check generation leads to line number
```
udalov, line 23: Looks like this sentence is a bit misleading. The line number after the default value is not caused by the parameter non-null checks because if we suppress their generation by making `ifoo` private, we still generate that line number. I still think that JVM IR is correct, it's just that the comment should be slightly modified

madsager: True, the line number is also generated bacause of the zero initialization of an inlining local. Updated the comment to just state that a line number is generated for the code following `alsoInline` without specifying exactly which part of the code generated after leads to it.

udalov:
Thanks!

PR #3868, opened 2020-10-26
FIR2IR: introduce & use implicit cast inserter

jsjeon:
The motivations are [KT-42889](https://youtrack.jetbrains.com/issue/KT-42889) (which isn't resolved yet in this PR) and some regressions from https://github.com/JetBrains/kotlin/pull/3857 (which aren't resolved in this PR either). Both cases reveal how shallow `fir2ir`'s implicit cast insertion is. As discussed in the shared doc, we need a dedicated visitor for implicit cast insertion, which would be counterpart of `psi2ir`'s `InsertImplicitCasts.kt`. This PR includes a couple steps towards that direction.

------

2nd, 3rd, and 4th commits are mechinical refactoring before making that visitor more comprehensive.

5th attempted to reproduce [KT-42889](https://youtrack.jetbrains.com/issue/KT-42889), but it works well with JVM_IR and FIR. Feel free to drop it.

6th commit is a IR text test version of the same example.

1st, 7th, and 8th commits are just minor cleanup. Although 8th one is a good consolidation before 9th commit.

------

With this chain of changes, now I can confirm both outer `when` branches of [KT-42889](https://youtrack.jetbrains.com/issue/KT-42889) are visited by implicit cast inserter. However, a condition to add coercion-to-Unit is not fulfilled yet. `@FlexibleNullability` annotation on `StringBuilder?` type is suspicious, and will keep investigating it.

Regressions from https://github.com/JetBrains/kotlin/pull/3857 are related to lambda target function's expected return types, which are type parameters for all those regressions. Alas, `@EnhancedNullability` is not propagated through type constraint system yet, so changing conditions to add `NOT_NULL` at the implicit cast inserter just flip-flopped test results in the directory where similar lambdas with `@EnhancedNullability` are tested in a various way: i.e., some became red; those regressions became green; but neither behavior (the current one from the former PR or the changed one) seems correct. This also needs further investigation.

Since this includes large refactoring as well as non-trivial changes in several visitors/generators, I think it's better to make a check point here. Tested full pipeline, and no differences were found.


mglukhikh (approved):

mglukhikh:
I looked through this PR today again. I think it should be accepted, because code becomes a bit nicer, and our behavior becomes a bit closer to PSI2IR. Also, performance is not affected significantly. So, the PR does nothing bad. I will recheck it again and merge.

jsjeon:
As you may have noticed, I focused on smartcast in/after `try` expression in another PR, and didn't have a chance to look further about `@EnhancedNullability` and `@FlexibleNullablity`. So, it sounds good to me if we can merge this at the moment and investigate further on those issues. Thanks!

jsjeon:
Oh, wait, it's already merged, thanks!

PR #3869, opened 2020-10-26
[FIR] fix KT-42825

juan-chen:
The issue is the type checker doesn't consider P? a subtype of
CapturedType<in P>?, whereas P a subtype of CapturedType<in P>?. In
AbstractTypeCheckerContext::checkSubtypeForSpecialCases, it checks if
P? is a subtype of the lower type of the captured type, which is
CapturedType<in P>, and returns false.

This fix uses non-nullable version of the type to check in the case of
CapturedType.

### DIFF
--- a/core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
+++ b/core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
@@ -407,9 +407,15 @@ object AbstractTypeChecker {
         val superTypeCaptured = superType.asCapturedType()
         val lowerType = superTypeCaptured?.lowerType()
         if (superTypeCaptured != null && lowerType != null) {
+            // If superType is nullable, e.g., to check if Foo? a subtype of Captured<in Foo>?, we check the LHS, Foo?,
+            // against the nullable version of the lower type of RHS. See KT-42825
+            val nullableLowerType =
+                if (superType.isMarkedNullable())
+                    lowerType.withNullability(true)
+                else lowerType
             when (getLowerCapturedTypePolicy(subType, superTypeCaptured)) {
-                CHECK_ONLY_LOWER -> return isSubtypeOf(this, subType, lowerType)
-                CHECK_SUBTYPE_AND_LOWER -> if (isSubtypeOf(this, subType, lowerType)) return true
+                CHECK_ONLY_LOWER -> return isSubtypeOf(this, subType, nullableLowerType)
+                CHECK_SUBTYPE_AND_LOWER -> if (isSubtypeOf(this, subType, nullableLowerType)) return true
                 SKIP_LOWER -> Unit
             }
         }
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -16036,6 +16036,11 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
                 runTest("compiler/testData/codegen/box/javaInterop/generics/javaNestedSamInterface.kt");
             }
 
+            @TestMetadata("kt42825.kt")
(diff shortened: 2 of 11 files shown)


core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
```diff
@@ -407,9 +407,10 @@ object AbstractTypeChecker {
         val superTypeCaptured = superType.asCapturedType()
         val lowerType = superTypeCaptured?.lowerType()
         if (superTypeCaptured != null && lowerType != null) {
+            val properSubType = if (subType.isMarkedNullable() && superType.isMarkedNullable()) subType.withNullability(false) else subType
```
jsjeon, line 410: How about adding some comments around here to explain the case? Like,

// If both are nullable, e.g., `T? <: Captured<in T>?`, we want non-null variant of LHS, `T`, to be compared
// against lower captured type of RHS, `Captured<in T>`.

if my understanding is correct.

Also the commit message can be more descriptive with a similar explanation.

juan-chen: done

mglukhikh:
The same as your other PR today: two JS BB tests fail here.

zarechenskiy:
I was asked to check these changes because they are in the core part, so here's a bit of investigation from my side:
- the root problem for this issue is that it's possible to get **nullable**-captured type like `Captured(in Foo)?`
- it's impossible to get such types in FE 1.0 as there is no way to specify nullability for a whole projection => there is no issue for FE 1.0
- in FIR we don't approximate types for methods with a captured receiver, so for a type `Processor<Captured(in Foo)>` all member scope will have that captured type in signatures => we can get a substitution of `T?` to `Captured(in Foo)` and get a nullable-captured type (it's started somewhere here https://github.com/JetBrains/kotlin/commit/b64b32e06b2063607018bd18e3a1de3f4db2cca6)

As a result, the main question for me is what actually represents `Captured(in Foo)?` type? The approximated type is obviously `Foo?`, seems like it should be also correct if `Foo` is a type variable (but if my understanding is correct, then it should be impossible to get nullable-captured types that have type variables inside).

core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
zarechenskiy: It's not clear to me why we should modify **subType** and not superType. Basically, here we're trying to represent (approximate) type `Captured(in Foo)?` somehow, so why not add nullability for lower type? There are some implications if `Foo` is a flexible type (we can add constraint `T <: Foo!` or `T <: Foo?`)

zarechenskiy: By the way, shouldn't we modify types when `subType` is not-nullable? Consider a situation like `T <: Captured(in Foo)?` where `T` is a type variable. Currently, we'll end up adding constraint like `T <: Foo` and, as a result, we'll infer `T` to `Foo`, while it should be inferred to `Foo?`

juan-chen: Updated the PR to use nullable version of the lower type if the captured type is marked nullable. Also updated the names.

juan-chen:
Thanks for the comments! I've updated the code, comments, and the subject.

core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
zarechenskiy: Sorry for the confusion but one more addition: let's remove `lowerType.isSimpleType()` as it seems we should perform the same approximation for kinds of lower types

zarechenskiy:
Sorry again, but it's also good to include `KT-XXX` inside the description. For, example, we use a notation `#KT-42825 Fixed` or `^KT-42825 Fixed` (mostly for our YouTrack system), example: https://github.com/JetBrains/kotlin/commit/106ee1d1ffd582a4f57b82ddac41d25e521ab835

mglukhikh:
Merged manually.

PR #3870, opened 2020-10-27
[FIR] fix KT-42824

juan-chen:
The current implemention doesn't consider T a subtype of Captured<in
T>!!, since AbstractTypeCheckerContext::checkSubtypeForSpecialCases
does not handle DefinitelyNotNullType cases. This PR adds handling of
DefinitelyNotNullType.

### DIFF
--- a/core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
+++ b/core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
@@ -404,12 +404,16 @@ object AbstractTypeChecker {
 
         if (subType.isStubType() || superType.isStubType()) return isStubTypeEqualsToAnything
 
-        val superTypeCaptured = superType.asCapturedType()
+        // superType might be a definitely notNull type (see KT-42824)
+        val superOriginalType = superType.asDefinitelyNotNullType()?.original() ?: superType
+        val superTypeCaptured = superOriginalType.asCapturedType()
         val lowerType = superTypeCaptured?.lowerType()
         if (superTypeCaptured != null && lowerType != null) {
+            val lowerTypeWNullability =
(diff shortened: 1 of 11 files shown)


mglukhikh:
You should fix failing JS BB tests here (probably by ignoring them): https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/3168251. About the general approach, personally I'm not sure. Somebody who knows type checker better will look at it.

core/compiler.common/src/org/jetbrains/kotlin/types/AbstractTypeChecker.kt
```diff
@@ -404,7 +404,9 @@ object AbstractTypeChecker {
 
         if (subType.isStubType() || superType.isStubType()) return isStubTypeEqualsToAnything
 
-        val superTypeCaptured = superType.asCapturedType()
+        // superType might be a definitely notNull type (see KT-42824)
+        val superOriginalType = superType.asDefinitelyNotNullType()?.original() ?: superType
```
zarechenskiy, line 408: I believe we should propagate `!!`-nullability inside `lowerType`, similar to what you did in https://github.com/JetBrains/kotlin/pull/3869. So, subtyping check `T <: Captured(in Foo)!!` will be transformed into `T <: Foo!!`, otherwise we'll just lose information about nullability here

zarechenskiy (changes requested):
Thank you for your work! Also, please don't forget to include `#KT-42824 Fixed` inside the description of  your commit message

mglukhikh:
Merged manually. Please pay attention to the failed tests next time.

> You should fix failing JS BB tests here (probably by ignoring them): https://teamcity.jetbrains.com/buildConfiguration/Kotlin_KotlinForGoogle_CompilerWithTests/3168251. About the general approach, personally I'm not sure. Somebody who knows type checker better will look at it.

PR #3871, opened 2020-10-27
[JVM+IR] Rebase LVT test of destructuring in lambda params

kandersen:
The debug experiece of destructuring patterns in lambdas is different
across the two backends due to the IR backend moving local variables
to fields.

However, since the destructuring variable is never actually visible in
the debugger (no linenumbers in the live range of the variable), and
the variable is never used for anything other than hiding it from the
debugger, we propose that it is not actually necessary to include it
in the LVT (and in fact, could be left out of the LVT on the old
backend).


kandersen:
Maybe @yanex knows if there are necessary reasons for having the variable in the LVT (IntelliJ debug filters etc), or if it's fine to elide it.

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
```diff
@@ -17,36 +17,50 @@ suspend fun box() = foo(A()) { (x_param, _, y_param) ->
...
 // test.kt:10 foo: a:A=A, block:kotlin.jvm.functions.Function2=TestKt$box$2, $completion:kotlin.coroutines.Continuation=helpers.ResultContinuation
 // CoroutineUtil.kt:28 getContext:
+
+// LOCAL VARIABLES JVM
 // test.kt:-1 <init>:
 // test.kt:-1 create: value:java.lang.Object=A, completion:kotlin.coroutines.Continuation=helpers.ResultContinuation
 // test.kt:-1 invoke:
+
+// LOCAL VARIABLES JVM_IR
+// test.kt:-1 <init>: $completion:kotlin.coroutines.Continuation=helpers.ResultContinuation
```
madsager, line 37: Instead of rebasing this, we should probably remove (or have an empty) local variable table for synthetic methods with no line numbers?

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
```diff
@@ -17,36 +17,50 @@ suspend fun box() = foo(A()) { (x_param, _, y_param) ->
 // However, in non-IR, the fields are first stored in local variables, and they are not read directly (even for destructuring components).
 // In IR, the fields are directly read from.
 
-// METHOD : UnderscoreNamesKt$test$2.invokeSuspend(Ljava/lang/Object;)Ljava/lang/Object;
+// The local variable for destructuring lambda arguments, in this case
+// `$dstr$x_param$_u24__u24$y_param`, is moved to a field in the IR backend,
```
max-kammerer, line 21: Is it right only for suspend lambdas? Or for general one too?

max-kammerer:
Please regenerate test data

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
kandersen: Will find a definitive answer, but I believe it's in general.

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
kandersen: @max-kammerer We would like your thoughts on this, too.

kandersen:
> Please regenerate test data

Done.

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
kandersen: Nope, just suspend functions. Will adjust the note. Also, turns out there are discrepancies in plain lambdas and destructuring as well! Next PR ... :)

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
max-kammerer: Sorry for delay. It's a good question. In old backend compiler doesn't write captured variables to list of local variables.
It would be same difference in next case:
```
fun main() {
    val z = 1;
    {
        z
    }()
}
```
But with old backend debugger able to show captrured variable without this information. So I'm not sure would this difference break smth in debugger experience but keeping old approach looks safer //cc @udalov

max-kammerer: But from other side:
```
fun main() {
    val z = 1;
        object {
            //JVM: $captured_local_variable$0=1
            //JVM_IR: $z=1
            val x = z;

            fun test() = 1
        }.test()
}
```

max-kammerer: Looks like lvt in lambda constructor is redundant, but make sense in object literals and local classes

kandersen: Excellent. I think with some additional coverage we can justify leaving out certain LVTs. Are you happy with this PR as is, or should we do more work here?

yanex:
Looks nice to me.
And I totally agree with the idea. Note that we [already hide](https://github.com/JetBrains/kotlin/blob/master/idea/jvm-debugger/jvm-debugger-core/src/org/jetbrains/kotlin/idea/debugger/stackFrame/KotlinStackFrame.kt#L181) these variables in the debugger.

compiler/testData/debug/localVariables/suspend/underscoreNames.kt
max-kammerer: Yes, some tuning could be done in furher prs

max-kammerer:
Thank you!

PR #3873, opened 2020-10-28
JVM IR: SAM wrapper ABI fixes

sfs:
This PR brings the ABI for SAM wrappers on the JVM IR backend more in line with the JVM backend.
- The JVM backend generates inline SAM wrappers (i.e., `...$sam$i$...` classes) for SAM wrappers in inline lambdas as well as for wrappers in inline methods. This seems like a bug to me, since there is no reason to generate public SAM wrapper classes outside of inline functions, but nevertheless the IR backend needs to do the same thing for compatibility.
- SAM wrapper classes are non-synthetic on the JVM backend, while the implementation methods and backing fields are synthetic.
- Finally, the JVM backend does not generate generic signatures for SAM wrappers. With the introduction of raw types in the IR backend we can finally match this behavior (at the moment we are generating wrong generic signatures in some cases).

I think that we actually want the SAM wrapper class to be synthetic and we don't want to generate inline SAM wrappers in inline lambdas, but that will probably have to wait for the next major version.


sfs:
The test failed since we were generating line numbers in the (non-synthetic!) constructor of a SAM wrapper. It's correct that the constructor is non-synthetic in the JVM backend and I've added a test for this. I've also removed the line numbers (to fix `IrKotlinSteppingTestGenerated.StepIntoOnly.testSamAdapter`) and the null check (because the JVM backend doesn't generate one here).

compiler/testData/codegen/bytecodeText/sam/samWrapperSyntheticFlags.kt
```diff
@@ -0,0 +1,23 @@
+// FILE: J.java
+
+public class J {
+    public static void g(Runnable r) {
+        r.run();
+    }
+}
+
+// FILE: test.kt
```
udalov, line 9: Maybe put these two functions to separate `FILE`s, so that the file facade name in the pattern below (`TestKt` now) would clearly indicate which one is which?

udalov: On second thought, maybe this test would even work better as a `bytecodeListing` test, instead of bytecode text?

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
```diff
@@ -723,6 +727,16 @@ class JvmSymbols(
 
     val runSuspendFunction: IrSimpleFunctionSymbol =
         kotlinCoroutinesJvmInternalRunSuspendKt.functionByName("runSuspend")
+
+    val rawTypeAnnotationClass: IrClassSymbol =
```
udalov, line 731: It would be great to use the same IR class symbol instance as in psi2ir, at least to avoid code duplication. I propose to pass the `JvmGeneratorExtensions` instance to `JvmBackendContext` (instead of `classNameOverride`), and use `context.generatorExtensions.rawTypeAnnotationConstructor` in `rawType`.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmSingleAbstractMethodLowering.kt
```diff
@@ -25,12 +32,31 @@ internal val singleAbstractMethodPhase = makeIrFilePhase(
...
         if (inInlineFunctionScope) DescriptorVisibilities.PUBLIC else JavaDescriptorVisibilities.PACKAGE_VISIBILITY
 
     override fun getSuperTypeForWrapper(typeOperand: IrType): IrType =
         typeOperand.erasedUpperBound.defaultType
 
+    override fun getWrappedFunctionType(klass: IrClass): IrType =
+        klass.symbol.rawType(context as JvmBackendContext)
+
+    // The constructor of a SAM wrapper is non-synthetic and should not have line numbers.
+    // Otherwise the debugger will try to step into it.
```
udalov, line 54: I suppose we could use a test on debugger behavior, if there isn't one yet?

compiler/testData/codegen/bytecodeText/sam/samWrapperSyntheticFlags.kt
sfs: There is a bytecode listing test in `bytecodeListing/samAdapterAndInlinedOne.kt`, but it would need to have a separate expectation file for the IR backend, or alternatively fixes in the JVM backend. The JVM backend doesn't generate the correct inner class attributes and omits the nullability annotations on non-synthetic SAM constructors.

Actually I've added the expectation file for the IR backend now, since these are the only remaining differences between the backends and the IR backend is arguably more correct here.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/JvmSymbols.kt
sfs: Good point, I didn't realize that we're already passing a part of the generator extensions to the backend context. Done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmSingleAbstractMethodLowering.kt
sfs: There already is a test in `stepping/stepInto/samAdapter.kt`. That's how I noticed this problem in the first place. Previously, the test was passing since the constructor was synthetic (which it isn't in the JVM BE).

compiler/testData/codegen/bytecodeText/sam/samWrapperSyntheticFlags.kt
udalov: Thanks! Yes, the expectation file for JVM IR looks good. So now that we have it, can we add `// WITH_SIGNATURES
` to `bytecodeListing/samAdapterAndInlinedOne.kt` to also test generic signatures, and remove the new bytecode text tests?

sfs: Sure! Done.

udalov:
Thank you!

PR #3874, opened 2020-10-29
[JVM_IR] Determine correct type of empty varargs array.

madsager:
When calling vararg methods with a generic vararg type without
passing explicit parameters, we have to allocate an empty array
of the right type. We failed to do so previously, as we did
not take the type arguments for the dispatch receiver into
account.

### DIFF
--- a/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
+++ b/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
@@ -16,10 +16,7 @@ import org.jetbrains.kotlin.ir.expressions.impl.IrCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstructorCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrTypeOperatorCallImpl
 import org.jetbrains.kotlin.ir.symbols.*
-import org.jetbrains.kotlin.ir.types.IrSimpleType
-import org.jetbrains.kotlin.ir.types.IrType
-import org.jetbrains.kotlin.ir.types.isAny
-import org.jetbrains.kotlin.ir.types.isSubtypeOf
+import org.jetbrains.kotlin.ir.types.*
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.SpecialNames
 import org.jetbrains.kotlin.utils.DFS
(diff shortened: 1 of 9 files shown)


madsager:
Please hold on with reviewing this. The last round of changes have introduced a lot of failures, I'll look into it.

madsager:
Alright, fixed the failures. This should be ready for review. Sorry for uploading before it was ready. :)

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -515,11 +512,27 @@ val IrFunction.allTypeParameters: List<IrTypeParameter>
...
+    val result = mutableMapOf<IrTypeParameterSymbol, IrType>()
+    if (dispatchReceiverTypeArguments.isNotEmpty()) {
+        val parentTypeParameters = extractTypeParameters(irFunction.parentClassOrNull!!)
+        parentTypeParameters.withIndex().forEach { (index, typeParam) ->
+            dispatchReceiverTypeArguments[index].typeOrNull?.let {
+                result[typeParam.symbol] = it
+            }
+        }
+    }
+    irFunction.allTypeParameters.withIndex().associateTo(result) {
```
udalov, line 532: (`associateTo` returns destination)
```suggestion
    return typeParameters.withIndex().associateTo(result) {
```

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
```diff
@@ -515,11 +512,27 @@ val IrFunction.allTypeParameters: List<IrTypeParameter>
...
+    val dispatchReceiverTypeArguments = (dispatchReceiver?.type as? IrSimpleType)?.arguments ?: emptyList()
+    if (typeParameters.isEmpty() && dispatchReceiverTypeArguments.isEmpty()) {
+        return emptyMap()
+    }
+
+    val result = mutableMapOf<IrTypeParameterSymbol, IrType>()
+    if (dispatchReceiverTypeArguments.isNotEmpty()) {
+        val parentTypeParameters = extractTypeParameters(irFunction.parentClassOrNull!!)
+        parentTypeParameters.withIndex().forEach { (index, typeParam) ->
+            dispatchReceiverTypeArguments[index].typeOrNull?.let {
```
udalov, line 527: Can we make this `typeOrNull!!` or would it fail in some case?

madsager: It will fail for instance if there is a star projection:

Java:
```
public class AT<G> {
  public void foo(G ...x) {
  }
}
```

Kotlin:
```
fun AT<*>.bar() {
  foo()
}
```

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
madsager: Done, thanks!

compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IrUtils.kt
udalov: Right... For some reason I though this would mean a star in function's type arguments, which would be prohibited.

udalov:
Thank you!

PR #3876, opened 2020-10-29
Corrected small grammatical error in KDocs

ciarant:
No related YouTrack id. Not sure if I should have created on for something so small.

### DIFF
--- a/compiler/tests-spec/testData/diagnostics/linked/expressions/constant-literals/character-literals/p-4/neg/1.1.fir.kt
+++ b/compiler/tests-spec/testData/diagnostics/linked/expressions/constant-literals/character-literals/p-4/neg/1.1.fir.kt
@@ -5,13 +5,13 @@
 // TESTCASE NUMBER: 1
 
 fun case1() {
-    //less then four hex digits
+    //less than four hex digits
     val c0 = <!ILLEGAL_CONST_EXPRESSION!>'\u'<!>
     val c1 = <!ILLEGAL_CONST_EXPRESSION!>'\uf'<!>
     val c2 = <!ILLEGAL_CONST_EXPRESSION!>'\u1f'<!>
     val c3 = <!ILLEGAL_CONST_EXPRESSION!>'\u1wf'<!>
 
-    //more then four hex digits
+    //more than four hex digits
     val c4 = <!ILLEGAL_CONST_EXPRESSION!>'\u1wF2f'<!>
 }
 
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/ConstructorCodegen.java
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/ConstructorCodegen.java
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.descriptors.*;
 import org.jetbrains.kotlin.psi.*;
 import org.jetbrains.kotlin.resolve.BindingContext;
 import org.jetbrains.kotlin.resolve.DescriptorToSourceUtils;
+import org.jetbrains.kotlin.resolve.InlineClassesUtilsKt;
 import org.jetbrains.kotlin.resolve.calls.model.ResolvedCall;
 import org.jetbrains.kotlin.resolve.jvm.InlineClassManglingRulesKt;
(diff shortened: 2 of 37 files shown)


udalov:
These source files are generated (see the file header), so I suppose the typo should be fixed in the generator as well

ciarant:
Ahh. I saw the `generated` directory but assumed that since they were under source control I might as well go ahead an submit a PR. I'll do the same for stdlib then. Sound ok?

ilya-g:
Better to fix it in the template (https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-stdlib-gen/src/templates/Generators.kt#L888) and then run the generator (Gradle task `:tools:kotlin-stdlib-gen:run` or run configuration "Generate standard library sources" in IDEA) and commit results.
`

ciarant:
Yeah, that makes more sense. I'll do that.

compiler/tests-spec/testData/diagnostics/linked/expressions/constant-literals/character-literals/p-4/neg/1.1.fir.kt
```diff
@@ -5,13 +5,13 @@
...
 fun case1() {
-    //less then four hex digits
+    //less than four hex digits
     val c0 = <!ILLEGAL_CONST_EXPRESSION!>'\u'<!>
     val c1 = <!ILLEGAL_CONST_EXPRESSION!>'\uf'<!>
     val c2 = <!ILLEGAL_CONST_EXPRESSION!>'\u1f'<!>
     val c3 = <!ILLEGAL_CONST_EXPRESSION!>'\u1wf'<!>
 
-    //more then four hex digits
+    //more than four hex digits
```
udalov, line 14: If you're fixing this here in `1.1.fir.kt`, please also do the same change for `1.1.kt`, otherwise the test will fail :) (`x.kt` and `x.fir.kt` should have the same sources modulo diagnostics)

ilya-g:
Thanks, I've squashed the first and third commits ("less -> fewer") and merged them to master. Could you rebase your branch on top of master to continue reviewing the rest?

compiler/tests-spec/testData/diagnostics/linked/expressions/constant-literals/character-literals/p-4/neg/1.1.fir.kt
ciarant: I'm afraid I don't understand. The comment typo 'less then' -> 'less than' only occurred in `.../character-literals/p-4/neg/1.1.fir.kt`. The spelling in the associated `1.1.kt` was correct. I'm probably missing something (new to Kotlin, and this codebase!)

udalov: No, here's `1.1.kt` in your branch, it says "more then": https://github.com/JetBrains/kotlin/blob/f280c0cf8216f659cedbfa373d06e982e5521d6a/compiler/tests-spec/testData/diagnostics/linked/expressions/constant-literals/character-literals/p-4/neg/1.1.kt#L26
while in `1.1.fir.kt` (this line) you've fixed it to "than".

I'm less concerned about the typos or your fixes, and more about the fact that sources of `1.1.kt` and `1.1.fir.kt` should be the same, or the test would fail :).

ciarant: > I'm less concerned about the typos or your fixes
I would be too :-)

Turns out I'd forgotten to commit that `1.1.kt` change. Done now.

ilya-g:
Something wrong went with the rebase. It seems that you rebased Kotlin master on top of your branch. Please do the opposite.

Another way is to create a new branch from the current Kotlin master and cherry-pick only required commits from your PR branch, then force-push that new branch over the branch of this PR

ciarant:
I've made a right mess out of that! I think I'll just close this PR and do a new one containing only the required commits you've referenced. How does that sound?

ilya-g:
If you make a new branch with the only required commits, you can then update the current PR with the following command:
`git push origin your_new_branch:master -f`, given that the `origin` remote is pointing to your fork.

PR #3879, opened 2020-10-30
[JVM_IR] Follow old backend in bridge visibility and varargs marking.

madsager:
The old backend makes bridges for protected and package-private
methods public. Also, for bridges for vararg methods, the vararg
marker is not on the bridge.

These differences seem minor but are visible via reflection, so
we might as well follow the old backend.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
@@ -149,8 +149,7 @@ class FunctionCodegen(
             }
         }
 
-        val isVararg = valueParameters.lastOrNull()?.varargElementType != null
-        val isBridge = origin == IrDeclarationOrigin.BRIDGE || origin == IrDeclarationOrigin.BRIDGE_SPECIAL
+        val isVararg = valueParameters.lastOrNull()?.varargElementType != null && !isBridge()
         val modalityFlag = when ((this as? IrSimpleFunction)?.modality) {
             Modality.FINAL -> when {
                 origin == JvmLoweredDeclarationOrigin.CLASS_STATIC_INITIALIZER -> 0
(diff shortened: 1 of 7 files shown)


compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
```diff
@@ -304,6 +253,8 @@ fun IrFunction.isInlineOnly() =
 
 fun IrFunction.isReifiable() = typeParameters.any { it.isReified }
 
+private fun IrFunction.isBridge() = origin == IrDeclarationOrigin.BRIDGE || origin == IrDeclarationOrigin.BRIDGE_SPECIAL
```
udalov, line 256: This is a really minor nitpick, but maybe make this `internal` and use it in `FunctionCodegen.calculateMethodFlags`? And possibly also in `MethodSignatureMapper.mapOverriddenSpecialBuiltinIfNeeded`

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
```diff
@@ -170,7 +170,7 @@ class FunctionCodegen(
 
         return getVisibilityAccessFlag() or modalityFlag or functionDeprecationFlags or
                 (if (isStatic) Opcodes.ACC_STATIC else 0) or
-                (if (isVararg) Opcodes.ACC_VARARGS else 0) or
+                (if (isVararg && !isBridge) Opcodes.ACC_VARARGS else 0) or
```
udalov, line 173: Another minor nitpick, but I'd prefer if `&& !isBridge` was a part of the `isVararg` value declaration, just to avoid misusing `isVararg` between its declaration and this usage

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/irCodegenUtils.kt
madsager: Good idea, done.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/FunctionCodegen.kt
madsager: Done, thanks!

udalov:
Thanks!

PR #3881, opened 2020-10-30
DRAFT1: FIR2IR: avoid implicit NOT_NULL cast if expected type accepts null

jsjeon:
This is the first attempt to resolve regressions added at https://github.com/JetBrains/kotlin/pull/3857 where implicit `NOT_NULL` cast was added whenever the expression's return type has `@EnhancedNullability`. Such implicit cast is not necessary if the expected type accepts `null`.

After https://github.com/JetBrains/kotlin/pull/3868, we're ready to compare the value type and the expected type in a similar way `psi2ir` does. Then, checking if the expected type accepts `null` or not isn't a big deal. Rather, the caveat is, we should distinguish whether such expected type is inferred type or user-contributed type.

To that end, this PR attempts to utilize `FirFakeSourceElementKind.ImplicitTypeRef`. That fake kind needs to be recorded even though there is no type reference (like `var x = nonNullString()`), and then propagated during declaration resolution. It might over-populate `source`; and I need to fine-tune error reports in some checkers.

### DIFF
--- a/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirTypeMismatchOnOverrideChecker.kt
+++ b/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/declaration/FirTypeMismatchOnOverrideChecker.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.kotlin.fir.analysis.checkers.declaration
 
+import org.jetbrains.kotlin.fir.FirFakeSourceElementKind
 import org.jetbrains.kotlin.fir.FirSourceElement
 import org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext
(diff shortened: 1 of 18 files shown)


mglukhikh:
So finally I decided to choose third way which looks less hacky for me than previous two. See here: https://github.com/JetBrains/kotlin/commits/rr/mg/fir-enhanced-nullability (two commits). In general, I implemented three things here:

- 'EnhancedNullability' annotations are now preserved during translation to IR
- FIR resolve does not use this annotation for declaration types
- Translator inserts implicit not null casts only when argument with 'EnhancedNullability' is used for not-null parameter

As you can see, this way does not require any hacks with fake source or delegated type reference. I will polish code a bit tomorrow (for example, it's better to move enhanced nullability stuff to JVM part of the translator).

RoundNutz4 (approved):

jsjeon:
Thanks for taking a look at drafts, and sharing another way to do that. Yeah, that looks much simpler and cleaner. I left a couple nits in those commits.

PR #3882, opened 2020-10-30
DRAFT2: FIR2IR: avoid implicit NOT_NULL cast if expected type accepts null

jsjeon:
This is the second attempt to resolve regressions added at #3857 where implicit `NOT_NULL` cast was added whenever the expression's return type has `@EnhancedNullability`. Such implicit cast is not necessary if the expected type accepts `null`.

After #3868, we're ready to compare the value type and the expected type in a similar way `psi2ir` does. Then, checking if the expected type accepts `null` or not isn't a big deal. Rather, the caveat is, we should distinguish whether such expected type is inferred type or user-contributed type.

To that end, the first approach (#3881) attempts to utilize `FirFakeSourceElementKind.ImplicitTypeRef`. That fake kind needs to be recorded even though there is no type reference (like `var x = nonNullString()`), and then propagated during declaration resolution. It might over-populate `source`; and I need to fine-tune error reports in some checkers.

The last part of the first approach doesn't seem good, so this PR attempts to utilize `FirResolvedTypeRef.delegatedType` as a source to indicate whether the given type is inferred or not, since that property is used to store the reference to original, unresolved type ref.

During the course of implementing it, found two minor fixes; see first two commits. Core parts are also intuitive, and much simpler than the first approach.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
@@ -207,7 +207,7 @@ class Fir2IrImplicitCastInserter(
                 coerceToUnitIfNeeded(type, irBuiltIns)
             }
             // TODO: Not exactly matched with psi2ir yet...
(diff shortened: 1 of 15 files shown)


mglukhikh:
See my comment to https://github.com/JetBrains/kotlin/pull/3881

PR #3883, opened 2020-10-30
Reuse more Native implementations for Wasm

ilya-g:
When you merge it to your branch, please squash it with your commits related to stdlib, because these commits partially revert them.


skuzmich:
Hierarchy `Native = IR | Wasm` is a bit confusing because:
* Wasm backend in question does not use Native's LLVM based backend
* There is already K/N LLVM-based Wasm target, which is a different thing
* All of them use IR :)

I would mark pure Kotlin implementations as `on(PureKotlin)` or similar and keep backend hierarchy logical. Then, if we don't have a platform-specific implementation, stdlib generator would pick pure Kotlin one. Just an idea for the future, since this is a lot of changes and probably is out of scope of this Wasm update.

PR #3884, opened 2020-10-31
allow multiple retry for AnalysisHandlerExtension

neetopia:
This is a compiler extension point change needed to allow multiple round processing for [Kotlin Symbol Processing](https://goo.gle/ksp), which is using [AnalysisHandlerExtension](https://github.com/JetBrains/kotlin/blob/master/compiler/frontend.java/src/org/jetbrains/kotlin/resolve/jvm/extensions/AnalysisHandlerExtension.kt) for hooking into compiler.

There is already some level of support for retry. However, the existing retry logic will only retry once upon receiving a `RetryWithAdditionalRoots` from `analyze()`. In order to achieve multiple retries, changes to this extension behavior is needed.


erokhins:
Merged manually. Will be available since 1.4.30-M2.
Commits in 1.4.30: 4731b9fe0cc, b5b37d2b60

In master: a6cb156, 1a37706

PR #3885, opened 2020-11-02
[JVM_IR] Enable KotlinAgainstJava and remaining JavaAgainstKotlin test.

madsager:


tests/mute-common.csv
```diff
@@ -47,6 +47,8 @@ org.jetbrains.kotlin.shortenRefs.ShortenRefsTestGenerated.testExtensionForObject
 org.jetbrains.kotlin.util.KotlinVersionsTest.testVersionsAreConsistent, KT-35567,,
 org.jetbrains.kotlin.jvm.compiler.CompileKotlinAgainstJavaTestGenerated.WithAPT.testClassWithTypeParameter, KT-36448,,
```
madsager, line 48: The YouTrack issue mentioned here is marked as fixed. I added a comment on the issue.

udalov:
Looks like for whatever reason, the general test mute mechanism doesn't work for tests with specified target backend. So since you've specified target backend `JVM` for the old test, the existing tests `{With,Without}APT.testClassWithTypeParameter` started to fail. The IR versions are also failing.

Maybe replace the entry in `mute-common.csv` with `// IGNORE_BACKEND: JVM, JVM_IR` in the test data file?

madsager:
> Looks like for whatever reason, the general test mute mechanism doesn't work for tests with specified target backend. So since you've specified target backend `JVM` for the old test, the existing tests `{With,Without}APT.testClassWithTypeParameter` started to fail. The IR versions are also failing.
> 
> Maybe replace the entry in `mute-common.csv` with `// IGNORE_BACKEND: JVM, JVM_IR` in the test data file?

That was silly, sorry. Done. :)

madsager:
Hold on, I think the muting issue is hitting `JavaAgainstKotlinBinariesCheckerTestGenerated` as well now for the same reason. I'll have a look.

madsager:
Hmm, this is confusing. This change shouldn't touch those tests but it does look like the FLAKY marker in the muted file does have the intended effect.

madsager:
Or maybe it does. A flaky marker in the muted file probably means that when running the test it can fail, but it is marked as flaky on teamcity? In any case, this is not related to this change. :)

udalov:
Yeah, looks like those are unrelated. All is passing now, thanks!

PR #3886, opened 2020-11-02
JVM IR: Fix names for SAM callable references with inline class return types

sfs:
Currently the names are mangled twice. The only reason things didn't break was that bridge generation added bridges to fix these naming issues.

This fixes the ABI issue in KT-42879 for the JVM IR backend, although KT-42879 remains broken on the JVM backend.


udalov:
Thanks!

PR #3889, opened 2020-11-03
FIR checker: make calls effect analyzer path-sensitive

jsjeon:
One of TODOs left after https://github.com/JetBrains/kotlin/pull/3813 is to migrate all existing checkers to use path-sensitive CFG traversals (and then deprecate/remove old, path-insensitive traversals; and then restore names of traversing utils). This PR revises calls effect analyzer to use path-sensitive version.

The underlying info for the calls effect analyzer is event occurrence range. This PR includes some refactoring regarding CFA info whose key and value are `EdgeLabel` and `EventOccurrencesRange`.

### DIFF
--- a/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
+++ b/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfaUtils.kt
@@ -11,7 +11,6 @@ import org.jetbrains.kotlin.contracts.description.EventOccurrencesRange
 import org.jetbrains.kotlin.fir.references.FirResolvedNamedReference
 import org.jetbrains.kotlin.fir.resolve.dfa.cfg.*
 import org.jetbrains.kotlin.fir.symbols.impl.FirPropertySymbol
-import java.lang.IllegalStateException
 
 abstract class EventOccurrencesRangeInfo<E : EventOccurrencesRangeInfo<E, K>, K : Any>(
     map: PersistentMap<K, EventOccurrencesRange> = persistentMapOf()
@@ -39,6 +38,8 @@ class PropertyInitializationInfo(
     override val constructor: (PersistentMap<FirPropertySymbol, EventOccurrencesRange>) -> PropertyInitializationInfo =
         ::PropertyInitializationInfo
 
+    override val empty: () -> PropertyInitializationInfo =
+        ::EMPTY
 }
 
 class LocalPropertyCollector private constructor() : ControlFlowGraphVisitorVoid() {
@@ -61,7 +62,7 @@ class LocalPropertyCollector private constructor() : ControlFlowGraphVisitorVoid
 
 class PathAwarePropertyInitializationInfo(
     map: PersistentMap<EdgeLabel, PropertyInitializationInfo> = persistentMapOf()
-) : ControlFlowInfo<PathAwarePropertyInitializationInfo, EdgeLabel, PropertyInitializationInfo>(map) {
+) : PathAwareControlFlowInfo<PathAwarePropertyInitializationInfo, PropertyInitializationInfo>(map) {
     companion object {
         val EMPTY = PathAwarePropertyInitializationInfo(persistentMapOf(NormalPath to PropertyInitializationInfo.EMPTY))
     }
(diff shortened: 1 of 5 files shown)


jsjeon:
@demiurg906 PTAL this one. After these, there is only one usage left, and I'll continue working on migration.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
```diff
@@ -0,0 +1,228 @@
...
+import org.jetbrains.kotlin.contracts.description.EventOccurrencesRange
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.CFGNode
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.EdgeLabel
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.NormalPath
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.TryExpressionExitNode
+import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
+import org.jetbrains.kotlin.fir.symbols.impl.FirPropertySymbol
+import java.lang.IllegalStateException
+
+// TODO: this can be `sealed` once all of its subtypes are migrated
```
demiurg906, line 19: I think there is no sense to make it `sealed` for two reasons:
1. It's better to write concrete implementation of `ControlFlowInfo` close to checker which uses it
2. In future there will be compiler plugins which may want to provide additional CFA with their own implementations of info

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
```diff
@@ -0,0 +1,228 @@
...
+    var result = one
+    for (symbol in one.keys.union(other.keys)) {
+        val kind1 = one[symbol] ?: EventOccurrencesRange.ZERO
+        val kind2 = other[symbol] ?: EventOccurrencesRange.ZERO
+        result = result.put(symbol, kind1 or kind2)
+    }
+    return result
+}
+
+class LambdaInvocationInfo(
```
demiurg906, line 57: According to previous comment, please move this class back to `FirCallsEffectAnalyzer.kt`

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
```diff
@@ -84,11 +84,126 @@ class PropertyInitializationInfo(
...
-        return super.put(key, value) as PropertyInitializationInfo
+    override val empty: () -> PropertyInitializationInfo =
+        ::EMPTY
+
+    override fun merge(other: PropertyInitializationInfo): PropertyInitializationInfo {
+        return merge(this, other)
+    }
+}
+
+sealed class PathAwareControlFlowInfo<S : ControlFlowInfo<S, *, *>>(
```
demiurg906, line 95: I think it's good to add another generic self-type `P : PathAwareControlFlowInfo<S, P>` (or something like that), to avoid `super.someCall(...) as SpecificImplementation` in all implementations

demiurg906 (changes requested):

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
jsjeon: I see. I'll keep this as a separate, stand-alone, abstract class here.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
jsjeon: Rewrote commits to not move this and other classes in the beginning.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/ControlFlowInfos.kt
jsjeon: Thanks! I wanted to do that, but somehow didn't find a good way to do so, now I understand why `ControlFlowInfo` has that self-type parameter!

jsjeon:
For other issues, https://github.com/JetBrains/kotlin/pull/3899 is about to change CFG and interpretation of labeled info at lambda exit. Then, migration of calls effect analyzer should take into account that too; hence marked this as draft for now.

jsjeon:
@demiurg906 This is ready again. Changes at #3899 and/or adjusted lambda exit handling don't affect this PR at all.

jsjeon:
Friendly ping.

demiurg906:
Sorry for delay.

Thank you for PR, I merged it manually

PR #3890, opened 2020-11-03
JVM IR: Don't generate null checks on value parameters of private operator functions

sfs:
The null check on the extension receiver of private operator functions is a special case and doesn't extend to value parameters.


udalov:
Nice find. Thanks.

PR #3893, opened 2020-11-03
[IR] Correct line number on suspend function with implicit unit return

kandersen:


max-kammerer:
Thank you!

max-kammerer:
Thank you!

PR #3895, opened 2020-11-05
Samples for string last

qurbonzoda:


libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -408,4 +408,17 @@ class Strings {
...
         assertPrints(matchDetails(inputString, toFind, 10), "Searching for 'ever' in 'Never ever give up' starting at position 10: Not found")
     }
+
+    @Sample
+    fun last() {
+        val string = "Kotlin 1.4.0"
+        assertPrints(string.last(), "0")
+        assertPrints(string.last { it.isLetter() }, "n")
+        assertFails { string.last { it > 'z' } }
+        assertPrints(string.lastOrNull { it > 'z' }, "null")
```
ilya-g, line 418: Better to swap `assertFais` and `assertPrints`, because the former will be commented out.

libraries/stdlib/common/src/generated/_Strings.kt
```diff
@@ -139,7 +139,10 @@ public inline fun CharSequence.indexOfLast(predicate: (Char) -> Boolean): Int {
 
 /**
  * Returns the last character.
+ *
  * @throws [NoSuchElementException] if the char sequence is empty.
+ *
+ * @sample samples.text.Strings.last
```
ilya-g, line 145: This is a generated file, please change the template as well.

ilya-g (changes requested):

libraries/stdlib/common/src/generated/_Strings.kt
qurbonzoda: done

qurbonzoda:
Merged manually

PR #3896, opened 2020-11-05
[JVM_IR] Deal with lowering ordering for JvmStatic function references

madsager:
Do not destructively update @JvmStatic functions in objects,
instead create a copy on first access, and replace the original with
the copy in the jvm static lowering. This ensures that the original
function is seen in other lowerings independently of file lowering
order.

### DIFF
--- a/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests/org/jetbrains/kotlin/codegen/ir/FirBlackBoxCodegenTestGenerated.java
@@ -17774,6 +17774,11 @@ public class FirBlackBoxCodegenTestGenerated extends AbstractFirBlackBoxCodegenT
             runTest("compiler/testData/codegen/box/jvmStatic/funAccess.kt");
         }
 
+        @TestMetadata("functionReference.kt")
+        public void testFunctionReference() throws Exception {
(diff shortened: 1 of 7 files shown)


madsager:
This is not pretty and it seems like we have to do these kind of things a lot. I wonder if it is time to reconsider if doing all lowerings one file at a time is the right approach. We might want to discuss if we should move to module level lowerings to avoid these issues.

/cc @pyos @sfs

madsager:
Friendly ping. :)

udalov:
Sorry for the delay, Mads. The changes look good, I'll merge them after a remote run.

The delay was caused by the fact that I was thinking about the file/module phase problem and lowering of lazy IR again after #3865 and this PR, and did an experiment where I've refactored the "JvmStatic in object" phase to a module phase, and also changed the lazy IR correspondingly. It turned out not that bad, and definitely more future-proof than the traditional "map in the context" approach we're using, and now I'm thinking if it's worth it to apply similar refactoring to other phases that change code structure in such non-local way. The branch itself is in progress (there are some FIR tests failing at the moment) at [KOTLIN-CR-4651](https://upsource.jetbrains.com/kotlin/review/KOTLIN-CR-4651).

In any case, it can be done after merging these changes which seem to safely fix this one issue.

udalov:
Thanks!

madsager:
Thanks Alex! :+1:

PR #3899, opened 2020-11-06
FIR checker: handle uncaught exception paths at lambda exits

jsjeon:
The motivation is [KT-43156](https://youtrack.jetbrains.com/issue/KT-43156):
```kt
class A {
    fun test() {
        val a: A
        synchronized(this) {
            if (bar()) throw RuntimeException()
            a = A()
        }
        <!UNINITIALIZED_VARIABLE!>a<!>.bar()  // should be OK
    }
    fun bar() = false
}
```
It turns out that the underlying issue is not quite specific to `synchronized`, since it's defined as:
```kt
inline fun <R> synchronized(lock: Any, block: () -> R): R {
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }

    @Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE", "INVISIBLE_MEMBER")
    monitorEnter(lock)
    try {
        return block()
    }
    finally {
        @Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE", "INVISIBLE_MEMBER")
        monitorExit(lock)
    }
}
```
and as shown in the 1st commit that simply reproduces the issue, we can see a similar problem if we replace `synchronized` with another function that inputs a lambda and has a contract that such lambda is invoked exactly once.

Rather, similar to ab/normal path after `try/finally`, the issue is that abnormal paths still pass after the lambda exit. To that end, the 2nd commit revisits `ThrowExceptionNode` and adds labeled edge to distinguish uncaught exception path; it also revises other uncaught exception paths in `try` expression (in particular, re`throw` in `catch` clauses). Then, at the 3rd commit, the checker is updated to filter out abnormal paths when exiting lambda, which resolves [KT-39709](https://youtrack.jetbrains.com/issue/KT-39709) as well.


jsjeon:
@demiurg906 PTAL this one before https://github.com/JetBrains/kotlin/pull/3889; I'll mark that one as draft for now.

jsjeon:
Friendly ping.

Btw, we plan to track `fir` (and other backends) performance internally, and this is the one last blocking issue for one benchmark. We know how to avoid this (by replacing `val` property with `var`), but would like to keep the benchmark as-is. As expected, the result is promising!

demiurg906:
Sorry, I was too busy last week and forgot to check this PR

Actually you problem that your fix here may be fixed in much more simple way, without any edge labeling. It's enough to don't add exits from anonymous functions to targets of abnormal execution exits (`exitTargetsForTry` stack in `ControlFlowGraphBuilder`) for in-place anonymous functions. With this solution last commit is redundant. 

So I think I accept first commit and remove changes about `addNodeThatReturnsNothing` from second one (all about try/catch/finally in it seems reasonable) and replace it with my fix

demiurg906:
Merged manually

jsjeon:
Good to learn how this could be addressed in a different way. Thank you for keeping changes regarding reconfiguration of `try` expression.

PR #3901, opened 2020-11-09
[Wasm] Enable for loops optimization

skuzmich:


compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/loops/HeaderProcessor.kt
```diff
@@ -382,7 +382,9 @@ internal class IndexedGetLoopHeader(
         with(builder) {
             // loopVariable = objectVariable[inductionVariable]
             val indexedGetFun = with(headerInfo.expressionHandler) { headerInfo.objectVariable.type.getFunction }
-            val get = irCall(indexedGetFun.symbol).apply {
+            // Making sure that expression type has type of the variable when it exists.
+            // Return type of get function can be a type parameter (for example Array<T>::get) which is not a subtype of loopVariable type.
+            val get = irCall(indexedGetFun.symbol, type = loopVariable?.type ?: indexedGetFun.returnType).apply {
```
skuzmich, line 387: @punzki could you take a look?

I think this used to work in JVM because bytecode generator is able to fix type mismatch by inserting casts even without explicit IrTypeOperator. I'm not sure if we should rely on this in IR for common lowerings. WebAssembly backend, I'm working on, never encountered a situation where initialization expression type is not a subtype of variable type other than in this case. Do you mind if I make type conversion explicit in this case? Type difference between generic function return type and call expression type is typically generated by frontend, so I guess it is OK.

bashor (approved):

PR #3902, opened 2020-11-10
FIR: handle reference to property with invisible setter

jsjeon:
The motivation is bb test `callableReference/property/privateSetterOutsideClass.kt`, which is originated from [KT-12337](https://youtrack.jetbrains.com/issue/KT-12337):
```kt
open class Bar(name: String) {
    var foo: String = name
        private set
}

fun box(): String {
    val p1: KProperty1<Bar, String> = Bar::foo
    if (p1 is KMutableProperty<*>) return "Fail: p1 is a KMutableProperty"
    ...
}
```
For `Bar::foo`, its type shouldn't be `KMutableProperty`. Also, `setter` for `IrPropertyReference` shouldn't be set if the target property reference is not mutable.

Checking mutability of a property needs more sophisticated logic than expected: we should consider _where_ the reference is used. E.g., even though a property has a `private` setter, it's still mutable if it's referenced in the same class's members or inner/local classes' members.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
@@ -5,10 +5,7 @@
 
 package org.jetbrains.kotlin.fir.resolve.calls
 
-import org.jetbrains.kotlin.descriptors.Visibilities
-import org.jetbrains.kotlin.fir.FirElement
-import org.jetbrains.kotlin.fir.FirSession
-import org.jetbrains.kotlin.fir.FirSourceElement
+import org.jetbrains.kotlin.fir.*
 import org.jetbrains.kotlin.fir.declarations.*
 import org.jetbrains.kotlin.fir.expressions.FirAnnotationCall
 import org.jetbrains.kotlin.fir.expressions.FirExpression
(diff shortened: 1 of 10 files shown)


jsjeon:
Friendly ping @mglukhikh

mglukhikh:
I remember about this PR. However, it looks scary for me, so I'd like somebody else to look at. Probably @dzharkov is the best person because he is the primary author of callable reference resolve.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
```diff
@@ -396,10 +395,67 @@ fun ConeKotlinType.isKCallableType(): Boolean {
...
+        receiverType,
+        propertyType,
+        isMutable = propertyOrField.canBeMutableReference(callInfo)
     )
 }
+
+private fun FirVariable<*>.canBeMutableReference(callInfo: CallInfo): Boolean {
+    if (!isVar) return false
+    if (this is FirField) return true
+    if (this is FirProperty && this.hasJvmFieldAnnotation) return true
```
dzharkov, line 412: Probably, this check is redundant
JvmField properties cannot have custom setters

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
```diff
@@ -396,10 +395,67 @@ fun ConeKotlinType.isKCallableType(): Boolean {
...
+    return setter == null || setter!!.isVisible(this, callInfo)
+}
+
+private fun FirPropertyAccessor.isVisible(variable: FirVariable<*>, callInfo: CallInfo): Boolean {
```
dzharkov, line 416: I wonder if it's possible to reuse the logic from org.jetbrains.kotlin.fir.FirVisibilityChecker

dzharkov (changes requested):

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
jsjeon: Right, not sure what regressions made me believe this is needed. Removing this doesn't affect anything.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/calls/CallableReferenceResolution.kt
jsjeon: Thanks for the pointer! In doing so, though, I found a couple missed implementations.

For `protected` declarations, owner id is a key to determine visibility, and property accessors, including default accessors, don't have `containingClassAttr`. I added that when possible.

For `private` declarations, if they are top-level, their containing file is a key. `FirProvider`, however, didn't have entries (in `callableContainerMap`) for property accessors. I added that too.

dzharkov:
I've merged your changes manually.
Thanks for your contribution!

PR #3905, opened 2020-11-10
Add always-do-smart-step setting to debugger

emrekultursay:
https://youtrack.jetbrains.com/issue/KT-31558

This feature already exists in Java debugger and should also exist
in Kotlin debugger in order to provide a consistent experience.

The core of the feature is documented below:
https://www.jetbrains.com/help/idea/stepping-through-the-program.html#step-into

### DIFF
--- a/idea/jvm-debugger/jvm-debugger-core/resources/messages/KotlinDebuggerCoreBundle.properties
+++ b/idea/jvm-debugger/jvm-debugger-core/resources/messages/KotlinDebuggerCoreBundle.properties
@@ -13,6 +13,7 @@ line.breakpoint=Line breakpoint
 line.and.lambda.breakpoint=Line and {0,choice,1#Lambda|2#Lambdas} Breakpoints
 
 filter.ignore.internal.classes=Do not step into Kotlin runtime library implementation classes
+always.do.smart.step.into=Always do smart step into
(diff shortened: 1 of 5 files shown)


emrekultursay:
Ping.

yalishevant:
Thank you for the PR, we will receive it but a bit later. For enabling this feature we need to fix SmartStepInto behavior before. There is a number of issues like [KT-35669](https://youtrack.jetbrains.com/issue/KT-35669), [KT-35701](https://youtrack.jetbrains.com/issue/KT-35701), [KT-31804](https://youtrack.jetbrains.com/issue/KT-31804), [KT-33738](https://youtrack.jetbrains.com/issue/KT-33738) and other, on what we are currently working. 
ETA: IJ 2021.2 - 2021.3

nikita-nazarov:
Hello @emrekultursay. Thanks for your contribution! It was a conscious and discussed decision to disable this feature, and as @yalishevant mentioned, the Kotlin debugger stepping is not in an appropriate state for enabling it now due to the complexity of the Kotlin language. Our team is currently working on the debugger quality, and we will definitely add this setting in the future. But it will be more comfortable to do it from our side.

emrekultursay:
FYI, we implemented and enabled this feature in Android Studio Dual (Java+Native) debugger.  Android Studio users can enable always-do-smart-step when debugging Kotlin as well as Java.

PR #3908, opened 2020-11-12
FIR2IR: add implicit NOT_NULL cast for @FlexibleNullability type

jsjeon:
Besides `@EnhancedNullability` type, which could be marked via some `@NotNull` annotations, there is another special type for which we need to add `NOT_NULL` implicit cast: `@FlexibleNullability`. This PR introduces a similar attribute to record `@FlexibleNullability` for platform type; adds `NOT_NULL` implicit cast for flexible type whose lower and upper bounds' nullability don't match; and converts type-associated `@FlexibleNullability`.


jsjeon:
But I have to admit that results aren't 100% good. Most cases look good, but some are in question; will leave some comments on those files directly.

compiler/fir/analysis-tests/testData/enhancement/jsr305/Strange.fir.txt
```diff
@@ -1,9 +1,9 @@
 public open class Strange : R|kotlin/Any| {
-    @R|javax/annotation/Nonnull|(R|javax/annotation/meta/When.UNKNOWN|()) public open field field: R|ft<kotlin/String, kotlin/String?>!|
+    @R|javax/annotation/Nonnull|(R|javax/annotation/meta/When.UNKNOWN|()) public open field field: R|ft<@FlexibleNullability kotlin/String, kotlin/String?>!|
```
jsjeon, line 2: Most of analysis test data changes are about adding either `@FlexibleNullability` or `@EnhancedNullability`. As you can see below, `@FlexibleNullability` is added only if no attributes are found, and then just recognized a few case where `@EnhancedNullability` should be attached. I just extended the logic to load attributes from annotations to cover more non-null annotations.

Anyway, my high-level question is, should this attribute be bound to upper type as well? Currently, it's only associated with lower type.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
```diff
@@ -138,3 +138,19 @@ private fun ConeTypeParameterType.hasNotNullUpperBound(): Boolean {
...
+    coneType.containsNull()
+
+fun ConeKotlinType.containsNull(): Boolean {
+    if (isMarkedNullable) {
+        return true
+    }
+    return when (this) {
+        is ConeFlexibleType -> upperBound.containsNull()
+        is ConeDefinitelyNotNullType -> false
+        is ConeTypeParameterType -> true
```
mglukhikh, line 152: Shouldn't it be dependent to parameter bound type? For example `T : Any` cannot be null.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
```diff
@@ -138,3 +138,19 @@ private fun ConeTypeParameterType.hasNotNullUpperBound(): Boolean {
         }
     }
 }
+
+fun FirTypeRef.containsNull(): Boolean =
```
mglukhikh, line 142: The name is a bit strange. `canBeNull`?

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
```diff
@@ -209,6 +210,9 @@ class Fir2IrImplicitCastInserter(
             expectedType.isUnit -> {
                 coerceToUnitIfNeeded(type, irBuiltIns)
             }
+            valueType.isNullabilityFlexible() && valueType.containsNull() && !expectedType.acceptsNullValues() -> {
```
jsjeon, line 213: For your convenience, this is the counterpart: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt#L352

Note that `IrType.isNullabilityFlexible()` is indeed checking if it's a flexible type and its lower/upper bound's nullability don't match.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
```diff
@@ -218,18 +222,28 @@ class Fir2IrImplicitCastInserter(
...
+        containsNull() || hasEnhancedNullability()
 
     private fun IrExpression.insertImplicitNotNullCastIfNeeded(expression: FirExpression): IrExpression {
         // [TypeOperatorLowering] will retrieve the source (from start offset to end offset) as an assertion message.
         // Avoid type casting if we can't determine the source for some reasons, e.g., implicit `this` receiver.
         if (expression.source == null) {
             return this
         }
+        // Cast type massage 1. Remove @EnhancedNullability
+        // Cast type massage 2. Convert it to a non-null variant (in case of @FlexibleNullability)
```
jsjeon, line 243: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/transformations/InsertImplicitCasts.kt#L392

Two lines there did the same massages.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/EnhancementSignatureParts.kt
```diff
@@ -219,6 +224,14 @@ internal class EnhancementSignatureParts(
                 )
             }
 
+        if (attributesCache.getValue(this) == ConeAttributes.Empty) {
+            val isJavaPlatformClass =
+                this is FirJavaTypeRef && this.toFqName() != null && JavaToKotlinClassMap.isJavaPlatformClass(this.toFqName()!!)
+            if (isJavaPlatformClass) {
```
jsjeon, line 230: According to IR text test data, there are more places that need to record `@FlexibleNullability`, e.g., type parameter, or many other declaration places. I'm just not that familiar with `:fir:java` module yet. :\

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/javaTypeUtils.kt
```diff
@@ -330,5 +330,7 @@ internal fun List<FirAnnotationCall>.computeTypeAttributesForJavaType(): ConeAtt
             CompilerConeAttributes.EnhancedNullability.ANNOTATION_CLASS_ID -> add(CompilerConeAttributes.EnhancedNullability)
             // TODO: Need to handle others too? E.g., COMPATQUAL_NONNULL_ANNOTATION_ID
             in NOT_NULL_ANNOTATION_IDS -> add(CompilerConeAttributes.EnhancedNullability)
+            JAVAX_NONNULL_ANNOTATION_ID -> add(CompilerConeAttributes.EnhancedNullability)
+            // TODO: What about nullable annotations?
```
jsjeon, line 334: Should we add `@FlexibleNullability` for `@Nullable` annotations?

And, what about `javax...CheckForNull`? That sounds like we need `@EnhancedNullability` and add implicit cast to materialize null check?

compiler/testData/codegen/bytecodeText/nullCheckOptimization/deterministicNotNullChecks.kt
```diff
@@ -1,8 +1,4 @@
 // FILE: test/CallableDescriptor.java
-// IGNORE_BACKEND_FIR: JVM_IR
-// Here FIR adds implicit NOT_NULL cast for `origin`, resulting in an assertion being added,
-// which is the correct (yet mismatching) behavior, according to https://youtrack.jetbrains.com/issue/KT-35656
-
```
jsjeon, line 5: Duh, I'm removing my own comments. 🤦  Well, on the bright side, at least now we match with JVM_IR behavior.

compiler/testData/ir/irText/classes/implicitNotNullOnDelegatedImplementation.fir.txt
```diff
@@ -22,9 +22,9 @@ FILE fqName:<root> fileName:/implicitNotNullOnDelegatedImplementation.kt
       BLOCK_BODY
         DELEGATING_CONSTRUCTOR_CALL 'public constructor <init> () [primary] declared in <root>.JFoo'
         INSTANCE_INITIALIZER_CALL classDescriptor='CLASS CLASS name:K1 modality:FINAL visibility:public superTypes:[<root>.JFoo]'
-    FUN FAKE_OVERRIDE name:foo visibility:public modality:OPEN <> ($this:<root>.JFoo) returnType:kotlin.String [fake_override]
+    FUN FAKE_OVERRIDE name:foo visibility:public modality:OPEN <> ($this:<root>.JFoo) returnType:@[FlexibleNullability] kotlin.String [fake_override]
```
jsjeon, line 25: As shown here: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/ir/irText/classes/implicitNotNullOnDelegatedImplementation.kt#L19 nullability assertions are expected at some places, and indeed added here, but with different attribute: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/ir/irText/classes/implicitNotNullOnDelegatedImplementation.txt#L30

Not so sure how it's the other way around... need to study https://github.com/JetBrains/kotlin/commit/8cb8284957f864615c81de8aaaa57781aec52afd :\

compiler/testData/ir/irText/expressions/implicitCastOnPlatformType.fir.txt
```diff
@@ -2,5 +2,6 @@ FILE fqName:<root> fileName:/implicitCastOnPlatformType.kt
   FUN name:test visibility:public modality:FINAL <> () returnType:kotlin.String
     BLOCK_BODY
       RETURN type=kotlin.Nothing from='public final fun test (): kotlin.String declared in <root>'
-        CALL 'public open fun getProperty (p0: kotlin.String?): kotlin.String? declared in java.lang.System' type=kotlin.String? origin=null
-          p0: CONST String type=kotlin.String value="test"
+        TYPE_OP type=kotlin.String origin=IMPLICIT_NOTNULL typeOperand=kotlin.String
+          CALL 'public open fun getProperty (p0: kotlin.String?): kotlin.String? declared in java.lang.System' type=kotlin.String? origin=null
```
jsjeon, line 6: parameter `p0`, return type `kotlin.String?` need `@FlexibleNullability`, but no idea where to look at yet...

compiler/testData/ir/irText/stubs/builtinMap.fir.txt
```diff
@@ -18,7 +18,7 @@ FILE fqName:<root> fileName:/builtinMap.kt
             if: CONST Boolean type=kotlin.Boolean value=true
             then: CALL 'public final fun apply <T> (block: @[ExtensionFunctionType] @[ExtensionFunctionType] kotlin.Function1<T of kotlin.apply, kotlin.Unit>): T of kotlin.apply [inline] declared in kotlin' type=java.util.LinkedHashMap<K1 of <root>.plus?, V1 of <root>.plus?> origin=null
               <T>: java.util.LinkedHashMap<K1 of <root>.plus?, V1 of <root>.plus?>
-              $receiver: CONSTRUCTOR_CALL 'public constructor <init> (p0: kotlin.collections.Map<out K of java.util.LinkedHashMap?, out V of java.util.LinkedHashMap?>?) declared in java.util.LinkedHashMap' type=java.util.LinkedHashMap<K1 of <root>.plus?, V1 of <root>.plus?> origin=null
+              $receiver: CONSTRUCTOR_CALL 'public constructor <init> (p0: @[FlexibleNullability] kotlin.collections.Map<out K of java.util.LinkedHashMap?, out V of java.util.LinkedHashMap?>?) declared in java.util.LinkedHashMap' type=java.util.LinkedHashMap<K1 of <root>.plus?, V1 of <root>.plus?> origin=null
```
jsjeon, line 21: Here, e.g., type parameters, `K`, `V`, 'K1`, and `V1`, need the same attribute, but not sure where it could be added/propagated...

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
jsjeon: I followed this one: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.psi2ir/src/org/jetbrains/kotlin/psi2ir/KotlinUtils.kt#L36, but yeah, sounds strange.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
jsjeon: Oh, you're right, I was too naive here. Will update it as per suggestion.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
mglukhikh: Yes, I confirm this is consistent with FE 1.0.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrImplicitCastInserter.kt
mglukhikh: Confirmed

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/EnhancementSignatureParts.kt
mglukhikh: This check are suspicious for me

`if (attributesCache.getValue(this) == ConeAttributes.Empty)`

I think you are trying to check something like "There is no enhanced nullability", aren't you? May be it's better to modify the check accordingly? (just in case some other attributes will appear in future, which can break logic here)

mglukhikh: About more places, well, it's a good place to start.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/javaTypeUtils.kt
mglukhikh: As I know, all Java not-null annotations should provoke `@EnhancedNullability`. About `@FlexibleNullability`, I think it should be only on flexible types when bounds nullability is different.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/EnhancementSignatureParts.kt
mglukhikh: Also, why this check `if (isJavaPlatformClass)` is needed? As I know, all Java types with flexible nullability should have this annotation...

compiler/testData/ir/irText/expressions/implicitCastOnPlatformType.fir.txt
mglukhikh: Citing myself:
```
Also, why this check `if (isJavaPlatformClass)` is needed? As I know, all Java types with flexible nullability should have this annotation...
```

compiler/testData/ir/irText/stubs/builtinMap.fir.txt
mglukhikh: I'd say that type parameter based type should have this attribute if the referenced Java type parameter does not have not-null bound.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
jsjeon: Rewrote this to consider parameter bound types.

compiler/fir/tree/src/org/jetbrains/kotlin/fir/types/FirTypeUtils.kt
jsjeon: Renamed.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/EnhancementSignatureParts.kt
jsjeon: Right, my intention was to check something like that, and right, again, that this is quite fragile.

Re: platform class, I saw many previous commits mentioned platform type for `@FlexibleNullability`, so I assumed it's bound to platform type only. :p

Anyway, this is completely rewritten, and not needed anymore.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/enhancement/javaTypeUtils.kt
jsjeon: Thanks to this comment, I completely rework how those attributes are added.

In the 1st commit, as a separate work, more Java non-null annotations are converted.

In the 2nd commit, instead of checking, e.g., if there isn't any other attributes, or if this is a platform type, I visited the places where _flexible_ type _with_ different nullability lower/upper bounds are generated, and added `@FlexibleNullability` to upper bound (where `nullable` is flipped).

While doing so, I relied on `union` of attributes, which seems like a natural way to accumulate attributes, and thanks to that, I found hidden bugs of that logic.

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/CompilerConeAttributes.kt
```diff
@@ -50,20 +50,32 @@ object CompilerConeAttributes {
     object ExtensionFunctionType : ConeAttribute<ExtensionFunctionType>() {
         val ANNOTATION_CLASS_ID = ClassId(FqName("kotlin"), Name.identifier("ExtensionFunctionType"))
 
-        override fun union(other: ExtensionFunctionType?): ExtensionFunctionType? = other
-        override fun intersect(other: ExtensionFunctionType?): ExtensionFunctionType? = this
+        override fun union(other: ExtensionFunctionType?): ExtensionFunctionType? = this
+        override fun intersect(other: ExtensionFunctionType?): ExtensionFunctionType? = other
```
jsjeon, line 54: So, the underlying logic is, for the same key, we will retrieve `ConeAttribute` from both side, and only non-null result (of operation) will be recorded.

It is obvious that `this` is non-null. For `union`, returning non-null `this` will act like `or`. For `intersect`, returning the other one, which could be `null`, will act like `and`.

So, with previous implementation (that returns `other` for `union`, which could be `null`), I saw that `union` always returns `EMPTY` attributes!

compiler/fir/analysis-tests/testData/enhancement/jsr305/Strange.fir.txt
jsjeon: > my high-level question is, should this attribute be bound to upper type as well? Currently, it's only associated with lower type.

Answer to my own question: since this attribute is added to flexible type whose lower/upper bounds' nullability don't match, it makes sense to add this to upper bound instead. However, as long as the retrieval of attributes takes into account both lower and upper bound, it doesn't matter technically. Such `union` logic is just added in the 2nd commit.

compiler/testData/ir/irText/classes/implicitNotNullOnDelegatedImplementation.fir.txt
jsjeon: After changing the whole logic (from enhancement point to flexible type creation point), this is no longer the case. I think there is an override checker in `java` module, and that's where to propagate attributes on the return type of overridden functions to overriding functions.

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/CompilerConeAttributes.kt
demiurg906: I strongly disagree with those changes. `union` and `intersect` define how attributes should be passed throw common super type calculation (`intersect` for intersection types and `union` for union types (which will be supported somewhere in future)). And for this attributes we have strong rules:
- `Exact` and `NoInfer` stays only on upper type and don't flow throw inference
- If on component of intersection type is `ExtensionFuncitonType` then the whole type also is

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/ConeAttributes.kt
```diff
@@ -65,7 +66,7 @@ class ConeAttributes private constructor(attributes: List<ConeAttribute<*>>) : A
     private inline fun perform(other: ConeAttributes, op: ConeAttribute<*>.(ConeAttribute<*>?) -> ConeAttribute<*>?): ConeAttributes {
         if (this.isEmpty() && other.isEmpty()) return this
         val attributes = mutableListOf<ConeAttribute<*>>()
-        for (index in indices) {
+        for (index in this.typeRegistry.indices.union(other.typeRegistry.indices)) {
```
demiurg906, line 69: `indicies` here is a reference to property of companion object of `ConeAttributes` (which is a `typeRegistry` for any `ConeAttributes` object), so there is no need in this change

compiler/fir/cones/src/org/jetbrains/kotlin/fir/utils/ArrayMapOwner.kt
```diff
@@ -70,6 +70,6 @@ abstract class TypeRegistry<K : Any, V : Any> {
         return idPerType.computeIfAbsent(kClass) { idCounter.getAndIncrement() }
     }
 
-    protected val indices: Collection<Int>
+    internal val indices: Collection<Int>
```
demiurg906, line 73: This is also not needed due to previous comment

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/ConeTypes.kt
```diff
@@ -126,7 +126,7 @@ open class ConeFlexibleType(
         get() = lowerBound.nullability.takeIf { it == upperBound.nullability } ?: ConeNullability.UNKNOWN
 
     override val attributes: ConeAttributes
-        get() = lowerBound.attributes
+        get() = lowerBound.attributes.union(upperBound.attributes)
```
demiurg906, line 129: It's designed that `FlexibleType.attributes == lowerBound.attributes`

demiurg906 (changes requested):

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/ConeTypes.kt
jsjeon: Thanks for the clarification. So, this attribute should be associated with lower bound. Reverted this part to the original expression.

compiler/fir/analysis-tests/testData/enhancement/jsr305/Strange.fir.txt
jsjeon: Correction to my own answer: this should be recorded at lower bound, since the attribute of flexible type is supposed to be that of lower bound.

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/ConeAttributes.kt
jsjeon: Yeah, all of this happened due to my misunderstanding on union/intersect types. Reverted.

compiler/fir/cones/src/org/jetbrains/kotlin/fir/utils/ArrayMapOwner.kt
jsjeon: Right, reverted.

compiler/fir/cones/src/org/jetbrains/kotlin/fir/types/CompilerConeAttributes.kt
jsjeon: Thanks for the detailed explanation. I somehow still/sometimes misunderstood `union` and `intersect` types. :\ If these attributes are used for such super type calculation, yeah, the original expression for `union`/`intersect` make sense. I just need to use `intersect` where I used `union` previously.

compiler/testData/ir/irText/stubs/builtinMap.fir.txt
jsjeon: At least `K` and `V` have that attribute for now. Not sure about `K1` and `V1`...

mglukhikh:
Merged manually, thank you!

PR #3910, opened 2020-11-13
FIR2IR: convert recursive captured type

jsjeon:
The motivation is [KT-43346](https://youtrack.jetbrains.com/issue/KT-43346) where type converter raises stack overflow while converting recursive captured type, e.g., `Recursive<R> where R : Recursive<R>, ...`. According to `psi2ir` result, it's okay to convert type parameter `R` in this case. So, we can keep track of what captured types are visited during the conversion, and when we detect the recursion, we can rather switch to convert type parameter, in lieu of super type chain.

### DIFF
--- a/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
+++ b/compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
@@ -58,7 +58,11 @@ class Fir2IrTypeConverter(
         StandardClassIds.Char to irBuiltIns.charType
     )
 
+    private val capturedTypeCache = mutableMapOf<ConeCapturedType, IrType>()
+    private val capturedTypeStack = mutableSetOf<ConeCapturedType>()
+
     fun FirTypeRef.toIrType(typeContext: ConversionTypeContext = ConversionTypeContext.DEFAULT): IrType {
+        capturedTypeStack.clear()
         return when (this) {
             !is FirResolvedTypeRef -> createErrorType()
             !is FirImplicitBuiltinTypeRef -> type.toIrType(typeContext, annotations)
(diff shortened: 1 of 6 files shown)


compiler/testData/ir/irText/firProblems/recursiveCapturedTypeInPropertyReference.txt
```diff
@@ -0,0 +1,68 @@
...
+            <T>: <root>.Recursive<*>
+            <R>: <root>.AbstractSymbol<out <root>.Recursive<*>>
+            $receiver: CALL 'public final fun filterIsInstance <R> (): kotlin.collections.List<@[NoInfer] R of kotlin.collections.filterIsInstance> [inline] declared in kotlin.collections' type=kotlin.collections.List<@[NoInfer] <root>.Recursive<*>> origin=null
+              <R>: <root>.Recursive<*>
+              $receiver: GET_VAR 'list: kotlin.collections.List<kotlin.Any> declared in <root>.AbstractSymbol.foo' type=kotlin.collections.List<kotlin.Any> origin=null
+            transform: PROPERTY_REFERENCE 'public abstract symbol: <root>.AbstractSymbol<R of <root>.Recursive> [val]' field=null getter='public abstract fun <get-symbol> (): <root>.AbstractSymbol<R of <root>.Recursive> declared in <root>.Recursive' setter=null type=kotlin.reflect.KProperty1<<root>.Recursive<*>, <root>.AbstractSymbol<out <root>.Recursive<*>>> origin=null
```
jsjeon, line 55: Note that the only differences I can see now are:
* lack of `@NoInfer` attribute, which is out of scope of this change
* lack of `out` variance: `AbstractSymbol<out <root>.Recursive<*>>>` v.s. `AbstractSymbol<<root>.Recursive<*>>`
See the 2nd type argument of property reference where the original stack overflow happened.

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrTypeConverter.kt
```diff
@@ -58,7 +58,11 @@ class Fir2IrTypeConverter(
         StandardClassIds.Char to irBuiltIns.charType
     )
 
+    private val capturedTypeCache = mutableMapOf<ConeCapturedType, IrType>()
+    private val capturedTypeStack = mutableSetOf<ConeCapturedType>()
```
mglukhikh, line 62: Not sure both these structures are needed.... Will do some experiments...

mglukhikh:
Merged with minor improvements.

PR #3911, opened 2020-11-17
FIR: resolve Throwable.stackTrace synthetic property

jsjeon:
The motivation is [KT-43339](https://youtrack.jetbrains.com/issue/KT-43339), where `java/lang/Throwable.(get|set)StackTrace` is not property processed by `JvmMappedScope`. The main reason is that return type (or parameter type) of `getStackTrace` (or `setStackTrace`) is `Array<java/lang/StackTraceElement>`, and JVM descriptor util didn't convert array appropriately, returning just `Array`, which should have been `[Ljava/lang/StackTraceElement;`.


mglukhikh:
You should probably write `// FULL_JDK` in test.

jsjeon:
> You should probably write `// FULL_JDK` in test.

Duh, yeah, that's magic :\ So, the issue was rather JVM signature conversion when array type is involved in.

mglukhikh:
Merged manually, thank you!

PR #3912, opened 2020-11-17
Make random implementations serializable

CommanderTvis:
Make Random and XorWowRandom expect classes and add java.io.Serializable
interface to JVM Random implementations where possible.

Related issue: https://youtrack.jetbrains.com/issue/KT-25571


CommanderTvis:
The problem is that `kotlin.random.Random` objects can't be passed as Serializable or stored inside other Serializable objects.

So, this PR makes `kotlin.random.Random` and `kotlin.random.XorWowRandom` `expect` classes to provide `java.io.Serializable` interface in the JVM platform. Functions `defaultPlatformRandom` and `doubleFromParts` were removed and inlined inside the `actual` declaration of `Random`. 

JVM platform `Random` implementations were made objects and were marked interface `java.io.Serializable` too.

ilya-g:
We'd better avoid code duplication at that scale. Could you elaborate why it is required to split so much code into expect and actual?

CommanderTvis:
> We'd better avoid code duplication at that scale. Could you elaborate why it is required to split so much code into expect and actual?

Because only `expect/actual class` can implement an interface only on a certain platform.

ilya-g:
Still better to find some workarounds. For example introduce an internal expect Serializable interface and implement it in common classes. In fact, there's already one in stdlib — `kotlin.io.Serializable`.

CommanderTvis:
JVM Serializable also requires weird methods like `readResolve`. It makes no sense to have them on other platforms (even as internal or private)

ilya-g:
Private unused methods are optimized out during tree shaking, so they won't hurt either. Still better than code duplication.

CommanderTvis:
@ilya-g OK, I've deduplicated the classes.

libraries/stdlib/jvm/src/kotlin/internal/PlatformImplementations.kt
```diff
@@ -51,20 +52,28 @@ internal val IMPLEMENTATIONS: PlatformImplementations = run {
     val version = getJavaVersion()
     if (version >= 0x10008) {
         try {
-            return@run castToBaseType<PlatformImplementations>(Class.forName("kotlin.internal.jdk8.JDK8PlatformImplementations").newInstance())
-        } catch (e: ClassNotFoundException) { }
+            return@run castToBaseType<PlatformImplementations>(
+                Class.forName("kotlin.internal.jdk8.JDK8PlatformImplementations").newInstance()
+            )
+        } catch (e: ClassNotFoundException) {
+        }
```
ilya-g, line 59: Please revert formatting changes unrelated to the code regions you have changed.

libraries/stdlib/jdk8/src/kotlin/random/jdk8/PlatformThreadLocalRandom.kt
```diff
@@ -8,15 +8,20 @@ package kotlin.random.jdk8
 import java.util.concurrent.ThreadLocalRandom
 
 @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER", "CANNOT_OVERRIDE_INVISIBLE_MEMBER")
-internal class PlatformThreadLocalRandom : kotlin.random.AbstractPlatformRandom() {
+internal object PlatformThreadLocalRandom : kotlin.random.AbstractPlatformRandom(), Serializable {
+    private const val serialVersionUID = -8182665159126504406L
```
ilya-g, line 12: Why such `serialVersionUID` values, do they mean something? I believe we can start with `serialVersionUID = 1L` for all serializable entities.

libraries/stdlib/jdk8/src/kotlin/random/jdk8/PlatformThreadLocalRandom.kt
```diff
@@ -8,15 +8,20 @@ package kotlin.random.jdk8
...
     override fun nextDouble(until: Double): Double = ThreadLocalRandom.current().nextDouble(until)
 
-//     do not delegate this, as it's buggy in JDK8+ (up to 11 at the moment of writing)
-//     override fun nextDouble(from: Double, until: Double): Double = ThreadLocalRandom.current().nextDouble(from, until)
-}
\ No newline at end of file
+    //     do not delegate this, as it's buggy in JDK8+ (up to 11 at the moment of writing)
+    //     override fun nextDouble(from: Double, until: Double): Double = ThreadLocalRandom.current().nextDouble(from, until)
+
+    private fun readResolve(): Any = PlatformThreadLocalRandom
```
ilya-g, line 26: I think such internal implementations should not be serializable at all. 
`Random.Default` manages its singleton serialization entirely and therefore it shouldn't expose any implementation details in a serialized form.

libraries/stdlib/jvm/src/kotlin/random/PlatformRandom.kt
```diff
@@ -47,17 +47,26 @@ internal abstract class AbstractPlatformRandom : Random() {
...
+
     override val impl: java.util.Random
         get() = implStorage.get()
+
+    private fun readResolve(): Any = FallbackThreadLocalRandom
 }
 
-private class PlatformRandom(override val impl: java.util.Random) : AbstractPlatformRandom()
+private class PlatformRandom(override val impl: java.util.Random) : AbstractPlatformRandom(), Serializable {
+    companion object {
```
ilya-g, line 66: Companion object with just a single private const should be private.

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -267,10 +267,12 @@ public abstract class Random {
      *
      * @sample samples.random.Randoms.defaultRandom
      */
-    companion object Default : Random() {
-
+    companion object Default : Random(), Serializable {
+        private const val serialVersionUID = 6449517406788167720L
         private val defaultRandom: Random = defaultPlatformRandom()
```
ilya-g, line 272: Better to use `writeReplace` to avoid exposing `defaultRandom` field.

libraries/stdlib/jdk8/src/kotlin/internal/jdk8/JDK8PlatformImplementations.kt
```diff
@@ -29,13 +29,13 @@ internal open class JDK8PlatformImplementations : JDK7PlatformImplementations()
...
 
-        val range = matcher.start(name)..matcher.end(name) - 1
-        return if (range.start >= 0)
+        val range = matcher.start(name) until matcher.end(name)
+        return if (range.first >= 0) {
             MatchGroup(matcher.group(name), range)
-        else
+        } else
             null
     }
```
ilya-g, line 37: This change should be in a separate commit, or even better, in a separate PR.

ilya-g (changes requested):

libraries/stdlib/jdk8/src/kotlin/random/jdk8/PlatformThreadLocalRandom.kt
CommanderTvis: I used JDK [serialver](https://docs.oracle.com/javase/7/docs/technotes/tools/solaris/serialver.html) utility that creates something like hash, which has to be changed if the class's structure is changed

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: What should `writeReplace` return then?

CommanderTvis: @ilya-g

ilya-g: Just an empty serializable class whose `readResolve` returns `Random.Default`.

libraries/stdlib/jdk8/src/kotlin/random/jdk8/PlatformThreadLocalRandom.kt
```diff
@@ -8,7 +8,7 @@ package kotlin.random.jdk8
 import java.util.concurrent.ThreadLocalRandom
 
 @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER", "CANNOT_OVERRIDE_INVISIBLE_MEMBER")
-internal class PlatformThreadLocalRandom : kotlin.random.AbstractPlatformRandom() {
+internal object PlatformThreadLocalRandom : kotlin.random.AbstractPlatformRandom() {
```
ilya-g, line 11: I don't see much sense in transforming it to `object`. It is instantiated once and then is kept in a private property of `Random` companion object. So no need to have another instance where its reference is stored.
Same for `FallbackThreadLocalRandom`

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -267,10 +267,12 @@ public abstract class Random {
      *
      * @sample samples.random.Randoms.defaultRandom
      */
-    companion object Default : Random() {
-
+    companion object Default : Random(), Serializable {
+        private const val serialVersionUID = 6449517406788167720L
```
ilya-g, line 271: I still don't get what bad can happen if we set `serialVersionUID = 1L` in all such classes?

libraries/stdlib/jvm/test/random/RandomJvmSerializationTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import test.io.serializeAndDeserialize
+import kotlin.random.Random
+import kotlin.random.asJavaRandom
+import kotlin.random.asKotlinRandom
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertSame
+
+class RandomJvmSerializationTest {
```
ilya-g, line 16: After we settle with serialized representation and serialVersionUID, let's add some serialization stability tests.
Something like https://github.com/JetBrains/kotlin/blob/1.4.20/libraries/stdlib/jvm/test/collections/CollectionJVMTest.kt#L192-L207

libraries/stdlib/js/src/kotlin/random/PlatformRandom.kt
```diff
@@ -13,5 +13,6 @@ internal actual fun defaultPlatformRandom(): Random =
 
 private val INV_2_26: Double = 2.0.pow(-26)
 private val INV_2_53: Double = 2.0.pow(-53)
+
 internal actual fun doubleFromParts(hi26: Int, low27: Int): Double =
-    hi26 * INV_2_26 + low27 * INV_2_53
\ No newline at end of file
+    hi26 * INV_2_26 + low27 * INV_2_53
```
ilya-g, line 18: Unrelated change

libraries/stdlib/src/kotlin/random/Random.kt
ilya-g: On the other hand, we could rely on the fact that companion object doesn't have any fields and keeps all its state as static in the containing class, but IMO that's a too fragile assumption.

CommanderTvis: Then I'll make defaultRandom a property without backing field, since *ThreadLocalRandom is an `object`.

.idea/misc.xml
```diff
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="ASMIdeaPluginConfiguration">
+    <asm skipDebug="false" skipFrames="false" skipCode="false" expandFrames="false" />
+    <groovy codeStyle="LEGACY" />
+  </component>
```
ilya-g, line 6: Unrelated change

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -267,9 +267,13 @@ public abstract class Random {
      *
      * @sample samples.random.Randoms.defaultRandom
      */
-    companion object Default : Random() {
+    companion object Default : Random(), Serializable {
+        private const val serialVersionUID = 0L
 
-        private val defaultRandom: Random = defaultPlatformRandom()
+        private val defaultRandom: Random
+            get() = defaultPlatformRandom()
```
ilya-g, line 274: We definitely do not want to evaluate this again on each call to Random.Default's methods. Please revert to storing the implementation here and resort to serialization surrogate approach.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun defaultIsSerializable() {
+        val instance = Random
+        assertSame(instance, serializeAndDeserialize(instance))
+    }
+
+    private fun testPersistedDeserialization(hexValue: String, expected: Any) =
+        assertEquals(expected = expected, actual = deserializeFromHex(hexValue))
+
+    @Test
+    fun deserializeDefault() = testPersistedDeserialization(
```
ilya-g, line 28: In this test we should use `assertSame` check. In the test with predictable Random, we can compare several `next*` values with the expected instance.

ilya-g (changes requested):

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: In JVM, this method will be 100% inlined.

.idea/misc.xml
CommanderTvis: Oh, sorry.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: How many of them?

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: I mean, method `kotlin/random/Random$Companion.getDefaultRandom()Lkotlin/random/Random;` will be replaced with direct access to `??????ThreadLocalRandom.INSTANCE : L???;`

.idea/misc.xml
CommanderTvis: It's so cumbersome to have IDEA settings saved automatically.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: @ilya-g

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: @ilya-g of course, I can discard this change, but it shouldn't make any overhead on all the platforms Kotlin supports.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlin.random.asKotlinRandom
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertSame
+
+class RandomSerializationTest {
+    @Test
+    fun defaultIsSerializable() {
+        val instance = Random
+        assertSame(instance, serializeAndDeserialize(instance))
```
CommanderTvis, line 21: @ilya-g this test already checks that Random.default instances are same when serialized.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun deserializeDefault() = testPersistedDeserialization(
+        "ac ed 00 05 73 72 00 1c 6b 6f 74 6c 69 6e 2e 72 61 6e 64 6f 6d 2e 52 61 6e 64 6f 6d 24 44 65 66 61 75 6c 74 59 81 49 e9 14 4e a8 28 02 00 00 78 70",
+        Random
+    )
+
+    @Test
+    fun xorwowIsSerializable() {
+        val instance = Random(0)
+        val deserialized = serializeAndDeserialize(instance)
+        assertEquals(instance.nextInt(), deserialized.nextInt())
```
CommanderTvis, line 37: @ilya-g of course, we can check more than first 4 bytes the deserialized random produces.

libraries/stdlib/src/kotlin/random/Random.kt
ilya-g: Better not to speculate about platform optimization capabilities, especially considering that this code can run on other VMs such as Android runtime, when we can write it in a way that make it sure happen without any optimization.
In other Kotlin platforms, the changes caused by this one are also totally unnecessary.

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
ilya-g: I believe one call for each of the following methods would be enough: `nextInt()`, `nextLong()`, `nextDouble()`, `nextBytes()`

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
ilya-g: It doesn't catch the situation when we change the default instance in an incompatible way.

ilya-g (changes requested):

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: Clear

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: In the current state of changes, `default` has no serialized fields and has constant `serialVersionUID`. Instances will be persistent until we add non-transient field or change `serialVersinoUID`. I'm not sure how to test such a situation well.

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -290,7 +295,11 @@ public abstract class Random {
 
         override fun nextBytes(array: ByteArray): ByteArray = defaultRandom.nextBytes(array)
         override fun nextBytes(size: Int): ByteArray = defaultRandom.nextBytes(size)
-        override fun nextBytes(array: ByteArray, fromIndex: Int, toIndex: Int): ByteArray = defaultRandom.nextBytes(array, fromIndex, toIndex)
+        override fun nextBytes(array: ByteArray, fromIndex: Int, toIndex: Int): ByteArray = defaultRandom.nextBytes(
+            array,
+            fromIndex,
+            toIndex
+        )
```
ilya-g, line 302: This change is not related. Better to keep it consistent with the other function definitions.

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -267,10 +268,14 @@ public abstract class Random {
      *
      * @sample samples.random.Randoms.defaultRandom
      */
-    companion object Default : Random() {
+    companion object Default : Random(), Serializable {
+        private const val serialVersionUID = 0L
 
+        @Transient
```
ilya-g, line 274: Ok, I think the variant with all transient fields and `readResolve` will do (without introducing a serialization surrogate).

ilya-g (commented):
Please rebase your branch on the latest `master` and ensure there are no merge commits in it.

libraries/stdlib/src/kotlin/random/Random.kt
```diff
@@ -267,10 +268,14 @@ public abstract class Random {
      *
      * @sample samples.random.Randoms.defaultRandom
      */
-    companion object Default : Random() {
+    companion object Default : Random(), Serializable {
+        private const val serialVersionUID = 0L
```
ilya-g, line 272: Bad news: this const doesn't work as we intended it to. The reason is that it is compiled as a static field to the `Random` class, and not to the `Random.Default`.
So changing its value doesn't fail the test with persistent deserialization, and on the opposite, adding a public method to `Random.Default` fails the test, which indicates that `serialVersionUID` for this class is still automatic.

Given the above I think we have no other way than to use a serialization surrogate.

ilya-g (changes requested):

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: What do you mean by serialization surrogate?

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals(first.nextInt(), second.nextInt())
+        assertEquals(first.nextDouble(), second.nextDouble())
+        assertEquals(first.nextLong(), second.nextLong())
+        assertTrue(first.nextBytes(64).contentEquals(second.nextBytes(64)))
+    }
+
+    @Test
+    fun deserializeDefault() = assertSame(
+        expected = Random,
+        actual = deserializeFromHex("ac ed 00 05 73 72 00 1c 6b 6f 74 6c 69 6e 2e 72 61 6e 64 6f 6d 2e 52 61 6e 64 6f 6d 24 44 65 66 61 75 6c 74 59 81 49 e9 14 4e a8 28 02 00 00 78 70")
```
ilya-g, line 43: Please also add a similar test for XorwowRandom.

libraries/stdlib/src/kotlin/random/Random.kt
CommanderTvis: Clear

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: OK

libraries/stdlib/src/kotlin/random/Random.kt
ilya-g: Can be a nested class in `Random` or `Random.Default`

libraries/stdlib/jvm/test/random/RandomSerializationTest.kt
CommanderTvis: It is impossible since serialization produces a new instance with state restored from bytes, so instances of XorWow will be neither same nor equal (since XorWow is not a data class)

ilya-g: Indeed, that's why I said "a similar", not "the same" :)
You can use `testRandomsHaveSameState` to check that the state has been restored correctly.

CommanderTvis: OK

ilya-g (approved):

CommanderTvis:
@ilya-g how to run API dump?

ilya-g:
To perform API dump, you can use two shared run configurations in Kotlin project:
- `Test: public API binary compatibility validator, overwrite results` (dumps JVM public binary API)
- `Test: stdlib-js public kotlin api test, overwrite results` (dumps JS source API)

ilya-g:
I've merged it. Thank you for the good work!

PR #3913, opened 2020-11-17
Add a `String.sarcastic()` easter egg. 

AdamMc331:
Easter eggs in programming languages can be a quick way to brighten a developers day. This has precedence in the Android ecosystem with things like [Log.wtf](https://developer.android.com/reference/android/util/Log#wtf(java.lang.String,%20java.lang.String)) or [isUserAGoat](https://developer.android.com/reference/android/os/UserManager#isUserAGoat()).

One way we could bring this simple humor to the Kotlin language is with an extension function that randomizes the case of each character, playing off of the [mocking SpongeBob](https://knowyourmeme.com/memes/mocking-spongebob) meme.  

Let me know what you think!

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
+++ b/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt
@@ -15,6 +15,7 @@ import java.nio.charset.Charset
 import java.nio.charset.CodingErrorAction
 import java.util.Locale
 import java.util.regex.Pattern
+import kotlin.random.Random
 
 
 /**
@@ -722,3 +723,14 @@ public actual fun CharSequence.repeat(n: Int): String {
  */
 public actual val String.Companion.CASE_INSENSITIVE_ORDER: Comparator<String>
     get() = java.lang.String.CASE_INSENSITIVE_ORDER
+
+/**
+ * Returns the string with the case of each character randomized, to emphasize sarcasm in the resulting text.
+ */
+public fun String.sarcastic(random: Random = Random): String = buildString {
+    this@sarcastic.forEach {
+        val mocking = if (random.nextBoolean()) Char::toLowerCase else Char::toUpperCase
+
+        append(mocking(it))
+    }
+}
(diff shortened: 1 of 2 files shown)


arkivanov:
Please mind the Turkish locale [problem](https://stackoverflow.com/questions/52402151/how-to-uppercase-turkish-i). 😄

handstandsam:
It's cute, but I always try to make software as small as possible. This is a fun parody extension function but doesn't belong in Kotlin in my opinion. Sorry Adam 🙁

AdamMc331:
> It's cute, but I always try to make software as small as possible. This is a fun parody extension function but doesn't belong in Kotlin in my opinion. Sorry Adam 🙁

This is a valid concern and something I weighed before opening a real PR. I agree with the principle that we should be avoiding unnecessary code that has to be maintained when there are definitely much more important problems for the language to be focusing on. The reason I decided it was worth having a conversation is because I felt a 7 line function was a small enough trade off for a sensible chuckle. 

I'd still love to hear what the JetBrains team thinks. Ultimately, I want what's best for the language, and if the team feels this isn't worth an exception there's definitely no hard feelings. :)

mikel063093:
Approve

tjbarber:
I love this. For only 7 lines, it gave me a good chuckle and I have to think it would do the same for other developers.

Mauker1:
I absolutely loved it. It's really short and gave me a good laugh, please merge this!

LifeIsStrange:
This should not be merged, it's not about the size it adds it's about reducing the signal to noise ratio of autocomplete!
When the ide autocomplete is triggered for myStr.s
I want to see slice(), split(), etc not 
saRcAStiC(). 
Such a method shouldn't be on the doc too as it reduce once again signal to noise ratio. 
So it would need to be the first method to be not shown of the official documentation and to be manually hidden from the IDE Autocompletion, such autocomplete exception add complexity and might be a slow down (and autocompletion really should be faster). And such exception should be added to the kotlin community lsp too.

All this burden for a joke that people will not really appreciate outside of this PR thread make it not worth it to merge, still thanks for the laugh!

AdamMc331:
I've decided to go back on this and close it. Grateful that we had the conversation about this though. :)

PR #3914, opened 2020-11-17
FIR2IR: convert annotations on delegated members

jsjeon:
The motivation is bb test `throws/delegationAndThrow(_1_3)?`:
```kt
interface A {
    @Throws(IOException::class)
    fun foo()
}
// FILE: B.kt
class B(a: A) : A by a
```
where `B::foo` is supposed to have the checked exception type. We simply need to convert annotations on delegated members.

Note that, as per [KT-35834](https://youtrack.jetbrains.com/issue/KT-35834), checked exception declared in Kotlin interface should not be carried to JVM bytecode (since language version 1.4). `throws/delegationAndThrow_1_3` is checking if such `@Throws` is associated with `B::foo`, whereas `throws/delegationAndThrow` is checking if such checked exception is correctly dropped. Since we don't convert annotations at all, the latter has passed. This change adds unconditional conversions, which is fine because JVM backend has a logic to discard `@Throws` according to language version and interface delegation.


compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
```diff
@@ -59,7 +59,7 @@ class Fir2IrDeclarationStorage(
     private val moduleDescriptor: FirModuleDescriptor
 ) : Fir2IrComponents by components {
 
-    internal var annotationGenerator: AnnotationGenerator? = null
+    private val annotationGenerator = AnnotationGenerator(this)
```
mglukhikh, line 62: I think it's better to add `annotationGenerator` to components

mglukhikh (changes requested):
Everything other looks good

compiler/fir/fir2ir/src/org/jetbrains/kotlin/fir/backend/Fir2IrDeclarationStorage.kt
jsjeon: Indeed, it is created five different places. :o Added it to `Fir2IrComponents`. I also grouped elements in the components into some relevant pieces; feel free to mingle them.

mglukhikh (approved):

mglukhikh:
Merged manually, thank you! 👍

PR #3917, opened 2020-11-19
1.4.20 changelog

woainikk:


schernov-xo:
Perfect version update. Only one comment: IMHO the release should not be announced (many people get e-mail notifications about it) until all artifacts are available on repos. I mean gradle plugins of `1.4.20` version (did not exist yesterday, today they are available).
Thanks! Kotlin 😍

PR #3918, opened 2020-11-19
Register Kotlin MPP source sets in AGP

sellmair:
^KT-43391 fixed


libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
```diff
@@ -0,0 +1,123 @@
...
+    Unfortunately, there does not exist any "ad-hoc" API like `all`.
+    Therefore we sync the directories once again after evaluation
+     */
+    target.project.whenEvaluated {
+        kotlinSourceSet.kotlin.srcDirs(*androidSourceSet.java.srcDirs.toTypedArray())
+        androidSourceSet.java.srcDirs(*kotlinSourceSet.kotlin.srcDirs.toTypedArray())
+    }
+}
+
+private fun syncKotlinAndAndroidResources(
```
gavra0, line 83: Can we do this only if MPP is applied?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
```diff
@@ -0,0 +1,123 @@
...
+        -- could be claimed by Android androidTest (instrumented test)
+
+    The Kotlin source set would win in this scenario.
+     */
+    if (disambiguationClassifier == null || !androidSourceSet.name.startsWith(disambiguationClassifier)) {
+        kotlinSourceSet.kotlin.srcDir("src/${androidSourceSet.name}/kotlin")
+    }
+
+    kotlinSourceSet.kotlin.srcDirs(*androidSourceSet.java.srcDirs.toTypedArray())
+    androidSourceSet.java.srcDirs(*kotlinSourceSet.kotlin.srcDirs.toTypedArray())
```
gavra0, line 70: I think we should avoid adding e.g `src/main/kotlin` to java source directories of AGP.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    /*
+    Make sure to include user configuration as well.
+    Unfortunately, there does not exist any "ad-hoc" API like `all`.
```
gavra0, line 74: AGP Variant API should give you a hook that you can use to read the final value of java source directories. See `AbstractAndroidProjectHandler.preprocessVariant`.

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/new-mpp-android-source-sets/lib/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+plugins {
+    id("com.android.library")
+    kotlin("multiplatform")
+}
+
+android {
```
gavra0, line 6: Can you add custom Android source sets. E.g. https://cs.android.com/android/platform/superproject/+/master:tools/build/tests/overlay3/build.gradle;l=1

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/new-mpp-android-source-sets/lib/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+kotlin {
+    android()
+    macosX64("macos")
+
+    sourceSets {
+        getByName("commonMain").dependencies {
+            implementation(kotlin("stdlib-common"))
+        }
+
+        getByName("commonMain").dependencies {
```
LouisCAD, line 24: Looks like there's a mistake there.
```suggestion
        getByName("commonTest").dependencies {
```

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
```diff
@@ -0,0 +1,123 @@
...
+    if (androidSourceSet.renderscript.srcDirs.isNotEmpty()) {
+        androidSourceSet.renderscript.srcDir(kotlinSourceSet.sourceFolderFor(project, "rs"))
+    }
+
+    if (androidSourceSet.jni.srcDirs.isNotEmpty()) {
+        androidSourceSet.jni.srcDir(kotlinSourceSet.sourceFolderFor(project, "jni"))
+    }
+
+    if (androidSourceSet.jniLibs.srcDirs.isNotEmpty()) {
+        androidSourceSet.jniLibs.srcDir(kotlinSourceSet.sourceFolderFor(project, "jniLibs"))
```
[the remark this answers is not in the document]
sellmair: Added shaders here as well! ✅ 
I noticed that androids `./gradlew sourceSets` task does not print shaders. 
Is there a reason for that, or can this be added later?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: Oh! I see: This *should* not make a difference for non-mpp project because the source sets will have the same name and therefore this paths *should* be the same. I added a check for this case. So if you would (in theory) change the name of the folders, it will only be minorly problematic for mpp projects. 
✅

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: Oh! I would be interested on why this could be harmful? 
I actually had a good intention for registering this code!

See this issue: [MPP: Instrumented Android Tests are not executed](https://youtrack.jetbrains.com/issue/KT-35016)
It seems that AGP is checking if test sources are empty before executing them on device. However, when a project does only include sources inside the kotlin source folder, then AGP does not take them into account and therefore the tests are not executed :( 

- I also experienced that test-gutters worked better when registering the source sets
- Source set colors were also correct when this is done. 

If adding the Kotln source folders here is a bad idea, then we probably should discuss a solution on AGP side?
I would be glad to help here ☺️

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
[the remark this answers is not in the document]
gavra0: It's just on oversight in AGP. We don't see high usage of that task, so no one reported the issue.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: The approach of registering the Kotlin source folder here is also done by the compose team: 

https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/buildSrc/src/main/kotlin/androidx/build/AndroidXUiPlugin.kt#195

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: Oh! I meant "ad-hoc" in the sense of "NamedDomainObjectContainer" or "FileCollection" which both are live. It seems that "SourceDirectorySet" is not live this sense (like executing code the moment a new source folder was added, or letting one SourceDirectorySet extend a live other one). 

I can't see any advantage for hooking into variants instead of this "whenEvaluated", since we would in both cases just fill in the set with missing entries. 

NB: "whenEvaluated" behaves different than "afterEvaluate". It also makes sure, that all of Androids "afterEvaluate" listeners have run.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: Actually, it seems that the same is happening for non-hmpp projects!

gavra0: This is definitely an issue on the AGP side, I've just filled https://issuetracker.google.com/issues/173770818.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
gavra0: This looks ok. Until AGP has better APIs for this, this should work perfectly fine.

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/new-mpp-android-source-sets/lib/build.gradle.kts
sellmair: Good idea!
Does this work for you? : https://github.com/JetBrains/kotlin/pull/3918/commits/2e234d08007118e2706718c64511d3c8b776fa47

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/new-mpp-android-source-sets/lib/build.gradle.kts
sellmair: Thanks a lot! Was indeed a mistake 🤷  ☺️

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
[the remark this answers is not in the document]
sellmair: Ah this is okay! Obviously I am using the output of that task for our integration test 👀 
But should be okay 👍

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: This is a very good argument ☺️ 👍 
Since there are potentially other issues fixed with registering those source folders in AndroidSourceSet, we might want to have a call on that? Maybe we could offer you a better API for looking at our source sets?

sellmair: @h0tk3y What do you think?
Registering Kotlin source sets in Android Source Set "java" directory:

+ Support for .java files in /kotlin source sets (I think some would intuitively expect this)
+ Will fix tests not running when only kotlin sources present (actually pretty severe issue IMHO)
- Potential cost of scanning more files for .java sources where we do not actually expect _many_ java files to begin with

libraries/tools/kotlin-gradle-plugin/src/functionalTest/kotlin/org/jetbrains/kotlin/gradle/SyncKotlinAndAndroidSourceSetsTest.kt
```diff
@@ -0,0 +1,206 @@
+/*
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+@file:Suppress("invisible_reference", "invisible_member", "FunctionName", "DuplicatedCode")
```
h0tk3y, line 5: (minor) Is this used for accessing `internal`s from the `main` source set while we don't have working associate compilations in the IDE? If so, then it's OK, but overall, suppressing this looks quite scary. Could you please add a comment explaining the purpose of this suppression?

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
h0tk3y: > + Support for .java files in /kotlin source sets (I think some would intuitively expect this)
In fact, I think there could be reasons why that's not what we want in MPP. For example, if we at some point support compiling JVM/Android-shared source sets to `klib`, I'm not sure we will want Java sources mixed with Kotlin in Android-shared source sets. However, this doesn't seem like a blocker for having Kotlin and Java mixed in the source directories.

> + Will fix tests not running when only kotlin sources present (actually pretty severe issue IMHO)
I agree that it's a major issue that we should fix this sooner rather than later. Until a proper fix is implemented on the AGP side, we might even want to choose the suboptimal solution.

However, note that allowing Java sources is not something that we can freely take back afterwards. Even if we decide to do that, we will have to provide migration assistance to users who will have already added their Java sources to the Kotlin source roots.

My opinion on this is: if we are sure that the fix is going to appear on the AGP side soon enough, then not registering the Kotlin sources as Java is a bit more preferable.

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    if (androidSourceSet.shaders.srcDirs.isNotEmpty()) {
+        androidSourceSet.shaders.srcDir(kotlinSourceSet.sourceFolderFor(project, "shaders"))
+    }
+}
+
+private fun KotlinSourceSet.sourceFolderFor(project: Project, type: String): File {
+    return project.file("src/${this.name}/$type")
```
h0tk3y, line 144: Maybe avoid duplication with that function? https://github.com/JetBrains/kotlin/blob/bd66e87440b9d2463437a4682128d4ce24de91ca/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/sources/KotlinSourceSetFactory.kt#L33. This is a very simple piece of logic, though, and finding the duplicate will be easy if we need to change it.

h0tk3y (approved):

libraries/tools/kotlin-gradle-plugin/src/functionalTest/kotlin/org/jetbrains/kotlin/gradle/SyncKotlinAndAndroidSourceSetsTest.kt
sellmair: Yes, non-supported associate compilations is the reason 😞 
Added a comment to all occasions of this suppress ✅

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
sellmair: Good point! See here https://github.com/JetBrains/kotlin/pull/3918/commits/b580114dd70a28b8d36297b42d86789ce579149a
✅

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/resources/testProject/new-mpp-android-source-sets/lib/build.gradle.kts
gavra0: Yup, looking good. Thanks!

libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/syncKotlinAndAndroidSourceSets.kt
gavra0: Regarding the https://issuetracker.google.com/issues/173770818, it will be fixed in next AGP alpha, and we may cherry-pick it to 4.2 branch as well, so one of the issues will be removed.

sellmair: Awesome! 🎉 
I already removed the automatic registration! 
Let's merge it without. I will try to find what other issues would have been mitigated by this change and we will find another solution afterwards ☺️

sellmair:
Merged in Space.

PR #3919, opened 2020-11-19
Added sample for String.replace()

zekehul:
Added a sample for `String.replace()` method.


libraries/stdlib/samples/test/samples/text/strings.kt
```diff
@@ -421,4 +421,13 @@ class Strings {
         assertPrints(emptyString.lastOrNull(), "null")
         assertFails { emptyString.last() }
     }
+  
+    @Sample
+    fun replace() {
+        val inputString0 = "Mississippi"
+        val inputString1 = "How can the net amount of entropy of the universe be massively decreased?"
```
qurbonzoda, line 428: Let's have a shorter sentence. It will be difficult for documentation readers to find out where exactly replace occurred in a long sentence.

qurbonzoda (changes requested):
Now that your wrote a sample, you have to go to [the declaration](https://github.com/JetBrains/kotlin/blob/ccc272c49f419fc59ccba80231b72c80a4f79ba7/libraries/stdlib/common/src/kotlin/TextH.kt#L237) of the functions and add reference to your sample.

libraries/stdlib/samples/test/samples/text/strings.kt
zekehul: I've shortened `inputString1`. If that's not short enough, I can shorten it further.

qurbonzoda:
Merged manually. Thank you for the contribution!

PR #3921, opened 2020-11-20
Use separate logic for filtering and skipping in collectAndFilterRealOverrides [KT-43487]

LepilkinaElena:


compiler/testData/codegen/box/traits/abstractClassWithFakeOverride.kt
```diff
@@ -0,0 +1,11 @@
...
+    fun foo(): String = "OK"
+}
+
+abstract class B : A {
+    abstract override fun foo(): String
+}
+
+abstract class C : B()
+
+fun box() = "OK"
```
udalov, line 11: Let's change the test case a little bit to use the classes from the hierarchy. (If the classes aren't loaded at runtime, they are not verified by the JVM and thus may contain invalid bytecode, which the test wouldn't catch)
```
interface A {
    fun foo(): String = "Fail"
}

abstract class B : A {
    abstract override fun foo(): String
}

abstract class C : B()

class D : C() {
    override fun foo(): String = "OK"
}

fun box() = D().foo()
```

udalov (approved):

alexander-gorshenev (approved):

PR #3922, opened 2020-11-20
KT-43489: KGP - Avoid storing build history mapping in a property

gavra0:
If tasks are created eagerly, it is possible for this
mapping to be initalized too early. This causes incremental
compilation to fail, as it will contain mappings only for
projects that have been evaluated thus far.

Fixes KT-43489


nav-nav:
Looks good, tests pass, merged into master

PR #3924, opened 2020-11-21
FIR: substitute stub types after builder inference

jsjeon:
The motivation is [KT-43340](https://youtrack.jetbrains.com/issue/KT-43340):
```kt
... : Map<Visibility, Int> = buildMap {
  ...
  put(Private, 0)
  ...
}
```
That `buildMap` has `@BuilderInference` in its `builderAction` parameter: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/collections/Maps.kt#L140, so it's resolved via `FirBuilderInferenceSession`, which eventually substitutes stub type variables, `K` and `V` (of `Map<K, V>`), with types resolved from lambda body (in this case, those would be `Visibility` and `Int` respectively).

The issue is, completed calls, such as `put`, can still have unsubstituted types, e.g., that `put`'s return type was still stub type variable `V`. In `FirBuilderInferenceSession`, we have a chance to `updateCalls` literally, but that just visited partially resolved calls only. According to `CoroutineInferenceSession` in FE 1.0 (which I believe is the counterpart of `FirBuilderInferenceSession`), even completed calls are revisited to deal with remaining, unsubstituted postponed type variables. We need to do the same here.

The caveat is, when `FirCallCompletionResultsWriterTransformer` revisits the completed call, that is literally completed already, and it bails out at that point. As a stopgap, this change manually rewrites the return type with the type substitutor, but we may need to recursively revisit all pieces of the completed call so as to cover all potential appearances of unsubstituted type variables.


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
```diff
@@ -203,9 +201,24 @@ class FirBuilderInferenceSession(
         val commonSystemSubstitutor = commonSystem.buildCurrentSubstitutor() as ConeSubstitutor
         val nonFixedTypesToResultSubstitutor = ConeComposedSubstitutor(commonSystemSubstitutor, nonFixedToVariablesSubstitutor)
         val completionResultsWriter = components.callCompleter.createCompletionResultsWriter(nonFixedTypesToResultSubstitutor)
+
+        for ((completedCall, candidate) in commonCalls) {
+            // TODO: Only update return type? Should we need to visit all appearances of unsubstituted postponed variables in types?
```
jsjeon, line 206: Now I found a concrete example that shows more places of remaining stub types: https://github.com/JetBrains/kotlin/blob/master/compiler/testData/codegen/box/builderInference/lackOfNullCheckOnNullableInsideBuild.kt#L25

The corresponding IR text looks like this:
```
                  CALL 'public final fun resolve (resolver: kotlin.Function1<<root>.Foo, T of <root>.Builder>): kotlin.Unit declared in <root>.Builder' type=kotlin.Unit origin=null
                    $this: GET_VAR '<this>: <root>.Builder<kotlin.String?> declared in <root>.LackOfNullCheckOnNullableInsideBuildKt.box.<anonymous>' type=<root>.Builder<IrErrorType> origin=null
                    resolver: BLOCK type=kotlin.Function1<<root>.Foo, IrErrorType> origin=LAMBDA
                      FUN LOCAL_FUNCTION_FOR_LAMBDA name:<anonymous> visibility:local modality:FINAL <> (it:<root>.Foo) returnType:IrErrorType
                        VALUE_PARAMETER name:it index:0 type:<root>.Foo
                        BLOCK_BODY
                          RETURN type=kotlin.Nothing from='local final fun <anonymous> (it: <root>.Foo): IrErrorType declared in <root>.LackOfNullCheckOnNullableInsideBuildKt.box.<anonymous>'
                            CALL 'public final fun <get-string> (): kotlin.String? declared in <root>.Foo' type=kotlin.String? origin=GET_PROPERTY
                              $this: GET_VAR 'it: <root>.Foo declared in <root>.LackOfNullCheckOnNullableInsideBuildKt.box.<anonymous>.<anonymous>' type=<root>.Foo origin=null
                      FUNCTION_REFERENCE 'local final fun <anonymous> (it: <root>.Foo): IrErrorType declared in <root>.LackOfNullCheckOnNullableInsideBuildKt.box.<anonymous>' type=kotlin.Function1<<root>.Foo, IrErrorType> origin=LAMBDA reflectionTarget=null
```
See 6 `IrErrorType` appearances, which are due to remaining stub type. By using the same substitutor, I can see that, e.g., dispatch receiver expression's type can be rescued easily.

But, still, that---dealing with type ref one-by-one---seems not scalable. I can see, e.g., `FirTypeResolver(Impl)?` and `Fir(Specific)?TypeResolveTransformer`. Does it make sense to introduce, say, `FirStubTypeSubstitutor` around this `FirBuilderInferenceSession`, which will be dedicated to visit only type references with stub types and substitute it with proper final type according to the composed substitutor here?

compiler/testData/ir/irText/types/castsInsideCoroutineInference.fir.txt
```diff
@@ -39,7 +39,7 @@ FILE fqName:<root> fileName:/castsInsideCoroutineInference.kt
               $receiver: VALUE_PARAMETER name:<this> type:<root>.FlowCollector<T of <root>.onCompletion>
               BLOCK_BODY
                 VAR name:safeCollector type:<root>.SafeCollector<IrErrorType> [val]
-                  CONSTRUCTOR_CALL 'public constructor <init> (collector: <root>.FlowCollector<T of <root>.SafeCollector>) [primary] declared in <root>.SafeCollector' type=<root>.SafeCollector<IrErrorType> origin=null
+                  CONSTRUCTOR_CALL 'public constructor <init> (collector: <root>.FlowCollector<T of <root>.SafeCollector>) [primary] declared in <root>.SafeCollector' type=<root>.SafeCollector<T of <root>.onCompletion> origin=null
```
jsjeon, line 42: Ah, there are more `IrErrorType` in this file already. 🤦

compiler/testData/ir/irText/firProblems/typeVariableAfterBuildMap.fir.txt
```diff
@@ -552,27 +552,32 @@ FILE fqName:<root> fileName:/typeVariableAfterBuildMap.kt
...
-                    value: CONST Int type=kotlin.Int value=1
-                  RETURN type=kotlin.Nothing from='local final fun <anonymous> (): kotlin.Unit declared in <root>.Visibilities.ORDERED_VISIBILITIES'
-                    CALL 'public abstract fun put (key: K of kotlin.collections.MutableMap, value: V of kotlin.collections.MutableMap): V of kotlin.collections.MutableMap? declared in kotlin.collections.MutableMap' type=IrErrorType origin=null
+                  TYPE_OP type=kotlin.Unit origin=IMPLICIT_COERCION_TO_UNIT typeOperand=kotlin.Unit
+                    CALL 'public abstract fun put (key: K of kotlin.collections.MutableMap, value: V of kotlin.collections.MutableMap): V of kotlin.collections.MutableMap? declared in kotlin.collections.MutableMap' type=kotlin.Int? origin=null
+                      $this: GET_VAR '<this>: kotlin.collections.MutableMap<<root>.Visibility, kotlin.Int> declared in <root>.Visibilities.ORDERED_VISIBILITIES.<anonymous>' type=kotlin.collections.MutableMap<IrErrorType, IrErrorType> origin=null
```
jsjeon, line 557: Also here. Only call's type is substituted, hm. We definitely need more scalable, dedicated transformer that can handle all the remaining stub types.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
jsjeon: I was convinced that we eventually need something similar, so I tried it. See my latest/5th commit, which can unmute some of builder inference bb tests.

compiler/testData/ir/irText/firProblems/typeVariableAfterBuildMap.fir.txt
jsjeon: All of stub types in this example are correctly substituted, hence no more error types at the end.

compiler/testData/ir/irText/types/castsInsideCoroutineInference.fir.txt
jsjeon: There are still some remaining `IrErrorType` in this case. Maybe the scope for stub type substitutor should be broader than what it covers for now: types in completed calls inside builder-inferred lambda. For example, `val safeCollector = ...` inside the lambda has a substituted type for the assigning expression, but its variable type is not substituted (since it's not a call). 

Another one above as type operand seems more complicated. It came from smartcast info, so any evidences inside type constraint system need some updates as well. Both seems out of scope of this PR, though.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
```diff
@@ -216,3 +226,26 @@ class ConeComposedSubstitutor(val left: ConeSubstitutor, val right: ConeSubstitu
         return left.substituteOrNull(rightSubstitution ?: type)
     }
 }
+
+class FirStubTypeTransformer(
+    override val session: FirSession,
+    private val substitutor: ConeSubstitutor
+) : FirAbstractTreeTransformer<Void?>(phase = FirResolvePhase.IMPLICIT_TYPES_BODY_RESOLVE) {
```
demiurg906, line 233: It's better to use `Nothing?` or `Unit` (I prefer the first one) in such cases

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
```diff
@@ -216,3 +226,26 @@ class ConeComposedSubstitutor(val left: ConeSubstitutor, val right: ConeSubstitu
...
+
+    override fun transformTypeRef(typeRef: FirTypeRef, data: Void?): CompositeTransformResult<FirTypeRef> {
```
demiurg906, line 235: This is not needed since you override `transformResolvedTypeRef`

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
demiurg906, line 233: `FirAbstractTreeTransformer` is used for global stage-dependent transformers. For regular transformers you should use `FirTransformer` or `FirDefaultTransformer` (second one includes default overrides like `override transformResolvedTypeRef(...) = transformTypeRef(...)`)

compiler/fir/tree/src/org/jetbrains/kotlin/fir/expressions/impl/FirResolvedArgumentList.kt
```diff
@@ -21,4 +24,9 @@ class FirResolvedArgumentList internal constructor(
             argument.accept(visitor, data)
         }
     }
+
+    override fun <D> transformArguments(transformer: FirTransformer<D>, data: D): FirArgumentList {
+        arguments.forEach { it.transformSingle(transformer, data) }
```
demiurg906, line 29: This solution doesn't satisfies contract of transformations. If transformer returns new instance of argument then it won't be written in `mapping`. Please replace it with this:
```
mapping = mapping.mapKeys { (it, _) -> it.transformSingle(transformer, data) }
```

And don't forget to make `mapping` mutable with private setter

compiler/testData/ir/irText/types/castsInsideCoroutineInference.fir.txt
demiurg906: I think it's fine to call `FirStubTypeTransformer` over whole top-level call

demiurg906 (changes requested):

demiurg906:
Generally your solution is OK, thank you. Please fix all notes and I'll accept it

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
jsjeon: Thanks, I prefer `Nothing?` too.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
jsjeon: Right, I somehow started with this and forgot to revert it, thanks.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/FirBuilderInferenceSession.kt
jsjeon: Good to learn these, thanks!

compiler/fir/tree/src/org/jetbrains/kotlin/fir/expressions/impl/FirResolvedArgumentList.kt
jsjeon: Also, good to learn these (and contract behind transformations). Edited as per suggestions.

compiler/testData/ir/irText/types/castsInsideCoroutineInference.fir.txt
jsjeon: I left a TODO at `updateCalls` to explicitly mention that we need to update more than just calls.

mglukhikh:
I think now everything is Ok, I'll merge this PR after remote run.

mglukhikh:
Merged manually 👍

PR #3927, opened 2020-11-23
FIR: add a resolution mode for property delegates

pyos:
Like function arguments, they are context-dependent, but unlike function arguments, callable references should be resolved eagerly as if they are explicit receivers.

Also, using PSI to determine the IR origin for callable references does not work if the delegate is a callable reference.


mglukhikh (approved):

mglukhikh:
This PR also fixes one of muted FIR old FE tests (I will unmute it).

mglukhikh:
Merged manually 👍

PR #3929, opened 2020-11-24
FIR checker: make unused checker path-sensitive

jsjeon:
This, as another follow-up for #3813, makes `UnusedChecker` path-sensitive. Together with #3889, all existing checkers have been migrated, and thus, the old variant---path-insensitive data collection---is deprecated as well.

### DIFF
--- a/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
+++ b/compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
@@ -31,14 +31,12 @@ fun ControlFlowGraph.traverse(
     traverse(direction, visitor, null)
 }
 
+// ---------------------- Path-sensitive data collection -----------------------
 
-// --------------------- Path-insensitive data collection ----------------------
-
-// TODO: Deprecate and make all existing checkers path-sensitive
 fun <I : ControlFlowInfo<I, K, V>, K : Any, V : Any> ControlFlowGraph.collectDataForNode(
     direction: TraverseDirection,
     initialInfo: I,
-    visitor: ControlFlowGraphVisitor<I, Collection<I>>
+    visitor: ControlFlowGraphVisitor<I, Collection<Pair<EdgeLabel, I>>>
 ): Map<CFGNode<*>, I> {
     val nodeMap = LinkedHashMap<CFGNode<*>, I>()
     val startNode = getEnterNode(direction)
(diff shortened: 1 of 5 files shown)


mglukhikh:
Hi, I remember about this PR but have to hold on it for a few more days due to internal reasons. Meanwhile, I've just found a few problems in unused checker, so, if you are interested...

- https://youtrack.jetbrains.com/issue/KT-43687
- https://youtrack.jetbrains.com/issue/KT-43688

mglukhikh:
And one more about FIR CFG: https://youtrack.jetbrains.com/issue/KT-43691

jsjeon:
Thanks for the pointers, but I'd like to avoid making conflicts to my own PR(s). May I assume these mechanical refactoring changes in this branch look okay/good and would be merged as-is so that I can try to resolve those issues on top of this branch?

jsjeon:
> * https://youtrack.jetbrains.com/issue/KT-43687

A draft fix on top of this branch: https://github.com/JetBrains/kotlin/pull/3948

jsjeon:
> * https://youtrack.jetbrains.com/issue/KT-43688

A draft fix on top of this branch: #3949 

Perhaps conflict with #3948 above; either one can be checked in first, then I can deal with merge conflict.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/UnusedChecker.kt
```diff
@@ -35,42 +38,53 @@ object UnusedChecker : FirControlFlowChecker() {
...
         override fun visitVariableAssignmentNode(node: VariableAssignmentNode) {
             val variableSymbol = (node.fir.calleeReference as? FirResolvedNamedReference)?.resolvedSymbol ?: return
-            val data = data[node]?.get(variableSymbol) ?: return
-            if (data == VariableStatus.ONLY_WRITTEN_NEVER_READ) {
-                // todo: report case like "a += 1" where `a` `doesn't writes` different way (special for Idea)
-                val source = node.fir.lValue.source
-                reporter.report(source, FirErrors.ASSIGNED_VALUE_IS_NEVER_READ)
+            val dataPerNode = data[node] ?: return
+            for (label in dataPerNode.keys) {
+                val data = dataPerNode[label]!![variableSymbol] ?: continue
```
mglukhikh, line 50: May be it's better to enumerate entries `for ((label, pathInfo) in dataPerNode)` and remove this `!!`?

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/UnusedChecker.kt
```diff
@@ -35,42 +38,53 @@ object UnusedChecker : FirControlFlowChecker() {
...
             }
         }
 
         override fun visitVariableDeclarationNode(node: VariableDeclarationNode) {
             val variableSymbol = node.fir.symbol
             if (variableSymbol.isLoopIterator) return
-            val data = data[node]?.get(variableSymbol) ?: return
+            val dataPerNode = data[node] ?: return
+            for (label in dataPerNode.keys) {
+                val data = dataPerNode[label]!![variableSymbol] ?: continue
```
mglukhikh, line 66: The same here

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
```diff
@@ -31,14 +31,12 @@ fun ControlFlowGraph.traverse(
...
+// ---------------------- Path-sensitive data collection -----------------------
 
-// --------------------- Path-insensitive data collection ----------------------
-
-// TODO: Deprecate and make all existing checkers path-sensitive
 fun <I : ControlFlowInfo<I, K, V>, K : Any, V : Any> ControlFlowGraph.collectDataForNode(
     direction: TraverseDirection,
     initialInfo: I,
-    visitor: ControlFlowGraphVisitor<I, Collection<I>>
+    visitor: ControlFlowGraphVisitor<I, Collection<Pair<EdgeLabel, I>>>
```
mglukhikh, line 39: I'm not sure, but may be some kind of map is better than `Collection<Pair>` here?

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/UnusedChecker.kt
```diff
@@ -197,7 +241,27 @@ object UnusedChecker : FirControlFlowChecker() {
...
+        }
+
+        private fun update(
+            info: PathAwareVariableStatusInfo,
+            symbol: FirPropertySymbol,
+            updater: (VariableStatus?) -> VariableStatus?,
+        ): PathAwareVariableStatusInfo {
+            var resultMap = persistentMapOf<EdgeLabel, VariableStatusInfo>()
+            var changed = false
+            for (label in info.keys) {
```
mglukhikh, line 254: Here also entry enumeration should be better

mglukhikh (changes requested):

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/checkers/extended/UnusedChecker.kt
jsjeon: Ah, right. I don't know why I've stuck to use `label`. There could be many different per-label iterations. Thanks to these comments, I revisited all my previous iterations of path-aware infos and remove `!!`.

compiler/fir/checkers/src/org/jetbrains/kotlin/fir/analysis/cfa/CfgTraverser.kt
jsjeon: Hm, not really. Say we use `Map<EdgeLabel, I>`. If a certain node has multiple incoming info with the same label (and usually will do so, because the common path is labeled with `NormalPath`), now the value part, `I`, needs to be `merge`d, but that's not what the traverser should do; instead, here we literally collect them as a collection of pair, and let the `visitor` do its job: It could apply labels; merge infos; etc. with its own orders/logics.

mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3930, opened 2020-11-25
FIR: import parents of companion objects first

pyos:
Otherwise, information about members moved from companion objects to the parent class (e.g. on JVM, companion object fields -> static fields in parent class) will be incorrect.

### DIFF
--- a/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/codegen/FirCompileKotlinAgainstKotlinTestGenerated.java
+++ b/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/codegen/FirCompileKotlinAgainstKotlinTestGenerated.java
@@ -148,6 +148,11 @@ public class FirCompileKotlinAgainstKotlinTestGenerated extends AbstractFirCompi
         runTest("compiler/testData/compileKotlinAgainstKotlin/expectClassActualTypeAlias.kt");
     }
 
+    @TestMetadata("importCompanion.kt")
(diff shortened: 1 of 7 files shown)


compiler/fir/fir-deserialization/src/org/jetbrains/kotlin/fir/deserialization/ClassDeserialization.kt
```diff
@@ -88,9 +88,10 @@ fun deserializeClassToSymbol(
             containerSource
         )
     if (status.isCompanion) {
-        parentContext?.let {
-            context.annotationDeserializer.inheritAnnotationInfo(it.annotationDeserializer)
+        require(parentContext != null) {
+            "companion object $classId deserialized without parent context - annotations for moved members will be incorrect"
```
mglukhikh, line 92: Sorry but one of tests from `FirTypeEnhancementTestGenerated` crashes with this exception. Please look at this test (`testIntrinsicCompanionObject`).

mglukhikh (changes requested):

compiler/fir/fir-deserialization/src/org/jetbrains/kotlin/fir/deserialization/ClassDeserialization.kt
pyos: Hm, I don't really know if this is correct for built-ins, so I've removed the exception

mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3931, opened 2020-11-25
[JS IR] Make WITH_RUNTIME imply KJS_WITH_FULL_RUNTIME

skuzmich:
There is a lot of intersection between these


bashor (approved):

udalov:
Do you plan to also remove KJS_WITH_FULL_RUNTIME from test data where there's already WITH_RUNTIME?

skuzmich:
@udalov no, but we plan to remove KJS_WITH_FULL_RUNTIME completely once we fix perf issues.

PR #3932, opened 2020-11-25
[JVM_IR] Use direct field access to backing fields on current class.

madsager:
The current backend uses direct field access to the backing field
instead of calling the companion object accessor, which calls
an accessibility bridge, which then gets the field for code such as:

```
class A {
  companion object {
    val s: String = "OK"
  }

  // f uses direct access to the A.s backing field.
  fun f() = s
}
```

This change does the same for the IR backend.


max-kammerer:
I've made brief search and didn't find any test with inline, i.e.:
```
class A {
  companion object {
    val s: String = "OK"
  }

  // f uses direct access to the A.s backing field.
  inline fun f() = s
}
```
in this case `s` should be accessed via accessor.
Maybe it's also  worth to add tests for lateinit, delegated vars

madsager:
This actually works for inline functions because the access ends up going through the accessibility bridge which is public and therefore works. So instead of calling `Companion.getS -> A.access$getS$cp` we directly call `A.access$getS$cp` which looks fine to me? I added a test both as a black box and bytecode text test to make this explicit.

lateinit properties are covered here:

codegen/bytecodeText/properties/lateinit/companionObject.kt

I added a blackbox test for delegated properties on companion objects to make sure that is covered.

compiler/testData/codegen/bytecodeText/companion/inlineFunctionCompanionPropertyAccess.kt
```diff
@@ -0,0 +1,45 @@
...
+// One call to the accessibility bridge `access$setV$cp` from Companion.setV.
+// 1 INVOKESTATIC A.access\$setV\$cp
+
+// One call to the accessibility bridge `access$getS$cp` from Companion.getS.
+// 1 INVOKESTATIC A.access\$getS\$cp
+
+// JVM_IR_TEMPLATES:
+
+// No calls to the getter/setter on the companion object.
+// 0 INVOKEVIRTUAL A\$Companion.getS
```
max-kammerer, line 36: >So instead of calling Companion.getS -> A.access$getS$cp we directly call A.access$getS$cp which looks fine to me?

I'm expect that next change wouldn't be binary compatible cause of synthetic accessor usage:
```
class A {
    companion object {
        val s: String
            get() = "Ok"
        var v : String 
            get() = "NOT OK"
            set(value) {}
    }

    inline fun f(): String = s

    inline fun g() {
        v = "OK"
    }
}
```

madsager: These are not `DEFAULT_PROPERTY_ACCESSORS` so they are filtered out and `A$Companion.setV` and `A$Companion.getS` will be used from `g` and `f`, so this should generate the same code as before. Let me add a bytecode text test checking that as well. :)

max-kammerer: I meant that if original code 
```
class A {
    companion object {
        val s = "OK"
        var v = "NOT OK"
    }

    inline fun f(): String = s

    inline fun g() {
        v = "OK"
    }
}
```
would be changed to example above it would break binary compatibility

madsager: Oh my, I'm sorry @max-kammerer I was moving way too fast here. I should have understood what you meant. Thank you for catching that! :-) 

I filtered out inline functions as well, so that we only use direct access for code that actually stays on the class of the property backing field.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmPropertiesLowering.kt
```diff
@@ -67,6 +67,18 @@ class JvmPropertiesLowering(private val backendContext: JvmBackendContext) : IrE
         return expression
     }
 
+    private fun isDefaultAccessorForCompanionPropertyBackingFieldOnCurrentClass(
+        property: IrProperty,
+        function: IrSimpleFunction
+    ): Boolean {
+        if (function.origin != IrDeclarationOrigin.DEFAULT_PROPERTY_ACCESSOR) return false
+        if (property.isLateinit) return false
+        if ((currentFunction?.irElement as? IrFunction)?.isInline == true) return false
```
max-kammerer, line 76: I'm not sure that lambda expression is already extracted to separate class before this lowering (or it could be changed in future), it's also good to have next cases in tests:
```
class A {
   ...
    inline fun g(crossinline s: () -> Unit) {
        {
            s()
            v = "OK"            
        } ()
    }

    inline fun g2(crossinline s: () -> Unit) {
        object {
            fun run() {
                s()
                v = "OK"
            }
        }.run ()
    }
}```

Comment (madsager) [reply compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/JvmPropertiesLowering.kt]: This did indeed expose another case that is not working. The code snippet above is currently not problematic. Both the object and the lambda are represented as classes and therefore the `currentClass` check will filter them out.

However, accesses from a LOCAL_FUNCTION_FOR_LAMBDA in an inline function are not counted as being in an inline function and therefore end up using the accessibility bridge instead of the getting on the Companion object. That is not good. I added a test and changed the code to filter out everything that is nested inside of an inline function and not just the functions itself. I incorrectly assumed that there would be no local functions inside an inline function (as local functions are rejected by the frontend), however, lambdas at this point are represented as nested functions.

This still leads to differences with the current backend. For the following example:

```
class A {
  companion object {
    val x = 32
  }

  inline fun f(g: () -> Unit) {
    g()
  }

  fun h() {
    f {
      x
    }
  }
}
```

The access to x in h will use direct property access. That happens on the current backend as well.

The difference is that with the current JVM_IR code, the same is true if the lambda is marked cross inline, in that case the accessibility bridge will be used from the $$inlined class. That should be fine for binary compatibility, I think, as this code belongs on this class and therefore if we change the companion object this code will be recompiled as well. But, what do you think @max-kammerer?

max-kammerer: > The difference is that with the current JVM_IR code, the same is true if the lambda is marked cross inline, in that case the accessibility bridge will be used from h. That should be fine for binary compatibility, I think, as this code belongs on this class and therefore if we change the companion object this code will be recompiled as well. But, what do you think @max-kammerer?

Yes, it's safe.
Thank you also for adding similar cases with `inline h` )

max-kammerer:
@madsager Many thanks!

PR #3934, opened 2020-11-26
FIR DFA: refine type info around null equality checks

jsjeon:
The motivation is [KT-43569](https://youtrack.jetbrains.com/issue/KT-43569) where variable `result` is updated multiple times in order to be non-null:
```kt
class Flaf(val javaName: String) {
    private val INSTANCES = mutableMapOf<String, Flaf>()
    fun forJavaName(javaName: String): Flaf {
      var result: Flaf? = INSTANCES[javaName]
      if (result == null) {
        result = INSTANCES["${javaName}_alternative"]
        if (result == null) {
          result = Flaf(javaName)
        }
        INSTANCES[javaName] = result
      }
      return result
    }
}
```
However, at the last index operator point, the smartcast type info (or info from FIR DFA) is still nullable, hence the resolution failed, which expected the argument for value `V` is not nullable.

The key point is:
```kt
if (result == null) {
  result = ... // not null anymore
} // synthetic else // result is not null according to the condition
// therefore, at the join of when conditions, result is not null
```

When the synthetic `else` of the inner `when` expression was visited, type info added at that point affected all the above expressions, including the first variable assignment: `result = INSTANCE[...]`. So, `result`, which was of `Flaf?` type, became `it(Any, Flaf?)` all the sudden. The effects of branches inside `when` expression as well as expressions before/after `when` expression should be isolated. The 2nd commit does so, and now the type information for the first variable assignment remains as-is.

But, inspired by that, when a variable is of a nullable class-like type, followed by a null check, we can learn that, for that branch and `else` branch, we can nail down that variable's nullability. For now, after `null` check, implications bound to the variable are `Any`. But, we can use the r-value's (non-null) type. The 3rd commit does so, and is able to deal with collection set, which expected non-null value type, properly.

### DIFF
--- a/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/fir/Fir2IrTextTestGenerated.java
+++ b/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/fir/Fir2IrTextTestGenerated.java
@@ -1807,6 +1807,11 @@ public class Fir2IrTextTestGenerated extends AbstractFir2IrTextTest {
             runTest("compiler/testData/ir/irText/firProblems/FirBuilder.kt");
         }
 
+        @TestMetadata("inapplicableCollectionSet.kt")
+        public void testInapplicableCollectionSet() throws Exception {
+            runTest("compiler/testData/ir/irText/firProblems/inapplicableCollectionSet.kt");
+        }
+
         @TestMetadata("InnerClassInAnonymous.kt")
         public void testInnerClassInAnonymous() throws Exception {
             runTest("compiler/testData/ir/irText/firProblems/InnerClassInAnonymous.kt");
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/LogicSystem.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/dfa/LogicSystem.kt
@@ -122,15 +122,21 @@ abstract class LogicSystem<FLOW : Flow>(protected val context: ConeInferenceCont
         return result
     }
 
-    protected fun or(statements: Collection<TypeStatement>): MutableTypeStatement {
+    private inline fun manipulateTypeStatements(
+        statements: Collection<TypeStatement>,
+        op: (Collection<Set<ConeKotlinType>>) -> MutableSet<ConeKotlinType>
+    ): MutableTypeStatement {
         require(statements.isNotEmpty())
         statements.singleOrNull()?.let { return it as MutableTypeStatement }
         val variable = statements.first().variable
         assert(statements.all { it.variable == variable })
-        val exactType = orForTypes(statements.map { it.exactType })
-        val exactNotType = orForTypes(statements.map { it.exactNotType })
+        val exactType = op.invoke(statements.map { it.exactType })
+        val exactNotType = op.invoke(statements.map { it.exactNotType })
         return MutableTypeStatement(variable, exactType, exactNotType)
     }
+
+    protected fun or(statements: Collection<TypeStatement>): MutableTypeStatement =
+        manipulateTypeStatements(statements, ::orForTypes)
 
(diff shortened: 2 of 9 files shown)


jsjeon:
Although these changes can resolve the original issue, I have to admit that each commit (except for the reproduction) has some regressions. (Some other changes look good, though.) So, I'm open to further discussions, and feel free to point out where things can go wrong.

demiurg906:
Since this PR introduce new regressions (spec tests) I can't accept it. But from my perspective you are trying to fix it from wrong side. Let's look at your original example:
```
interface A {
    fun foo()
}

fun getA(): A = null!!

fun getNullableA(): A? = null

fun test() {
    var a: A? = getNullableA()
    if (a == null) {
        a = getNullableA()
        if (a == null) {
            a = getA()
        } // <------- here
    }
    a.foo()
}
```

At the end of inner `if` we have two flows (from then branch and synthetic else):
1. `a hasType A`
2. `a hasTypes A?, Any`

And for those flows `or` in logic system returns empty set but it should be `A`. So main problem somewhere around this place

jsjeon:
`or` in logic system computes the common super types, so `or(A, A?, Any)` returns `Any?`. Having `Any` makes it reach `Any` anyway, and having `A?` makes it nullable, so I don't understand how it should be `A`.

But, I have to admit that those regressions in spec tests originated from removing `Any` implications. I still don't understand why all the sudden the variable with null eq check could be either of `Any` type or shouldn't be of `Any` type if we know the given operand has a more specific type. Anyway, restoring (or reverting) changes regarding `Any` implications removed regressions.

Also, it looks like `when` expression/branch already have its own way to isolate effects between flows. So I dropped the 2nd commit. In sum, all regressions are gone, while the original example is handled properly.

jsjeon:
> `or` in logic system computes the common super types, so `or(A, A?, Any)` returns `Any?`. Having `Any` makes it reach `Any` anyway, and having `A?` makes it nullable, so I don't understand how it should be `A`.

Oh, perhaps, you meant `it(A, A?, Any)` should be `A`? As long as all those possibilities are left until smartcast addition, where the intersection of possible types (plus the original type) is used, that's right. But, that means we need to `union` at the end of inner `if`, which doesn't sound right.

Anyway, I retraced how my change made the original problem work, and found that, at the end of inner `if`, two flows look like:
1. `a hasType A` // then branch
2. `a hasType A?, Any, A` // synthetic else, the last `A` is added

then, `or` of type info for `a` from both flows ended up with `a hasType A`. :)

demiurg906:
> Oh, perhaps, you meant it(A, A?, Any) should be A? 

No. When we perform `or` operation on flows we want to get next result: `or(x has type (A, B), x has type (C, D)) = cst(it(A, B), it(C, D)). Current implementation is doing slightly different thing which was made for optimizations but your case shows that it is wrong. 

If apply this rule to flows from example then we get `A` for them: `or(a has A, a has (A?, Any)) = cst(A, it(A?, Any)) = cst(A, A) = A`

jsjeon:
> When we perform `or` operation on flows we want to get next result: `or(x has type (A, B), x has type (C, D)) = cst(it(A, B), it(C, D)). Current implementation is doing slightly different thing which was made for optimizations but your case shows that it is wrong.
> 
> If apply this rule to flows from example then we get `A` for them: `or(a has A, a has (A?, Any)) = cst(A, it(A?, Any)) = cst(A, A) = A`

Thanks for the clarification with concrete examples. So, the issue we found demonstrates a corner case where there is no common type between given sets, resulting in different types to be equal to original given sets, which in turn led to flattening, which isn't correct.

I guess the intersection type is also based on super type computation, so the optimization delegates it to the outer super type computation (`cst(A, it(A?, Any)) => cst(A, A?, Any)`), but due to the nullability adjustment, I think we can't do that. Reimplementing `or` operation for types as you described resolved the issue.

jsjeon:
Friendly ping.

demiurg906:
Yeah, sorry

I just send your branch to `rr` before push

demiurg906:
Merged manually

PR #3936, opened 2020-11-27
Solves issue KT-43615.

stephanenicolas:
It seems to work well locally.


max-kammerer:
Is it possible to add any tests?

max-kammerer:
cc @gavra0

gavra0:
I'm able to reproduce now, and I'm looking into it. Most likely the root cause is something else.

max-kammerer:
@gavra0 Thank you for investigation!

PR #3937, opened 2020-11-27
FIR: serialize HAS_CONSTANT at least for const properties

pyos:
Non-const properties may need them too with if the 1.4 feature NoConstantValueAttributeForNonConstVals is disabled.


mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3940, opened 2020-11-27
FIR2IR: fix the way annotations are moved to fields

pyos:
1. When an annotation has multiple targets, the priority goes like this: constructor parameter (if applicable) -> property -> backing field.
 2. The argument to `kotlin.annotation.Target` is a vararg, so that should be handled as well.
 3. `AnnotationTarget.VALUE_PARAMETER` allows receivers, constructor parameters, and setter parameters, while `AnnotationTarget.FIELD` allows both backing fields and delegates.

Known issue: java.lang.annotation.Target is not remapped to the Kotlin equivalent, so things are still broken for pure Java annotations.


mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3941, opened 2020-11-29
FIR: refine type of when expression when possible

jsjeon:
The motivation is [KT-43616](https://youtrack.jetbrains.com/issue/KT-43616) where a local property is initialized with `when` expression whose type was marked nullable, whereas all branches' types were not nullable. We can safely refine the return type of `when` expression if 1) all branches have the same type; and 2) that type is a sub type of the original type of `when` expression.

### DIFF
--- a/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/fir/Fir2IrTextTestGenerated.java
+++ b/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/fir/Fir2IrTextTestGenerated.java
@@ -1817,6 +1817,11 @@ public class Fir2IrTextTestGenerated extends AbstractFir2IrTextTest {
             runTest("compiler/testData/ir/irText/firProblems/MultiList.kt");
         }
 
+        @TestMetadata("NamedCompilerPhase.kt")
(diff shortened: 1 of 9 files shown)


mglukhikh:
Well, I saw your solution, thanks. Yet I'm not sure it should be fixed this way. We will discuss it internally.

jsjeon:
> Well, I saw your solution, thanks. Yet I'm not sure it should be fixed this way. We will discuss it internally.

Sure, no problem. Just let me know if you came up with other ways to fix and if it's doable for me.

I updated the commit a little bit to describe some corner cases to ponder as well as to not directly use the intersection type since it's not necessary at all.

jsjeon:
As per @juan-chen 's suggestion, at the 3rd commit, I tweaked the logic a bit to update the type of `when` expression with the common super types of branches. According to the use of common super types, e.g., lambda's return type, which is computed as the common super types of all `return`'ed types, I think it's still conservative, yet intuitive, and still resolve the original issue. :)

jsjeon:
I saw you fixed different way: https://github.com/JetBrains/kotlin/commit/2429f429c5e2012499b3dcd33bf07139144d725e

PR #3939, opened 2020-11-27
[IR] Make isHidden and isAssignable explicit on IrValueParameter.

madsager:
There were a couple of places where they were confused and
isAssignable was passed as a positional parameter in the position
of isHidden.

PR #3943, opened 2020-11-30
[IR] Fix default arguments in local functions for Compose default argument handling

madsager:
Compose does its own default argument handling. This exposes an issue where LocalDeclarationsLowering does not remap symbols in IrSetValue which is used to set the parameter which has a default value.

Additionally, add tests that perform default arguments rewriting as a generation extension in the style of Compose. The code is almost exactly what Compose currently does.


madsager:
Friendly ping. This fixes the default argument handling for compose. Therefore, it would be great to get this in and hopefully cherry-picked to 1.4.30.

compiler/testData/codegen/composeLikeBytecodeText/default.kt
```diff
@@ -0,0 +1,8 @@
+fun foo(s: String = "O") = s
+
+fun box() = foo() + foo("K")
+
+// For Compose special default arugment handling, we still do not want
+// the default argument mask in the local variable table.
+
+// 0 LOCALVARIABLE \$default
```
udalov, line 8: Maybe just `0 \$default` to check that there are no `$default` methods at all? Local variable names aren't a part of ABI, so they can possibly change in the future and the test will not check what it's checking now anymore

udalov:
I think it'd be great either to (preferably) somehow simplify `ComposeLikeGenerationExtension` to remove non-essential parts, or to add more tests that check everything there. Otherwise we could easily break this code suddenly by any major change in the IR API, and since the tests won't show it, this would defeat the purpose of having this code in our project.

Also, please fix a new compilation error which appeared after merging #3939.

compiler/testData/codegen/composeLikeBytecodeText/default.kt
madsager: Done.

madsager:
I fixed the compilation error.

There are two things that I would like this to test; both of which the compiler doesn't generate as is and therefore both of which needs extensions:

1. inlining of custom default argument handling; I cannot see that I can do that in a simpler way that this. I have tried to cut it to the bone and only do what I have to in order to actually make that work.

2. the missing remapping in local declarations lowering of an IrSetValue for a parameter. For that one, I could make a smaller extension that only remaps a given local variable to an argument. However, it would still be an extension, and I would then have to have two extensions as I really would like to have a regression test for 1.

To me, the benefit of having this test extension do something that is much like what Compose is doing is that if we break this with large changes to IR structures, we will know how to fix it in Compose.

I'm definitely open to changing the tests if we can find an easier way to test this. I'm afraid I don't know how to do something that is simpler than this that tests the changes we have made to make sure that Compose default argument handling works.

udalov:
Merged to [master](https://github.com/JetBrains/kotlin/commit/fadedc84dbd3ac9aba1d7224732f53105692ec47) and [1.4.30](https://github.com/JetBrains/kotlin/commit/583c9d783760045a4fdd1be6d203124efbcae9c0), sorry for delays. Thank you!

PR #3944, opened 2020-11-30
[FIR] fix property accessors from fakeoverriding delegated properties.

juan-chen:
When a property in a subclass is created from fakeoverriding a delegated property in a super class, the delegated property does not have FIR accessors, and DeleagatedMemberGenerator creates IR accessors. This causes IrProperty.discardAccessorsAccordingToBaseVisibility to discard the subclass property accessors.

This PR adds a check in IrProperty.discardAccessorsAccordingToBaseVisibility for the delegated property case, so that subclass property accessors can be kept.

^Fix KT-43342.


mglukhikh:
Yesterday I fixed the situation this way: https://github.com/JetBrains/kotlin/commit/78c35100add90072b0e4389275ae2831446336cb, it looks a bit more consistent.

PR #3945, opened 2020-12-01
FIR: remap Java meta-annotations to Kotlin equivalents

pyos:
This is a direct port of JavaAnnotationMapper from the old frontend.

### DIFF
--- a/compiler/fir/java/src/org/jetbrains/kotlin/fir/java/JavaUtils.kt
+++ b/compiler/fir/java/src/org/jetbrains/kotlin/fir/java/JavaUtils.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.kotlin.fir.java
 
+import org.jetbrains.kotlin.builtins.StandardNames
 import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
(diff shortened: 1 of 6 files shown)


compiler/fir/java/src/org/jetbrains/kotlin/fir/java/JavaUtils.kt
```diff
@@ -421,52 +428,120 @@ private fun FirRegularClass.createRawArguments(
...
 
 internal fun JavaAnnotation.toFirAnnotationCall(
     session: FirSession, javaTypeParameterStack: JavaTypeParameterStack
 ): FirAnnotationCall {
     return buildAnnotationCall {
-        val mapping = buildArgumentMapping(session, javaTypeParameterStack, classId, arguments)
-        argumentList = if (mapping != null) {
-            buildResolvedArgumentList(mapping)
+        val lookupTag = when (classId) {
+            JAVA_TARGET_CLASS_ID -> ClassId.topLevel(StandardNames.FqNames.target)
```
mglukhikh, line 518: You can use `JavaAnnotationMapper.javaToKotlinNameMap` access here, instead of writing this `when`

mglukhikh (changes requested):
I will run some tests today later and may be add a bit more information.

compiler/fir/java/src/org/jetbrains/kotlin/fir/java/JavaUtils.kt
pyos: JavaAnnotationMapper is in the core/descriptors.jvm module, so it's not currently listed as a dependency, and probably shouldn't be.

mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3946, opened 2020-12-01
[JVM] Do not put the name of default lambda parameter in LVT.

madsager:
If we do, the local variable table will not make sense. As as
example:

```
inline fun foo(getString: () -> String = { "OK" }) {
  println(getString())
}

inline fun bar() {
}

fun main() {
    bar()
    foo()
}
```

leads to the following bytecode:

```
  public static final void main();
    descriptor: ()V
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=2, locals=4, args_size=0
         0: iconst_0
         1: istore_0
         2: nop
         3: nop
         4: iconst_0
         5: istore_1
         6: nop
         7: ldc           #53                 // String OK
         9: astore_2
        10: iconst_0
        11: istore_3
        12: getstatic     #30                 // Field java/lang/System.out:Ljava/io/PrintStream;
        15: aload_2
        16: invokevirtual #36                 // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
        19: nop
        20: return
      LineNumberTable:
        line 9: 0
        line 13: 2
        line 10: 3
        line 14: 4
        line 15: 6
        line 16: 7
        line 17: 19
        line 11: 20
      LocalVariableTable:
            2       1     0 $i$f$bar   I
            6      14     1 $i$f$foo   I
            4      16     0 getString$iv   Lkotlin/jvm/functions/Function0;
```

The `getString$iv` local should not be there. It has been inlined away.
Leaving it in the local variable table leads to inconsistent locals
info. Local 0 contains an int but we declare a local of type
Function0.

Strengthen the checking that is done with D8 so that we catch issues
like these. Also, mute a test that generates invalid inner class attributes.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrTypeMapper.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrTypeMapper.kt
@@ -53,6 +53,10 @@ class IrTypeMapper(private val context: JvmBackendContext) : KotlinTypeMapperBas
         }
 
     private fun computeClassInternalName(irClass: IrClass): StringBuilder {
+        context.getLocalClassType(irClass)?.internalName?.let {
+            return StringBuilder(it)
+        }
+
         val shortName = SpecialNames.safeIdentifier(irClass.name).identifier
 
         when (val parent = irClass.parent) {
@@ -71,11 +75,6 @@ class IrTypeMapper(private val context: JvmBackendContext) : KotlinTypeMapperBas
                     return computeClassInternalName(parent.parentAsClass.parentAsClass)
                         .append("$").append(parent.name.asString())
                 }
-        }
-
-        val localClassType = context.getLocalClassType(irClass)
-        if (localClassType != null) {
-            return StringBuilder(localClassType.internalName)
         }
 
         error(
--- a/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/defaultMethodUtil.kt
+++ b/compiler/backend/src/org/jetbrains/kotlin/codegen/inline/defaultMethodUtil.kt
@@ -130,7 +130,9 @@ fun <T, R : DefaultLambda> expandMaskConditionsAndUpdateVariableNodes(
         node.instructions.insert(position, newInsn)
     }
 
-    node.localVariables.removeIf { it.start in toDelete && it.end in toDelete }
+    node.localVariables.removeIf {
+        (it.start in toDelete && it.end in toDelete) || defaultLambdas.contains(it.index)
+    }
 
     node.remove(toDelete)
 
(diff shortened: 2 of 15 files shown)


madsager:
@pyos @max-kammerer is this the right way to fix this? I'm not sure I understand the details well enough here, so I could use your help double checking that this fix makes sense. Thanks!

madsager:
Fixed the incorrect names in inner class attributes as well in the second commit.

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/codegen/IrTypeMapper.kt
```diff
@@ -65,10 +65,10 @@ class IrTypeMapper(private val context: JvmBackendContext) : KotlinTypeMapperBas
                     append(shortName)
                 }
             is IrClass ->
-                return computeClassInternalName(parent).append("$").append(shortName)
+                return StringBuilder(classInternalName(parent)).append("$").append(shortName)
```
udalov, line 68: I suppose the problem started happening after f2b0d057b9ec17686deb5d25990624b378ef3e3a where I've tried to optimize this code to minimize string concatenations and map lookups for deeply nested classes. Is there any way we could keep the optimization and also fix the class names, maybe by storing the class names in the `context.getLocalClassType` map? It'd be great to have names for all local classes in that map, so that only one lookup would be necessary in the type mapper.

madsager: This type is actually in the localClassType map. The problem is that we do not look it up when we do a recursive call here. Let me try to restructure this so that we do the localClassType map lookup on recursive calls but thread around a StringBuilder so we only build the strings once. I'll benchmark and see how that performs compared to what is there now for some deep hierarchies.

madsager: Thanks for catching that @udalov. I verified that for deeply nested classes my initial attempt added a 10% overhead on the full compilation time. That is not good and is indeed caused by the string building changing.

I updated to always lookup types in the local class type mapping first in computeClassInternalName. That fixes the issue and has only on the order of ~1 percent compile time penalty.

max-kammerer:
Looks like there is smth wrong in `expandMaskConditionsAndUpdateVariableNodes` cause it already tries to delete entries from LVT

max-kammerer:
Looks like it tries to delete lv entries inside removed block, but not entries for lambda itself. It best to move this logic in `expandMaskConditionsAndUpdateVariableNodes`

madsager:
I moved it there. But what I did just checks for the index in the locals table. I'm concerned if there could be other locals in the MethodNode locals table that use the same index as a default lambda of the inline function. We probably need a more conservative condition here...

max-kammerer:
> I'm concerned if there could be other locals in the MethodNode locals table that use the same index as a default lambda of the inline function

No, in general compiler never writes  in method parameters slots except case of $default function where default value is written into parameter slot

madsager:
> > I'm concerned if there could be other locals in the MethodNode locals table that use the same index as a default lambda of the inline function
> 
> No, in general compiler never writes in method parameters slots except case of $default function where default value is written into parameter slot

Makes sense, thanks! :)

max-kammerer:
Many thanks!

madsager:
This one is not critical, but it would be nice to get this merged to 1.4.30 if possible to fix a bunch of build info messages for Android users.

udalov:
I've pushed this to [1.4.30](1ca28f941eb57e187e36e14d072e1ef38a68d9b2)

PR #3948, opened 2020-12-01
FIR unused checker: visit qualified accesses in annotations

jsjeon:
The motivation is [KT-43687](https://youtrack.jetbrains.com/issue/KT-43687) where a variable used as annotation arguments is determined as unused variable. It's simply because unused checker has not considered annotations.

PR #3949, opened 2020-12-02
FIR unused checker: handle invoke properly

jsjeon:
The motivation is [KT-43688](https://youtrack.jetbrains.com/issue/KT-43688) where local properties that are used via `invoke` are determined as unused variable. It's simply because unused checker has not handled `invoke` properly.


jsjeon:
No conflict after https://github.com/JetBrains/kotlin/pull/3948 being merged. @mglukhikh PTAL

mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3950, opened 2020-12-02
Check analysis error before checking compiler args

tdn120:
ExperimentalUsageChecker.checkCompilerArguments() will crash if there was an error running analyze().
This checks for the error first so it can be properly reported.

### DIFF
--- a/compiler/cli/src/org/jetbrains/kotlin/cli/common/messages/AnalyzerWithCompilerReport.kt
+++ b/compiler/cli/src/org/jetbrains/kotlin/cli/common/messages/AnalyzerWithCompilerReport.kt
@@ -113,11 +113,13 @@ class AnalyzerWithCompilerReport(
 
     override fun analyzeAndReport(files: Collection<KtFile>, analyze: () -> AnalysisResult) {
         analysisResult = analyze()
-        ExperimentalUsageChecker.checkCompilerArguments(
-            analysisResult.moduleDescriptor, languageVersionSettings,
-            reportError = { message -> messageCollector.report(ERROR, message) },
-            reportWarning = { message -> messageCollector.report(WARNING, message) }
-        )
+        if (!analysisResult.isError()) {
+            ExperimentalUsageChecker.checkCompilerArguments(
+                analysisResult.moduleDescriptor, languageVersionSettings,
+                reportError = { message -> messageCollector.report(ERROR, message) },
+                reportWarning = { message -> messageCollector.report(WARNING, message) }
+            )
+        }
         reportSyntaxErrors(files)
         reportDiagnostics(analysisResult.bindingContext.diagnostics, messageCollector, renderDiagnosticName)
         reportIncompleteHierarchies()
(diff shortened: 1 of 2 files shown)


udalov:
Hi! As discussed in the email thread, let's remove the logic that outputs error because it doesn't work correctly if there was a _compilation_ error (not _internal_ error). Instead, let's simply guard the `ExperimentalUsageChecker.checkCompilerArguments` call with `if (!analysisResult.isError())`.

Also, we'll need some kind of test to verify that this won't break in the future. @yanex could you suggest what's the right place for such a test? I suppose `plugins/kapt3/kapt3-base/test` would work?

tdn120:
Updated the code, but could use some advice on testing.  I don't really see a good entry point beyond just calling `analyzeAndReport()` with an errored `AnalysisResult` and confirming it returns with a proper report.
If that's good enough I can put that together.

udalov:
I meant to add an integration test which performs the whole compilation and running kapt with a processor such as the one in [KT-30172](https://youtrack.jetbrains.com/issue/KT-30172). It does look like tests in `plugins/kapt3/kapt3-base/test` are fairly elaborate, so you could try using one of them as an example. Without the test, we can't be sure that the issue is fixed. :)

tdn120:
Still working on that, but just to update, I was able to put together a similar test to the ones in `JavaKaptContextTest`, but it turns out that you can't differentiate between the old error (compiler crash) and new error (error message) - both come back with KaptBaseError.Kind.ERROR_RAISED.  Will probably be a few days before I can dig in further.

rybalkinsd:
Commented in YT, but will also past here:
Latest repro: 1.3.61
Fixed in: 1.3.70
Fixing commit https://github.com/JetBrains/kotlin/commit/958e6623fcefabf0f1f1267c40d254ded63614ee

Kapt uses `KaptWithoutKotlincTask` instead of `KaptWithKotlincTask` and make a graceful shutdown.
Setting `kapt.use.worker.api=true` fixes bug in 1.3.61.

All cases were tested with a free compiler arg `-Xuse-experimental=kotlin.contracts.ExperimentalContracts`

rybalkinsd:
Commented in YT, but will also past here:

Latest repro: 1.3.61
Fixed in: 1.3.70
Fixing commit https://github.com/JetBrains/kotlin/commit/958e6623fcefabf0f1f1267c40d254ded63614ee

Kapt uses KaptWithoutKotlincTask instead of KaptWithKotlincTask and make a graceful shutdown.
Setting kapt.use.worker.api=true fixes bug in 1.3.61.

All cases were tested with a free compiler arg `-Xuse-experimental=kotlin.contracts.ExperimentalContracts

udalov:
@tdn120 Is this change still needed for you? Can you confirm or deny that the solution for @rybalkinsd works for you?

tdn120:
Hmm I was assuming the solution would work, but now that I look, that seems to be a Gradle-specific plugin option.  Is there an analogue CLI option?  I tried a few guesses, but consistently get "unsupported plugin option" or "invalid argument".

udalov:
Sorry for the enormous delay. I finally got around to reproducing the issue, and this change indeed fixes it, for which I've added a test. The fix will be available in the first preview build of 1.7.0.

PR #3951, opened 2020-12-02
[IR] Align debugging of suspend lambdas with old BE

kandersen:
The existing backend restores LVs and parameters from the suspend lambda
fields used for spilling between suspension points, hence they are
visible in the debugger as local variables, plain and simple.

This PR introduces the same pattern to the IR backend, to bring the
debugging experience in line with the existing backend.

Both backends are still at the mercy of the liveness analysis
performed in the coroutine transformer where a liveness analysis
minimizes live ranges of entries in the LVT. E.g. an unused parameter
will be dropped entirely.

Adjusted existing test expectations accounting for the differences in
LV behavior.

### DIFF
--- a/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SuspendLambdaLowering.kt
+++ b/compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SuspendLambdaLowering.kt
@@ -5,12 +5,10 @@
 
 package org.jetbrains.kotlin.backend.jvm.lower
 
+import com.intellij.lang.jvm.source.JvmDeclarationSearch
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
 import org.jetbrains.kotlin.backend.common.IrElementTransformerVoidWithContext
-import org.jetbrains.kotlin.backend.common.ir.copyTo
-import org.jetbrains.kotlin.backend.common.ir.createImplicitParameterDeclarationWithWrappedDescriptor
-import org.jetbrains.kotlin.backend.common.ir.isSuspend
-import org.jetbrains.kotlin.backend.common.ir.moveBodyTo
+import org.jetbrains.kotlin.backend.common.ir.*
 import org.jetbrains.kotlin.backend.common.lower.LocalDeclarationsLowering
 import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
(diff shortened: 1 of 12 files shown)


ilmirus (approved):

compiler/ir/backend.jvm/src/org/jetbrains/kotlin/backend/jvm/lower/SuspendLambdaLowering.kt
```diff
@@ -209,19 +208,43 @@ private class SuspendLambdaLowering(context: JvmBackendContext) : SuspendLowerin
...
-                override fun visitGetValue(expression: IrGetValue): IrExpression {
-                    val parameter = (expression.symbol.owner as? IrValueParameter)?.takeIf { it.parent == irFunction }
-                        ?: return expression
-                    val field = fields[parameter.index + if (irFunction.extensionReceiverParameter != null) 1 else 0]
-                    val receiver = IrGetValueImpl(UNDEFINED_OFFSET, UNDEFINED_OFFSET, dispatchReceiverParameter!!.symbol)
-                    return IrGetFieldImpl(expression.startOffset, expression.endOffset, field.symbol, field.type, receiver)
-                }
-            }, null)
+            val localVals: List<IrVariable> = fields.mapIndexed { index, field ->
+                val varDescriptor = WrappedVariableDescriptor()
```
udalov, line 212: Maybe refactor this variable creation to `buildVariable`?

kandersen: Yup, happy to! Should I move it to that irFactory in the context?

udalov: I mean use [this existing function](https://github.com/JetBrains/kotlin/blob/5aaaa3881dfea4b58473dc3572be6f9ad220ea2c/compiler/ir/backend.common/src/org/jetbrains/kotlin/ir/builders/declarations/declarationBuilders.kt#L322) here, mainly to get rid of manual descriptor creation and binding

kandersen: Perfect, thanks! Also caught an "optimized import" in an unrelated file.

udalov:
Thanks!

PR #3954, opened 2020-12-03
KT-43745 Use UNICODE_CASE in String.replace to enable unicode-aware case check

kozaxinan:
Fixes KT-43745.

New String.replace implementation only works with `US-ASCII` characters. This is not same behaviour as before 1.4.20. It is also not same behaviour as Java. 

This changes add UNICODE_CASE to pattern check to enable unicode-aware case check.

Performance note.
I couldn't find a test to compare performance of new String.replace and old String.replace implementation and I couldn't verify the performance gain announcement in 1.4.20 release. Using UNICODE_CASE might affect the performance as mentioned it java doc of Pattern.UNICODE_CASE


ilya-g:
Thanks for the PR, but we decided to fix it differently: by removing the branch with `Pattern` at all.

I've used the test cases from your PR to improve our, though, see 149bcc2d22e21cc7ed99a23d21504573e3be559a

PR #3955, opened 2020-12-04
Remove wrapper class constructors

Sineaggi:
As per [JEP-390](https://openjdk.java.net/jeps/390), wrapper class constructors which have been deprecated since java 9, will now be marked deprecated for removal.

This will be interesting going forward, as this could mean older library might break when being run on a newer jdk.

### DIFF
--- a/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/boxing.kt
+++ b/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/boxing.kt
@@ -25,24 +25,24 @@ internal fun boxByte(primitive: Byte): java.lang.Byte = java.lang.Byte.valueOf(p
 
 @SinceKotlin("1.3")
 @PublishedApi
-internal fun boxShort(primitive: Short): java.lang.Short = java.lang.Short(primitive)
+internal fun boxShort(primitive: Short): java.lang.Short = java.lang.Short.valueOf(primitive) as java.lang.Short
 
 @SinceKotlin("1.3")
 @PublishedApi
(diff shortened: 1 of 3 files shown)


libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/boxing.kt
```diff
@@ -25,24 +25,24 @@ internal fun boxByte(primitive: Byte): java.lang.Byte = java.lang.Byte.valueOf(p
 
 @SinceKotlin("1.3")
 @PublishedApi
-internal fun boxShort(primitive: Short): java.lang.Short = java.lang.Short(primitive)
+internal fun boxShort(primitive: Short): java.lang.Short = java.lang.Short.valueOf(primitive) as java.lang.Short
```
ilmirus, line 28: These constructor calls are here for the reason: they enable sophisticated HotSpot optimizations in coroutines heavy code. In this case, HotSpot scalarizes primitive wrappers on fast-path, while on slow-path (i.e. suspension) it keeps the wrappers. Using `valueOf` breaks the optimization, slowing down coroutines code.
Besides, we do know about the warning, it has been there since JDK 9 and we deliberately use wrapper constructors for the reason above.

ilmirus (changes requested):

libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/boxing.kt
udalov: Yep, it's also explained in a comment at the top of this file.

PR #3956, opened 2020-12-04
Improve error parsing

YaPravaATyNet:
Motivation to create this pull request was fact that syntax errors can lead to weird semantic errors. I made some changes in parser to make this situations less frequent.

1. **Change for-expression parsing**
In the case when error occurred in "for" declaration it starts to parse loop body right after error token although usually next tokens are also error and they are not body. As a result there can be messages about non-existent semantic errors. E.g.:
`for (i in 1...3) {break;}`
Here among messages we can see _"'break' and 'continue' are only allowed inside a loop"_ despite of "break" being in loop body. 
Or if "in" is absent:
`for (i 1..3){}`
We get a message _"For is not an expression, and only expressions are allowed here"_ that is pretty unexpected and not very clear.
So i made some changes and now, if there is some error in "for" declaration, body will not be parsed until right parenthesis or braces appears. It allows to avoid messages like above.

2. **Type parsing in value argument list**
In value argument parsing there is already a case when the type for it is specified but only if it's simply type consisted from one token. E.g. 
`foo(a:Int) // error: Unexpected type specification`
But if the type is more complex:
`foo(a:kotlin.Any)`
we get much more error messages and there is _"Unresolved reference: Any"_ among them. Besides it makes parsing worse when value argument list is inside more complex constructions, e.g. annotations.
So i also added parsing for more complex types in this case.

### DIFF
--- a/compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
+++ b/compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
@@ -1841,8 +1841,17 @@ public class KotlinExpressionParsing extends AbstractKotlinParsing {
                 while (true) {
                     while (at(COMMA)) errorAndAdvance("Expecting an argument");
                     parseValueArgument();
-                    if (at(COLON) && lookahead(1) == IDENTIFIER) {
-                        errorAndAdvance("Unexpected type specification", 2);
+                    if (at(COLON) && (lookahead(1) == IDENTIFIER || lookahead(1) == LPAR)) {
+                        PsiBuilder.Marker type = mark();
+                        advance(); // COLON
+                        myKotlinParsing.parseTypeRef();
+                        int end = myBuilder.rawTokenIndex();
+                        type.rollbackTo();
+                        PsiBuilder.Marker err = mark();
+                        while (myBuilder.rawTokenIndex() < end) {
+                            advance();
+                        }
+                        err.error("Unexpected type specification");
                     }
                     if (!at(COMMA)) {
                         if (atSet(EXPRESSION_FIRST)) {
--- a/compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
+++ b/compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
@@ -1493,6 +1493,11 @@ public class KotlinExpressionParsing extends AbstractKotlinParsing {
                 error("Expecting a variable name");
             }
 
+            TokenSet recoverySet = TokenSet.create(RPAR, LBRACE, RBRACE, EOL_OR_SEMICOLON, EOF);
+            if (!atSet(recoverySet)) {
+                errorUntil("Unexpected tokens in loop", recoverySet);
+            }
+
             expectNoAdvance(RPAR, "Expecting ')'");
             myBuilder.restoreNewlinesState();
         }
--- a/compiler/testData/diagnostics/tests/incompleteCode/checkNothingIsSubtype.fir.kt
+++ b/compiler/testData/diagnostics/tests/incompleteCode/checkNothingIsSubtype.fir.kt
@@ -11,6 +11,6 @@ fun test(nothing: Nothing?) {
 }
 
 fun sum(a : IntArray) : Int {
-<!UNRESOLVED_REFERENCE, UNRESOLVED_REFERENCE, UNRESOLVED_REFERENCE!>for (n
-<!SYNTAX!>return<!><!SYNTAX!><!> "?"<!>
-}
+for (n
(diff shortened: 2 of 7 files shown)


compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
```diff
@@ -1836,8 +1841,17 @@ public void parseValueArgumentList() {
                 while (true) {
                     while (at(COMMA)) errorAndAdvance("Expecting an argument");
                     parseValueArgument();
-                    if (at(COLON) && lookahead(1) == IDENTIFIER) {
-                        errorAndAdvance("Unexpected type specification", 2);
+                    if (at(COLON) && (lookahead(1) == IDENTIFIER || lookahead(1) == LPAR)) {
```
PetukhovVictor, line 1844: I suggest removing checking `LPAR`: in `foo(x:)`,`Unexpected type specification` error may be confused because in fact, there isn't any type here.

PetukhovVictor, line 1844: BTW a type may not be started with `IDENTIFIER`. It can be a functional type, an annotated type or suspend functional type. See [the grammar](https://kotlinlang.org/docs/reference/grammar.html#types) for more details.

You can just try to parse a type after `:` and report `Unexpected type specification ` if it succeeded.

compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
```diff
@@ -1493,6 +1493,11 @@ private void parseFor() {
                 error("Expecting a variable name");
             }
 
+            TokenSet recoverySet = TokenSet.create(RPAR, LBRACE, RBRACE, EOL_OR_SEMICOLON, EOF);
+            if (!atSet(recoverySet)) {
+                errorUntil("Unexpected tokens in loop", recoverySet);
```
PetukhovVictor, line 1498: Minor: I believe `in "for" loop` would be clearer than `in loop`.

compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
PetukhovVictor: It seems, it can be simplified (in this version):
```
if (at(COLON) && (lookahead(1) == IDENTIFIER || lookahead(1) == LPAR)) {
    PsiBuilder.Marker err = mark();
    advance(); // COLON
    myKotlinParsing.parseTypeRef();
    err.error("Unexpected type specification");
}
```
But see the comment above.

compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
```diff
@@ -1493,6 +1493,11 @@ private void parseFor() {
                 error("Expecting a variable name");
             }
 
+            TokenSet recoverySet = TokenSet.create(RPAR, LBRACE, RBRACE, EOL_OR_SEMICOLON, EOF);
```
PetukhovVictor, line 1496: What about cases with chars from recovery set like `for (i in 1..3)) {break;}`? The strange semantic errors will still be reported there. I think we should report an error not only on chars not from the recovery set.

PetukhovVictor:
It would be fine if you add the separated PSI tests with given examples in the PR description.

PetukhovVictor:
Also, please advance `SOURCE_STUB_VERSION` (see `compiler/psi/src/org/jetbrains/kotlin/psi/stubs/KotlinStubVersions.kt`) because an error code with a type specification in calls may appear in signatures (for instance, as a delegation call). I'm not sure if it appears in stubs in the Kotlin world, but just in case it's better to advance the stub version.

compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
YaPravaATyNet: I don't fully understand what this means. It seems that it's unclear in general when to parse char from recovery set as _"unexpected tokens in 'for' loop"_. For example the case `for (i in 1..2) foo())`: foo() looks more like loop body with extra ')' and regard ') foo()' as error in 'for' declaration can be unexpected. And here` { for (i in 1..2 }` the try to parse '}' inside 'for' will break all parsing further. So should we handle  only case when there is char from recovery set and ')' right after it, e.g. `for (i in 1..2)) {}` or `for (i in 1..2;) {}`?

compiler/psi/src/org/jetbrains/kotlin/parsing/KotlinExpressionParsing.java
udalov: `in "the" for loop`

PR #3957, opened 2020-12-04
FIR: pass elvis expected type to lhs/rhs

jsjeon:
The motivation is bb test `kt6694...` where lhs/rhs of elvis expression has too specific type than expected type, which is specified as the enclosing function's return type. During body resolution, we had a chance to propagate the expected type in a top-down manner, like from function return type through function body to that elvis expression. It's just not passed to lhs/rhs at the last moment.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirControlFlowStatementsResolveTransformer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirControlFlowStatementsResolveTransformer.kt
@@ -216,9 +216,12 @@ class FirControlFlowStatementsResolveTransformer(transformer: FirBodyResolveTran
     ): CompositeTransformResult<FirStatement> {
         if (elvisExpression.calleeReference is FirResolvedNamedReference) return elvisExpression.compose()
         elvisExpression.transformAnnotations(transformer, data)
-        elvisExpression.transformLhs(transformer, ResolutionMode.ContextDependent)
+        val expectedArgumentType =
+            if (data is ResolutionMode.WithExpectedType && data.expectedType !is FirImplicitTypeRef) data
+            else ResolutionMode.ContextDependent
+        elvisExpression.transformLhs(transformer, expectedArgumentType)
         dataFlowAnalyzer.exitElvisLhs(elvisExpression)
-        elvisExpression.transformRhs(transformer, ResolutionMode.ContextDependent)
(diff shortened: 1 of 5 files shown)


compiler/testData/codegen/bytecodeText/boxingOptimization/maxMinByOrNull.kt
```diff
@@ -29,7 +29,7 @@ fun box(): String {
 // 12 ICONST_0
 // 1 IF_ICMPGE
 // 1 IF_ICMPLE
-// 4 LCMP
+// 2 LCMP
```
mglukhikh, line 32: This change will probably break IR / FE 1.0 codegen tests

mglukhikh:
The whole thing now looks very questionable. Will try to think more about it, when I have time.

compiler/testData/codegen/bytecodeText/boxingOptimization/maxMinByOrNull.kt
jsjeon: Uh, forgot that it's shared across other tests. I dug into what made this change, and found that only explicit, resolved type is worth being propagated to lhs/rhs. With that change, all the spec/diagnostic changes are gone. (Most of previous changes look correct to me, though.) Only changes about the bb test and IR text tests are remained as-is.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/transformers/body/resolve/FirControlFlowStatementsResolveTransformer.kt
```diff
@@ -216,9 +216,12 @@ class FirControlFlowStatementsResolveTransformer(transformer: FirBodyResolveTran
     ): CompositeTransformResult<FirStatement> {
         if (elvisExpression.calleeReference is FirResolvedNamedReference) return elvisExpression.compose()
         elvisExpression.transformAnnotations(transformer, data)
-        elvisExpression.transformLhs(transformer, ResolutionMode.ContextDependent)
+        val expectedArgumentType =
+            if (data is ResolutionMode.WithExpectedType && data.expectedType !is FirImplicitTypeRef) data
```
jsjeon, line 220: Not sure what makes things look questionable, but it wouldn't be a coincidence that top-down propagation of types (from explicit info of enclosing one to inner item) appears again at https://github.com/JetBrains/kotlin/pull/3968/commits/3815dca0a0298b3698a0ae1b911d8030de4266ac

mglukhikh:
Merged manually. After some consideration I decided the solution is Ok. 👍

PR #3958, opened 2020-12-04
Ant: Add support for fork-mode

scaventz:
Without support for fork-mode, a OOM error can be produced like below.  Since tasks ```java``` and ```javac``` both support fork-mode in Ant, it is better to have a fork-mode supported by the task ```kotlinc``` I think.
```
D:\Github-scaventz\lwjgl3\build.xml:204: java.lang.OutOfMemoryError: Java heap space
	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:109)
	at org.apache.tools.ant.Task.perform(Task.java:350)
	at org.apache.tools.ant.Target.execute(Target.java:449)
	at org.apache.tools.ant.Target.performTasks(Target.java:470)
	at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1388)
	at org.apache.tools.ant.Project.executeTarget(Project.java:1361)
	at org.apache.tools.ant.helper.DefaultExecutor.executeTargets(DefaultExecutor.java:41)
	at org.apache.tools.ant.Project.executeTargets(Project.java:1251)
	at org.apache.tools.ant.Main.runBuild(Main.java:834)
	at org.apache.tools.ant.Main.startAnt(Main.java:223)
	at org.apache.tools.ant.Main.start(Main.java:190)
	at org.apache.tools.ant.Main.main(Main.java:274)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.ant.execution.AntMain2.main(AntMain2.java:31)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.util.Arrays.copyOf(Arrays.java:3236)
	at java.io.ByteArrayOutputStream.toByteArray(ByteArrayOutputStream.java:191)
	at org.jetbrains.kotlin.preloading.ClassPreloadingUtils.loadAllClassesFromJars(ClassPreloadingUtils.java:146)
	at org.jetbrains.kotlin.preloading.ClassPreloadingUtils.preloadClasses(ClassPreloadingUtils.java:49)
	at org.jetbrains.kotlin.preloading.ClassPreloadingUtils.preloadClasses(ClassPreloadingUtils.java:70)
	at org.jetbrains.kotlin.ant.KotlinAntTaskUtil.getOrCreateClassLoader(KotlinAntTaskUtil.kt:60)
	at org.jetbrains.kotlin.ant.KotlinCompilerBaseTask.execute(KotlinCompilerBaseTask.kt:117)
	at org.jetbrains.kotlin.ant.Kotlin2JvmTask.execute(Kotlin2JvmTask.kt:85)
	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:99)
	... 16 more
```

### DIFF
--- a/ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
+++ b/ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
@@ -16,9 +16,12 @@
 
 package org.jetbrains.kotlin.ant
 
-import org.apache.tools.ant.types.Path
-import org.apache.tools.ant.types.Reference
+import org.apache.tools.ant.BuildException
+import org.apache.tools.ant.taskdefs.Execute
+import org.apache.tools.ant.taskdefs.Redirector
+import org.apache.tools.ant.types.*
 import java.io.File.pathSeparator
+import java.io.File.separator
 
 class Kotlin2JvmTask : KotlinCompilerBaseTask() {
     override val compilerFqName = "org.jetbrains.kotlin.cli.jvm.K2JVMCompiler"
@@ -27,6 +30,9 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
     var moduleName: String? = null
 
     var noReflect: Boolean = false
+
+    private val cmdl = CommandlineJava()
+    var fork: Boolean = false
 
     private var compileClasspath: Path? = null
 
@@ -72,5 +78,48 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
         if (noStdlib) args.add("-no-stdlib")
         if (noReflect) args.add("-no-reflect")
         if (includeRuntime) args.add("-include-runtime")
+    }
+
+    override fun execute() {
+        if (!fork)
+            super.execute()
+        else {
+            exec()
+        }
+    }
+
+    private fun exec() {
+        val javaHome = System.getProperty("java.home")
+        val javaBin = javaHome + separator + "bin" + separator + "java"
+        val redirector = Redirector(this)
+
+        fillArguments()
+
+        val command = ArrayList<String>()
+        command.add(javaBin)
+        command.addAll(cmdl.vmCommand.arguments) // jvm args
+        command.add("-Dorg.jetbrains.kotlin.cliMessageRenderer=FullPath") // same MessageRenderer as non-forking mode
+        command.add("-cp")
+        command.add(KotlinAntTaskUtil.compilerJar.canonicalPath)
+        command.add(compilerFqName)
+        command.addAll(args) // compiler args
+
+        // streamHandler: used to handle the input and output streams of the subprocess.
+        // watchdog: a watchdog for the subprocess or <code>null</code> to disable a timeout for the subprocess.
+        // TODO: support timeout for the subprocess
+        val exe = Execute(redirector.createHandler(), null)
+        exe.setAntRun(getProject())
+        exe.commandline = command.toTypedArray()
+        log("Executing command: ${command.joinToString(" ")}", LogLevel.DEBUG.level)
+        log("Compiling ${src!!.list().toList()} => [${output!!.canonicalPath}]")
+        val exitCode = exe.execute()
+        redirector.complete()
+        if (failOnError && exitCode != 0) {
+            throw BuildException("Compile failed; see the compiler error output for details.")
+        }
+    }
+
+    fun createJvmarg(): Commandline.Argument {
+        return cmdl.createVmArgument()
     }
 }
--- a/compiler/testData/integration/ant/jvm/fork/build.log.expected
+++ b/compiler/testData/integration/ant/jvm/fork/build.log.expected
@@ -1,0 +1,16 @@
+OUT:
+Buildfile: [TestData]/build.xml
+
+premain:
+    [javac] Compiling 1 source file to [Temp]
+      [jar] Building jar: [Temp]/premain.jar
+
+build:
+  [kotlinc] Compiling [[TestData]] => [[Temp]/fork.jar]
+  [kotlinc] -Xms64m
+  [kotlinc] -Xmx128m
+
+BUILD SUCCESSFUL
+Total time: [time]
+
+Return code: 0
--- a/compiler/testData/integration/ant/jvm/fork/build.xml
+++ b/compiler/testData/integration/ant/jvm/fork/build.xml
@@ -1,0 +1,25 @@
+<project name="Ant Task Test" default="build">
+    <taskdef resource="org/jetbrains/kotlin/ant/antlib.xml" classpath="${kotlin.lib}/kotlin-ant.jar"/>
+
+    <target name="premain">
+        <sequential>
+            <javac srcdir="${test.data}/premain" destdir="${temp}" includeAntRuntime="false"/>
+            <jar destfile="${temp}/premain.jar"
+                 basedir="${temp}"
+                 excludes="*.java *.jar">
+                <manifest>
+                    <attribute name="Premain-Class" value="Premain"/>
+                </manifest>
+            </jar>
+        </sequential>
+    </target>
+
+    <target name="build" depends="premain">
+        <kotlinc src="${test.data}" output="${temp}/fork.jar" fork="true">
+            <jvmarg value="-Xms64m"/>
+            <jvmarg value="-Xmx128m"/>
+            <jvmarg value="-javaagent:${temp}/premain.jar"/>
+            <compilerarg value="-nowarn"/>
+        </kotlinc>
+    </target>
+</project>
--- a/compiler/testData/integration/ant/jvm/forkOnError/build.xml
+++ b/compiler/testData/integration/ant/jvm/forkOnError/build.xml
@@ -1,0 +1,25 @@
+<project name="Ant Task Test" default="build">
+    <taskdef resource="org/jetbrains/kotlin/ant/antlib.xml" classpath="${kotlin.lib}/kotlin-ant.jar"/>
+
+    <target name="premain">
+        <sequential>
+            <javac srcdir="${test.data}/premain" destdir="${temp}" includeAntRuntime="false"/>
+            <jar destfile="${temp}/premain.jar"
+                 basedir="${temp}"
+                 excludes="*.java *.jar">
+                <manifest>
+                    <attribute name="Premain-Class" value="Premain"/>
+                </manifest>
+            </jar>
+        </sequential>
+    </target>
+
+    <target name="build" depends="premain">
+        <kotlinc src="${test.data}" output="${temp}/fork.jar" fork="true">
+            <jvmarg value="-Xms64m"/>
+            <jvmarg value="-Xmx128m"/>
+            <jvmarg value="-javaagent:${temp}/premain.jar"/>
+            <compilerarg value="-Werror"/>
+        </kotlinc>
+    </target>
+</project>
--- a/ant/src/org/jetbrains/kotlin/ant/KotlinCompilerBaseTask.kt
+++ b/ant/src/org/jetbrains/kotlin/ant/KotlinCompilerBaseTask.kt
@@ -80,7 +80,7 @@ abstract class KotlinCompilerBaseTask : Task() {
         fillSpecificArguments()
     }
 
-    final override fun execute() {
+    override fun execute() {
         fillArguments()
 
         val compilerClass = KotlinAntTaskUtil.getOrCreateClassLoader().loadClass(compilerFqName)
--- a/compiler/testData/integration/ant/jvm/fork/hello.kt
+++ b/compiler/testData/integration/ant/jvm/fork/hello.kt
@@ -1,0 +1,5 @@
+package hello
+
+fun main() {
+    println(0 as String)
+}
--- a/compiler/testData/integration/ant/jvm/fork/premain/Premain.java
+++ b/compiler/testData/integration/ant/jvm/fork/premain/Premain.java
@@ -1,0 +1,14 @@
+import java.lang.instrument.Instrumentation;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.util.List;
+
+public class Premain {
+    public static void premain(String agentArgs, Instrumentation inst) {
+        RuntimeMXBean bean = ManagementFactory.getRuntimeMXBean();
+        List<String> arguments = bean.getInputArguments();
+        arguments.stream().filter(param ->
+            param.startsWith("-Xmx") || param.startsWith("-Xms")
+        ).sorted().forEach(System.out::println);
+    }
+}
--- a/compiler/testData/integration/ant/jvm/forkOnError/build.log.expected
+++ b/compiler/testData/integration/ant/jvm/forkOnError/build.log.expected
@@ -1,0 +1,24 @@
+OUT:
+Buildfile: [TestData]/build.xml
+
+premain:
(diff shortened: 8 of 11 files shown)


udalov:
Can we add a test for this behavior? There are some existing integration tests in `compiler/tests-gen/org/jetbrains/kotlin/integration/AntTaskTestGenerated.java` with test data at `compiler/testData/integration/ant/`

onurberberoglu10 (commented):
Xx

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
```diff
@@ -73,4 +81,43 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
...
+        val javaHome = System.getProperty("java.home")
+        val javaBin = javaHome + File.separator + "bin" + File.separator + "java"
+        val classpath = System.getProperty("java.class.path")
+        val className = clazz.name
+
+        val command = ArrayList<String>()
+        command.add(javaBin)
+        command.addAll(jvmArgs)
+        command.add("-cp")
+        command.add(classpath + File.separator + compileClasspath.toString())
```
ilya-g, line 105: Did you mean `pathSeparator` here?

scaventz: Yes, the `compileClasspath.toString()` is `<pathelement location="${lib}/kotlinc/lib/kotlin-compiler.jar"/>` in my case.

scaventz:
> Can we add a test for this behavior? There are some existing integration tests in `compiler/tests-gen/org/jetbrains/kotlin/integration/AntTaskTestGenerated.java` with test data at `compiler/testData/integration/ant/`

Thanks for let me know this. A new test `testFork` has been added to AntTaskTestGenerated.java, it will print parameters of the forked JVM in `build.log.expected`

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
```diff
@@ -73,4 +81,42 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
         if (noReflect) args.add("-no-reflect")
         if (includeRuntime) args.add("-include-runtime")
     }
+
+    override fun execute() {
```
udalov, line 85: It seems that overriding `execute` resulted in the fork mode not having all the logic that was there, such as:
* filling different compiler arguments (`-nowarn`, `-verbose`, `-include-runtime`, `-module-name`, etc.)
* running the compiler with `execFullPathsInMessages` as opposed to the default mode when paths in compiler messages are relative
* logging what's being compiled
* throwing `BuildException` on compilation error

compiler/testData/integration/ant/jvm/fork/build.xml
```diff
@@ -0,0 +1,18 @@
+<project name="Ant Task Test" default="build">
```
udalov, line 1: Looks like you didn't commit the `build.log.expected` file

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
scaventz: Thanks for your view and pointed out those mistakes. 
I have made some modifications about you mentioned above. But I'm confused about 

> running the compiler with execFullPathsInMessages as opposed to the default mode when paths in compiler messages are relative

Did you mean MessageRenderer? I ran some tests, but I didn't see difference between `RelativePath MessageRenderer `and `FullPath MessageRenderer`. Is it should be MessageRenderer in fork mode?

Besides, I find the prefix "[kotlinc]" missing in my `build.log.expected`, trying to find out a solution about it.

scaventz: > Besides, I find the prefix "[kotlinc]" missing in my build.log.expected, trying to find out a solution about it.

This output format problem is solved.

scaventz:
@udalov  
Hi, sorry for disturbing, can you take a look at this when you available? 

I think I have fixed those problems as far as I can see (Except the `MessageRenderer` relevant stuff, because it seems there are no differences between `RelativePath MessageRenderer` and `FullPath MessageRenderer` in this case, `RelativePath MessageRenderer` is acceptable to me). 

And there are two tests now, one for compiling success, one for compiling error, both would print JVM args.

If this pull request is acceptable, I would like to continue work and create another PR to enhance it, by adding support timeout for the subprocess.

Thanks a lot.

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
udalov: Thanks! I meant that the compiler invoked from the Ant task uses full paths in the messages by default, and it's a bit weird that enabling `fork` changes that to relative paths.

However, it's not a big problem at all, and seeing that there's no API in the compiler for running it _externally_ with full paths in messages, I suppose we can leave it as is, with relative paths in the fork mode.

compiler/testData/integration/ant/jvm/forkOnError/build.xml
```diff
@@ -0,0 +1,19 @@
...
+    <target name="build">
+        <kotlinc src="${test.data}" output="${temp}/fork.jar" fork="true">
+            <jvmarg value="-Xms64m"/>
+            <jvmarg value="-Xmx128m"/>
+            <jvmarg value="-Xmn32m"/>
+            <jvmarg value="-XX:ReservedCodeCacheSize=64m"/>
+            <jvmarg value="-XX:CICompilerCount=4"/>
+            <jvmarg value="-XX:+PrintCommandLineFlags"/>
+
+            <compilerarg line="-jvm-target 1.8"/>
```
udalov, line 12: Looks like both in this and in the other test, we're not checking that the compiler arguments are properly applied.

I propose to do the following to improve that:
- in this test, use just one compiler argument `-Werror`, and change the error in the code to some warning. For example, replace `println1(...)` with `println(... as String)`.
- in the other test (`integration/ant/jvm/fork/`), change the code in the same way but use just one compiler argument `-nowarn`.

This will test not only the non-zero exit code and the error message text/paths, but also the fact that compiler arguments are applied.

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
```diff
@@ -73,4 +79,48 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
...
+        else {
+            exec()
+        }
+    }
+
+    private fun exec() {
+        val javaHome = System.getProperty("java.home")
+        val javaBin = javaHome + separator + "bin" + separator + "java"
+        val classpath = System.getProperty("java.class.path")
+        var redirector = Redirector(this)
```
udalov, line 95: Please fix the warning here and on `exitCode` below ("Variable is never modified and can be declared immutable using 'val'")

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
```diff
@@ -73,4 +79,48 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
...
+
+        fillArguments()
+
+        val command = ArrayList<String>()
+        command.add(javaBin)
+        command.addAll(cmdl.vmCommand.arguments) // jvm args
+        command.add("-cp")
+        command.add(classpath + pathSeparator + KotlinAntTaskUtil.compilerJar.canonicalPath)
```
udalov, line 103: Is `classpath` really necessary? If I understand correctly, it's the classpath of the currently-running Ant task, which is not needed for the compiler to work

udalov (changes requested):

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
scaventz: It seems a mistake and have been removed by now.

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
scaventz: Warnings have been fixed.

compiler/testData/integration/ant/jvm/forkOnError/build.xml
scaventz: Thanks for your suggestion! Now the tests look better.

ant/build.gradle.kts
```diff
@@ -17,5 +18,5 @@ sourceSets {
 }
 
 runtimeJar {
-    manifest.attributes["Class-Path"] = "$compilerManifestClassPath kotlin-preloader.jar"
+    manifest.attributes["Class-Path"] = "$compilerManifestClassPath kotlin-preloader.jar kotlin-compiler.jar"
```
scaventz, line 21: This modification is used to support of setting propertiy: ```org.jetbrains.kotlin.cliMessageRenderer=FullPath```. This property is used in ```CLITool.kt```, line 189, to get the instance ```MessageRenderer.PLAIN_FULL_PATHS```, which is the same MessageRenderer as non-forking mode.

udalov: It's nice that there actually is a way to select the message renderer! I didn't know that.

However, adding kotlin-compiler.jar to the classpath here is risky. At the very least, it can significantly slow down the loading time of the Ant task, since the compiler jar is very large. Let's just inline the constants from `MessageRenderer` instead, to avoid compile-time dependency on the compiler jar.

compiler/testData/integration/ant/jvm/fork/build.log.expected
```diff
@@ -0,0 +1,11 @@
+OUT:
+Buildfile: [TestData]/build.xml
+
+build:
+  [kotlinc] Compiling [[TestData]] => [[Temp]/fork.jar]
+  [kotlinc] -XX:CICompilerCount=4 -XX:InitialHeapSize=67108864 -XX:MaxHeapSize=134217728 -XX:MaxNewSize=33554432 -XX:NewSize=33554432 -XX:+PrintCommandLineFlags -XX:ReservedCodeCacheSize=67108864 -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:-UseLargePagesIndividualAllocation -XX:+UseParallelGC
```
udalov, line 6: On our buildserver, both tests fail because we don't have the `-XX:-UseLargePagesIndividualAllocation` argument. Can we do something about it?

Actually these arguments can be quite different on different JDKs, and to be honest, I don't see how to make this test JDK-independent... Maybe we'd need some other "custom" kind of test, which won't check for the exact match of the output with the `.expected` file, but will look for some specific substring in the output to verify that the JVM args have been applied correctly (e.g. like just `-XX:CICompilerCount=4` in this case)?

udalov: Or, as a dirty solution, we could change the output normalization logic in `AbstractAntTaskTest.normalizeOutput`, so that it would replace the string with a bunch of `-XX` arguments, printed by java, to a string containing only the known arguments?

So that this line:
```
-XX:CICompilerCount=4 -XX:InitialHeapSize=67108864 -XX:MaxHeapSize=134217728 -XX:MaxNewSize=33554432 -XX:NewSize=33554432 -XX:+PrintCommandLineFlags -XX:ReservedCodeCacheSize=67108864 -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:-UseLargePagesIndividualAllocation -XX:+UseParallelGC
```

would be transformed into something like:
```
-XX:CICompilerCount=4 [JavaXXArgs]
```

(similarly to how we replace test data path with `[TestData]`, time with `[time]`, etc.)

I don't like this though, since some JDKs can also print some -XX args before CICompilerCount.

Maybe there's some easier way to write this test (which is: to verify that JVM args have been applied) that I'm not seeing, I don't know.

udalov (changes requested):

compiler/testData/integration/ant/jvm/fork/build.log.expected
scaventz: Thanks for you suggestions, it is indeed a tricky problem. 

According to [Oracle's documentation](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/java.html) , it seems all ```Non-Standard Options```(start with -X) and ```Advanced Runtime Options```(start with -XX) are not guaranteed to be supported by all JVM implementations. 

So I reduced the number of parameters,  leaving only two parameters ```-Xms``` and ```-Xmx```, and hope common JVM implementations support them...

What do you think? is this feasible?

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
```diff
@@ -94,8 +93,8 @@ class Kotlin2JvmTask : KotlinCompilerBaseTask() {
         val javaBin = javaHome + separator + "bin" + separator + "java"
         val redirector = Redirector(this)
 
-        val msgRendererPropKey = MessageRenderer.PROPERTY_KEY
-        val msgRendererPropVal = MessageRenderer.PLAIN_FULL_PATHS.name
+        val msgRendererPropKey = "org.jetbrains.kotlin.cliMessageRenderer"
```
udalov, line 96: Please inline these variables

compiler/testData/integration/ant/jvm/fork/build.log.expected
udalov: Yeah, this looks like a great idea :). :+1:

ant/src/org/jetbrains/kotlin/ant/Kotlin2JvmTask.kt
scaventz: I'm afraid I don't quite understand what you mean, should I change it from 
```
val msgRendererPropKey = "org.jetbrains.kotlin.cliMessageRenderer"
val msgRendererPropVal = "FullPath"
command.add("-D$msgRendererPropKey=$msgRendererPropVal")
```
to
```
command.add("-Dorg.jetbrains.kotlin.cliMessageRenderer=FullPath")
```
?

udalov: Yes, exactly

udalov: (I'm saying "inline" because this is how the corresponding action is called in IntelliJ, see the menu -> Refactor -> _Inline_ property)

scaventz: Understood, thanks.

udalov (approved):

udalov:
Good job! Thanks for the contribution :).

PR #3960, opened 2020-12-05
FIR: extend suspend conversion to intersection type

jsjeon:
The motivation is bb test `coroutines/suspendConversion/intersectionTypeToSubtypeConversion.kt` where an argument with multiple type bounds could be passed to `suspend` function consumers each of which expects different yet part of bound types. This PR is kind of an extension of #3718 in the sense that it covers more sorts of types when searching for `invoke` symbol in the given type.


mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3961, opened 2020-12-05
Added String.replaceAfter() sample

zekehul:
Added a sample for `String.replaceAfter()` method


fzhinkin:
Unfortunately, the PR didn't get an attention before (sorry for that). I'm closing the PR as it seems to be out of date, feel free to reopen it.

PR #3962, opened 2020-12-05
FIR: keep nullability of lambda return type

jsjeon:
The motivation is bb test `inference/noCoercionToUnitWithEqualityConstraintForNullableReturnType` where the expected return type of the lambda is supposed to be `Unit?`, according to the parameter type, `() -> R?`. When a postponed lambda is analyzed, from the accumulated type constraints, we compute the expected type of lambda return type. There, when the raw return type has a constraint where it's upper or equal to `Unit`, we simply assume the expected type is `Unit`, but the nullability of the raw return type should have been considered as well.


mglukhikh (approved):

mglukhikh:
Merged manually 👍

PR #3964, opened 2020-12-07
Allow to inline anonymous objects from `init` block to several constructors

max-kammerer:
Allow to inline anonymous objects from `init` block to several constructors

Mostly it's a workaround to avoid assertion on generating `init` block in several constructors

#KT-42815 Fixed

### DIFF
--- a/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/codegen/ir/FirBlackBoxInlineCodegenTestGenerated.java
+++ b/compiler/fir/fir2ir/tests-gen/org/jetbrains/kotlin/codegen/ir/FirBlackBoxInlineCodegenTestGenerated.java
@@ -246,6 +246,16 @@ public class FirBlackBoxInlineCodegenTestGenerated extends AbstractFirBlackBoxIn
             runTest("compiler/testData/codegen/boxInline/anonymousObject/kt38197.kt");
         }
 
+        @TestMetadata("kt42815.kt")
+        public void testKt42815() throws Exception {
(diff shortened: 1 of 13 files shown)


max-kammerer:
@pyos Not sure that it's a best approach to avoid problem. Maybe the best one to distinguish `init` code from constructor one and pass null parent for `init` code but in such case it would be much more complicated //cc @udalov

pyos:
JvmLocalClassPopupLowering already has a condition to move classes out of lambdas inlined into `init` blocks to the class body. The problem is that `IrAnonymousInitializer` is not an `IrDeclarationParent`, so when there's some other declaration between the `init` and the lambda (like `val y` in KT-42815) that's the one that gets stored in `inlineLambdaToScope` and there's no way to detect that this declaration is itself inside an `init` block.

This was actually broken by 1e712f8 I think, because before then `IrElementVisitorVoidWithContext` only added certain kinds of declarations to the stack, while now `IrInlineReferenceLocator` overrides `visitDeclaration`.

max-kammerer:
>JvmLocalClassPopupLowering already has a condition to move classes out of lambdas inlined into init blocks to the class body. The problem is that IrAnonymousInitializer is not an IrDeclarationParent, so when there's some other declaration between the init and the lambda (like val y in KT-42815) that's the one that gets stored in inlineLambdaToScope and there's no way to detect that this declaration is itself inside an init block.

@pyos Thank you for pointing to this, I'm totally forget about it

max-kammerer:
Now it's fixed according to original approach (bug introduced in refactoring should gone away)

udalov (approved):

PR #3965, opened 2020-12-08
Introduce Int <-> Char conversion functions

qurbonzoda:


libraries/stdlib/src/kotlin/text/Char.kt
```diff
@@ -8,6 +8,72 @@
...
+ * Throws an exception if the [radix] is not in the range `2..36` or if this value is not less than the specified [radix].
+ *
+ * If this value is less than `10`, the decimal digit Char with code `'0'.code + this` is returned.
+ * Otherwise, the uppercase Latin letter with code `'A'.code + this - 10` is returned.
+ */
+public fun Int.digitToChar(radix: Int = 10): Char {
+    if (radix !in 2..36) {
+        throw IllegalArgumentException("Invalid radix: $radix. Valid radix values are in range 2..36")
+    }
+    if (this >= radix) {
```
ilya-g, line 67: What about negatives?

qurbonzoda: The check above (`radix !in 2..36`) excludes negative radix

ilya-g: I mean negative digit

libraries/stdlib/src/kotlin/text/Char.kt
```diff
@@ -8,6 +8,72 @@
...
+        throw IllegalArgumentException("Digit $this does not represent a valid digit in radix $radix")
+    }
+    return if (this < 10) {
+        Char(code = '0'.code + this)
```
ilya-g, line 71: `'0' + this` should work too.

qurbonzoda: done

libraries/stdlib/src/kotlin/text/Char.kt
qurbonzoda: done

qurbonzoda:
The review branch was renamed. PR has moved to https://github.com/JetBrains/kotlin/pull/3969

PR #3967, opened 2020-12-08
[JVM] Do not collapse unrelated locals in state machine transform

madsager:


madsager:
@ilmirus can you have a look at this. The first commit adds a local variable stepping test that shows the collapsing of unrelated locals in a simple setting. The second commit attempts to fix it by only ever extending LVT nodes if they correspond to the same original LVT node. Looks like the test added when this collapsing was put in place still passes.

ilmirus (approved):
Thanks!

PR #3968, opened 2020-12-08
FIR: no constraint for coerced-to-Unit expression in lambda

jsjeon:
The motivation is bb test `inference/coercionToUnitForLambdaReturnTypeWithFlexibleConstraint`:
```java
// FILE: TestJ.java

public class TestJ {
    public static <T> In<T> materialize() {
        return null;
    }
}
```
```kt
// FILE: test.kt

class In<in T>

fun <T> inferred(e: In<T>?, l: () -> T): T = l()

fun box(): String {
    // coercion to Unit for T!
    val inferred = (inferred(TestJ.materialize<Unit>(), { null })).toString()
    return if (inferred == "kotlin.Unit") "OK" else "fail : $inferred"
}
```
which currently fails with `"fail: null"` because the lambda `{ null }`, whose only expression (`null`) is supposed to be wrapped with coercion-to-Unit, still returns `null`. The lambda's return type and its block type should be `Unit` too, instead of `Nothing?`. To do so, we need a chain of changes as follows.

First, to change the lambda's return type as `Unit`, when analyzing the postponed lambda, we should be able to conclude that the expected return type of that lambda is `Unit`. In the collected type constraints, we were close but not quite yet: `ft<Unit?, Unit> <: T` is the constraint for the not fixed type. NI allows only simple type marker, like `Unit <: T`. The first commit extends it to flexible type. See the detailed message of the first commit.

Second, to change the block type, during expression resolution, we need top-down propagation---from the explicit type of the enclosing declaration to the block type---as well. Currently, only bottom-up propagation---from the type of the last expression of the block to the block type---is implemented. This also helps avoid adding redundant `return` expression with `Unit` type.

The last thing that bothers coercion-to-Unit insertion is the unnecessary type constraint from the argument resolution. After analyzing the postponed lambda, we visit return arguments to add more type constraints. However, `null`, which is of `Nothing?`, will be coerced to `Unit` (if no more type constraints are added), should have not visited. Otherwise, a constraint like `Nothing? <: T` will be added and change the block type and lambda return type again. In a similar manner, we don't constraint for the return expression of lambda if it is of `Unit` type. We should do so for the last expression which would be coerced to `Unit` without any constraints.

### DIFF
--- a/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/PostponedArgumentsAnalyzer.kt
+++ b/compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/PostponedArgumentsAnalyzer.kt
@@ -130,7 +130,7 @@ class PostponedArgumentsAnalyzer(
             rawReturnType,
             stubsForPostponedVariables
         )
-        applyResultsOfAnalyzedLambdaToCandidateSystem(c, lambda, candidate, results, ::substitute)
+        applyResultsOfAnalyzedLambdaToCandidateSystem(c, lambda, candidate, results, expectedTypeForReturnArguments, ::substitute)
         return results
     }
 
@@ -139,6 +139,7 @@ class PostponedArgumentsAnalyzer(
         lambda: ResolvedLambdaAtom,
         candidate: Candidate,
         results: ReturnArgumentsAnalysisResult,
+        expectedReturnType: ConeKotlinType? = null,
         substitute: (ConeKotlinType) -> ConeKotlinType = c.createSubstituteFunctorForLambdaAnalysis()
     ) {
         val (returnArguments, inferenceSession) = results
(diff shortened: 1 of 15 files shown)


compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/PostponedArgumentsAnalyzer.kt
```diff
@@ -164,21 +164,30 @@ class PostponedArgumentsAnalyzer(
 
         val checkerSink: CheckerSink = CheckerSinkImpl(candidate)
 
+        val lastExpression = lambda.atom.body?.statements?.lastOrNull() as? FirExpression
         var hasExpressionInReturnArguments = false
+        // Don't constraint for return expressions of lambda if it has Unit return type.
```
jsjeon, line 169: Better to be added with https://github.com/JetBrains/kotlin/commit/def647c0941769e3f7e08ecf242174f74c626e80#diff-5fb4206d83fbcd3e7c827c3b12541010eaf8de08ed8dd2bd470b993cfc8760d9

Anyway, this also indicates why `Unit` return needs a special handling.

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/PostponedArgumentsAnalyzer.kt
```diff
@@ -164,21 +164,30 @@ class PostponedArgumentsAnalyzer(
...
-                context = resolutionContext,
-                isReceiver = false,
-                isDispatch = false
-            )
+            // If it is the last expression, and the resolved type (via expected type) is Unit, that expression will be coerced to Unit.
+            // If the last expression is of Unit type, of course it's not coercion-to-Unit case.
+            val lastExpressionCoercedToUnit =
+                it == lastExpression && lambda.atom.returnTypeRef.coneType.isUnit && !it.typeRef.coneType.isUnit
+            // Don't constraint for the last expression of lambda if it will be coerced to Unit.
+            if (!lastExpressionCoercedToUnit) {
```
jsjeon, line 179: Here is the counterpart in old FE: https://github.com/JetBrains/kotlin/blob/master/compiler/resolution/src/org/jetbrains/kotlin/resolve/calls/components/PostponedArgumentsAnalyzer.kt#L159

demiurg906 (commented):
Hi @jsjeon. Thanks for PR, it's ok modulo one note. Will accept it after you fix it

compiler/fir/resolve/src/org/jetbrains/kotlin/fir/resolve/inference/PostponedArgumentsAnalyzer.kt
```diff
@@ -164,21 +164,30 @@ class PostponedArgumentsAnalyzer(
...
-                lambda.atom.returnTypeRef, // TODO: proper ref
-                checkerSink,
-                context = resolutionContext,
-                isReceiver = false,
-                isDispatch = false
-            )
+            // If it is the last expression, and the resolved type (via expected type) is Unit, that expression will be coerced to Unit.
+            // If the last expression is of Unit type, of course it's not coercion-to-Unit case.
+            val lastExpressionCoercedToUnit =
+                it == lastExpression && lambda.atom.returnTypeRef.coneType.isUnit && !it.typeRef.coneType.isUnit
```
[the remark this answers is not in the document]
jsjeon: Indeed, lambda return type is not inferred _yet_ here, and we don't want to constrain it if the last expression is going to be coerced-to-Unit (as we did for the inferred lambda return type if that inferred type is `Unit`). Otherwise, the last expression is revisited, and in the original example, `null` of `Nothing?` type will force the lambda return type and the block type again to be `Nothing?`.

Regarding `lambdaReturnType` variable, yeah, it's still type variable `T` in this example, since it's not fixed yet. But, again, we don't want to constrain it as `Nothing?`, since we will do coercion-to-Unit. To be clearer, I changed the call site to pass the explicit expected return type.

FYI, here is the counterpart in old FE: https://github.com/JetBrains/kotlin/blob/master/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/tower/KotlinResolutionCallbacksImpl.kt#L230 where `expectedReturnType` is indeed `expectedTypeForReturnArguments` (https://github.com/JetBrains/kotlin/blob/master/compiler/resolution/src/org/jetbrains/kotlin/resolve/calls/components/PostponedArgumentsAnalyzer.kt#L130), which matches with the change I add here.

demiurg906:
Merged manually

PR #3970, opened 2020-12-09
Avoid resolving deprecated configurations in tests

bingranl:
This PR enables more integration tests run with warning-mode=fail by
updating "resolveAllConfigurations" task in a way where we don't resolve
deprecated configurations and instead check dependencies on alternative
configurations.


libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/VariantAwareDependenciesIT.kt
```diff
@@ -287,13 +283,13 @@ class VariantAwareDependenciesIT : BaseGradleIT() {
...
-            gradleBuildScript().appendText("\ndependencies { compile 'org.jetbrains.kotlin:kotlin-compiler-embeddable' }")
+            gradleBuildScript().appendText("\ndependencies { implementation 'org.jetbrains.kotlin:kotlin-compiler-embeddable' }")
 
             testResolveAllConfigurations {
-                assertContains(">> :compile --> kotlin-compiler-embeddable-${defaultBuildOptions().kotlinVersion}.jar")
+                assertContains(">> :compileClasspath --> kotlin-compiler-embeddable-${defaultBuildOptions().kotlinVersion}.jar")
 
                 // Check that the transitive dependencies with 'runtime' scope are also available:
-                assertContains(">> :compile --> kotlin-script-runtime-${defaultBuildOptions().kotlinVersion}.jar")
+                assertContains(">> :runtimeClasspath --> kotlin-script-runtime-${defaultBuildOptions().kotlinVersion}.jar")
```
bingranl, line 292: @h0tk3y I am not sure about your intention of this test, could you double check this change?

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/VariantAwareDependenciesIT.kt
```diff
@@ -287,13 +283,13 @@ class VariantAwareDependenciesIT : BaseGradleIT() {
         // variants chosen over the runtime ones when resolving a configuration with no required Usage:
         with(Project("simpleProject")) {
             setupWorkingDir()
-            gradleBuildScript().appendText("\ndependencies { compile 'org.jetbrains.kotlin:kotlin-compiler-embeddable' }")
+            gradleBuildScript().appendText("\ndependencies { implementation 'org.jetbrains.kotlin:kotlin-compiler-embeddable' }")
```
h0tk3y, line 286: To keep this test still checking [the requirement](https://github.com/JetBrains/kotlin/pull/3970/files#diff-e4cb0de25baada0e7d2426d8854b5d3addb949a3f3ed109836e198117f4ff79dR281-R283), I think we need to add a custom resolvable configuration that extends from `implementation` and has no explicit `Usage` attribute.

libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/VariantAwareDependenciesIT.kt
h0tk3y: To keep this test still checking [the requirement](https://github.com/JetBrains/kotlin/pull/3970/files#diff-e4cb0de25baada0e7d2426d8854b5d3addb949a3f3ed109836e198117f4ff79dR281-R283), I think we need to add a custom resolvable configuration that extends from `implementation` and has no explicit `Usage` attribute.

bingranl: The link "this requirement" leads me this PR page, do you intend to share some other resource?

h0tk3y: Sorry, I meant these lines, the link to the line range from the diff viewer didn't work: https://github.com/JetBrains/kotlin/blob/d63f495fe7f867583008d04a4b11b383f3805f7b/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/VariantAwareDependenciesIT.kt#L281-L283
```
// Starting with Gradle 5.0, plain Maven dependencies are represented as two variants, and resolving them to the API one leads
// to transitive dependencies left out of the resolution results. We need to ensure that our attributes schema does not lead to the API
// variants chosen over the runtime ones when resolving a configuration with no required Usage:
```

h0tk3y: @bingranl Could you please adjust the test to follow this statement?

bingranl: Hi Sergey, I was on vacation and sorry for the late reply. I just uploaded a new patch to address your comments. Does that look good? Thanks!

bingranl:
Friendly ping.

h0tk3y:
The tests went fine, the branch is good to merge. Thanks a lot for the changes!

bingranl:
Thanks for your help!

